PR #641, opened 2022-05-24
Added publish api docs to release workflow

AlexejPenner:
## Describe changes
We could try getting the api docs fixed for this release already.

## Pre-requisites

## Types of changes


schustmi (commented):
The only issue I see with that is the following: With the existing solution we could backport doc changes to the release branch and the api docs were automatically rebuilt. This won't be possible with the new solution, do you think that's a concern or not something we should care about?

schustmi (approved):

PR #667, opened 2022-05-30
Integration test examples on kubeflow pipelines

AlexejPenner:
## Describe changes

Running with:
```pytest tests/integration/test_examples.py -s --on-kubeflow --use-virtualenv``` 

Will now run the chosen examples on kubeflow. This should manly happen from within the github actions and is less intended for local use.

All examples that rely on a configurable stack component get their secrets from the gh action secret manager. Some examples are still missing from this but should follow in the future (wandb_tracking, whylogs, step operator, cloud_secrets_manager).

These tests will be run when '''LTKF!' has been commented on a PR - on success a comment is left on the PR:

TestingQuo

## Pre-requisites

## Types of changes

### CODE REVIEW

**Comment by AlexejPenner** [comment]
LTKF!

**Comment by AlexejPenner** [comment]
LTKF!

File: .github/workflows/infrastructure_tests.yml [hunk 1]
```diff
@@ -0,0 +1,59 @@
+name: Test Examples on Remote Infrastructure
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [ opened, synchronize ]
```
schustmi, line 6: Could we make this test run only once by removing these triggers here?

examples/scipy/steps/importer/importer_step.py
```diff
@@ -0,0 +1,42 @@
...
+#  permissions and limitations under the License.
+import os
+import re
+
+import numpy as np
+import pandas as pd
+
+from zenml.steps import Output, step
+
+TRAIN_PATH = os.path.join(os.path.dirname(__file__), "../../data", "train.csv")
```
schustmi, line 22: This joining might not work on windows due to the explicit `/` in the path. Can we use `pathlib` or `os.path.join`/`os.path.dirname` to get the correct path on all operating systems?

.github/workflows/kubeflow.yml
```diff
@@ -24,20 +27,29 @@ jobs:
       # Workaround from FuseML (https://github.com/fuseml/fuseml/blob/main/.github/workflows/ci.yml)
       # as the TF images are too large for the GH action runner disk
       - name: Free disk space
+        shell: bash
```
schustmi, line 30: Any reason that the android removal was deleted below?

.github/workflows/kubeflow.yml
```diff
@@ -24,20 +27,29 @@ jobs:
...
           echo
           echo "Available storage:"
           df -h
 
+      - name: Install Linux System Dependencies
+        if: runner.os=='Linux'
+        run: sudo apt install graphviz
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
schustmi, line 47: Just out of curiosity, how can this work without using any secrets to authenticate? Or are those secret keys hardcoded somewhere and just not visible in this yaml file?

.github/workflows/kubeflow.yml
```diff
@@ -50,21 +62,33 @@ jobs:
...
+          pip install notebook==6.4.11
 
-      - name: Install Linux System Dependencies
-        if: runner.os=='Linux'
-        run: sudo apt install graphviz
+      - name: Perform docker login and get Kubecontext
+        run: bash scripts/setup_kubeflow_aws_connection.sh
 
       - name: Build docker image
         # this step can be much faster if we use an image that copies the
```
schustmi, line 71: I'm guessing the current image now uses the lock file, can we remove the comment?

.github/workflows/kubeflow.yml
```diff
@@ -50,21 +62,33 @@ jobs:
...
         # lockfile that was generated in the previous step
-        run: docker build -t zenml-base-image:latest -f docker/base-dev.Dockerfile .
+        run: docker build -t test-base-image:latest -f docker/base-test.Dockerfile .
+
+      - name: Install Prerequisites
+        run: |
+          source $VENV
+          zenml integration install s3 kubeflow \
+            evidently facets huggingface mlflow tensorflow sklearn xgboost lightgbm slack neural_prophet -f
```
schustmi, line 79:
```suggestion
            evidently facets huggingface mlflow tensorflow sklearn xgboost lightgbm slack neural_prophet -y
```

tests/integration/test_examples.py
```diff
@@ -15,22 +15,39 @@
...
-from typing import Callable, NamedTuple, Optional
+from typing import Callable, List, Optional
 
 import pytest
+from pydantic import BaseModel
 
-from zenml.cli import EXAMPLES_RUN_SCRIPT, SHELL_EXECUTABLE, LocalExample
+from zenml.cli import set_active_stack
+
+# from zenml.integrations.slack.alerters import SlackAlerter
```
schustmi, line 30: Can be removed I guess?

tests/integration/test_examples.py
```diff
@@ -175,31 +246,38 @@ def test_run_example(
...
-        # instead windows tries to delete immediately and fails with a
-        # PermissionError: [WinError 32] The process cannot access the
-        # file because it is being used by another process
-        logging.debug(
-            "Skipping deletion of temp dir at teardown, due to "
-            "Windows Permission error"
-        )
+        time.sleep(1)
+        # Validate the result
+        # example_configuration.assert_successful(repo)
```
schustmi, line 265: Should this be in?

tests/integration/test_examples.py
```diff
@@ -175,31 +246,38 @@ def test_run_example(
...
+    except Exception as e:
+        pass
+    finally:
+        # clean up
+        try:
+            # subprocess.check_call(["docker", "system", "prune", "-a", "-f"])
```
schustmi, line 271: Should this be in?

tests/integration/test_examples.py
```diff
@@ -46,102 +63,153 @@ def copy_example_files(example_dir: str, dst_dir: str) -> None:
...
 
     name: str
-    validation_function: Callable[[Repository], None]
-    setup_function: Optional[Callable[[Repository], None]] = None
-    skip_on_windows: bool = False
+    runs_on_windows: bool
+    required_stack_components: List[StackComponent] = list()
+    pipeline_name: str
+    pipeline_path: str
+    step_count: Optional[int]
```
schustmi, line 78:
```suggestion
    step_count: Optional[int] = None
```

schustmi (changes requested):

tests/integration/conftest.py
```diff
@@ -46,40 +53,78 @@ def shared_kubeflow_profile(
...
+        t = threading.Thread(target=daemon_function, args=args, kwargs=kwargs)
+        t.setDaemon(True)
+        t.start()
+
+        # As we are creating a daemon thread instead of a real daemon process,
+        #  no pid file is created, so we now need to pretend the pid file exists
+        def fake_pid_file_exists(pid_file: str):
+            return True
+
+        module_mocker.patch(
```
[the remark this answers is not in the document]
AlexejPenner: Yes, this ensures, that the method is patched only as from the second time that it is called. For the first call it should return false so that the daemon thread can run and pretend to have written a pid file

tests/integration/conftest.py
```diff
@@ -117,7 +162,7 @@ def clean_kubeflow_profile(
     Yields:
         An empty repository with a provisioned local kubeflow stack.
     """
-    cleanup_active_profile()
+    # cleanup_active_profile()
```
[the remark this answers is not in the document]
AlexejPenner: yes, been commented in now

.github/workflows/infrastructure_tests.yml
AlexejPenner: yes, once everything else works well together

docker/base-test.Dockerfile
```diff
@@ -0,0 +1,51 @@
+FROM ubuntu:20.04
```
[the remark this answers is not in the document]
AlexejPenner: Actually I am not sure myself, didn't have an in-depth look at this

examples/scipy/steps/__init__.py
```diff
@@ -11,7 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 #  or implied. See the License for the specific language governing
 #  permissions and limitations under the License.
-from .loader.loader_step import importer
+from .importer.importer_step import importer
```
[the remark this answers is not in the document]
AlexejPenner: I was under the impression that most (if not all examples) use relative imports here, at least coudln't find one that doesn't

tests/integration/conftest.py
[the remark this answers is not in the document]
schustmi: Ah that makes sense, thanks! I have one follow up-question, to explain let me quickly describe how I think the tests are running currently (and correct me if I'm wrong here, as then the question probably doesn't make sense 😄):
- All the examples get "started" from the same python process, which uses the module-level fixture `shared_kubeflow_profile`.
- The example copies and updates the stack with additional components that it needs.
- The example gets run and the copied stack etc get deleted.

Now to my questions: Do we call `zenml stack up` or the code equivalent at some point after copying and updating the stack to make sure the newly added components also get a chance to work?

If we don't call it: Should we?
If we do call it: Due to using a `module_mocker` to patch the logic of writing the PID file, now the thing you described above does not work anymore as it's already patched right?

examples/scipy/steps/__init__.py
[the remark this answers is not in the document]
schustmi: You're completely right, I was somehow under the impression that we always used absolute imports. I just checked and it seems like
- some examples import nothing in these init files
- kubernetes and the quickstart do absolute imports
- the rest does relative imports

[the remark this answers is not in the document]
AlexejPenner: if we want to do this uniformly I would capture it in a separate ticket if thats ok with you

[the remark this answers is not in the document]
schustmi: Sounds good

PR #691, opened 2022-06-09
Sane defaults for MySQL

htahir1:
## Describe changes
I added sane defaults for mysql and more docs.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/metadata_stores/mysql_metadata_store.py
+++ b/src/zenml/metadata_stores/mysql_metadata_store.py
@@ -26,10 +26,10 @@ from zenml.repository import Repository
 class MySQLMetadataStore(BaseMetadataStore):
     """MySQL backend for ZenML metadata store."""
 
-    username: Optional[str] = None
+    port: int = 3306
+    username: Optional[str] = "root"
     password: Optional[str] = None
     host: str
-    port: int
     database: str
 
     secret: Optional[str] = None


AlexejPenner (approved):
:1st_place_medal:

src/zenml/metadata_stores/mysql_metadata_store.py
```diff
@@ -26,10 +26,10 @@
 class MySQLMetadataStore(BaseMetadataStore):
     """MySQL backend for ZenML metadata store."""
 
-    username: Optional[str] = None
+    port: int = 3306
+    username: Optional[str] = "root"
     password: Optional[str] = None
```
fa9r, line 31:
```suggestion
    password: Optional[str] = ""
```

fa9r (approved):
LGTM. Would also add default `password=""`.

src/zenml/metadata_stores/mysql_metadata_store.py
htahir1: In this case might make sense to have a default database name as well?

schustmi (changes requested):
Actually I change my approval after just having worked with the MySQL secret schema. We throw an error if the username is defined both in the stack component as well as the secret, so I don't think we should set a default value for either `username` or `password` here.

schustmi (approved):

PR #695, opened 2022-06-09
Stack Component registering made easier

AlexejPenner:
## Describe changes
Multiple times today I was registering stack components with flavors that I was unfamiliar with. Going into the docs or code takes quite a bit of time, so I simply registered them and waited for the error message to show me which fields I was missing.

Old

In this Pull request I introduce code that will enter interactive mode, in case a required field is not set in the first command. Now it will look a bit like this:

New

The only disadvantage that I can see in this, is that the reverse-i-search will become less useful to figure out which command was run to register a given stack component. Looking forward to your comments.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/stack_components.py
+++ b/src/zenml/cli/stack_components.py
@@ -340,7 +340,7 @@ def generate_stack_component_register_command(
         flavor: str,
         old_flavor: str,
         interactive: bool,
-        args: List[str]
+        args: List[str],
     ) -> None:
         """Registers a stack component.
 
@@ -421,7 +421,7 @@ def generate_stack_component_register_command(
                         user_input = click.prompt(f"{field}")
                     else:
                         prompt = f"{field} (Optional)'"
-                        user_input = click.prompt(prompt, default='')
+                        user_input = click.prompt(prompt, default="")
                     if user_input:
                         completed_fields[field] = user_input
 
--- a/src/zenml/cli/stack_components.py
+++ b/src/zenml/cli/stack_components.py
@@ -296,7 +296,7 @@ def _register_stack_component(
 
 def generate_stack_component_register_command(
     component_type: StackComponentType,
-) -> Callable[[str, str, str, List[str]], None]:
+) -> Callable[[str, str, str, bool, List[str]], None]:
     """Generates a `register` command for the specific stack component type.
 
     Args:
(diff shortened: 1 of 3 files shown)


htahir1:
I added @bcdurak and @schustmi as reviewers to get their opinions on this. I am personally unsure myself whether this is a good default way or not. For me the error seems clear what to add and id rather not go in interactive mode

AlexejPenner:
> I added @bcdurak and @schustmi as reviewers to get their opinions on this. I am personally unsure myself whether this is a good default way or not. For me the error seems clear what to add and id rather not go in interactive mode

What downside do you see in going interactive?

htahir1:
In CI/CD settings, when I want to automate the construction of the stack in as script, I want it to fail gracefully rather than go in interactive mode and hang my machine :-)

schustmi:
Love the idea, super useful to register unknown components. 

I also agree with @htahir1 that the "defaulting" to interactive might not be a good idea for any automated script, in that case you want it to just fail. Can we instead add this functionality with a `-i/--interactive` flag similar to the secret registration? We could even include optional attributes as well to make it a full-featured registration. (Btw we should definitely add this for stacks as well!)

AlexejPenner:
> Love the idea, super useful to register unknown components.
> 
> I also agree with @htahir1 that the "defaulting" to interactive might not be a good idea for any automated script, in that case you want it to just fail. Can we instead add this functionality with a `-i/--interactive` flag similar to the secret registration? We could even include optional attributes as well to make it a full-featured registration. (Btw we should definitely add this for stacks as well!)

Have implemented this and pushed :)

src/zenml/cli/stack_components.py
```diff
@@ -361,28 +374,73 @@ def register_stack_component_command(
...
+                }
+
+                completed_fields = parsed_args.copy()
+                for field, field_req in missing_fields.items():
+                    if field_req:
+                        user_input = click.prompt(f"{field}")
+                    else:
+                        prompt = f"{field} (Optional)'"
+                        user_input = click.prompt(prompt, default="")
+                    if user_input:
```
fa9r, line 425: I think this might not work if you want to set the empty string as a parameter somewhere. E.g., for mysql `password=""`. However, I also can't think of any way how this can be supported without screwing other things up, so LGTM! 😅

fa9r (approved):
Awesome functionality, looking forward to have this in dev so I can start using it myself 👍

schustmi (approved):

safoinme (approved):

PR #705, opened 2022-06-14
Add stack component copy command

schustmi:
## Describe changes
This PR adds functionality to copy an existing `StackComponent`.

Other small changes:
- The`zenml stack copy` command now doesn't load the actual stack components anymore and is therefore much quicker
- The `zenml stack-component down` command `--yes` flag is updated to `--force` to be inline with `zenml stack down`

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/stack.py
+++ b/src/zenml/cli/stack.py
@@ -1045,32 +1045,56 @@ def import_stack(
 
 @stack.command(
     "copy",
-    help="SOURCE_NAME: The name of the stack to copy.\n\nTARGET_NAME: Name of the copied stack.",
+    help="SOURCE_STACK: The name of the stack to copy.\n\nTARGET_STACK: Name of the copied stack.",
 )
-@click.argument("source_name", type=str, required=True)
-@click.argument("target_name", type=str, required=True)
-def copy_stack(source_name: str, target_name: str) -> None:
+@click.argument("source_stack", type=str, required=True)
+@click.argument("target_stack", type=str, required=True)
+@click.option(
+    "--from",
+    "source_profile",
+    type=str,
+    required=False,
+    help="The profile from which to copy the stack.",
+)
+def copy_stack(
+    source_stack: str, target_stack: str, source_profile: Optional[str] = None
+) -> None:
     """Copy a stack.
 
     Args:
-        source_name: The name of the stack to copy.
-        target_name: Name of the copied stack.
+        source_stack: The name of the stack to copy.
+        target_stack: Name of the copied stack.
+        source_profile: Name of the profile to copy from.
     """
     track_event(AnalyticsEvent.COPIED_STACK)
 
-    with console.status(f"Copying stack `{source_name}`...\n"):
-        repo = Repository()
+    repo = Repository()
+    if source_profile:
         try:
-            stack_ = repo.get_stack(source_name)
+            profile = GlobalConfiguration().profiles[source_profile]
         except KeyError:
             cli_utils.error(
-                f"Stack `{source_name}` cannot be copied as it does not exist."
+                f"Unable to find source profile '{source_profile}'."
+            )
+    else:
+        profile = repo.active_profile
+
+    with console.status(f"Copying stack `{source_stack}`...\n"):
+        try:
+            # Use a different `Repository` here which is configured to read from
+            # the source profile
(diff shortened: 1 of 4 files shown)


src/zenml/cli/stack.py
```diff
@@ -1045,32 +1045,56 @@ def import_stack(
...
-def copy_stack(source_name: str, target_name: str) -> None:
+@click.argument("source_stack", type=str, required=True)
+@click.argument("target_stack", type=str, required=True)
+@click.option(
+    "--from",
+    "source_profile",
+    type=str,
+    required=False,
+    help="The profile from which to copy the stack.",
+)
```
strickvl, line 1058: Would it be too confusing to have a target_profile option as well? So that if you're in one profile but you realise you need to copy your stack to somewhere else, then you can do it directly from there. This way, it feels like maybe it's forcing someone to switch profiles first, which they might not want to do?

src/zenml/cli/stack_components.py
```diff
@@ -828,6 +830,92 @@ def delete_stack_component_command(name: str) -> None:
...
+
+    @click.argument("source_component", type=str, required=True)
+    @click.argument("target_component", type=str, required=True)
+    @click.option(
+        "--from",
+        "source_profile",
+        type=str,
+        required=False,
+        help=f"The profile from which to copy the {display_name}.",
+    )
```
strickvl, line 854: Same question here about source and target profiles...

strickvl (changes requested):
I had a couple of implementation questions, and I think it also needs `cli/__init__.py` to be updated as well.

src/zenml/cli/stack.py
schustmi: Mhm I guess not, I'll add it for both commands. Do you have any naming suggestions for these options?

`zenml stack copy SOURCE_NAME TARGET_NAME --from SOURCE_PROFILE --to TARGET_PROFILE` 

Or are `--from/to` to ambiguous as we essentially have 2 sources and 2 targets?

src/zenml/cli/stack.py
```diff
@@ -1045,32 +1045,56 @@ def import_stack(
...
-            stack_ = repo.get_stack(source_name)
+            profile = GlobalConfiguration().profiles[source_profile]
+        except KeyError:
+            cli_utils.error(
+                f"Unable to find source profile '{source_profile}'."
+            )
+    else:
+        profile = repo.active_profile
+
+    with console.status(f"Copying stack `{source_stack}`...\n"):
```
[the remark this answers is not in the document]
schustmi: Yes that is handled by the call to `repo.register_stack(...)` which also registers all missing components (Unless there's a conflict)

src/zenml/cli/stack.py
strickvl: To start with, the defaults should be the active component / stack and the active profile.

`zenml stack copy default new_default` (takes from current profile and copies into current profile)

`zenml stack copy default new_default --from default --to not_default` (when you have multiple profiles involved?

Not sure if the  default / not default stuff is worth the hassle. Maybe we should just make people specify everything? That *is* pretty annoying, though...

schustmi: Default profile for source/target profile is quite easy to implement. However would you also expect `zenml stack copy new_Name` to work? That would default to the active stack as the source, but would be a little confusing imo

strickvl: yeah defaulting to the active stack was how I'd thought. would be same with the profiles where you'd default to the active profile, no?

In any case, I agree that it's a bit confusing. Users shouldn't have to remember this stuff.

What'd be really nice now would be that CLI functionality @bcdurak showed off a few weeks back. But anyway, maybe for now we should just require people to be explicit.

schustmi: I think with profiles it's a little less confusing as it's defined as options, but calling a `copy ...` command with one argument seems a little unclear what it does.
`zenml stack copy test` could refer to both:
- copy the `test` stack and assign it a name like `test_1` automatically
- copy the active stack and name it `test`

I'd probably say we should keep those required, and make both the source and target profile default to the current profile. Would that be acceptable for you?

strickvl: Yes great

Issue #600, opened 2022-05-17
[FEATURE]: Add `ignored_columns` to Evidently StandardStep to filter dataframe columns


strickvl:
### Contact Details [Optional]

support@zenml.io

### Describe the feature you'd like

ZenML currently implements a way to detect drift using [Evidently](https://evidentlyai.com/). We created a standard step that can be used to generate and access Evidently's core visualisations here: `zenml/src/zenml/integrations/evidently/steps/evidently_profile.py`. (Learn more about Evidently and drift detection by checking out our example [here](https://github.com/zenml-io/zenml/tree/main/examples/drift_detection).

The entrypoint function for the `EvidentlyProfileStep` currently takes a reference dataset and a comparison dataset. We would like to add an extra optional argument where users could pass in `ignored_columns` (a List of strings, most likely) which would then be ignored when making the drift comparison.

### How do you solve your current problem with the current status-quo of ZenML?

Currently you would have to do the pre-processing in a separate step or process (i.e. removing the specific columns).

### Any other comments?

This is related to #ENG-328 (Jira).

ketangangal:
Hey @strickvl 
 
 While passing reference dataset and a comparison dataset, if you keep ref columns equal to comparison columns then there is no point of adding ignored_columns . User has to manage this scenario. I can't see any specific use case for it. Please let me know more about your thoughts. I would like to contribute.
 
 Thanks,

htahir1:
@ketangangal What about the case when you want to calculate drift on not all columns but only certain ones? For example if you have a mixed dataset and only want to focus on the numerical ones? I think there is a use-case for it

PR #706, opened 2022-06-14
Ignore columns for drift detection

SangamSwadiK:
## Describe changes
Added ignored_columns to Evidently StandardStep to filter dataframe columns, 
to enable drift detection on selected columns.
- This fixes the **(Issue : #600 )**

## Pre-requisites

## Types of changes

@strickvl Hi, Submitting another PR, on the same issue.


strickvl:
Thank you for this. It would be useful to have some tests for this addition so that we know it's doing what it's supposed to do. Would you be able to add those?

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -127,6 +127,7 @@ def entrypoint(  # type: ignore[override]
         reference_dataset: pd.DataFrame,
         comparison_dataset: pd.DataFrame,
         config: EvidentlyProfileConfig,
+        ignored_columns=Union[None, List[str]],
```
strickvl, line 130: This might better be expressed as `Optional[List[str]]`. I think they're equivalent?

SangamSwadiK: > This might better be expressed as `Optional[List[str]]`. I think they're equivalent?

Yes you are right, I updated it.

SangamSwadiK:
> Thank you for this. It would be useful to have some tests for this addition so that we know it's doing what it's supposed to do. Would you be able to add those?

Hi, I would like to add some tests, but I'm not sure where to do it. I did go through the `tests/integration/test_examples.py` and `tests/integration/example_validation.py` and the below in `tests/integration/test_examples.py`
```
    ExampleIntegrationTestConfiguration(
        name="evidently_drift_detection",
        validation_function=drift_detection_example_validation,
    )
```
If you can please point me in the right direction, I would be able to add the tests.
Thanks.

SangamSwadiK:
> I think at a minimum we'd want some unit tests to ensure that the ignored columns were really being ignored. I would maybe ignore the actual 'integration' tests (of the integration) since we are currently reworking them.
> 
> (Flagging @AlexejPenner so we remember to add testing for this as part of the new testing suite.)

Got it. Thanks for clarification. I'll write the unit tests at `zenml/tests/unit/integrations/evidently` submit a PR.

SangamSwadiK:
@strickvl  Hi, I have a noob question. I've written the tests, but I'm facing a type issue and getting a subclass error, its expecting the **ignored_columns** argument to be of **type class**, **Optional[List[str]] gives a subclass error**, however **List[str]** gets passed.

The below works
```
    def entrypoint(  # type: ignore[override]
        self,
        reference_dataset: pd.DataFrame,
        comparison_dataset: pd.DataFrame,
        config: EvidentlyProfileConfig,
        ignored_columns:List[str],
```
But changing **List[str]** to **Optional[List[str]]** (which is correct) results in typing.Union type which gives a 
**subclass error : arg1 must be of type class.**
I tried using dataclasses, but that too gave a typing.Union.
Any suggestions will be helpful.

Thanks.

htahir1:
@SangamSwadiK seems like you now used an email that hasnt signed the CLA yet. Would you mind doing it?

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -137,13 +138,48 @@ def entrypoint(  # type: ignore[override]
...
 
         Returns:
             profile: dictionary report extracted from an Evidently Profile
               generated for the data drift
             dashboard: HTML report extracted from an Evidently Dashboard
               generated for the data drift
         """
+        if ignored_columns is not None:
+            if not (isinstance(ignored_columns, list)):
+                raise ValueError(
```
strickvl, line 156: I think a `TypeError` would be more appropriate here.

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -137,13 +138,48 @@ def entrypoint(  # type: ignore[override]
...
+                    f"Expects a list of features but got type {type(ignored_columns)}"
```
strickvl, line 157:
```suggestion
                    f"Expects a list of columns but got type '{type(ignored_columns)}'."
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -137,13 +138,48 @@ def entrypoint(  # type: ignore[override]
...
+                )
+
+            if len(ignored_columns) == 0:
+                raise ValueError(
+                    "Expects a list of features or None but got empty list"
+                )
+
+            if sorted(set(ignored_columns)) != sorted(ignored_columns):
+                raise ValueError("Duplicate features found")
```
strickvl, line 166: I think maybe these two could be ignored. 

- For when the list of features is empty, then basically we will do nothing, so no need to specify a minimum number of items, I think.
- I think we can handle any duplication on our end, perhaps, so that we don't raise unnecessarily large numbers of errors here. (We could remove duplicates from the array, or convert it to a set and then back to a list etc etc.

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -137,13 +138,48 @@ def entrypoint(  # type: ignore[override]
             comparison_dataset: a Pandas DataFrame of new data you wish to
                 compare against the reference data
             config: the configuration for the step
+            ignored_columns: columns to ignore while drift detection
```
strickvl, line 141:
```suggestion
            ignored_columns: columns to ignore when creating Evidently profile
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -137,13 +138,48 @@ def entrypoint(  # type: ignore[override]
...
+            if sorted(set(ignored_columns)) != sorted(ignored_columns):
+                raise ValueError("Duplicate features found")
+
+            if not (
+                set(ignored_columns).issubset(set(reference_dataset.columns))
+            ) or not (
+                set(ignored_columns).issubset(set(comparison_dataset.columns))
+            ):
+                raise ValueError(
+                    "Feature not found in reference/comparison datasets"
```
strickvl, line 174:
```suggestion
                    "Column name is not found in reference/comparison datasets."
```

strickvl (changes requested):
Thanks for these new additions. Especially thanks for the neat tests! I made some comments on the implementation so I didn't look too closely at the tests for now since I think those changes might break some of the tests, but otherwise I think you're on the right track for the most part!

schustmi:
I haven't taken a look at the entire PR, but just quickly glanced over the comments and saw the discussion about an empty list as a default argument to a function. In general mutable default values (e.g. an emtpy list `[]` or an emtpy dict `{}`) as function arguments should be avoided in python, and I'm pretty sure mypy will also complain about that.

The reason for this is that it often leads to unwanted behaviour, like for example this:
```python
def append(value, list_: List = []):
    list_.append(value)
    return list_
    
# the first time this works as expected, appending the value 3 to an empty list
print(append(3))  # [3]

# the second time, python uses the same instance of the list which already 
# contains elements from the previous function call
print(append(4))  # [3, 4]
```

I think there are two main ways to avoid this:
- Use an emtpy tuple (immutable) as default value:
    ```python
    def f(value: Tuple = ()):
        ...
    ```
- Make the argument Optional and set the default value inside the function:
    ```python
    def f(value: Optional[List] = None):
        value = value or []
        ...
    ```

schustmi:
@SangamSwadiK As far as I can understand from code, we'll want an if condition that checks if the `ignore_columns` are emtpy/None anyway right? In that case I'd personally go with the `Optional[List[str]]`, but both should be fine.

```python
def entrypoint(..., ignored_columns: Optional[List[str]] = None):
    if ignored_columns:
        # In here you can treat it as a List[str] with elements in it, and mypy should also
        # be aware of that automatically due to the if-condition above
        ...
```

schustmi:
Ah sorry my bad, I didn't read the previous comments. I now see that this is part of a ZenML step, which unfortunately doesn't allow `Optional` inputs which explains the error. Additionally, it right now doesn't work with default values (definitely something we want to tackle in the future to make these steps more flexible).

We could put the `ignore_columns` in the `EvidentlyProfileConfig` for that step instead, which would allow both `Optional` and default values. However I'm not familiar enough with this integration to know if this is something that makes sense or is required on the step, maybe @strickvl has an opinion on that.

strickvl:
@schustmi @SangamSwadiK I think it's reasonable to add the columns to the `EvidentlyProfileConfig` object. I don't see any reasonable why that wouldn't be work. But it will change a few things to the current implementation. (And the tests would need to change as well). But then once implemented that way you can get the values by calling `config.ignored_columns` or something to that equivalent.

stefannica:
> @schustmi @SangamSwadiK I think it's reasonable to add the columns to the `EvidentlyProfileConfig` object. I don't see any reasonable why that wouldn't be work. But it will change a few things to the current implementation. (And the tests would need to change as well). But then once implemented that way you can get the values by calling `config.ignored_columns` or something to that equivalent.

I completely support this idea. The columns are definitely part of the step configuration instead of being step artifacts. The difference is this:

* use a step artifact as input for your step if you need information to be dynamically generated from another step in a pipeline. In other words, you would need another step in your pipeline to dynamically generate the list of columns to be used as input by the Evidently profiler step.
* use a step configuration as input for your step if you know what the configuration is before you run the pipeline.

This definitely feels like a configuration attribute, not like a pipeline artifact.

stefannica:
As a heads-up, we've updated Evidently to the last version in the last release, so you'll have to rebase your PR and test with the latest changes again.

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -111,6 +112,7 @@ class EvidentlyProfileConfig(BaseDriftDetectionConfig):
     """
 
     column_mapping: Optional[EvidentlyColumnMapping] = None
+    ignore_cols: Optional[List[str]] = None
```
strickvl, line 115:
```suggestion
    ignored_cols: Optional[List[str]] = None
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
-            ):
-                raise ValueError(
-                    "Feature not found in reference/comparison datasets"
-                )
+        data_validator = cast(
+            EvidentlyDataValidator,
+            EvidentlyDataValidator.get_active_data_validator(),
+        )
+        column_mapping = None
+        if config.ignore_cols is not None and not (
```
strickvl, line 162:
```suggestion
        if config.ignored_cols is not None and not (
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+            isinstance(config.ignore_cols, list)
```
strickvl, line 163:
```suggestion
            isinstance(config.ignored_cols, list)
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+        ):
+            raise TypeError(
+                f"Expects a List of columns but got type {type(config.ignore_cols)}"
```
strickvl, line 166:
```suggestion
                f"Expects a List of columns but got type {type(config.ignored_cols)}"
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+            )
+        elif len(config.ignore_cols) == 0:
```
strickvl, line 168:
```suggestion
        elif len(config.ignored_cols) == 0:
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+            raise ValueError(
+                f"Expects None or list of columns in strings, but got {config.ignore_cols}"
```
strickvl, line 170:
```suggestion
                f"Expects None or list of columns in strings, but got {config.ignored_cols}"
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+            )
+
+        elif not (all((isinstance(ele, str) for ele in config.ignore_cols))):
```
strickvl, line 173:
```suggestion
        elif not (all((isinstance(ele, str) for ele in config.ignored_cols))):
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+            raise TypeError(
+                "One or more columns to be ignored are not type string"
+            )
 
+        elif not (
+            set(config.ignore_cols).issubset(set(reference_dataset.columns))
```
strickvl, line 179:
```suggestion
            set(config.ignored_cols).issubset(set(reference_dataset.columns))
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+        ) or not (
+            set(config.ignore_cols).issubset(set(comparison_dataset.columns))
```
strickvl, line 181:
```suggestion
            set(config.ignored_cols).issubset(set(comparison_dataset.columns))
```

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -140,55 +141,57 @@ def entrypoint(  # type: ignore[override]
...
+        ):
+            raise ValueError(
+                "Column name is not found in reference or comparison datasets"
+            )
+
+        else:
             reference_dataset = reference_dataset.drop(
-                labels=ignored_columns, axis=1
+                labels=list(config.ignore_cols), axis=1
```
strickvl, line 189:
```suggestion
                labels=list(config.ignored_cols), axis=1
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -36,26 +36,27 @@
...
-    col_list = ["target"]
-
-    clmn_map = ColumnMapping(*col_list)
+    clmn_map = EvidentlyColumnMapping(target_names=["target"])
 
     profile_config = EvidentlyProfileConfig(
         profile_sections=["datadrift"], column_mapping=clmn_map
     )
 
+    print("config created sucessfully")
```
strickvl, line 49:
```suggestion
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -36,26 +36,27 @@
...
     profile_step = EvidentlyProfileStep()
 
+    print(" step created sucessfully")
```
strickvl, line 52:
```suggestion
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -168,22 +144,21 @@ def test_incorrect_datatype() -> None:
         )
 
 
-def test_empty_list() -> None:
-    """Tests ignore cols parameter for empty list
+def test_ignored_cols_elements() -> None:
+    """Tests ignore cols parameter for type of its elements
```
strickvl, line 148:
```suggestion
    """Tests ignored_cols parameter for type of its elements
```

strickvl (commented):
Just an initial few comments. I'll take a deeper look first thing in the morning on this. Thanks for resubmitting!

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -0,0 +1,167 @@
...
+
+    all_cols = target + prediction + cat_feat_names + num_feat_names
+    assert "target" in all_cols
+    assert "prediction" in all_cols
+    assert "ignore_col" in all_cols
+    assert "var_A" in all_cols
+    assert "var_B" in all_cols
+
+
+def test_non_existing_col() -> None:
```
strickvl, line 107:
```suggestion
def test_ignoring_non_existing_column() -> None:
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -0,0 +1,167 @@
...
+
+    with pytest.raises(ValueError):
+        drift_obj, dash_obj = profile_step.entrypoint(
+            reference_dataset=ref,
+            comparison_dataset=comp,
+            config=profile_config,
+        )
+
+
+def test_incorrect_datatype() -> None:
```
strickvl, line 128:
```suggestion
def test_ignoring_incorrect_datatype() -> None:
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -0,0 +1,167 @@
...
+    """Tests ignored_cols parameter for incorrect datatype
+    and raises Error"""
+
+    clmn_map = EvidentlyColumnMapping(target_names=["target"])
```
strickvl, line 132:
```suggestion
    column_map = EvidentlyColumnMapping(target_names=["target"])
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -0,0 +1,167 @@
...
+    profile_config = EvidentlyProfileConfig(
+        profile_sections=["datadrift"],
+        column_mapping=clmn_map,
```
strickvl, line 135:
```suggestion
        column_mapping=column_map,
```

tests/unit/integrations/evidently/steps/test_ignore_columns.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def test_ignored_cols_elements() -> None:
+    """Tests ignored_cols parameter for type of its elements
+    and raises Error"""
+
+    clmn_map = EvidentlyColumnMapping(target_names=["target"])
+    profile_config = EvidentlyProfileConfig(
+        profile_sections=["datadrift"],
+        column_mapping=clmn_map,
```
strickvl, line 156:
```suggestion
    column_map = EvidentlyColumnMapping(target_names=["target"])
    profile_config = EvidentlyProfileConfig(
        profile_sections=["datadrift"],
        column_mapping=column_map,
```

strickvl (commented):
Some more comments here. More to follow.

src/zenml/integrations/evidently/steps/evidently_profile.py
```diff
@@ -150,6 +159,39 @@ def entrypoint(  # type: ignore[override]
             EvidentlyDataValidator.get_active_data_validator(),
         )
         column_mapping = None
+        if config.ignored_cols is not None and not (
```
strickvl, line 162:
```suggestion
        if config.ignored_cols and not (
```

I think maybe you can refactor it down to this..

strickvl (changes requested):
I left some more comments earlier. Hope this is all clear!

Thank you for this implementation.

strickvl:
One other point... in order for the tests to run on our Github CI system, you'll need to add the Evidently integration to be installed for unit-tests.

You can do this in `.github/workflows/unit-test.yml`. You'll find a line that reads:

```shell
zenml integration install kubeflow s3 gcp azure gcp_secrets_manager vertex vault pillow -f
```

You just have to add `evidently` to that list so it becomes:

```shell
zenml integration install kubeflow s3 gcp azure gcp_secrets_manager vertex vault pillow evidently -f
```

And then the tests will be able to run!

Thanks!

strickvl:
@SangamSwadiK Seems like all the tests are failing currently.

strickvl:
@SangamSwadiK There are various ways to test elements that have stack components, like Evidently. ([See the docs](https://docs.zenml.io/mlops-stacks/data-validators/evidently) on Evidently for an overview of how that works.)

For Evidently and the tests you've written, I think probably the best way to do this will be to mock the following method call within your tests:

```python
EvidentlyDataValidator.get_active_data_validator()
```

The mocked function should return an object you create at the beginning of the tests (i.e. your Evidently data validator). Then you don't need to mess around with registering stack components (though that would be an alternative method to do this test).

Does that make sense?

strickvl:
I might suggest the following: how about we merge this onto our `develop` branch once we make our next release (today or tomorrow) and then we just handle the testing part ourselves? I appreciate that the codebase has changed a few times since you started this PR and we're also making some internal changes around how we test things, so it probably is an unfair ask for us to get you to write these tests.

In that case, I'd suggest you just stop working on the tests and let us merge this all in onto develop in a day or two and then write the tests.

Would that be ok, @SangamSwadiK?

schustmi:
@SangamSwadiK Yes, creating a new PR is probably the easiest option, but that PR can also only have commits from the account with which you can sign the CLA. Thanks again for your contribution!

PR #708, opened 2022-06-16
Remove `force` flag from secrets managers' implementation

strickvl:
We had an inconsistently used `force` flag which was defined in the base secrets manager. I removed this, since we already have checks at the CLI level to confirm that people really want to delete all secrets.

I added some missing license statements where I found them.

I also looked to see if there were any tests that were going to break as a result of these changes, but I see that we currently have no tests covering these changes (or really *any* effective testing of our cloud secrets managers).

### DIFF
--- a/docs/book/extending-zenml/secrets-managers.md
+++ b/docs/book/extending-zenml/secrets-managers.md
@@ -63,7 +63,7 @@ class BaseSecretsManager(StackComponent, ABC):
         """Delete an existing secret."""
 
     @abstractmethod
-    def delete_all_secrets(self, force: bool = False) -> None:
+    def delete_all_secrets(self) -> None:
         """Delete all existing secrets."""
 ```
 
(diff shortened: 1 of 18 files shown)


schustmi (changes requested):
I think the GitHubSecretsManager might still need an update, otherwise looks good

strickvl:
Ah yes, I forgot that that was even a thing!

schustmi (approved):

AlexejPenner (approved):
Good Catch :)

strickvl:
I don't think this PR will include any more tests, @AlexejPenner. I think that'll come either with #ENG-934 or some more extended integration tests.

AlexejPenner:
> I don't think this PR will include any more tests, @AlexejPenner. I think that'll come either with #ENG-934 or some more extended integration tests.

Oh wasn't talking about new tests, its just that it was failing unit tests because there is a force flag being passed in the tests: https://github.com/zenml-io/zenml/runs/6914867938?check_suite_focus=true

PR #713, opened 2022-06-20
Misc bugfixes

schustmi:
## Describe changes
This PR fixes multiple small things I found when writing the GH actions tutorial:
* Expand file paths during interactive secret registration
* Fix typo in github example
* Allow ~ in paths when passing files as secrets
* Quote pip requirements during docker install 
* Updating stack component now uses pydantic validation

## Pre-requisites

## Types of changes


src/zenml/cli/secret.py
```diff
@@ -215,9 +218,12 @@ def register_secret(
             while True:
                 k = click.prompt("Please enter a secret-key")
                 if k not in secret_contents:
-                    secret_contents[k] = getpass.getpass(
+                    v = getpass.getpass(
                         f"Please enter the secret_value for the key [{k}]:"
                     )
```
strickvl, line 223: Maybe choose whether we're using underscores or hyphens? At moment possibly the inconsistency is confusing?

strickvl (approved):
Just noticed one inconsistency in here (from before you made the PR). Otherwise looks good to me.

src/zenml/cli/secret.py
schustmi: We can probably even remove it and use spaces instead right?

src/zenml/cli/secret.py
```diff
@@ -206,17 +207,22 @@ def register_secret(
             for k in secret_keys:
                 v = getpass.getpass(f"Secret value for {k}:")
                 if v:
-                    secret_contents[k] = v
+                    secret_contents[k] = expand_argument_value_from_file(
```
stefannica, line 210: When adding the ability to expand secret values from file contents, I was thinking of also doing it in the interactive mode, but then decided not to, because 1. you can't see what you're typing and 2. you don't get auto-completion to help you to point to an existing file on disk. These reasons make it really error prone to specify a path in interactive mode.

src/zenml/cli/stack_components.py
```diff
@@ -643,9 +643,13 @@ def update_stack_component_command(
...
 
-            updated_component = component_wrapper.to_component().copy(
-                update=parsed_args
-            )
+            # Initialize a new component object to make sure pydantic validation
+            # is used
+            new_attributes = {
+                **component_wrapper.to_component().dict(),
+                **parsed_args,
+            }
```
stefannica, line 651: Nice ! :)

stefannica (approved):
Nice work !

src/zenml/cli/secret.py
schustmi: Do you think it's better to remove it again? I've just without checking expected the interactive version to work the same way and tried copy-pasting the paths there with a prepended @ and got confused when it didn't work, that's why I implemented it.

PR #715, opened 2022-06-20
Fix MLflow repeated deployment error

fa9r:
## Describe changes
I fixed a bug that would say `NoSuchProcess: process no longer exists (pid=...)` when running a pipeline with MLflow deployer twice. 

The issue was that the BaseService would issue a command to kill the currently running process and then immediately poll that process to make sure it went down, which would lead to a race condition since the process was stopped in the middle of the polling.

## Pre-requisites

## Types of changes


examples/quickstart/run.py
```diff
@@ -66,31 +66,29 @@ def main():
...
-        -1
-    ]
+    inf_run = repo.get_pipeline(pipeline_name="inference_pipeline").runs[-1]
 
     # visualize training pipeline
     PipelineRunLineageVisualizer().visualize(train_run)
 
     # visualize inference pipeline
-    PipelineRunLineageVisualizer().visualize(inference_run)
+    PipelineRunLineageVisualizer().visualize(inf_run)
```
strickvl, line 80: what was this about?

strickvl (commented):
I like how this is handled. I'm not a super fan of insertions of `sleep()` into things, but maybe there are benefits to that approach I'm not seeing.

examples/quickstart/run.py
fa9r: Just making L74 a bit more readable by reducing it to <80 chars 😄

stefannica (approved):
Clever catch ! Race conditions are notoriously difficult to debug and fix.

PR #723, opened 2022-06-24
Add missing help text for CLI commands 

safoinme:
## Describe changes
Add the CLI help text commands for commands using the docstring, whenever a command using a docstring that have args as a help. it shows the args in an unformatted way.

While there is a solution to rewrap this for click to show the args each in a line. this doesn't work with our docstring checking library.

image

## Pre-requisites

## Types of changes


fa9r (approved):
LGTM

src/zenml/cli/secret.py
```diff
@@ -92,56 +96,10 @@ def register_secret(
 
     Use this command to store sensitive information into a ZenML secret. The
     secret data consists of key-value pairs that can be configured interactively
-    (if the `--interactive` option is set) or via command line arguments.
+    (if the `--interactive` option is set) or via command-line arguments.
     If a schema is indicated, the secret key-value pairs will be validated
     against the schema.
 
-    When passed as command line arguments, the secret field values may also be
```
strickvl, line 99: Does this text just get deleted? Are we moving it somewhere else?

safoinme: Ah, that was deleted with the old way cause it didn't make sense to show it in CLI. I return it back until figure out where that should be written.

strickvl (approved):
Yeah once the linting issues with darglint etc are fixed, this looks good to me now!

Issue #727, opened 2022-06-26
[FEATURE]: Allow fetching from Repository by name or type


strangemonad:
### Contact Details [Optional]

shawnmorel@gmail.com

### Describe the feature you'd like

This is tangentially related to #726 

Currently, when you want to fetch items from a `Repository` you need to use strings to identify the keys. I would like to be able to use the pipeline and step objects to fetch views from the repository.

In general, we'd like to limit how much we rely on string naming agreement. This allows refactoring tools to do a better job but also means you never have to copy paste strings since your editor can auto-complete pipeline and step definition names.

e.g.

```python
@step
def my_step():
  ...

@pipeline
def my_pipeline(my_step):
  ...

p = my_pipeline(my_step())
p.run()

repo = Repository()
```

I'd like the following to work

```python
pipeline_view = repo.get_pipeline(p)
pipeline_view = repo.get_pipeline(my_pipeline)
pipeline_view = repo.get_pipeline("my_pipeline")

step_view = pv.steps[my_step]
s = p.steps[...]
step_view = pv.steps[s]
```

### Is your feature request related to a problem?

Sort of, we're anticipating maintenance problems as we scale and step and pipeline names might change over time but we wouldn't be able to catch all usages in notebooks. In our own codebase, we'd like to enforce never using string names in a custom flake8 code style plugin. 

Imaging we've refactored `myproj.pipelines.my_fancy_pipeline.my_fancy_pipeline` -> `myproj.pipelines.my_awesome_pipeline import my_awesome_pipeline`

the following code would fail with an import error:

```python
from  myproj.pipelines.my_fancy_pipeline import my_fancy_pipeline
# ^^^ module not found

repo = Repository()
repo.get_pipelines(my_fancy_pipeline)
```

### How do you solve your current problem with the current status-quo of ZenML?

We've written the following utility code 

```python
PipelineName = Union[str, BasePipeline, BasePipelineMeta]


@beta
def get_last_run(pipeline: PipelineName) -> PipelineRunView:
    """Hacky convenience method to fetch the last run of a pipeline.

    Notes
    -----
    This is inherently racy, but it's currently the only way to fetch the pipeline run details.
    This approach is also what's currently suggested in the docs.

    See Also
    --------
    https://github.com/zenml-io/zenml/issues/726

    """
    repo = Repository()

    return repo.get_pipeline(pipeline_name=to_pipeline_name(pipeline)).runs[-1]


@beta
[...]

PR #733, opened 2022-06-28
Allow fetching pipeline/step by name, class or instance

AlexejPenner:
## Describe changes
This relates to [Issue 727](https://github.com/zenml-io/zenml/issues/727)

The following code should demonstrate how the access to pipelines and steps has been expanded. Examples and Tests still need to be adjusted to these changes.
`
import random

from zenml.pipelines import pipeline
from zenml.repository import Repository
from zenml.steps import Output, step

@step
def get_first_num() -> Output(first_num=int):
    """Returns an integer."""
    return 10

@step(enable_cache=False)
def get_random_int() -> Output(random_num=int):
    """Get a random integer between 0 and 10"""
    return random.randint(0, 10)

@step
def subtract_numbers(first_num: int, random_num: int) -> Output(result=int):
    """Subtract random_num from first_num."""
    return first_num - random_num

@pipeline
def example_pipeline(get_first_num, get_random_int, subtract_numbers):
    # Link all the steps artifacts together
    first_num = get_first_num()
    random_num = get_random_int()
    subtract_numbers(first_num, random_num)

# Initialize a new pipeline run
pipe = example_pipeline(
    get_first_num=get_first_num(),
    get_random_int=get_random_int(),
    subtract_numbers=subtract_numbers(),
)

# Run the new pipeline
pipe.run()

repo = Repository()

a = repo.get_pipeline(pipeline_name="example_pipeline").runs[-1]
b = repo.get_pipeline(pipeline=example_pipeline).runs[-1]
c = repo.get_pipeline(pipeline=pipe).runs[-1]

w = a.get_step(name="get_first_num")
x = a.get_step(step_name="get_first_num")
y = b.get_step(step=get_first_num)
z = c.get_step(step=get_first_num())
`

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/pipelines/base_pipeline.py
+++ b/src/zenml/pipelines/base_pipeline.py
@@ -72,7 +72,7 @@ class BasePipelineMeta(type):
     """Pipeline Metaclass responsible for validating the pipeline definition."""
 
     def __new__(
-            mcs, name: str, bases: Tuple[Type[Any], ...], dct: Dict[str, Any]
+        mcs, name: str, bases: Tuple[Type[Any], ...], dct: Dict[str, Any]
     ) -> "BasePipelineMeta":
         """Saves argument names for later verification purposes.
 
@@ -309,7 +309,7 @@ class BasePipeline(metaclass=BasePipelineMeta):
                 ) from e
 
         if isinstance(self._requirements, str) and fileio.exists(
-                self._requirements
+            self._requirements
         ):
             with fileio.open(self._requirements, "r") as f:
                 requirements.update(
@@ -399,9 +399,8 @@ class BasePipeline(metaclass=BasePipelineMeta):
 
         for step in self.steps.values():
             if (
-                    step.custom_step_operator
-                    and step.custom_step_operator not in
-                    available_step_operators
+                step.custom_step_operator
+                and step.custom_step_operator not in available_step_operators
             ):
                 raise StackValidationError(
                     f"Step '{step.name}' requires custom step operator "
@@ -419,11 +418,11 @@ class BasePipeline(metaclass=BasePipelineMeta):
             step._has_been_called = False
 
     def run(
-            self,
-            *,
-            run_name: Optional[str] = None,
-            schedule: Optional[Schedule] = None,
-            **additional_parameters: Any,
+        self,
+        *,
+        run_name: Optional[str] = None,
+        schedule: Optional[Schedule] = None,
+        **additional_parameters: Any,
     ) -> Any:
         """Runs the pipeline on the active stack of the current repository.
 
@@ -498,7 +497,7 @@ class BasePipeline(metaclass=BasePipelineMeta):
         )
 
     def with_config(
-            self: T, config_file: str, overwrite_step_parameters: bool = False
+        self: T, config_file: str, overwrite_step_parameters: bool = False
     ) -> T:
         """Configures this pipeline using a yaml file.
 
@@ -526,7 +525,7 @@ class BasePipeline(metaclass=BasePipelineMeta):
         return self
 
     def _read_config_steps(
-            self, steps: Dict[str, Dict[str, Any]], overwrite: bool = False
+        self, steps: Dict[str, Dict[str, Any]], overwrite: bool = False
     ) -> None:
         """Reads and sets step parameters from a config file.
 
--- a/src/zenml/pipelines/base_pipeline.py
+++ b/src/zenml/pipelines/base_pipeline.py
@@ -31,7 +31,6 @@ from typing import (
     cast,
 )
 
-import zenml.cli.utils as cli_utils
 from zenml import constants
 from zenml.config.config_keys import (
     PipelineConfigurationKeys,
@@ -73,7 +72,7 @@ class BasePipelineMeta(type):
     """Pipeline Metaclass responsible for validating the pipeline definition."""
 
     def __new__(
-        mcs, name: str, bases: Tuple[Type[Any], ...], dct: Dict[str, Any]
+            mcs, name: str, bases: Tuple[Type[Any], ...], dct: Dict[str, Any]
     ) -> "BasePipelineMeta":
         """Saves argument names for later verification purposes.
 
@@ -116,7 +115,8 @@ class BasePipeline(metaclass=BasePipelineMeta):
         requirements_file: DEPRECATED: Optional path to a pip requirements file
             that contains all requirements to run the pipeline. (Use
             `requirements` instead.)
-        requirements: Optional list of (string) pip requirements to run the pipeline, or a string path to a requirements file.
+        requirements: Optional list of (string) pip requirements to run the
+        pipeline, or a string path to a requirements file.
         required_integrations: Optional set of integrations that need to be
             installed for this pipeline to run.
     """
@@ -138,7 +138,10 @@ class BasePipeline(metaclass=BasePipelineMeta):
         self.requirements_file = kwargs.pop(PARAM_REQUIREMENTS_FILE, None)
         if self.requirements_file:
             logger.warning(
-                "The `requirements_file` argument has been deprecated. Please use `requirements` instead to pass in either a string path to a file listing your 'requirements' or a list of the individual requirements."
+                "The `requirements_file` argument has been deprecated. Please "
+                "use `requirements` instead to pass in either a string path "
+                "to a file listing your 'requirements' or a list of the "
+                "individual requirements."
             )
         self._requirements = kwargs.pop(PARAM_REQUIREMENTS, None)
         self.dockerignore_file = kwargs.pop(PARAM_DOCKERIGNORE_FILE, None)
@@ -306,7 +309,7 @@ class BasePipeline(metaclass=BasePipelineMeta):
                 ) from e
 
         if isinstance(self._requirements, str) and fileio.exists(
-            self._requirements
+                self._requirements
         ):
             with fileio.open(self._requirements, "r") as f:
                 requirements.update(
@@ -316,8 +319,9 @@ class BasePipeline(metaclass=BasePipelineMeta):
                     }
                 )
             if self.requirements_file:
-                cli_utils.warning(
-                    f"Using the file path passed in as `requirements` and ignoring the file '{self.requirements_file}'."
+                logger.warning(
+                    f"Using the file path passed in as `requirements` and "
+                    f"ignoring the file '{self.requirements_file}'."
                 )
         # Add this logic back in (described in #ENG-882)
         #
@@ -342,8 +346,9 @@ class BasePipeline(metaclass=BasePipelineMeta):
         elif isinstance(self._requirements, List):
             requirements.update(self._requirements)
             if self.requirements_file:
-                cli_utils.warning(
-                    f"Using the values passed in as `requirements` and ignoring the file '{self.requirements_file}'."
+                logger.warning(
+                    f"Using the values passed in as `requirements` and "
+                    f"ignoring the file '{self.requirements_file}'."
                 )
         elif self.requirements_file and fileio.exists(self.requirements_file):
             # TODO [ENG-883]: Deprecate the `requirements_file` option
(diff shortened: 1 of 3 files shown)


src/zenml/post_execution/pipeline_run.py
```diff
@@ -156,11 +157,18 @@ def get_step_names(self) -> List[str]:
         self._ensure_steps_fetched()
         return list(self._steps.keys())
 
-    def get_step(self, name: str) -> StepView:
+    def get_step(
+        self,
+        step: Optional["BaseStep"] = None,
+        step_name: Optional[str] = "",
```
schustmi, line 163:
```suggestion
        step_name: Optional[str] = None,
```

src/zenml/post_execution/pipeline_run.py
```diff
@@ -169,11 +177,34 @@ def get_step(self, name: str) -> StepView:
...
         self._ensure_steps_fetched()
+        if step and step_name:
+            raise TypeError(
+                "'step' and 'step_name' both set for the"
+                "get_step() method. This is not supported. "
+                "Please set either the 'step' or the "
+                "'step_name' parameter."
+            )
+        elif step:
+            if isinstance(step, zenml.steps.base_step.BaseStepMeta):
```
schustmi, line 188: I'm not sure why this check is needed here? Is it to also allow step classes being passed?

If yes I suggest we replace it by `issubclass(step, BaseStep` instead of a check of the metaclass. It would also need to be included in the type annotation of the function signature, e.g. 
```python
step: Optional[Union["BaseStep", Type["BaseStep"]] = None,
```

I also have one more "suggestion" (I'm not quite sure if I like it myself), but just throwing it our there in case you like it:
Instead of having two parameters `step_name` and `step`, we could make it one parameter which can be either a string or a step. I like two things about it:
- We don't need a check if both/none of them are set
- Usage of the method is much easier and doesn't require a keyword argument each time:
```python
# both of these work without any step=..., step_name=...
run.get_step(MyStepClass)
run.get_step("name_of_my_step")
```

src/zenml/post_execution/pipeline_run.py
```diff
@@ -169,11 +177,34 @@ def get_step(self, name: str) -> StepView:
             KeyError: If there is no step with the given name.
         """
         self._ensure_steps_fetched()
+        if step and step_name:
+            raise TypeError(
```
schustmi, line 181: ValueError or RuntimeError imo

src/zenml/repository.py
```diff
@@ -1078,11 +1085,15 @@ def get_pipelines(
 
     @track(event=AnalyticsEvent.GET_PIPELINE)
     def get_pipeline(
-        self, pipeline_name: str, stack_name: Optional[str] = None
+        self,
+        pipeline: Optional["BasePipeline"] = None,
+        pipeline_name: Optional[str] = "",
```
schustmi, line 1090:
```suggestion
        pipeline_name: Optional[str] = None,
```

src/zenml/repository.py
```diff
@@ -1096,6 +1107,27 @@ def get_pipeline(
             RuntimeError: If no stack name is specified and no active stack name
                 is configured.
         """
+        if pipeline and pipeline_name:
+            raise TypeError(
```
schustmi, line 1111: ValueError or RuntimeError

src/zenml/repository.py
```diff
@@ -1096,6 +1107,27 @@ def get_pipeline(
...
+            if isinstance(
+                pipeline, zenml.pipelines.base_pipeline.BasePipelineMeta
+            ):
+                pipeline_name = pipeline.__name__
+            elif isinstance(
+                pipeline, zenml.pipelines.base_pipeline.BasePipeline
+            ):
+                pipeline_name = pipeline.name
+        elif not pipeline_name:
+            raise TypeError(
```
schustmi, line 1127: ValueError or RuntimeError

src/zenml/repository.py
```diff
@@ -1096,6 +1107,27 @@ def get_pipeline(
...
         """
+        if pipeline and pipeline_name:
+            raise TypeError(
+                "'pipeline' and 'pipeline_name' both set for the"
+                "get_pipeline() method. This is not supported. "
+                "Please set either the 'pipeline' or the "
+                "'pipeline_name' parameter."
+            )
+        elif pipeline:
+            if isinstance(
```
schustmi, line 1118: Same as the comment on the `get_step`, this should probably use `issubclass(...)` if the behavior is intended and also needs an update on the type annotation

schustmi (changes requested):

src/zenml/post_execution/pipeline_run.py
```diff
@@ -169,11 +177,34 @@ def get_step(self, name: str) -> StepView:
...
+                "'step_name' parameter."
+            )
+        elif step:
+            if isinstance(step, zenml.steps.base_step.BaseStepMeta):
+                step_name = step.__name__
+            elif isinstance(step, zenml.steps.base_step.BaseStep):
+                step_name = step.name
+        elif "name" in kwargs and not step_name:
+            logger.warning(
+                "Using 'name' to get a step from "
```
strangemonad, line 194: super nit. Warnings like this can be much more friendly for newcomers if you can put a link to the specific API docs containing examples. `https://apidocs.zenml.io/latest/api_docs/post_execution/#zenml.post_execution.pipeline_run.PipelineRunView.get_step`


you can streamline this as

```
def api_link() -> str:
   # stack trace get caller name
   return "https://apidocs.zenml.io/latest/api_docs/post_execution/#{caller_name}"
```

strangemonad:
@AlexejPenner @schustmi left a couple comments and nits. Thanks so much for tackling this usability feature so quickly. These seem trivial but go a long way to create buy-in with newcomers.

One comment, not sure if this is just because it's just a draft or what your stance is on unit tests but certainly as the complexity of the zenml code base grows, having more unit tests will probably be helpful to prevent regressions.

thanks so much for this framework!

AlexejPenner:
Hi @strangemonad and @schustmi, thanks for your thoughts, I will be applying the suggestions today!

Yes, as this was a draft I did not add any unit-tests, but they will be part of the finished PR.

src/zenml/post_execution/pipeline_run.py
AlexejPenner: I love this suggestion, I'll see what I can do within this ticket to make the warnings more user friendly.

src/zenml/pipelines/base_pipeline.py
```diff
@@ -588,3 +594,22 @@ def _read_config_steps(
                             parameter, step_name, previous_value, value
                         )
                     )
+
+    def get_runs(self):
```
schustmi, line 598: Nice idea to put this here. Can we potentially make this a `classmethod` so that users don't even have to create a pipeline instance for this?

tests/unit/post_execution/test_pipeline_run.py
```diff
@@ -11,3 +11,80 @@
...
+    for input_arg in input_args:
+        returned_sv = sample_pipeline_run_view.get_step(**input_arg)
+        assert sample_stepview._id == returned_sv._id
+        assert sample_stepview._name == returned_sv._name
+
+
+def test_get_step_raises_exception(
+    sample_step: "BaseStep", sample_stepview: StepView, sample_pipeline_run_view
+):
+    """Test that the `get_step` method returns the correct step_view"""
```
schustmi, line 76: wrong docstring

tests/unit/test_repository.py
```diff
@@ -474,3 +475,53 @@ def test_get_pipelines_forwards_to_metadata_store(clean_repo, mocker):
...
+    for input_arg in input_args:
+        SQLiteMetadataStore.get_pipeline.reset_mock()
+        clean_repo.get_pipeline(**input_arg)
+        SQLiteMetadataStore.get_pipeline.assert_called_once_with(
+            "some_pipeline"
+        )
+
+
+def test_get_pipeline_raises_exception(clean_repo, mocker):
+    """Tests that getting post-execution pipelines forwards calls to the
```
schustmi, line 507: wrong dockstring

tests/unit/post_execution/test_pipeline_run.py
```diff
@@ -11,3 +11,80 @@
...
+    ]  # calling with deprecated kwarg
+
+    for input_arg in input_args:
+        returned_sv = sample_pipeline_run_view.get_step(**input_arg)
+        assert sample_stepview._id == returned_sv._id
+        assert sample_stepview._name == returned_sv._name
+
+
+def test_get_step_raises_exception(
+    sample_step: "BaseStep", sample_stepview: StepView, sample_pipeline_run_view
```
schustmi, line 74: I think we don't need the `sample_step` and `sample_stepview` here right? And the `sample_pipeline_run_view` is missing a type annotation.

tests/unit/post_execution/test_pipeline_run.py
```diff
@@ -11,3 +11,80 @@
...
+    """Return sample pipeline run view for testing purposes"""
+    sample_pipeline_run_view = PipelineRunView(
+        id_=1, name="sample_run_name", executions=None, metadata_store=None
+    )
+    setattr(sample_pipeline_run_view, "_steps", {"some_step": sample_stepview})
+    return sample_pipeline_run_view
+
+
+def test_get_step_returns_stepview(
+    sample_step: "BaseStep", sample_stepview: StepView, sample_pipeline_run_view
```
schustmi, line 56: The `sample_pipeline_run_view` is missing a type annotation.

schustmi (changes requested):

src/zenml/logger.py
```diff
@@ -238,3 +239,19 @@ def disable_logging(log_level: int) -> Iterator[None]:
...
+
+    Args:
+        caller_path: Path to the class, method or function to which the api
+            doc-link should point.
+
+    Returns:
+        Link to the api docs for the given caller_path
+    """
+    return (
+        f"https://apidocs.zenml.io/{zenml.__version__}/api_docs/"
```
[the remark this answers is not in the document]
AlexejPenner: any suggestions where to move this?

src/zenml/pipelines/base_pipeline.py
AlexejPenner: definitely

src/zenml/logger.py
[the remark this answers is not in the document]
schustmi: Not really 😄 
Maybe in some utils file?

schustmi (approved):

src/zenml/pipelines/base_pipeline.py
```diff
@@ -588,3 +594,24 @@ def _read_config_steps(
...
                     )
+
+    @classmethod
+    def get_runs(cls) -> List["PipelineRunView"]:
+        """Get all past runs from the associated PipelineView.
+
+        Returns:
+            A list of all past PipelineRunViews.
+        """
+        return Repository().get_pipeline(cls).runs  # type: ignore[no-any-return]
```
schustmi, line 605: Shouldn't this code have a check if the pipeline exists?
`Repository().get_pipeline(...)` might return `None` and crash here?

src/zenml/pipelines/base_pipeline.py
```diff
@@ -588,3 +594,24 @@ def _read_config_steps(
...
+    def get_run(cls, run_name: str) -> "PipelineRunView":
+        """Get a specific past run from the associated PipelineView.
+
+        Args:
+            run_name: Name of the run
+
+        Returns:
+            The PipelineRunView of the specific pipeline run.
+        """
+        return Repository().get_pipeline(cls).get_run(run_name)  # type: ignore[no-any-return]
```
schustmi, line 617: Same as above

schustmi (changes requested):

src/zenml/pipelines/base_pipeline.py
AlexejPenner: Wouldn't it already fail with the correct RuntimeError from within Repository().get_pipeline(), what would be the purpose of catching and reraising the error here?

schustmi: If I understood correctly, the `RuntimeError` only happens if a wrong type/nothing is passed to get the pipeline name.
But if you use the method correctly, it might still happen that this pipeline has never been run and isn't stored in the metadata store. That's why the `get_pipeline` method returns an `Optional[PipelineView]`. It would then fail here with an error that it's unable to find the attribute `runs` on an object of type `None` is my guess

Issue #736, opened 2022-06-28
[BUG]: can't register a feast stack component because the feast feature store flavor has an incorrect FlavorWrapper source module


strangemonad:
### Contact Details [Optional]

shawnmorel@gmail.com

### System Information

ZenML version: 0.9.0
Install path: /Users/shawnmorel/Code/rivian/me/factory-data/factory-data-algorithms/projects/common/process-health/.venv/lib/python3.9/site-packages/zenml
Python version: 3.9.9
Platform information: {'os': 'mac', 'mac_version': '12.4'}
Environment: native
Integrations: ['feast', 'kubeflow', 'scipy', 'seldon']

### What happened?

From `zenml.integrations.feast`

```
FlavorWrapper(
    name=FEAST_FEATURE_STORE_FLAVOR,
    source="zenml.integrations.feast.feature_store.FeastFeatureStore",
    type=StackComponentType.FEATURE_STORE,
    integration=cls.NAME,
)
```

notice the incorrect `...feast.feature_store` package name, should be `...features_stores` to match reality.

May I suggest that the pattern used to define integrations be made less fragile rather than a human provided string, source should come from `f{FeastFeatureStore.__module__}.{FeastFeatureStore.__qualname__}`. This could be streamlined in the `Integration` base class.

### Reproduction steps

```console
zenml integration install feast
zenml feature-store register blah --flavor feast
Error: The feature_store flavor 'feast' is a part of ZenML's 'feast' integration, which is currently not installed on your system. You can install it by executing: 'zenml integration install feast'.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

strickvl:
Thanks for reporting this. I was just going to write a quick reply to say that this has already been fixed, but when I go to the `develop` branch to confirm it then it seems like it wasn't fixed. Not sure what happened there. I'll pass on your suggestion on having a less flaky way to pass in that pattern to the core team. You're right it does seem error prone..

PR #737, opened 2022-06-29
Fix broken Feast integration

strickvl:
Referenced in #736, this small fix updates the source string passed in on initialising the Feast feature store integration.


schustmi (approved):

AlexejPenner (approved):
Let's merge this in, the test-failures are sporadic due to a broken cache on Linux with python 3.8

PR #739, opened 2022-06-29
Fix broken Azure Secrets Manager (and associated schema)

strickvl:
Internally transform the key names for the Azure Secret Schema so that underscores get stripped out and added back in.

## Pre-requisites

## Types of changes


AlexejPenner (commented):
Definitely couldn't come up with a better solution than this. As we are piggybacking on the logic inside the prepend_group_name_to_keys function, it might make sense renaming it to prepare_data_for_saving or something similar to make it a bit more obvious that more is going on inside this function. other than that, I'm happy that you found a way :)

strickvl:
@AlexejPenner to be perfectly transparent, it was @schustmi who suggested this (better) way somewhere on Discord :)

schustmi:
Actually when I suggested it I had a general solution dealing with underscores in mind and not hard-coded to make it work with a single secret schema. This will work with this exact schema, but as soon as someone wants to store for example our MySQL authentication credentials it breaks again. I think the secret manager in general needs to have a way to support different keys with underscores and handle the mapping to/from those names in a generic way.

I don't have any concrete implementation suggestion, but here are a few ideas that might or might not work in the special Azure case:
- replacing underscores with some other special characters
- encode the key name in some way
- Use metadata that is associated with a secret to store the real key names

PR #757, opened 2022-07-07
New Docker images

schustmi:
## Describe changes

This PR adds Dockerfiles to
- build release images for all supported python versions
- build ZenServer images for all supported python versions
- quickly build a development image of the local environment

## Pre-requisites

## Types of changes


docker/base.Dockerfile
```diff
@@ -1,33 +1,16 @@
...
-RUN curl -sSL https://bootstrap.pypa.io/get-pip.py | python && \
-  pip install --no-cache-dir --upgrade --pre pip
+    PIP_NO_CACHE_DIR=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1
 
 ARG ZENML_VERSION
+
 # install the given zenml version (default to latest)
-RUN pip install --no-cache-dir zenml${ZENML_VERSION:+==$ZENML_VERSION}
+RUN pip install zenml${ZENML_VERSION:+==$ZENML_VERSION}
```
stefannica, line 13: I thought using `--no-cache-dir` in your dockerfile results in a smaller image. Was that not the case ?

That being said, caching may speed up the process of building the image, but in that case you might want to [remove the cache](https://pip.pypa.io/en/stable/topics/caching/#removing-the-cache) at the end of the build.

stefannica (approved):
Great work on building slimmer container images. I especially like the poetry.lock hack you included. I had one question, nothing major though.

docker/base.Dockerfile
schustmi: I ran some tests and the environment variable `PIP_NO_CACHE_DIR` set at the top seemed to be enough

stefannica: Ah, I missed that :) good work !

PR #760, opened 2022-07-08
Encode azure secrets manager secret names

schustmi:
## Describe changes
This PR encodes Azure secret names in order to avoid issues with special characters like underscores.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/azure/secrets_managers/azure_secrets_manager.py
+++ b/src/zenml/integrations/azure/secrets_managers/azure_secrets_manager.py
@@ -143,10 +143,10 @@ class AzureSecretsManager(BaseSecretsManager):
         self._ensure_client_connected(self.key_vault_name)
 
         for key, value in secret.content.items():
-            encoded_name = base64.b64encode(
+            encoded_key = base64.b64encode(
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/azure/secrets_managers/azure_secrets_manager.py
```diff
@@ -199,18 +142,24 @@ def update_secret(self, secret: BaseSecretSchema) -> None:
         """
         self._ensure_client_connected(self.key_vault_name)
 
-        adjusted_content = prepend_group_name_to_keys(secret)
+        for key, value in secret.content.items():
+            encoded_name = base64.b64encode(
```
strickvl, line 146: `encoded_name` feels like maybe it's the wrong name for it? aren't we talking about what the key is called here? so maybe call it `encoded_key` or `encoded_key_name`?

schustmi: Good point, I updated it

AlexejPenner (approved):
LGTM! :)

PR #762, opened 2022-07-08
Add optional machine specs to vertex orchestrator

felixthebeard:
Draft implementation to allow specifying machine specs for the vertex orchestrator:
The following optional fields can be configured:
- cpu_limit
- memory_limit
- node_selector_constraint
- gpu_limit

These are set for the entire pipeline for now. Once a more generic interface for specifying machine specs on a step level is defined this can be extended.

## Pre-requisites

## Types of changes


schustmi (approved):
Thanks for the PR, no complaints from my side!

Two small questions:
- You made the `node_selector_constraint` a Literal with those exact values, I'm guessing that Vertex only allows these exact machines?
- Does Vertex have any boundaries for the other attributes (e.g. `memory_limit`)? If you we could potentially include lower/upper bounds for those integers for pydantic validation.

gabrielmbmb:
> Thanks for the PR, no complaints from my side!
> 
> Two small questions:
> 
>     * You made the `node_selector_constraint` a Literal with those exact values, I'm guessing that Vertex only allows these exact machines?
> 
>     * Does Vertex have any boundaries for the other attributes (e.g. `memory_limit`)? If you we could potentially include lower/upper bounds for those integers for pydantic validation.

From the official docs https://cloud.google.com/vertex-ai/docs/pipelines/machine-types, the max CPU limit is 96 and the max memory is 624G. Also, the values for the `Literal` of `node_selector_constraint` are correct according to the documentation.

schustmi:
Thanks for looking it up @gabrielmbmb.

My idea would have been to check this using some options on pydantics `Field` class:
```python
from pydantic import BaseModel, Field

class Test(BaseModel):
    cpu_limit: int = Field(gt=0, le=96)  # greater than 0, lower or equal to 96
```

However after taking a second look these fields allow string values as well, which would make this entire check more complicated and probably unnecessary.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -30,7 +30,17 @@
...
 
 import os
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Optional, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    ClassVar,
+    Dict,
+    List,
+    Literal,
```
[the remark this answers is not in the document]
schustmi: I don't think we use it ourselves, but it's installed by some intermediate dependency.
Another solution would be to make this an enum, which would enforce similar validation using pydantic.
However I think the literal approach is quite nice, in case of `typing_extensions` I'd just suggest doing this:
```python
from typing_extensions import Literal
```
instead as that already uses `typing.Literal` on python 3.8+ and the imports look cleaner

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -414,6 +438,23 @@ def _construct_kfp_pipeline() -> None:
...
+
+                if self.memory_limit is not None:
+                    container_op = container_op.set_memory_limit(
+                        self.memory_limit
+                    )
+
+                if self.node_selector_constraint is not None:
+                    container_op = container_op.add_node_selector_constraint(
+                        self.node_selector_constraint
+                    )
```
AlexejPenner, line 453: Not sure which version of kfp you were running with, for me when trying this out in my dev-environment i ran into issues with the .add_node_selector_constraint as the second positional argument is missing, would this work for you:

```suggestion
                    container_op = container_op.add_node_selector_constraint(
                        label_name="cloud.google.com/gke-accelerator",
                        value=self.node_selector_constraint
                    )
```

If I understood the docs correctly, this "cloud.google.com/gke-accelerator" is currently the only available label name

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -414,6 +438,23 @@ def _construct_kfp_pipeline() -> None:
                     ]
                     container_op.after(upstream_container_op)
 
+                # Set optional CPU, RAM and GPU constraints for the pipeline
+                if self.cpu_limit is not None:
+                    container_op = container_op.set_cpu_limit(self.cpu_limit)
```
AlexejPenner, line 443: Were you able to run with integer values for the cpu limit? When I set --cpu-limit=1000m it works but when I run --cpu-limit=1, I get the following failure:

TypeError: expected string or bytes-like object

at this exact line in code.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -135,6 +145,20 @@ class VertexOrchestrator(BaseOrchestrator, GoogleCredentialsMixin):
...
+    node_selector_constraint: Optional[
+        Literal[
+            "NVIDIA_TESLA_A100",
+            "NVIDIA_TESLA_K80",
+            "NVIDIA_TESLA_P4",
+            "NVIDIA_TESLA_P100",
+            "NVIDIA_TESLA_T4",
+            "NVIDIA_TESLA_V100",
+        ]
+    ] = None
```
AlexejPenner, line 159: It took me some trial and error until I realized that I was using a region without any available gpu's - would you be willing to add some information to the docstrings regarding these fields? I would include links like this: https://cloud.google.com/compute/docs/gpus/gpu-regions-zones. At a later stage we'll then expand on these docstrings to give users all the relevant information

AlexejPenner (commented):
Hey Felix, thanks so much for this PR, this is definitely something that was missing and that we'll need to eventually make configurable on a per-step level. 

I took your code for a test-drive and I left some thoughts for you.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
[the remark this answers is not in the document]
felixthebeard: The easiest would be to 
```
from pydantic.typing import Literal
```
I removed the Literal in favor of a Tuple that allows specifying other constraints apart from the accelerator.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
felixthebeard: Added to the docstring.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
felixthebeard: Fixed.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
felixthebeard: Changed the node selector to contain the label name. See above.

PR #763, opened 2022-07-11
Prevent extra attributes when initializing StackComponents

schustmi:
## Describe changes

Previous to this PR, our StackComponent class allowed you to pass arbitrary parameters to its __init__ method and silently ignores them. This could lead to unexpected behavior when accidently mistyping an attribute name when registering a component on the CLI.

Example:
`zenml secrets-manager register NAME --flavor=aws --region=eu-west-1` (Note that the attribute on the `AWSSecretsManager` is called `region_name`, not `region` as passed in the command here).

When later running with this secrets manager, it would unexpectedly create secrets in the default region, not the one (wrongly) specified on the CLI.

After this PR, the CLI call would fail instead and let the user know that they passed a value for an attribute that doesn't exist.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/aws/secrets_managers/aws_secrets_manager.py
+++ b/src/zenml/integrations/aws/secrets_managers/aws_secrets_manager.py
@@ -27,7 +27,6 @@ from zenml.secrets_managers.base_secrets_manager import BaseSecretsManager
 
 logger = get_logger(__name__)
 
-DEFAULT_AWS_REGION = "us-east-1"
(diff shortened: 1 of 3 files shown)


strickvl (approved):
Looks good to me, though could you explain the problem context in which this was revealed to be a problem? Would help when writing release notes etc. If you could just add to the description on this PR?

schustmi:
@strickvl Updated the description

stefannica (approved):

PR #764, opened 2022-07-11
Annotation stack component and Label Studio integration

strickvl:
ScreenShot 2022-07-11 at 14 35 26

**FIRST STEP**: [Watch the mini-video](https://www.loom.com/share/045cd21242bd40d0863761c86ee6a05e) I made introducing this PR + how to get started

**Optional second step**: [Watch the second mini video](https://www.loom.com/share/72aab4908cfe4011bcd04b30f5ad6806) that shows what happens when you run the pipeline + what's happening in the pipeline itself

This PR contains an implementation of our first annotation integration. It ONLY works if you use a cloud artifact store, so you'll have to set that up. Instructions (as mentioned in the video) for all setup is listed in `examples/annotation_label_studio`.

## TODO items

- [ ] add some smaller tests (though the main integration is untestable until we implement a way to run our pipelines / tests on specific cloud infrastructure)
- [ ] finish implementing the actual inference part of the pipeline. (currently this isn't functional, though I did some work on writing the new materializers for the training steps)
- [ ] documentation + docstrings + linting + mypy etc (i.e. all the small things)
- [ ] (possibly) more default annotation options offering a wider range than the two computer vision ones I created so far.


examples/annotation_label_studio/materializers/fastai_learner_materializer.py
```diff
@@ -0,0 +1,62 @@
+import os
```
htahir1, line 1: Shouldnt this be a part of a standard fastai integration? On second thought, arent all fastai models just pytorch? So we hvae a problem like NeuralProphet here that @fa9r pointed out

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
```diff
@@ -0,0 +1,62 @@
...
+import tempfile
+from typing import Type
+
+from fastai.learner import Learner
+from fastai.vision.all import *
```
htahir1, line 6: `import *` probably a bad pattern, would encourage explicit imports

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
```diff
@@ -0,0 +1,62 @@
...
+        Only loads the model, not the checkpoint.
+
+        Args:
+            data_type: The type of the model to load.
+        Returns:
+            A loaded fastai model.
+        """
+        super().handle_input(data_type)
+
+        def is_cat(x):
```
htahir1, line 35: What does this do?

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def handle_return(self, model: Learner) -> None:
+        """Writes a fastai model along with its optimizer state.
+
+        Args:
+            model: A fastai model
+        """
+        super().handle_return(model)
+
+        def is_cat(x):
```
htahir1, line 53: What does this do?

examples/annotation_label_studio/materializers/pillow_image_materializer.py
```diff
@@ -0,0 +1,45 @@
...
+from zenml.artifacts import DataArtifact
+from zenml.io import fileio
+from zenml.logger import get_logger
+from zenml.materializers.base_materializer import BaseMaterializer
+from zenml.utils import io_utils
+
+logger = get_logger(__name__)
+
+
+class PillowImageMaterializer(BaseMaterializer):
```
htahir1, line 16: So this materializer takes a dict of files and returns a dict of images? Perhaps we should make that clear in the docstrings as it might be confusing

examples/annotation_label_studio/pipelines/continuous_training_pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+LOCAL_IMAGE_FILES = "./assets/images/"
+IMAGE_REGEX_FILTER = ".*(jpe?g|png)"
+
+logger = get_logger(__name__)
+
+
+def is_cat(x):
+    # NEEDED FOR FASTAI MODEL IMPORT (when / if we use it)
+    # return labels[x] == "cat"
+    return True
```
htahir1, line 27: This might be necessary but why is it in this file?

examples/annotation_label_studio/run.py
```diff
@@ -0,0 +1,112 @@
...
+gcs_data_sync = sync_new_data_to_label_studio(
+    config=zenml_gcs_artifact_store_sync_config,
+)
+
+s3_data_sync = sync_new_data_to_label_studio(
+    config=zenml_s3_artifact_store_sync_config,
+)
+
+
+def is_cat(x):
```
htahir1, line 90: I see this function the fourth time now

examples/annotation_label_studio/steps/batch_inference_step.py
```diff
@@ -0,0 +1,49 @@
...
+from zenml.logger import get_logger
+from zenml.steps import step
+
+LOCAL_IMAGE_FILES = "./assets/images/"
+IMAGE_REGEX_FILTER = ".*(jpe?g|png)"
+
+logger = get_logger(__name__)
+
+
+@step(enable_cache=True)
```
htahir1, line 27: No need to `enable_cache` explicitly

examples/annotation_label_studio/steps/batch_inference_step.py
```diff
@@ -0,0 +1,49 @@
...
+def batch_inference(image_dict: Dict, model: Learner) -> List:
+    """Execute batch inference on some images.
+
+    Returns a list of predictions compatible with Label Studio.
+    """
+    # TODO: Actually implement the batch inference step
```
htahir1, line 33: We might still want to retain our habit for doing `TODO[HIGH]` etc to indicate the severity

src/zenml/integrations/constants.py
```diff
@@ -25,11 +25,15 @@
...
 GRAPHVIZ = "graphviz"
+HUGGINGFACE = "huggingface"
 GREAT_EXPECTATIONS = "great_expectations"
 KUBEFLOW = "kubeflow"
+LABEL_STUDIO = "label_studio"
 KUBERNETES = "kubernetes"
 LIGHTGBM = "lightgbm"
 MLFLOW = "mlflow"
+NEURAL_PROPHET = "neural_prophet"
+# PIGEON = "pigeon"
```
htahir1, line 36: Remove ? :-)

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,646 @@
...
+
+        Raises:
+            ValueError: If the storage type is not supported.
+        """
+        if self._storage_source_already_exists(uri, config, dataset):
+            return
+        if config.storage_type == "azure":
+            if not config.azure_account_name or not config.azure_account_key:
+                logger.warn(
+                    "Authentication credentials for Azure aren't fully provided. Please update the storage synchronization settings in the Label Studio web UI as per your needs."
```
htahir1, line 588: 80 char

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,646 @@
...
+                presign_ttl=config.presign_ttl,
+                title=dataset.get_params()["title"],
+                description=config.description,
+                account_name=config.azure_account_name,
+                account_key=config.azure_account_key,
+            )
+        elif config.storage_type == "gcs":
+            if not config.google_application_credentials:
+                logger.warn(
+                    "Authentication credentials for Google Cloud Storage aren't fully provided. Please update the storage synchronization settings in the Label Studio web UI as per your needs."
```
htahir1, line 605: 80 char

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,646 @@
...
+                presign=config.presign,
+                presign_ttl=config.presign_ttl,
+                title=dataset.get_params()["title"],
+                description=config.description,
+                google_application_credentials=config.google_application_credentials,
+            )
+        elif config.storage_type == "s3":
+            if not config.aws_access_key_id or not config.aws_secret_access_key:
+                logger.warn(
+                    "Authentication credentials for S3 aren't fully provided. Please update the storage synchronization settings in the Label Studio web UI as per your needs."
```
htahir1, line 621: 80 char

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
```diff
@@ -0,0 +1,234 @@
...
+        dataset_name: Name of the dataset.
+        context: The StepContext.
+
+    Returns:
+        List of labeled data.
+
+    Raises:
+        StackComponentInterfaceError: If no active annotator could be found.
+    """
+    # TODO: have this check for new data *since the last time this step ran*
```
htahir1, line 156: Please do follow the format as before

src/zenml/steps/step_interfaces/base_annotation_step.py
```diff
@@ -0,0 +1,44 @@
...
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Base annotation step."""
+
+from abc import abstractmethod
+
+from zenml.steps import BaseStep, BaseStepConfig, StepContext
+
+
+class BaseAlerterStepConfig(BaseStepConfig):
```
htahir1, line 21: These are alerter steps but in the base annotation file?

htahir1 (changes requested):
Wow what a great PR! I love it, and i did a fast swipe over it and gave the obvious reviews first. Can do the rest later

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
strickvl: yes I have a branch open where I started working on that, but doing the whole integration with fastai + example etc was too much to do simultaneously. We can move it out later on.

And yeah it's PyTorch under the hood, but not completely so we have to handle it in a custom way.

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
strickvl: This is the way you do it with fastai. They don't do explicit imports.

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
strickvl: See above.

examples/annotation_label_studio/pipelines/continuous_training_pipeline.py
strickvl: It needs to be in the same namespace. This stuff will be cleared up in any case when I finish the actual example.

src/zenml/steps/step_interfaces/base_annotation_step.py
strickvl: Good catch. This should be deleted.

src/zenml/cli/stack_components.py
```diff
@@ -1351,12 +1362,165 @@ def flavor_group() -> None:
...
+                    "No active annotator found. Please register an annotator "
+                    "first and add it to your stack."
+                )
+                return
+
+            ctx.obj = annotator_wrapper.to_component()
+
+        @dataset.command(
+            "list",
+            context_settings={"ignore_unknown_options": True},
```
schustmi, line 1397: Any reason we need this ignore here?

src/zenml/cli/stack_components.py
```diff
@@ -1351,12 +1362,165 @@ def flavor_group() -> None:
...
+        )
+        @click.pass_obj
+        def dataset_list(annotator: "BaseAnnotator") -> None:
+            """List the available datasets.
+
+            Args:
+                annotator: The annotator stack component.
+            """
+            dataset_names = [
+                dataset.get_params()["title"]
```
schustmi, line 1408: I think this is a little LabelStudio specific? The output datasets of `any_annotator.get_datasets()` won't have a `get_params` method, maybe this method should return a list of dataset names instead?

src/zenml/cli/stack_components.py
```diff
@@ -1351,12 +1362,165 @@ def flavor_group() -> None:
...
+                for dataset in annotator.get_datasets()
+            ]
+            cli_utils.print_list_items(
+                list_items=dataset_names, column_title="DATASETS"
+            )
+
+        @dataset.command(
+            "stats", context_settings={"ignore_unknown_options": True}
```
schustmi, line 1416: Same as above, I guess we can remove this?

src/zenml/cli/stack_components.py
```diff
@@ -1351,12 +1362,165 @@ def flavor_group() -> None:
...
+            """Display statistics about a dataset.
+
+            Args:
+                annotator: The annotator stack component.
+                name: The name of the dataset.
+            """
+            projects = annotator.get_datasets()
+            # TODO: make it impossible to create new datasets with same name
+            try:
+                project = list(
```
schustmi, line 1433: Lots of LabelStudio specific code from here on

src/zenml/cli/stack_components.py
```diff
@@ -1351,12 +1362,165 @@ def flavor_group() -> None:
...
+            """Command to launch the annotation interface for a dataset.
+
+            Args:
+                annotator: The annotator stack component.
+                name: Name of the dataset
+            """
+            cli_utils.declare(
+                f"Launching the annotation interface for dataset '{name}'."
+            )
+            annotator.launch(url=annotator.get_annotation_url(name))
```
schustmi, line 1513: Maybe a check here to see if a dataset with `name` exists before we try to launch it?

src/zenml/integrations/label_studio/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    """Definition of Label Studio integration for ZenML."""
+
+    NAME = LABEL_STUDIO
+    REQUIREMENTS = ["label-studio", "label-studio-sdk"]
+
+    @classmethod
+    def activate(cls) -> None:
+        """Activates the integration."""
+        from zenml.integrations.label_studio import annotators  # noqa
+
```
schustmi, line 35:
```suggestion
```

I think the `activate` is only necessary for materializers at the moment and not for stack components.

src/zenml/integrations/label_studio/label_config_generators/label_config_generators.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import List, Tuple
+
+from zenml.enums import AnnotationTasks
+from zenml.logger import get_logger
+
+logger = get_logger(__name__)
+
+
+TASK_TO_FILENAME_REFERENCE_MAPPING = {
+    str(AnnotationTasks.IMAGE_CLASSIFICATION): "image",
```
schustmi, line 25: Any reason this needs to be a `str` key and can't just be the enum?

src/zenml/integrations/label_studio/label_studio_utils.py
```diff
@@ -0,0 +1,119 @@
...
+        url: Azure Label Studio URL.
+
+    Returns:
+        Tuple of (full URL, filename).
+    """
+    pth = urlparse(url).path
+    return f"az://{pth}", pth.split("/")[-1]
+
+
+def download_image(url: str, destination: str) -> None:
```
schustmi, line 36: This uses the `parse_azure_url` function, but the name implies it's a general utility to download images.

src/zenml/integrations/label_studio/label_studio_utils.py
```diff
@@ -0,0 +1,119 @@
...
+    """Downloads an image using fileio.
+
+    Args:
+        url: URL of the image to download.
+        destination: Path to the destination file.
+    """
+    full_url, filename = parse_azure_url(url)
+    fileio.copy(full_url, f"{destination}/{filename}")
```
schustmi, line 44: If destination is a local windows path, can we run into some issue here due to not using `os.path.join`?

src/zenml/integrations/label_studio/label_studio_utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+def get_azure_credentials() -> Tuple[Optional[str], Optional[str]]:
+    """Returns access credentials for Azure from the environment.
+
+    Returns:
+        Tuple of (account_name, account_key).
+    """
+    account_key = os.environ.get("AZURE_STORAGE_ACCOUNT_KEY")
+    account_name = os.environ.get("AZURE_STORAGE_ACCOUNT_NAME")
+    return account_name, account_key
```
schustmi, line 55: This approach unfortunately won't work when running in e.g. Kubeflow, as these environment variables (even when set locally) won't be set when executing the steps. We should probably refactor this to also include the `authentication_secret` defined on the artifact stores.

schustmi: Same for GCS/AWS below

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+    """Class to interact with the Label Studio annotation interface.
+
+    Attributes:
+        port: The port to use for the annotation interface.
+        api_key: The API key to use for authentication.
+        project_name: The name of the project to interact with.
+    """
+
+    port: int = DEFAULT_LABEL_STUDIO_PORT
+    api_key: str
```
schustmi, line 50: This should probably be stored as a secret?

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+    project_name: Optional[str]
```
schustmi, line 51: I couldn't find any place where this is used?

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+        Returns:
+            The LabelStudio Dataset object (a 'Project') for the given name.
+        """
+        # TODO: check for and raise error if client unavailable
+        ls = self._get_client()
+        dataset_id = self.get_id_from_name(dataset_name)
+        return ls.get_project(dataset_id)
+
+    def _dataset_name_to_project(self, dataset_name: str) -> Optional[Project]:
+        """Finds the project id for a specific dataset name.
```
schustmi, line 365:
```suggestion
        """Finds the project for a specific dataset name.
```

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+                use_blob_urls=config.use_blob_urls,
+                presign=config.presign,
+                presign_ttl=config.presign_ttl,
+                title=dataset.get_params()["title"],
+                description=config.description,
+                google_application_credentials=config.google_application_credentials,
+            )
+        elif config.storage_type == "s3":
+            if not config.aws_access_key_id or not config.aws_secret_access_key:
+                logger.warn(
```
schustmi, line 665: I think `logger.warn` is deprecated and `logger.warning` should be used instead. (same above)

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+        if self._storage_source_already_exists(uri, config, dataset):
+            return None
+        if config.storage_type == "azure":
+            if not config.azure_account_name or not config.azure_account_key:
+                logger.warn(
+                    "Authentication credentials for Azure aren't fully "
+                    "provided. Please update the storage synchronization "
+                    "settings in the Label Studio web UI as per your needs."
+                )
+            storage = dataset.connect_azure_import_storage(
```
schustmi, line 632: I think you could reduce a lot of duplicated code here:
```python
args  = {
   prefix=config.prefix,
   ...  # add all the identical args here
}
if config.storage_type == "azure":
    storage = dataset.connect_azure_import_storage(
        container=uri,
        ...,  # add all the azure custom args here
        **args
    )
```

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+
+        Returns:
+            A list of Azure import storage sources.
+
+        Raises:
+            ConnectionError: If the connection to the Label Studio backend is unavailable.
+        """
+        # TODO: check if client actually is connected etc
+        ls = self._get_client()
+        query_url = f"/api/storages/azure?project={dataset_id}"
```
schustmi, line 483: This seems to be identical with the two methods below apart from the `azure/gcs/s3` part of the URL? We could probably combine those right?

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
```diff
@@ -0,0 +1,234 @@
...
+        config: Step config.
+        context: Step context.
+
+    Returns:
+        The dataset name.
+
+    Raises:
+        StackComponentInterfaceError: If no active annotator could be found.
+    """
+    annotator = context.stack.annotator
```
schustmi: I'm guessing this needs a check like this:
```python
if not isinstance(annotator, LabelStudioAnnotator):
    raise TypeError()
```

schustmi: Same for the other steps below

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
```diff
@@ -0,0 +1,234 @@
...
+    # TODO: check that annotator is connected before querying it
+    dataset = annotator.get_dataset(dataset_name=dataset_name)
+    artifact_store = context.stack.artifact_store
+    if not uri.startswith(artifact_store.path):
+        raise ValueError(
+            "ZenML only currently supports syncing data passed from other ZenML steps and via the Artifact Store."
+        )
+
+    # removes the initial backslash from the prefix attribute by slicing
+    config.prefix = urlparse(uri).path[1:]
```
schustmi, line 204: Maybe a `rstrip("\")` is safer here?

schustmi (changes requested):
Awesome work! I've found a few things that could be improved, but the overall flow is really nice!

- `examples/annotation_label_studio/config.yaml` is an empty file

examples/annotation_label_studio/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+Data annotation / labeling is a core part of MLOps that is frequently left out
+of the conversation. ZenML offers a way to build continuous annotation (combining
+training and annotation into a loop) with Label Studio. This uses a combination
+of user-defined steps as well as some built-in steps that ZenML provides.
+
+## Basic guide to running this example
+
+Running this example on Azure is the quickest way to get going. You can use the
+`annotationartifactstore` storage account and its
```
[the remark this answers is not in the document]
strickvl: Yes sorry this was just for reviewing. The README + the example will get a big revamp.

examples/annotation_label_studio/README.md
```diff
@@ -0,0 +1,219 @@
...
+zenml annotator describe
+zenml annotator dataset list
+zenml annotator dataset stats
+```
+
+
+
+## 🗺 Overview
+
+The example pipeline is simple as can be. In our one and only step we access the
```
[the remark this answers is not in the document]
strickvl: Yep this will all be rewritten.

src/zenml/integrations/label_studio/label_config_generators/label_config_generators.py
strickvl: `mypy` was complaining.

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
htahir1: Ok lets leave it in then although i am at a loss why they promote this

strickvl: I think you would not like many things they do. That's why we have to be careful about the integration when we do actually do it...

examples/annotation_label_studio/materializers/fastai_learner_materializer.py
strickvl: Not sure, but that will all be ironed out + we will no longer have all those duplicates lying around when it's done.

examples/annotation_label_studio/materializers/pillow_image_materializer.py
strickvl: Done

examples/annotation_label_studio/steps/batch_inference_step.py
strickvl: Fixed

examples/annotation_label_studio/steps/batch_inference_step.py
strickvl: Done

src/zenml/integrations/constants.py
strickvl: Done

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: Fixed.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: Fixed.

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
strickvl: Fixed.

examples/annotation_label_studio/pipelines/continuous_training_pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+
+from zenml.logger import get_logger
+from zenml.pipelines import pipeline
+
+LOCAL_IMAGE_FILES = "./assets/images/"
```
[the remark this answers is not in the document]
strickvl: Something like this?

```python
LOCAL_IMAGE_FILES = os.path.join(
    Path(__file__).parent.absolute(), "assets/images"
)
```

examples/annotation_label_studio/steps/batch_inference_step.py
```diff
@@ -0,0 +1,49 @@
...
+    # below are just dummy values
+    return [
```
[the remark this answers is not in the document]
strickvl: You don't think it's enough just to point them to LS docs? This is just an example. Any time someone does something different, then they'll have to deal with a whole different output format.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,696 @@
...
+from zenml.io import fileio
+from zenml.logger import get_logger
+from zenml.utils import io_utils, networking_utils
+
+logger = get_logger(__name__)
+
+DEFAULT_LABEL_STUDIO_PORT = 8093
+
+
+class LabelStudioAnnotator(BaseAnnotator):
```
schustmi, line 40: I forgot to mention this earlier: This component should also overwrite the `validator` property of the `StackComponent` base class and make sure it's only place in stacks that it can work with. That essentially means an azure, s3 or gcp artifact store I guess?

examples/annotation_label_studio/steps/convert_annotation_step.py
```diff
@@ -0,0 +1,39 @@
...
+
+logger = get_logger(__name__)
+
+
+@step
+def convert_annotation(
+    label_studio_annotations: List[Dict[Any, Any]]
+) -> Output(image_urls=List, image_labels=List,):
+    """Converts the annotation from Label Studio to a dictionary."""
+    image_urls = [
```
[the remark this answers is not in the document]
strickvl: I don't remember that. My recollection was that in the end we decided to just make this a user step until we have a reliable converter to and from LS format for all annotation tasks.

examples/annotation_label_studio/steps/fine_tuning_step.py
```diff
@@ -0,0 +1,55 @@
...
+
+from fastai.learner import Learner
+from fastai.vision.all import *
+
+from zenml.integrations.label_studio.label_studio_utils import download_image
+from zenml.logger import get_logger
+from zenml.steps import step
+from zenml.steps.step_context import StepContext
+
+LOCAL_IMAGE_FILES = "./assets/images/"
```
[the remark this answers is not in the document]
strickvl: Removed.

examples/annotation_label_studio/steps/model_loader_step.py
```diff
@@ -0,0 +1,29 @@
...
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+
+from fastai.learner import Learner
+from huggingface_hub import from_pretrained_fastai
+
+from zenml.logger import get_logger
+from zenml.steps import step
+
+LOCAL_IMAGE_FILES = "./assets/images/"
```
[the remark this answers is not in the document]
strickvl: Removed.

src/zenml/annotators/base_annotator.py
```diff
@@ -0,0 +1,110 @@
...
+        Args:
+            *args: positional arguments.
+            **kwargs: keyword arguments.
+
+        Returns:
+            The labeled data for the given dataset.
+        """
+
+    @abstractmethod
+    def get_unlabeled_data(self, **kwargs: str) -> Any:
```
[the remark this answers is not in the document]
strickvl: I think this no longer applies. I've changed this all up to satisfy mypy.

src/zenml/cli/stack_components.py
strickvl: copy-paste artifact. now deleted.

src/zenml/cli/stack_components.py
strickvl: Good point. I added a new method.

src/zenml/cli/stack_components.py
strickvl: Yes. Done.

src/zenml/cli/stack_components.py
strickvl: Fixed now

src/zenml/cli/stack_components.py
strickvl: Added

src/zenml/integrations/label_studio/__init__.py
strickvl: Got it. Removed now.

src/zenml/integrations/label_studio/label_studio_utils.py
strickvl: You're right. Fixed.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: Correct. Removed now.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: Done.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: Good point. Reimplemented this.

strickvl: This is a cool trick! TIL.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: What do you mean combine them? Those methods are implemented by Label Studio. Do you mean we could implement a wrapper around them? We'd still have the conditional living somewhere...

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
strickvl: Good point. Fixed / done.

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
strickvl: I think I went with my way because I wasn't sure if something weird happens with Windows for these paths. Also, why did you use the backslash and not the forward slash?

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
strickvl: Ok I think I added this correctly...

examples/annotation_label_studio/steps/convert_annotation_step.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Any, Dict, List
+
+from zenml.logger import get_logger
+from zenml.steps import Output, step
+
+logger = get_logger(__name__)
+
+
+@step
+def convert_annotation(
```
safoinme, line 23: is the convert function something specific to this example only or it could be moved to some built-in function instead of implementing it as part of the example

src/zenml/annotators/base_annotator.py
```diff
@@ -0,0 +1,140 @@
...
+    @abstractmethod
+    def get_dataset_names(self) -> List[str]:
+        """Gets the names of the datasets currently available for annotation.
+
+        Returns:
+            The names of the datasets currently available for annotation.
+        """
+
+    @abstractmethod
+    def get_dataset_stats(self, dataset_name: str) -> Tuple[int, int]:
```
safoinme, line 72: is Tuple[int, int] general enough for the return of dataset stats for other annotation tools not only label studio?

safoinme (commented):
Amazing work @strickvl I was waiting for this! I really enjoyed reading this while this is just the first time reading it. I will make sure to have more time to try this and view some of the left files

examples/annotation_label_studio/README.md
```diff
@@ -0,0 +1,221 @@
...
+```shell
+# choose a username and password for your label-studio account
+label-studio reset_password --username <username> --password <password>
+# start a temporary / one-off label-studio instance to get your API key
+label-studio start -p 8094
+```
+
+Then visit
+[http://localhost:8094/](http://localhost:8094/) to log in, then visit [http://localhost:8094/user/account](http://localhost:8094/user/account) and get
+your Label Studio API key (from the upper right hand corner). You will need it
```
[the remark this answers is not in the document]
strickvl: This was really just a placeholder... Most of this README will change once I properly fix it. This was mainly just for internal use if someone wanted to try it out... So this is just temporary text.

examples/annotation_label_studio/steps/convert_annotation_step.py
strickvl: Unfortunately, due to the way that Label Studio outputs their annotations, (and how they import them), there will always need to be some kind of conversion step. I've been saying a bunch on Discord that I'd love it if ZenML was the one to build a universal conversion tool between all the formats, but I think for this PR that's too much :)

src/zenml/annotators/base_annotator.py
strickvl: Generally you get the number of unlabeled and labeled examples for a particular project. That's been the same across most of the tools I've used.

That said, I have no problem making a breaking change later :)

examples/annotation_label_studio/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+# pull example
+zenml example pull annotation_label_studio
+cd zenml_examples/annotation_label_studio
+
+# Initialize ZenML repo
+zenml init
+```
+
+### 🥞 Set up your stack for Microsoft Azure
```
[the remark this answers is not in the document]
htahir1: I also think the example should have a clear path of doing it with AWS/Azure/GCP. Perhaps you can write it in the relevant docs pages?

htahir1 (commented):
I had a look again and for the purposes of the review ill stick to the example for now. Please do let me know when its ready for review. I also feel like we should add new doc pages somewhere to be incoorporated into the new docs structure

examples/annotation_label_studio/pipelines/continuous_training_pipeline.py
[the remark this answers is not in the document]
schustmi:
```python
LOCAL_IMAGE_FILES = Path(__file__).parent.absolute() / "assets"/ "images"
```
I'd separate the final string as well to make sure it works on windows as well

examples/annotation_label_studio/steps/batch_inference_step.py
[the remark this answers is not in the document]
schustmi: No that would be fine, I'd just put this link to their docs in all the places we use a LS format

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
schustmi: Ah sorry, it should be a slash. The comment above that line talk about a backslash and that's why I wrote it.
It should also be `lstrip` instead to remove it from the beginning. Regarding windows, this should be a remote url and not have any backslashes right?

examples/annotation_label_studio/pipelines/continuous_training_pipeline.py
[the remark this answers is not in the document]
strickvl: Done.

examples/annotation_label_studio/steps/batch_inference_step.py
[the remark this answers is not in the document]
strickvl: Done

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
```diff
@@ -0,0 +1,272 @@
...
+                config
+            )
+        elif preexisting_dataset_list:
+            return cast(str, preexisting_dataset_list[0].get_params()["title"])
+        else:
+            raise StackComponentInterfaceError("No active annotator.")
+
+        return cast(str, registered_dataset.get_params()["title"])
+    else:
+        raise StackComponentInterfaceError("No active annotator.")
```
fa9r, line 149: I think this could be simplified:

```suggestion
    if annotator and annotator._connection_available():
        for dataset in annotator.get_datasets():
            if dataset.get_params()["title"] == config.dataset_name:
                return cast(str, dataset.get_params()["title"])
        
        dataset = annotator.register_dataset_for_annotation(config)
        return cast(str, dataset.get_params()["title"])
        
    raise StackComponentInterfaceError("No active annotator.")
```

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
```diff
@@ -0,0 +1,272 @@
...
+        raise TypeError(
+            "This step can only be used with the Label Studio annotator."
+        )
+    if annotator._connection_available():
+        dataset = annotator.get_dataset(dataset_name=dataset_name)
+        return dataset.get_labeled_tasks()  # type: ignore[no-any-return]
+    else:
+        raise StackComponentInterfaceError(
+            "Unable to connect to annotator stack component."
+        )
```
fa9r, line 186:
```suggestion
    raise StackComponentInterfaceError(
        "Unable to connect to annotator stack component."
    )
```

fa9r (changes requested):
Two more code suggestions that I couldn't add before since my code was outdated:

examples/annotation_label_studio/materializers/pillow_image_materializer.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    ASSOCIATED_TYPES = (dict,)
+    ASSOCIATED_ARTIFACT_TYPES = (DataArtifact,)
+
+    def handle_input(self, data_type: Type[Dict]) -> Dict:
+        """Read from artifact store"""
+        super().handle_input(data_type)
+        temp_dir = tempfile.TemporaryDirectory()
+        io_utils.copy_dir(self.artifact.uri, temp_dir.name)
```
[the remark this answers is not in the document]
strickvl: Exactly. We have this weird pattern with materializers sometimes where we have to first copy it locally and then use fileio to copy it to the artifact store. I mean to write a blog about this, and about materializers in general as I think they're a pretty critical part of ZenML that I feel like we don't explain well enough for beginners.

src/zenml/annotators/base_annotator.py
```diff
@@ -0,0 +1,140 @@
...
+    @abstractmethod
+    def get_url(self) -> str:
+        """Gets the URL of the annotation interface.
+
+        Returns:
+            The URL of the annotation interface.
+        """
+
+    @abstractmethod
+    def get_annotation_url(self, dataset_name: str) -> str:
```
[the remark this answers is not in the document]
strickvl: Done. And good point!

src/zenml/cli/stack_components.py
```diff
@@ -1351,12 +1363,167 @@ def flavor_group() -> None:
...
+
+            Args:
+                annotator: The annotator stack component.
+                dataset_name: The name of the dataset.
+            """
+            try:
+                (
+                    labeled_task_count,
+                    unlabeled_task_count,
+                ) = annotator.get_dataset_stats(dataset_name)
```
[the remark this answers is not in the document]
strickvl: Done

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+        port: The port to use for the annotation interface.
+        api_key: The API key to use for authentication.
+    """
+
+    port: int = DEFAULT_LABEL_STUDIO_PORT
+    api_key: str
+
+    FLAVOR: ClassVar[str] = LABEL_STUDIO_ANNOTATOR_FLAVOR
+
+    def validator(self) -> Optional["StackValidator"]:
```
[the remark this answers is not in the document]
strickvl: Yep. This came up when I tested it and it didn't work :) Fixed now.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+        """Gets the ID of the given dataset.
+
+        Args:
+            dataset_name: The name of the dataset.
+
+        Returns:
+            The ID of the dataset.
+        """
+        ls = self._get_client()
+        projects = ls.get_projects()
```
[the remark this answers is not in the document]
strickvl: Yeah the `ls` variable name is a convention from the label_studio documentation / guides and SDK. I thought maybe better to stick to what people are familiar with, but I agree it's not what I'd have started with myself.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+        try:
+            project = [
+                project
+                for project in projects
+                if project.get_params()["title"] == dataset_name
+            ][0]
+        except IndexError:
+            return None
+        return cast(int, project.get_params()["id"])
```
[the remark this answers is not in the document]
strickvl: Much nicer. There's even:

```python
return next((cast(int, project.get_params()["id"]) for project in projects if project.get_params()["title"] == dataset_name), None)
```

...but I don't find it as readable even if it is a bit more compact.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+                )
+            )[0]
+        except IndexError as e:
+            raise IndexError(
+                f"Dataset {dataset_name} not found. Please use `zenml annotator dataset list` to list all available datasets."
+            ) from e
+
+        labeled_task_count = len(project.get_labeled_tasks())
+        unlabeled_task_count = len(project.get_unlabeled_tasks())
+        return (labeled_task_count, unlabeled_task_count)
```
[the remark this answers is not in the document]
strickvl: Very nice. Thank you, and TIL a new pattern that is much cleaner :)

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+            Path to the root directory.
+        """
+        return os.path.join(
+            io_utils.get_global_config_directory(),
+            "annotators",
+            str(self.uuid),
+        )
+
+    @property
+    def _pid_file_path(self) -> str:
```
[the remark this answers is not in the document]
strickvl: Yes all this stuff was copied from there. It's basically identical. 100% ripe for refactoring it out...

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+            raise ValueError("`dataset_name` keyword argument is required.")
+
+        dataset_id = self.get_id_from_name(dataset_name)
+        if not dataset_id:
+            raise ValueError(
+                f"Dataset name '{dataset_name}' has no corresponding `dataset_id` in Label Studio."
+            )
+        return ls.get_project(dataset_id)
+
+    def _dataset_name_to_project(self, dataset_name: str) -> Optional[Project]:
```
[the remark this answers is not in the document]
strickvl: Even better, I can just delete this function since `get_dataset` replicates it entirely... Good catch! Thank you

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+        #     raise ValueError("`dataset_name` keyword argument is required.")
+
+        # dataset_id = self.get_id_from_name(dataset_name)
+        # if not dataset_id:
+        #     raise ValueError(
+        #         f"Dataset name '{dataset_name}' has no corresponding `dataset_id` in Label Studio."
+        #     )
+        # ls.delete_project(dataset_id)
+
+    def get_dataset(self, **kwargs: Any) -> Any:
```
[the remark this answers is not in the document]
strickvl: Unsure how to move the checks for `kwargs` out into a separate function. Would you know how I do that without some acrobatics?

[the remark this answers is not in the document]
strickvl: Just take `kwargs` as an argument for the helper function, do you mean?

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+
+        Args:
+            dataset_name: Id of the dataset.
+            output_format: Output format.
+
+        Returns:
+            A dictionary containing the converted dataset.
+        """
+        # project = self._dataset_name_to_project(dataset_name)
+        self._get_client()
```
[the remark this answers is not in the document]
strickvl: Haha good question. Nothing. It's a relic... I've deleted those two lines now.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+        project = self.get_dataset(dataset_name=dataset_name)
+        return project.export_tasks(export_type=output_format)  # type: ignore[no-any-return]
+
+    def get_labeled_data(self, **kwargs: Any) -> Any:
```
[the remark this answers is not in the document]
strickvl: As above, not quite sure how to handle the checks beyond passing `kwargs` into a helper function. Is that how you'd suggest I handle this?

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+            raise ValueError("`dataset_name` keyword argument is required.")
+
+        dataset_id = self.get_id_from_name(dataset_name)
+        if not dataset_id:
+            raise ValueError(
+                f"Dataset name '{dataset_name}' has no corresponding `dataset_id` in Label Studio."
+            )
+        return ls.get_project(dataset_id).get_labeled_tasks()
+
+    def get_unlabeled_data(self, **kwargs: str) -> Any:
```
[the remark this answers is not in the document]
strickvl: As above, not quite sure how to handle the checks beyond passing `kwargs` into a helper function. Is that how you'd suggest I handle this?

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+        Args:
+            config: Configuration for the dataset.
+
+        Returns:
+            A Label Studio Project object.
+        """
+        ls = self._get_client()
+
+        if self.get_id_from_name(config.dataset_name):
+            dataset = ls.get_project(self.get_id_from_name(config.dataset_name))
```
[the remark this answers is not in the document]
strickvl: Done.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -0,0 +1,781 @@
...
+                aws_secret_access_key=config.aws_secret_access_key,
+                aws_session_token=config.aws_session_token,
+                region_name=config.region_name,
+                s3_endpoint=config.s3_endpoint,
+                **storage_connection_args,
+            )
+        else:
+            raise ValueError(
+                f"Invalid storage type. '{config.storage_type}' is not supported by ZenML's Label Studio integration. Please choose between 'azure', 'gcs' and 'aws'."
+            )
```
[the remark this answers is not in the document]
strickvl: Not sure how this works if I still have to return something afterwards...

src/zenml/integrations/label_studio/label_config_generators/label_config_generators.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    label_config_end = "</Choices>\n</View>"
+
+    return (
+        label_config_start + label_config_choices + label_config_end,
+        label_config_type,
+    )
+
+
+def generate_basic_object_detection_bounding_boxes_label_config(
```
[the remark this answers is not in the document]
strickvl: Yeah I don't much like how we do this currently... I'll see what I can do.

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
strickvl: Thank you. Agreed. I think I sometimes end up with code like this following tussles with `mypy`, but then you're right to then take a step back afterwards and see if what you have at the end really couldn't be simplified down.

src/zenml/integrations/label_studio/steps/label_studio_standard_steps.py
strickvl: I always wonder if you don't lose something by doing it as you suggested: i.e. you lose the connection with the conditionals + the lineage that the error was raised because it failed these conditionals and not just because we got to the end of the method.

examples/annotation_label_studio/run.py
```diff
@@ -0,0 +1,112 @@
...
+)
+
+zenml_gcs_artifact_store_sync_config = LabelStudioDatasetSyncConfig(
+    storage_type="gcs",
+    label_config_type=label_config_type,
+    regex_filter=IMAGE_REGEX_FILTER,
+)
+
+zenml_s3_artifact_store_sync_config = LabelStudioDatasetSyncConfig(
+    storage_type="s3",
```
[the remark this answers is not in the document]
schustmi: Yes but that'll be almost impossible to handle I guess, we'd need to provide a way for them to give ZenML credentials for all their synced storages.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
schustmi: Yes exactly, what I meant here is:
`_get_azure_import_storage_sources`, `_get_gcs_import_storage_sources` and `_get_s3_import_storage_sources` are identical, apart from one tiny part of the URL.

We could simply have a method like
```python
def _get_import_storage_sources(..., provider: str):
    ls = self._get_client()
    query_url = f"/api/storages/{provider}?project={dataset_id}"
    ...
```

which removes the duplicated code. Probably `provider` is the wrong name here, but I hope you get the idea)

src/zenml/integrations/label_studio/label_config_generators/label_config_generators.py
schustmi: Mhm that's strange, I don't see any reason why it would complain about enum as dict keys.
I'd still suggest using `AnnotationTasks.IMAGE_CLASSIFICATION.value` here instead of converting that to a string using `str(...)`.

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
schustmi: Yes, looks good!

docs/book/mlops-stacks/annotators/label-studio.md
```diff
@@ -0,0 +1,193 @@
...
+and get your Label Studio API key (from the upper right hand corner). You will
+need it for the next step. `Ctrl-c` out of the Label Studio server that is
+running on the terminal.
+
+At this point you should register the API key with your secrets manager under a
+custom secret name, making sure to replace the two parts in `<>` with whatever
+you choose:
+
+```shell
+zenml secret register <label_studio_secret_name> --api_key="<your_label_studio_api_key>"
```
schustmi, line 92: I personally put the placeholders in capital letter to make them stick out even more, not sure if we have any convention on that though

schustmi (changes requested):

docs/book/mlops-stacks/annotators/label-studio.md
strickvl: Done.

PR #792, opened 2022-07-22
Replace the restriction to use Repository inside step with a warning

stefannica:
## Describe changes
The restriction to use Repository inside step is an antiquated measure that has become a nuisance. This PR replaces it with a warning and removes some other unneeded code designed to address this same problem.

## Pre-requisites

## Types of changes


schustmi (approved):

examples/quickstart/notebooks/quickstart.ipynb
```diff
@@ -435,7 +435,7 @@
     "@step(enable_cache=False)\n",
     "def prediction_service_loader() -> BaseService:\n",
     "    \"\"\"Load the model service of our train_evaluate_deploy_pipeline.\"\"\"\n",
-    "    repo = Repository(skip_repository_check=True)\n",
```
fa9r, line 438: why was this removed? Isn't it still nicer to suppress those warning when running the quickstart?

fa9r (approved):
Love the change! Only major question I have: are there some methods in Repository() that really should not be used from within a step and might lead to weird behavior?

examples/quickstart/notebooks/quickstart.ipynb
fa9r: same for the other change below

stefannica: you had a valid point here. I didn't want to expose users to this hidden variable, so I removed it from here. But the correct way to hide it is to detect whether a step has caching enabled _from within the step_ and only show that warning if that is the case. Which is what I did in the last commit.

fa9r (approved):

PR #794, opened 2022-07-25
Per-step resource configuration

schustmi:
## Describe changes

Some steps of your machine learning pipeline might be more resource-intensive
and require special hardware to execute. This PR enables this by allowing the specification of per-step resources as follows:

```python
from zenml.steps import step, ResourceConfiguration
@step(resource_configuration=ResourceConfiguration(cpu_count=8, gpu_count=2))
def training_step(...) -> ...:
    # train a model
```

This generic language is currently limited to the `KubeflowOrchestrator` and `VertexOrchestrator`, but will be extended in the upcoming weeks to support the `KubernetesOrchestrator`.
Our existing set of step operators unfortunately don't allow specific resource configurations but instead work with `instance/machine types`.

## Pre-requisites

## Types of changes


src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
```diff
@@ -55,14 +57,14 @@ class VertexStepOperator(BaseStepOperator, GoogleCredentialsMixin):
...
-        accelerator_count: [Optional] Defines number of accelerators to be
+        accelerator_type: Accelerator type from list: https://cloud.google.com/vertex-ai/docs/reference/rest/v1/MachineSpec#AcceleratorType
+        accelerator_count: Defines number of accelerators to be
             used for the job.
-        machine_type: [Optional] Machine type specified here: https://cloud.google.com/vertex-ai/docs/training/configure-compute#machine-types
-        base_image: [Optional] Base image for building the custom job container.
-        encryption_spec_key_name: [Optional]: Encryption spec key name.
+        machine_type: Machine type specified here: https://cloud.google.com/vertex-ai/docs/training/configure-compute#machine-types
+        base_image: Base image for building the custom job container.
+        encryption_spec_key_name: Encryption spec key name.
```
AlexejPenner, line 67: Why do these stop being optional, are we replacing them with sane defaults?

AlexejPenner (commented):
lgtm

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
schustmi: They stay optional, but it's already specified in the type annotation and shouldn't be in the docstring.

See for example here for the duplicated `Optional`:
https://apidocs.zenml.io/0.11.0/api_docs/integrations/#zenml.integrations.gcp.step_operators.vertex_step_operator.VertexStepOperator

src/zenml/integrations/github/orchestrators/github_actions_orchestrator.py
```diff
@@ -469,6 +469,14 @@ def prepare_or_run_pipeline(
 
         jobs = {}
         for step in sorted_steps:
+            if not step.resource_configuration.empty:
```
schustmi, line 472:
```suggestion
            if self.requires_resources_in_orchestration_environment(step):
```

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -443,22 +443,26 @@ def _construct_kfp_pipeline() -> None:
                     container_op.after(upstream_container_op)
 
                 # Set optional CPU, RAM and GPU constraints for the pipeline
-                if self.cpu_limit is not None:
-                    container_op = container_op.set_cpu_limit(self.cpu_limit)
+                step_resources = step.resource_configuration
```
wjayesh, line 446: Can this configuration for CPU, memory and GPU be a separate function that you call from here? I think that'll make it a bit cleaner to look at.

wjayesh: Oh, I see something similar in the `kubeflow_orchestrator` case in the form of a `_configure_container_resources` function.

wjayesh (commented):
Great work on the PR! Just have a minor suggestion.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
schustmi: Makes sense, I'll refactor this into a separate method similar to the kubeflow one

htahir1 (changes requested):
This is awesome work! I would just ask to update the examples of the corresponding orchestrators with resource configurations to indicate to users how to use them! Perhaps we can even add an example to show the interoperability of it? This is a big user facing feature and would to expedite it.

schustmi:
@htahir1 I mentioned it in the relevant examples. IMO we should hold off on adding a separate example until we overhaul the general pipeline config, wdyt?

wjayesh (approved):
LGTM!

src/zenml/step_operators/step_executor_operator.py
```diff
@@ -128,6 +129,50 @@ def _resolve_user_modules(
 
         return main_module_path, step_source_path
 
+    @staticmethod
+    def _get_step(
+        pipeline_node: pipeline_pb2.PipelineNode,
+    ) -> "BaseStep":
```
AlexejPenner, line 135: How is this related to the Step Resource story? Drive-by improvement?

AlexejPenner (approved):
LGTM

src/zenml/step_operators/step_executor_operator.py
schustmi: I use the step instance to fetch the resource config below. But actually that whole code can go now that the resource configuration is also in the MLMD contexts. I'll remove it :)

PR #796, opened 2022-07-25
Fix the mismatch KFP version between Kubeflow and GCP integration

safoinme:
## Describe changes
Use the same KFP version between GCP and Kubeflow Integration. Now creating a docker image with both integrations fails.

## Pre-requisites

## Types of changes


schustmi (approved):

fa9r (changes requested):
I think this will create a new dependency conflict between Kubeflow and (Seldon, Kubernetes) since all of them use `kubernetes` package under the hood. Maybe @stefannica knows more about this?

schustmi:
> 

Isn't seldon etc all mostly tested with the KubeflowOrchestrator (which isn't changed in this PR) and not at all with Vertex (which now should use the same, tested version of KFP)?

fa9r (approved):

fa9r:
You're absolutely right, for some reason I thought the KFP version changed, nevermind 🤪

AlexejPenner (approved):
Tested on vertex and works for me - LGTM

PR #799, opened 2022-07-25
Allow standard library types in steps 

stefannica:
## Describe changes
Some types in the standard Python library, such as `collections.OrderedDict` were not correctly identified as built-in types in the source utils:

* they do have an associated source file and python module
* they are not part of the site packages or the user site packages
* they are part of the standard OS-wide Python distribution files

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/source_utils.py
+++ b/src/zenml/utils/source_utils.py
@@ -33,6 +33,7 @@ import site
 import sys
 import types
 from contextlib import contextmanager
+from distutils.sysconfig import get_python_lib
 from types import (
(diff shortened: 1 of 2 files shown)


schustmi (commented):
Nice catch! I think this could benefit from some small unit test as well to make sure we don't break it in the future

stefannica:
> Nice catch! I think this could benefit from some small unit test as well to make sure we don't break it in the future

Done !

schustmi (approved):
LGTM

PR #804, opened 2022-07-27
Improve error message for subscripted generic step output types.

fa9r:
## Describe changes
I modified the step interface to raise an error during initialization if subscripted generics were specified as a step output, e.g., `Outputs(a=List[str], b=Dict[str, str])` should be `Outputs(a=List, b=Dict)` instead.

### Previous behavior:
Step would get created, but raise `TypeError: Subscripted generics cannot be used with class and instance checks` at runtime.

### New behavior:
A more concise error is raised during step initialization, e.g.: `StepInterfaceError: Subscripted generics cannot be used as step outputs. For step 'load_image_data', use images=typing.Dict instead of images=typing.Dict[str, PIL.Image.Image].`

## Pre-requisites

## Types of changes

### DIFF
--- a/.pyspelling-ignore-words
+++ b/.pyspelling-ignore-words
@@ -1092,6 +1092,7 @@ subnetIds
 subnets
 subpath
 subprocess
+subscripted
 subtypes
 superclass
 superclasses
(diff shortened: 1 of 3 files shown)


fa9r:
Changed this back to draft since the current solution still has various issues:
- doesn't check single outputs
- isn't compatible with Python 3.9
- doesn't have tests
- fails spelling and linting checks

fa9r:
Changed to draft again since I might have just figured out how to support subscripted returns, which would be much better than raising an error.

fa9r:
Closing in favor of #806.

PR #805, opened 2022-07-27
retrieve pipeline requirement within running step

safoinme:
## Describe changes
This code is to get running pipeline requirements with a step environment. 

This will be mainly used for the custom code deployment and particularly building docker image for custom deployment when a local orchestrator is used! Unlike the remote orchestrator where we can use the same created docker image to run the pipeline on the deployment, the local orchestrator requires building a new docker image and this build happens when the custom deployment step is running, for that we need the requirement to be already within step environment.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/steps/utils.py
+++ b/src/zenml/steps/utils.py
@@ -58,11 +58,13 @@ from tfx.types import component_spec
 from tfx.types.channel import Channel
 from tfx.utils import json_utils
 
+import zenml
 from zenml.artifacts.base_artifact import BaseArtifact
 from zenml.exceptions import MissingStepParameterError, StepInterfaceError
 from zenml.io import fileio
 from zenml.logger import get_logger
 from zenml.materializers.base_materializer import BaseMaterializer
+from zenml.repository import Repository
 from zenml.steps.base_step_config import BaseStepConfig
 from zenml.steps.step_context import StepContext
 from zenml.steps.step_environment import StepEnvironment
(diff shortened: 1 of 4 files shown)


src/zenml/steps/utils.py
```diff
@@ -452,6 +454,33 @@ def check_output_types_match(
                 f"{getattr(self, PARAM_STEP_NAME)}"
             )
 
+    def _collect_requirements(
```
schustmi, line 457: Can we refactor this in some function which can be used both here and in the step executor operator (which has code very similar to this)?

safoinme: @schustmi  I think it's possible but somehow I would want to leave it like this for the moment as refactor this, would change some stuff related to the step that would require some more testing :)

tests/unit/test_environment.py
```diff
@@ -49,6 +49,7 @@ def test_step_is_running():
         pipeline_run_id="run_id",
         step_name="step",
         cache_enabled=True,
+        pipeline_requirements=["zenml"],
```
fa9r, line 52: Why is this needed? Shouldn't line 710 of steps/utils.py already take care of this?

fa9r (approved):
The code itself looks good to me. 

One question I have: when creating a pipeline you can currently specify `requirements=...` and `required_integrations=...` separately. Are both of those already included in the output of `collect_requirements()`? maybe @schustmi knows more about that?

Also, are these changes still valid after #794 gets merged?

TL;DR: code looks good, but should wait for @schustmi 's review before merging :)

tests/unit/test_environment.py
safoinme: I don't think it does because that's just for testing of StepEnvironment, it tries to create a new object from that class. also, that list could be just empty rather than have "zenml" on it!

schustmi:
@fa9r @safoinme My PR will indeed change the way those requirements are configured and stored, but we can go ahead and merge this and I'll take care of the rest.

tests/unit/test_environment.py
schustmi: I think an empty list would be nicer indeed just to show that this is a "dummy" value.

schustmi (approved):

PR #807, opened 2022-07-28
Add zenml stack recipe CLI commands

wjayesh:
### Stack Recipes CLI commands

This PR introduces the `zenml stack recipe` CLI commands that allow the following functionalities.

  - `zenml stack recipe list`: List all available stack recipes.
  - `zenml stack recipe pull <recipe-name>`: Pull a recipe to your local filesystem.
  - `zenml stack recipe deploy <recipe-name>`: Deploy a specific stack recipe and auto-import the resources created, as a ZenML stack.
  - `zenml stack recipe destroy <recipe-name>`: Destroy all resources created by a recipe.
  - `zenml stack recipe clean`: Deletes the pulled recipes from the local filesystem.

Extensive error messages are put in place to guide users if the commands fail. This has been done keeping in mind that the feature would be in alpha, and things could break unexpectedly.

### Terraform wrapper

A `Terraform` class is implemented as a wrapper around the `terraform` CLI. It adds the following features:
- Check if the user has `terraform` installed. Raise an exception and point them to the installation page if they don't have it.
- Implement an `apply` command that performs a `terraform init` and then `terraform apply` on the recipe. The `init` is only performed once to save time.
- A `destroy` command that performs a simple `terraform destroy` but can potentially include a `--force` parameter in the future to force deletion of resources. Keeping it out for now, in the absence of tests, to avoid unintended consequences.
- Option to set log level for the `terraform` command executions.

**Features yet to be added:**
- Ability to run these commands in the background using the ZenML service concept.
- Check user credentials to verify if they have permission to execute the recipe.

### Minor changes

- While importing a stack, there's now an option to force-import a YAML configuration with a different version than the current ZenML version.

## Pre-requisites

## Types of changes


src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,840 @@
...
+    @property
+    def recipes_destroy_bash_script(self) -> str:
+        """Path to the bash script that destroys the recipe.
+
+        Returns:
+            Path to the bash script that destroys the recipe.
+        """
+        return os.path.join(self.path, RECIPE_DESTROY_SCRIPT)
+
+    # @property
```
safoinme, line 91: can this be removed?

src/zenml/cli/stack.py
```diff
@@ -986,11 +986,10 @@ def up_stack() -> None:
 )
 def down_stack(force: bool = False, old_force: bool = False) -> None:
     """Suspends resources of the active stack deployment.
-
     Args:
         force: Deprovisions local resources instead of suspending them.
         old_force: DEPRECATED: Deprovisions local resources instead of
-            suspending them. Use `-y/--yes` instead.
+            suspending them. Use `-f/--force` instead.
```
safoinme, line 992: is this the right message or the opposite "Use `-y/--yes` instead."

wjayesh: I was confused by this too but Michael mentioned that for stack down specifically, we are going with -f

src/zenml/cli/stack_recipes.py
wjayesh: yes, I'll remove it. Thanks!

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,840 @@
...
+    #             "of the stack recipe to find out more."
+    #         )
+
+    def is_present(self) -> bool:
+        """Checks if the stack_recipe exists at the given path.
+
+        Returns:
+            True if the stack_recipe exists at the given path, else False.
+        """
+        return fileio.exists(str(self.path)) and fileio.isdir(str(self.path))
```
safoinme, line 97: i. think `fileio.isdir(str(self.path))` is enough here. as it would verify if the file exists or not

bcdurak (commented):
The new CLI commands look great! Looking forward to using them in the future. I left a few small comments.

src/zenml/cli/stack.py
```diff
@@ -1141,19 +1141,49 @@ def _import_stack_component(
 @stack.command("import", help="Import a stack from YAML.")
 @click.argument("stack_name", type=str, required=True)
 @click.argument("filename", type=str, required=False)
+@click.option(
+    "--yes",
+    "-y",
+    "force",
+    is_flag=True,
```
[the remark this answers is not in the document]
wjayesh: Yes, it was intentional. I think @schustmi mentioned that the `down_stack` function was an exception to this convention.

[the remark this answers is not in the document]
schustmi: Correct, this is intentional. `--yes` as an option only makes sense if it's asking for confirmation, but in the case of `zenml stack down`, the `--force` actually completely changes the behavior of the command.

src/zenml/cli/__init__.py
```diff
@@ -823,5 +823,6 @@
 from zenml.cli.server import *  # noqa
 from zenml.cli.stack import *  # noqa
 from zenml.cli.stack_components import *  # noqa
+from zenml.cli.stack_recipes import *  # noqa
```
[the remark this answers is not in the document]
schustmi: I'd also add that general docs on this would be super helpful

src/zenml/cli/stack.py
```diff
@@ -1141,19 +1141,49 @@ def _import_stack_component(
...
+    help="Force the import of stack components irrespective of ZenML version matching.",
+)
+@click.option(
```
schustmi, line 1151: I think this entire option should be removed to be honest. We only added the deprecated options because we previously had the old option and didn't want it to break for users. But when implementing this functionality for some entirely new feature, there is no need for this.

Similar to my message above, I'm also not sure whether `--force` or `--yes` would be the right flag here. In this case it's not automatically confirming something, but instead "forcing" the import even if the versions don't match. Personally something like `--ignore-version-mismatch` would be my preferred option, but I think others might disagree and prefer the ease of use of just using the same flag everywhere.

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,806 @@
...
+                    )
+                raise
+        else:
+            raise FileNotFoundError(
+                "Bash File(s) to run recipes not found at"
+                f"{stack_recipe_runner}"
+            )
+
+        # Telemetry
+        if identifier == "deploy":
```
schustmi, line 135: I'm not a fan of passing in some `identifier` as a string here to be honest. I think making this an enum like `RecipeExecutionMode` would be a little nicer, but seeing as this is only used to track analytics, I don't see any reason to even pass this and not call `track_event` in the `deploy` and `destroy` CLI functions below?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,806 @@
...
+                pull,
+                stack_recipe_name=stack_recipe_name,
+                path=path,
+                force=force,
+            )
+
+        stack_recipe_runner = (
+            [] if shell_executable is None else [shell_executable]
+        ) + [local_stack_recipe.recipes_deploy_bash_script]
+        try:
```
schustmi, line 728: I think the `track_event` should be called here instead of in the `local_stack_recipe.run_stack_recipe` method.

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,806 @@
...
+    "--shell-executable",
+    "-x",
+    type=click.Path(exists=True),
+    required=False,
+    envvar=SHELL_EXECUTABLE,
+    help="Manually specify the path to the executable that runs .sh files. "
+    "Can be helpful for compatibility with Windows or minimal linux "
+    "distros without bash.",
+)
+@pass_git_stack_recipes_handler
```
schustmi, line 673: An option to automatically import the stack would be nice here, maybe something like `zenml stack recipe deploy <RECIPE_NAME> --import` (Or even let users specify an import name)

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,806 @@
...
+            )
+
+        stack_recipe_runner = (
+            [] if shell_executable is None else [shell_executable]
+        ) + [local_stack_recipe.recipes_deploy_bash_script]
+        try:
+            local_stack_recipe.run_stack_recipe(
+                stack_recipe_runner=stack_recipe_runner,
+                identifier=RECIPE_DEPLOY_IDENTIFIER,
+            )
```
schustmi, line 732: If the user chose to not import the stack automatically, we should log the location of the stack yaml as well as how to import it here.

schustmi (changes requested):

src/zenml/cli/stack_recipes.py
wjayesh: The identifier isn't passed as a string but by using a constant value. I agree that the string shouldn't be used in the comparison here. How would you feel if I made the following change?
```py
if identifier == RECIPE_DEPLOY_IDENTIFIER
```
The identifier is also used to make error messages more friendly and specific above as this function is called by both deploy and destroy functions.

schustmi: Yes that would be similar to the enum approach, definitely better than the previous solution and fine with me.

I think passing the identifier could still be avoided by doing something like this:
```python
def deploy(recipe_name):
    try:
        local_stack_recipe.run_stack_recipe(...)
    except SomeError as e:
        cli_utils(f"Failed to deploy recipe {recipe_name}. Reason: {e}")
```
In any case, as long as the string constants are gone I can live with it :)

wjayesh: That's a really elegant solution. Thanks so much for the idea! :)

src/zenml/cli/stack.py
wjayesh: I totally agree with the `--ignore-version-mismatch` idea. I would prefer something that is explicit about what it does too.

src/zenml/cli/stack_recipes.py
wjayesh: Hamza mentioned that the import should be the intended behaviour so I had made that the default and added `--no-import` as an option instead.

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,871 @@
...
+                subprocess.check_call(
+                    call,
+                    cwd=str(self.path),
+                    shell=click._compat.WIN,
+                    env=os.environ.copy(),
+                )
+            except RuntimeError:
+                raise NotImplementedError(
+                    "Please raise an issue at "
+                    "https://github.com/zenml-io/mlops-stacks"
```
htahir1, line 128: The repo should be refactored out to a constant

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,871 @@
...
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 38:
+                    raise NotImplementedError(
+                        "Please raise an issue at "
+                        "https://github.com/zenml-io/mlops-stacks"
+                    )
+                raise
+        else:
+            raise FileNotFoundError(
+                "Bash File(s) to run recipes not found at"
```
htahir1, line 139: What should the user do at this point? Perhaps guide them there

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,871 @@
...
+                )
+            except RuntimeError:
+                raise NotImplementedError(
+                    "Please raise an issue at "
+                    "https://github.com/zenml-io/mlops-stacks"
+                )
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 38:
+                    raise NotImplementedError(
+                        "Please raise an issue at "
```
htahir1, line 133: Its not implemented so raising an issue should not be the behavior. Maybe rather point them to the hellonext?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,871 @@
...
+
+        Raises:
+            NotImplementedError: If the stack_recipe hasn't been implement yet.
+            FileNotFoundError: If the stack_recipe runner script is not found.
+            subprocess.CalledProcessError: If the stack_recipe runner script fails.
+        """
+        if all(map(fileio.exists, stack_recipe_runner)):
+            call = stack_recipe_runner
+            try:
+                subprocess.check_call(
```
htahir1, line 119: As suggested in the PRD I would re-think should subprocess to call a bash script. Id rather use a the `python-terraform` library and potentially do the calls here? WDYT

htahir1 (commented):
Left some minor feedback as I read this over the weekend again. Ignore it if you have already addressed it

src/zenml/cli/stack_recipes.py
wjayesh: I think we can switch to the library. The only thing we lose is the ability to customize commands per recipe and considering that only the `aws-minimal` recipe requires additional clean-up at the moment, I'm thinking we can just add that condition in the python code itself.

src/zenml/cli/stack_recipes.py
wjayesh: This not implemented error points to the absence of a `run_recipe` script within an already existing stack recipe. So, I feel an issue should be created to investigate why the script isn't implemented or why it couldn't be located by the CLI code. 
Raising a feature request on hellonext might not be ideal since a stack recipe already exists and only a modification/fix is needed.

wjayesh: True. Moreover, I think a not implemented error here doesn't make a lot of sense too because the run command is called on an existing stack recipe so it should exist. I'll remove the `NotImplementedError` and just keep the `CalledProcessError`.

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
+Optionally, you can specify the relative path at which you want to install the
+stack recipe(s). Use the `-p` or `--path` flag.
+```bash
+zenml stack recipe pull <stack-recipe-name> --path=<PATH>
+```
+By default, all recipes get downloaded under a directory by the name `zenml_stack_recipes`.
+
+To deploy a recipe, use the `deploy` command. Before running deploy, review the `locals.tf`
+file for configuring non-sensitive variables and the `values.tfvars` file to add sensitive
+information like access keys and passwords.
```
AlexejPenner, line 855:
```suggestion
To deploy a recipe, use the `deploy` command. Before running deploy, review the 
`zenml_stack_recipes/<stack-recipe-name>/locals.tf` file for configuring non-sensitive 
variables and the `zenml_stack_recipes/<stack-recipe-name>/values.tfvars` file to 
add sensitive information like access keys and passwords.
```

src/zenml/cli/stack.py
```diff
@@ -1143,16 +1143,28 @@ def _import_stack_component(
...
 ) -> None:
     """Import a stack from YAML.
 
     Args:
         ctx: The click context.
         stack_name: The name of the stack to import.
         filename: The filename to import the stack from.
+        ignore_version_mismatch: Import stack components even if
+            the installed version of ZenML is different from the
+            one specified in the stack YAML file.
```
AlexejPenner, line 1167: It feels strange that the yaml file generated has a zenml version at all or even a zenml version different for the zenml package that calls this command, is there no way of avoiding this?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
+#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
AlexejPenner, line 1:
```suggestion
#  Copyright (c) ZenML GmbH 2022. All Rights Reserved.
```

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
 ```
 
+Deploying cloud resources using Stack Recipes
+-----------------------------------------------
+
+Stack Recipes allow you to quickly deploy fully-fledged MLOps stacks with just a few
+commands. Each recipe uses Terraform modules under the hood and once executed can set up
+a ZenML stack, ready to run your pipelines!
+
+A number of stack recipes are already available at the `mlops-stacks` repository. List them
```
strickvl, line 827:
```suggestion
A number of stack recipes are already available at [the `mlops-stacks` repository](https://github.com/zenml-io/mlops-stacks/). List them
```

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
+
+```bash
+zenml stack recipes list
+```
+
+If you want to pull any specific recipe to your local system, use the `pull` command:
+```bash
+zenml stack recipe pull <stack-recipe-name>
+```
+Not specifying a name will pull all the recipes.
```
strickvl, line 838:
```suggestion
If you don't specify a name, `zenml stack recipe pull` will pull all the recipes.
```

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
+
+If you notice any inconsistency with the locally pulled version and the GitHub repository,
```
strickvl, line 840:
```suggestion
If you notice any inconsistency with the locally-pulled version and the GitHub repository,
```

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
+```bash
+zenml stack recipe pull <stack-recipe-name> -y
+```
+
+Optionally, you can specify the relative path at which you want to install the
+stack recipe(s). Use the `-p` or `--path` flag.
+```bash
+zenml stack recipe pull <stack-recipe-name> --path=<PATH>
+```
+By default, all recipes get downloaded under a directory by the name `zenml_stack_recipes`.
```
strickvl, line 851:
```suggestion
By default, all recipes get downloaded under a directory called `zenml_stack_recipes`.
```

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
+zenml stack recipe deploy <stack-recipe-name>
+```
+
+Running deploy without any options will create a new ZenML stack with the same name as
+the stack recipe name. Use the `--stack-name` option to specify your own name.
+```bash
+zenml stack recipe deploy <stack-recipe-name> --stack-name=my_stack
+```
+
+If you wish to review the stack information from the newly generated resources before
```
strickvl, line 866:
```suggestion
If you wish to review the stack information from the newly-generated resources before
```

src/zenml/cli/__init__.py
```diff
@@ -817,6 +817,71 @@
...
+```
+This will still create a stack YAML configuration file but will not auto-import it. You can
+make any changes you want to the configuration and then run `zenml stack import` manually.
+
+To remove all resources created as part of the recipe, run the `destroy` command.
+```bash
+zenml stack recipe destroy <stack-recipe-name>
+```
+
+For deleting all the recipe files from your system, you can use the `clean` command.
```
strickvl, line 879:
```suggestion
To delete all the recipe files from your system, you can use the `clean` command.
```

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+        previous_run_dir = os.path.join(self.tf.working_dir, ".ignoreme")
+        if fileio.exists(previous_run_dir):
+            logger.info(
+                "Terraform already initialized, "
+                "terraform init will not be executed."
+            )
+        else:
+            ret_code, _, _ = self.tf.init(capture_output=False)
+            if ret_code != 0:
+                raise RuntimeError("The command 'terraform init' failed.")
```
AlexejPenner, line 104:
```suggestion
                raise RuntimeError("The command 'terraform init' failed.")
```

Is there any way we could tell the user what to do to make it work?

src/zenml/cli/stack.py
```diff
@@ -1171,14 +1183,22 @@ def import_stack(
...
-            f"Cannot import stacks from other ZenML versions. "
-            f"The stack was created using ZenML version "
-            f"{data['zenml_version']}, you have version "
-            f"{zenml.__version__} installed."
-        )
+        if ignore_version_mismatch:
+            cli_utils.warning(
+                f"The stack that will be installed is using ZenML version "
+                f"{data['zenml_version']}. You have version "
+                f"{zenml.__version__} installed. Some components might not"
```
strickvl, line 1192:
```suggestion
                f"{zenml.__version__} installed. Some components might not "
```

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+        _, _, _ = self.tf.apply(
+            var=vars,
+            input=False,
+            capture_output=False,
+            raise_on_error=True,
+        )
+
+        # return the path of the stack yaml file
+        _, stack_file_path, _ = self.tf.output("stack-yaml-path")
+        return str(stack_file_path)
```
AlexejPenner, line 120: Any chance this could fail in a predictable way that we might want to catch?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+            ret_code, _, _ = self.tf.init(capture_output=False)
+            if ret_code != 0:
+                raise RuntimeError("The command 'terraform init' failed.")
+            fileio.mkdir(previous_run_dir)
+
+        # get variables from the recipe as a python dictionary
+        vars = self._get_vars(self.tf.working_dir)
+
+        # once init is successful, call terraform apply
+        _, _, _ = self.tf.apply(
```
strickvl, line 111:
```suggestion
        self.tf.apply(
```

Do we need to assign to anything, if nothing is being used from what is returned?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+            FileNotFoundError: if the values.tfvars.json file is not
+                found in the stack recipe.
+        """
+        import json
+
+        variables_file_path = os.path.join(path, VARIABLES_FILE)
+        if not fileio.exists(variables_file_path):
+            raise FileNotFoundError(
+                "The file values.tfvars.json was not found in the "
+                "recipe's directory. Please verify if it exists."
```
AlexejPenner, line 142: Maybe print out the recipes directory here. The user might not be aware what the location of the recipe is.

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+    def is_stack_recipe(self, stack_recipe_name: Optional[str] = None) -> bool:
+        """Checks if the supplied stack_recipe_name corresponds to a stack_recipe.
+
+        Args:
+            stack_recipe_name: The name of the stack_recipe to check.
+
+        Returns:
+            Whether the supplied stack_recipe_name corresponds to an stack_recipe.
+        """
+        stack_recipe_dict = {e.name: e for e in self.stack_recipes}
```
strickvl, line 411: what is `e`? Maybe `recipe` instead?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+        stack_recipe_name: The name of the stack_recipe.
+        force: Force the redownload of the stack_recipes folder to the ZenML config
+            folder.
+        path: The path at which you want to install the stack_recipe(s).
+    """
+    cli_utils.warning(
+        "The stack recipes CLI is in alpha and actively being developed. "
+        "Please avoid running mission-critical workloads on resources deployed "
+        "through these commands."
+    )
```
strickvl, line 653: Maybe this message (and elsewhere where it is used) can be turned into a CONSTANT somewhere to reduce the duplication?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+        {"stack_recipe_name": stack_recipe.name}
+        for stack_recipe in git_stack_recipes_handler.get_stack_recipes()
+    ]
+    cli_utils.print_table(stack_recipes)
+
+    cli_utils.declare("\n" + "To get the latest list of stack recipes, run: ")
+    text = Text("zenml stack recipe pull -y", style="markdown.code_block")
+    cli_utils.declare(text)
+
+    cli_utils.declare("\n" + "To pull any individual stack_recipes, type: ")
```
strickvl, line 542:
```suggestion
    cli_utils.declare("\n" + "To pull any individual stack recipe, type: ")
```

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+        )
+
+
+@stack_recipe.command(help="Find out more about a stack recipe.")
+@pass_git_stack_recipes_handler
+@click.argument("stack_recipe_name")
+def info(
+    git_stack_recipes_handler: GitStackRecipesHandler, stack_recipe_name: str
+) -> None:
+    """Find out more about an stack recipe.
```
strickvl, line 589:
```suggestion
    """Find out more about a stack recipe.
```

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+    stack_recipe_name: str,
+    path: str,
+    force: bool,
+    no_import: bool,
+    log_level: str,
+    stack_name: Optional[str],
+) -> None:
+    """Run the stack_recipe at the specified relative path.
+
+    `zenml stack_recipe pull stack_recipe_name` has to be called with the same relative
```
AlexejPenner, line 757:
```suggestion
    `zenml stack_recipe pull <STACK_RECIPE_NAME>` has to be called with the same relative
```

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+                    stack_name=stack_name,
+                    filename=stack_yaml_file,
+                    ignore_version_mismatch=True,
+                )
+
+                cli_utils.declare(
+                    "Please consider creating any secrets that your stack "
+                    "components like the metadata store might need. "
+                    "You can inspect the fields of a stack component by "
+                    "running a describe command on them."
```
AlexejPenner, line 859: Are these secrets still a part of the terraform outputs? If yes it might be great to instruct the user how to find the secrets as well

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+    "stack_recipe folder and force installs all necessary integration "
+    "requirements.",
+)
+@click.option(
+    "--stack-name",
+    "-n",
+    type=click.STRING,
+    required=False,
+    help="Set a name for the ZenML stack that will be imported from the YAML"
+    "configuration file which gets generated after deploying the stack recipe."
```
strickvl, line 725:
```suggestion
    help="Set a name for the ZenML stack that will be imported from the YAML "
    "configuration file which gets generated after deploying the stack recipe. "
```

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+    "--log-level",
+    type=click.Choice(
+        ["TRACE", "DEBUG", "INFO", "WARN", "ERROR"], case_sensitive=False
+    ),
+    help="Choose one of TRACE, DEBUG, INFO, WARN or ERROR (case insensitive) as "
+    "log level for the destroy operation.",
+    default="ERROR",
+)
+@pass_git_stack_recipes_handler
+# @pass_tf_client
```
strickvl, line 906: Can be deleted?

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1016 @@
...
+@click.pass_context
+def destroy(
+    ctx: click.Context,
+    git_stack_recipes_handler: GitStackRecipesHandler,
+    # tf_client: Terraform,
```
strickvl, line 911: Same?

strickvl (commented):
Just some small comments here.

AlexejPenner (commented):
Great work, It might be worth a thought removing all the non-cli related code into a separate file, but I do understand that this is a similar case to the examples which we have handled similarly until now. While reading though the code I wasn'T entirle sure how the terraform outputs are parsed into the stack yaml, specifically the secrets. How would that be handled in the current version of the code?

src/zenml/cli/stack.py
```diff
@@ -1141,19 +1140,49 @@ def _import_stack_component(
...
+    help="Force the import of stack components irrespective of ZenML version matching.",
+)
+@click.option(
+    "--force",
+    "-f",
+    "old_force",
+    is_flag=True,
+    help="DEPRECATED: Force the import of stack components irrespective of ZenML version matching"
+    "Use `-y/--yes` instead.",
+)
```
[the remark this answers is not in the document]
wjayesh: You're right. I had replaced this with a new flag which is more explicit in what it does, `--ignore-version-mismatch`. What do you think?

src/zenml/cli/stack.py
wjayesh: The generated YAML needs to have a version number because syntax for some stack components might differ across versions, if we don't have entirely new components altogether. As for why it differs from the installed zenml version: the version on the stack YAML file comes from the stack recipe (it shows that the given yaml is valid for this specific version at least) and is therefore independent of the installed zenml on the user end.

src/zenml/cli/stack_recipes.py
wjayesh: Yes, I agree. I catch this error in the `deploy` and `destroy` CLI functions and have added extensive instructions there on potential actions to take.

src/zenml/cli/stack_recipes.py
wjayesh: The `raise_on_error` option in the `apply` function raises a `TerraformCommandError` if the function fails. I catch this in the `deploy` CLI function with some information on what next steps the user should take. This info is derived from the common problems I've faced when ruuning the recipes.

src/zenml/cli/stack_recipes.py
wjayesh: Great idea!

src/zenml/cli/stack_recipes.py
wjayesh: That would be really convenient, I agree. I'll add instructions on how to access them.

wjayesh:
> While reading though the code I wasn'T entirle sure how the terraform outputs are parsed into the stack yaml, specifically the secrets. How would that be handled in the current version of the code?

@AlexejPenner The parsing of the terraform outputs into a stack YAML is done as part of a stack recipe itself and is not handled here at all. The secret outputs can be accessed individually by running `terraform output <var-name>` inside the stack recipe directory. I've added the instructions for it in the code, as you suggested. 
I'm planning to create secrets directly as part of the recipe going forward. It would require changes only at the stack recipe level and can thus be handled independently of our release cycles.

> It might be worth a thought removing all the non-cli related code into a separate file, but I do understand that this is a similar case to the examples which we have handled similarly until now.

I think that since the Terraform code is only used inside the `stack_recipes` CLI, it would be okay to place it in the same file as the commands. However, I'm open to putting it separately as I agree it would look cleaner but can't decide on a suitable home for it. Do you have any potential candidates in mind?

src/zenml/cli/stack_recipes.py
wjayesh: I agree, thanks for the catch!

src/zenml/cli/stack_recipes.py
wjayesh: You're right, I'll remove the assignment.

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1013 @@
...
+                "The file values.tfvars.json was not found in the "
+                f"recipe's directory at {variables_file_path}. Please "
+                "verify if it exists."
+            )
+
+        # read values into a dict and return
+        with fileio.open(variables_file_path, "r") as f:
+            variables = json.load(f)
+            f.close()
+            return variables
```
schustmi, line 155:
```suggestion
        with fileio.open(variables_file_path, "r") as f:
            variables = json.load(f)
        return variables
```
The `with` statement already takes care of closing the file for us here.

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1013 @@
...
+
+
+@click.option(
+    "--path",
+    "-p",
+    type=click.STRING,
+    default="zenml_stack_recipes",
+    help="Relative path at which you want to clean the stack_recipe(s)",
+)
+@stack_recipe.command(help="Deletes the ZenML stack recipes directory.")
```
schustmi, line 560: Apparently this works as is, but I think it would be nicer for readability to have the `.command()` above all the possible options

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1013 @@
...
+@click.option(
+    "--path",
+    "-p",
+    type=click.STRING,
+    default="zenml_stack_recipes",
+    help="Relative path at which you want to install the stack recipe(s)",
+)
+def pull(
+    git_stack_recipes_handler: GitStackRecipesHandler,
+    stack_recipe_name: str,
```
schustmi, line 636:
```suggestion
    stack_recipe_name: Optional[str],
```
This might be none

pyproject.toml
```diff
@@ -70,7 +70,10 @@ nbconvert = "6.4.4"
 
 # Optional dependencies for the ZenServer
 fastapi = { version = "~0.75.0", optional = true }
-uvicorn = { extras = ["standard"], version = "~0.17.5" , optional = true}
+uvicorn = { extras = ["standard"], version = "~0.17.5",  optional = true}
+
+# dependencies for stack recipes
+python-terraform = "^0.10.1"
```
schustmi, line 76: This line makes the terraform python package an explicit dependency of ZenML, is this intended? I haven't checked how large this package is, but we could maybe think about having it optional as well?

schustmi (changes requested):

pyproject.toml
htahir1: I would suggest we move it as an extra so `pip install zenml[terraform]` or something?

htahir1: If we do the above, we should add some logic that tries to import this in the __init__ of stack recipes and fails gracefully if not installed

src/zenml/cli/stack.py
```diff
@@ -1143,16 +1143,28 @@ def _import_stack_component(
...
+    help="Import stack components even if the installed version of ZenML "
+    "is different from the one specified in the stack YAML file",
+)
 @click.pass_context
 def import_stack(
-    ctx: click.Context, stack_name: str, filename: Optional[str]
+    ctx: click.Context,
+    stack_name: str,
+    filename: Optional[str],
+    ignore_version_mismatch: bool,
```
htahir1, line 1157: This should have a default behavior, no? Like False

htahir1, line 1157: One more thing id like here is this --decouple-stores flag that is bothering me a lot and is available in the zenml stack register command natively

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1013 @@
...
+from zenml.utils import io_utils
+from zenml.utils.analytics_utils import AnalyticsEvent, track_event
+
+logger = get_logger(__name__)
+
+EXCLUDED_RECIPE_DIRS = [""]
+STACK_RECIPES_GITHUB_REPO = "https://github.com/zenml-io/mlops-stacks.git"
+STACK_RECIPES_REPO_DIR = "zenml_stack_recipes"
+VARIABLES_FILE = "values.tfvars.json"
+ALPHA_MESSAGE = (
```
htahir1, line 45: I would even add here that they should go to the `STACK_RECIPES_GITHUB_REPO` to report issues

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1013 @@
...
+            io_utils.create_dir_if_not_exists(destination_dir)
+            git_stack_recipes_handler.copy_stack_recipe(
+                stack_recipe, destination_dir
+            )
+            cli_utils.declare(
+                f"Stack recipe pulled in directory: {destination_dir}"
+            )
+            track_event(
+                AnalyticsEvent.PULL_STACK_RECIPE,
+                {"stack_recipe_name": stack_recipe.name},
```
htahir1, line 694: Would it not be nice to also list the stack recipe version here so we can backcheck the version of the recipe?

htahir1: This applies to all track_events in this  PR

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1013 @@
...
+            )
+            text = Text("zenml stack describe", style="markdown.code_block")
+            cli_utils.declare(text)
+            cli_utils.declare(
+                "\n" + "to investigate and switch to a new stack if needed."
+            )
+
+        except python_terraform.TerraformCommandError as e:
+            force_message = ""
+            if stack_recipe_name == "aws_minimal":
```
htahir1, line 993: This is a bit too hard-coded IMO. We shouldnt have any recipe specific logic in here. We can for now keep it in but remove it later

htahir1 (changes requested):
Ok this is a good but as its so close to examples we might want to refactor at some point. Otherwise, left comments! Cheers!!

src/zenml/cli/stack.py
schustmi: FYI I think this flag will be gone with the new ZenServer/MLMD changes, probably still makes sense to include/forward it here for now right?

pyproject.toml
schustmi: Yes I agree, `zenml stack recipe` could simply print a warning/info message that it's missing this package.

```python
try:
  import python_terraform
except ImportError:
  # add just the stack recipe group here with info on how to install it/make it available
else:
  # the normal stack recipe CLI code here
```

src/zenml/cli/stack_recipes.py
wjayesh: That's a good point but unfortunately, I haven't implemented versions on the recipes so far. I'll get to it.

src/zenml/cli/stack_recipes.py
wjayesh: Agreed

src/zenml/cli/stack_recipes.py
wjayesh: I completely agree but the eks terraform module currently doesn't get cleaned up properly. I first thought of adding a `--force`  option with logic that can destroy the eks module resources and then clean up the rest of the recipe. But owing to the very risky nature of the commands I decided against adding it to the CLI in any form and instead, went ahead with this very specific warning. 

I'll take another shot at resolving the `eks` problem and if resolved, this error message can be removed.

src/zenml/cli/stack.py
wjayesh: Okay, I have added the `decouple_stores` option.

src/zenml/cli/stack_recipes.py
wjayesh: agreed!

src/zenml/cli/stack_recipes.py
wjayesh: good catch!

src/zenml/cli/stack_recipes.py
wjayesh: okay, thanks for pointing out!

pyproject.toml
wjayesh: That makes sense! Should I go with `zenml[recipes]`  or `zenml[terraform]` for the extras installation?

htahir1: `zenml[recipes]` might make sense

src/zenml/cli/stack.py
```diff
@@ -1171,14 +1193,22 @@ def import_stack(
...
+                f"{data['zenml_version']}. You have version "
+                f"{zenml.__version__} installed. Some components might not "
+                "work as expected."
+            )
+        else:
+            cli_utils.error(
+                f"Cannot import stacks from other ZenML versions. "
+                f"The stack was created using ZenML version "
+                f"{data['zenml_version']}, you have version "
+                f"{zenml.__version__} installed."
```
AlexejPenner, line 1210:
```suggestion
                f"Cannot import stacks from other ZenML versions. "
                f"The stack was created using ZenML version "
                f"{data['zenml_version']}, you have version "
                f"{zenml.__version__} installed. You can "
                retry using the `--ignore-version-mismatch` 
                flag. However, be aware that this might
                fail or lead to other unexpected behavior."
```

wjayesh: Great addition!

src/zenml/cli/stack_recipes.py
```diff
@@ -0,0 +1,1044 @@
...
+                logger.info(
+                    "\nA stack configuration YAML file has been generated as "
+                    f"part of the deployment of the {stack_recipe_name} recipe. "
+                    f"Find it at {stack_yaml_file}."
+                )
+
+                if not no_import:
+                    logger.info(
+                        "\nThe flag --no-import is not set. Proceeding "
+                        "to import a new ZenML stack from the created resources."
```
AlexejPenner, line 862:
```suggestion
                        "\nThe flag `--no-import` is not set. Proceeding "
                        "to import a new ZenML stack from the created resources."
```

AlexejPenner (approved):
Beautifully done, love all the usability goodies and clear error/warning messages :fire:

PR #811, opened 2022-07-29
New Docker build configuration

schustmi:
## Describe changes

This PR greatly increases the flexibility in which users can specify how ZenML builds the Docker image that is used to run their pipeline steps in remote orchestrators/step operators. Additionally, it reduces the code duplication of the Docker image building code that was found in many stack components by introducing a simple pydantic mixin `PipelineDockerImageBuilder` which all stack components can inherit from to easily build Docker images.

## Pre-requisites

## Types of changes


docs/book/developer-guide/advanced-usage/docker.md
```diff
@@ -31,60 +41,155 @@ following two ways:
...
+
+{% hint style="warning" %}
+This is an advanced feature and will most likely break your pipelines. If you use this,
+you're on your own and need to copy a stack configuration to the correct path yourself.
+{% endhint %}
+
+### How to install additional pip dependencies
+
+By default, ZenML will automatically install all the packages required by your
+active ZenML stack. There are however many additional ways in which you can specify
```
bcdurak, line 88:
```suggestion
active ZenML stack. There are, however, various ways in which you can specify
```

src/zenml/pipelines/base_pipeline.py
```diff
@@ -136,23 +134,50 @@ def __init__(self, *args: BaseStep, **kwargs: Any) -> None:
...
+            kwargs.pop(PARAM_DOCKER_CONFIGURATION, None)
+            or DockerConfiguration()
+        )
+        self._migrate_to_docker_config(kwargs)
+
         self.name = self.__class__.__name__
         self.__steps: Dict[str, BaseStep] = {}
         self._verify_arguments(*args, **kwargs)
 
+    def _migrate_to_docker_config(self, kwargs: Dict[str, Any]) -> None:
```
bcdurak, line 149: The docstring for the method is missing here.

src/zenml/steps/step_environment.py
```diff
@@ -102,10 +105,10 @@ def cache_enabled(self) -> bool:
         return self._cache_enabled
 
     @property
-    def pipeline_requirements(self) -> List[str]:
-        """The name of the currently running step.
+    def docker_configuration(self) -> "DockerConfiguration":
```
bcdurak, line 108: Is this used anywhere? I am not sure if I am missing anything.

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -505,10 +505,12 @@ def _configure_node_context(
             runtime_config_copy, sort_keys=True, default=pydantic_encoder
         )
 
+        docker_config_json = pipeline.docker_configuration.json(sort_keys=True)
```
bcdurak, line 508: Just out of curiosity, I see two different paradigms here. While the `runtime_configuration` is given to this function by the orchestrator, the `docker_configuration` is treated as a class variable of the pipeline. Is there a specific reason why it is like this?

bcdurak (commented):
I have been reading this for a while and everything looks amazing. I have left a few small comments.

I am so looking forward to merging this into my branch. I will try it right away and share my findings from a more hands-on review process :)

src/zenml/pipelines/base_pipeline.py
schustmi: Good catch, I added it!

src/zenml/steps/step_environment.py
schustmi: Not in my PR, but @safoinme added the `pipeline_requirements` property to the step environment so he could then use it in this custom code deployment PR. Now the pipeline requirements don't exist anymore but got replaced by the docker config.

src/zenml/orchestrators/base_orchestrator.py
schustmi: Mhm correct, I think this is mainly due to the `docker_configuration` being specified on the `@pipeline` decorator right now, while the runtime config is passed to `.run(...)`. However I think the entire configurations are quite messy at the moment, hopefully the design document we're creating this week will bring some more clarity in how we handle these things in the future :)

safoinme (commented):
Half through The PR, such an amazing implementation to something that really bothered me while working on the custom deployment!
@schustmi one question regarding the additional build argument, what would be the best way to add an additional argument (e.g. `expose 5000`)

schustmi:
@safoinme I'm assuming that by build argument in this case you mean adding an additional line to the generated Dockfile. 

Here would be one potential solution how to do that:

```python
MyModelDeployer(BaseModelDeployer, PipelineDockerImageBuilder):
    @staticmethod
    def _generate_zenml_pipeline_dockerfile(
        parent_image: str,
        docker_configuration: DockerConfiguration,
        requirements_files: Sequence[str] = (),
        entrypoint: Optional[str] = None,
    ) -> List[str]:
        lines  = super()._generate_zenml_pipeline_dockerfile(parent_image, docker_configuration, requirements_files, entrypoint)
        lines.append("EXPOSE 5000")
        return lines
```

src/zenml/orchestrators/base_orchestrator.py
bcdurak: Looking forward to it :)

safoinme (approved):
LGTM!! Looking forward to see this merged

PR #812, opened 2022-07-29
Implement Recursive Built-In Container Materializer

fa9r:
## Describe changes

I revamped `materializers.built_in_materializer.py` to add support for `bytes`, `set`, and non-JSON-serializable `dict`, `list`, and `tuple` objects. 

The main implication of this is that you can now use arbitrary lists/sets/tuples/sets of materializable data types in your steps without having to write a custom materializer for it.

E.g., the following data types can now be handled automatically:
- `Dict[str, np.ndarray]`,
- `Set[pd.DataFrame]`,
- `Dict[str, List[torch.nn.module]]`, if pytorch integration is installed,
- `Dict[str, List[MyCustomClass]]`, if a materializer for `MyCustomClass` was defined,
- `List[Union[Dict[str, Union[Tuple[int, float, str, bool], Dict[str, List[np.ndarray]], bytes]], Set[Union[float, int]]]]`,
- ...

### Implementation Details

The original `BuiltInMaterializer` was split into three separate classes:
- `BuiltInMaterializer` now only handles `bool`, `float`, `int`, `str`; otherwise its behavior is unchanged (materialization via serialization to JSON file),
- `bytes` type is now handled by a separate `BytesMaterializer` since `bytes` is not JSON-serializable (this was broken before, seems like no one ever tried to use `bytes` artifacts in ZenML).
- `list`, `dict`, `tuple` and (new) `set` are now handled by a separate `BuiltInContainerMaterializer`.

`BuiltInContainerMaterializer` works like this:
- If the given container (dict/list/set/tuple) is JSON-serializable, write it to JSON (similar as before).
- Otherwise, recursively materialize all elements in the container into a subdirectory by finding the corresponding materializer from the `default_materializer_registry` at runtime.
- Tuples and sets are cast to `list` before materialization (and back to the original type after loading).
- Non-serializable dicts are materialized as a list of lists `[keys, values]` (and reconstructed as `dict` after loading).

I also added 9 unit tests covering the materialization of all supported built-in data types.

### Potential Issues

- Materializers are strongly linked to artifacts. Thus, when creating a materializer for an element at runtime, we need to create a mock artifact in order to initialize a corresponding materializer. This might have uninteneded side-effects since it goes deep into TFX-land. Maybe @htahir1 or @bcdurak might know whether this is problematic or not?
- When loading elements at run time, the data type is found by iterating through all artifact types registered in the `default_materializer_registry` and checking if any of them have a simiilar string representation as the type of the element had before materializing. This means, loading data will fail if any element had a type that is not explicitly registered in the registry and was instead materialized by a materializer linked to a superclass. However, this should very rarely happen in practice and is IMO a design flaw of the `default_materializer_registry` since implicitly using the materializer of a superclass is already quite unpredictable in itself.

## Pre-requisites

## Types of changes


src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,11 +71,246 @@ def handle_input(self, data_type: Type[Any]) -> Any:
...
+
+    Args:
+        obj: The object to check.
+
+    Returns:
+        True if the entire object is JSON-serializable, else False.
+    """
+    if any(isinstance(obj, basic_type) for basic_type in BASIC_TYPES):
+        return True
+    if any(isinstance(obj, iterable_) for iterable_ in (list, tuple)):
```
schustmi, line 145: Maybe include `set` here as well? Also, I think `isinstance(obj, ...)` allows a tuple as the second value to simplify the code a little

src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,11 +71,246 @@ def handle_input(self, data_type: Type[Any]) -> Any:
...
+        # non-serializable dict: Handle as non-serializable list of lists.
+        if isinstance(data, dict):
+            data = [list(data.keys()), list(data.values())]
+
+        # non-serializable list: Materialize each element into a subfolder.
+        # Get path, type, and corresponding materializer for each element.
+        paths, types, materializers = [], [], []
+        for i, element in enumerate(data):
+            element_path = os.path.join(self.artifact.uri, str(i))
+            os.mkdir(element_path)
```
schustmi, line 292: Needs to use fileio to work for remote artifact stores

src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,11 +71,246 @@ def handle_input(self, data_type: Type[Any]) -> Any:
...
+            # Write metadata as JSON.
+            metadata = {"length": len(data), "paths": paths, "types": types}
+            yaml_utils.write_json(self.metadata_path, metadata)
+            # Materialize each element.
+            for element, materializer in zip(data, materializers):
+                materializer.handle_return(element)
+        # If an error occurs, delete all created files.
+        except Exception as e:
+            # Delete metadata
+            if os.path.exists(self.metadata_path):
```
schustmi, line 311: Fileio here, in the `remove` line and the `rmtree`

src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,11 +71,246 @@ def handle_input(self, data_type: Type[Any]) -> Any:
...
+
+        # non-serializable dict: Handle as non-serializable list of lists.
+        if isinstance(data, dict):
+            data = [list(data.keys()), list(data.values())]
+
+        # non-serializable list: Materialize each element into a subfolder.
+        # Get path, type, and corresponding materializer for each element.
+        paths, types, materializers = [], [], []
+        for i, element in enumerate(data):
+            element_path = os.path.join(self.artifact.uri, str(i))
```
schustmi, line 291: This might lead to strange results when using a remote artifact store on a windows machine. The artifact uri will be something like `s3://my_bucket/step_1` and will be joined using backslashes: `s3://my_bucket/step_1\\3`.

schustmi (changes requested):

src/zenml/materializers/built_in_materializer.py
fa9r: ugh why do we support Windows again? 😆

schustmi: Actually I remembered that I implemented something a while ago to catch this exact case, so I think this should actually be fine here.

strickvl (commented):
Just checked this out + tried it using the Pillow materializer and a list etc. All works well, AFAICS.

I think you need to add something about this in the docs, since it's a pretty great update / change, and I think we can be explicit about it.

I did notice that it did NOT work when I specified the materializer using the `with_return_materializers(...)` method. Perhaps that can be part of the docs update. The problem is that the step is now returning a `List` or whatever and the `with_return...` means it's expecting `Image.Image`. In any case, maybe just worth being explicit about this in the docs?

src/zenml/materializers/built_in_materializer.py
bcdurak: also for the `exists`?

src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,11 +71,246 @@ def handle_input(self, data_type: Type[Any]) -> Any:
...
+        Returns:
+            The data read.
+
+        Raises:
+            RuntimeError: If the data was not found.
+        """
+        super().handle_input(data_type)
+
+        # If the data was not serialized, there must be metadata present.
+        if not os.path.exists(self.data_path) and not os.path.exists(
```
bcdurak, line 219: fileio here

src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,11 +71,246 @@ def handle_input(self, data_type: Type[Any]) -> Any:
...
+            self.metadata_path
+        ):
+            raise RuntimeError(
+                f"Materialization of type {data_type} failed. Expected either"
+                f"{self.data_path} or {self.metadata_path} to exist."
+            )
+
+        # If the data was serialized as JSON, deserialize it.
+        if os.path.exists(self.data_path):
```
bcdurak, line 228: fileio here

bcdurak (commented):
I left a few small comments. Other than those, everything seems great!

schustmi:
> Just checked this out + tried it using the Pillow materializer and a list etc. All works well, AFAICS.
> 
> I think you need to add something about this in the docs, since it's a pretty great update / change, and I think we can be explicit about it.
> 
> I did notice that it did NOT work when I specified the materializer using the `with_return_materializers(...)` method. Perhaps that can be part of the docs update. The problem is that the step is now returning a `List` or whatever and the `with_return...` means it's expecting `Image.Image`. In any case, maybe just worth being explicit about this in the docs?

@strickvl The materializer you passed to `with_return_materializers(...)` was a custom one that is able to handle a list? Or one handling a single pillow image that should be used for elements in the returned list?

From my understanding the first option should work as is, while the second one doesn't (and also won't be able without some bigger changes I think)

strickvl:
@schustmi it was "one handling a single pillow image that should be used for elements in the returned list". I think my expectation was that this new code could use that to handle the non-serializable elements of the list, but actually it makes a hard link between the return type which is why it raised an error I imagine.

schustmi:
@strickvl Yes exactly. I think what you tried would actually be a really nice feature, but would require some extra effort: We would need a separate dict which maps data types to materializers (this is similar to the default materializer registry, but on a per-step basis). Additionally, @fa9r logic to serialize/deserialize the materializers to a string and read from the default registry would break down.
Not sure if we should do this in this PR (if it doesn't make it into the release), but should definitely create a ticket to keep this in mind.

strickvl:
@schustmi yeah let's leave it for this ticket. It's already pretty nice, esp when it has some extra docs to show it off. I'll make the ticket for the remaining part @fa9r.

src/zenml/materializers/built_in_materializer.py
fa9r: Great suggestion, applied both!

src/zenml/materializers/built_in_materializer.py
fa9r: Okay perfect! 😄

PR #816, opened 2022-08-02
Add docs for local kubeflow + local metadata store

schustmi:
## Describe changes

Based on a Slack-Thread, this PR adds information on how to access the local metadata store with local kubeflow.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/mlops-stacks/orchestrators/kubeflow.md
+++ b/docs/book/mlops-stacks/orchestrators/kubeflow.md
@@ -131,6 +131,17 @@ cluster.
 * A [local container registry](../container-registries/default.md) as part of your stack.
 
 {% hint style="warning" %}
+If you're using the Kubeflow orchestrator locally in combination with a
+[local metadata store](../metadata-stores/sqlite.md), you need to make sure the SQLite
+database is writable from inside the Kubeflow Pods. To do so, first figure our where
+your global config directory is using [this guide](../../resources/global-config.md#the-global-config)
+and then run:
+```shell
+chmod -R a+rw <GLOBAL_CONFIG_DIR>/local_stores
+```
+{% endhint %}
+
+{% hint style="warning" %}
 The local Kubeflow Pipelines deployment requires more than 2 GB of RAM,
 so if you're using Docker Desktop make sure to update the resource
 limits in the preferences.


docs/book/mlops-stacks/orchestrators/kubeflow.md
```diff
@@ -130,6 +130,17 @@ When using the Kubeflow orchestrator locally, you'll additionally need
...
 * A [local container registry](../container-registries/default.md) as part of your stack.
 
+{% hint style="warning" %}
+If you're using the Kubeflow orchestrator locally in combination with a
+[local metadata store](../metadata-stores/sqlite.md), you need to make sure the SQLite
+database is writable from inside the Kubeflow Pods. To do so, first figure our where
+your global config directory is using [this guide](../../resources/global-config.md#the-global-config)
+and then run:
+```shell
+chmod -R a+rw <GLOBAL_CONFIG_DIR>/local_stores
```
stefannica, line 140: Doing this on the local filesystem basically gives everyone on the machine read/write access to all these files in your home directory (on Linux at least), which includes your local secrets manager stores.

stefannica (changes requested):
Didn't we already cover this use-case by using the local UID/GID in the container image built locally for Kubeflow ?

docs/book/mlops-stacks/orchestrators/kubeflow.md
```diff
@@ -130,6 +130,17 @@ When using the Kubeflow orchestrator locally, you'll additionally need
 cluster.
 * A [local container registry](../container-registries/default.md) as part of your stack.
 
+{% hint style="warning" %}
+If you're using the Kubeflow orchestrator locally in combination with a
+[local metadata store](../metadata-stores/sqlite.md), you need to make sure the SQLite
+database is writable from inside the Kubeflow Pods. To do so, first figure our where
```
fa9r, line 136:
```suggestion
database is writable from inside the Kubeflow Pods. To do so, first figure out where
```

fa9r (changes requested):
As @stefannica pointed out, having to change the global config access is somewhat unsafe und definitely feels hacky. Since we're reworking the metadata store anyways, I would propose to hold off on these docs changes, and, if needed, document how to do use local Kubeflow under the new structure later.

docs/book/mlops-stacks/orchestrators/kubeflow.md
schustmi: The user in slack still ran into issues, even with the UID/GID commands in the image. It might not be related to Kubeflow after all, as there seems to be some SQLite issues when the user doesn't have write access to the entire directory containing the database.
I guess this call could also be limited to exactly this directory by doing `chmod -R a+rw <GLOBAL_CONFIG_DIR>/local_stores/<UUID>`, which only contains the sqlite database. However I also don't have a problem not adding that at all if you think it's not a common error

PR #820, opened 2022-08-04
Pillow Image materializer

strickvl:
I ported a version of our Pillow image materializer from the annotation example such that it works with #812.

## Pre-requisites

## Types of changes


htahir1 (changes requested):
Hey! We should add this materializer to a Pillow integration rather than base ZenML, right? Its not part of any core dependency of ZenML?

schustmi (commented):
More of a general question, but why isn't this part of an integration but instead part of core ZenML? I don't think we install Pillow by default, which means this would crash for all users that don't have it installed right?

src/zenml/materializers/pillow_image_materializer.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def handle_return(self, image: PIL.Image.Image) -> None:
+        """Write to artifact store.
+
+        Args:
+            image: A PIL.Image.Image object.
+        """
+        super().handle_return(image)
+        temp_dir = tempfile.TemporaryDirectory()
+        file_extension = image.format
```
stefannica, line 82:
```suggestion
        file_extension = image.format.lower()
```

Just a tiny nit. IIRC, the format was uppercase.

src/zenml/materializers/pillow_image_materializer.py
```diff
@@ -0,0 +1,93 @@
...
+        full_filename = f"{DEFAULT_IMAGE_FILENAME}.{file_extension}"
+        temp_image_path = os.path.join(temp_dir.name, full_filename)
+        artifact_store_path = os.path.join(self.artifact.uri, full_filename)
+
+        # save the image in a temporary directory
+        image.save(temp_image_path)
+
+        # copy the saved image to the artifact store
+        io_utils.copy(temp_image_path, artifact_store_path, overwrite=True)  # type: ignore[attr-defined]
+        fileio.remove(temp_image_path)
```
stefannica, line 93:
```suggestion
        artifact_store_path = os.path.join(self.artifact.uri, full_filename)

        # save the image in a temporary file
        with tempfile.NamedTemporaryFile() as f:
            image.save(f.name)
            # copy the saved image to the artifact store
            io_utils.copy(f.name, artifact_store_path, overwrite=True)  # type: ignore[attr-defined]
```

You can use the `NamedTemporaryFile` utility to get there faster. The file is deleted automatically when the context exits.

src/zenml/materializers/pillow_image_materializer.py
```diff
@@ -0,0 +1,93 @@
...
+            str(temp_dir),
+            f"{DEFAULT_IMAGE_FILENAME}{os.path.splitext(filepath)[1]}",
+        )
+
+        # copy from artifact store to temporary file
+        fileio.copy(filepath, temp_file)
+        image = PIL.Image.open(temp_file)
+
+        # Cleanup and return
+        fileio.rmtree(temp_dir)
```
stefannica, line 71:
```suggestion
        file_extension = os.path.splitext(filepath)[-1]
        with tempfile.NamedTemporaryFile(suffix=f".{file_extension}") as f:
                # copy from artifact store to temporary file
                fileio.copy(filepath, f.name)
                image = PIL.Image.open(f.name)
```

You can use the `NamedTemporaryFile` utility to get there faster. The file is deleted automatically when the context exits.

stefannica (changes requested):
A couple of code suggestions to make it easier for you to handle temporary files. If you decide to keep this a builtin materializer, you should also add `types-Pillow` to the zenml pip dependencies. 

@htahir1 @schustmi Might be a good idea to keep this as a builtin materializer, to add more data types for computer vision datasets to the ZenML core.

tests/unit/materializers/test_pillow_image_materializer.py
```diff
@@ -0,0 +1,57 @@
...
+
+import PIL
+
+from zenml.materializers.pillow_image_materializer import (
+    PillowImageMaterializer,
+)
+from zenml.pipelines import pipeline
+from zenml.steps import step
+
+JPEG_FILE_PATH = os.path.join(
```
fa9r, line 25: feels a bit weird to load an image from examples/ here, I would suggest to create a random image at the start of the test instead and try to load that. E.g. you could use something like this to create a random image: https://stackoverflow.com/questions/10901049/create-set-of-random-jpgs

fa9r (changes requested):
As other reviewers pointed out, this should probably be an integration, not built-in. I also added a few suggestions to improve the test. Otherwise looks good to me.

src/zenml/materializers/pillow_image_materializer.py
strickvl: I originally had it like that, but then mypy complains a lot that the return type of `image.format` is `Optional[str]` so then I'd need to add various bits of extra guard logic, and for the purposes of materialization it doesn't make a difference if it's lower or upper case. WDYT?

tests/unit/materializers/test_pillow_image_materializer.py
strickvl: Thanks, yes I had the same idea overnight. No need to load an image at all, though this did reveal a potential bug that I've also now fixed. (TL;DR: if you create an image de novo in Pillow, it doesn't come with an inbuilt image format (vs if you load one from a file) so actually that didn't work with our materializer.. It now does).

strickvl:
Thanks for the reviews and comments. I've now updated / amended things as per the feedback.

I think my feeling around having it as a built-in materializer was that Pillow feels like at least as fundamental as Pandas or Numpy for computer vision work, but since consensus wanted it in integrations instead, that's where I've put it for now.

src/zenml/integrations/pillow/materializers/pillow_image_materializer.py
```diff
@@ -0,0 +1,89 @@
...
+        """Write to artifact store.
+
+        Args:
+            image: A PIL.Image.Image object.
+        """
+        super().handle_return(image)
+        file_extension = image.format
+        # handle images generated in Pillow itself
+        if file_extension is None:
+            file_extension = "PNG"
```
schustmi, line 77:
```suggestion
        file_extension = image.format or "PNG
```

.github/workflows/unit-test.yml
```diff
@@ -41,7 +41,8 @@ jobs:
       - name: Install Prerequisites
         run: |
           source $VENV
-          zenml integration install kubeflow s3 gcp azure gcp_secrets_manager vertex vault -f
+          zenml integration install kubeflow s3 gcp azure gcp_secrets_manager
+          vertex vault pillow -f
```
fa9r, line 45: does this still work when it's split over two lines like this?

fa9r (approved):
LGTM!

.github/workflows/unit-test.yml
strickvl: Hmmm good point. I've updated. I'm pretty sure it was like that before I got there, but in any case it's all on a single line now.

htahir1 (approved):

stefannica (approved):
I agree with you Alex, I still think this should be a builtin materializer. Having to install an entire integration just to have access to a data type feels unnecessarily complicated and I haven't seen any arguments against it other than "this is how we've always done it". But I'm fine with it either way.

schustmi:
I can give you a pretty simple reason, ZenML does not have a dependency on pillow last time I checked. Why should we force every user to install Pillow just so the ones that want to use it can save a single line of installation?

stefannica:
> I can give you a pretty simple reason, ZenML does not have a dependency on pillow last time I checked. Why should we force every user to install Pillow just so the ones that want to use it can save a single line of installation?

The same reason we force every user to install numpy: the only thing requiring numpy in ZenML is the materializer, yet we have it as a ZenML dependency.

fa9r:
@stefannica good point, I think having individual materializers as separate integrations feels really weird in general. However, having them as dead-weight dependencies is even worse IMO.

Would it be possible that we somehow only load materializer code at runtime so we can make them automatically available without having them as a dependency of ZenML itself?

schustmi:
> > I can give you a pretty simple reason, ZenML does not have a dependency on pillow last time I checked. Why should we force every user to install Pillow just so the ones that want to use it can save a single line of installation?
> 
> @schustmi the same reason we force every user to install numpy: the only thing requiring numpy in ZenML is the materializer, yet we have it as a ZenML dependency. Screenshot from 2022-08-11 10-41-01

I'm quite certain that numpy is required by some dependency of ZenML and that's why it was added as a built-in. The reason was definitely not "we force every user to install this", and I don't think we should ever do that with something we don't actively require.

Don't get me wrong, I dislike having them in integrations as well, but simply requiring it for the sake of having the materializer built-in seems wrong, especially with quite heavy-weight dependencies like Pillow.

Possible solutions:
- **Option 1**: Loosen the restrictions on materializer integration requirements: If an integration would for example just require "Pillow" for it to be activated, there are 0 user steps involved in getting this to run. If they are using Pillow in their code, they will need to run `pip install Pillow` at some point, at which point ZenML will detect that and "activate" the materializer. This however still leaves us with a long list of essentially empty integrations which contain a single materializer.
- **Option 2**: Keep all the materializer integrations in a separate folder in core-zenml. Then once we get to a point where materializers might be needed (I can only think of running a pipeline at the moment), we can try to import all those files. The import will fail if the package doesn't exist, but if it succeeds the materializer is registered.

The downside to both of those: A materializer might be broken in certain package versions, and users will have to deal with that themselves.

AlexejPenner (approved):

PR #826, opened 2022-08-08
Decouple metadata store from stack

fa9r:
## Describe changes
I decoupled the metadata store from the stack and added it as an attribute of the `SQLZenStore` instead. I also updated/removed all CLI commands related to metadata stores and adjusted all unit tests to the new structure.

### Not Implemented (future TODO):
- implementations for new methods of RestZenStore and corresponding entrypoints of ZenServer

### Technical Details

In particular, I refactored the following classes to not use the metadata store:
- `BaseOrchestrator` calls `repo.zen_store._get_tfx_metadata_config()` now.
- All post execution classes (`PipelineView`, `PipelineRunView`, `StepView`, `ArtifactView`) no longer have the metadata store as an internal attribute, and query `repo.zen_store` instead.
- `StepContext` no longer has method `get_pipeline()`. This can only be queried via `repo.zen_store.get_pipeline()` now.

This introduces the following changes to the ZenStore API:
- `get_pipeline_run()` -> `get_pipeline_run_wrapper()`
- `get_pipeline_runs()` -> `get_pipeline_run_wrappers()`
- + `get_pipeline()`
- + `get_pipelines()`
- + `get_pipeline_run()`
- + `get_pipeline_runs()`
- + `get_pipeline_run_steps()`
- + `get_step_by_id()`
- + `get_step_status()`
- + `get_step_artifacts()`
- + `get_producer_step_from_artifact()`
- + `_get_tfx_metadata_config()`

### Open Questions
- Will the `StepContext` changes work correctly when running steps in remote environments?
- If yes, should `StepContext` as a whole become part of `Repository`?
- Should the post execution classes be further refactored to retrieve the metadata via a separate endpoint `get_pipeline_run_metadata()` instead of the current `get_pipeline_run_wrapper()`?

## Pre-requisites

## Types of changes


tests/unit/zen_server/test_zen_server.py
```diff
@@ -32,7 +32,9 @@ def running_zen_server(tmp_path_factory: pytest.TempPathFactory) -> ZenServer:
     tmp_path = tmp_path_factory.mktemp("local_zen_store")
     global_cfg = GlobalConfiguration()
 
-    backing_zen_store = LocalZenStore().initialize(str(tmp_path))
+    backing_zen_store = SqlZenStore().initialize(
+        f"sqlite:///{tmp_path / 'store.db'}"
```
schustmi, line 36: Is this something that needs backslashes on windows instead or fine with the hardcoded `/`?

src/zenml/metadata_stores/base_metadata_store.py
```diff
@@ -48,12 +49,10 @@
...
 
 
-class BaseMetadataStore(StackComponent, ABC):
+class BaseMetadataStore(BaseModel, ABC):
     """Base class for all ZenML metadata stores."""
 
-    # Class Configuration
-    TYPE: ClassVar[StackComponentType] = StackComponentType.METADATA_STORE
-
+    uuid: UUID = Field(default_factory=uuid4)
```
schustmi, line 55: Any reason for this UUID still existing? For some potential scoping inside the metadata store?

src/zenml/zen_stores/local_zen_store.py
```diff
@@ -921,7 +921,7 @@ def get_role_assignments_for_team(
 
     # Pipelines and pipeline runs
 
-    def get_pipeline_run(
+    def get_pipeline_run_wrapper(
```
schustmi, line 924: Shouldn't this entire yaml store be deleted?

src/zenml/post_execution/pipeline.py
```diff
@@ -72,10 +63,10 @@ def runs(self) -> List["PipelineRunView"]:
         """
         # Do not cache runs as new runs might appear during this objects
         # lifecycle
-        runs = list(self._metadata_store.get_pipeline_runs(self).values())
+        runs = list(Repository().zen_store.get_pipeline_runs(self).values())
```
schustmi, line 66: A quick explanation on why this contained the metadata store as a private attribute before:
```python
pipeline = Repository().get_pipeline(...)
# switch to a different profile
GlobalConfig().activate_profile("other_profile")
# the following call now fails or results in unexpected behaviour as it gets
# runs from a different profile and the associated zen store
# that's the issue with having this huge repo singleton that we query everywhere,
# changing the active state in a python process messes with a lot of things
pipeline.get_runs()
```

I'm not sure if this is a concern for now, but it could certainly bite us at some point. (Same for all the other post-execution classes)

schustmi (changes requested):

src/zenml/metadata_stores/base_metadata_store.py
fa9r: I guess we can take it out. Initially I kept this since the post execution classes used it, but I refactored that out already. The only consideration, as you said, might be to keep it in case we need to have multiple metadata stores within the zenstore (which I think we want?). TL;DR not sure whether this can go or not :D

src/zenml/zen_stores/local_zen_store.py
fa9r: Yeah, I have another branch for that, but didn't go ahead yet since we haven't made a final decision of whether to migrate it or not. If we do migrate, then keeping (and potentially adjusting) the class will be required.

src/zenml/post_execution/pipeline.py
fa9r: Yes, this will be an issue. However, I don't think switching profiles in the middle of your workflows should be supported anyways. I'd propose to check in each post execution class call that the active profile is the same as the profile that the class was queried from and raise an error otherwise. Any opinions?

tests/unit/zen_server/test_zen_server.py
fa9r: I guess it should be fine since all the Windows unit tests passed in the CI?

AlexejPenner (approved):
As there is many more changes to come in the coming week I would advocate for merging this in to the 1103 branch soon and continue with incremental improvements and changes throughout the next few weeks.

src/zenml/metadata_stores/base_metadata_store.py
schustmi: Alright we can just keep it for now till we're sure how we handle scoped metadata stores I guess.

tests/unit/zen_server/test_zen_server.py
schustmi: I think we actually skip all the ZenServer tests on windows as it doesn't support the daemon process?

src/zenml/post_execution/pipeline.py
schustmi: Seems like almost as much (if not more) work than actually supporting this feature?
However I agree that probably switching profiles should not be something that is happening while running code, however we won't be able to prevent users from doing that if they want. Anyway I'll leave this up to you, nothing critical imo.

tests/unit/zen_server/test_zen_server.py
fa9r: So then it doesn't matter anyways? 😄

fa9r:
> As there is many more changes to come in the coming week I would advocate for merging this in to the 1103 branch soon and continue with incremental improvements and changes throughout the next few weeks.

Yeah makes sense, let's do it.

PR #828, opened 2022-08-08
Add automated link check github actions

dnth:
## Describe changes
This PR adds automated link checks on all markdown files in the repo. The workflow will be triggered upon a push to the repo.

## Pre-requisites

## Types of changes


dnth:
Here is an sample output of the check

image

.github/workflows/markdown-link-check.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    env:
+      ZENML_DEBUG: 1
+      ZENML_ANALYTICS_OPT_IN: false
+    steps:
+    - uses: actions/checkout@master
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        use-quiet-mode: 'no'
+        use-verbose-mode: 'no'
```
strickvl, line 16: Not sure what this looks like when it runs, but is it worth keeping the verbose mode in? Some of our links are redirects, and I'm not sure if those fail or not?

strickvl (commented):
I guess when we turn this on it'll find a TON of broken links? Will we fix those in a separate PR, or do we just have no broken links?

fa9r (approved):
Overall looks good to me, but hard to review until it is live :D

Also, when exactly will this run? As one of the checks on each PR?

.github/workflows/markdown-link-check.yml
dnth: This is my latest run. We can change that if its needed. But I find a little clutter with the verbose mode.

https://github.com/dnth/zenml/runs/7726267706?check_suite_focus=true

dnth:
> 

From what I understand, yes it's one of the checks in each PR.

.github/workflows/markdown-link-check.yml
strickvl: Yeah I guess let's start with just the non-verbose mode and if we need more info we can always add it back in?

strickvl (commented):
Looks ok from what I can tell now. The main disadvantage of this method is that we aren't able to run it locally.

Usually with these kinds of things we're able to run the script locally so you can iterate quickly via your own terminal. The disadvantage from doing it all in Github is that you can't CTRL-CLICK the errors to open them directly in your code editor.

If there was a way to include this into a script that can run locally, like you have with something like [`deadlink`](https://github.com/nschloe/deadlink), I'd personally prefer that, but not sure whether once most links are fixed after we merge it, then perhaps we won't have so many to fix on a regular basis?

.github/workflows/markdown-link-check.yml
dnth: Sure. I'm open to changes to use other tools too in case this is a little cumbersome that the checks only happens on Github

dnth:
> Looks ok from what I can tell now. The main disadvantage of this method is that we aren't able to run it locally.
> 
> Usually with these kinds of things we're able to run the script locally so you can iterate quickly via your own terminal. The disadvantage from doing it all in Github is that you can't CTRL-CLICK the errors to open them directly in your code editor.
> 
> If there was a way to include this into a script that can run locally, like you have with something like [`deadlink`](https://github.com/nschloe/deadlink), I'd personally prefer that, but not sure whether once most links are fixed after we merge it, then perhaps we won't have so many to fix on a regular basis?

Is it viable if we use tools like deadlinks for local check and include it as part of our contributing guide and this tool we keep it on Github to catch any dead links before a merge? It's similar to what we do for spell checks.

strickvl:
I'd encourage something which can run locally and on Github. Any fixes that'll take place will happen locally, after all. Maybe compare the output for `deadlink` vs the GH cli tool on a few files? Now that we've fixed the broken URLs over lines issue, the output should be more or less identical?

dnth:
> I'd encourage something which can run locally and on Github. Any fixes that'll take place will happen locally, after all. Maybe compare the output for `deadlink` vs the GH cli tool on a few files? Now that we've fixed the broken URLs over lines issue, the output should be more or less identical?

I find that this tool can be run locally with an npm install which may introduce complexities. Here's the link showing how to run locally:

https://github.com/tcort/markdown-link-check

Also I find that the output for this tool is more verbose compared to deadlink in which it shows the exact file the dead links are from. 

Felix suggests that we could use different tools for local checks and leave this one on Github as an automated measure to catch any dead links in case the dead links are not checked locally.

So, to anyone who's trying to fix a dead link locally, use `deadlink` (or any other link checker tool) and fix it locally. And once the PR is finalized, then this github actions will run another check before merging.

Do you think this is something we can try now?

PR #829, opened 2022-08-08
Fix the SQL zenstore to work with MySQL

### LINKED ISSUE #788: [BUG]: profile with MySQL store doesn't work

_No response_

### System Information

System details are irrelevant

### What happened?

ZenML profiles with a MySQL server as back-end are not functional. E.g. the following command:

```
zenml profile create --store-type sql --url "mysql://zenml:password@localhost/zenml" mysql_profile
```

fails when trying to create the default components & stack:

```
OperationalError: (MySQLdb.OperationalError) (1822, "Failed to add the foreign key constraint. Missing index for constraint 'zenstackdefinition_ibfk_3' in the referenced table 'zenstackcomponent'")[SQL: 
CREATE TABLE zenstackdefinition (
stack_name VARCHAR(255) NOT NULL,
component_type VARCHAR(255) NOT NULL,
component_name VARCHAR(255) NOT NULL,
PRIMARY KEY (stack_name, component_type, component_name),
FOREIGN KEY(stack_name) REFERENCES zenstack (name),
FOREIGN KEY(component_type) REFERENCES zenstackcomponent (component_type),
FOREIGN KEY(component_name) REFERENCES zenstackcomponent (name)
)
```

### Reproduction steps

1. provision or use an existing MySQL server (e.g. the one provided with Kubeflow works nicely)
2. create a ZenML profile connected to the remote MySQL server. This will fail.

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefannica:
## Describe changes
SQLModel doesn't support compound foreign keys at the moment and some other foreign key constraints modeled with SQLModel were incorrect. I had to removed or replace them to make the SQL zenstore work with MySQL.

Unfortunately, this also means that this change breaks backwards compatibility.

We should re-add the removed constraints when we redefine the store to match the ZenML REST API updates.

Fixes #788

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -87,9 +87,9 @@ class ZenStackComponent(SQLModel, table=True):
     """SQL Model for stack components."""
 
     id: UUID = Field(primary_key=True)
-    component_type: StackComponentType
+    type: StackComponentType
     name: str
-    component_flavor: str
+    flavor: str
     configuration: bytes  # e.g. base64 encoded json string
 
 
@@ -109,11 +109,8 @@ class ZenStackDefinition(SQLModel, table=True):
     """
 
     stack_name: str = Field(primary_key=True, foreign_key="zenstack.name")
-    component_type: StackComponentType = Field(
-        primary_key=True, foreign_key="zenstackcomponent.component_type"
-    )
-    component_name: str = Field(
-        primary_key=True, foreign_key="zenstackcomponent.name"
+    component_id: UUID = Field(
+        primary_key=True, foreign_key="zenstackcomponent.id"
     )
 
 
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -517,16 +507,6 @@ def _update_stack_component(
             # handle any potential renamed component
             updated_component.name = component.name
 
-            # rename components inside stacks
```
stefannica, line 520: These are no longer needed if the component UUID is used as a foreign key instead of the name+type

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -592,25 +576,24 @@ def _save_stack(
                     session.delete(result)
 
             for ctype, cname in stack_configuration.items():
-                statement = (
```
stefannica, line 595: This was never needed, because the `ZenStackDefinition.stack_name <-> ZenStack.name` foreign key constraint prevents it, and we delete the definitions above, if the stack exists anyway.

schustmi (approved):

PR #833, opened 2022-08-09
Unify CLI concepts (removing `secret`, `feature` and `served-models`)

strickvl:
I merged 3 dual-concept CLI commands:

- `secrets-manager` + `secret`
- `feature-store` + `feature`
- `model-deployer` + `served-models`

The second of each pair now works as a subcommand, so:

```shell
zenml secrets-manager secret register ...

# or

zenml model-deployer models list
```

In this PR I updated the code, the docs and tests. I will make separate PRs to:

- the blog
- zenfiles
- zenbytes
- (maybe?) zenml.io website itself

Note that this is (obviously) a breaking change.

## Pre-requisites

## Types of changes


src/zenml/cli/stack_components.py
```diff
@@ -1362,6 +1380,988 @@ def flavor_group() -> None:
     )(explain_command)
 
 
+def register_secrets_manager_subcommands() -> None:
```
stefannica, line 1383: This file is already as bloated as it gets. Do you think you could keep these new `register_*` functions inside their original files, secret.py and served-models.py (perhaps renamed to model.py) instead ?

That should give some much needed structure to this already convoluted sub-command logic. Not to mention it should make the review easier :)

stefannica (changes requested):

src/zenml/cli/stack_components.py
strickvl: Yes of course. I'll do that.

strickvl: Done

src/zenml/cli/stack_components.py
```diff
@@ -1362,154 +1365,21 @@ def flavor_group() -> None:
...
-                raise ValueError("Dataset does not exist.") from e
-
-
 def register_all_stack_component_cli_commands() -> None:
     """Registers CLI commands for all stack components."""
     for component_type in StackComponentType:
         register_single_stack_component_cli_commands(
             component_type, parent_group=cli
         )
+
```
schustmi, line 1374: Do we need these calls actually in the for loop and check for the `component_type`, or can we simply call all of them after the call to `register_all_stack_component_cli_commands()` below?

schustmi (approved):
LGTM, just one minor comment that might make it a little cleaner

src/zenml/cli/stack_components.py
strickvl: Good point. Fixed.

src/zenml/cli/secret.py
```diff
@@ -39,472 +39,483 @@
...
+            active_stack = repo.zen_store.get_stack(name=repo.active_stack_name)
+            secrets_manager_wrapper = active_stack.get_component_wrapper(
+                StackComponentType.SECRETS_MANAGER
+            )
+            if secrets_manager_wrapper is None:
+                error(
+                    "No active secrets manager found. Please create a secrets manager "
+                    "first and add it to your stack."
+                )
+                return
```
stefannica, line 68: Here's an interesting idea: what if you don't even make these commands visible (don't register them) if the current stack doesn't include a secrets manager ? Would that be too confusing for users ?

src/zenml/cli/secret.py
```diff
@@ -39,472 +39,483 @@
...
-            "first and add it to your stack."
-        )
-        return
-
-    ctx.obj = secrets_manager_wrapper.to_component()
-
+def register_secrets_manager_subcommands() -> None:
+    """Registers CLI subcommands for the Secrets Manager."""
+    secrets_manager_group = cast(TagGroup, cli.commands.get("secrets-manager"))
+    if secrets_manager_group:
```
stefannica, line 45: tiny hint for reducing the indentation here:

```suggestion
    if not secrets_manager_group:
        return
        
    # the rest of the code follows with less indentation
```

src/zenml/cli/model.py
```diff
@@ -0,0 +1,438 @@
...
+from zenml.repository import Repository
+
+if TYPE_CHECKING:
+    from zenml.model_deployers import BaseModelDeployer
+
+
+def register_model_deployer_subcommands() -> None:
+    """Registers CLI subcommands for the Model Deployer."""
+    model_deployer_group = cast(TagGroup, cli.commands.get("model-deployer"))
+    if model_deployer_group:
```
stefannica, line 40: same hint here about reducing the code indentation

stefannica (approved):
Looks great ! Had a few minor suggestions for reducing indentation, up to you if you want to include them or not.

src/zenml/cli/secret.py
strickvl: I think I'd find that confusing, and also possibly would lead to confusion when debugging with a user, perhaps? I like the principle of the idea, but presumably we should also do the same for everything else? I.e. the CLI only shows you commands that are available to you?

strickvl:
@stefannica reworked the files with your nice indentation trick. Thanks! Will just wait for tests to pass and then will merge.

PR #836, opened 2022-08-09
Fix served models logs formatting error

safoinme:
## Describe changes
Add a try expect to the served models get logs to CLI function to pass formatting error that rich.console fails on

## Pre-requisites

## Types of changes


src/zenml/cli/served_models.py
```diff
@@ -418,4 +419,7 @@ def get_model_service_logs(
         if raw or line.startswith("\x1b["):
             print(line)
         else:
-            console.print(line)
+            try:
+                console.print(line)
+            except MarkupError:
+                print(line)
```
strickvl, line 425: Which `print` does this use? `rich` print or the Python print?

safoinme: @strickvl  That's python print as rich fail to format it! i couldn't find a way of how to do it with rich print

strickvl: Got it. If it's not too much extra work, it'd be nice to reproduce this error in a test, but otherwise the fix looks ok to me :)

strickvl (approved):
LGTM!

fa9r (commented):
Is this an issue that we might also encounter in other places? If so, maybe we can patch the `zenml.console.print` to always catch markdown errors like this. Actually, that should be a feature of rich, did you check whether `rich.console.Console` has an option to configure markdown error handling?

safoinme:
@fa9r  rich documentation doesn't provide enough information about this. it only says that 
```
Rich will raise MarkupError if the markup contains one of the following errors:

Mismatched tags, e.g. "[bold]Hello[/red]"

No matching tag for implicit close, e.g. "no tags[/]"

```
Regarding the question, I think yes if there is one of these two cases any other `console.print` may fail and I don't think `rich.console.Console` has any argument for this

fa9r (approved):

fa9r:
@safoinme Okay, I see. Yeah then let's just merge it like this for now and potentially revisit and generalize the fix once we encounter the same issue in other places.

PR #837, opened 2022-08-10
Spark Integration

bcdurak:
Hey everyone and welcome to the long-awaited Spark PR. 

With this PR, we go over a new integration with ZenML and you can think of it as our first steps toward the world of distributed programming.

**CAUTION**: This PR already has the docker configuration branch `feature/ENG-1071-docker-flexibility` merged.

# Summary

Within the integration, you will find two distinct groups of implementations:

- Materializers: `SparkDataFrameMaterializer` and `SparkModelMaterializers` 
- Step Operator: `SparkStepOperator` (base) and `KubernetesSparkStepOperator`

While materializers are quite straightforward and support the PySpark DataFrame API and the models in their machine learning library MLLib respectively, the step operators are a bit more complicated. They are responsible for launching a Spark job based on the given step and there are a few things to consider here.

Let's talk about the configuration first. Under the base class, you will find 4 instance parameters.
	- `master` is the master URL for the cluster. You might see different schemes for different cluster managers which are supported by Spark like Mesos, YARN, or Kubernetes. Within the context of this PR, the implementation supports **Kubernetes** as a cluster manager.
	- `deploy_mode` can either be 'cluster' (default) or 'client' and it decided where the driver node of the application will run.
	- `submit_args` is the JSON string of a dict, which will be used to define additional params if required (Spark has quite a lot of different parameters, so including them, all in the step operator was not implemented). 

For Kubernetes, there are also some additional important parameters: 
	- `kubernetes_namespace` is the namespace under which the driver and executor pods will run.
	- `kubernetes_service_account` is the service account that will be used by various Spark components (to create and watch the pods).
	- `docker_parent_image` (which originally comes from the `PipelineDockerImageBuilder` base class) indicates the name of a base image that has Spark enabled.

In addition to these parameters, the `launch` method of the step operator gets a `DockerConfiguration` and a `ResourceConfiguration`. Utilizing all of this, the overall configuration happens in 4 base methods:

- `_resource_configuration` which translates our configuration object to Spark's resource configuration.
- `_backend_configuration` which is responsible for cluster-manager-specific configuration.
- `_io_configuration` is one of the critical ones. Even though we have materializers, Spark might require additional packages and configuration to work with a specific filesystem. For this implementation, I have implemented a way for Spark to recognize that it is operating with an S3 artifact store and configure itself accordingly. While this works on S3 at the moment, this also means that other artifact store flavors might be problematic if left without any additional configuration.
- `_additional_configuration` takes the submit args and utilizes them to create a further configuration.

Once the configuration is completed, `_launch_spark_job` comes into play. This takes the completed config object and is responsible to run it on the given `master` URL with the specified `deploy_mode`. In the base class, this is achieved by using `spark-submit` as using pyspark makes us use `spark-shell` which does not support `cluster` as a `deploy-mode`.

All the configuration methods and the launch method are extendible. In fact, the `KubernetesSparkStepOperator` is implemented by just overwriting the `_backend_configuration` method of the base class.

# Possible areas for improvement

- **IMPORTANT**: This is a bit challenging to explain but I will give it a shot here. In an ideal case, when you are submitting an application to Spark, you have various ways to define additional packages which are required within the environment where Spark is running. As I was using an S3 artifact store, I quickly found out that this was the case for me. When I tried to install these packages (through various means) in the cluster mode, they only got installed to the driver pods but not the executor pods, which ended up failing my jobs. As a solution, I figured out that I can install these packages myself and create the base image using Spark's `docker-image-tool` locally (which copies over your local Spark installation along with the packages). I would very much like to avoid this solution and not force people to do what I did if they want to make it work with any artifact store which requires additional packages.
- Once we figure out how to parse it, we can add a new `ResourceConfiguration` to the implementation for Spark.
- Adding some integration tests

# Small TODOs

- Write the corresponding docs pages
- Finishing up the docstrings
- Finishing up the README of the example
- Detect whether Spark is installed
- Update the integration table
 
## Pre-requisites

## Types of changes


src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,325 @@
...
+
+        This is especially critical when using 'spark-submit' as it defines the
+        path (to the application in the environment where Spark is running)
+        which is used within the command.
+
+        For more information on how to set this property please check:
+
+        https://spark.apache.org/docs/latest/submitting-applications.html#advanced-dependency-management
+        """
+        return None
```
AlexejPenner, line 60: This is a property that always returns None? What am I missing, why can't this be an optional instance parameter?

src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,325 @@
...
+        else:
+            logger.warning(
+                "In most cases, the spark step operator requires additional "
+                "configuration based on the artifact store flavor you are "
+                "using. That also means, that when you use this step operator "
+                "with certain artifact store flavor, ZenML can take care of "
+                "the pre-configuration. However, the artifact store flavor "
+                f"'{artifact_store.FLAVOR}' featured in this stack is not "
+                f"known to this step operator and it might require additional "
+                f"configuration."
```
AlexejPenner, line 203: Maybe it could be helpful to link to docs here or give an example of where and how the user would specify these additional parameters.

AlexejPenner, line 203: Maybe it could be helpful to link to docs here or give an example of where and how the user would specify these additional parameters.

src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,325 @@
...
+            "--input_artifact_types_path",
+        ]:
+            i = entrypoint_command.index(arg)
+            command.extend([arg, entrypoint_command[i + 1]])
+
+        command = " ".join(command)
+
+        # Execute the spark-submit
+        process = subprocess.Popen(
+            command,
```
AlexejPenner, line 264: I think its not necessary to join all the command items into one single string, to the best of my knowledge Popen will also take a list of strings.

AlexejPenner (changes requested):
Looks good to me on the Spark abstraction - I left some comments and I explicitely skipped any changes related to https://github.com/zenml-io/zenml/pull/811

docs/book/developer-guide/advanced-usage/docker.md
```diff
@@ -31,60 +41,155 @@ following two ways:
...
-You don't need to install the `zenml` pip package as well as
-any integration that is used for components of your active stack.
+@pipeline(docker_configuration=docker_config)
+def my_pipeline(...):
+    ...
+```
+
+{% hint style="warning" %}
+This is an advanced feature and will most likely break your pipelines. If you use this,
+you're on your own and need to copy all the necessary files to the correct paths yourself.
```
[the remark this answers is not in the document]
schustmi: I'll answer here as these are my changes from the docker PR: I actually intended this as "harsh" as possible. In my opinion this option is super super advanced, and we definitely don't want to support/help debug any issue that comes from it. Same for the stack config below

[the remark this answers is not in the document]
AlexejPenner: Makes sense, I'll retract the comment then

src/zenml/integrations/spark/step_operators/spark_step_operator.py
bcdurak: The application path is a property that _might_ be used within the `_launch_spark_job` method depending on the case. This is mainly the reason why I did not want to turn it into an instance parameter.

In cases such as the `KubernetesSparkStepOperator`, it points to where the entrypoint for the step lies (for the `spark-submit` command). This way, if our users decide to use a different base image (with a different application path), they can just change this property without changing the `_launch_spark_job` method.

src/zenml/integrations/spark/step_operators/spark_step_operator.py
bcdurak: You are absolutely right. I have kept it there for debugging purposes. I will update and resolve it.

src/zenml/integrations/spark/step_operators/spark_step_operator.py
schustmi: Does it make sense to at least make it return `Optional[str]` in this case?

bcdurak: Makes sense. Fixed it.

examples/spark/steps/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    SparkStepOperator,
+)
+from zenml.repository import Repository
+
+step_operator = Repository().active_stack.step_operator
+if not step_operator or not isinstance(step_operator, SparkStepOperator):
+    raise RuntimeError(
+        "Your active stack needs to contain a Spark step operator for this "
+        "example to work."
+    )
```
fa9r, line 27: This is really nice. Maybe we should also do this for other examples?

examples/spark/steps/importer_step/importer_step.py
```diff
@@ -0,0 +1,30 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+
+from pyspark.sql import DataFrame, SparkSession
+
+from zenml.repository import Repository
+from zenml.steps import BaseStepConfig, step
+
+step_operator = Repository().active_stack.step_operator
```
fa9r, line 20: maybe here you could make sure that this really is a Spark step operator instead

examples/spark/steps/transformer_step/transformer_step.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@step(custom_step_operator=step_operator.name)
+def transformer_step(
+    train: DataFrame,
+) -> Output(train_xf=DataFrame, scaler=Transformer, encoder=Transformer,):
+    # Cast the columns to the right types
+    for column in train.columns:
+        train = train.withColumn(column, col(column).cast("float"))
+    train = train.withColumn("label", col("label").cast("int"))
```
fa9r, line 33: feels a bit odd to cast label twice

src/zenml/integrations/spark/materializers/spark_dataframe_materializer.py
```diff
@@ -0,0 +1,51 @@
...
+        Returns:
+            A loaded spark dataframe.
+        """
+        super().handle_input(data_type)
+        # Create the Spark session
+        spark = SparkSession.builder.getOrCreate()
+
+        # Read the data
+        path = os.path.join(self.artifact.uri, DEFAULT_FILEPATH)
+        return spark.read.parquet(path)
```
schustmi, line 40: Does this actually work with any artifact stores, e.g. Azure?

schustmi: Same for the other materializer read/write methods

src/zenml/integrations/spark/step_operators/kubernetes_step_operator.py
```diff
@@ -0,0 +1,121 @@
...
+
+        spark_config: a SparkConf object which collects all the
+                configuration parameters
+        docker_configuration: The Docker configuration for this step.
+        pipeline_name: Name of the pipeline which the step to be executed
+                is part of.
+        """
+        # Copy the entrypoint
+        entrypoint_path = os.path.join(get_source_root_path(), ENTRYPOINT_NAME)
+        copy(LOCAL_ENTRYPOINT, entrypoint_path, overwrite=True)
```
schustmi, line 100: This file needs to be removed after the docker build. I also think it should fail instead of potentially overwriting a user file with the same name?

src/zenml/integrations/spark/step_operators/kubernetes_step_operator.py
```diff
@@ -0,0 +1,121 @@
...
+from zenml.repository import Repository
+from zenml.runtime_configuration import RuntimeConfiguration
+from zenml.integrations.spark.step_operators import spark_entrypoint
+from zenml.utils.pipeline_docker_image_builder import PipelineDockerImageBuilder
+from zenml.utils.source_utils import get_source_root_path
+
+logger = get_logger(__name__)
+
+LOCAL_ENTRYPOINT = spark_entrypoint.__file__
+APP_DIR = "/app"
```
schustmi, line 35: This is already defined in the pipeline_docker_image_builder.py as the constant `DOCKER_IMAGE_WORKDIR`

src/zenml/integrations/spark/step_operators/kubernetes_step_operator.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Class configuration
+    FLAVOR: ClassVar[str] = SPARK_KUBERNETES_STEP_OPERATOR
+
+    @property
+    def application_path(self) -> Any:
+        """Provides the application path in the corresponding docker image."""
+        return f"local://{APP_DIR}/{ENTRYPOINT_NAME}"
+
+    # @staticmethod
```
schustmi, line 56: Can we get rid of this?

examples/spark/pipelines/spark_pipeline/spark_pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+@pipeline
+def spark_pipeline(
+    importer,
+    analyzer,
+    splitter,
+    processer,
+    trainer,
+):
+    dataset = importer()
+    analyzer(dataset)
```
fa9r, line 28: this `analyzer` step feels a bit incomplete here, if there is a step like this in the pipeline I would naturally be curious to know how this can now be accessed/visualized. If we have some way to visualize the pandas `describe` as part of the post execution, then I would include that code in `run.py`, otherwise I would propose to remove this step.

examples/spark/README.md
```diff
@@ -0,0 +1,325 @@
+# The distributed programming with Spark
```
fa9r, line 1:
```suggestion
# Distributed Data Processing with Spark
```

schustmi (changes requested):

examples/spark/pipelines/spark_pipeline/spark_pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pipeline
+def spark_pipeline(
+    importer,
+    analyzer,
+    splitter,
+    processer,
+    trainer,
+):
```
fa9r, line 26: Needs a docstring that explains the pipeline a bit

fa9r (approved):
Awesome integration, I love the interface and basic abstractions. In particular, `launch()` feels very nicely decoupled and generally polished.

Regarding the code, the only major question I had was whether the `SparkStepOperator` should be an abstract class instead. 

The example still needs quite a bit of work I think, especially the README intro. I will already approve anyways since we can backport the example in a separate PR if necessary.

Great job overall, can't wait to finally get this live!

Time to celebrate! 🎉 😄

examples/spark/steps/__init__.py
AlexejPenner: It

src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,321 @@
...
+
+        if resource_configuration.memory:
+            # TODO[LOW]: Fix the conversion of the memory unit with a new
+            #   type of resource configuration.
+            spark_config.set(
+                "spark.executor.memory",
+                resource_configuration.memory.lower().strip("b"),
+            )
+
+    def _backend_configuration(
```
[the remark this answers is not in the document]
AlexejPenner: +1

src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,321 @@
...
+                        ("spark.hadoop.fs.s3a.access.key", f"{key}"),
+                        ("spark.hadoop.fs.s3a.secret.key", f"{secret}"),
+                    ]
+                )
+            else:
+                raise RuntimeError(
+                    "When you use an Spark step operator with an S3 artifact "
+                    "store, please make sure that your artifact store has"
+                    "defined the required credentials namely the access key "
+                    "and the secret access key."
```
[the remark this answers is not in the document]
AlexejPenner: s3 artifact store does no necessarily mean AWS

AlexejPenner (approved):
This integration **spark**s joy - LGTM

src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,321 @@
...
+if TYPE_CHECKING:
+    from zenml.config.resource_configuration import ResourceConfiguration
+
+LOCAL_ENTRYPOINT = entrypoint.__file__
+APP_DIR = "/app/"
+CONTAINER_ZENML_CONFIG_DIR = ".zenconfig"
+ENTRYPOINT_NAME = "zenml_spark_entrypoint.py"
+
+
+class SparkStepOperator(BaseStepOperator):
```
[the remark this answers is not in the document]
fa9r: Update: no, it should not be an abstract class, since you can technically use the base class itself and specify everything as submit args instead.

src/zenml/integrations/spark/step_operators/spark_step_operator.py
```diff
@@ -0,0 +1,321 @@
...
+                )
+            else:
+                raise RuntimeError(
+                    "When you use an Spark step operator with an S3 artifact "
+                    "store, please make sure that your artifact store has"
+                    "defined the required credentials namely the access key "
+                    "and the secret access key."
+                )
+        else:
+            logger.warning(
```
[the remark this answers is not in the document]
fa9r: Update: the users can specify themselves using submit args

src/zenml/utils/source_utils.py
```diff
@@ -98,15 +96,7 @@ def is_third_party_module(file_path: str) -> bool:
...
-
-    for path in site.getsitepackages() + [
-        site.getusersitepackages(),
-        get_python_lib(standard_lib=True),
-    ]:
-        if pathlib.Path(path).resolve() in absolute_file_path.parents:
-            return True
-
-    return False
+    return get_source_root_path() not in absolute_file_path.parents
```
AlexejPenner, line 99: get_source_root_path returns a str. However, a casting to Path would be necessary as the current implementation  seems to not work as expected:

```python
'/home/zenml/tests/unit/utils' in Path('/home/zenml/src/zenml/utils/source_utils.py').parents

False
```

I'm also not sure this work as expected in general. The root repo might very well be at the same level as the venv:

|-.zen
|-venv
|user_src

As such this will make all packages in venv non-third party packages:

e.g.:
```python
`/home/apenner/PycharmProjects/zenml` not in `/home/apenner/PycharmProjects/zenml/venv/lib/python3.8/site-packages/_pytest/cacheprovider.py`

False
```

PR #841, opened 2022-08-11
Custom deployment with KServe and Seldon Core

safoinme:
Hey everyone, This is the first version of the custom deployment story with Seldon and KServe.

With this PR, we will see how we can extend the deployment story with ZenML to not only support custom deployment within Seldon and KServe but also simplify it as much as possible while still allowing as much customizability as possible for advanced use cases.

**CAUTION**: This PR already has merged the docker configuration branch `feature/ENG-1071-docker-flexibility` into it, feel free to skip the docker and orchestrators-related files or if you are interested you can check it from [here](https://github.com/zenml-io/zenml/pull/811).

# Introduction
With ZenML model deployers integrations users can already deploy their trained ml models either into a local (dev) or remote (prod) environment! While these integrations provide a powerful set of features and functionalities to deploy different frameworks such as _**(SKLearn, TensorFlow and PyTorch)**_ it does this with something called built-in servers _**(MLServer, TFServing and TorchServe)**_ which are built or added services for each of the famous ml frameworks. 
> Note: This model server usually requires the model to be in an exact format and to be saved in central storage that the serving tool can access to be able to load it! 💡💡 **_ZenML already takes care of all the preparation and saving of the model artifact into the central repository so the model deployer can access it. using the built-in deployment step ZenML takes a specific model artifact and the framework type from the parameters prepare it to the right format, saves it in the artifact store part of the Stack and provides the serving tool URI for where the new artifact is saved_** 💡💡

While it's true that the out-of-the-box model servers are low engineering effort and only require simple preparing and saving <ins>[except for PyTorch/TorchServe which is only supported in KServe](https://kserve.github.io/website/0.9/modelserving/v1beta1/torchserve/)</ins>. They are either limited to simple use cases where the checkpoint of the trained model is all you need **_(That's how ZenML saves the models with the materializers)_** or they are very complicated and challenging to adapt to more advanced use cases that require some sort of preprocessing or postprocessing functionalities before running prediction with the model. 

For those reasons, all serving tools provide a way for the users to create and deploy Custom Python Models
- KServe call this: A custom predictor and it can be done by extending their Model class `kserve.Model` and packaging the new model class and the artifact into a docker image using [Buildpacks](https://buildpacks.io/docs/tools/pack/) after that user can create an inferenceservice CRD manifest that uses the new docker image as the predictor. [read more about it from here](https://kserve.github.io/website/0.9/modelserving/v1beta1/custom/custom_model/)
- Seldon calls this:  A Non-Reusable Model Server which can be done by creating a class that shares the name of the file where it's defined. The class should also have a defined `predict` function that the seldon-core-microservice will be using to run prediction on the loaded model. Everything then should be packaged in a docker image that Seldon deployment CRD can use. [read more about it from here](https://docs.seldon.io/projects/seldon-core/en/latest/python/python_wrapping_docker.html)

# Custom Code Implementation
Now that we have a general overview of the current deployment story and how it's using only the model servers. and we mentioned a small summary of the required process that Seldon and KServe require to do custom deployment let's see how we can achieve that within ZenML scope.

We have mainly 5 components (pieces of code) that create the custom deployment story and they are: 

1. `custom predict function`: the code defined by the user to do custom operation needed! in the current implementation, the custom predict function must get 2 parameters: * model: which will be the model loaded in memory. * request: and that would be the request coming from external. any pre or post-process can be defined with the custom predict and be called within the definition of it. [Please click here to see an example of the code](https://github.com/zenml-io/zenml/blob/feature/unified-custom-deployment/examples/custom_code_deployment/steps/pytorch_steps/custom_deploy_functions.py).

2. `zenml_custom_model`: This is a file that has an implementation on every model deployer integration and it contains the custom class definition that each of the tools. This file has 3 objectives: load the model into memory using the materializers, call the custom to predict the function defined by the user, and start the server using the main function implemented that can be triggered using an entry point. [Please click here to see an example of the code for kserve](https://github.com/zenml-io/zenml/blob/feature/unified-custom-deployment/src/zenml/integrations/kserve/custom_deployer/zenml_custom_model.py) or [seldon](https://github.com/zenml-io/zenml/blob/feature/unified-custom-deployment/src/zenml/integrations/seldon/custom_deployer/zenml_custom_model.py) 

3. `custom_deployment_step`: The built-in custom deployment step is responsible for preparing the service config with all required configurations to deploy the model and this includes: * The different parameters for the deployment. * entry point used with the new custom docker image. * call function responsible for saving the model metadata that will be used to load model into memory. * save the model checkpoint and the model metadata in the artifact store and start deployment with all required configurations. [Please click here to see an example of the code](https://github.com/zenml-io/zenml/blob/feature/unified-custom-deployment/src/zenml/integrations/seldon/steps/seldon_deployer.py#L245).

4. `model_deployer`: the integration implementation of the model deployer. it has one new role which is to prepare a custom deployment docker image. if the stack is using a remote orchestrator it will return the same image if the stack has a local orchestrator it will build a new image. [Please click here to see an example of the code](https://github.com/zenml-io/zenml/blob/feature/unified-custom-deployment/src/zenml/integrations/kserve/model_deployers/kserve_model_deployer.py#L518).

5. `materializer_utils`: a save and load functionalities. save takes the model artifact and extracts necessary information such as the used materializer and datatype class and saves it to the YAML file. and load uses the saved data to load the model back to memory so it can be used. with these 2 functionalities, users don't need to write the load function as long as they provide the model artifact. (Also this way we don't need access to the stack, metadata store, or even artifact store. This could be a start point for how a model registry can be used in a remote environment) [Please click here to see an example of the code](https://github.com/zenml-io/zenml/blob/feature/unified-custom-deployment/src/zenml/utils/materializer_utils.py).
 
# potential improvements

## TODO List:
- [ ] Write detailed docs on this
- [ ] Finish Readme of the example 
- [ ] Add a TensorFlow example

## Pre-requisites

## Types of changes


examples/custom_code_deployment/steps/prediction_service_loader/prediction_service_loader.py
```diff
@@ -0,0 +1,102 @@
...
+
+from zenml.integrations.kserve.model_deployers import KServeModelDeployer
+from zenml.integrations.kserve.services import KServeDeploymentService
+from zenml.integrations.seldon.model_deployers.seldon_model_deployer import (
+    SeldonModelDeployer,
+)
+from zenml.integrations.seldon.services.seldon_deployment import (
+    SeldonDeploymentService,
+)
+from zenml.steps import BaseStepConfig, step
```
AlexejPenner, line 24: Seems to me like this would import both the Kserve as well as the Seldon specific integrations which would fail if not both are installed. how would you feel about moving the imports into the respective functions. 

In general one should be able to run the example wiuth either seldon or kserve installed.

AlexejPenner: As the two functions are very similar they could maybe even be unified and use the active_stack to get the active model deployer, in that case you might not even need explicit imports?

examples/custom_code_deployment/steps/predictor/predictor.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    service.start(timeout=120)  # should be a NOP if already started
+    response = service.predict(data)
+    rich_print(response)
+
+
+@step
+def seldon_predictor(
+    service: KServeDeploymentService,
```
AlexejPenner, line 43: seldon not kserve

AlexejPenner (commented):
I have read through the examples so far, and I like the overall deployment story that is shaping up :)

examples/custom_code_deployment/steps/prediction_service_loader/prediction_service_loader.py
safoinme: Yes, that's true, users would be required to have both installed! it should be either changed to be imported after getting the model deployer option in the command line or would be even better to get it using the active stack.

One question, would it be better to just split the examples or create multiple sub-examples in the custom code example? cause there will be a TensorFlow (maybe sklearn) example added?

src/zenml/integrations/seldon/services/seldon_deployment.py
```diff
@@ -33,7 +33,7 @@
 from zenml.services.service_type import ServiceType
 
 if TYPE_CHECKING:
```
fa9r, line 35: if you don't need any imports here, you can just delete this line and the next

src/zenml/integrations/seldon/model_deployers/seldon_model_deployer.py
```diff
@@ -470,3 +476,46 @@ def delete_model_server(
         # secret used to store the authentication information for the Seldon
         # Core model server storage initializer
         self._delete_kubernetes_secret()
+
+    def prepare_custom_deployment_image(
```
fa9r, line 480: duplicate code with kserve?

src/zenml/integrations/seldon/steps/seldon_deployer.py
```diff
@@ -36,10 +43,49 @@
...
+        Args:
+            v: predict function path
+
+        Returns:
+            predict function path
+
+        Raises:
+            ValueError: if predict function path is not valid
+        """
+        if not v:
```
fa9r, line 80: bad variable name `v`

fa9r: also, duplicate code from Kserve

src/zenml/integrations/seldon/steps/seldon_deployer.py
```diff
@@ -192,3 +239,149 @@ def prepare_service_config(model_uri: str) -> SeldonDeploymentConfig:
     )
 
     return service
+
+
+@step(enable_cache=False)
+def seldon_custom_model_deployer_step(
```
fa9r, line 245: this also has a lot of duplicate code with kserve

fa9r (approved):
Awesome PR, looking forward to get this live 👍 

The core code is very clean and nicely commented. Really great job on that! The example is also really cool. Love it! 

Most review comments I left are related to code duplication - I think there is a lot of duplicate code between KServe and Seldon. It might make sense to refactor this into a common base class or something similar. You don't necessarily need to do this in this PR, but I'd suggest to create a ticket for it so we don't forget.

Another general question I had, why are the custom models defined in a new directory `integrations/.../custom_deployer/` instead of `integrations/.../model_deployers/`? Since this is directly related to deployment, that's where I would have expected it.

Apart from that all good. I'll already approve since I think none of my remarks are critical issues. Feel free to rerequest review though if you think it might be helpful.

examples/custom_code_deployment/run_pytorch.py
```diff
@@ -0,0 +1,232 @@
...
+                    f"To stop the service, run "
+                    f"[italic green]`zenml served-models delete "
+                    f"{str(service.uuid)}`[/italic green]."
+                )
+            elif service.is_failed:
+                print(
+                    f"The KServe prediction server is in a failed state:\n"
+                    f" Last state: '{service.status.state.value}'\n"
+                    f" Last error: '{service.status.last_error}'"
+                )
```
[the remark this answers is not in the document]
safoinme: Changed the structure a bit, to avoid require to install kserve and seldon integration at the same time which doesn't make sense

examples/kserve_deployment/steps/pytorch_steps/pytorch_inference_processor.py
```diff
@@ -46,5 +46,6 @@ def pytorch_inference_processor(
     img = urlopen(config.img_url).read()
     image_64_encode = base64.b64encode(img)
     bytes_array = image_64_encode.decode("utf-8")
-    request = {"data": bytes_array}
-    return json.dumps([request])
+    instances = {"data": bytes_array}
+    request = [instances]
+    return json.dumps(request)
```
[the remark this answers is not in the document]
safoinme: because the built-in predict function on both kserve and seldon are changed to support also custom dictionaries

examples/kserve_deployment/steps/pytorch_steps/pytorch_inference_processor.py
```diff
@@ -27,7 +27,7 @@ class PyTorchInferenceProcessorStepConfig(BaseStepConfig):
 
     img_url: Optional[
         str
-    ] = "https://raw.githubusercontent.com/kserve/kserve/master/docs/samples/v1beta1/torchserve/v1/imgconv/0.png"
+    ] = "https://raw.githubusercontent.com/kserve/kserve/master/docs/samples/v1beta1/torchserve/v1/imgconv/1.png"
```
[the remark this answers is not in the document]
safoinme: there were two links one for the image of "0" and the other one for "1", I was running an example and showed that we would give the model a picture of "0" while actually the one that had priority was the one for "1" so it was like the model is wrong! to avoid confusion make just pic of one of them

src/zenml/integrations/kserve/custom_deployer/zenml_custom_model.py
```diff
@@ -0,0 +1,170 @@
...
+    type=click.STRING,
+    help="The name of the model to deploy. This important for the Kserve server.",
+)
+@click.option(
+    "--predict_func",
+    required=True,
+    type=click.STRING,
+    help="The path to the custom predict function defined by the user.",
+)
+def main(model_name: str, model_uri: str, predict_func: str) -> None:
```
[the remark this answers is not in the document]
safoinme: The way the custom deployment server works with the KServe server is by implementing a custom model class and passing it to the KServe server and then starting the server. Because custom classes usually need some parameters to be passed to the model, the parameters are passed from the entry point to the main function as arguments and then passed to the model class constructor.

src/zenml/integrations/kserve/steps/kserve_step_utils.py
```diff
@@ -27,8 +27,14 @@
...
 )
 from zenml.io import fileio
+from zenml.logger import get_logger
 from zenml.utils import io_utils
 
+logger = get_logger(__name__)
+
+
+ARTIFACT_FILE = "artifact.json"
+
```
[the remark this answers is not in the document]
safoinme: should be deleted! that's part of the step file now

src/zenml/integrations/seldon/model_deployers/seldon_model_deployer.py
safoinme: Yes, I had a discussion with @schustmi  about this. since not all deployers have to do this step and it will be different definition for bentoML and MLFlow it is okay to have it duplicated

src/zenml/integrations/kserve/custom_deployer/zenml_custom_model.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        if self.predict_func is not None:
+            try:
+                prediction = {
+                    "predictions": self.predict_func(
+                        self.model, request["instances"]
+                    )
+                }
+            except Exception as e:
+                raise Exception("Failed to predict: {}".format(e))
```
schustmi, line 110: I think raising an `Exception` isn't recommended, instead we should probably use a `RuntimeError` here

src/zenml/integrations/kserve/custom_deployer/zenml_custom_model.py
```diff
@@ -0,0 +1,172 @@
...
+            if isinstance(prediction, dict):
+                return prediction
+            else:
+                raise Exception("Prediction is not a dictionary.")
```
schustmi, line 114:
```suggestion
                raise TypeError("Prediction is not a dictionary.")
```
Maybe the error message could also be a little more expressive, e.g. that the predict function returned an object which was not a dict, and then include the type of the returned object (`type(prediction)`)

src/zenml/utils/materializer_utils.py
```diff
@@ -0,0 +1,103 @@
...
+        "materializer"
+    ]
+    materializer_class = source_utils.load_source_path_class(
+        model_artifact.properties["materializer"].string_value
+    )
+    model_class = source_utils.load_source_path_class(
+        model_artifact.properties["datatype"].string_value
+    )
+    materialzer_object = materializer_class(model_artifact)
+    model = materialzer_object.handle_input(model_class)
```
schustmi, line 94: typo in materializer

src/zenml/utils/materializer_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        import torch.nn as nn
+
+        if issubclass(model_class, nn.Module):  # type: ignore
+            model.eval()
+    except ImportError:
+        pass
+    logger.debug(f"model loaded successfully :\n{model}")
```
schustmi, line 102:
```suggestion
    logger.debug(f"Model loaded successfully :\n{model}")
```

src/zenml/integrations/kserve/steps/kserve_step_utils.py
```diff
@@ -27,8 +27,11 @@
     KServeDeployerStepConfig,
 )
 from zenml.io import fileio
+from zenml.logger import get_logger
 from zenml.utils import io_utils
 
+logger = get_logger(__name__)
```
schustmi, line 33: Not needed I guess?

src/zenml/integrations/kserve/services/kserve_deployment.py
```diff
@@ -58,16 +57,19 @@ class KServeDeploymentConfig(ServiceConfig):
...
         resources: the Kubernetes resources to allocate for the prediction service.
+        containers: the containers to use for the custom prediction services.
     """
 
     model_uri: str = ""
     model_name: str
     secret_name: Optional[str]
     predictor: str
     replicas: int = 1
+    containers: Optional[Dict[str, Any]]
```
schustmi, line 72: From what I understand, this is actually a single container right? Would it make sense to rename this to `container` instead?

src/zenml/integrations/kserve/steps/kserve_deployer.py
```diff
@@ -163,6 +169,37 @@ def torch_config_validate(cls, v: Optional[str]) -> Optional[str]:
...
+        Args:
+            predict_func_path: predict function path
+
+        Returns:
+            predict function path
+
+        Raises:
+            ValueError: if predict function path is not valid
+        """
+        if not predict_func_path:
```
schustmi, line 194: I think pydantic already covers this check as this is a required, non-optional string attribute

src/zenml/integrations/kserve/steps/kserve_deployer.py
```diff
@@ -163,6 +169,37 @@ def torch_config_validate(cls, v: Optional[str]) -> Optional[str]:
...
+            raise ValueError("Predict function path is required.")
+        try:
+            import_class_by_path(predict_func_path)
+        except AttributeError:
+            raise ValueError("Predict function can't be found.")
+        return predict_func_path
```
schustmi, line 200: Should we also add a check here to make sure the predict function is actually a function/something callable?

src/zenml/integrations/kserve/steps/kserve_deployer.py
```diff
@@ -278,3 +316,151 @@ def kserve_model_deployer_step(
...
+        ValueError: if the custom deployer parameters is not defined
+        DoesNotExistException: if no active stack is found
+
+
+    Returns:
+        KServe deployment service
+    """
+    # verify that a custom deployer is defined
+    if not config.custom_deploy_parameters:
+        raise ValueError("Custom deploy parameters are required.")
```
schustmi, line 349: I agree with felix here, this could benefit from a detailed error.

src/zenml/integrations/kserve/steps/kserve_deployer.py
```diff
@@ -278,3 +316,151 @@ def kserve_model_deployer_step(
...
+    # copy the model files to a new specific directory for the deployment
+    served_model_uri = os.path.join(context.get_output_artifact_uri(), "kserve")
+    fileio.makedirs(served_model_uri)
+    io_utils.copy_dir(model.uri, served_model_uri)
+
+    # Get the model artifact to extract information about the model
+    # and how it can be loaded again later in the deployment environment.
+    artifact = stack.metadata_store.store.get_artifacts_by_uri(model.uri)
+    if not artifact:
+        raise DoesNotExistException("No artifact found at {}".format(model.uri))
```
schustmi, line 430:
```suggestion
        raise DoesNotExistException(f"No artifact found at {model.uri}.")
```

src/zenml/integrations/seldon/custom_deployer/zenml_custom_model.py
```diff
@@ -0,0 +1,188 @@
...
+
+logger = get_logger(__name__)
+
+DEFAULT_MODEL_NAME = "model"
+DEFAULT_LOCAL_MODEL_DIR = "/mnt/models"
+
+Array_Like = Union[np.ndarray, List[Any], str, bytes, Dict[str, Any]]
+
+
+class ZenMLCustomModel(object):
```
schustmi, line 33:
```suggestion
class ZenMLCustomModel:
```

src/zenml/integrations/seldon/custom_deployer/zenml_custom_model.py
```diff
@@ -0,0 +1,188 @@
...
+        (
+            f'[{{"name":"model_uri","value":"{model_uri}","type":"STRING"}},'
+            f'{{"name":"model_name","value":"{model_name}","type":"STRING"}},'
+            f'{{"name":"predict_func","value":"{predict_func}","type":"STRING"}}]'
+        ),
+    ]
+    try:
+        subprocess.check_call(command)
+    except subprocess.CalledProcessError:
+        logger.error("Failed to start the seldon-core-microservice process.")
```
schustmi, line 183: Maybe also log the error here?

src/zenml/integrations/seldon/custom_deployer/zenml_custom_model.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        if self.predict_func is not None:
+            try:
+                prediction = {"predictions": self.predict_func(self.model, X)}
+            except Exception as e:
+                raise Exception("Failed to predict: {}".format(e))
+            if isinstance(prediction, dict):
+                return prediction
+            else:
+                raise Exception("Prediction is not a dictionary.")
```
schustmi, line 118: Similar comment to Kserve, should be a `TypeError` and more descriptive I think

src/zenml/utils/materializer_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    materializer: the materializer class. This is the path to the materializer class.
+
+    Args:
+        model_artifact: the artifact to extract the metadata from.
+
+    Returns:
+        The path to the temporary file where the model metadata is saved
+    """
+    metadata = dict()
+    metadata["datatype"] = model_artifact.properties["datatype"].string_value
```
schustmi, line 48: These `datatype` and `materializer` string constants are used quite a lot, probably makes sense to have them as a constant somewhere?

src/zenml/integrations/seldon/steps/seldon_deployer.py
```diff
@@ -36,10 +43,49 @@
...
+            predict function path
+
+        Raises:
+            ValueError: if predict function path is not valid
+        """
+        if not predict_func_path:
+            raise ValueError("Predict function path is required.")
+        try:
+            import_class_by_path(predict_func_path)
+        except AttributeError:
```
schustmi, line 84: Similar to kserve, this could also check that the function is a function/callable to catch errors early

src/zenml/integrations/seldon/steps/seldon_deployer.py
```diff
@@ -192,3 +239,151 @@ def prepare_service_config(model_uri: str) -> SeldonDeploymentConfig:
...
+
+
+@step(enable_cache=False)
+def seldon_custom_model_deployer_step(
+    deploy_decision: bool,
+    config: SeldonDeployerStepConfig,
+    context: StepContext,
+    model: ModelArtifact,
+) -> SeldonDeploymentService:
+    """Seldon Core custom model deployer pipeline step.
```
schustmi, line 251: If I recall correctly, this step is actually very similar to the one of Kserve, might be nice if we could abstract the differences away in some functions on the `ModelDeployer` subclasses, but definitely not necessary in this PR

schustmi (changes requested):
Awesome job @safoinme, loved reading this PR!

I have a few minor questions and requests, but overall this looks great already

src/zenml/integrations/kserve/services/kserve_deployment.py
safoinme: sure

src/zenml/integrations/kserve/steps/kserve_deployer.py
safoinme: I think that's a great idea! do you think this is enough @schustmi ?
```
        try:
            predict_function = import_class_by_path(predict_func_path)
        except AttributeError:
            raise ValueError("Predict function can't be found.")
        if not callable(predict_function):
            raise ValueError("Predict function must be callable.")
```

schustmi: Tiny nit: I would use a `TypeError` for the last one, as it is related to the value having the wrong type. Other than that looks good!

src/zenml/integrations/seldon/steps/seldon_deployer.py
safoinme: I will mark it as the first thing to do just after release, there is a lot of functions that got repeated and can benefit of some cleaning

src/zenml/utils/materializer_utils.py
schustmi: Ah no, it was simply a wrong spelling:`materialzer_object` instead of `materializer_object`

AlexejPenner (approved):
Simply awesome! LGTM - the two failing tests seem like sporadic cache issues to me and the two broken links will exists after the upcoming release.

src/zenml/utils/materializer_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    materializer_object: BaseMaterializer = materializer_class(model_artifact)
+    model = materializer_object.handle_input(model_class)
+    try:
+        import torch.nn as nn
+
+        if issubclass(model_class, nn.Module):  # type: ignore
+            model.eval()
+    except ImportError:
+        pass
```
stefannica, line 109: Up until this point, the materializer_utils.py code is generic and reusable for things other than models. I would even have suggested naming these methods `save_artifact_metadata` and `load_artifact_from_metadata`. However, this code is pytorch specific code in an otherwise generic zenml code area. I recommend moving it to where it is being used, namely seldon and kserve, even though it will lead to some minor code duplication.

src/zenml/integrations/kserve/steps/kserve_deployer.py
```diff
@@ -278,3 +317,154 @@ def kserve_model_deployer_step(
...
     return service
+
+
+@step(enable_cache=False)
+def kserve_custom_model_deployer_step(
+    deploy_decision: bool,
+    config: KServeDeployerStepConfig,
+    context: StepContext,
+    model: ModelArtifact,
+) -> KServeDeploymentService:
```
stefannica, line 328: The custom code functionality could just be added to the existing `kserve_model_deployer_step` and have some code executed conditionally based on whether `custom_deploy_parameters` has been set or not. Some code duplication could have been avoided that way.

src/zenml/integrations/kserve/steps/kserve_deployer.py
```diff
@@ -278,3 +317,154 @@ def kserve_model_deployer_step(
...
+        "-m",
+        "zenml.integrations.kserve.custom_deployer.zenml_custom_model",
+        "--model_name",
+        config.service_config.model_name,
+        "--predict_func",
+        config.custom_deploy_parameters.predict_function,
+    ]
+
+    # verify if there is an active stack before starting the service
+    if not context.stack:
```
stefannica, line 409: An stack is guaranteed to be active, because you already called `model_deployer = KServeModelDeployer.get_active_model_deployer()`, but this is probably just for mypy.

src/zenml/integrations/seldon/steps/seldon_deployer.py
```diff
@@ -192,3 +241,151 @@ def prepare_service_config(model_uri: str) -> SeldonDeploymentConfig:
...
+
+    # prepare the service configuration for the deployment
+    service_config = config.service_config.copy()
+    service_config.model_uri = served_model_uri
+
+    # create the specification for the custom deployment
+    service_config.spec = create_seldon_core_custom_spec(
+        model_uri=service_config.model_uri,
+        custom_docker_image=custom_docker_image_name,
+        secret_name=model_deployer.kubernetes_secret_name,
```
stefannica, line 374: This secret contains credentials that allow the seldon predictor containers to authenticate to the remote object storage service (S3, GCS, etc.). What you need here is a secret that contains credentials that allow the kube cluster to authenticate to the container registry.

stefannica (approved):
I have a few improvement suggestions, nothing that would block this PR though. Thanks for all the hard work you put into making custom code serving a reality!

PR #844, opened 2022-08-12
Add Tekton orchestrator

schustmi:
## Describe changes

This PR adds an alpha-version of a potential Tekton orchestrator in ZenML.

**Missing**:
- Docs
- Example
- Tests

## Pre-requisites

## Types of changes


src/zenml/integrations/tekton/orchestrators/__init__.py
```diff
@@ -0,0 +1,18 @@
+#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
htahir1, line 1: 2022

htahir1 (approved):
Looks great!

htahir1:
Added the example and docs

fa9r (approved):
Awesome, looks good!

The only suggestion I'd have is to refactor some of the duplicate code from KFP orchestrator, but I guess we can do that at a later point as part of ENG-982.

Otherwise looks very polished, ready to be merged 👍

htahir1:
@strickvl if you give an approval we can merge!

docs/book/mlops-stacks/orchestrators/tekton.md
```diff
@@ -0,0 +1,139 @@
...
+
+To use the Tekton orchestrator, we need:
+* The ZenML `tekton` integration installed. If you haven't done so, run 
+    ```shell
+    zenml integration install tekton -y
+    ```
+* [Docker](https://www.docker.com) installed and running.
+* [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) installed.
+
+When using the Tekton orchestrator with a remote cluster, you'll additionally need
```
schustmi, line 98: There is no `local` case at the moment, I think we can remove this sentence right?

examples/tekton_pipeline_orchestration/setup.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
schustmi, line 1: I think the `setup.sh` file should only exist for examples that can be run locally, so should be removed in this case

htahir1: Ok! Shouldn't we also remove this from the kubeflow pipelines example too?

docs/book/mlops-stacks/orchestrators/tekton.md
htahir1: Hmm but the local case isn't mentioned here. Do you mean saying the word "additionally"?

examples/tekton_pipeline_orchestration/setup.sh
schustmi: I think the kubeflow example actually runs locally (assuming that you have the pre-requisites installed), but Tekton doesn't support that (for now)

docs/book/mlops-stacks/orchestrators/tekton.md
schustmi: Yeah, I think we could just merge these two lists and remove the intermediate sentence, these are all equivalent requirements to run it.

PR #858, opened 2022-08-24
Add option to specify task dependencies

schustmi:
## Describe changes
This PR adds a way to specify non-data dependencies between steps of a pipeline:
```python
@pipeline
def p(step_1, step_2, step_3):
    step_1.after(step_2)
    step_3(step_1(), step_2())
```

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/developer-guide/advanced-usage/step-execution-order.md
+++ b/docs/book/developer-guide/advanced-usage/step-execution-order.md
@@ -1,0 +1,37 @@
+---
+description: How to control the order in which steps get executed
+---
+
+By default, ZenML uses the data flowing between steps of your pipeline to 
+determine the order in which steps get executed. The following example
+shows a pipeline in which `step_3` depends on the outputs of `step_1`
+and `step_2`. This means that ZenML can execute both `step_1` and `step_2` in
+parallel but needs to wait until both are finished before `step_3` can be
+started.
+
+```python
+from zenml.pipelines import pipeline
+
+@pipeline
+def example_pipeline(step_1, step_2, step_3):
+    step_1_output = step_1()
+    step_2_output = step_2()
+    step_3(step_1_output, step_2_output)
+```
+
+If you have additional constraints on the order in which steps get executed,
+you can specify non-data dependencies by calling `step.after(some_other_step)`:
+
+```python
+from zenml.pipelines import pipeline
+
+@pipeline
+def example_pipeline(step_1, step_2, step_3):
+    step_1.after(step_2)
+    step_1_output = step_1()
+    step_2_output = step_2()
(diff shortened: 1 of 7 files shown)


src/zenml/orchestrators/utils.py
```diff
@@ -54,7 +69,7 @@ def create_tfx_pipeline(
     # step is executed (see `BaseOrchestrator.run_step(...)`)
     return tfx_pipeline.Pipeline(
         pipeline_name=zenml_pipeline.name,
-        components=tfx_components,  # type: ignore[arg-type]
+        components=list(tfx_components.values()),  # type: ignore[arg-type]
```
htahir1, line 72: Why was this neccessary? :-)

schustmi: Always needed a list, I just changed it above to a `Dict` so I could more easily fetch the components to add the manual upstream steps.

AlexejPenner (approved):
Didn't imagine this to be so straightforward - looks good to me

docs/book/developer-guide/advanced-usage/step-execution-order.md
```diff
@@ -0,0 +1,37 @@
...
+from zenml.pipelines import pipeline
+
+@pipeline
+def example_pipeline(step_1, step_2, step_3):
+    step_1_output = step_1()
+    step_2_output = step_2()
+    step_3(step_1_output, step_2_output)
+```
+
+If you have additional constraints on the order in which steps get executed,
```
htahir1, line 22: I feel like this should be also in other parts of the doc. Maybe somewhere more visibile than the guide? API Docs perhaps? Im not sure

htahir1 (approved):
One comment otherwise LGTM pending that we test this on most orchestrators

htahir1:
Accidentally approved, was more of a Comment

docs/book/developer-guide/advanced-usage/step-execution-order.md
schustmi: Yeah I wasn't sure where to put it tbh. It seems too advanced for the developer guide, which is why I put it here.
The feature is documented in the API docs as that just uses the docstring of the `.after(...)` method right? I wouldn't know where to put more generic feature descriptions in the API docs (and also don't think they're the right place), but happy to add it on other docs pages if there are any suggestions

htahir1: hmm maybe make it more prominent with an example?

schustmi: Added a simple example in the docstring

htahir1 (approved):
LGTM!

PR #864, opened 2022-08-26
Split docstring check into a separate action that only runs once

fa9r:
## Describe changes
I made two changes to our GitHub Actions:
- Docstring checking was removed from linting is now in a separate job that is only run on once machine.
- Markdown checking now runs in quiet mode (only prints links that are broken) and its job can no longer fail so it does not affect the overall CI result

## Pre-requisites

## Types of changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,14 +135,13 @@ Warning: This might take a while for both (~ 15 mins each, depending on your mac
 time, please run it as it will make the
 next commands error-free.
 
-It is easy to run the following scripts via Poetry to ensure code formatting and
-linting is in order:
+You can now run the following scripts via Poetry to automatically format your
+code and to check whether the code formatting, linting, docstrings, and
+spelling is in order:
 
 ```
 poetry run bash scripts/format.sh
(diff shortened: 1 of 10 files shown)


strickvl (approved):
Let's go to main!

CONTRIBUTING.md
```diff
@@ -141,6 +141,7 @@ linting is in order:
 ```
 poetry run bash scripts/format.sh
 poetry run bash scripts/lint.sh
+poetry run bash scripts/docstring.sh
 poetry run bash scripts/check-spelling.sh
```
strickvl, line 145: wonder whether this a *lot* for people to do & if we should consider consolidating some of these into a single command for the sake of external contributors, albeit keeping the individual scripts / commands for ourselves...)

strickvl (approved):

CONTRIBUTING.md
fa9r: Great idea! I added a `run-ci-checks.sh` script that combines linting, docstring checks, and spell checks and adjusted the contributing guide accordingly.

PR #865, opened 2022-08-26
Remove profiles and local zenml store

stefannica:
## Describe changes
Remove the profile concept:

* remove profile related CLI commands and configurations
* a single global zenml store can be configured in the global configuration via a new `zenml config set` CLI command
* the zen store concept is also hidden in the CLI as the user can only configure either a local database or connect to a remote Zen server via the new `zenml config set` CLI command.
* remove the local zenml store implementation
* replace the `zenml profile` CLI command with `zenml config` to manage the global store configuration
* implement new `zenml profile list` and `zenml profile import` CLI commands only to be used as a migration path from legacy profiles to the new global configuration approach. These commands still retain some parts of the local zen store implementation that are required for backwards compatibility. We can remove all this code at a later time, after users have had enough time to migrate their local profiles.
* moved all concerns related to Zen stores from Repository into the BaseZenStore class

## Pre-requisites

## Types of changes


docs/book/collaborate/zenml-store.md
```diff
@@ -1,240 +0,0 @@
----
-description: How to manage stacks in a centralized location with ZenStores
```
fa9r, line 2: Since ZenStore is still a concept, so I don't think this should be totally removed.

src/zenml/cli/user_management.py
```diff
@@ -32,40 +32,24 @@ def user() -> None:
 @user.command("get")
 def get_user() -> None:
     """Get the active user."""
-    cli_utils.print_active_profile()
-    cli_utils.declare(f"Active user: '{Repository().active_user_name}'")
-
-
-@user.command("set", help="Set the active user.")
-@click.argument("user_name", type=str)
```
fa9r, line 40: Why was this removed? I guess we should still support switching users locally, or how would that work now?

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -381,20 +381,20 @@ def build_and_push_docker_image(
...
             )
             build_context_root = (
                 source_utils.get_source_root_path()
                 if requires_build_context
                 else None
             )
-            maybe_include_active_profile = (
-                _include_active_profile(build_context_root=build_context_root)  # type: ignore[arg-type]
-                if docker_configuration.copy_profile
+            maybe_include_global_config = (
```
fa9r, line 391: what is `maybe_include_global_config`? :D

fa9r (commented):
I only checked it very superficially. Personally, there are too many big changes for me to review it thoroughly, I don't even know where to start 😅

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -63,38 +64,45 @@
...
-            ValueError: if the url is not a valid URL.
-        """
-        if not self.is_valid_url(url.strip("/")):
-            raise ValueError("Invalid URL for REST store: {url}")
-        self._url = url.strip("/")
-        super().initialize(url, *args, **kwargs)
-        return self
+        # Validate attributes when assigning them. We need to set this in order
+        # to have a mix of mutable and immutable attributes
+        validate_assignment = True
```
schustmi, line 84: Any reason why extras should be ignored here? And the validate assignment is already set on the superclass and can probably be removed?

src/zenml/config/store_config.py
```diff
@@ -0,0 +1,50 @@
...
+
+    class Config:
+        """Pydantic configuration class."""
+
+        # Validate attributes when assigning them. We need to set this in order
+        # to have a mix of mutable and immutable attributes
+        validate_assignment = True
+        # Allow extra attributes to be set in the base class. The concrete
+        # classes are responsible for validating the attributes.
+        extra = "allow"
```
schustmi, line 47: Any specific reason to have this here? I'm guessing at some point the config get instantiated as a `StoreConfiguration` instead of the correct subclass?

schustmi (changes requested):
So much code gone, love it! I couldn't find anything major, just have two questions related to pydantic

src/zenml/config/store_config.py
stefannica: Ah, yes, you're correct. This allows me to reference `StoreConfiguration` in `GlobalConfiguration`, and I can also assign a `StoreConfiguration` object to a `SqlZenStoreConfiguration` field and let the `SqlZenStoreConfiguration` class deal with the validation.

src/zenml/zen_stores/rest_zen_store.py
stefannica: I'm ignoring extras in all configuration related models persisted to disk as a general strategy to ensure backwards compatibility with previous ZenML versions.

src/zenml/utils/pipeline_docker_image_builder.py
stefannica: this is an option part of the docker image build configuration that copies the global configuration, including the local zen store contents, into the docker image. It replaces the previous option that was used to copy the active profile along with the active stack and its components into the docker image.

src/zenml/zen_stores/rest_zen_store.py
schustmi: Mhm I see, I think for `StackComponent`s we recently did the opposite, we might want to unify that. I think in that case it was to prevent typos on the CLI though that parses arbitrary arguments, I'm guessing that's not a problem here right, or could something like this happen?
```bash
zenml store register --passwor=...
```

src/zenml/config/store_config.py
schustmi: I ran into the exact same issue in the thing I'm currently working on. I found a different approach that makes sense in my situation, I'll just leave it here for you to decide if it's worth in this case as well:
```python
class StoreConfig(BaseModel):
    class Config:
        extra = "forbid"

class Container(BaseModel):
    store: StoreConfig
    
    @validator("config", pre=True)
    def _convert_config(cls, config: Any) -> StoreConfig:
        if isinstance(config, Dict):
            # this happens when it gets read from disk
            store_type = config["type"]
            ...  # create the right object here
```

src/zenml/zen_stores/rest_zen_store.py
stefannica: No, the CLI is highly regulated concerning what users can configure for the stores. In fact, the local SQLite store is completely hidden through `zenml config set --local-store` and accepts no user parameters. Creating a REST store through the CLI is also tightly regulated through the use of specified `--url`, `--username` and  `--password` arguments.

src/zenml/config/store_config.py
stefannica: That's exactly what I was using, a validator to convert the base `StoreConfig` into the proper subclass, but what I found is that it's not even needed, because Pydantic does the conversion automatically. To expand on your example: 

```python

class StoreConfig(BaseModel):
    name: str

    class Config:
        extra = "allow"
        validate_assignment = True

class Container(BaseModel):
    store: StoreConfig

    class Config:
        extra = "forbid"
        validate_assignment = True

class CustomStoreConfig(StoreConfig):
    custom_attr1: str
    custom_attr2: int

    class Config:
        extra = "forbid"

class CustomContainer(Container):
    store: CustomStoreConfig

store = StoreConfig(name="test", custom_attr1="value", custom_attr2="1")
cont = CustomContainer(store=store)
```

schustmi: Ah I see, in my case the `CustomContainer` class doesn't exist and instead the `Container` can hold potentially different "stores", which is why the automatic conversion doesn't work

fa9r:
Some feedback on the user flow (detailed code review will follow later):
- `zenml config set --local-store` should be `zenml disconnect` IMO
- other cases of `zenml config set ...` should be `zenml connect ...` instead IMO
- `zenml profile ...` CLI commands should show a warning every time that profile are deprecated and explain how profiles can be migrated now
- `zenml profile import` should be renamed to `zenml profile migrate`
- when no args are provided to `zenml profile import`, I would expect it to just import everything (components, flavors, ...)
- for `zenml profile import`, instead of the `--prefix` arg I would prefer a `--project` arg that you can use to either import the profile into an existing project or to automatically create a new project that includes all stacks from the old profile; for this change, it probably makes more sense to implement it after rebasing onto the mega branch since we already have some implementation on projects (incl. default projects) there

stefannica:
@fa9r I implemented some of your suggestions (see below).

> Some feedback on the user flow (detailed code review will follow later):
> 
> * `zenml config set --local-store` should be `zenml disconnect` IMO
> * other cases of `zenml config set ...` should be `zenml connect ...` instead IMO

`zenml connect` and `zenml disconnect` are indeed better, but when I make this change, I would like to do it in the context of adding ZenML server deployments to the CLI, because they are tightly related to `zenml server deploy` and `zenml server connect|disconnect`.

> * `zenml profile ...` CLI commands should show a warning every time that profile are deprecated and explain how profiles can be migrated now

Done, for the `zenml profile list` command only. It doesn't make sense to do it for `zenml profile migrate`.

> * `zenml profile import` should be renamed to `zenml profile migrate`

Done.

> * when no args are provided to `zenml profile import`, I would expect it to just import everything (components, flavors, ...)

Done.

> * for `zenml profile import`, instead of the `--prefix` arg I would prefer a `--project` arg that you can use to either import the profile into an existing project or to automatically create a new project that includes all stacks from the old profile; for this change, it probably makes more sense to implement it after rebasing onto the mega branch since we already have some implementation on projects (incl. default projects) there

Great suggestion. I will replace `--prefix` with `--project` after rebase if project scoping for stacks/components/flavors works as expected.

stefannica:
Manually rebased and merged into the `feature/ENG-1103-merging-metadata-stores-and-zen-server` branch.

PR #867, opened 2022-08-29
Update cheat sheet with latest CLI commands from 0.13.0

dnth:
## Describe changes
This PR updates the cheat sheet with latest CLI command breaking changes from 0.13.0.

## Pre-requisites

## Types of changes


strickvl:
The big thing I see needing changing is the `zenml served-models...` command which no longer exists. It is now `zenml model-deployer models ...` for equivalent functionality. Also the following annotation commands:

```shell
zenml annotator up — Spin up the annotation server in the background

zenml annotator dataset list — List the datasets that are available to use in your annotation interface
zenml annotator dataset stats [DATASET_NAME] — Learn how much of your dataset you have labelled and how much remains
zenml annotator dataset annotate [DATASET_NAME] — Launch the annotation interface for a particular dataset
```

strickvl:
Also might be worth waiting for Stefan's branch #865 to be merged in since that'll remove a section + add some new commands...

strickvl:
Ok then keep the `profile` commands until that happens... That makes sense.

fa9r:
Another thing to keep in mind for the next version: we also have a whole bunch of CLI commands related to user/permission management (project, user, team, role). Currently it makes sense that these are not on the cheat sheet since they are basically useless, but we might want to include some of them once the ZenServer is out.

dnth:
@strickvl added annotation CLI :)

strickvl:
@dnth wondering whether the headers for each command should be human-comprehensible headlines and not just the first part of the command? i.e. "Annotation" instead of `zenml annotator` etc? and 'Deployed Models' instead of `zenml model-deployer models`?

strickvl:
@dnth looks great to me!

dnth:
@strickvl Done! :fire:

strickvl (approved):

PR #868, opened 2022-08-29
Add source to segment identify call

htahir1:
## Describe changes
I did this because in the future we might have identify using different sources (like ZenServer), so we want to be aware of it in the telemetry

## Pre-requisites
- [x] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/mlops-stacks/integrations) table and the [corresponding website section](https://zenml.io/integrations).

## Types of changes

### DIFF
--- a/src/zenml/enums.py
+++ b/src/zenml/enums.py
@@ -117,6 +117,6 @@ class SecretValidationLevel(StrEnum):
 
 
 class AnalyticsEventSource(StrEnum):
-    """Analytics identify sources."""
+    """Enum to identify analytics events source."""
 
     ZENML_GO = "zenml go"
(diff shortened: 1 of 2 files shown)


strickvl (approved):

src/zenml/enums.py
```diff
@@ -114,3 +114,9 @@ class SecretValidationLevel(StrEnum):
     SECRET_AND_KEY_EXISTS = "SECRET_AND_KEY_EXISTS"
     SECRET_EXISTS = "SECRET_EXISTS"
     NONE = "NONE"
+
+
+class IdentifySource(StrEnum):
```
schustmi, line 119: IMO the name could be a little more expressive here, something with analytics in there? `AnalyticsEmailSource`/`AnalyticsEventSource`?

fa9r: Docstring too ;)

fa9r (approved):

htahir1:
@fa9r @schustmi took your suggestions in thanks!

PR #880, opened 2022-09-07
Unit tests for (some) integrations

strickvl:
Added some unit tests for integrations. There's probably more that could be added, but these were the low-hanging fruit. There is one final change that will be made once this approved which will be splitting the unit tests into `unit-core` and `unit-integration` directories, following discussion a while back with @AlexejPenner.

@stefannica this probably interfaces with the work on the testing, so if you'd rather I just leave things as they are, then I'm happy to do that as well.

## Pre-requisites

## Types of changes


tests/unit/integrations/great_expectations/materializers/test_ge_materializer.py
```diff
@@ -0,0 +1,47 @@
...
+    with does_not_raise():
+        test_pipeline(
+            read_suite=read_suite().with_return_materializers(
+                GreatExpectationsMaterializer
+            )
+        ).run()
+
+    last_run = clean_repo.get_pipeline("test_pipeline").runs[-1]
+    expectation_suite = last_run.steps[-1].output.read()
+    assert isinstance(expectation_suite, ExpectationSuite)
```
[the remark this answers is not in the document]
schustmi: This is actually True for most of the materializer tests below, they could all in addition to doing instance checks perform some additional validation that makes sure the object is correctly reconstructed and no information is lost during writing/reading.

tests/unit/integrations/xgboost/materializers/test_xgboost_booster_materializer.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_xgboost_booster_materializer():
+    """Tests whether the steps work for the xgboost booster materializer."""
+
+    @step
+    def some_step() -> xgb.Booster:
+        return xgb.Booster(params={"objective": "reg:linear"})
+
+    with does_not_raise():
+        some_step().with_return_materializers(XgboostBoosterMaterializer)()
```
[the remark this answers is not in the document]
schustmi: This is also the case for some other tests in this PR, i'll just leave it here as a general note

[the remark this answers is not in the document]
strickvl: Ah yes. This must have been from when I first started the PR / ticket weeks ago. I think I took this from some tests that already exist, but in any case it does nothing. I'll amend to the pipeline format as you suggest.

tests/unit/integrations/great_expectations/materializers/test_ge_materializer.py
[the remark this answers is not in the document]
strickvl: Yep I'll add some checks, as appropriate.

tests/unit/integrations/huggingface/materializers/test_huggingface_tf_model_materializer.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_pipeline(read_model) -> None:
+        """Tests the Huggingface Tensorflow Pretrained Model materializer."""
+        read_model()
+
+    with does_not_raise():
+        test_pipeline(
+            read_model=read_model().with_return_materializers(
+                HFTFModelMaterializer
+            )
+        ).run()
```
fa9r, line 45: I see this pattern a lot: 1) define a step with no inputs that returns a specific data type 2) create pipeline with only this step in it 3) make sure the pipeline runs without error. Perhaps this could be generalized.

tests/unit/integrations/neural_prophet/materializers/test_neural_prophet_materializer.py
```diff
@@ -0,0 +1,59 @@
...
+    def read_forecaster() -> NeuralProphet:
+        """Reads and materializes a Neural Prophet forecaster."""
+        sample_df = pd.DataFrame(
+            {
+                "ds": pd.date_range("2018-01-01", periods=10),
+                "y": np.random.rand(10),
+            }
+        )
+        m = NeuralProphet(epochs=7, batch_size=37)
+        m.fit(sample_df)
```
fa9r, line 40: how long does this run?

fa9r: if it's not instant I'd reduce epochs to either 1 or 0

fa9r (approved):
Great PR, this is much needed!

The overall pattern for testing materializers seems to be almost always the same:
1. Define a step with no inputs that returns <DATA_TYPE>
2. Define a pipeline with only one step
3. Run the pipeline `with_return_materializers(<MATERIALIZER>)` and check that there's no error

Perhaps this could be generalized into a `_test_materializer()` util function that you can just call with different data types and materializer classes?

tests/unit/integrations/huggingface/materializers/test_huggingface_tf_model_materializer.py
strickvl: That's the basic version of it, but then there's the extended version which @schustmi encouraged me to do: i.e. actually checking that the artifacts are really functional (via the post-execution checks / calls). Agreed that probably this could all be simplified somehow with an extra function. I'll try to get this in.

tests/unit/integrations/neural_prophet/materializers/test_neural_prophet_materializer.py
strickvl: I'll reduce it, though it takes basically no time at all to run. It can't be 0, I think, based on the errors I was getting last week. I've reduced it to 2 now.

tests/unit/integrations/evidently/materializers/test_evidently_profile_materializer.py
```diff
@@ -17,27 +17,21 @@
...
-def materialize_profile() -> Profile:
-    """Materialize the profile."""
-    return Profile(sections=[DataDriftProfileSection()])
-
-
-@pipeline
-def evidently_pipeline(materializer) -> None:
-    """Test pipeline for the Evidently profile materializer."""
-    materializer()
+from ....test_general import _test_materializer
```
fa9r, line 24: Wouldn't this import work as well: `from tests/unit/test_general import _test_materializer`? If so, I would prefer that over `from ....`

fa9r (approved):
Awesome, really clean now 👍

tests/unit/integrations/evidently/materializers/test_evidently_profile_materializer.py
strickvl: I wasn't able to get that to work, but maybe I'll try again later.

strickvl: Figured out a way to get this to work 🥳. Still doesn't completely make sense to me why it works that way, but at least it beats the old way I was doing it.

schustmi (approved):
This looks really clean now, nice job!

PR #882, opened 2022-09-07
Mlflow Databricks connection

gabrielmbmb:
## Describe changes
I've updated the `MLFlowExperimentTracker` so it can be connected to an [MLflow tracking server managed by Databricks](https://www.databricks.com/product/managed-mlflow). To do so, I followed the [instructions](https://docs.databricks.com/applications/mlflow/access-hosted-tracking-server.html) provided by Databricks.

In addition, I have added some unit tests for the MLflow experiment tracker, since there were none. I manually tested the changes using the [MLflow Tracking example](https://github.com/zenml-io/zenml/tree/main/examples/mlflow_tracking) and it worked fine!

image

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/mlops-stacks/experiment-trackers/mlflow.md
+++ b/docs/book/mlops-stacks/experiment-trackers/mlflow.md
@@ -66,12 +66,20 @@ what other types of components it can be combined with. This option requires
 [authentication related parameters](#authentication-methods) to be configured
 for the MLflow Experiment Tracker.
 
+* [Databricks scenario](https://www.databricks.com/product/managed-mlflow):
+This scenario assumes that you have a Databricks workspace and you want to
+use the managed MLflow Tracking server it provides. This option requires
+[authentication related parameters](#authentication-methods) to be configured
+for the MLflow Experiment Tracker.
+
 ### Authentication Methods
 
 You need to configure the following credentials for authentication to a remote
 MLflow tracking server:
 
-* `tracking_uri`: The URL pointing to the MLflow tracking server.
+* `tracking_uri`: The URL pointing to the MLflow tracking server. If using
+an MLflow Tracking Server managed by Databricks, then the value of this
+attribute should be `"databricks"`.
 * `tracking_username`: Username for authenticating with the MLflow tracking
 server. 
 * `tracking_password`: Password for authenticating with the MLflow tracking
--- a/docs/book/mlops-stacks/experiment-trackers/mlflow.md
+++ b/docs/book/mlops-stacks/experiment-trackers/mlflow.md
@@ -80,6 +80,10 @@ server.
 Token for authenticating with the MLflow tracking server.
 * `tracking_insecure_tls` (optional): Set to skip verifying the MLflow tracking server SSL
 certificate.
+* `databricks_host`: The host of the Databricks workspace with the MLflow managed
+server to connect to. This is only required if `tracking_uri` value is set to
+`"databricks"`. More information:
+[Access the MLflow tracking server from outside Databricks](https://docs.databricks.com/applications/mlflow/access-hosted-tracking-server.html)
     
 Either `tracking_token` or `tracking_username` and `tracking_password` must be
 specified.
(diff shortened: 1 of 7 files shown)


htahir1:
Wow great contribution! I do have some comments off the bat:

* Why do we need to add the databricks logic inside the standard mlflow tracker? Cant we make another one with a seperate integration?
* On the other hand, I do not really see the difference between the DATABRICKS_HOST and TRACKING_URI params (and the other three). They are all analagous i suppose? Why cant we just re-use the params already set?

Happy to hear your thoughts and again thank you for the awesome contrib!

gabrielmbmb:
@htahir1 hey Hamza, so Databricks MLflow Tracking Server is exactly the same as the vanilla MLflow Tracking Server, but the only thing that changes is the way to connect to it. If we want to connect to it, we need to explicitly set the value of the `MLFLOW_TRACKING_URI` to `"databricks"` or `mlflow.set_tracking_uri("databricks")`, this way `mlflow` Python package will know how to handle the connection to the MLflow Tracking Server managed by Databricks. Having that said, if we set the tracking uri to `"databricks"`, then MLflow will need to know the Databricks authentication either provided by environment variables (`DATABRICKS_USERNAME`, `DATABRICKS_PASSWORD` or `DATABRICKS_TOKEN`) or having a `.databrickscfg file containing the authentication in the root of your user home. Additionally, we need to let MLflow to which Databricks to connect, so we need to set the `DATABRICKS_HOST` env variable, and therefore that's why I added the `databricks_host` attribute.

htahir1:
@gabrielmbmb Understood...

 I would be curious of the thoughts of @schustmi and @AlexejPenner here. Guys should this be a seperate component?

htahir1:
@gabrielmbmb We had a discussion internally and we decided we'll merge for now but we might need to have a seperate databricks_mlflow experiment tracking flavor in the future once we have more databricks integrations!

Thank you again for the contribution! 🚀 🚀

docs/book/mlops-stacks/experiment-trackers/mlflow.md
```diff
@@ -80,6 +80,10 @@ server.
 Token for authenticating with the MLflow tracking server.
 * `tracking_insecure_tls` (optional): Set to skip verifying the MLflow tracking server SSL
 certificate.
+* `databricks_host`: The host of the Databricks workspace with the MLflow managed
```
schustmi, line 83: Does it make sense to add a short mention of the `databricks` constant in the `tracking_uri` bullet point above?

src/zenml/integrations/mlflow/experiment_trackers/mlflow_experiment_tracker.py
```diff
@@ -74,13 +79,17 @@ def my_step(context: StepContext, ...)
...
+            managed server to connect to. This is only required if
+            `tracking_uri` value is set to `"databricks"`.
     """
 
     tracking_uri: Optional[str] = None
     tracking_username: Optional[str] = SecretField()
     tracking_password: Optional[str] = SecretField()
     tracking_token: Optional[str] = SecretField()
     tracking_insecure_tls: bool = False
+    databricks_host: Optional[str] = SecretField()
```
schustmi, line 92: The secret field right now warns users that they should specify the value as a secret, and in the future might even raise an exception if they don't. I'm not familiar with Databricks, but from a quick search the URL doesn't necessarily contain any sensitive information right? In that case I'd suggest making this just a normal `Optional[str]` field similar to the `tracking_uri` (it can still be specified as a secret, just isn't mandatory anymore).

docs/book/mlops-stacks/experiment-trackers/mlflow.md
schustmi: Along the same lines, I think a short mention in the `how to deploy it` section that mentions databricks would probably make this easier to find?

schustmi (changes requested):
Awesome PR @gabrielmbmb, thanks! I have a few minor suggestions, but overall the code looks really good

src/zenml/integrations/mlflow/experiment_trackers/mlflow_experiment_tracker.py
gabrielmbmb: You're right, this one doesn't need to be a secret

schustmi (approved):
Looks good to me!

schustmi:
@gabrielmbmb Anything left from your side or is this good to go?

gabrielmbmb:
@schustmi nothing left!

PR #888, opened 2022-09-12
Remove explicit nbconvert and markupsafe dependency

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,12 +65,11 @@ rich = {extras = ["jupyter"], version = "^12.0.0"}
 httplib2 = "<0.20,>=0.19.1"
 pyparsing = "<3,>=2.4.0"
 sqlmodel = "~0.0.8"
-markupsafe = "1.1.1"
-nbconvert = "6.4.4"
 passlib = { extras = ["bcrypt"], version = "~1.7.4"}
 python-terraform = { version = "^0.10.1" }
 pymysql = { version = "~1.0.2"}
 alembic = { version = "~1.8.1"}
+
 
 # Optional dependencies for the ZenServer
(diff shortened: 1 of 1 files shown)


htahir1 (commented):
Wow markupsafe was so annoying, and im very happy to get rid of it. However, one second thought I am unsure why have nbconvert as a dependency in the first place?

htahir1 (commented):
is this a deptree thing? What about getting rid of pyarrow?

schustmi:
Yes deptree is a actually a really good idea, we'd just need to make sure we install all the integrations before running it (as from what I've seen this update here clashes not with the zenml installation, but only once we try to install mlflow).

Regarding pyarrow: 
This will break backwards compatability with previously saved artifacts, should we do it anyway?

schustmi:
My findings on why the integration tests are failing for this PR:

On Windows: we're trying to install `markupsafe` in the virtual environment while it is currently imported.

On Linux: it's a combination of the integrations `airflow`, `neural_prophet` and `mlflow`
- The default installation of ZenML leaves us with `Jinja2==3.1.2` and `MarkupSafe==2.1.1`
- Installing airflow on top of that installs (among other things) `Jinja2==2.11.3`, `MarkupSafe==1.1.1` and `Flask==1.1.4`
- Installing neural prophet after that: `Jinja2==3.1.2`, `MarkupSafe==2.1.1` and`Flask==1.1.4`  ⚠️ This is now broken as the installed flask version requires older versions of jinja and markupsafe
- Installing mlflow after that: Downgrades `Jinja2==2.11.3` but the markupsafe version still doesn't match and leads to the import error in the tests

schustmi:
Solved by #1042

PR #891, opened 2022-09-14
Runtime Lineage Graph

htahir1:
## Describe changes
I implemented a lineage graph based on pydantic and the old PipelineRunView. I do not have the capability at this moment to merge this into the new way of doing things. Perhaps @AlexejPenner  or @fa9r can help?


fa9r:
@htahir1 cool PR, I'll make it work with the new views 👍

fa9r (approved):

PR #895, opened 2022-09-16
component flavor models

fa9r:
## Describe changes

We split all stack component flavors into three classes:
- An implementation class, which subclasses `StackComponent` and defines the logic of stack component.
- A config class, which subclasses the new `StackComponentConfiguration` class, defines the attributes of the flavor, and, optionally, performs input validations.
- A flavor class, which subclasses the new `Flavor` class, replaces the old `TYPE` and `FLAVOR` attributes, and links to the respective implementation and config classes via its properties.

By lazy-loading the implementation class, this change allows us to register stack components without having to load the implementation logic (and, therefore, without having to have any of the required integrations installed).

## Pre-requisites

## Types of changes


src/zenml/stack/stack_component.py
```diff
@@ -109,6 +95,71 @@ def __init__(self, **kwargs: Any) -> None:
...
+            user=self.user,
+            project=self.project,
+            id=self.id,
+            type=self.type,
+            flavor_name=self.flavor,
+            name=self.name,
+            configuration=self.config.dict(),
+        )
+
     def __custom_getattribute__(self, key: str) -> Any:
```
schustmi, line 163: This `__custom_getattribute__` (and setting it to `__getattribute__` below needs to be on the config class so that secret references work

src/zenml/stack/stack_component.py
```diff
@@ -33,21 +33,7 @@
...
-    """
-
-    name: str
-    uuid: UUID = Field(default_factory=uuid4)
-
-    # Class Configuration
-    TYPE: ClassVar[StackComponentType]
-    FLAVOR: ClassVar[str]
-
+class StackComponentConfig(BaseModel, ABC):
```
schustmi, line 36: I haven't looked into this PR in detail, but I'd be interested how you solved that mypy correctly recognizes the type of `self.config` in all the various stack component subclasses?

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -80,16 +80,11 @@ def _include_global_config(
         fileio.rmtree(config_path)
 
 
-class PipelineDockerImageBuilder(BaseModel):
-    """Builds Docker images to run a ZenML pipeline.
+class PipelineDockerImageBuilderConfigMixin(BaseModel):
```
schustmi, line 83: I suggest leaving the `PipelineDockerImageBuilder` class untouched and without a separate config class. The only reason that stack components subclass this is the `docker_parent_image` attribute which was kept for backwards compatability. In this breaking release instead I suggest removing this attribute entirely and have users use the docker configuration instead.

src/zenml/stack/stack_component.py
fa9r: We haven't yet. Any suggestions how to do that cleanly?

src/zenml/utils/pipeline_docker_image_builder.py
fa9r: Makes sense. How should we fix this for now? I assume the orchestrators/step operators still need to subclass this class, so we simply remove the `docker_parent_image` attribute and all logic associated to it?

src/zenml/models/component_model.py
```diff
@@ -53,10 +49,10 @@ class ComponentModel(AnalyticsTrackedModelMixin):
         "is_shared",
     ]
 
-    id: Optional[UUID] = Field(
-        default=None,
-        title="The id of the Stack Component.",
+    id: UUID = Field(
```
AlexejPenner, line 52: Check out the Updates to the StackModel, specifically how it inherits from ShareableProjectScopedDomainModel - this might make handling our models easier if you also apply it here

src/zenml/models/flavor_models.py
```diff
@@ -13,103 +13,57 @@
...
-                )
+    ANALYTICS_FIELDS: ClassVar[List[str]] = [
+        "id",
+        "type",
+        "integration",
+        "project_id",
+        "user_id",
+    ]
+
+    id: Optional[UUID] = Field(
```
AlexejPenner, line 40: Same as the ComponentModel it might make sense here to also inherit from ShareableProjectScopedDomainModel

src/zenml/repository.py
```diff
@@ -692,53 +786,133 @@ def deregister_stack(self, stack: "StackModel") -> None:
...
+            return self.zen_store.list_stack_components(
+                project_name_or_id=self.active_project.id,
+                user_name_or_id=self.active_user.id,
+                type=type,
+            )
+
+    def get_stack_component_by_name_and_type(
+        self, type: StackComponentType, name: str, is_shared: bool = False
+    ) -> ComponentModel:
+        """Fetches a stack by name within the active stack"""
```
AlexejPenner, line 914:
```suggestion
        """Fetches a stack component by name and type"""
```

src/zenml/stack/stack.py
```diff
@@ -121,6 +121,21 @@ def __init__(
...
+        """Creates a StackModel from an actual Stack instance.
+
+        Returns:
+            a StackModel
+        """
+        return StackModel(
+            id=self.id,
+            name=self.name,
+            components={
+                type_: component.to_model()
```
AlexejPenner, line 134: shouldn't this be `type_: component.id`

src/zenml/stack/stack_component.py
```diff
@@ -109,6 +95,71 @@ def __init__(self, **kwargs: Any) -> None:
...
+
+    def to_model(self) -> "ComponentModel":
+        return ComponentModel(
+            user=self.user,
+            project=self.project,
+            id=self.id,
+            type=self.type,
+            flavor_name=self.flavor,
+            name=self.name,
+            configuration=self.config.dict(),
```
AlexejPenner, line 160: I'm not really sure here, but the default factory of the created field of the ComponentModel will be triggered here and a false creation date for the component will be generated. We have to make sure this isn't falsely stored or reported to the user. Alternativley it could be queried from the zen_store using the id

src/zenml/zen_server/models/component_models.py
```diff
@@ -0,0 +1,86 @@
...
+from typing import Any, Dict, Optional
+from uuid import UUID
+
+from pydantic import BaseModel, Field
+
+from zenml.enums import StackComponentType
+from zenml.models import ComponentModel
+
+
+class CreateComponentModel(BaseModel):
```
AlexejPenner, line 24: Have  alook at how the other models have been updated - subclassing from ProjectScopedCreateRequest - we can hopwever also change this after the merge

AlexejPenner (commented):
I got some notes, but i don't really care if they are fixed on the separate branch or merged and then addressed - however we should then track these as todos

src/zenml/stack/stack_component.py
fa9r: Good catch. Done.

PR #904, opened 2022-09-27
Warn when registering non-local component with local ZenServer

strickvl:
- Added stack component configuration properties to determine whether a component is local (uses local resources) or is remote (running in a location different than the local host)
- I use this property inside the `client` when registering and updating stacks and stack components to warn in the following cases:
  a) a user is trying to register a remote component while using a local database or connected to a local ZenML Server
  b) a user is trying to register a local component while connected to a remote ZenML Server
  c) a user is trying to register a stack that combines local and remote components

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -27,18 +27,20 @@ from zenml.constants import (
     REPOSITORY_DIRECTORY_NAME,
     handle_bool_env_var,
 )
-from zenml.enums import StackComponentType
+from zenml.enums import StackComponentType, StoreType
 from zenml.environment import Environment
 from zenml.exceptions import (
     AlreadyExistsException,
     IllegalOperationError,
     InitializationException,
+    ValidationError,
 )
 from zenml.io import fileio
 from zenml.logger import get_logger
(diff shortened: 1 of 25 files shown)


src/zenml/client.py
```diff
@@ -774,6 +775,13 @@ def register_stack_component(
 
         # Update the configuration in the model
         component.configuration = configuration
+        if (
+            self.zen_store.is_local_store()
+            and not StackComponent.from_model(component).is_local
+        ):
+            logger.warn(
+                "You are registering a non-local stack component while using a local ZenServer. This may lead to unexpected behavior."
```
AlexejPenner, line 783: Shouldn't we be more forcefull? A remote orchestrator with a local zensotre will certainly fail, so we should maybe even strongly discourage this and explain why this might not work

stefannica: I updated the message to convey more information. I would still not make this an error, as history showed people have all sorts of deployment use-cases, some of which are on-prem or even on the same machine, but we have no easy way of detecting those.

PR #905, opened 2022-09-28
Pretty flavor describe command for the CLI

bcdurak:
## Describe changes
I have implemented a CLI command to describe different flavors based on their configuration.

Example:
```
zenml model-deployer flavor describe mlflow

Configuration class: MLFlowModelDeployerConfig

Configuration for the MLflow model deployer.

Attributes:
    service_path: the path where the local MLflow deployment service
        configuration, PID and log files are stored.

Properties
service_path, string
```

## Pre-requisites

## Types of changes


src/zenml/cli/utils.py
```diff
@@ -765,6 +768,40 @@ def print_server_deployment(server: "ServerDeployment") -> None:
     console.print(rich_table)
 
 
+def describe_pydantic_object(schema_json: str):
+    """Describes a pydantic object based on the json of its schema.
+
+    Attributes:
+        schema_json: str, represents the schema of a pydantic object, which
+            can be obtained through BaseModelCLass.schema_json()
```
strickvl, line 776:
```suggestion
    """Describes a Pydantic object based on the json of its schema.

    Attributes:
        schema_json: str, represents the schema of a Pydantic object, which
            can be obtained through BaseModelClass.schema_json()
```

AlexejPenner (approved):
Had a quick look and this looks good, would this also tell the user which of the fields are required and if any special validation conditions need to be met? Like s3:// as prefix to any path for S3

bcdurak:
Yes and no. It will show below which fields are required but it wont show any further validation criteria coming from pydantic.

PR #906, opened 2022-09-29
Backend changes to enable the new Telemetry flow

AlexejPenner:
The Basic functionality is implemented, I will test this after lunch, I just wanted to share as early as possible to get super quick iterations in :

UserModels no longer have an email field as this is not used for anything besides telemetry
UserModels now have the email_opted_in field with three states (`null`, True, False)

User table in db now has an optional email field and the optional email_opted_in field with the above mentioned three states.

The ZenStore now has the .user_email_opt_in() method to allow setting theses fields
```    
def user_email_opt_in(
        self,
        user_name_or_id: Union[str, UUID],
        email: str,
        user_opt_in_response: bool
    ) -> None:
        """Persist user response to the email prompt.

        Args:
            user_name_or_id: The name or the ID of the user.
            email: The users email
            user_opt_in_response: Whether this email should be associated
                                  with the user id in the telemtry
        Raises:
            KeyError: If no user with the given name exists.
        """
```

The Endpoint is as follows: 
Either the email is set and the opt_in is set to True, or the opt_in is set to false, even in this case the email would be persisted, but we can also add some logic to ignore the email field if the opt_in is set to false

image

## Describe changes

## Pre-requisites

## Types of changes


stefannica (approved):
Looks green !

htahir1 (approved):

PR #921, opened 2022-10-04
Prettify pipeline runs list CLI output.

fa9r:
## Describe changes
I made `zenml pipeline runs list` more pretty.

## Pre-requisites

## Types of changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -183,3 +183,6 @@ zenml_examples/
 
 # stack recipes
 test_terraform/
+
+# static dashboard files
+src/zenml/zen_server/dashboard/
--- a/src/zenml/cli/utils.py
+++ b/src/zenml/cli/utils.py
@@ -1109,7 +1109,7 @@ def print_pipeline_runs_table(
     runs_dicts = []
     for pipeline_run in pipeline_runs:
         if pipeline_run.pipeline_id is None:
-            pipeline_name = "N/A"
+            pipeline_name = "unlisted"
         else:
             pipeline_name = client.zen_store.get_pipeline(
                 pipeline_run.pipeline_id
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -182,4 +182,7 @@ zenml_examples/
 !examples/**/*.yaml
 
 # stack recipes
-test_terraform/
\ No newline at end of file
+test_terraform/
+
+# static dashboard files
+src/zenml/zen_server/dashboard/
```
strickvl, line 188: Not sure if this interferes with how we build our package? I just added it into my `~/.gitignore_global` file. but maybe it's ok?

src/zenml/cli/utils.py
```diff
@@ -1068,3 +1070,60 @@ def get_component_by_id_or_name_or_prefix(
...
+    Args:
+        client: Repository instance
+        pipeline_runs: List of pipeline runs
+    """
+    runs_dicts = []
+    for pipeline_run in pipeline_runs:
+        try:
+            pipeline_name = client.zen_store.get_pipeline(
+                pipeline_run.pipeline_id
+            ).name
```
strickvl, line 1116: This is potentially a lot of queries. Wondering if we could improve this somehow so that we're not iterating over dozens, potentially hundreds of these?

strickvl: For now it's fine, though, probably, so long as we don't overwhelm the server.

.gitignore
fa9r: @stefannica said it's fine, so should be good

src/zenml/cli/utils.py
fa9r: If this really makes problems then we could rewrite it to query all pipelines, stacks, and users and then find the respective pipeline/user/stack for each pipeline run from the list. However, since that is quite convoluted, I would propose we keep it as is for now and only do this if it really causes issues.

htahir1 (approved):
For me this works

PR #924, opened 2022-10-04
Fix the CI unit test and integration test failures

stefannica:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/steps/utils.py
+++ b/src/zenml/steps/utils.py
@@ -44,7 +44,6 @@ from typing import (
 )
 
 import pydantic
-from tfx.dsl.component.experimental.decorators import _SimpleComponent
 from tfx.dsl.components.base.base_executor import BaseExecutor
 from tfx.dsl.components.base.executor_spec import ExecutorClassSpec
 from tfx.orchestration.portable import outputs_utils
@@ -64,6 +63,8 @@ from zenml.steps.step_output import Output
 from zenml.utils import proto_utils, source_utils
 
 if TYPE_CHECKING:
+    from tfx.dsl.component.experimental.decorators import _SimpleComponent
+
     from zenml.config.step_configurations import ArtifactConfiguration
     from zenml.steps.base_step import BaseStep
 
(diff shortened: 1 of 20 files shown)


src/zenml/steps/utils.py
```diff
@@ -150,6 +151,8 @@ def create_component_class(step: "BaseStep") -> Type[_SimpleComponent]:
     Returns:
         The component class.
     """
+    from tfx.dsl.component.experimental.decorators import _SimpleComponent
```
strickvl, line 154: For future curiosity: what is this and what does importing it fix? Looks like magic!

tests/conftest.py
```diff
@@ -194,88 +194,88 @@ def clean_client(
 
 @pytest.fixture
 def sql_store() -> BaseZenStore:
-    with tempfile.TemporaryDirectory(suffix="_zenml_sql_test") as temp_dir:
-        store = SqlZenStore(
```
strickvl, line 198: For my understanding, what was wrong with the context? It was deleting the temporary files while they were still needed?

strickvl (approved):
This looks good to me. Just had a couple of questions, which you're welcome to ignore :)

src/zenml/steps/utils.py
stefannica: This import drags along the whole tfx.experimental module, which loads a lot of other modules. Among them, cryptography, which, if loaded in memory, won't allow any package updates through `zenml integration install` on Windows.

The root problem is that running _any_ CLI command (`zenml integration *` included) imports all the modules from _all_ CLI commands in memory, which 1. takes time and 2. makes uninstalling/upgrading some packages with pip impossible on Windows.

tests/conftest.py
fa9r: Would be curious as well

stefannica: the SQLite store database created in the temporary directory is still in use when the tests complete. Windows doesn't like if you're trying to remove open files. Without the context, the temp directory is left behind.

I vaguely remember you facing a similar issue with using this context approach in materializers.

fa9r (approved):
Thanks for fixing them!

PR #925, opened 2022-10-04
Fix running pipelines from different locations.

fa9r:
## Describe changes
I adjusted the pipeline registration to allow rerunning pipelines from different locations.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/step_configurations.py
+++ b/src/zenml/config/step_configurations.py
@@ -105,6 +105,37 @@ class StepSpec(StrictBaseModel):
         _, class_name = self.source.rsplit(".", maxsplit=1)
         return class_name
 
+    def __eq__(self, other: Any) -> bool:
+        """Returns whether the other object is referring to the same step.
+        This is the case if the other objects is a `StepSpec` with the same
+        `upstream_steps` and a `source` that meets one of the following
+        conditions:
+            - it is the same as the `source` of this step
+            - it refers to the same absolute path as the `source` of this step
+
+        Args:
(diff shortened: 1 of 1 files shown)


strickvl (approved):
Looks good to me, but does this solve the issue around absolute paths that we were also discussing earlier today?

fa9r:
> Looks good to me, but does this solve the issue around absolute paths that we were also discussing earlier today?

Not fully, but I would say it fixes 99% of practical issues, especially when combined with `zenml init`. 

Let's assume your folder structure is `/A/B/C/run.py` and you call the initial run from `/A/B/`. Then this fix allows you to run from `/A/B/C/`, `/A/`, `/A/D/`, `/A/E/F/`, ... Basically any of the parent folders of where `run.py` lies or folders higher than where the initial call was made.

What this doesn't fix is running from another different subdirectory within the same root folder, e.g., calling `python ../C/run.py` from `/A/B/G/`. However, this works too if you have run `zenml init` within `/A/B/`.

Now the 1% that is still not working is if you have your code in one folder and run it twice from two different unrelated locations, e.g. calling `../A/B/C/run.py` first from `/H/` and then from `/I/`.

strickvl:
> Now the 1% that is still not working is if you have your code in one folder and run it twice from two different unrelated locations, e.g. calling `../A/B/C/run.py` first from `/H/` and then from `/I/`.

Yeah that makes sense + I think that's a limited scenario that we can just fix later.

PR #932, opened 2022-10-05
Update cheat sheet

dnth:
## Describe changes
This PR adds server related commands to the ZenML cheat sheet.

## Pre-requisites

## Types of changes


fa9r:
This needs some adjustment since we have added several new commands in #879 and removed others.

In particular, of the commands on the cheat sheet, `profile create` and `profile set` no longer exist and need to be removed.

Also the following commands were added in #879:
- `zenml server connect / disconnect / down / up / explain / list / logs / status`
- `zenml config describe / explain / set`
- `zenml pipeline list / runs / delete`
- `zenml profile list / migrate`

`profile list` and `profile migrate` are mostly there for backwards compatibility, so I guess we don't need to mention them, but the others are all quite useful IMO.

PR #936, opened 2022-10-06
Implemented `docstring` Attribute of StepModel

fa9r:
## Describe changes
I implemeneted the logic that allows us to track docstrings that users have defined for their steps. This is useful, e.g., in post execution where you can now call `step.docstring` (which previously always returned None).

In order to implement this, I also added a `docstring` attribute into the `StepConfiguration` class.

## Pre-requisites

## Types of changes


htahir1 (commented):
Looks good. Can we write a small test for this where we check that its tracked properly? With empty docstrings and no docstring and a normal docstring

fa9r:
@htahir1 good idea, I'll add that.

PR #938, opened 2022-10-06
Fixing the bug in the registration of custom flavors

bcdurak:
## Describe changes
I fixed the registration of the custom flavors.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/stack_components.py
+++ b/src/zenml/cli/stack_components.py
@@ -926,11 +926,15 @@ def generate_stack_component_flavor_register_command(
             client = Client()
 
             # Create a new model
+            # TODO: Investigate how we can create this model without empty
+            #   strings as values for name and config_schema
             flavor_create_model = FlavorModel(
                 source=source,
                 type=component_type,
                 user=client.active_user.id,
                 project=client.active_project.id,
+                name="",
+                config_schema="",
             )
 
             # Register the new model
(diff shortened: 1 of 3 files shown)


src/zenml/cli/stack_components.py
```diff
@@ -926,11 +926,15 @@ def register_stack_component_flavor_command(source: str) -> None:
...
 
             # Create a new model
+            # TODO: Investigate how we can create this model without empty
+            #   strings as values for name and config_schema
             flavor_create_model = FlavorModel(
                 source=source,
                 type=component_type,
                 user=client.active_user.id,
                 project=client.active_project.id,
+                name="",
```
fa9r, line 936: shouldn't the name already be defined by the user in the source code? I believe we have some util functions for loading a class from path, in that case we could load `source` and simply call `.name` right?

fa9r: Ah, just saw you're doing exactly that in the client. In that case, I would suggest we just change the input argument types of the client function to expect something different than a `FlavorModel`.

Currently, you're also setting `type` and `user` both in the CLI and the client which isn't ideal either.

I would suggest to change `client.create_flavor` to `create_flavor(source, component_type)` and then have all the logic there.

fa9r (changes requested):
Good fix, but seems unfinished. The TODO you added seems not too hard to resolve, so why not fix it directly. Otherwise good to go!

fa9r (approved):
Nothing changed but looks good to me now 🤔

PR #943, opened 2022-10-07
Fix shared stacks not being allowed to be set as active.

fa9r:
## Describe changes
We previously had a check that automatically set the default stack as active if the currently active stack was not owned by the active user, which prevented shared stacks from being used by other users.

## Pre-requisites

## Types of changes


strickvl (approved):
🙌

bcdurak (approved):
looks good to me!

PR #946, opened 2022-10-07
Fix Kubeflow run name

safoinme:
## Describe changes
quick fix to Kubeflow run name placeholder 

## Pre-requisites

## Types of changes


strickvl (commented):
Was this the quick-fix? If so, is there any point making this fix if we're just going to do more serious fixes next week?

safoinme:
> @strickvl 
I'm not sure there is a straightforward fix to what's causing the error before this change, not sure how much it would take to fix/refactor the kubeflow orchestrator code. so maybe having it work at least is better than not. maybe @htahir1 have better argument for this

PR #951, opened 2022-10-10
Fix pipeline run dashboard URL for unlisted runs

fa9r:
## Describe changes
Calling `run.py(unlisted=True)` previously raised an error when trying to define the dashboard URL.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/pipelines/base_pipeline.py
+++ b/src/zenml/pipelines/base_pipeline.py
@@ -48,7 +48,6 @@ from zenml.config.pipeline_configurations import (
 from zenml.config.pipeline_deployment import PipelineDeployment
 from zenml.config.schedule import Schedule
 from zenml.config.step_configurations import StepConfigurationUpdate
-from zenml.enums import StoreType
 from zenml.environment import Environment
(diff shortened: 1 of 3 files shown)


src/zenml/pipelines/base_pipeline.py
```diff
@@ -553,9 +553,10 @@ def run(
             if runs:
                 # For now, take the first index to get the latest run
                 run = runs[0]
-                url = (
-                    f"{gc.store.url}/pipelines/{pipeline_id}/runs/{run.id}/dag"
-                )
+                url = gc.store.url
+                if register_pipeline:
```
htahir1, line 557: Maybe add a comment this if for unlisted runs, otherwise LGTM

src/zenml/pipelines/base_pipeline.py
```diff
@@ -553,9 +553,10 @@ def run(
             if runs:
```
schustmi, line 553: Can we maybe extract this functionality to display the dashboard URL into a separate method?

PR #961, opened 2022-10-13
Fix class resolving in Colab.

fa9r:
## Describe changes
Class resolving previously failed on Colab when importing steps from other files.

E.g. in ZenBytes we do:
```
!git clone https://github.com/zenml-io/zenbytes.git  # noqa
!mv zenbytes/steps .
!mv zenbytes/pipelines .

from steps.evaluator import evaluator
from steps.importer import importer
from steps.sklearn_trainer import svc_trainer

digits_svc_pipeline = digits_pipeline(
    importer=importer(), trainer=svc_trainer(), evaluator=evaluator()
)
```

which failed with `RuntimeError: Main module was not started from a file. Cannot determine the module root path.`
raised in `get_source_root_path` in `is_third_party_module` in `resolve_class`.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/source_utils.py
+++ b/src/zenml/utils/source_utils.py
@@ -117,9 +117,11 @@ def is_third_party_module(file_path: str) -> bool:
         if pathlib.Path(path).resolve() in absolute_file_path.parents:
             return True
 
-    return (
-        pathlib.Path(get_source_root_path()) not in absolute_file_path.parents
-    )
+    try:
+        source_root_path = pathlib.Path(get_source_root_path())
+        return source_root_path not in absolute_file_path.parents
+    except RuntimeError:
+        return False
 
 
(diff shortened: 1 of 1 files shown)


schustmi:
@htahir1 ran into this exact issue yesterday, and it is fixed by doing a `zenml init`. What do the colab files in your case get resolved to after this change?

Overall the logic here seems a little flawed to me: This change essentially says that when running in colab, nothing can be a third party module?

fa9r:
@schustmi thanks for the hint, `zenml init` fixes this indeed.

fa9r:
@schustmi something about `is_third_party_module` raising an error still seems odd to me though.

schustmi:
Yes I agree. Maybe we need a special case here that detects that we're running in a notebook environment and maybe uses the notebook path as the "source root"?

PR #962, opened 2022-10-13
First iteration of the CLI optimization

bcdurak:
## Describe changes
I have optimized the general interaction around our CLI and removed a lot of unnecessary imports. To sum it up with some numbers:

at `0.20.0`: `from zenml import cli` would import 4388 modules taking 4.2 seconds :turtle:
now with this PR: `from zenml import cli` is importing 451 modules taking 0.25 seconds :rabbit2: 

Some operations might still take some time and I am hoping to fix those with the next iteration.

## Pre-requisites

## Types of changes


src/zenml/cli/stack_recipes.py
```diff
@@ -60,159 +53,6 @@
     f'`pip install "zenml[stacks]=={zenml.__version__}"`.'
 )
 
-try:
```
strickvl, line 63: what happened here with this mega-deletion?

src/zenml/exceptions.py
```diff
@@ -173,6 +172,8 @@ def __init__(
             parameters_class: Class of the parameters object for which
                 the parameters are missing.
         """
+        import textwrap
```
strickvl, line 175: only thinking about this since it came up earlier today when we were talking. I'm not entirely sure `textwrap` from the `stdlib` is the best way to do this? Or I'm not sure if there's not a better way using `rich` or our cli_utils etc?

strickvl: Can fully ignore this...

strickvl (approved):
Nice work so far, pending failing tests etc! I manually tested it out a bit and the speedup is noticeable if you're looking for it, though I guess our CLI still isn't the fastest car on the block due to things more out of our control...

src/zenml/cli/stack_recipes.py
wjayesh: we shifted the stack recipe service to a separate class under zenml.recipes

src/zenml/cli/annotator.py
```diff
@@ -112,7 +113,7 @@ def dataset_stats(annotator: "BaseAnnotator", dataset_name: str) -> None:
     )
     @click.pass_obj
     def dataset_delete(
-        annotator: "BaseAnnotator", dataset_name: str, all: bool
+        annotator: "BaseAnnotator", dataset_name: str, all_: bool
```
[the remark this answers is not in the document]
bcdurak: Yes. As I was scrolling through the pages for the sake of optimization, I was warned that we are using names that were shadowing some built-in names in Python. I thought I would change them as well.

src/zenml/exceptions.py
bcdurak: I am open to any suggestions tbh. I didn't know any alternatives so I left it as it is.

src/zenml/cli/secret.py
```diff
@@ -79,7 +79,7 @@ def secret(ctx: click.Context) -> None:
         "--schema",
         "-s",
         "secret_schema_type",
-        default=ARBITRARY_SECRET_SCHEMA_TYPE,
+        default="arbitrary",  # TODO: Place in a constant outside secret module
```
stefannica, line 82: can't you just move `ARBITRARY_SECRET_SCHEMA_TYPE` to `zenml.constants` or something ?

src/zenml/cli/server.py
```diff
@@ -178,12 +173,17 @@ def up(
     if ip_address is not None:
         config_attrs["ip_address"] = ip_address
 
+    from zenml.zen_server.deploy.deployment import ServerDeploymentConfig
```
stefannica, line 176: the linter isn't going to like these mid-block imports :)

src/zenml/client.py
```diff
@@ -37,19 +35,24 @@
...
+    from zenml.enums import StackComponentType
+    from zenml.models import (
+        ComponentModel,
+        FlavorModel,
+        HydratedStackModel,
+        PipelineModel,
+        ProjectModel,
+        StackModel,
+        UserModel,
+    )
```
stefannica, line 53: I think you may have overdone it here. Not sure I understand why removing the model imports makes the CLI or the client faster. The models are pretty thin in terms of libraries that they depend on. This is where I would have drawn the line.

stefannica (approved):
Amazing work ! Can't wait to use these awesome improvements !

src/zenml/cli/server.py
bcdurak: fixed it!

src/zenml/cli/secret.py
bcdurak: moved it!

PR #935, opened 2022-10-05
`VertexOrchestrator` apply node selector constraint if `gpu_limit > 0`

gabrielmbmb:
## Describe changes
When using the `VertexOrchestrator` and providing a value for the `node_selector_constraint` to get a GPU, all the steps will get a GPU even if it is not needed for that step. This is undesired, because you may want the training step to use GPU, but not all steps in the pipeline.

With this PR you can do something like:

```python
from zenml.steps import step, ResourceConfiguration

@step(resource_configuration=ResourceConfiguration(cpu_count=4.0, gpu_count=0, memory="64GB"))
def preprocess() -> int:
    ...

@step(resource_configuration=ResourceConfiguration(cpu_count=4.0, gpu_count=1, memory="64GB"))
def train(a: int, b:int) -> int:
    ...
```
and expect only the `train` step to request the GPU specified in the `node_selector_constraint` attribute to Vertex.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [ ] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/mlops-stacks/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #963, opened 2022-10-13
Fix `gpu_limit` condition in `VertexOrchestrator`

gabrielmbmb:
## Describe changes

In #935, I added the condition to check if it was needed to add the node selector constraint if `gpu_limit` was `0`, but there was another bug when the value for `gpu_limit` was set, and the value could never be `0`. As `0` is evaluated like a falsy value, the value of `gpu_limit` would always be the value of `self.config.gpu_limit` in the case `resource_settings.gpu_count == 0`.

I've also added unit tests for `VertexOrchestrator._configure_container_resources` method.

## Pre-requisites

## Types of changes

### DIFF
--- a/tests/unit/integrations/gcp/orchestrators/test_vertex_orchestrator.py
+++ b/tests/unit/integrations/gcp/orchestrators/test_vertex_orchestrator.py
@@ -137,7 +137,7 @@ def test_vertex_orchestrator_stack_validation(
         # ResourceSettings specified for the step, should take values from here
         (
             ResourceSettings(cpu_count=1, gpu_count=1, memory="1GB"),
-            {"cpu_limit": None, "gpu_limit": 1, "memory_limit": "1G"},
+            {"cpu_limit": 4, "gpu_limit": 4, "memory_limit": "4G"},
             {
                 "accelerator": {"count": "1", "type": "NVIDIA_TESLA_K80"},
                 "cpuLimit": 1.0,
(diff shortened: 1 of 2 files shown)


tests/unit/integrations/gcp/orchestrators/test_vertex_orchestrator.py
```diff
@@ -125,3 +129,96 @@ def test_vertex_orchestrator_stack_validation(
...
         ).validate()
+
+
+@pytest.mark.parametrize(
+    "resource_settings, orchestrator_resource_settings, expected_resources",
+    [
+        # ResourceSettings specified for the step, should take values from here
+        (
+            ResourceSettings(cpu_count=1, gpu_count=1, memory="1GB"),
+            {"cpu_limit": None, "gpu_limit": 1, "memory_limit": "1G"},
```
fa9r, line 140: Since GPU and memory have the same values for both settings, those might not necessarily come from the ResourceSettings

fa9r (approved):
Looks good to me, great unit test!

schustmi (approved):
Looks good to me!

tests/unit/integrations/gcp/orchestrators/test_vertex_orchestrator.py
gabrielmbmb: updated!

PR #964, opened 2022-10-13
Detect failed executions

schustmi:
## Describe changes
Currently whenever an execution fails due to input resolving (this happens for example when the metadata store is not shared between steps), we wrongly log that the step was cached. This PR patches the TFX launcher to we can detect whether this happens and raise an exception.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/orchestrators/base_orchestrator.py
+++ b/src/zenml/orchestrators/base_orchestrator.py
@@ -440,6 +440,9 @@ class BaseOrchestrator(StackComponent, ABC):
 
         Returns:
             Optional execution info returned by the launcher.
+
+        Raises:
+            RuntimeError: If the execution failed during preparation.
         """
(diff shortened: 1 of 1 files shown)


htahir1:
This is much needed. I also notice the cast that when someone interupts a pipeline in between steps, the current ZenML Server detects it as always Running. Is that also something we can fix?

htahir1 (approved):
Using `nonlocal` like a boss :P

schustmi:
I think handling the keyboard interrupt would be much more complicated as it can happen in many stages during the process:
- Before the execution is registered in the MLMD. At that point the execution ID doesn't even exist yet
- During input resolution
- While the executor is running

At this point I think we would have to write a custom launcher class to handle all of this, do you think this is worth it or should we instead wait until we can remove it entirely?

htahir1:
@schustmi Ok lets merge this and wait for the removal of our dependencies

PR #967, opened 2022-10-13
Fix the API docs builder

stefannica:
## Describe changes
Implemented a strategy for installing integrations in a way that doesn't break the API docs builder.
This also required a few additional changes:

1. no longer instantiating the ZenML store when the REST API utils module is being imported
2. using lazy initialization for the zen store in general, to avoid having to initializing the local sqlite zen store or having to connect to a zen server when not necessary (e.g. when running `zenml integration install`)
3. updating some integration package requirements:
  * kfp and kfp-tekton were out of sync with eachother
  * feast was using an older package version that was no longer working 

Tested that this works (i.e. it builds the API docs successfully in release branches) here: https://github.com/zenml-io/zenml/actions/runs/3245990909/jobs/5324213166

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/feast/__init__.py
+++ b/src/zenml/integrations/feast/__init__.py
@@ -32,7 +32,7 @@ class FeastIntegration(Integration):
     """Definition of Feast integration for ZenML."""
 
     NAME = FEAST
-    REQUIREMENTS = ["feast[redis]~=0.26.0", "redis-server"]
+    REQUIREMENTS = ["feast[redis]~=0.26.0", "redis-server>=6.0.9"]
 
     @classmethod
     def flavors(cls) -> List[Type[Flavor]]:
--- a/src/zenml/integrations/feast/__init__.py
+++ b/src/zenml/integrations/feast/__init__.py
@@ -32,7 +32,7 @@ class FeastIntegration(Integration):
     """Definition of Feast integration for ZenML."""
 
     NAME = FEAST
-    REQUIREMENTS = ["feast[redis]>=0.19.4", "redis-server"]
+    REQUIREMENTS = ["feast[redis]~=0.26.0", "redis-server"]
 
     @classmethod
     def flavors(cls) -> List[Type[Flavor]]:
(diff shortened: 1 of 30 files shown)


src/zenml/integrations/feast/__init__.py
```diff
@@ -32,7 +32,7 @@ class FeastIntegration(Integration):
     """Definition of Feast integration for ZenML."""
 
     NAME = FEAST
-    REQUIREMENTS = ["feast[redis]>=0.19.4", "redis-server"]
+    REQUIREMENTS = ["feast[redis]~=0.26.0", "redis-server"]
```
strickvl, line 35:
```suggestion
    REQUIREMENTS = ["feast[redis]~=0.26.0", "redis-server>=6.0.9"]
```

I made the same change in my branch working on fixing the Feast example, but added this. Hope pinning Redis doesn't mess with anything.

strickvl (changes requested):
Looks great to me, but hope you can add in the `redis-server` change I suggested + that it doesn't break other things since I've added that same extended change elsewhere.

strickvl:
Am certain @AlexejPenner will be happy about this PR.

AlexejPenner (approved):
Hallelujah, lgtm!

src/zenml/integrations/feast/__init__.py
stefannica: That shouldn't be a problem. I'll test it, just in case.

fa9r (approved):
Awesome, finally automated again 🙌

stefannica:
New try to build the API docs with these changes and the GitHub action: https://github.com/zenml-io/zenml/actions/runs/3249084707/jobs/5331068077

src/zenml/zen_server/zen_server_api.py
```diff
@@ -73,6 +77,16 @@ def relative_path(rel: str) -> str:
     allow_headers=["*"],
 )
 
+
+@app.on_event("startup")
+def initialize() -> None:
+    """Initialize the ZenML server."""
+    # IMPORTANT: this needs to be done before the fastapi app starts, to avoid
+    # race-conditions
```
strickvl, line 85:
```suggestion
    # race conditions
```

strickvl (approved):

PR #972, opened 2022-10-14
K3d local registry pod spec updated

wjayesh:
## Describe changes
I fixed a problem with the k3d cluster when a default container registry is used. 
The pod spec should have the `k3d-zenml...` prefix on the image name for it to be pulled successfully inside the pod. This PR adds that.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
+++ b/src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
@@ -549,7 +549,7 @@ class KubeflowOrchestrator(BaseOrchestrator):
         if (
             self.is_local
             and stack.container_registry.flavor
-            is ContainerRegistryFlavor.DEFAULT.value
+            == ContainerRegistryFlavor.DEFAULT.value
         ):
             image_name = f"k3d-zenml-kubeflow-registry.{image_name}"
         is_scheduled_run = bool(deployment.schedule)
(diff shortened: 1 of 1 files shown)


schustmi:
Can you explain in a little more detail what the issue was here? From what I know this worked fine with the existing code, what lead to the failure and why does this prefix solve it?

The final image name would be something like `k3d-zenml-kubeflow-registry.localhost:5000/zenml:some_pipeline`, which seems very strange to me

wjayesh:
@schustmi 
In the k3d [docs](https://k3d.io/v5.4.4/usage/registries/#testing-your-registry), it says:
> Note: as per the explanation in the [preface](https://k3d.io/v5.4.4/usage/registries/#preface-referencing-local-registries), you could replace k3d-registry.localhost:12345 with localhost:12345 in the docker tag and docker push commands below (but not in the kubectl part!)

It means that the image name should indeed be: `k3d-zenml-kubeflow-registry.localhost:5000/zenml:some_pipeline` and not just `localhost:..`. I also verified that it didn't work for me earlier (after a user raised this as a problem on Slack).

schustmi:
@wjayesh You're right, I just ran some additional tests:
- When using a local container registry with port 5000 (`container_registry.uri="localhost:5000"`), the pods can pull images both using `localhost:5000/image-name:tag` as well as `k3d-zenml-kubeflow-registry.localhost:5000/image-name:tag`
- When using a different port, only the latter works. Not sure why that is, but your fix makes it work with all additional ports. Great job!

I have one additional suggestion to make this even cleaner: Instead of checking `stack.container_registry.flavor== ContainerRegistryFlavor.DEFAULT.value`, we can use the property that checks if a container registry is local: `stack.container_registry.config.is_local`

wjayesh:
Thanks for testing the fix, Michael! I have added your suggestion.

schustmi (approved):
LGTM!

PR #977, opened 2022-10-17
Feature/eng 1394 pipeline run export import migrate

fa9r:
## Describe changes

## Pre-requisites

## Types of changes


fa9r:
@htahir1 wasn't fully ready for review yet :D

src/zenml/zen_stores/schemas/pipeline_schemas.py
```diff
@@ -138,6 +138,7 @@ class PipelineRunSchema(SQLModel, table=True):
     )
     pipeline: PipelineSchema = Relationship(back_populates="runs")
 
+    status: ExecutionStatus
```
stefannica, line 141: Could you please make this field optional to maintain backwards compatibility with existing database entries ? Until we implement a mechanism for database migration, this should be our strategy, otherwise users will have to drop their DBs when they update zenml, which is not desirable. I'm getting this if I use my current database with this PR: 
```OperationalError: (sqlite3.OperationalError) no such column: pipelinerunschema.status
[SQL: SELECT pipelinerunschema.id 
FROM pipelinerunschema 
WHERE pipelinerunschema.status = ?]
[parameters: (<ExecutionStatus.RUNNING: 'running'>,)]```

File: src/zenml/zen_stores/schemas/pipeline_schemas.py [hunk 2]
```diff
@@ -227,12 +225,13 @@ class StepRunSchema(SQLModel, table=True):
 
     pipeline_run_id: UUID = Field(foreign_key="pipelinerunschema.id")
 
+    status: ExecutionStatus
```
stefannica, line 228: Same here: status should be optional/nullable

src/zenml/zen_server/routers/artifacts_endpoints.py
```diff
@@ -36,15 +39,40 @@
...
+    )
+
+
+@router.post(
+    "",
+    response_model=ArtifactModel,
+    responses={401: error_response, 409: error_response, 422: error_response},
+)
+@handle_exceptions
+def create_artifact(step: ArtifactModel) -> ArtifactModel:
```
schustmi, line 69:
```suggestion
def create_artifact(artifact: ArtifactModel) -> ArtifactModel:
```

src/zenml/zen_server/routers/artifacts_endpoints.py
```diff
@@ -36,15 +39,40 @@
...
+    """Create a new artifact.
+
+    Args:
+        step: The artifact to create.
+
+    Returns:
+        The created artifact.
+    """
+    return zen_store().create_artifact(step)
```
schustmi, line 78:
```suggestion
    return zen_store().create_artifact(artifact)
```

src/zenml/zen_server/routers/artifacts_endpoints.py
```diff
@@ -36,15 +39,40 @@
...
+    "",
+    response_model=ArtifactModel,
+    responses={401: error_response, 409: error_response, 422: error_response},
+)
+@handle_exceptions
+def create_artifact(step: ArtifactModel) -> ArtifactModel:
+    """Create a new artifact.
+
+    Args:
+        step: The artifact to create.
```
schustmi, line 73:
```suggestion
        artifact: The artifact to create.
```

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -1371,26 +1371,24 @@ def get_run_step(self, step_id: UUID) -> StepRunModel:
...
-        }
+        filters = locals()
+        filters.pop("self")
+        return self._list_resources(
+            route=STEPS,
+            resource_model=StepRunModel,
+            **filters,
+        )
 
     def get_run_step_inputs(self, step_id: UUID) -> Dict[str, ArtifactModel]:
```
stefannica, line 1393: Just a side-question: if `get_run_step_outputs` is replaced by `list_artifacts`, shouldn't `get_run_step_inputs` also be achievable through a `list_artifacts` call ?

src/zenml/client.py
```diff
@@ -1205,6 +1209,234 @@ def register_pipeline(
...
+        """
+        import json
+
+        from zenml.utils.yaml_utils import write_yaml
+
+        pipeline_runs = self.zen_store.list_runs(
+            project_name_or_id=self.active_project.id
+        )
+        if not pipeline_runs:
+            raise RuntimeError("No pipeline runs found.")
```
schustmi, line 1229: This could be handled more gracefully with a simple warning message that nothing got exported IMO

src/zenml/client.py
```diff
@@ -1205,6 +1209,234 @@ def register_pipeline(
...
+        yaml_data = []
+        for pipeline_run in pipeline_runs:
+            pipeline_run.status = self.zen_store.get_run(pipeline_run.id).status
```
schustmi, line 1232: Why do we need an additional call to `get_run` here? Isn't `pipeline_run.status` the same thing that we got back from `list_runs`?

src/zenml/client.py
```diff
@@ -1205,6 +1209,234 @@ def register_pipeline(
...
+            run_dict = json.loads(pipeline_run.json())
+            run_dict["steps"] = []
+            steps = self.zen_store.list_run_steps(run_id=pipeline_run.id)
+            for step in steps:
+                step.status = self.zen_store.get_run_step_status(step.id)
```
schustmi, line 1237: I think you changed this for pipeline runs to be `get_run(...).status`. Does it make sense to have the same API for steps as well, `get_step(...).status`?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2611,169 +2648,468 @@ def list_runs(
...
+    ) -> Dict[str, Any]:
+        """Gets the side effects for a component in a pipeline run.
 
         Args:
-            step_id: The id of the step to get outputs for.
+            run_id: The ID of the pipeline run to get.
+            component_id: The ID of the component to get.
+        """
+        # TODO: raise KeyError if run doesn't exist
+        pass  # TODO
```
schustmi, line 2694: I'm wondering how mypy isn't complaining about the return type of this method, do we have it disabled for the entire file or something?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3011,34 +3351,47 @@ def _sync_runs(self) -> None:
...
+        # Sync steps of all unfinished runs.
+        with Session(self.engine) as session:
+            unfinished_run_ids = session.exec(
+                select(PipelineRunSchema.id).where(
+                    PipelineRunSchema.status == ExecutionStatus.RUNNING
+                )
+            ).all()
+        for run_id in unfinished_run_ids:
+            self._sync_run_steps(run_id)
+            self.get_run(run_id)  # Updates run status if it has changed.
```
schustmi, line 3394: This seems like an expensive call just to sync the status to the run. Can we maybe include that logic in the `sync_run_steps` if that is already doing all the queries?

src/zenml/zen_server/routers/artifacts_endpoints.py
stefannica, line 78:
```suggestion
def create_artifact(artifact: ArtifactModel) -> ArtifactModel:
    """Create a new artifact.

    Args:
        artifact: The artifact to create.

    Returns:
        The created artifact.
    """
    return zen_store().create_artifact(artifact)
```

schustmi (changes requested):

stefannica (changes requested):
It's great to see the new create run/step endpoints and the simplification of "get_status" functionality. This brings us one step closer to removing the MLMD dependency. It would be better if existing users didn't have to drop their DBs to gain access to the new import and migration features and the pipeline run sync performance improvements. Hopefully, it's not too difficult to maintain backwards compatibility.

src/zenml/zen_stores/schemas/pipeline_schemas.py
stefannica: On second thought, the solution I suggested won't work, because the sqlmodel schema will still want to access that new field on reads/writes.

stefannica: Perhaps it's time to introduce SQLModel to alembic: https://github.com/tiangolo/sqlmodel/issues/85

src/zenml/models/pipeline_models.py
```diff
@@ -100,14 +101,16 @@ class StepRunModel(DomainModel):
 
     pipeline_run_id: UUID
     parent_step_ids: List[UUID]
+    input_artifacts: Dict[str, UUID]  # mapping from input name to artifact ID
```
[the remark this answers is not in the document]
fa9r: Which artifact is an output of which step is a 1-to-N relationship defined by the `ArtifactModel.parent_step_id` attribute, so, for internal purposes, we don't necessarily need to store this link in the `StepRunModel` again.

Input artifacts are more tricky since that is an N-to-M relationship, which we currently store as a separate table `StepRunInputSchema`. `StepRunModel.input_artifacts` is the equivalent on model level.

src/zenml/zen_server/routers/artifacts_endpoints.py
fa9r: Good catch!

src/zenml/client.py
fa9r: I agree, makes sense.

src/zenml/client.py
fa9r: Yes, good point, I think we could remove the `get_run_status()` method as well. Will do.

src/zenml/client.py
fa9r: Initially no, but now yes :D

Good catch, I'll remove the call.

src/zenml/zen_stores/sql_zen_store.py
fa9r: We do have mypy enabled for the file. I guess mypy doesn't complain about return types if you use `pass`?

src/zenml/zen_stores/rest_zen_store.py
fa9r: Unfortunately not since that information is not contained in the `ArtifactSchema`, see https://github.com/zenml-io/zenml/pull/977#discussion_r998077458

src/zenml/zen_stores/sql_zen_store.py
schustmi: Mhm I didn't know (and expect) that, but I just checked and you're correct. I guess it would be safer if we raised/asserted something instead, but doesn't matter in this case I think

src/zenml/zen_stores/sql_zen_store.py
fa9r: I like the idea of merging step and status sync into a single method `sync_run()`. It feels very clean at first glance. Unfortunately, after implementing it, I noticed that this leads to a difficult circular dependency:
- `sync_run()` calls `get_run()` to sync the status
- `get_run()` needs to have all steps to determine the status and accordingly needs to call `sync_run()`

src/zenml/cli/pipeline.py
```diff
@@ -156,3 +157,78 @@ def list_pipeline_runs(
...
+    """
+    cli_utils.print_active_config()
+    client = Client()
+    try:
+        client.import_pipeline_runs(filename=filename)
+    except EntityExistsError as err:
+        cli_utils.error(str(err))
+
+
+@runs.command("migrate", help="Migrate pipeline runs from a MySQl DB file.")
```
schustmi, line 191: This isn't limited to mysql right?

fa9r: Good point, I changed the wording.

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -1340,22 +1324,51 @@ def list_runs(
...
+    def create_run_step(self, step: StepRunModel) -> StepRunModel:
+        """Creates a step.
+
+        Args:
+            step: The step to create.
+
+        Returns:
+            The created step.
+        """
+        return self._create_resource(
```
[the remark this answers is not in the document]
fa9r: It is not a stupid question at all. 

Currently, steps and artifacts cannot exist without a corresponding run. As such, the owner/project of those resources are defined only at run level and all steps and artifacts of a run are assumed to have the same project/owner. You can also see this in our internal models, e.g. `zenml.models.pipeline_models.StepRunModel` is only a `DomainModel`, not a `ProjectScopedDomainModel`.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2611,169 +2670,436 @@ def list_runs(
             runs = session.exec(query).all()
             return [run.to_model() for run in runs]
```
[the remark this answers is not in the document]
fa9r: Yes, you're right, when running inside a server this status might be outdated until the next sync happens. I changed it to also update the status here.

src/zenml/zen_stores/migrations/versions/ccd68b7825ae_add_status_to_pipeline_and_step_run.py
```diff
@@ -0,0 +1,56 @@
...
+                "status", sqlmodel.sql.sqltypes.AutoString(), nullable=True
+            ),
+        )
+    op.execute("UPDATE pipelinerunschema SET status = 'running'")
+    with op.batch_alter_table("pipelinerunschema", schema=None) as batch_op:
+        batch_op.alter_column(
+            "status",
+            nullable=False,
+            existing_type=sqlmodel.sql.sqltypes.AutoString(),
+        )
```
AlexejPenner, line 33: You might be able to get away with a server_default instead of this three step approach  - https://www.loom.com/share/6b57aa48248245f9904de4ce1782a2e4

This will anchor the default value in the table definitions as well, not sure if that is wanted behaviour, that whenever the field is not specified explicitely it defaults to running. What do you think?

```suggestion
    with op.batch_alter_table("pipelinerunschema", schema=None) as batch_op:
        batch_op.add_column(
            sa.Column(
                "status", sqlmodel.sql.sqltypes.AutoString(), nullable=True, server_default="running"
            ),
        )
```

src/zenml/zen_stores/migrations/versions/ccd68b7825ae_add_status_to_pipeline_and_step_run.py
```diff
@@ -0,0 +1,56 @@
...
+                "status", sqlmodel.sql.sqltypes.AutoString(), nullable=True
+            ),
+        )
+    op.execute("UPDATE steprunschema SET status = 'running'")
+    with op.batch_alter_table("steprunschema", schema=None) as batch_op:
+        batch_op.alter_column(
+            "status",
+            nullable=False,
+            existing_type=sqlmodel.sql.sqltypes.AutoString(),
+        )
```
AlexejPenner, line 47: Same question as above

AlexejPenner (commented):
Just left one small comment on the migration script, looking forward to read your thoughts

stefannica:
It took a bit of doing, but I tested that this PR picks up pre-0.20.0 pipeline runs correctly when connected to a database previously used for the metadata store component:

```
┏━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━┯━━━━━━━━━━━┯━━━━━━━┓
┃ PIPELINE NAME │ RUN NAME                                           │ RUN ID                               │ STATUS │ STACK     │ OWNER ┃
┠───────────────┼────────────────────────────────────────────────────┼──────────────────────────────────────┼────────┼───────────┼───────┨
┃   unlisted    │ drift_detection_pipeline-24_Oct_22-22_19_50_727314 │ 00b994c3-15c8-4c55-a676-a1a2f6b2db69 │ ✅     │ [DELETED] │ zenml ┃
┠───────────────┼────────────────────────────────────────────────────┼──────────────────────────────────────┼────────┼───────────┼───────┨
┃   unlisted    │ airflow_example_pipeline-24_Oct_22-21_54_02_806786 │ b3b05b2e-96e9-4ab2-b4bc-33c5e53cb0c3 │ ✅     │ [DELETED] │ zenml ┃
┠───────────────┼────────────────────────────────────────────────────┼──────────────────────────────────────┼────────┼───────────┼───────┨
┃   unlisted    │ drift_detection_pipeline-24_Oct_22-22_10_45_219072 │ b4a96c36-aa26-42b7-8084-a81e52235534 │ ✅     │ [DELETED] │ zenml ┃
┠───────────────┼────────────────────────────────────────────────────┼──────────────────────────────────────┼────────┼───────────┼───────┨
┃   unlisted    │ airflow_example_pipeline-24_Oct_22-21_59_32_145383 │ bb703caa-310a-4043-95aa-7ce6c8348454 │ ✅     │ [DELETED] │ zenml ┃
┗━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━┷━━━━━━━━━━━┷━━━━━━━┛
```

src/zenml/zen_stores/migrations/versions/ccd68b7825ae_add_status_to_pipeline_and_step_run.py
stefannica: @AlexejPenner , we explicitly tried to avoid using a server default, because that configures an additional layer of logic in the database that can create unpredictable behavior.

stefannica:
@fa9r the new pipeline runs show up incomplete in the UI, any idea why ? This is for an airflow example run:

Screenshot from 2022-10-24 23-01-08

src/zenml/zen_stores/rest_zen_store.py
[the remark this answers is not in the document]
schustmi: Thanks for the explanation!

fa9r:
@stefannica should be fixed now, could you try again?

schustmi (approved):

docker/zenml-dev.Dockerfile
```diff
@@ -0,0 +1,19 @@
+ARG PYTHON_VERSION=3.9
+FROM python:${PYTHON_VERSION}-slim AS base
+
+ENV PYTHONFAULTHANDLER=1 \
+    PYTHONUNBUFFERED=1 \
+    PYTHONHASHSEED=random \
+    PIP_NO_CACHE_DIR=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1
+
```
schustmi, line 9: Can you add   `ZENML_CONTAINER=1` here as well please

PR #978, opened 2022-10-17
Automatically open browser upon `zenml up` command

dnth:
## Describe changes
This PR adds a feature to automatically open the ZenML dashboard in a browser upon calling `zenml up`.

## Pre-requisites

## Types of changes


src/zenml/cli/server.py
```diff
@@ -242,6 +244,7 @@ def up(
                 f"'{server.status.url}'. You can connect to it using the "
                 f"'{DEFAULT_USERNAME}' username and an empty password."
             )
+            webbrowser.open(server.status.url)
```
htahir1, line 247: What is this fails? Maybe we should have try catch here for known errors

src/zenml/cli/server.py
```diff
@@ -242,6 +244,7 @@ def up(
                 f"'{server.status.url}'. You can connect to it using the "
```
htahir1, line 244: This log should change now to reflect that the browser is opening?

src/zenml/cli/server.py
```diff
@@ -21,6 +21,8 @@
 import yaml
 from rich.errors import MarkupError
 
+import webbrowser
```
htahir1, line 24: Is there a way to override this behavior? Though some env variable?

htahir1 (changes requested):
Great! left a few comments. Also a question, is the webbrowser library part of core python or part of a dependency

src/zenml/cli/server.py
fa9r: I agree, there should be some environment variable that you can set if you don't want it to pop up. Also, I would recommend to move the import into the method where it is used, then you can check 
```
if not <ENV_VAR>:
    import webbrowser
    webbrowser.open(...)
```

src/zenml/cli/server.py
fa9r: IMO not necessary, people will notice :D

src/zenml/cli/server.py
fa9r: Also, moving the import is needed anyways to not slow down other CLI calls

fa9r (changes requested):
Very cool feature. I agree with @htahir1 that there should be an environment variable that you can set if you don't want to have it pop up, and I would recommend to move the import into the function where it is used. Otherwise looks good to me!

src/zenml/cli/server.py
dnth: Since we're using env variables, I'm putting a statement here to inform users about it. Is this necessary you think?

src/zenml/cli/server.py
dnth: I've added a try catch statement here.

dnth:
> Great! left a few comments. Also a question, is the webbrowser library part of core python or part of a dependency

The webbrowser library is part of the python standard library so we don't have to add it to our dependencies.

src/zenml/cli/server.py
```diff
@@ -240,9 +241,17 @@ def up(
...
                 f"'{server.status.url}'. You can connect to it using the "
-                f"'{DEFAULT_USERNAME}' username and an empty password."
+                f"'{DEFAULT_USERNAME}' username and an empty password. To open the dashboard in a browser automatically, set the env variable AUTO_OPEN_DASHBOARD=true."
             )
 
+            if os.environ.get(ENV_AUTO_OPEN_DASHBOARD) == "true":
+                try:
+                    import webbrowser
+                    webbrowser.open(server.status.url)
+                    cli_utils.declare(f"Automatically opening the dashboard in your browser. To disable this, set the the env variable AUTO_OPEN_DASHBOARD=false.")
```
fa9r, line 251:
```suggestion
                    cli_utils.declare(f"Automatically opening the dashboard in your browser. To disable this, set the env variable AUTO_OPEN_DASHBOARD=false.")
```

src/zenml/cli/server.py
```diff
@@ -240,9 +241,17 @@ def up(
             cli_utils.declare(
                 f"The local ZenML dashboard is available at "
                 f"'{server.status.url}'. You can connect to it using the "
-                f"'{DEFAULT_USERNAME}' username and an empty password."
+                f"'{DEFAULT_USERNAME}' username and an empty password. To open the dashboard in a browser automatically, set the env variable AUTO_OPEN_DASHBOARD=true."
             )
 
+            if os.environ.get(ENV_AUTO_OPEN_DASHBOARD) == "true":
```
fa9r, line 247: Here I would expect the default behavior to be that the dashboard opens unless you explicitly set this env variable. In `zenml.constants` we have a util function which can do exactly that (and is more lenient in which values the user may set for the variable).

```suggestion
            if handle_bool_env_var(ENV_AUTO_OPEN_DASHBOARD, default=True):
```

src/zenml/cli/server.py
```diff
@@ -30,6 +30,7 @@
 from zenml.logger import get_logger
 from zenml.utils import yaml_utils
 from zenml.utils.analytics_utils import AnalyticsEvent, track_event
+from zenml.constants import ENV_AUTO_OPEN_DASHBOARD
```
fa9r, line 33: If you would like to use that function below :D

```suggestion
from zenml.constants import ENV_AUTO_OPEN_DASHBOARD, handle_bool_env_var
```

fa9r (approved):
LGTM now!

Only suggestion: If the env var is not set, I'd change the default behavior to still open the dashboard, see code suggestions below.

src/zenml/constants.py
```diff
@@ -79,6 +79,7 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 ENV_ZENML_SKIP_PIPELINE_REGISTRATION = "ZENML_SKIP_PIPELINE_REGISTRATION"
 ENV_ZENML_SERVER_ROOT_URL_PATH = "ZENML_SERVER_ROOT_URL_PATH"
 ENV_ZENML_SERVER_DEPLOYMENT_TYPE = "ZENML_SERVER_DEPLOYMENT_TYPE"
+ENV_AUTO_OPEN_DASHBOARD = "AUTO_OPEN_DASHBOARD"
 ENV_ZENML_LOCAL_STORES_PATH = "ZENML_LOCAL_STORES_PATH"
```
htahir1, line 85: extra line

src/zenml/cli/server.py
```diff
@@ -240,9 +241,17 @@ def up(
             cli_utils.declare(
                 f"The local ZenML dashboard is available at "
                 f"'{server.status.url}'. You can connect to it using the "
-                f"'{DEFAULT_USERNAME}' username and an empty password."
+                f"'{DEFAULT_USERNAME}' username and an empty password. To open the dashboard in a browser automatically, set the env variable AUTO_OPEN_DASHBOARD=true."
```
htahir1, line 244:
```suggestion
                f"`{DEFAULT_USERNAME}` username and an empty password. To open the dashboard"
                " in a browser automatically, set the env variable AUTO_OPEN_DASHBOARD=true."
```

src/zenml/cli/server.py
```diff
@@ -240,9 +241,17 @@ def up(
...
             )
 
+            if handle_bool_env_var(ENV_AUTO_OPEN_DASHBOARD, default=True):
+                try:
+                    import webbrowser
+                    webbrowser.open(server.status.url)
+                    cli_utils.declare(f"Automatically opening the dashboard in your browser. To disable this, set the env variable AUTO_OPEN_DASHBOARD=false.")
```
htahir1, line 251:
```suggestion
                    cli_utils.declare(f"Automatically opening the dashboard in your browser... "
                                      "To disable this, set the env variable AUTO_OPEN_DASHBOARD=false.")
```

htahir1 (changes requested):
Almost there. Some lines are a bit too long and i left some suggestions, however those suggestions might be syntactically wrong so please check. Maybe try running formatting and linting locally too?

htahir1 (approved):

PR #979, opened 2022-10-17
Add a `just_mine` flag for `zenml stack list`

strickvl:
This allows users to see only the stacks that they created, useful when there are many shared stacks on a particular ZenML server.

I'll write a test to cover this if we think it's something we could potentially merge. In an ideal world, we'd replicate the ability to filter this list just as we have on the ZenServer API, but in the meanwhile this would be a useful addition.

## Pre-requisites

## Types of changes


schustmi (approved):
I like the idea of having this while we work on the more generic filtering 👍 

Personally I'd prefer having the option called something like `--exclude-shared` or `--no-shared` as I've seen those used in other CLIs, but happy to merge in any case!

strickvl:
@schustmi what if you share some of your own stacks? wouldn't excluding shared stacks also mean those aren't visible?

src/zenml/cli/stack.py
```diff
@@ -685,12 +685,21 @@ def rename_stack(
...
+def list_stacks(just_mine: str = False) -> None:
+    """List all available stacks.
+
+    Args:
+        just_mine: To list only the stacks that the current user has created.
+    """
     cli_utils.print_active_config()
 
     client = Client()
     stacks = client.stacks
```
fa9r, line 698: If you look at the implementation of `client.stacks`, that method is first querying all your stacks, then all shared stacks, and then it combines them. Instead of then taking the shared stacks out again if `just_mine`, I'd suggest to not add the shared stacks in the first place. Perhaps this could be done by splitting the `client.stacks` logic into two separate methods `get_my_stacks()` and `get_shared_stacks()`, then you could just call `client.get_my_stacks()` directly.

fa9r (changes requested):
I think this could be optimized a bit, see below. Otherwise LGTM!

schustmi:
@strickvl Mhm good point, `shared` in this case can mean both being shared with me and me sharing my stacks. Is there a word which signals only the former?

strickvl:
> @strickvl Mhm good point, `shared` in this case can mean both being shared with me and me sharing my stacks. Is there a word which signals only the former?

just-mine-and-shared? ;) Not really sure. I don't think there's a single word. Also I think I wouldn't want stacks shared with me in this list. Then we're back to a list of 30 stacks.

schustmi:
Yeah I agree, stacks shared with me definitely shouldn't be included in this call, that's the entire purpose of the flag right?

Can't come up with any other name for the option, but i would suggest naming it `--just-mine` instead of the underscore to keep it in line with (most) of our other options

fa9r (approved):
:)

PR #982, opened 2022-10-18
Add option to specify apt packages

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/graphviz/__init__.py
+++ b/src/zenml/integrations/graphviz/__init__.py
@@ -22,7 +22,6 @@ class GraphvizIntegration(Integration):
 
     NAME = GRAPHVIZ
     REQUIREMENTS = ["graphviz>=0.17"]
-    SYSTEM_REQUIREMENTS = {"graphviz": "dot"}
 
 
 GraphvizIntegration.check_installation()
--- a/src/zenml/utils/pipeline_docker_image_builder.py
+++ b/src/zenml/utils/pipeline_docker_image_builder.py
@@ -209,6 +209,7 @@ class PipelineDockerImageBuilder:
                 docker_settings.required_integrations,
                 docker_settings.replicate_local_python_environment,
                 docker_settings.install_stack_requirements,
+                docker_settings.apt_packages,
                 docker_settings.environment,
                 docker_settings.copy_files,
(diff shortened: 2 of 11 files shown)


stefannica (approved):
Looks great !

src/zenml/integrations/graphviz/__init__.py
```diff
@@ -22,7 +22,6 @@ class GraphvizIntegration(Integration):
 
     NAME = GRAPHVIZ
     REQUIREMENTS = ["graphviz>=0.17"]
-    SYSTEM_REQUIREMENTS = {"graphviz": "dot"}
```
fa9r, line 25: graphviz seems to be available via apt, how come we can't use the new `APT_PACKAGES ` here?

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -265,10 +266,21 @@ def build_docker_image(
             )
             requirements_file_names = [f[0] for f in requirement_files]
 
+            apt_packages = docker_settings.apt_packages
+            if docker_settings.install_stack_requirements:
+                apt_packages += stack.apt_packages
```
fa9r, line 271: would this still work if we have the same package in both `docker_settings.apt_packages ` and `stack.apt_packages`?

fa9r (approved):
Just had two minor questions I was wondering about, otherwise LGTM :)

src/zenml/utils/pipeline_docker_image_builder.py
schustmi: Yep that works without any issues from what I've tested

src/zenml/integrations/graphviz/__init__.py
schustmi: The `APT_PACKAGES` are essentially just used when a stack component of an integration is in the active stack when running a pipeline, it wouldn't really have any impact to add it for graphviz as that integration contains no stack component and also serves no real purpose while running pipeline. Would you still add it in this case for completeness?

fa9r: Ah I see. No, then it's fine as it is 👍

PR #984, opened 2022-10-18
Include deployment in azureml docker build

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


htahir1 (commented):
Shouldnt we, on principle, use fileio for all these remove and edit commands?

schustmi:
@htahir1 I think we don't have a convention for this case yet (this is guaranteed to be a local path -> fileio is not necessary) and I've seen both fileio as well as the default python method being used. Fine with whatever we decide though, I guess not using fileio signals that this is a local path which is why I've used it in these cases before

strickvl (approved):
This looks good to me. Thank you for the fix!

PR #988, opened 2022-10-19
Replace old flavor references, fix the windows local ZenML server and other fixes

stefannica:
## Describe changes
This PR is a mix of different fixes:

1. fix the remaining invalid references to `flavor` as a class field no longer available in the base StackComponent class
2. fix the windows local zenml server and add more information about missing support and available options
3. allow the zenml client to function when the server uses a default project name other than `default`
4. make the MLMD pipeline run sync period configurable with an env variable and in the zenml-server helm chart

## Pre-requisites

## Types of changes


src/zenml/cli/server.py
```diff
@@ -153,8 +155,23 @@ def up(
     gc = GlobalConfiguration()
 
     if docker:
+        if not check_docker():
+            cli_utils.error(
+                "Docker does not seem to be installed on your system. Please "
```
schustmi, line 160: Maybe add a "or running" here somewhere?

schustmi (approved):
All the changes look good to me and I'm happy to merge this as is!

I'll still leave one additional suggestion (no good deed goes unpunished 😝) here that would make this even nicer in my opinion:

You've already generalized the `get_active_...` methods to the base classes of some types, but I think we could go one step further and even implement it on the base `StackComponent` itself as a `get_active_component()` classmethod.

As an additional improvement which get's rid of the need of casting the return type of this function:

```python
from typing import TypeVar, Type

T = TypeVar("T", bound="C")

class C:
    @classmethod
    def get(cls: Type[T]) -> T:
        return cls()


class D(C):
    pass


reveal_type(C.get())  # Revealed type is "C*"
reveal_type(D.get())  # Revealed type is "D*"
```

This example uses a `TypeVar` to signal that the return of the classmethod is an instance of that class, which I think we could apply to the method on the stack component as well (or just the `get_active_model_deployer` and `get_active_data_validator` for now)

PR #990, opened 2022-10-19
Implement automatic alembic migration

AlexejPenner:
## Describe changes
Work in Progress

Remaining tasks listed on Jira on ticket ENG-1407
Design Doc in progress here: https://www.notion.so/zenml/How-to-Migrate-7ab7a42daab848b0977f3eb3b035dd85

## Pre-requisites

## Types of changes


src/zenml/alembic.ini
```diff
@@ -0,0 +1,108 @@
...
+# version_path_separator = :
+# version_path_separator = ;
+# version_path_separator = space
+version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url = sqlite:////home/apenner/.config/zenml/local_stores/default_zen_store/zenml.db
```
AlexejPenner, line 58: I'm still looking for a sane default for this

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -546,8 +545,35 @@ def _initialize(self) -> None:
...
+            revision: Version of the db-schema to up/downgrade to
+        """
+        from alembic.config import Config
+        from alembic import command
+
+        url, connect_args, engine_args = self.config.get_sqlmodel_config()
+        connection = self._engine.connect()
+
+        # TODO: This feels wrong, any suggestions how to improve this
+        ini_file_path = Path.absolute(Path(__file__).parents[1] / "alembic.ini")
```
AlexejPenner, line 568: Would really like to find a better way to do this

schustmi: `Path.absolute(Path(zenml.__file__).parents[0] / "alembic.ini")` 😛

schustmi: Potentially a tiny bit better, as we can now freely move the `sql_zen_store.py` without messing this up

src/zenml/zen_server/deploy/docker/docker_zen_server.py
```diff
@@ -190,6 +191,8 @@ def _get_container_cmd(self) -> Tuple[List[str], Dict[str, str]]:
             SERVICE_CONTAINER_GLOBAL_CONFIG_PATH,
             LOCAL_STORES_DIRECTORY_NAME,
         )
+        env[ENV_ZENML_DISABLE_DATABASE_MIGRATION] = "True"
```
schustmi, line 194: Can you explain why this is necessary? To me this seems the same as the local case (and should also be running a docker container which contains the same ZenML version as the local python environment) and therefore the migration is necessary? Or will the migration always happen locally (e.g. triggered by the call to `zenml up --docker` and this is included to save the check if migration should be done?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -546,7 +551,18 @@ def _initialize(self) -> None:
         self._engine = create_engine(
             url=url, connect_args=connect_args, **engine_args
         )
-        SQLModel.metadata.create_all(self._engine)
+        self._alembic = Alembic(self.engine)
+        if ENV_ZENML_DISABLE_DATABASE_MIGRATION not in os.environ:
+            self.migrate_database()
+        # SQLModel.metadata.create_all(self._engine)
```
schustmi, line 557: Can be removed I think

src/zenml/zen_stores/migrations/README.md
```diff
@@ -0,0 +1,51 @@
...
+SQLModel Schemas are used to create the zenml tables. This means, anyone using
+zenml in a fresh environment gets all migration scripts executed in historical 
+order to create the initial state of their database.
+
+
+## Create a revision
+
+
+1) Make sure the `store.url` of your global config points at an instance of the 
+   database that represents the status-quo before your changes to teh database
```
schustmi, line 32:
```suggestion
   database that represents the status-quo before your changes to the database
```

src/zenml/zen_stores/migrations/README.md
```diff
@@ -0,0 +1,51 @@
...
+## How alembic is works
+
+src/zenml/alembic.ini is used for the configuration of alembic in general. The
+env.py file is the entrypoint for alembic where we are able to influence some
+settings, like for example the location of our sqlite db.
+
+Now when you create an alembic revision using `alembic revision ...` the 
+database instance will be compared to the state of all SQLModels under 
+src/zenml/zen_stores/schemas. This means, in order for alembic to work properly
+for you, make sure you have a fully instantiated version of the `develop` state
```
schustmi, line 12: Instead of doing multiple revisions for each PR/change, does it make sense to instead do a single revision per release? Or would the speedup of this be negligible?

src/zenml/zen_stores/migrations/README.md
```diff
@@ -0,0 +1,51 @@
+# How to create a migration
+
+## How alembic is works
+
+src/zenml/alembic.ini is used for the configuration of alembic in general. The
```
schustmi, line 5: I think the `alembic.ini` file is missing in this PR? Also would it be possible to move that file into `src/zenml/zen_stores/migrations` or does it need to stay at the top level?

src/zenml/zen_server/deploy/docker/docker_zen_server.py
stefannica: The migration will always happen locally. This is to prevent the server from accidentally migrating a local database and leading to unexpected results (especially useful for us in development).

src/zenml/zen_stores/migrations/README.md
stefannica: I actually moved it out of the python package, into the repository root. It won't be part of the python package (still debatable if it should be).

src/zenml/zen_stores/migrations/README.md
AlexejPenner: I had a rather long debate with Stefan about this. Here is the reason that convinced me

Let's say Stefan has a PR with a Migration "3ak2l", on top of that you have a PR with a Migration "4owl42". Whenever we push these into develop, our dev servers are updated through these migrations and the version is stored. If we now combine these two migrations into a release migration "v0.20.5" we lose the history ... ->  "3ak2l" -> "4owl42" -> ...
and replace it with ... -> "v0.20.5" -> ... - This will lead to problems if we now try to upgrade our dev servers again, our dev server will be at version "4owl42" and alembic will not know anything about this version. So we will have to nuke our dev servers after every release. This same effect will also impact all of us during development.

fa9r: So which migrations to apply is determined by the migration tag in alembic, not by the state of the DB schema?

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: No longer necessary as alembic won't be packaged

src/zenml/zen_stores/migrations/README.md
AlexejPenner: We defined a release workflow that will give us the best of both worlds and resolve any branches: https://www.notion.so/zenml/Make-A-Release-adb5e415107e4189bbb894a4c08c00f1#8027d6d8331243f49cf24e3c0b9abaed

This will now create an empty revision at release time that gives us a version stamp matching our code version.

src/zenml/zen_stores/sql_zen_store.py
stefannica: Even better: we don't involve the `alembic.ini` file at all in the zenml db auto-migration.

src/zenml/zen_stores/migrations/README.md
stefannica: After more debating with Alexej, we found that alembic has a nice feature called "branches" that allow developers to add migration scripts in parallel, then merge them at a later point, if necessary. We made the alembic merge a mandatory step in the release process. This also allows us to "stamp" the state of the migration scripts with every release, so we can run things like "alembic downgrade 0.20.5" and "alembic upgrade 0.20.6" manually, if necessary.

stefannica (approved):

src/zenml/zen_stores/migrations/README.md
schustmi: That sounds great. I also think that database migration will be rather a sporadic event in the future and something we consider carefully instead of everyone simply modifying them on random branches ;)

schustmi (approved):

src/zenml/zen_stores/migrations/alembic.py
```diff
@@ -0,0 +1,236 @@
...
+    "ParentType",
+    "ParentContext",
+    "Context",
+    "Type",
+    "ContextProperty",
+    "sqlite_sequence",
+]
+
+
+def include_object(
```
fa9r, line 71: should this be named `exclude_object` instead?

fa9r: the docstring and function name seem to describe the opposite

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -547,7 +577,62 @@ def _initialize(self) -> None:
...
+                    "The ZenML database has more than one migration head "
+                    "revision. This is not expected and might indicate a "
+                    "database migration problem. Please raise an issue on "
+                    "GitHub if you encounter this."
+                )
+            # Case 3: the database has been migrated with alembic before. Just
+            # upgrade to the latest revision.
+            self.alembic.upgrade()
+        else:
+            if self.alembic.db_is_empty():
```
fa9r, line 625: the conditionals here could perhaps be flattened a bit via `elif` and/or early returns

fa9r (approved):
LGTM

PR #992, opened 2022-10-20
Further CLI optimization

bcdurak:
## Describe changes
I optimized the number of API calls when doing `zenml stack list`.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -404,7 +405,7 @@ def create_stack(
             request_model=CreateStackRequest,
         )
 
-    def get_stack(self, stack_id: UUID) -> StackModel:
+    def get_stack(self, stack_id: UUID, hydrated: bool = False) -> StackModel:
```
schustmi, line 408: Does `get_stack` also support the `hydrated` flag? If yes we should probably also include it on the `SqlZenStore`, correct?

stefannica: I agree, this is probably a left-over.

src/zenml/cli/utils.py
```diff
@@ -864,7 +864,7 @@ def get_stack_by_id_or_name_or_prefix(
         is_shared=True,
     )
 
-    named_stacks = user_only_stacks + shared_stacks
+    named_stacks = user_only_stacks + shared_stacks  # type: ignore[operator]
```
stefannica, line 867: As a general strategy, please try to not silence mypy this way. Instead, add some `assert isinstance(...)` statements after `list_stacks` calls

stefannica (changes requested):

stefannica (approved):

PR #994, opened 2022-10-20
Add Tolerations and NodeAffinity to Kubernetes executor

wefner:
## Describe changes
I implemented a `KubernetesOrchestratorSettings` to allow pods to use Node Affinity and Tolerations in order to choose the right nodes to run on.

> However, the `settings` will only be available for the first `step`. It won't update the manifest on subsequent `steps`.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-gallery/orchestrators/kubernetes.md
+++ b/docs/book/component-gallery/orchestrators/kubernetes.md
@@ -73,16 +73,32 @@ For additional configuration of the Kubernetes orchestrator, you can pass
 from zenml.integrations.kubernetes.flavors import KubernetesOrchestratorSettings
 
 kubernetes_settings = KubernetesOrchestratorSettings(
-    node_affinity=[{
-        "key": "node.kubernetes.io/name",
-        "operator": "In",
-        "values": ['my_powerful_node_group']
-        }],
-    tolerations=[{"key": "node.kubernetes.io/name",
-                  "operator": "Exists",
-                  "value": "training",
-                  "effect": "NoSchedule"
-      }])
+    affinity={
+        "nodeAffinity": {
+            "requiredDuringSchedulingIgnoredDuringExecution": {
+                "nodeSelectorTerms": [
+                    {
+                        "matchExpressions": [
+                            {
+                                "key": "node.kubernetes.io/name",
+                                "operator": "In",
+                                "values": ["my_powerful_node_group"],
+                            }
+                        ]
+                    }
+                ]
+            }
+        }
+    },
+    tolerations=[
+        {
+            "key": "node.kubernetes.io/name",
+            "operator": "Exists",
+            "value": "",
+            "effect": "NoSchedule",
+        }
+    ],
+)
 
 @pipeline(
     settings={
--- a/docs/book/component-gallery/orchestrators/kubernetes.md
+++ b/docs/book/component-gallery/orchestrators/kubernetes.md
@@ -66,8 +66,8 @@ python file_that_runs_a_zenml_pipeline.py
 For additional configuration of the Kubernetes orchestrator, you can pass
 `KubernetesOrchestratorSettings` which allows you to configure the following attributes:
 
-* `node_affinity`: NodeAffinity match expressions.
-* `tolerations`: Toleration settings for the pod.
+* `affinity`: Affinity constraints (`spec.affinity`) that can be defined to a pod.
+* `tolerations`: Toleration (`spec.tolerations`) so the pod can be scheduled.
 
 ```python
 from zenml.integrations.kubernetes.flavors import KubernetesOrchestratorSettings
--- a/docs/book/component-gallery/orchestrators/kubernetes.md
+++ b/docs/book/component-gallery/orchestrators/kubernetes.md
@@ -63,6 +63,35 @@ You can now run any ZenML pipeline using the Kubernetes orchestrator:
 python file_that_runs_a_zenml_pipeline.py
 ```
 
+For additional configuration of the Kubernetes orchestrator, you can pass
+`KubernetesOrchestratorSettings` which allows you to configure the following attributes:
+
+* `node_affinity`: NodeAffinity match expressions.
+* `tolerations`: Toleration settings for the pod.
+
+```python
+from zenml.integrations.kubernetes.flavors import KubernetesOrchestratorSettings
+
+kubernetes_settings = KubernetesOrchestratorSettings(
+    node_affinity=[{
+        "key": "node.kubernetes.io/name",
+        "operator": "In",
+        "values": ['my_powerful_node_group']
+        }],
+    tolerations=[{"key": "node.kubernetes.io/name",
+                  "operator": "Exists",
+                  "value": "training",
+                  "effect": "NoSchedule"
+      }])
+
+@pipeline(
+    settings={
+        "orchestrator.kubernetes": kubernetes_settings
+    }
+)
+  ...
+```
+
 A concrete example of using the Kubernetes orchestrator can be found 
 [here](https://github.com/zenml-io/zenml/tree/main/examples/kubernetes_orchestration).
 
--- a/src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
@@ -41,7 +41,7 @@ def build_pod_manifest(
         service_account_name: Optional name of a service account.
             Can be used to assign certain roles to a pod, e.g., to allow it to
             run Kubernetes commands from within the cluster.
-        settings: KubernetesOrchestratorSettings object
+        settings: `KubernetesOrchestratorSettings` object
 
     Returns:
         Pod manifest.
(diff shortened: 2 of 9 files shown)


schustmi:
Hi @wefner, thanks for the contribution!

The Kubernetes orchestrator is quite involved, but I'll try to explain why I think only the first pod gets the affinity/tolerations applied and how you could update it to make it apply to all pods.

The following steps happen when running a pipeline with the Kubernetes orchestrator:
- initially a single "orchestrator" pod is created. This pod is created inside the `KubernetesOrchestrator.prepare_or_run_pipeline(...)` method, and you've successfully applied the affinity and tolerations to it by passing the settings to `build_pod_manifest(...)`. **This pod however is only responsible for creating the subsequent pods which actually run the steps of the pipeline.**
- the orchestrator pod creates new pods for each step in the pipeline. This happens in `kubernetes_orchestrator_entrypoint.py` where `build_pod_manifest(...)` is called again to create the manifest for the pods running the actual steps. (https://github.com/zenml-io/zenml/blob/main/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py#L124-L131) You would have to extract the settings from the `deployment_config` and pass the settings to this call in order to apply the affinity and tolerations.

wefner:
Thank you @schustmi for the detailed explanation! 

I have updated it now to retrieve the settings from the orchestrator and the entrypoint. They are applied to all the pods.

docs/book/component-gallery/orchestrators/kubernetes.md
```diff
@@ -63,6 +63,35 @@ You can now run any ZenML pipeline using the Kubernetes orchestrator:
 python file_that_runs_a_zenml_pipeline.py
 ```
 
+For additional configuration of the Kubernetes orchestrator, you can pass
+`KubernetesOrchestratorSettings` which allows you to configure the following attributes:
+
+* `node_affinity`: NodeAffinity match expressions.
+* `tolerations`: Toleration settings for the pod.
```
fa9r, line 70: This could need a bit more detail on what those attributes actually do and why/when you would want to configure them.

docs/book/component-gallery/orchestrators/kubernetes.md
```diff
@@ -63,6 +63,35 @@ You can now run any ZenML pipeline using the Kubernetes orchestrator:
...
+    node_affinity=[{
+        "key": "node.kubernetes.io/name",
+        "operator": "In",
+        "values": ['my_powerful_node_group']
+        }],
+    tolerations=[{"key": "node.kubernetes.io/name",
+                  "operator": "Exists",
+                  "value": "training",
+                  "effect": "NoSchedule"
+      }])
```
fa9r, line 85: this isn't formatted correctly.

I'd suggest to copy it into some random file, run the format script, then copy it back.

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -74,9 +77,61 @@ def build_pod_manifest(
...
         spec = cast(Dict[str, Any], manifest["spec"])  # mypy stupid
         spec["serviceAccountName"] = service_account_name
+
+    if settings is not None:
+        spec = cast(Dict[str, Any], manifest["spec"])
+        spec.update(add_pod_settings(settings))
     return manifest
 
 
+def add_pod_settings(settings):
```
fa9r, line 87: these three functions need docstrings and type annotations

fa9r (approved):
Great PR, looks good to me from a conceptual level, just needs a little bit of cleanup before merging. Once all CI checks pass, this should be good to go!

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
fa9r: I just enabled workflows for your PR @wefner, these will give you more details on what's missing, see below :)

fa9r: To fix the spell-check issues, you will need to add `tolerations` into `.pyspelling-ignore-words`.

The other issues can be solved by wrapping class names in backticks (`), e.g. `KubernetesOrchestratorSettings`

schustmi:
I have two more points I'd like to discuss and potentially update, otherwise looks great!
- Following up on my explanation above (at first an "orchestrator" pod gets launched which then launches the pods for all steps): Do you think it makes sense to apply the tolerations and affinity to the orchestrator pod as well (this is what's currently happening), or should we only apply them to the pods running the actual steps?
- Users of the `KubernetesOrchestrator` can specify the settings you created either for a ZenML pipeline or a single ZenML step (this might be useful to target different nodes for specific steps). Right now you're getting the settings only from the pipeline and disregard any settings specified for specific steps: 
```python

for name_in_pipeline, step in deployment_config.steps.items(): 
    settings  = step.config.settings.get("orchestrator.kubernetes")  # if someone specified settings on the pipeline, they get automatically passed down to each step
```

wefner:
> Do you think it makes sense to apply the tolerations and affinity to the orchestrator pod as well (this is what's currently happening), or should we only apply them to the pods running the actual steps?

Good question. Yes, I do. Because this is exactly the point of having node affinity and tolerations. I think it makes sense to have "zenml workloads" in specific node groups so they don't interfere with general workloads or any other possible taints in the cluster. Although the orchestrator pod is small, it still adds to the pod count by node, preventing other workloads that won't use "zenml nodegroups" to be scheduled. Does this make sense?

> Users of the KubernetesOrchestrator can specify the settings you created either for a ZenML pipeline or a single ZenML step (this might be useful to target different nodes for specific steps). Right now you're getting the settings only from the pipeline and disregard any settings specified for specific steps:

Ah, good shout! I thought specific steps for `KubernetesOrchestrator` wasn't supported but I might have confused it with `ResourceSettings`.

--- 

On another note, I have been thinking of refactoring the interface of this Kubernetes Settings, and just passing whatever the user input is. 

[Example for Tolerations](https://github.com/zenml-io/zenml/pull/994/files#diff-ac5a803600177072a20eecdc3074ce345afb5840ae65beb75ff697141ce72022R98): This `add_pod_tolerations(...)` will unpack a list to create the same data structure. We can just pass `settings.tolerations`.

[Example for NodeAffinity](https://github.com/zenml-io/zenml/pull/994/files#diff-ac5a803600177072a20eecdc3074ce345afb5840ae65beb75ff697141ce72022R114): This will only support `requiredDuringSchedulingIgnoredDuringExecution `. 

We could just allow the user to pass any data structure they want according to Kubernetes `spec.<field>`.

In this way, we will not have to maintain our code to Kubernetes API versions.

Also, if we want to add more settings such as `resources`, `initContainers`, `volumes`, etc we won't need to create more functions to update the `spec`. We could even allow the users to pass their own manifest too. What do you think?

schustmi:
Alright yes that does make sense, let's keep it for the orchestrator pod as well then.

You're right, `ResourceSettings` haven't been implemented for the Kubernetes orchestrator yet. However that shouldn't interfere with the specification of the `KubernetesOrchestratorSettings` on a step level as I outlined above.

I just created a branch (still work in progress) which aims to solve something similar to your suggestion: https://github.com/zenml-io/zenml/compare/develop...feature/generic-kubernetes-settings

In there I created a generic `KubernetesPodSettings` object (which can/will be used in hopefully all of our orchestrators, e.g. Kubernetes, Kubeflow, Tekton, Vertex) which allows exactly what you just described. In addition, it can be used in code like this using the actual k8s objects:
```python
from kubernetes.client.models import V1Affinity

settings = KubernetesPodSettings(affinity=V1Affinity(...))
```

Let me know if this sounds good to you or if there's something missing that I should include in there.

wefner:
This sounds good to me. Having pod settings makes it more flexible 👍🏿 

I can move the settings to every step as suggested but besides this, wouldn't this work clash with your new branch? Your changes will supersede this PR if I understood correctly. Let me know if you still want this PR going through.

I have also found out that, although `KubernetesOrchestratorSettings` has default values for its attributes, if all of them are not passed to the orchestrator settings, it will raise an `AttributeError`. As an example, let's not add tolerations:

```
settings = {"orchestrator.kubernetes": KubernetesOrchestratorSettings(
      node_affinity=[{
        "key": "node.kubernetes.io/name",
        "operator": "In",
        "values": ['zenml']
        }])
}

```

```
│ /home/vscode/.local/lib/python3.9/site-packages/zenml/integrations/kubernetes/orchestrators/mani │
│ fest_utils.py:94 in add_pod_settings                                                             │
│    91 │   if getattr(settings, "node_affinity"):                                                 │
│    92 │   │   spec["affinity"] = add_pod_affinity(settings.node_affinity)                        │
│    93 │                                                                                          │
│ ❱  94 │   if settings.tolerations:                                                               │
│    95 │   │   # spec["tolerations"] = add_pod_tolerations(settings.tolerations)                  │
│    96 │   │   spec["tolerations"] = settings.tolerations                                         │
│    97                                                                                            │
│ /home/vscode/.local/lib/python3.9/site-packages/zenml/config/secret_reference_mixin.py:98 in     │
│ __custom_getattribute__                                                                          │
│    95 │   │   Returns:                                                                           │
│    96 │   │   │   The (potentially resolved) attribute value.                                    │
│    97 │   │   """                                                                                │
│ ❱  98 │   │   value = super().__getattribute__(key)                                              │
│    99 │   │                                                                                      │
│   100 │   │   if not secret_utils.is_secret_reference(value):                                    │
│   101 │   │   │   return value 
```

schustmi:
No I'd absolutely want to merge this PR in, you've did lots of the hard work of applying the settings, I'll just try to find a way to include my pod settings as part of the `KubernetesOrchestratorSettings` via inheritance or adding an attribute to it.

Regarding the other issue, I have a potential idea what is happening: When you're trying to get the settings from `deployment.pipeline.settings.get("orchestrator.kubernetes")` or `step.config.settings.get("orchestrator.kubernetes")` it returns you an instance of `BaseSettings`. To use the default values specified on the subclass (`KubernetesOrchestratorSettings` in your case), you need to convert the settings to this class, for example like this:
```python
settings = KubernetesOrchestratorSettings.parse_obj(step.config.settings.get("orchestrator.kubernetes"))
```

wefner:
Hi @schustmi @fa9r ,

this is the closest I got and it's working fine for me. I'm only concerned about `kubernetes_orchestrator_entrypoint.py` where I would have liked to get `self.get_settings(...)`  inherited from `StackComponent` but this solution also returns a `<class 'zenml.integrations.kubernetes.flavors.kubernetes_orchestrator_flavor.KubernetesOrchestratorSettings'>`

I updated the docs, lint, format, etc. Let me know if there's something else.

schustmi (approved):
Looks good!

PR #995, opened 2022-10-21
Fix `FileNotFoundError` with remote path in HuggingFace Dataset materializer

gabrielmbmb:
## Describe changes

The HuggingFace Dataset/DatasetDict materializer was raising a `FileNotFoundError` exception when passing a remote path such as `gs://artifacts/23/hf_datasets` or `s3://artifacts/23/hf_datasets`. This was caused because the remote path was passed to the `load_from_disk` function, but a specific `fs` was not being passed, so `datasets` code was treating the path as a local path.

## Pre-requisites

## Types of changes


gabrielmbmb:
Not sure if it is better to just pass the filesystem to the `datasets.load_from_disk` function or download the dataset to the local filesystem using the `io_utils.copy_dir` and then pass the local path to `datasets.load_from_disk`.

Either way, I think we should use the same approach in `handle_input` and `handle_return`, so, what do you think is better?

1. Use `io_utils` to download/upload the files of the dataset from/to the remote path
2. Use `fsspec` to create a filesystem handler than can be passed to `datasets.load_from_disk` and `Dataset.save_to_disk`

bcdurak:
Hey @gabrielmbmb,

Thank you for reporting this issue and you are absolutely right. For the time being, I would recommend the first solution. We are already using this method in some of our materializers such as the [KerasMaterializer](https://github.com/zenml-io/zenml/blob/main/src/zenml/integrations/tensorflow/materializers/keras_materializer.py). Perhaps, you can use it as an example.

bcdurak (approved):
Hey @gabrielmbmb,

Thank you once again for fixing this issue. I will merge the PR now and it is going to be a part of our next release.

PR #998, opened 2022-10-22
Fix GCP Artifact Store listdir empty path

safoinme:
## Describe changes
GCP listdir function returns an empty path, which causes an infinite loop! this fix adds a condition to the process to ignore that.

## Pre-requisites

## Types of changes


fa9r (approved):

safoinme:
Have someone tried to run the Kubeflow example on Azure before? I think it may have the same problem (considering the same error was happening with AWS and GCP and it's fixed in the same way) @htahir1 @schustmi

schustmi (approved):
LGTM! I'll check the azure artifact store for the same issue later

schustmi:
@safoinme Azure seems to work fine for me!

PR #1004, opened 2022-10-25
Delay user instruction in dockerfile generation

schustmi:
## Describe changes
This PR delays the `USER` instructions when generating dockerfiles and makes sure the specified user is the owner of the directory.

Adding the `USER` instruction early leads to some permission issues when installing pip packages on a system level. 
## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/pipeline_docker_image_builder.py
+++ b/src/zenml/utils/pipeline_docker_image_builder.py
@@ -444,9 +444,6 @@ class PipelineDockerImageBuilder:
         """
         lines = [f"FROM {parent_image}", f"WORKDIR {DOCKER_IMAGE_WORKDIR}"]
 
-        if docker_settings.user:
-            lines.append(f"USER {docker_settings.user}")
-
         if docker_settings.copy_global_config:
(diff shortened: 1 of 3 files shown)


src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -474,6 +471,10 @@ def _generate_zenml_pipeline_dockerfile(
 
         lines.append("RUN chmod -R a+rw .")
 
+        if docker_settings.user:
+            lines.append(f"USER {docker_settings.user}")
+            lines.append(f"RUN chown -R {docker_settings.user} .")
```
fa9r, line 476: Why is this necessary? Seems like this is the meat of the PR :D

schustmi: This user might not have the necessary permissions to install the `pip` packages in the above commands.
The `chown` command makes sure that the `/app` directory (which contains the user files/code) is owned by the provided user instead of `root`

fa9r (approved):

PR #1007, opened 2022-10-25
Introduce Permissions, Link Permissions to Roles, Restrict Access to endpoints based on Permission

AlexejPenner:
## Describe changes
Please find the Design Doc [here](https://www.notion.so/zenml/Roles-Permissions-Scopes-669b936335ae41009bcab1a786c9a59a)

In Scope:
* Introduce scoped endpoints enabling read-only access to the dashboard
* Apply scopes to users
* Make the transition unnoticeable to current users
* Minimal interference for new users
* Global scopes

Out of Scope:
* Project specific scopes
* UI representation of scopes

Tests before merge:

- [x] New installation, verify default user can call all entry points in local deployment
- [x] New installation, verify new users can call all entry points in Scenario local deployment
- [x] Existing db, verify default user can call all entry points in local deployment
- [x] Existing db, verify all users can call all entry points in  local deployment

## Pre-requisites

## Types of changes


src/zenml/zen_server/auth.py
```diff
@@ -207,3 +210,39 @@ def authentication_provider() -> Callable[..., AuthContext]:
...
+    roles = zen_store().list_role_assignments(
+        user_name_or_id=current_user.id,
+        project_name_or_id=None
+    )
+
+    for role in roles:
+        for permission in zen_store().get_role(role.role).permissions:
+            if permission == required_permission:
+                return True
+    raise NotAuthorizedError("Unauthorized")
```
fa9r, line 248: this is mostly duplicate code from above, I'd suggest define a helper method `user_has_permission(auth_context, permission)` and then just call `user_has_permission(auth_context, PermissionType.READ.value` here

src/zenml/zen_stores/schemas/user_management_schemas.py
```diff
@@ -187,14 +188,33 @@ def to_model(self) -> TeamModel:
         )
 
 
+class RolePermissionsSchema(SQLModel, table=True):
+    """SQL Model for team assignments."""
+
+    permission_id: int = Field(primary_key=True, foreign_key="permissionschema.id")
```
fa9r, line 194: why is permission ID `int` not `UUID`?

src/zenml/zen_stores/schemas/user_management_schemas.py
```diff
@@ -187,14 +188,33 @@ def to_model(self) -> TeamModel:
...
+    role_id: UUID = Field(primary_key=True, foreign_key="roleschema.id")
+
+
+class PermissionSchema(SQLModel, table=True):
+    """SQL Model for roles."""
+
+    id: int = Field(primary_key=True)
```
fa9r, line 201: same here, why int?

fa9r (commented):
Great PR, I really like the design, feels intuitive from a UX perspective and also relatively easy to use/maintain for us.

src/zenml/zen_server/routers/users_endpoints.py
```diff
@@ -66,7 +70,7 @@
 current_user_router = APIRouter(
     prefix=API + VERSION_1,
     tags=["users"],
-    dependencies=[Depends(authorize)],
+    dependencies=[Security(authorize, scopes=["read"])],
```
stefannica, line 73: Even users with no permissions should be able to fetch information about their account.

stefannica (changes requested):
Looks good for the most part. I particularly like the scoped token approach, it makes for a quick and clean implementation. I think what this needs is a couple of built-in default `admin` and `guest` roles that the user can quickly add to newly created users. The `guest` role should be the most restrictive (basically read-only) and should be assigned by default if a role isn't explicitly specified.

There doesn't seem to be a point to creating users that have no roles.

src/zenml/zen_server/auth.py
AlexejPenner: I moved away from this so no longer necessary

src/zenml/zen_stores/schemas/user_management_schemas.py
AlexejPenner: I would argue the permissions are maintained by us, the permissions_table could be seen as a db representation of an enum. I don't think uuid's would be necessary here

src/zenml/zen_server/routers/users_endpoints.py
AlexejPenner: Good point

src/zenml/zen_stores/migrations/versions/5994f9ad0489_introduce_role_permissions.py
```diff
@@ -0,0 +1,58 @@
...
+    sa.ForeignKeyConstraint(['permission_id'], ['roleschema.id'], ),
+    sa.ForeignKeyConstraint(['role_id'], ['permissionschema.id'], ),
+    sa.PrimaryKeyConstraint('permission_id', 'role_id')
+    )
+    # get metadata from current connection
+    meta = sa.MetaData(bind=op.get_bind())
+
+    # pass in tuple with tables we want to reflect, otherwise whole database will get reflected
+    meta.reflect(only=('permissionschema',))
+
```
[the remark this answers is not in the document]
AlexejPenner: As I understand it its a way of getting information about a tbale from the database in order to construct an incode representation similar to the SQLModels that we are using - you can find information about this here: https://docs.sqlalchemy.org/en/14/core/reflection.html

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -98,7 +99,20 @@ def token(
...
+
+    permissions = set().union(*[
+        zen_store().get_role(ra.role).permissions for ra in role_assignments
+    ])
+
+    access_token = JWTToken(
+        token_type=JWTTokenType.ACCESS_TOKEN,
+        user_id=auth_context.user.id,
+        permissions=permissions
+    ).encode()
```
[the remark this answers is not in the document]
AlexejPenner: Most of the logic was actually a method on the UserModel - as this accesses and has dependecies on the zen_store I didn't want this logic in the UserModel - i would not touch this until Baris' changes to the Models have made it into develop, maybe all the relevant information will be accessible on the new model

.gitignore
```diff
@@ -178,12 +178,14 @@ zenml_examples/
...
-
 # some examples have yamls; don't ignore them
 !examples/**/*.yaml
 
 # stack recipes
 test_terraform/
 
+# dashboard build files
+src/zenml/zen_server/dashboard/
+
```
stefannica, line 189: hehe :) we tried this before, but then the dashboard files are not included in the pypi package when we build it. I botched a release because of this.

src/zenml/zen_stores/migrations/versions/5994f9ad0489_introduce_role_permissions.py
```diff
@@ -0,0 +1,121 @@
...
+        sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("role_id", sqlmodel.sql.sqltypes.GUID(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["role_id"],
+            ["roleschema.id"],
+        ),
+        sa.PrimaryKeyConstraint("name", "role_id"),
+    )
+    # get metadata from current connection
+    meta = sa.MetaData(bind=op.get_bind())
```
stefannica, line 35: I didn't realize how painful it can be to populate the database without SQLModel. I thought you might use SQLModel here to add the default roles and update all existing users instead of alembic and raw sqlalchemy, but then I remembered that SQLModel schemas are a living thing that may change even more in the future.

stefannica (changes requested):
The changes look great, but you have to leave the dashboard alone :smile:

.gitignore
AlexejPenner: <iframe src="https://giphy.com/embed/afVwAjxEfqofu" width="480" height="323" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/reaction-the-office-no-afVwAjxEfqofu">via GIPHY</a></p>

AlexejPenner: moving it into .git/info/exclude locally and removing it here

stefannica (approved):

PR #1010, opened 2022-10-26
Docs for using CUDA-enabled docker images

strickvl:
I added a core page which describes how to update your pipeline / step definitions to use a parent base image with CUDA installed. I updated orchestrator and step operator pages to reflect this new addition.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/advanced-guide/pipelines/gpu-hardware.md
+++ b/docs/book/advanced-guide/pipelines/gpu-hardware.md
@@ -27,7 +27,7 @@ an example, if you wanted to use the latest CUDA-enabled official PyTorch image
 for your entire pipeline run, you could include the following code:
 
 ```python
-docker_settings = DockerSettings(parent_image="pytorch/pytorch:latest")
+docker_settings = DockerSettings(parent_image="pytorch/pytorch:1.12.1-cuda11.3-cudnn8-runtime")
 
 @pipeline(settings={"docker": docker_settings})
 def my_pipeline(...):
@@ -42,7 +42,7 @@ be installed. There are lots of ways to specify this, but as one example, you
 could do the following (updating the code from above):
 
 ```python
-docker_settings = DockerSettings(parent_image="pytorch/pytorch:latest", requirements=["zenml==0.20.5", "torchvision"])
+docker_settings = DockerSettings(parent_image="pytorch/pytorch:1.12.1-cuda11.3-cudnn8-runtime", requirements=["zenml==0.20.5", "torchvision"])
 
 @pipeline(settings={"docker": docker_settings})
 def my_pipeline(...):
--- a/docs/book/advanced-guide/pipelines/gpu-hardware.md
+++ b/docs/book/advanced-guide/pipelines/gpu-hardware.md
@@ -49,9 +49,8 @@ def my_pipeline(...):
     ...
 ```
 
-Adding these two extra settings options and applying them to whatever resource
-is appropriate for your pipelines and use cases will be enough to ensure that
-CUDA is enabled for those specific steps. Note that these configuration changes
+Adding these two extra settings options will be enough to ensure that
+CUDA is enabled for the specific steps that require GPU acceleration. Note that these configuration changes
 are **required** for the GPU hardware to be properly utilized. If you don't
 update the settings, your steps might run but they will not see any boost in
 performance from the custom hardware.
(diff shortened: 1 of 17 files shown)


docs/book/advanced-guide/pipelines/gpu-hardware.md
```diff
@@ -0,0 +1,57 @@
...
+```python
+docker_settings = DockerSettings(parent_image="pytorch/pytorch:latest", requirements=["zenml==0.20.5", "torchvision"])
+
+@pipeline(settings={"docker": docker_settings})
+def my_pipeline(...):
+    ...
+```
+
+Adding these two extra settings options and applying them to whatever resource
+is appropriate for your pipelines and use cases will be enough to ensure that
```
fa9r, line 53: somewhat hard to read sentence

docs/book/advanced-guide/pipelines/gpu-hardware.md
```diff
@@ -0,0 +1,57 @@
...
+
+```python
+docker_settings = DockerSettings(parent_image="pytorch/pytorch:latest")
+
+@pipeline(settings={"docker": docker_settings})
+def my_pipeline(...):
+    ...
+```
+
+2. Add ZenML as an explicit pip requirement
```
fa9r, line 37: why is this necessary, don't we do this under the hood when building images? Or, asked differently, why don't we need to do this when running on CPU?

fa9r (approved):
LGTM!

docs/book/advanced-guide/pipelines/gpu-hardware.md
strickvl: Not entirely sure, but this is what @schustmi told me this morning. Maybe he has the answer. I think step #2 is just precautionary and perhaps not essential?

docs/book/advanced-guide/pipelines/gpu-hardware.md
strickvl: Fixed now.

docs/book/advanced-guide/pipelines/gpu-hardware.md
htahir1: You always need ZenML installed in every case, because the code that runs inside the docker image is an entrypoint file which is inside ZenML.That code actually kicks off the pipeline.

In the `CPU` case, you dont specify a parent_image @fa9r, so you the default zenml base image is used which has zenml already installed. With this current approach, a pytorch base image with cuda is being used and of course that means zenml needs to be installed. So its quite mandatory

docs/book/advanced-guide/pipelines/gpu-hardware.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. Specify a CUDA-enabled parent image in your `DockerSettings`
+
+For full details, see the whole section where we explain how to do this [on the
+containerization page](../../advanced-guide/pipelines/containerization.md). As
+an example, if you wanted to use the latest CUDA-enabled official PyTorch image
+for your entire pipeline run, you could include the following code:
+
+```python
+docker_settings = DockerSettings(parent_image="pytorch/pytorch:latest")
```
htahir1, line 30: Are you sure `pytorch/pytorch:latest` has CUDA installed? Maybe specify a CUDA-specific image to make this very explicit.

htahir1 (changes requested):
Cool - I think Im hoping for a more in-depth analysis of this CUDA business and I left my thoughts in the review. One thing I would like to see on top of my comments is the changing of `examples` of step operators to include the methodology mentioned

docs/book/advanced-guide/pipelines/gpu-hardware.md
strickvl: I am sure, but I can specify something directly so that it's obvious maybe.

docs/book/advanced-guide/pipelines/gpu-hardware.md
```diff
@@ -0,0 +1,78 @@
...
+The core cloud operators all offer prebuilt Docker images that fit with their
+hardware. You can find more information on them here:
+
+- [AWS](https://github.com/aws/deep-learning-containers/blob/master/available_images.md)
+- [GCP](https://cloud.google.com/deep-learning-vm/docs/images)
+- [Azure](https://learn.microsoft.com/en-us/azure/machine-learning/concept-prebuilt-docker-images-inference)
+
+Not all of these images are available on DockerHub, so your orchestrator node
+would have to have permission, for example, to pull images from registries if
+you were to use one of those.
```
htahir1, line 78:
```suggestion
Not all of these images are available on DockerHub, so your please ensure
that the orchestrator environment your pipeline runs in have to have sufficient 
permission to pull images from registries if you were to use one of those.
```

htahir1 (changes requested):
Nice, thanks for the quick turnaround! The examples update is still left and i left one more suggestion

PR #1017, opened 2022-10-27
Adding zenserver integration to stack recipe CLI 

wjayesh:
## Describe changes
I made the following changes to the stack recipe CLI to make it work with the ZenServer. 

- on running deploy, the CLI now checks if the user has a remote ZenServer configured. If not, then the CLI goes ahead to create one on the same cloud as the recipe and using outputs from the recipe such as the ingress controller hostname.
- there's a flag `--no-server` that can be passed to avoid creation of zen server even if there is none present.
- some recipes don't create a Kubernetes cluster and there's a warning message with a link to the docs that can help people deploy ZenML in that scenario.

### Some minor changes
- `create_sql` is now a general option across all clouds. This makes it easy to deploy the server through the recipe CLI.

This PR goes in tandem with [this PR](https://github.com/zenml-io/mlops-stacks/pull/21) on the mlops-stacks repository.

## Pre-requisites

## Types of changes


src/zenml/cli/stack_recipes.py
```diff
@@ -867,33 +874,81 @@ def deploy(
...
+                        vars = stack_recipe_service.get_vars()
+                        filter = [
+                            "aws-stores-minimal",
+                            "azureml-minimal",
+                            "vertex-ai",
+                        ]
+                        if Path(
+                            stack_recipe_service.terraform_client.working_dir
+                        ).name in filter and (
+                            "enable_mlflow" not in vars
```
schustmi, line 908: I'm not sure I understand why we care if MLflow is enabled/disabled here?

src/zenml/cli/stack_recipes.py
```diff
@@ -905,45 +960,52 @@ def deploy(
...
 
+def zen_server_exists() -> bool:
+    """Check if a remote ZenServer is active.
+
+    Returns:
+        True if active, false otherwise.
+    """
+    gc = GlobalConfiguration()
+
+    return gc.store and gc.store.type == StoreType.REST
```
schustmi, line 1006: I think this only checks if a ZenServer is used, but also returns True for a local (daemon or Docker) ZenServer.
```python
GlobalConfiguration().zen_store.is_local_store()
```
instead would give the correct result

schustmi (changes requested):

src/zenml/zen_server/deploy/terraform/terraform_zen_server.py
```diff
@@ -99,6 +99,7 @@ class TerraformServerDeploymentConfig(ServerDeploymentConfig):
         ingress_controller_hostname: The ingress controller hostname to use for
             the ingress self-signed certificate and to compute the ZenML server
             URL.
+        create_sql: Whether to create a SQL database as part of the recipe.
```
stefannica, line 102: nit: the name and description are a bit misleading. You're not creating a SQL database, you're deploying a SQL database service. `deploy_db` would be a better name.

src/zenml/zen_server/deploy/terraform/recipes/gcp/zen_server.tf
```diff
@@ -47,7 +47,7 @@ resource "helm_release" "zen-server" {
   }
   set {
     name = "ingress.host"
-    value = var.create_ingress_controller? "${data.kubernetes_service.ingress-controller[0].status.0.load_balancer.0.ingress.0.ip}.nip.io" : var.ingress_controller_hostname
+    value = var.create_ingress_controller? "${data.kubernetes_service.ingress-controller[0].status.0.load_balancer.0.ingress.0.ip}.nip.io" : "${var.ingress_controller_hostname}.nip.io"
```
stefannica, line 50: I didn't notice this before, but thought I should mention it now: .nip.io only works with IP addresses, but you're using the ingress hostname here, which may or may not be an IP address.

src/zenml/cli/stack_recipes.py
wjayesh: Only if mlflow is enabled on the recipe side, a Kubernetes cluster gets created to host it. Otherwise, no cluster and therefore we can't do a `zenml deploy` right away.

src/zenml/zen_server/deploy/terraform/recipes/gcp/zen_server.tf
wjayesh: You're right but I had noticed that in the case of GKE and AKS, the ingress controller host is an IP address and so I have added this suffix only to those two recipes.

schustmi (approved):

stefannica (approved):

PR #1022, opened 2022-10-27
Update link checker configs for faster, more accurate checks

dnth:
## Describe changes
This PR adds 2 things -

1. Update `markdown_check_config.json` to consider certain HTTP codes as alive. They are specified in these lines in the file.

```
"aliveStatusCodes": [
    999,
    403,
    503,
    200,
    0
  ],
```

Also added a few other configs for more accurate checks like 

```
"timeout": "20s",
"retryOn429": true,
"retryCount": 5,
"fallbackRetryDelay": "30s"
```

The description of each parameter is as follows -

image

2. Adds configs in `markdown-link-check.yml` to only check in specific folders/files instead of checking the entire repo -

```
folder-path: './examples, ./docs/book'

file-path: './README.md, ./LICENSE, ./RELEASE_NOTES.md, CODE-OF-CONDUCT.md, CONTRIBUTING.md, CLA.md, RELEASE_NOTES.md, ROADMAP.md'
```

These 2 additions should reduce the time it takes to check links and reduce false positives in the results.

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -14,5 +14,7 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'no'
+        folder-path: './examples, ./docs/book'
+        file-path: './README.md, ./LICENSE, ./RELEASE_NOTES.md, CODE-OF-CONDUCT.md, CONTRIBUTING.md, CLA.md, RELEASE_NOTES.md, ROADMAP.md'
         config-file: .github/workflows/markdown_check_config.json
       continue-on-error: true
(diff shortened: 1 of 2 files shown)


.github/workflows/markdown-link-check.yml
```diff
@@ -14,5 +14,7 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'no'
+        folder-path: './examples, ./docs/book'
```
fa9r, line 17: don't we need to check src/ as well? We might have log messages or similar that point to old links. Or can't we detect those anyways?

dnth: Absolutely! This tool now only checks markdown files. So if the links are in a `.py` file it will miss it. But I noted there are also a couple of `.md` files in src/ 

I will add the src/ folder in :)

PR #1025, opened 2022-10-28
Unify run names across orchestrators

schustmi:
## Describe changes

This PR addresses the inconsistency in run names using different orchestrators.

Previously, we used run names returned by the respective orchestrators as the names for ZenML pipeline runs.
This means that run names were varying from orchestrator to orchestrator, often not even containing the pipeline name.
Furthermore, we were often ignoring the explicit run name given by the user when calling `pipeline.run(run_name="something")`.

This PR makes sure that we have consistent run names across all orchestrators and respect the configured run name.
Additionally, the `run_name` can now be specified as a template like this: `something-{date}-{time}` and ZenML will replace the placeholders when the actual run gets started. This is especially useful for scheduled runs as well as restarting/rerunning a run from the orchestrator UI.

## Pre-requisites

## Types of changes


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -208,6 +213,20 @@ def config(self) -> BaseOrchestratorConfig:
         """
         return cast(BaseOrchestratorConfig, self._config)
 
+    @abstractmethod
+    def get_orchestrator_run_id(self) -> str:
```
fa9r, line 217: does this even need to be abstract? How about we just define a single env variable that we set for all orchestrators?

src/zenml/zen_stores/migrations/versions/d02b3d3464cf_add_orchestrator_run_id_column.py
```diff
@@ -0,0 +1,40 @@
...
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("pipelinerunschema", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "orchestrator_run_id",
+                sqlmodel.sql.sqltypes.AutoString(),
+                nullable=False,
```
fa9r: This breaks existing DBs if nullable is False. I would propose to simply allow it to be None for backwards compatibility. If you really want to enforce that it is not None for all entries, you need to a migration in three steps: 1) add column with nullable True 2) run SQL command to define values for all null columns 3) update column to set nullable False

fa9r: See how we added status here for example: https://github.com/zenml-io/zenml/pull/977/files#diff-a276a0accba80e17b01aa2f049ee0affeed819b35743017dbfbc23fd931df277

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -405,6 +423,70 @@ def _cleanup_run(self) -> None:
         self._active_deployment = None
         self._active_pb2_pipeline = None
 
+    def get_run_id_for_orchestrator_run_id(
```
fa9r, line 423: This could move to general utils perhaps

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -405,6 +423,70 @@ def _cleanup_run(self) -> None:
...
+        self, orchestrator_run_id: str
+    ) -> UUID:
+        run_id_seed = f"{self.id}-{orchestrator_run_id}"
+
+        m = hashlib.md5()
```
fa9r: `m` not the best variable name :D

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -292,12 +308,13 @@ def run_step(
         step_name = step.config.name
         pb2_pipeline = self._active_pb2_pipeline
 
-        run_name = run_name or self._active_deployment.run_name
+        run = self._create_or_reuse_run()
```
fa9r, line 308: `run` shadows a built-in name, I've used `run_model` everywhere for that reason

fa9r (changes requested):
Great PR, finally we have consistent run names!

Unfortunately, the DB migration code will not work as is, see comment for details.

Apart from that, the PR looks good to me!

src/zenml/orchestrators/base_orchestrator.py
schustmi: Environment variables are not the only way to do this, see e.g. non-docker based orchestrators like airflow or the local orchestrator. I also think having it abstract makes it much more visible than silently requiring an environment variable.

src/zenml/zen_stores/migrations/versions/d02b3d3464cf_add_orchestrator_run_id_column.py
schustmi: Actually this is outdated already, I've changed it to be `nullable=True`. Do I still need a three step approach in this case, or can I leave the migration unchanged otherwise?

src/zenml/orchestrators/base_orchestrator.py
schustmi: Already renamed ;)

src/zenml/orchestrators/base_orchestrator.py
schustmi: I'll rename it

src/zenml/orchestrators/base_orchestrator.py
schustmi: I had it in `orchestrators/utils.py` before, but had to import it in many orchestrator files then. Would there be any explicit upside to having it in some utils file?

src/zenml/orchestrators/base_orchestrator.py
fa9r: Hmm yeah same on my side, sorry for the confusion then 😅

src/zenml/zen_stores/migrations/versions/d02b3d3464cf_add_orchestrator_run_id_column.py
fa9r: In that case you can leave it unchanged 👍

src/zenml/orchestrators/base_orchestrator.py
fa9r: I was thinking this could become a general util function `generate_uuid_from_str(str)` or similar

fa9r: but it's not that important

fa9r (approved):

src/zenml/orchestrators/base_orchestrator.py
schustmi: Ah yeah that's a good idea, I'll still keep this method in the `BaseOrchestrator` but move the logic outside!

schustmi: Move it into `uuid_utils`

PR #1029, opened 2022-11-02
Before deleting the global zen_server files, spin it down

AlexejPenner:
## Describe changes
To avoid leaving a dangeling process that needs to be manually killed.

## Pre-requisites

## Types of changes


fa9r (commented):
Should we perhaps fix this in the server code instead? Otherwise you can still get this behaviour  if you manually remove your global config without calling `down` first

AlexejPenner:
> Should we perhaps fix this in the server code instead? Otherwise you can still get this behaviour if you manually remove your global config without calling `down` first

I think that is not trivial. If someone manually deletes the .config/zenml/zen_server/local folder, the service.pid file is deleted. Either the server would need to have a listener to kill itself when those files are deleted, or when you try spinning up a new zen_server at the same port, you would somehow figure out its a zen_server blocking the port, and then you would need to kill it by pid, which might not be desired by the user.

stefannica (approved):

fa9r (approved):

PR #1033, opened 2022-11-03
Allow non-local Label Studio instances for annotation

strickvl:
I updated the label studio annotator stack component to allow for deployed (i.e. non-local) instances of Label Studio to be used in stacks / ZenML.

## Pre-requisites

## Types of changes


fa9r (approved):

htahir1 (commented):
Cool, easy win. Just one comment

src/zenml/integrations/label_studio/flavors/label_studio_annotator_flavor.py
```diff
@@ -26,7 +26,8 @@
     from zenml.integrations.label_studio.annotators import LabelStudioAnnotator
 
 
-DEFAULT_LABEL_STUDIO_PORT = 8093
+DEFAULT_LOCAL_INSTANCE_URL = "https://localhost"
```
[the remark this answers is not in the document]
strickvl: You're totally right. Good catch!

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -87,7 +88,10 @@ def get_url(self) -> str:
         Returns:
             The URL of the annotation interface.
         """
-        return f"http://localhost:{self.config.port}"
+        if self.config.instance_url == DEFAULT_LOCAL_INSTANCE_URL:
```
htahir1, line 91: Why do we need this condition? Isn't it okay just to return self.config.instance_url?

strickvl: Not quite, but you were right that there was a simpler way to do this. I encountered some errors when testing it out on a deployed instance earlier, but I realised what I was doing wrong. I removed the conditional now and updated the docs to make it clearer what the user needs to do.

schustmi (approved):

PR #1036, opened 2022-11-03
Feature/eng 1402 consolidate stack sharing

AlexejPenner:
## Describe changes
WIP

## Pre-requisites

## Types of changes


htahir1 (changes requested):
IM a bit unsure bout my review, i.e., im not sure if i understood your logic too much. IMO people should  be guided to using the `-r` option a bit more in the messages but i think you might have intentionally not asked this of them? Anyway left some comments

htahir1 (approved):

src/zenml/cli/stack.py
```diff
@@ -447,13 +465,21 @@ def update_stack(
...
+@click.option(
+    "--recursive",
+    "-r",
+    "recursive",
+    is_flag=True,
+    help="Recursively also share all stack components if they are private.",
+)
 def share_stack(
-    stack_name_or_id: Optional[str],
+    stack_name_or_id: Optional[str], recursive: bool = False
```
fa9r, line 476: This changes the default behavior from sharing to not sharing, is that intended?

src/zenml/cli/stack.py
```diff
@@ -422,13 +431,22 @@ def update_stack(
...
+                    client=client,
+                    component_type=c_type,
+                    id_or_name_or_prefix=name,
+                )
+                if stack_to_update.is_shared and not component.is_shared:
+                    cli_utils.error(
+                        "You attempted to add a private component"
+                        f"{c_type}:{name} to a shared stack. This "
+                        f"is not supported. Set the {c_type} to"
+                        f"shared like this and then re-share your stack: "
```
fa9r, line 444: you don't need to reshare the stack but rather run the command again right?

src/zenml/cli/stack.py
```diff
@@ -230,13 +230,22 @@ def register_stack(
...
+                    id_or_name_or_prefix=name,
+                )
+                stack_components[c_type] = [component.id]
+                if share:
+                    if not component.is_shared:
+                        cli_utils.error(
+                            "You attempted to create a shared stack containing "
+                            f"a private component {c_type}: {name}. This "
+                            f"is not supported. Set the {c_type} to "
+                            f"shared like this and then try re-sharing your "
```
fa9r, line 245: same as below, you shouldn't reshare your stack but rather run the register command again I guess

fa9r (approved):
Code looks great and seems well tested.

This changes the default behavior of `zenml stack share` though, is that intended?

src/zenml/cli/stack.py
```diff
@@ -1213,10 +1250,18 @@ def copy_stack(
...
+            share = False
+
+        stack_to_copy.is_shared = share
+
         from zenml.models import StackModel
 
         copied_stack = StackModel.parse_obj(
-            stack_to_copy.dict(exclude={"id", "created", "updated"})
+            stack_to_copy.dict(
+                exclude={"id", "created", "updated", "is_shared"}
```
schustmi, line 1263: Does this not overwrite the `is_shared` that you set just a few lines above?

schustmi (changes requested):

src/zenml/cli/stack.py
AlexejPenner, line 1263:
```suggestion
            stack_to_copy.dict(
                exclude={"id", "created", "updated"}
```

PR #1041, opened 2022-11-04
Fix the pipeline run sync on sqlite and the --blocking zenml server deployment

stefannica:
## Describe changes
This PR fixes two issues:

- the recently introduced mechanism that uses database locking ("select for update") to synchronize pipeline runs was not applicable to SQLite databases
- the `zenml up --blocking` command did not use the local database and instead ended up creating a new one

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -18,6 +18,7 @@ import os
 import re
 from datetime import datetime, timedelta
 from pathlib import Path, PurePath
+from threading import Lock
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -528,6 +529,8 @@ class SqlZenStore(BaseZenStore):
         CONFIG_TYPE: The type of the store configuration.
         _engine: The SQLAlchemy engine.
         _metadata_store: The metadata store.
+        _lock: A thread mutex used to ensure thread safety during the pipeline
+            run synchronization.
     """
 
     config: SqlZenStoreConfiguration
@@ -539,6 +542,7 @@ class SqlZenStore(BaseZenStore):
     _metadata_store: Optional["MetadataStore"] = None
     _highest_synced_run_mlmd_id: int = 0
     _alembic: Optional[Alembic] = None
+    _sync_lock: Optional[Lock] = None
 
     @property
     def engine(self) -> Engine:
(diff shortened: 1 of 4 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3568,16 +3588,30 @@ def _sync_runs(self) -> None:
...
-                # can lock to prevent multiple processes from syncing runs at
-                # the same time.
-                logger.debug("Syncing pipeline runs...")
-                session.query(AlembicVersion).with_for_update().all()
-                self._sync_runs_with_lock(session)
-                logger.debug("Pipeline runs sync complete")
+            # This is to synchronize the locally running threads so that only
+            # one thread attempts to sync the runs at any given time. The
+            # timeout is to ensure that we don't block the threads for too long.
+            if self.sync_lock.acquire(timeout=30):
```
fa9r, line 3594: Can we make the timeout shorter? Perhaps this could also improve our performance issues since if there are a lot of requests trying to sync at the same time this should then only allow one of them to do the actual sync and skip syncing for the others, right?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3568,16 +3588,30 @@ def _sync_runs(self) -> None:
...
+                with Session(self.engine) as session:
+                    logger.debug("Syncing pipeline runs...")
+                    if self.config.driver != SQLDatabaseDriver.SQLITE:
+                        # This is to synchronize all server processes trying to
+                        # sync the pipeline runs at the same time. We use the
+                        # alembic version table as a shared resource that we can
+                        # lock to prevent multiple processes from syncing runs
+                        # at the same time.
+                        session.query(AlembicVersion).with_for_update().all()
```
fa9r, line 3603: Do we even still need this code now that we lock it using `threading.Lock`?

stefannica: As the comment says, this is needed you're using multiple replicas in your server (multiple zenml server instance clones connected to the same database).

src/zenml/zen_stores/sql_zen_store.py
stefannica: yes, good idea, I will make it 20 seconds. That way, the client won't have time to timeout (default client timeout is 30 seconds now)

fa9r (approved):
LGTM!

PR #1045, opened 2022-11-06
Fix bug when running non-local annotator instance.

sheikhomar:
## Describe changes
`zenml stack up` attempts to start a local Label Studio instance even though the annotator is configured to use a non-local instance. I fixed it by checking if the instance URL is the same as `DEFAULT_LOCAL_INSTANCE_URL` before starting/stopping local annotator daemon.

## Pre-requisites

## Types of changes


stefannica (changes requested):
@sheikhomar many thanks for fixing this ! Could you please also patch the `is_running` method in that same file ? Otherwise it looks good.

strickvl:
This looks good to me! Thanks for making this PR, @sheikhomar! I'll get it merged in.

stefannica (approved):

PR #1050, opened 2022-11-07
Fix Null Step Configuration/Parameters Error

fa9r:
## Describe changes

Some users faced errors when writing steps to our database with `None` step configuration or parameters. This was caused by too large configurations/parameters that exceed the 4K character limit after encoding, which SQLModel apparently silently set to `None` then. To fix this, I changed the type of those columns to TEXT.

While at it, I also adjusted some code in the metadata store to ensure those parameters can never be set to `None`, no matter what is written in MLMD.

## Pre-requisites

## Types of changes


schustmi (approved):

strickvl (commented):
This generally looks good to me. I'm wondering whether something like [this](https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?redirectedfrom=MSDN&view=sql-server-ver16) is going to bite us at some point in the future if we use the `text` type too liberally?:

> "The ntext, text, and image data types will be removed in a future version of SQL Server. Avoid using these data types in new development work, and plan to modify applications that currently use them. Use nvarchar(max), varchar(max), and varbinary(max) instead."

src/zenml/zen_stores/migrations/versions/7280c14811d6_use_text_type.py
```diff
@@ -0,0 +1,111 @@
...
+            existing_type=sa.TEXT(),
+            type_=sa.VARCHAR(length=4096),
+            existing_nullable=False,
+        )
+
+    with op.batch_alter_table("pipelineschema", schema=None) as batch_op:
+        batch_op.alter_column(
+            "spec",
+            existing_type=sa.TEXT(),
+            type_=sa.VARCHAR(length=4096),
```
stefannica, line 93: Not that it really matters, but will existing entries be truncated or set to null ?

stefannica (approved):

fa9r:
> This generally looks good to me. I'm wondering whether something like [this](https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?redirectedfrom=MSDN&view=sql-server-ver16) is going to bite us at some point in the future if we use the `text` type too liberally?:

I think it should be fine since we are using SQLAlchemy's TEXT, which is translated to different types depending on the underlying engine, so I would expect that TEXT will simply get translated to another data type (perhaps `VarChar(65535)`?) once the TEXT type no longer exists in MySQL.

strickvl (approved):

src/zenml/zen_stores/migrations/versions/7280c14811d6_use_text_type.py
fa9r: Interestingly, on SQLite nothing happens, once you change to TEXT the downgrade will not change the limit back. I guess this is because SQLite doesn't enforce any length restrictions anyways so using type TEXT/VarChar work with arbitrary length strings there.

As for MySQL, I didn't try since I don't even know how I could downgrade it within k8s and it's probably not worth the effort to investigate this :D

PR #1055, opened 2022-11-08
Sync Speedup

fa9r:
## Describe changes
I made various changes to the sync logic to speed it up:
- I reintroduced the logic to run the sync asynchronously on the server. To prevent missing data during post execution, the RestZenStore explicitly invokes a sync before each GET method. This way, the sync is called on every GET that a user issues, but not on requests from the UI.
- Runs with status=RUNNING that are older than 1 week will no longer be synced. This is to prevent syncing old runs from zenml<0.21.0 that were incorrectly left in RUNNING state after being keyboard-interrupted.
- The status of a step is now only updated if all of its output artifacts have been written. This allows us to ignore all finished runs in future sync calls. To do so, `StepRunModel.num_outputs` was introduced.
- The status of a run/step is now only updated as part of the sync. All other methods that were calling the status update directly (`get_run()` and `get_run_step()`) now also call `_sync_run()` like all other GET methods.
- Output artifacts and input artifacts are now queried separately from the metadata store to prevent unnecessary recursive calls.
- Unnecessary `get_run()` calls in the server/client were removed.

## Known Concurrency Issues
Running pipelines in parallel (e.g. using multi-threading) can still cause several issues unrelated to syncing:
- `ml_metadata.errors.UnknownError: Stream removed` raised in TFX launcher
-  running the same pipeline in parallel with caching enabled might set `step.parent_step_ids` incorrectly. This is because our metadata store code defines a parent step of a step as the last step that produced one of its input artifacts, which might not be correct in case of concurrency.

## Pre-requisites

## Types of changes


src/zenml/zen_server/routers/metadata_sync_endpoints.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@router.get(
+    "",
+    response_model=None,
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+@handle_exceptions
+def sync_runs(
+    _: AuthContext = Security(authorize, scopes=[PermissionType.WRITE])
```
AlexejPenner, line 37: You scoped this to need WRITE permissions, what happens if a user with read permissions pings this through the rest_zen_store?

src/zenml/zen_stores/migrations/versions/8a64fbfecda0_add_num_outputs_to_run_step.py
```diff
@@ -0,0 +1,45 @@
...
+    """Upgrade database schema and/or data, creating a new revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("pipeline_run", schema=None) as batch_op:
+        batch_op.alter_column(
+            "num_steps", existing_type=sa.INTEGER(), nullable=True
+        )
+
+    with op.batch_alter_table("step_run", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("num_outputs", sa.Integer(), nullable=True)
```
AlexejPenner, line 28: are sa.Integer and sa.INTEGER the same thing?

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -2028,10 +2034,5 @@ def _delete_resource(
...
     def _sync_runs(self) -> None:
-        """Syncs runs from MLMD.
-
-        Raises:
-            NotImplementedError: This internal method may not be called on a
-                `RestZenStore`.
-        """
-        raise NotImplementedError
+        """Syncs runs from MLMD."""
+        self.get(METADATA_SYNC)
```
AlexejPenner, line 2038: I think we should have some error handling on this, if it fails, we wouldn't want to throw an error for the user, the user will be helpless anyways, no need to bother them?

AlexejPenner (commented):
Congrats on getting this over with

stefannica (approved):
You even added back the helm configuration option, thank you !

src/zenml/zen_stores/rest_zen_store.py
fa9r: Good idea, I added error handling inside the endpoint so we only log an exception to the server logs instead of raising it

src/zenml/zen_stores/migrations/versions/8a64fbfecda0_add_num_outputs_to_run_step.py
fa9r: Great catch, they are not the same, `sa.INTEGER` will fail if the underlying DB doesn't have integer as data types. TL;DR: Always use CamelCase types, never UPPERCASE, see https://docs.sqlalchemy.org/en/14/core/type_basics.html#the-camelcase-datatypes

src/zenml/zen_server/routers/metadata_sync_endpoints.py
fa9r: You're right, it would fail which is unexpected behaviour

src/zenml/zen_stores/migrations/versions/8a64fbfecda0_add_num_outputs_to_run_step.py
```diff
@@ -0,0 +1,45 @@
...
+Revises: 5330ba58bf20
+Create Date: 2022-11-08 16:20:35.241562
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "8a64fbfecda0"
+down_revision = "5330ba58bf20"  # 0.21.1
```
AlexejPenner, line 13:
```suggestion
down_revision = "5330ba58bf20"
```

Why the comment here?

AlexejPenner (approved):

src/zenml/zen_server/routers/metadata_sync_endpoints.py
fa9r: Changed it to READ

PR #1056, opened 2022-11-08
Fix role permission migration

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/migrations/versions/5994f9ad0489_introduce_role_permissions.py
+++ b/src/zenml/zen_stores/migrations/versions/5994f9ad0489_introduce_role_permissions.py
@@ -50,24 +50,24 @@ def upgrade() -> None:
     conn.execute(
         sa.text(
             """
-            DELETE FROM userroleassignmentschema
-            WHERE role_id
-            IN (SELECT id FROM roleschema WHERE name=='admin' OR name=='guest')
+            DELETE U FROM userroleassignmentschema U
+            LEFT JOIN roleschema R ON U.role_id = R.id
+            WHERE R.name IN ('guest', 'admin');
             """
         )
     )
     conn.execute(
         sa.text(
             """
-            DELETE FROM teamroleassignmentschema
-            WHERE role_id
-            IN (SELECT id FROM roleschema WHERE name=='admin' OR name=='guest')
+            DELETE T FROM teamroleassignmentschema T
+            LEFT JOIN roleschema R ON T.role_id = R.id
+            WHERE R.name IN ('guest', 'admin');
             """
         )
     )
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/migrations/versions/5994f9ad0489_introduce_role_permissions.py
```diff
@@ -50,24 +50,24 @@ def upgrade() -> None:
     conn.execute(
         sa.text(
             """
-            DELETE FROM userroleassignmentschema
-            WHERE role_id
-            IN (SELECT id FROM roleschema WHERE name=='admin' OR name=='guest')
+            DELETE U FROM userroleassignmentschema U
+            LEFT JOIN roleschema R ON U.role_id = R.id
```
fa9r, line 54: Why do you need a left join here? `role_id` should never be `null`, and even if it were, those entries would be filtered by the WHERE clause later, so this seems unnecessary.

```suggestion
            JOIN roleschema R ON U.role_id = R.id
```

src/zenml/zen_stores/migrations/versions/5994f9ad0489_introduce_role_permissions.py
```diff
@@ -50,24 +50,24 @@ def upgrade() -> None:
...
         )
     )
     conn.execute(
         sa.text(
             """
-            DELETE FROM teamroleassignmentschema
-            WHERE role_id
-            IN (SELECT id FROM roleschema WHERE name=='admin' OR name=='guest')
+            DELETE T FROM teamroleassignmentschema T
+            LEFT JOIN roleschema R ON T.role_id = R.id
```
fa9r, line 63: Same as above, LEFT JOIN seems unnecessary.

```suggestion
            JOIN roleschema R ON T.role_id = R.id
```

fa9r (approved):
Very happy this is finally fixed so I can merge develop into my branches again 😄 

I am curious though about why the migration failed before. `WHERE x = a OR x = b` should be valid syntax in all SQL versions, no? Perhaps it's because of the `==`?

EDIT: Actually I retract my approval here, if the issue is indeed the `==` instead of `=` then I would prefer the old implementation

fa9r (commented):
Actually I retract my approval here, if the issue is indeed the `==` instead of `=` then I would prefer the old implementation

fa9r (changes requested):
See comment above

fa9r (approved):
LGTM

stefannica (approved):
Why did we need an `__init__.py` in the `versions` folder ? that does not need to be a python package AFAICT.

PR #1058, opened 2022-11-09
Rename DB Tables and Fix Foreign Keys

fa9r:
## Describe changes
I renamed all DB tables from `Xschema` to `X`.

As part of this change, I also unified all foreign key constraints in our DB to:
- have a name
- have a clearly defined delete behavior
- have a clearly defined `nullable` state

Loom PR Walkthrough:
- https://www.loom.com/share/39dc46568f5047daad3dc7ec28b4b427
- https://www.loom.com/share/72a552e131b04104a6ac9d98ec85a6ef

## Pre-requisites

## Types of changes


src/zenml/zen_stores/schemas/schema_utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def build_foreign_key_field(
+    source: str,
+    target: str,
+    source_column: str,
+    target_column: str,
+    ondelete: Optional[str] = None,
+    **sa_column_kwargs: Any,
+) -> Any:
```
strickvl, line 48: Why is this returning `Any` and not values of type `Field`?

fa9r: That's what I wanted to do initially as well. However, `sqlmodel.Field` is actually not a class, it is a function that returns type `Any` 😅

fa9r:
Doesn't work on MySQL yet: ALTER commands may be called on foreign keys there, so I need to merge my two migration scripts to perform the ALTER statements before recreating the foreign keys...

fa9r:
Now it should work for both MySQL and SQLite at least.

src/zenml/zen_stores/schemas/schema_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        target: Target table name.
+        source_column: Source column name.
+
+    Returns:
+        Name of the foreign key constraint.
+    """
+    return f"fk_{source}_{source_column}_{target}"
+
+
+def build_foreign_key_field(
```
AlexejPenner, line 41: This is freaking awesome!

AlexejPenner (commented):
This is really cool, all changes in code look great. To what extend do you want the migration script analyzed? Is it tested against SQLite and MySQL backends?

fa9r:
> This is really cool, all changes in code look great. To what extend do you want the migration script analyzed? Is it tested against SQLite and MySQL backends when those are prepopulated with data?

Yep, tested for both backends, both with a fresh install and with a DB prepopulated on develop :)

src/zenml/zen_stores/schemas/schema_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        target: Target table name.
+        source_column: Source column name.
+        target_column: Target column name.
+        ondelete: On delete behavior.
+        **sa_column_kwargs: Keyword arguments for the SQLAlchemy column.
+
+    Returns:
+        SQLModel foreign key field.
+    """
+    constraint_name = foreign_key_constraint_name(
```
schustmi, line 62: Does it make sense to include additional checks here, e.g. to prevent a case where `ondelete="SET NULL`, `nullable=False`?

src/zenml/zen_stores/schemas/user_management_schemas.py
```diff
@@ -35,13 +36,31 @@
 class TeamAssignmentSchema(SQLModel, table=True):
     """SQL Model for team assignments."""
 
-    user_id: UUID = Field(primary_key=True, foreign_key="userschema.id")
-    team_id: UUID = Field(primary_key=True, foreign_key="teamschema.id")
+    __tablename__ = "team_assignment"
+
+    user_id: UUID = build_foreign_key_field(
+        source=__tablename__,
+        target="user",  # TODO: how to reference `UserSchema.__tablename__`?
```
AlexejPenner, line 43: Are these todo's here still relevant?

AlexejPenner (approved):

schustmi (approved):

src/zenml/zen_stores/schemas/user_management_schemas.py
fa9r: If there's a way to do it I would like to reference the class instead of hardcoding the table name, but it seems like a cyclic import issue

src/zenml/zen_stores/schemas/schema_utils.py
fa9r: Depends how strict we want to be on this, but yeah we could do it. I was also thinking to mke `nullable` and `ondelete` non-optional so you have to define them properly whenever you add a foreign key.

stefannica (approved):
I love how you were able to synthesize what would've probably been thousands of lines of auto-generated alembic statements into a few hundred lines of code ! Amazing PR !

src/zenml/zen_stores/schemas/schema_utils.py
fa9r: I like the idea, this should prevent unexpected behavior down the line. I made both `nullable` and `ondelete` non-optional and added the check you suggested.

src/zenml/zen_stores/schemas/user_management_schemas.py
fa9r: I fixed all the TODOs that were fixable. The only ones left are in tables that are used as `link_model` somewhere. In these cases, the class of the foreign key tables cannot be referenced since those classes need to reference the link model already.

PR #1061, opened 2022-11-10
Docker ZenML deployment improvements and docs

stefannica:
## Describe changes
Docker ZenML deployment improvements and docs.

This PR makes some improvements to how the server MLMD configuration is advertised to clients and how the global configuration is copied into Docker containers used by local Docker based orchestrators to overall improve the local Docker ZenML experience and to enable the deployment of a MySQL + ZenML setup with docker-compose.

Deployment scenarios unlocked and/or tested with this PR:

- [x] zenml server started with `zenml up --docker` available on `http://localhost:8080`; local, local-kubeflow and docker orchestrators work with default settings. 
- [x] both zenml server and MySQL server started as local Docker containers with docker-compose (see docker-compose example in included docs) and available on `http://localhost:8080` and `mysql://127.0.0.1:3306`; local, local-kubeflow and docker orchestrators work with default settings. 
- [x] cloud zenml server and MySQL server: local, docker, local-kubeflow and remote Kubeflow orchestrators work
- [x] local zenml client connected directly to local MySQL database running in container: local, local-kubeflow and docker orchestrators work with default settings. 

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/getting-started/deploying-zenml/docker.md
+++ b/docs/book/getting-started/deploying-zenml/docker.md
@@ -223,6 +223,9 @@ docker run -it -d -p 8080:80 --name zenml \
 If you're looking for a quick way to run both the ZenML server and a MySQL
 database with Docker, you can [deploy the ZenML server with Docker Compose](#zenml-server-with-docker-compose).
 
+The rest of this guide covers various advanced use cases for running the ZenML
+server with Docker.
+
 ### Persisting the SQLite database
 
 Depending on your use case, you may also want to mount a persistent volume or
@@ -287,6 +290,30 @@ Connecting your client to the ZenML server is the same as before:
 ```shell
 zenml connect --url http://localhost:8080 --username default --password ''
 ```
+
+### Direct MySQL database connection
+
+This scenario is similar to the previous one, but instead of running a ZenML
+server, the client is configured to connect directly to a MySQL database running
+in a Docker container.
+
+As previously covered, the containerized MySQL database service can be started
+with a command like the following:
+
+```shell
+docker run --name mysql -d -p 3306:3306 -e MYSQL_ROOT_PASSWORD=password mysql:5.7
+```
+
+The ZenML client on the host machine can then be configured to connect directly
+to the database with a slightly different `zenml connect` command:
+
+```shell
+zenml connect --url mysql://127.0.0.1/zenml --username root --password password
+```
+
+> **Note**
+> The `localhost` hostname will not work with MySQL databases. You need
+> to use the `127.0.0.1` IP address instead.
 
 ### ZenML server with Docker Compose
 
--- a/docs/book/getting-started/deploying-zenml/docker.md
+++ b/docs/book/getting-started/deploying-zenml/docker.md
@@ -371,13 +371,25 @@ To start the containers, run the following command from the directory where
 the `docker-compose.yml` file is located:
 
 ```shell
-docker-compose up -d
+docker-compose -p zenml up  -d
+```
+
+or, if you need to use a different filename or path:
+
+```shell
+docker-compose -f /path/to/docker-compose.yml -p zenml up -d
 ```
 
 Connecting your client to the ZenML server is the same as before:
 
 ```shell
 zenml connect --url http://localhost:8080 --username default --password ''
+```
+
+Tearing down the installation is as simple as running:
+
+```shell
+docker-compose -p zenml down
 ```
 
 ## Troubleshooting
@@ -393,7 +405,7 @@ If you used the `zenml up --docker` CLI command to deploy the Docker ZenML
 server, you can check the logs with the command:
 
 ```shell
-zenml logs
+zenml logs -f
 ```
 
 ### Manual Docker Deployments
@@ -402,5 +414,12 @@ If you used the `docker run` command to manually deploy the Docker ZenML server,
 you can check the logs with the command:
 
 ```shell
-docker logs zenml
+docker logs zenml -f
+```
+
+If you used the `docker compose` command to manually deploy the Docker ZenML
+server, you can check the logs with the command:
+
+```shell
+docker compose -p zenml logs -f
 ```
--- a/docs/book/getting-started/deploying-zenml/docker.md
+++ b/docs/book/getting-started/deploying-zenml/docker.md
@@ -126,7 +126,8 @@ default SQLite database is located can optionally be overridden by setting the
 These configuration options are not required for most use cases, but can be
 useful in certain scenarios that require mirroring the same ZenML server
 configuration across multiple container instances (e.g. a Kubernetes
-deployment with multiple replicas):
+deployment with multiple replicas) or connecting the ZenML server to a
+MySQL database over a private network:
 
 - **ZENML_USER_ID**:
     This is a UUID value that is used to uniquely identify the server's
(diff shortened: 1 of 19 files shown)


docs/book/getting-started/deploying-zenml/docker.md
```diff
@@ -1,88 +1,359 @@
...
-    The client certificate required to connect to the MySQL database, if you
-    are using SSL client certificates to connect to the database.
-- ZENML_STORE_SSL_KEY
-    The client certificate private key required to connect to the MySQL
-    database, if you are using SSL client certificates to connect to the
-    database.
-- ZENML_STORE_SSL_VERIFY_SERVER_CERT
-    Whether to verify SSL certificate for the MySQL server.    
+{% hint style="warning" %}
+The SQLite database backend option currently has one major limitation:
```
fa9r, line 65: Why would that be? If you mount the DB, shouldn't it be available both in the container and outside where the pipelines are run? Which information will be missing exactly?

docs/book/getting-started/deploying-zenml/docker.md
```diff
@@ -1,88 +1,359 @@
...
+dependency on the ml-metadata library for pipeline run tracking.
+
+This situation will be signalled by a warning message in the ZenML client and
+pipeline logs that looks like this:
+
+    The ZenML server is using a SQLite database at /zenml/.zenconfig/local_stores/default_zen_store/zenml.db
+    that is not available locally. Using the default local SQLite database
+    instead.
+
+In the meantime, it is recommended to configure the ZenML server to connect to a
```
fa9r, line 81: TL;DR: You can either use MySQL or SQLite, but don't use SQLite :D

Why don't we tell people to use MySQL directly?

fa9r: Nevermind, just finished reading, I agree it is pretty clean to have sensible defaults that allow you to just run the image without having to set up a MySQL DB first.

Let me change my question then: If the runs are broken, when would you use the default deployment with SQLite instead of regular `zenml up`? What's the point? If there is no use case for it, I would still propose to hide the SQLite option for now and require users to set up a MySQL with their local Docker deployments

docs/book/getting-started/deploying-zenml/docker.md
```diff
@@ -1,88 +1,359 @@
...
+
+Connecting your client to the ZenML server is the same as before:
+
+```shell
+zenml connect --url http://localhost:8080 --username default --password ''
+```
+
+## Troubleshooting
+
+You can check the logs of the container to verify if the server is up and depending on where you have deployed it, you can also access the dashboard at a `localhost` port (if running locally) or through some other service that exposes your container to the internet.
```
fa9r, line 358: Wouldn't hurt to explicitly write the logs CLI command here again

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -142,6 +142,27 @@ def get_store_class(store_type: StoreType) -> Type["BaseZenStore"]:
...
+        """
+        from zenml.zen_stores.rest_zen_store import RestZenStoreConfiguration
+        from zenml.zen_stores.sql_zen_store import SqlZenStoreConfiguration
+
+        if SqlZenStoreConfiguration.supports_url_scheme(url):
+            return StoreType.SQL
+        elif RestZenStoreConfiguration.supports_url_scheme(url):
+            return StoreType.REST
+
+        return None
```
fa9r, line 164: When would this happen? Intuitively, I feel like this should raise an error instead of the `None` return in such cases

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -399,6 +422,8 @@ def get_metadata_config(
         Args:
             expand_certs: Whether to expand the certificate paths to their
                 contents.
+            external: Whether to return the external metadata configuration
+                that is advertised to clients.
```
fa9r, line 424: I don't see any arg like this

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -252,9 +266,20 @@ def copy_configuration(
...
+        from zenml.utils.docker_utils import (
+            replace_localhost_with_internal_hostname,
+        )
+
         assert isinstance(config, RestZenStoreConfiguration)
         config = config.copy(deep=True)
 
+        # If the URL is localhost, then we make the assumption that the
+        # copied configuration will be used in a docker container running
+        # on the same machine (other scenarios are not technically possible)
```
fa9r: I don't know where exactly this method is called, but why are other scenarios not possible?

fa9r (approved):
Both docs and code look good and are very well explained.

Only thing I was wondering about is the SQLite limitation and whether we should even expose it as an option if runs aren't working there.

Otherwise LGTM!

src/zenml/environment.py
```diff
@@ -44,6 +44,8 @@ def get_environment() -> str:
         return "kubernetes"
     elif Environment.in_docker():
         return "docker"
+    elif Environment.in_container():
```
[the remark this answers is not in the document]
stefannica: Not all containers use docker as a container runtime. All "docker containers" are "containers" but the reverse is not valid: not all "containers" are "docker containers". Other container runtime engines exist (runc, containerd or cri-o for example).

docs/book/getting-started/deploying-zenml/docker.md
stefannica: Even if you mount the SQLite DB in the container, it's too complicated to get the client and local (Docker) orchestrators to all access (mount and access in the case of local Docker orchestrators) the same file:

- the client and server need to use the exact same path
- the SQLite database file must reside under the `local_stores` folder, which is the only folder mounted in the docker orchestrator containers
- other complications that I can't think of right now

I prefer to discourage users from using this DB type rather than to explain the intricacies of manually managing what ZenML does automatically.

docs/book/getting-started/deploying-zenml/docker.md
stefannica: When we remove the need for clients to connect directly to the server's database, this SQLite scenario will become a very important easy-to-set-up ZenML server deployment option.

My (not so secret) plan is to keep the option documented and just update the docs to reflect it as the simpler option when we finally remove MLMD.

src/zenml/zen_stores/base_zen_store.py
stefannica: Good point. Done

src/zenml/zen_stores/rest_zen_store.py
stefannica: The method is only called when docker images are built for orchestrators and operators that rely on containers. The technical impossibility comes from the range of orchestrators and operators that are available at the moment. They either use Docker locally (in which case the localhost mapping works as it should) or they use a remote cloud (in which case, the setup is not going to work either way, so it's irrelevant what we do here).

src/zenml/environment.py
```diff
@@ -158,8 +171,8 @@ def in_docker() -> bool:
             container, `False` otherwise.
         """
         # TODO [ENG-167]: Make this more reliable and add test.
-        if INSIDE_ZENML_CONTAINER:
-            return True
+        if not INSIDE_ZENML_CONTAINER:
```
schustmi, line 174: I think this logic is not optimal now, as something like the following won't be detected by either of `in_container` or `in_docker` anymore.

```bash
docker run -it ubuntu bash
pip install zenml
zenml stack register ...
```

docs/book/getting-started/deploying-zenml/docker.md
stefannica: Done

src/zenml/zen_stores/base_zen_store.py
stefannica: Done

src/zenml/zen_stores/sql_zen_store.py
stefannica: Thanks for noticing, docstring linter was complaining too. Removed it.

src/zenml/environment.py
stefannica: Right, this is for the case when the CLI is also run in a container. I'll remove the extra condition.

stefannica: Done

schustmi (approved):

PR #1066, opened 2022-11-13
Refactor Kubeflow Orchestrator - Disentangle K3d

safoinme:
## Describe changes
I think the kubernetes_context should return the active stack if one is set in the system Kubernetes configuration, only if it's not set to return to the k3d k8s cluster.

## Pre-requisites

## Types of changes


stefannica (changes requested):

fa9r (commented):
Overall this looks good to me.

One thing I was wondering though is whether you could further disentangle the k3d code from the local Kubeflow orchestrator, i.e., move all k3d functionality to some separate module that the local Kubeflow orchestrator imports.

The reason I am asking is that this might then directly enable local Kubernetes as well for which we have a separate open ticket (ENG-983), so you could potentially implement two big features at once with this PR :D

stefannica (changes requested):
Without looking at the code changes in depth, I imagine this change also requires a lot of updates to our documentation and examples, but I didn't see any.

fa9r (approved):
Code looks good to me.

It seems k3d and the local kubeflow are still tighly coupled, but I guess we can merge this PR and then do those other changes later without release pressure. Perhaps trying to implement a local k3d Kubernetes orchestrator while refactoring the k3d code out might also be helpful since you will directly notice which code parts need to be duplicated and should therefore be extracted/abstracted.

@stefannica regarding docs I think we decided that we do not need separate examples/docs pages for the local k3d orchestrators for now. However, you are right in that there might be places in the docs or examples that will be outdated after this PR. @safoinme did you go through our content once and check that?

fa9r:
Also, one of the kfp unit tests it failing

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -106,46 +96,18 @@ def config(self) -> KubeflowOrchestratorConfig:
...
             The kubernetes context associated with the orchestrator.
         """
-        if self.config.kubernetes_context:
-            return self.config.kubernetes_context
-        return self._get_k3d_kubernetes_context(self.id)
+        return (
+            self.config.kubernetes_context
+            if self.config.kubernetes_context
+            else None
+        )
```
schustmi, line 110:
```suggestion
        return self.config.kubernetes_context
```

schustmi:
I'm guessing this also breaks all existing stacks with local kubeflow orchestrators because of the new flavor name?
In which case this needs also some comments on how to migrate these

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -124,8 +130,6 @@ class KubeflowOrchestratorConfig(BaseOrchestratorConfig):
             block until all steps finished running on KFP.
         skip_local_validations: If `True`, the local validations will be
             skipped.
-        skip_cluster_provisioning: If `True`, the k3d cluster provisioning will
-            be skipped.
         skip_ui_daemon_provisioning: If `True`, provisioning the KFP UI daemon
```
schustmi, line 133: This is not currently doing anything for the remote orchestrator, should we move it or re-introduce this functionality?

schustmi (changes requested):

PR #1069, opened 2022-11-14
Make additional stack component config options available as a setting

schustmi:
There are two changes in this PR:

### Default values for all stack component settings can now be configured on the stack component itself
This is solved by making sure the respective `StackComponentConfig` inherits from the settings class specified for the component. For example:
```python
class MLFlowExperimentTrackerSettings(BaseSettings):
    experiment_name: Optional[str] = None
    nested: bool = False
    ...

class MLFlowExperimentTrackerConfig(
    BaseExperimentTrackerConfig, MLFlowExperimentTrackerSettings
):
    tracking_uri: Optional[str] = None
    tracking_username: Optional[str] = SecretField()
    tracking_password: Optional[str] = SecretField()
    ...
```
The config inherits from the settings, which means that users can now for example specify that all runs using this experiment tracker component should be nested by default:
`zenml experiment-tracker register mlflow --flavor=mlflow --nested=True`

This default value can then be overwritten as before by specifying settings on a step/pipeline/run:
```python
@pipeline(
    settings={
        "experiment_tracker.mlflow": MLFlowExperimentTrackerSettings(nested=False)
    }
)
def my_pipeline(...):
    ...
```

### Update to existing stack components

Existing stack component configurations which are "dynamic" and make sense to be configurable on a run-by-run basis were included as settings for that component. Due to technical limitations this right now isn't possible for all options as settings can only be configured when running a pipeline, and not for example when calling `zenml stack up`.

## Pre-requisites

## Types of changes


htahir1:
@schustmi a description for this PR would be helpul

schustmi:
@hamza This is for the CI, not ready for review that's why no one is tagged

htahir1 (changes requested):
Hard one to review but I think there are no glaring errors, except there are no docs. We should have docs for these

src/zenml/integrations/gcp/flavors/vertex_step_operator_flavor.py
```diff
@@ -28,28 +29,39 @@
...
+class VertexStepOperatorSettings(BaseSettings):
+    """Settings for the Vertex step operator.
+
+    Attributes:
+        accelerator_type: Accelerator type from list
+            https://cloud.google.com/vertex-ai/docs/reference/rest/v1/MachineSpec#AcceleratorType
+        accelerator_count: Defines number of accelerators to be used for the
+            job.
+        machine_type: Machine type specified here
+            https://cloud.google.com/vertex-ai/docs/training/configure-compute#machine-types
```
safoinme, line 41: does it make sense to add this here or in docs [compatibility-table](https://cloud.google.com/vertex-ai/docs/training/configure-compute#gpu-compatibility-table)

schustmi: Absolutely, I'll add it!

safoinme (approved):
I really enjoyed going through the changes! I think all looks good regarding the PR. 🚀

docs/book/advanced-guide/pipelines/settings.md
```diff
@@ -45,10 +45,20 @@ Settings are categorized into two types:
...
 A good example of this is the [`MLflow Experiment Tracker`](../../component-gallery/experiment-trackers/mlflow.md), where configuration which remains static such as the `tracking_url` is sent through at registration time, while runtime configuration such as the `experiment_name` (which might change every pipeline run) is sent through as runtime settings.
 
+{% hint style="info" %}
+All ZenML default stack components however also allow you to specify default values for settings in their configuration. For example,
+you could set a default value for the `nested` setting of your MLflow experiment tracker:
+`zenml experiment-tracker register <NAME> --flavor=mlflow --nested=True`
+
+This means that all pipelines that run using this experiment tracker use nested MLflow runs unless overwritten by specifying settings
+for the pipeline.
+{% endhint %}
```
htahir1, line 60:
```suggestion
Even though settings can be overriden at runtime, you can also specify *default* values for settings while configuring a stack component. For example, you could set a default value for the `nested` setting of your MLflow experiment tracker:
`zenml experiment-tracker register <NAME> --flavor=mlflow --nested=True`

This means that all pipelines that run using this experiment tracker use nested MLflow runs unless overwritten by specifying settings for the pipeline at runtime.
```

PR #1070, opened 2022-11-15
Add get_or_create_run() ZenStore method

fa9r:
## Describe changes
I created a `get_or_create_run()` method in the ZenStores.

This is currently only used by the orchestrator, which calls this method at the start of each step.

The main reason for moving this logic from the orchestrator into the SQLZenStore is that the server logs are now no longer polluted with error messages.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -2898,6 +2898,25 @@ class SqlZenStore(BaseZenStore):
             self.update_run(run_model)
         return run_model
 
+    def get_or_create_run(
+        self, pipeline_run: PipelineRunModel
+    ) -> PipelineRunModel:
+        """Gets or creates a pipeline run.
+
+        If a run with the same ID already exists, it is returned. Otherwise, a
+        new run is created.
+
+        Args:
+            pipeline_run: The pipeline run to get or create.
+
+        Returns:
+            The pipeline run.
+        """
+        try:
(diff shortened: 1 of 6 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2898,6 +2898,25 @@ def _update_run_status(
...
+        new run is created.
+
+        Args:
+            pipeline_run: The pipeline run to get or create.
+
+        Returns:
+            The pipeline run.
+        """
+        try:
+            return self.get_run(pipeline_run.id)
```
schustmi, line 2916: Swap call order

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2903,19 +2903,27 @@ def get_or_create_run(
...
+        # We want to have the create statement in the try block since running it
+        # first will reduce concurrency issues.
         try:
-            return self.get_run(pipeline_run.id)
-        except KeyError:
             return self.create_run(pipeline_run)
+        except EntityExistsError:
+            # Currently, an `EntityExistsError` is raised if either the run ID
+            # or the run name already exists. Therefore, we need to have another
+            # try block since getting the run by ID might still fa
```
schustmi, line 2922:
```suggestion
            # try block since getting the run by ID might still fail.
```

schustmi (approved):

PR #1080, opened 2022-11-17
Add `zenml pipeline schedules list` CLI Command

fa9r:
## Describe changes
ZenML now keeps track of your pipeline run schedules. Whenever you call `pipeline.run(schedule=...)`, we now save this information in our DB and you can get a list of all your schedules via `zenml pipeline schedules list`:

```
┏━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━┯━━━━━━━━━━┯━━━━━━━━━━━━━━━━━┯━━━━━━━━━┯━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━┓
┃        NAME        │ CRON_EXPRESSION │ START_TIME │ END_TIME │ INTERVAL_SECOND │ CATCHUP │ ACTIVE │ ORCHESTRATOR_ID    │ PIPELINE_ID       ┃
┠────────────────────┼─────────────────┼────────────┼──────────┼─────────────────┼─────────┼────────┼────────────────────┼───────────────────┨
┃ first_pipeline-202 │ */5 * * * *     │ None       │ None     │ None            │ False   │ True   │ 8f79e086-3eb4-4297 │ 9b14a487-93d6-47f ┃
┃ 2_11_17-23_50_55_4 │                 │            │          │                 │         │        │ -b4dc-34df1050fe88 │ 6-ba40-87591d5a46 ┃
┃       84621        │                 │            │          │                 │         │        │                    │ 6f                ┃
┗━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━┷━━━━━━━━━━┷━━━━━━━━━━━━━━━━━┷━━━━━━━━━┷━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━┛
```

### Detailed changes
- New table `schedule` (`ScheduleSchema`)
- 3 new models `ScheduleRequestModel`, `ScheduleResponseModel`, `ScheduleUpdateModel`
- 4 new API endpoints (create, get, list, update)
- `zenml.config.schedule.Schedule` now has a `name`
- pipeline runs now have an optional `schedule_id`

TODO:
- figure out when/where exactly the schedule registration should be called
- unit tests
- docs
- test whether scheduled runs have `schedule_id` set correctly
- schedule deletion

## Pre-requisites

## Types of changes


fa9r:
@schustmi would love some conceptual feedback on this, in particular:
- Where should schedule registration be called in your opinion? Perhaps somewhere after we check whether the specific orchestrator even supports scheduling
- Should we link schedules to the orchestrator in the DB?
- Is there any way we can disable a schedule again from within ZenML?

schustmi:
The schedule probably needs to be connected not only to the ID of an orchestrator, but also to the ID of the job/recurring run of the orchestrator. This only exists once the orchestrator has submitted the request (or in the case of Airflow, never).
So the generic way to handle this would maybe be a return value from `orchestrator.run(...)` which returns this ID, after which we could register the schedule.

I'm not really sure this is something we want to do/give users access to anyway, because even if we might be able to stop schedules for some orchestrators like KFP, others like Airflow won't support this and the schedule will be in there forever.
Additionally, there probably won't be an easy way if the user cancels the schedule on the orchestrator directly

fa9r:
Closed in favor of #1227

PR #1083, opened 2022-11-21
Remove  source pins and ignore source pins during step spec comparisons

schustmi:
## Describe changes
This PR removes our version pins and updates the step spec comparison to ignore them if added during legacy runs.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/source_utils.py
+++ b/src/zenml/utils/source_utils.py
@@ -53,7 +53,6 @@ from typing import (
     Union,
 )
 
-from zenml import __version__, constants
 from zenml.enums import StackComponentType
 from zenml.environment import Environment
 from zenml.logger import get_logger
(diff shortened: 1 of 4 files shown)


src/zenml/utils/source_utils.py
```diff
@@ -393,16 +371,16 @@ def resolve_class(class_: Type[Any], replace_main_module: bool = True) -> str:
         source_path e.g. this.module.Class.
     """
     initial_source = class_.__module__ + "." + class_.__name__
-    if is_standard_source(initial_source):
-        return resolve_standard_source(initial_source)
+    if is_internal_source(initial_source):
+        return initial_source
```
fa9r, line 375: So if I get this right, we now no longer add pins anywhere now, right? Why were we doing this in the first place then?

fa9r (commented):
I think this not only removes the source pin from the spec, but also from all other places where the pins were used before. If that is intended, then the changes look great and ready to merge to me.

fa9r (approved):

htahir1 (approved):

PR #1089, opened 2022-11-22
Bugfix/eng 1485 fix api docs build

AlexejPenner:
## Describe changes
The API Docs build was failing because of sagemaker, airflow and bentoml having conflicting pins on their attrs version. On top of that, the neweset version on jinja2 deprecated the contexthandler which is used by mkdocstrings.

 
## Pre-requisites

## Types of changes


htahir1 (approved):
This is getting weirder and weirder. I would suggest we use @schustmi's approach here where we just install a depenedency, run the apidocs on the folder of that dependency, and keep going until we get the docs compiled. Is that possible?

PR #1103, opened 2022-11-23
Remove `ml-pipelines-sdk` dependency

schustmi:
## Describe changes

We removed the `ml-pipelines-sdk` dependency (including `ml-metadata`).

This enables a variety of new features:
- Python 3.10 support
- Native M1 Mac support
- Significantly faster pipeline runs
- Significantly faster pipeline-related DB queries
- Fewer dependencies

### Breaking Changes
- `zenml pipeline runs migrate` CLI command removed
- gRPC metadata service removed
- `/metadata-config` and `/metadata-sync` API endpoints removed

### Detailed Changes
- New `zenml.orchestrators.launcher` and `zenml.orchestrators.executor` that replace the respective TFX classes and write all data directly into the ZenML DB (#1100),
- New `zenml.io.filesystem` that replaces the TFX filesystem (#1095, #1093),
- Redesigned artifact schemas and models that now only store one entry per physical artifact, similar to MLMD (#1078, #1090),
- Caching-related information is now stored in our DB for both pipeline and step runs. In particular, there now is a `cache_key` field that can be used to determine whether a step can be cached or not. See `zenml.orchestrators.cache_utils.py` for details on how caching now works (#1085).

## Pre-requisites

## Types of changes


bcdurak:
LGTM!

.github/workflows/integration-test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       matrix:
         # TODO: readd macos-latest once we fix the integration tests
         os: [ ubuntu-latest, windows-latest ]
-        python-version: [ 3.7, 3.8, 3.9 ]
+        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
```
AlexejPenner, line 25: :fire:

strickvl, line 25: oh cool, you got it working!

schustmi: Unfortunately not yet, GH actions just decided to not run for this commit, I think they're having issues today as I've had that a few times

src/zenml/config/compiler.py
```diff
@@ -346,58 +326,11 @@ def _compile_step(
         )
 
         complete_step_configuration = StepConfiguration(
-            docstring=step.__doc__, **step.configuration.dict()
```
AlexejPenner, line 349: At what point are we now setting the step docstring?

schustmi: `BaseStep.__init__`

.github/workflows/integration-test.yml
schustmi: Had to remove this again due to some integrations (tensorflow, pytorch and probably others) don't support Python 3.11 yet.

src/zenml/zen_stores/migrations/versions/10a907dad202_delete_mlmd_tables.py
```diff
@@ -0,0 +1,109 @@
...
+# revision identifiers, used by Alembic.
+revision = "10a907dad202"
+down_revision = "6a28c4fd0ef2"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    # Bye bye MLMD :wave: :wave: :wave:
```
AlexejPenner, line 20: :laughing:

tests/unit/steps/test_base_step.py
```diff
@@ -740,100 +602,115 @@ def my_step() -> None:
...
-    @step
-    def some_step_3() -> Output(out1=int, out2=int):
-        return 1, "not_an_int"
-
-    for step_function in [some_step_1, some_step_2, some_step_3]:
+    for step_function in [
+        wrong_int_output_step_1,
+        wrong_int_output_step_2,
+        wrong_int_output_step_3,
+    ]:
```
AlexejPenner, line 660: could also be done with @pytest.mark.parametrize("step_function", [wrong_int_output_step_1, wrong_int_output_step_2, wrong_int_output_step_3]) - That way these show up as three separate tests and this test becomes more obviously extensible to more failing scenarios

tests/unit/steps/test_base_step.py
```diff
@@ -740,100 +602,115 @@ def my_step() -> None:
...
-        some_step_5,
-        some_step_6,
-        some_step_7,
+        wrong_num_outputs_step_1,
+        wrong_num_outputs_step_2,
+        wrong_num_outputs_step_3,
+        wrong_num_outputs_step_4,
+        wrong_num_outputs_step_5,
+        wrong_num_outputs_step_6,
+        wrong_num_outputs_step_7,
```
AlexejPenner, line 713: Same as above with parametrize?

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -375,21 +369,7 @@ def _configure_container_op(
...
-        # security context:
-        # https://www.kubeflow.org/docs/components/pipelines/installation/localcluster-deployment/#deploying-kubeflow-pipelines
-        # https://argoproj.github.io/argo-workflows/empty-dir/
-        # KFP will switch to the Emissary executor (soon), when this emptyDir
-        # mount will not be necessary anymore, but for now it's still in alpha
-        # status (https://www.kubeflow.org/docs/components/pipelines/installation/choose-executor/#emissary-executor)
-        volumes: Dict[str, k8s_client.V1Volume] = {
-            "/outputs": k8s_client.V1Volume(
-                name="outputs", empty_dir=k8s_client.V1EmptyDirVolumeSource()
-            ),
```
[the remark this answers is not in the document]
schustmi: Not necessarily TFX, but some tfx util function wrote markdown files there to display in the Kubeflow UI. The same information now exists (in much nicer form) in the ZenML dashboard and isn't needed anymore

src/zenml/models/step_run_models.py
```diff
@@ -75,6 +69,11 @@ class StepRunRequestModel(StepRunBaseModel, BaseRequestModel):
...
-@update_model
-class StepRunUpdateModel(StepRunRequestModel):
+
+
+class StepRunUpdateModel(BaseModel):
     """Update model for step runs."""
+
+    output_artifacts: Dict[str, UUID] = {}
+    status: Optional[ExecutionStatus] = None
+    end_time: Optional[datetime] = None
```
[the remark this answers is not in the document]
schustmi: Yes, we intentionally made this as restrictive as possible to make it clear what can/can't be updated

src/zenml/orchestrators/output_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    return output_artifacts
+
+
+def remove_artifact_dirs(artifacts: Sequence[BaseArtifact]) -> None:
+    """Removes the artifact directories.
+
+    Args:
+        artifacts: Artifacts for which to remove the directories.
+    """
```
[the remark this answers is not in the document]
schustmi: If a step fails due to something in ZenML or the user code going wrong, we clean up the (usually empty) artifact dirs

src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -56,10 +56,7 @@
 from zenml.models.server_models import ServerModel
 
 if TYPE_CHECKING:
-    from ml_metadata.proto.metadata_store_pb2 import (
-        ConnectionConfig,
-        MetadataStoreClientConfig,
-    )
+    pass
```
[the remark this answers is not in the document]
schustmi: Absolutely, I think our formatting script leaves this after removing unused imports

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -675,26 +564,43 @@ def sync_lock(self) -> Lock:
...
+        ):
+            master_url = self._engine.url._replace(database="mysql")
+            master_engine = create_engine(
+                url=master_url, connect_args=connect_args, **engine_args
+            )
+            query = f"CREATE DATABASE IF NOT EXISTS {self.config.database}"
+            conn = master_engine.connect()
+            conn.execute(text(query))
+            conn.close()
+
```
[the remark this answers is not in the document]
schustmi: Previously MLMD was still imported before alembic ran and created these tables for us. After removing this, we had to make sure for both SQLite and MySQL that the parent dir/database exists before connecting to it

src/zenml/orchestrators/step_launcher.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+class StepLauncher:
+    """This class is responsible for launching a step of a ZenML pipeline.
+
+    It does the following:
+    1. Query ZenML to resolve the input artifacts of the step,
+    2. Generate a cache key based on the step and its artifacts,
+    3. Build a `StepRunModel` for the step run,
+    4. Check if the step run can be cached, and if so, register it as cached.
```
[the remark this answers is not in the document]
schustmi: Just rewrote this entire docstring

src/zenml/orchestrators/step_runner.py
```diff
@@ -0,0 +1,283 @@
...
+                )
+
+        output_annotations = parse_return_type_annotations(spec.annotations)
+        if len(output_annotations) > 0:
+            # if there is only one output annotation (either directly specified
+            # or contained in an `Output` tuple) we treat the step function
+            # return value as the return for that output
+            if len(output_annotations) == 1:
+                return_values = [return_values]
+            elif not isinstance(return_values, Sequence):
```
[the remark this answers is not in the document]
schustmi: Mhm this is old code from the TFX executor, but very interesting suggestion, I'll debug and see if this leads to any issues

src/zenml/zen_stores/migrations/versions/10a907dad202_delete_mlmd_tables.py
schustmi: Haha @fa9r, didn't see that before 😄

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3165,125 +3060,201 @@ def _run_step_schema_to_model(
...
             # Update the step
-            existing_step.update(step_update)
-            session.add(existing_step)
-            session.commit()
+            existing_step_run.update(step_run_update)
+            session.add(existing_step_run)
+
+            # Update the output artifacts.
+            outputs = step_run_update.output_artifacts
+            if outputs is not None:
```
AlexejPenner, line 3167: Won't this always be True? The default value for output_artifacts is `{}` and `({} is not None) == True`

src/zenml/orchestrators/publish_utils.py
```diff
@@ -0,0 +1,179 @@
...
+        The updated step run.
+    """
+    return Client().zen_store.update_run_step(
+        step_run_id=step_run_id,
+        step_run_update=StepRunUpdateModel(
+            status=ExecutionStatus.COMPLETED,
+            end_time=datetime.now(),
+            output_artifacts=output_artifact_ids,
+        ),
+    )
```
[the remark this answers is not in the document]
schustmi: I think the create/update methods for pipeline runs and step runs currently exist only on the zen store for this exact reason, they should be internal and not used. Do you think that's okay for this case or something we should expose nontheless?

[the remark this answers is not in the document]
AlexejPenner: Works for me, if we hide create and update but expose get and list. What are your opinions on delete?

src/zenml/zen_stores/sql_zen_store.py
schustmi: Good catch, I simplified it 👍

src/zenml/orchestrators/publish_utils.py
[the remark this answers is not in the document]
schustmi: Delete doesn't even exist on the ZenStore at the moment, I would probably keep it hidden as well if we decide to add it at some point

src/zenml/zen_stores/migrations/versions/10a907dad202_delete_mlmd_tables.py
fa9r: 😜

src/zenml/orchestrators/step_runner.py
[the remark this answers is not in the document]
schustmi: Nice catch, this would indeed have passed (since probably the intial 0.5.0 release):

```python
@step
def my_step() -> Output(a=str, b=str):
  return "ab"
```

The two outputs would have been "a" and "b". I prevented this and added a test for it.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -675,26 +564,43 @@ def sync_lock(self) -> Lock:
...
         url, connect_args, engine_args = self.config.get_sqlmodel_config()
         self._engine = create_engine(
             url=url, connect_args=connect_args, **engine_args
         )
+
+        # SQLite: As long as the parent directory exists, SQLAlchemy will
+        # automatically create the database.
+        local_path = self.config.get_local_path(url)
+        if local_path and not fileio.exists(local_path):
+            fileio.makedirs(os.path.dirname(local_path))
```
stefannica, line 578:
```suggestion
        if (
            self.config.driver == SQLDatabaseDriver.SQLITE
            and self.config.database
            and not fileio.exists(self.config.database)
        ):
            fileio.makedirs(os.path.dirname(self.config.database))
```

And with this you can remove the `get_local_path` method from the configuration, as it's already overcrowded with all kinds of helper methods.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -675,26 +564,43 @@ def sync_lock(self) -> Lock:
...
+        # MySQL: We might need to create the database manually.
+        # To do so, we create a new engine that connects to the `mysql` database
+        # and then create the desired database.
+        # See https://stackoverflow.com/a/8977109
+        if (
+            self.config.driver == SQLDatabaseDriver.MYSQL
+            and self.config.database
+        ):
+            master_url = self._engine.url._replace(database="mysql")
+            master_engine = create_engine(
```
stefannica, line 589: A good ops practice is to manually create the database and then restrict the DB user permissions to only have access to that database. In that case, I imagine that trying to access the mysql database will fail.
You should first try to find out if the database exists in the context of the current user.

stefannica (commented):
Just a couple of suggestions, looks good otherwise.

.github/workflows/integration-test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       matrix:
         # TODO: readd macos-latest once we fix the integration tests
         os: [ ubuntu-latest, windows-latest ]
-        python-version: [ 3.7, 3.8, 3.9 ]
+        python-version: [3.7, 3.8, 3.9, '3.10']
```
stefannica, line 25: One more thing: we also need to update `release-cloudbuild.yml` to build images for python 3.10.

src/zenml/client.py
```diff
@@ -2190,24 +2169,32 @@ def export_pipeline_runs(self, filename: str) -> None:
         if not pipeline_runs:
             logger.warning("No pipeline runs found. Nothing to export.")
             return
-        yaml_data = []
+
+        yaml_data: Dict[str, List[Dict[str, Any]]] = {
+            "artifacts": [],
+            "runs": [],
+        }
```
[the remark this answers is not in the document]
schustmi: +1, @fa9r what do you think?

.github/workflows/integration-test.yml
schustmi: Good point, I thought of this a few times but never got around to doing it!

src/zenml/zen_stores/sql_zen_store.py
schustmi: Very nice, was looking for something like that but apparently missed it 👍

stefannica (approved):

PR #1107, opened 2022-11-25
Rename project -> workspace

AlexejPenner:
## Describe changes
**Only merge after the Optimization branch**

I renamed project -> workspace. In the migration all foreign key constraints are updated appropriately.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/1de87a8c8e8c_rename_project_to_workspace.py
```diff
@@ -0,0 +1,214 @@
...
+    table_names, new_fk_constraints, old_fk_constraints = _get_changes()
+
+    engine_name = op.get_bind().engine.name
+
+    # Under MySQL, we need to sort the old foreign keys by source table and
+    # source column first since the default foreign key names contain the
+    # foreign key number.
+    if engine_name == "mysql":
+        old_fk_constraints.sort(key=lambda x: (x[0], x[2]))
+    source_table_fk_constraint_counts: Dict[str, int] = defaultdict(int)
```
fa9r, line 141: this shouldn't be needed anymore since all the foreign keys already have an explicit name now

src/zenml/zen_stores/migrations/versions/1de87a8c8e8c_rename_project_to_workspace.py
```diff
@@ -0,0 +1,214 @@
...
+    for source, target, source_column, _, _ in old_fk_constraints:
+        if engine_name == "sqlite":
+            constraint_name = _fk_constraint_name(source, target, source_column)
+        elif engine_name == "mysql":
+            source_table_fk_constraint_counts[source] += 1
+            fk_num = source_table_fk_constraint_counts[source]
+            constraint_name = f"{source}_ibfk_{fk_num}"
+        else:
+            raise NotImplementedError(f"Unsupported engine: {engine_name}")
+        _drop_fk_constraint(source, constraint_name)
```
fa9r, line 153: Same here, checking DB engines should no longer be needed 
```suggestion
        constraint_name = _fk_constraint_name(source, target, source_column)
        _drop_fk_constraint(source, constraint_name)
```

src/zenml/zen_stores/migrations/versions/1de87a8c8e8c_rename_project_to_workspace.py
```diff
@@ -0,0 +1,214 @@
...
+    for source, target, source_column, _, _ in new_fk_constraints:
+        if engine_name == "sqlite":
+            constraint_name = _fk_constraint_name(source, target, source_column)
+        elif engine_name == "mysql":
+            source_table_fk_constraint_counts[source] += 1
+            fk_num = source_table_fk_constraint_counts[source]
+            constraint_name = f"{source}_ibfk_{fk_num}"
+        else:
+            raise NotImplementedError(f"Unsupported engine: {engine_name}")
+        _drop_fk_constraint(source, constraint_name)
```
fa9r, line 197: Same as above; also, I think there is no need to write the downgrade for this one since we won't support the downgrade in the TFX removal branch either

fa9r (commented):
migration should work but you shouldn't need the DB engine checks since all of the foreign keys already have explicit names now.

To check whether it worked, you can also just generate another alembic revision on top of this one which should ideally do nothing then

PR #1110, opened 2022-11-25
Validate Stack Before Provision

safoinme:
## Describe changes
This call the `validate function`  when the stack is provisioned which will validate if the stack configuration is correct! 

## Pre-requisites

## Types of changes


schustmi (approved):
LGTM. Maybe something we can add a test for?

safoinme:
@stefannica Before 0.20 the validation was done at the stack register level. the validate function was being called pre-register stack. while now that isn't called anymore. Do you think we should be back to the register function or provision?

safoinme:
@schustmi a test for this would be creating a stack with a remote orchestrator and no remote registry or do you think there is a better way to test it?

schustmi:
@safoinme The easiest test to me would be to create a stack object with a mock `validate` method and make sure that it gets called when calling `stack.provision()`

PR #1113, opened 2022-11-28
Add Pagination to all List commands

AlexejPenner:
**Please Note** This PR can not be merged to develop until our dashboard has been adjusted to for backend pagination. The Motivation for this Feature is that data should be filtered as early as possible (in the persistence layer) rather than client-side. To achieve this the query parameters of our endpoint routes need to deliver all the necessary information. 

The proposed design is as follows:

`/stacks?name=contains:default`

This would then be interpreted correctly on the backend side.

The CLI looks like this now:

```
(devenv) apenner@apenner-XPS:~/PycharmProjects/zenml$ zenml stack list --help

Usage: zenml stack list [OPTIONS]

  List all stacks that fulfill the filter requirements. By default all filters
  are interpreted as a check for equality. However advanced filter operators
  can be used to tune the filtering by writing the operator and separating it
  from the query parameter with a colon `:`, e.g. --field='operator:query'.

  Each datatype supports a specific set of filter operations, here are the
  relevant ones for the parameters of this command:

  [STRING] supported filter operators: ['equals', 'startswith', 'contains',
  'endswith']

  [UUID] supported filter operators: ['equals', 'startswith', 'contains',
  'endswith']

  [DATETIME] supported filter operators: ['equals', 'gt', 'gte', 'lt', 'lte']

  [BOOL] supported filter operators: ['equals']

Options:
  --sort_by TEXT           [STRING] Example: --sort_by='contains:example' to
                           filter everything that contains the query somewhere
                           in its sort_by.
  --logical_operator TEXT
  --page TEXT
  --size TEXT
  --id TEXT                [UUID] Example: --id='startswith:ab53ca' to filter
                           for all UUIDs starting with that prefix.
  --created TEXT           [DATETIME] The following datetime format is
                           supported: '%y-%m-%d %H:%M:%S'. Make sure to keep
                           it in quotation marks. Example:
                           --created='gte:%y-%m-%d %H:%M:%S' to filter for
                           everything created on or after the given date.
  --updated TEXT           [DATETIME] The following datetime format is
                           supported: '%y-%m-%d %H:%M:%S'. Make sure to keep
                           it in quotation marks. Example:
                           --updated='gte:%y-%m-%d %H:%M:%S' to filter for
                           everything created on or after the given date.
  --is_shared TEXT         [BOOL] Example: --is_shared='startswith:ab53ca' to
                           filter for all UUIDs starting with that prefix.
  --name TEXT              [STRING] Example: --name='contains:example' to
                           filter everything that contains the query somewhere
                           in its name.
  --description TEXT       [STRING] Example: --description='contains:example'
                           to filter everything that contains the query
                           somewhere in its description.
  --project_id TEXT        [UUID] Example: --project_id='startswith:ab53ca' to
                           filter for all UUIDs starting with that prefix.
  --user_id TEXT           [UUID] Example: --user_id='startswith:ab53ca' to
                           filter for all UUIDs starting with that prefix.
  --component_id TEXT      [UUID] Example: --component_id='startswith:ab53ca'
                           to filter for all UUIDs starting with that prefix.
  --help                   Show this message and exit.
```

In order to filter stacks, this could look like this:

`zenml stack list --created="gt:22-12-04 17:00:00" --name contains:def`

## Additional Changes

The fact that we had UserRoleAssignmentSchemas and TeamRoleAssignmentSchemas but only a RoleAssignmentModel broke our general customs around Model-Schema relationships and would have made a sensible Filter/Pagination abstraction much more involved/impossible, 
To deal with this, this PR introduces separate methods/endpoints and models around User and Team Role Assignments.

## Leftover TODOs:

[] Implement TeamRoleAssignment
[] Write ZenStore tests for all filter methods (currently there is too much changes happening on that side in the tests, and it makes sense implementing this when the merge is drawing closer)
[] Fix all failing tests, preferably once the first batch of changed tests made it in
[] Somehow in the neweset merge the default stack is no longer created on this stack if the db is empty

## Pre-requisites

## Types of changes


src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def generate_filter(self, table: Type[SQLModel]):
+        ands = []
+        for column_filter in self.list_of_filters:
+            ands.append(column_filter.generate_query_conditions(table=table))
+
+        return ands
+
+    @classmethod
+    def click_list_options(cls):
```
schustmi, line 363: Our core models should not have any connection to `click`, I think this should rather be utility functions in `zenml/cli`.

src/zenml/models/page_model.py
```diff
@@ -0,0 +1,166 @@
...
+
+from zenml.models.base_models import BaseResponseModel
+from zenml.models.filter_models import FilterBaseModel
+from zenml.zen_stores.schemas.base_schemas import BaseSchema
+
+T = TypeVar("T", bound=BaseSchema)
+B = TypeVar("B", bound=BaseResponseModel)
+
+
+@dataclass
```
schustmi, line 64: This seems to be a redefinition of the class in `filter_models.py`?

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,383 @@
...
+        elif self.operation == GenericFilterOps.LT:
+            return getattr(table, self.column) < self.value
+
+
+# ---------------- #
+# PAGINATION PARAM #
+# -----------------#
+
+
+@dataclass
```
schustmi, line 177: Any specific reason for using `dataclass` here instead of making this a pydantic model as all the others?

src/zenml/models/page_model.py
AlexejPenner: jup its a duplicate that will be removed

src/zenml/models/filter_models.py
AlexejPenner: nope, copy pasted from somewhere

src/zenml/models/filter_models.py
AlexejPenner: i can move it there, (it will make the decorator a bit more involved to use as the Filter class would need to be passed into the decorator as a parameter

AlexejPenner: done

src/zenml/models/page_model.py
AlexejPenner: cleaned it, moved the offset into a property and removed all the other param-releated things from the code

AlexejPenner: cleaned it, moved the offset into a property and removed all the other param-releated things from the code

src/zenml/models/filter_models.py
AlexejPenner: its gone now

tests/unit/models/test_filter_models.py
```diff
@@ -0,0 +1,187 @@
...
+
+def test_filter_model_uuid_fields_equality_with_invalid_uuid_fails():
+    """Test filtering for equality with invalid UUID fails."""
+
+    with pytest.raises(ValueError):
+        uuid_value = "a92k34"
+        TestFilterModel(uuid_field=f"{GenericFilterOps.EQUALS}:{uuid_value}")
+
+
+def test_filter_model_string_fields_accept_correct_filter_ops():
```
fa9r, line 167: a lot of those tests are doing the same logic, perhaps you could abstract all this into a util `_test_filter_model_accept_correct_filter_ops(filter: Filter, filter_model: FilterBaseModel, filter_field: str, filter_value: Any)` that all these tests then just call with different args. That would make it much easier to write new tests for potential new filters in the future.

fa9r: also why do they all need to define custom filter models actually, can't they just use `TestFilterModel` for the successful checks as well?

src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -176,30 +191,16 @@ def get_stack(self, stack_id: UUID) -> StackResponseModel:
...
-            project_name_or_id: ID or name of the Project containing the stack
-            user_name_or_id: Optionally filter stacks by their owner
-            component_id: Optionally filter for stacks that contain the
-                          component
-            name: Optionally filter stacks by their name
-            is_shared: Optionally filter out stacks by whether they are shared
-                or not
-
+            stack_filter_model: All filter parameters including pagination
+                                params
```
fa9r, line 200: formatting off

fa9r: same for a lot of the docstrings below

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -585,6 +610,65 @@ def alembic(self) -> Alembic:
...
 
+    @classmethod
+    def filter_and_paginate(
+        cls,
+        session: Session,
+        query: Union[Select[AnySchema], SelectOfScalar[AnySchema]],
+        table: Type[AnySchema],
+        filter_model: FilterBaseModel,
+    ) -> Page[B]:
+        """Given a query, return aPage instance with a list of filtered Models.
```
fa9r, line 621:
```suggestion
        """Given a query, return a `Page` instance with a list of filtered models.
```

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+
+class FilterBaseModel(BaseModel):
+    """Class to unify all filter, paginate and sort request parameters in one place.
+
+    This Model allows fine-grained filtering, sorting and pagination of
+    resources.
+
+    Usage for a given Child of this class:
+    ```
+    ResourceListModel(
```
fa9r, line 257: really helpful to have this example here 👍

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -585,6 +610,65 @@ def alembic(self) -> Alembic:
...
+        """
+
+        # Filtering
+        filters = filter_model.generate_filter(table=table)
+
+        if filters is not None:
+            query = query.where(filters)
+
+        # Sorting
+        query = query.order_by(getattr(table, filter_model.sort_by))
```
fa9r, line 640: if you move this down you don'y need to set `query.order_by(None)` below I think

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -585,6 +610,65 @@ def alembic(self) -> Alembic:
...
+
+        # Get the total amount of pages in the database for a given query
+        total_pages = math.ceil(total / filter_model.size)
+
+        # Get a page of the actual data
+        item_schemas: List[AnySchema] = (
+            session.exec(
+                query.limit(filter_model.size).offset(filter_model.offset)
+            )
+            .unique()
```
fa9r, line 657: I suspect this hidden `unique` call might create some hard to debug issues in the future, what is this needed for actually?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -138,8 +157,14 @@
     UserRoleAssignmentSchema,
     UserSchema,
 )
+from zenml.zen_stores.schemas.base_schemas import BaseSchema
 
 AnyNamedSchema = TypeVar("AnyNamedSchema", bound=NamedSchema)
+AnySchema = TypeVar("AnySchema", bound=BaseSchema)
+B = TypeVar("B", bound=BaseResponseModel)
```
fa9r, line 164: why can't you use `BaseResponseModel` directly in the type annotations? That would make it more readable IMO

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -824,50 +908,27 @@ def get_stack(self, stack_id: UUID) -> StackResponseModel:
...
-                or not
+            stack_filter_model: All filter parameters including pagination
+                                params
 
         Returns:
             A list of all stacks matching the filter criteria.
         """
         with Session(self.engine) as session:
-            # Get a list of all stacks
+            # Manually create the query and add any custom clauses
```
fa9r, line 923: I would remove this comment, after your changes all the list commands feel very clean and at least to me it is obvious what they do / how they work now. Same for other list methods below

```suggestion
```

src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -59,23 +70,24 @@
...
+    project_filter_model: ProjectFilterModel = Depends(),
     _: AuthContext = Security(authorize, scopes=[PermissionType.READ]),
-) -> List[ProjectResponseModel]:
+) -> Page[ProjectResponseModel]:
     """Lists all projects in the organization.
 
     Args:
-        name: Optional name of the project to filter by.
+        project_filter_model: Filter model used for pagination, sorting,
+                              filtering
```
fa9r, line 85: docstring formatting off

src/zenml/cli/pipeline.py
```diff
@@ -52,19 +55,23 @@ def cli_pipeline_run(python_file: str, config_path: str) -> None:
...
 
 @pipeline.command("list", help="List all registered pipelines.")
-def list_pipelines() -> None:
+@list_options(PipelineFilterModel)
+def list_pipelines(**kwargs) -> None:
     """List all registered pipelines."""
     cli_utils.print_active_config()
-    pipelines = Client().list_pipelines()
+    client = Client()
+    with console.status("Listing roles...\n"):
```
fa9r, line 63:
```suggestion
    with console.status("Listing pipelines...\n"):
```

src/zenml/cli/pipeline.py
```diff
@@ -107,47 +114,25 @@ def runs() -> None:
...
-            pipeline_id = client.get_pipeline(pipeline).id
-        if user:
-            user_id = client.get_user(user).id
-        pipeline_runs = client.list_runs(
-            pipeline_id=pipeline_id,
-            stack_id=stack_id,
-            user_name_or_id=user_id,
-            unlisted=unlisted,
-        )
+        with console.status("Listing roles...\n"):
```
fa9r, line 125:
```suggestion
        with console.status("Listing pipeline runs...\n"):
```

src/zenml/cli/stack.py
```diff
@@ -30,9 +35,13 @@
     ProvisioningError,
     StackExistsError,
 )
+from zenml.models import StackFilterModel
 from zenml.utils.analytics_utils import AnalyticsEvent, track
 from zenml.utils.yaml_utils import read_yaml, write_yaml
 
+if TYPE_CHECKING:
+    pass
+
```
fa9r, line 44:
```suggestion
```

src/zenml/cli/utils.py
```diff
@@ -61,12 +72,11 @@
         PipelineRunResponseModel,
         StackResponseModel,
     )
-    from zenml.secret import BaseSecretSchema
-    from zenml.services import BaseService, ServiceState
-    from zenml.zen_server.deploy import ServerDeployment
 
 MAX_ARGUMENT_VALUE_SIZE = 10240
 
+M = TypeVar("M", bound=BaseResponseModel)
```
fa9r, line 78: Why can't you use `BaseResponseModel` directly?

src/zenml/cli/utils.py
```diff
@@ -1047,3 +1058,153 @@ def warn_unsupported_non_default_project() -> None:
...
+        )
+    elif filter_model.is_uuid_field(field):
+        return (
+            f"[UUID] Example: --{field}='{GenericFilterOps.STARTSWITH}:ab53ca' to "
+            f"filter for all UUIDs starting with that prefix."
+        )
+    elif filter_model.is_int_field(field):
+        return (
+            f"[INTEGER] Example: --{field}='{GenericFilterOps.STARTSWITH}:ab53ca' to "
+            f"filter for all UUIDs starting with that prefix."
```
fa9r, line 1104: this seems wrong for ints

src/zenml/cli/utils.py
```diff
@@ -1047,3 +1058,153 @@ def warn_unsupported_non_default_project() -> None:
...
+        )
+    elif filter_model.is_bool_field(field):
+        return (
+            f"[BOOL] Example: --{field}='{GenericFilterOps.STARTSWITH}:ab53ca' to "
+            f"filter for all UUIDs starting with that prefix."
```
fa9r, line 1109: also seems wrong for bools

src/zenml/cli/utils.py
```diff
@@ -1047,3 +1058,153 @@ def warn_unsupported_non_default_project() -> None:
...
+        )
+    elif filter_model.is_str_field(field):
+        return (
+            f"[STRING] Example: --{field}='{GenericFilterOps.CONTAINS}:example' to "
+            f"filter everything that contains the query somewhere in its "
+            f"{field}."
+        )
+
+
+def create_data_type_help_text(
```
fa9r, line 1119: can't we get this automatically somehow without this helper function?

src/zenml/cli/utils.py
```diff
@@ -1047,3 +1058,153 @@ def warn_unsupported_non_default_project() -> None:
...
+        return (
+            f"[STRING] supported filter operators: "
+            f"{[str(op) for op in StrFilter.ALLOWED_OPS]}"
+        )
+    else:
+        return f"{field}"
+
+
+def list_options(filter_model: Type[FilterBaseModel]) -> F:
+    """Create a decorator to generate the correct list of parameters to use for filtering."""
```
fa9r, line 1161: wow that's so cool

src/zenml/cli/pipeline.py
```diff
@@ -52,19 +55,23 @@ def cli_pipeline_run(python_file: str, config_path: str) -> None:
 
 
 @pipeline.command("list", help="List all registered pipelines.")
-def list_pipelines() -> None:
+@list_options(PipelineFilterModel)
+def list_pipelines(**kwargs) -> None:
```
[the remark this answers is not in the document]
fa9r: nvm, just saw `list_options` does that, in that case this change is just AWESOME

fa9r (changes requested):
What an awesome PR, very well designed. I really like the new abstractions and the new CLI, and I love how clean the ZenStore list commands are now!

Overall, I think the design is great, we just need to clean everything up. In particular, I was wondering whether we could somehow create filter models from one of the existing model types, similarly as we do with `@update_model` for update models. That would get rid of A LOT of boilerplate code.

Some more high-level suggestions:
- one thing that is unintuitive to me is that the filter args are always called `project_id` / `user_id` but request and update models have `project` / `user` even though those are also just UUIDs. Should we rename this for the request and update models as well then?
- post execution feels like it might still require a few more changes, but I guess our new integration tests will reveal potential issues there
- docstring indents are off in various places, subsequent lines should continue with a single indent usually
- CLI changes look great but make sure to manually test them thoroughly since quite a lot has changed there

I also left quite a lot of detailed suggestions below, let me know if you need help with some of the refactoring, really looking forward to get this merged!

src/zenml/cli/utils.py
AlexejPenner: The only alternative I saw was mobing this onthe the BaseFilterModel but that feels more wrong, as its a cli concern

src/zenml/models/artifact_models.py
```diff
@@ -65,6 +67,45 @@ class ArtifactResponseModel(ArtifactBaseModel, ProjectScopedResponseModel):
...
+# ------ #
+# FILTER #
+# ------ #
+
+
+class ArtifactFilterModel(ProjectScopedFilterModel):
+    """Model to enable advanced filtering of all Artifacts."""
+
+    name: str = Query(
+        None,
```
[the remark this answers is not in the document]
AlexejPenner: Ctrl+Shift+F and Ctrl+Shift+R made this super easy :)

src/zenml/models/base_models.py
```diff
@@ -84,7 +91,7 @@ class UserScopedResponseModel(BaseResponseModel):
     Used as a base class for all domain models that are "owned" by a user.
     """
 
-    user: Optional["UserResponseModel"] = Field(
+    user: Union["UserResponseModel", None] = Field(
```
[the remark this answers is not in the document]
AlexejPenner: This was mysteriously failing in some docker environments (most probably due to some specific package version having issues) - this change solved these issues

src/zenml/models/base_models.py
```diff
@@ -45,6 +47,11 @@ class BaseResponseModel(AnalyticsTrackedModelMixin):
     created: datetime = Field(title="Time when this resource was created.")
     updated: datetime = Field(title="Time when this resource was last updated.")
 
+    class Config:
+        """Allow extras on Response Models"""
```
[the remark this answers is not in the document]
AlexejPenner: I had a reason last year, I wish I remembered it

src/zenml/models/base_models.py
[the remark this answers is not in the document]
fa9r: wat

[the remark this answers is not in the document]
fa9r: Perhaps mypy master Michael @schustmi knows why? 😄

src/zenml/post_execution/pipeline.py
```diff
@@ -123,9 +123,9 @@ def get_pipeline(
         name=pipeline_name,
         project_name_or_id=active_project_id,
     )
-    if len(pipeline_models) == 1:
+    if pipeline_models.total == 1:
```
schustmi, line 126: This change made me think that it might be a good idea to implement the `__len__` method on that class so that we can use the previous syntax?

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -1951,6 +1933,40 @@ def _get_resource(
         body = self.get(f"{route}/{str(resource_id)}")
         return response_model.parse_obj(body)
 
+    def _list_paginated_resources(
+        self,
+        route: str,
+        response_model: Type[AnyResponseModel],
+        list_model: FilterBaseModel,
```
schustmi, line 1940: Is `filter_model` a better name here?

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,697 @@
...
+        else:
+            return getattr(table, self.column) == self.value
+
+
+# ---------------- #
+# PAGINATION PARAM #
+# -----------------#
+
+
+class FilterBaseModel(BaseModel):
```
schustmi, line 277: Maybe `BaseFilterModel` to keep it in line with our `BasePipeline` etc?

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+
+    id: Union[UUID, str] = Query(None, description="Id for this resource")
+    created: Union[datetime, str] = Query(None, description="Created")
+    updated: Union[datetime, str] = Query(None, description="Updated")
+
+    class Config:
+        """Configure all Filter Models."""
+
+        extras = False
+        fields = {"list_of_filters": {"exclude": True}}
```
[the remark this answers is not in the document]
schustmi: It excludes `list_of_filters` from `.dict()` and `.json()` calls if I remember correctly

src/zenml/models/artifact_models.py
```diff
@@ -13,9 +13,10 @@
 #  permissions and limitations under the License.
 """Models representing artifacts."""
 
-from typing import Optional
+from typing import Optional, Union
 from uuid import UUID
 
+from fastapi import Query
```
schustmi, line 19: I find it a little strange that our generic model classes (which are also used in the SQLZenStore) now depend on some fastapi class. Can you explain why this is necessary?

schustmi (changes requested):
What a great PR! I did a quick scan and only left a few minor comments, but the overall implementation looks amazing.

One more additional question: Now that our models are always hydrated, how do we handle cases like a `PipelineModel` which includes `runs`. Is the plan to add pagination for those sub-lists as well, or do we somehow limit those to just a few elements?

src/zenml/zen_stores/zen_store_interface.py
fa9r: fixed

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+    ) -> Union["BinaryExpression", "BooleanClauseList"]:
+        """Concatenate all filters together with the chosen operator."""
+        from sqlalchemy import and_
+
+        user_created_filter = self._base_filter(table=table)
+        scope_filter = self._scope_filter(table=table)
+        if scope_filter is not None:
+            return and_(user_created_filter, scope_filter)
+        else:
+            return user_created_filter
```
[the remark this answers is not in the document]
fa9r: I refactored these methods as I outlined above

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+# -----------------#
+
+
+class FilterBaseModel(BaseModel):
+    """Class to unify all filter, paginate and sort request parameters in one place.
+
+    This Model allows fine-grained filtering, sorting and pagination of
+    resources.
+
+    Usage for a given Child of this class:
```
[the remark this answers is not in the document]
fa9r: Changed

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+            scope_filter.append(
+                or_(
+                    getattr(table, "user_id") == self._scope_user,
+                    getattr(table, "is_shared") is True,
+                )
+            )
+        if self._scope_project:
+            scope_filter.append(
+                getattr(table, "project_id") == self._scope_project
+            )
```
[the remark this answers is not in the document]
fa9r: Refactored as outlined above

src/zenml/models/filter_models.py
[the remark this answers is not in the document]
fa9r: this is deprecated since `list_of_filters` is no longer a field of this model; deleted the custom config accordingly.

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+            return getattr(table, self.column) > self.value
+        elif self.operation == GenericFilterOps.LTE:
+            return getattr(table, self.column) <= self.value
+        elif self.operation == GenericFilterOps.LT:
+            return getattr(table, self.column) < self.value
+        else:
+            return getattr(table, self.column) == self.value
+
+
+class DatetimeFilter(Filter):
```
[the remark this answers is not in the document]
fa9r: Deleted `DatetimeFilter` in favor of `NumericFilter`

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+            return getattr(table, self.column).like(f"%{self.value}%")
+        elif self.operation == GenericFilterOps.STARTSWITH:
+            return getattr(table, self.column).startswith(f"%{self.value}%")
+        elif self.operation == GenericFilterOps.CONTAINS:
+            return getattr(table, self.column).endswith(f"%{self.value}%")
+        else:
+            return getattr(table, self.column) == self.value
+
+
+class UUIDFilter(Filter):
```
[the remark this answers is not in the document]
fa9r: Refactored this to make `Filter` subclasses override `generate_query_conditions_from_column` instead. This allows us to use inheritance and also gets rid of a lot of repeated `getattr(table, self.column)` calls

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+                    StrFilter(
+                        operation=GenericFilterOps(operator),
+                        column=key,
+                        value=str(value),
+                    )
+
+        return values
+
+    @property
+    def list_of_filters(self) -> List[Filter]:
```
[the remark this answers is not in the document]
fa9r: Split it up into several smaller methods as suggested above

src/zenml/models/filter_models.py
```diff
@@ -0,0 +1,650 @@
...
+        page=2,
+        size=50
+    )
+    ```
+    """
+
+    FILTER_EXCLUDE_FIELDS: ClassVar[List[str]] = [
+        "sort_by",
+        "_scope_user",
+        "_scope_project",
```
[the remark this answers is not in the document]
fa9r: redefined these class vars in subclasses

tests/unit/models/test_filter_models.py
```diff
@@ -0,0 +1,187 @@
...
+def test_filter_model_sort_by_for_existing_field_succeeds(
+    correct_sortable_column: Any,
+):
+    """Test that the filter model sort_by field enforces valid filter fields"""
+    FilterBaseModel(sort_by=correct_sortable_column)
+    assert FilterBaseModel(sort_by=correct_sortable_column).sort_by
+
+
+@pytest.mark.parametrize(
+    "correct_sortable_column", ["catastic_column", "zenml", "page", 1]
```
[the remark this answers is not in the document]
fa9r: Adjusted

tests/unit/models/test_filter_models.py
```diff
@@ -0,0 +1,187 @@
...
+def test_filter_model_sort_by_for_non_filter_fields_fails(
+    correct_sortable_column: Any,
+):
+    """Test that the filter model sort_by field enforces valid filter fields"""
+    with pytest.raises(ValidationError):
+        FilterBaseModel(sort_by=correct_sortable_column)
+
+
+def test_filter_model_datetime_fields_accept_correct_filter_ops():
+    """Test that the flavor base model fails with long names."""
```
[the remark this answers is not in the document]
fa9r: Fixed docstrings

tests/unit/models/test_filter_models.py
```diff
@@ -0,0 +1,187 @@
...
+
+@pytest.mark.parametrize(
+    "correct_sortable_column", ["created", "updated", "id"]
+)
+def test_filter_model_sort_by_for_existing_field_succeeds(
+    correct_sortable_column: Any,
+):
+    """Test that the filter model sort_by field enforces valid filter fields"""
+    FilterBaseModel(sort_by=correct_sortable_column)
+    assert FilterBaseModel(sort_by=correct_sortable_column).sort_by
```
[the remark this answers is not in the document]
fa9r: Outdated

tests/unit/models/test_filter_models.py
fa9r: Abstracted filter model test logic into helper method `_test_filter_model` as outlined above

src/zenml/zen_stores/rest_zen_store.py
fa9r: Agreed, changed it 👍

src/zenml/models/filter_models.py
fa9r: Makes sense, renamed it 👍

src/zenml/post_execution/pipeline.py
fa9r: The question here is would you expect `len(Page)` to return the total number of items available, or the number of items in that specific page? Probably the number of items in the page, right? In that case, using `len` for this check could lead to issues if users set the page size limit to 1 for some reason.

fa9r: I added a couple of container methods to `Page` so we can now do `page[0]`, `x in page`, `len(page)`, `[a for a in page]`

src/zenml/zen_stores/sql_zen_store.py
fa9r: All removed already

src/zenml/post_execution/pipeline.py
schustmi: Yeah I was thinking of just getting the amount of items in that page. But you're right, the `total` property is still needed and also the right one to use in this case.

src/zenml/zen_stores/sql_zen_store.py
schustmi: The logic would still work with `BaseResponseModel`, but the type hints would be much worse:
For example

```python
class Base:
  def do_something():
    pass

class A(Base):
  pass

class B(Base):
  pass

def f1(item: Base) -> Base:
  item.do_something()
  return item

B = TypeVar("B", bound=Base)

def f2(item: B) -> B:
  item.do_something()
  return item

a = A()
f1(a)  # the return value will have type `Base`
f2(a)  # the return value will have type `A`
```

src/zenml/models/component_models.py
```diff
@@ -65,6 +80,76 @@ class ComponentResponseModel(ComponentBaseModel, ShareableResponseModel):
...
+        description="Flavor of the stack component",
+    )
+    type: str = Query(
+        default=None,
+        description="Type of the stack component",
+    )
+    project_id: Union[UUID, str] = Query(
+        default=None, description="Project of the stack"
+    )
+    user_id: Union[UUID, str] = Query(None, description="User of the stack")
```
fa9r, line 125: @AlexejPenner Why does this have both `_scope_type` and `type`? Similarly, I noticed many project-scoped filter models like this one also have both `project_id`, `user_id` and `_scope_project`, `_scope_user`. I didn't fully understand yet when you would use which here.

AlexejPenner: Its two different things the scope is always a separate circle in the Venn diagram of conditions (so Scope ∩ Filter is what you get) - However if you want to filter with an or operator within the filter you can always do this with type (e.g Filter=(type  ∪ flavor ∪ is_shared).

fa9r: Ok makes sense, I think. So the point is that project/user scoped resources usually always require an AND filter on those entities and should still allows OR filters on other fields, but perhaps in some other queries you don't have this scoping and want to also include project/user in an OR filter?

src/zenml/models/artifact_models.py
fa9r: @AlexejPenner

src/zenml/zen_stores/sql_zen_store.py
fa9r: @AlexejPenner

src/zenml/models/base_models.py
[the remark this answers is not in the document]
fa9r: 🤷

PR #1115, opened 2022-11-28
Add failed Status to component when step fails

safoinme:
## Describe changes
This PR adds step_failed to base stack_component so it will be used to pass a failing state of a step run to integration. For example, when a step fails, the MLFlow experiment tracker will show the run as failed rather than completed. 

## Pre-requisites

## Types of changes


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -376,6 +376,7 @@ def run_step(self, step: "Step") -> Optional[data_types.ExecutionInfo]:
                 execution_info = self._execute_step(component_launcher)
             except:  # noqa: E722
                 self._publish_failed_run(run_name_or_id=run_model.id)
+                stack.cleanup_step_run(info=step_run_info, step_failed=True)
```
fa9r, line 379: this try-except logic might change a bit after #1103, @schustmi will the general flow of calling `stack.cleanup_step_run(info=step_run_info, step_failed=True)` after failures still work with our changes?

schustmi: I think that should still work as expected yes

schustmi: Ah I just saw, in general it will work but the current implementation has a flaw:
In case the step execution fails, it would call `cleanup_step_run` twice, first with `step_failed=True` followed by `step_failed=False` (due to the `finally` clause which always gets executed.

Two potentials solutions:

```python
step_failed = False
try:
  ...
except:
  self._publish_failed_run(run_name_or_id=run_model.id)
  step_failed = True
finally:
  self.cleanup_step_run(info=step_run_info, step_failed=step_failed)
```

```python
try:
  ...
except:
  self._publish_failed_run(run_name_or_id=run_model.id)
  self.cleanup_step_run(info=step_run_info, step_failed=True)
else:
  self.cleanup_step_run(info=step_run_info, step_failed=False)
```

schustmi (changes requested):

src/zenml/orchestrators/base_orchestrator.py
safoinme: @schustmi what about including this in `StepRunInfo` as @htahir1 suggested or is `StepRunInfo` only for the information to run the step not the state of the run?

AlexejPenner (approved):
:seal: of approval

schustmi (approved):

schustmi (changes requested):
Actually there is another call to `stack.cleanup_step_run(...)` happening in the step operator entrypoint (https://github.com/zenml-io/zenml/blob/dd876c4e989d0026a2504229525ffa73222f2bda/src/zenml/step_operators/step_operator_entrypoint_configuration.py#L114-L117) which also needs to be adjusted.

I would also suggest removing the default value `step_failed=False` on the stack method, in which case linting would have caught this error

htahir1 (commented):
I would also wait for neptune to be merged and add a case for that

safoinme:
@htahir1 I don't think that would be required, Neptune seems to be detecting the state of the run and declare run as failed if the stop wasn't called manually which is the case only when step fails

schustmi:
I think it's mostly so the neptune PR doesn't need any updates, we can instead merge neptune first and then solve the issues on this branch which makes it easier for them.

docs/book/component-gallery/experiment-trackers/experiment-trackers.md
```diff
@@ -75,6 +75,9 @@ independently of ZenML
 * finally, you can access the Experiment Tracker UI to browse and visualize the
 information logged during your pipeline runs
 
+Note: the Expirement Tracker will declare run as failed if the pipeline step
```
schustmi, line 78: I'd personally remove this note (not crucial information) as it doesn't apply to all the experiment trackers (neptune)

schustmi (approved):
LGTM! Just a minor comment, but feel free to ignore

PR #1117, opened 2022-11-29
Improving error messages when fetching entitites

bcdurak:
## Describe changes
I implemented a minor fix that improves the error messages when fetching entities. There is a new KeyError that allows error messages with line breaks.

## Pre-requisites

## Types of changes


schustmi:
My personal opinion on this: I think code clarity (in this case using well known errors like the `KeyError` from the standard library in code/the API docs) is much more important than a slightly more readable error message.

bcdurak:
@schustmi You are right.

The newly added error is still a `KeyError` though, so programmatically the `try-except KeyError`s will still function but it is less readable now. 

They are solving a very specific problem that can be critical in the user flow. At the same time, with the addition of our new filtering approach, these changes may become redundant. 

Not really sure how to move forward here.

schustmi (approved):
LGTM

PR #1123, opened 2022-11-30
Introduce username and password to kubeflow for more native multi-tenant support

htahir1:
## Describe changes
I moved Kubeflow Multi Tenancy username password flow into the orchestrator itself. 

TODO:

- [x] Catch error and handle case wrong username and password are given
- [x] Catch case when its multi-tenant deployment  but username and password (or client_args) are not given

## Pre-requisites

## Types of changes


src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -780,6 +781,28 @@ def _get_kfp_client(
...
+        # Handle username and password, ignore the case if one is passed and not the other
+        # Also do not attempt to get cookie if cookie is already passed in client_args
+        if (
+            settings.client_username is not None
+            and settings.client_password is not None
+        ):
+            # If cookie is already set, then ignore
+            if "cookie" in client_args:
+                logger.warning(
+                    "Cookie already set in `client_args` so ignoring `client_username` and `client_password`..."
```
schustmi, line 793:
```suggestion
                    "Cookie already set in `client_args`, ignoring `client_username` and `client_password`..."
```

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -780,6 +781,28 @@ def _get_kfp_client(
...
+                )
+            else:
+                logger.info(
```
schustmi, line 796: To clean this up a little, I would move this log and the one below into the `_get_session_cookie` method

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +867,54 @@ def _get_kfp_ui_daemon_port(self) -> int:
             port = networking_utils.find_available_port()
         return port
 
+    def _get_session_cookie(self, username: str, password: str) -> str:
+        session = requests.Session()
+        if (
+            self.config.kubeflow_hostname is None
+            or "/pipeline" not in self.config.kubeflow_hostname
+        ):
+            raise AssertionError(
```
schustmi, line 876:
```suggestion
            raise ValueError(
```

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +867,54 @@ def _get_kfp_ui_daemon_port(self) -> int:
             port = networking_utils.find_available_port()
         return port
 
+    def _get_session_cookie(self, username: str, password: str) -> str:
+        session = requests.Session()
+        if (
+            self.config.kubeflow_hostname is None
```
schustmi, line 873: Are we 100% sure that this needs to be the case and can't be different in some custom deployments? We already log a warning in the config when setting the hostname, not sure it's necessary to always fail here?

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +867,54 @@ def _get_kfp_ui_daemon_port(self) -> int:
...
+                f"Error while trying to fetch kubeflow cookie: {errc}"
+            )
+        except requests.exceptions.Timeout as errt:
+            raise RuntimeError(
+                f"Error while trying to fetch kubeflow cookie: {errt}"
+            )
+        except requests.exceptions.RequestException as err:
+            raise RuntimeError(
+                f"Error while trying to fetch kubeflow cookie: {err}"
+            )
```
schustmi, line 902:
```suggestion
        except (requests.exceptions.HTTPError, requests.exceptions.ConnectionError, requests.exceptions.Timeout,  requests.exceptions.RequestException) as e:
            raise RuntimeError(
                f"Error while trying to fetch kubeflow cookie: {e}"
            )
```

This will need some formatting, but you can catch all of them at once like this

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -41,6 +42,8 @@ class KubeflowOrchestratorSettings(BaseSettings):
             specified on steps.
         timeout: How many seconds to wait for synchronous runs.
         client_args: Arguments to pass when initializing the KFP client.
+        client_username: Username to use for the kubeflow client. To be used together with `client_password`.
```
schustmi, line 45: I think this should mention that we're trying to generate the cookie from those to (and that they both need to be set for this to work).

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -116,6 +121,37 @@ def _migrate_pod_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
             values["node_affinity"] = {}
             values["node_selectors"] = {}
```
schustmi, line 123: Maybe we can put this into a separate root validator so it's a little cleaner? If not, the method name should change at least

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -116,6 +121,37 @@ def _migrate_pod_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
 
+        client_creds_error = AssertionError(
```
schustmi, line 124:
```suggestion
        client_creds_error = ValueError(
```

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -116,6 +121,37 @@ def _migrate_pod_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                    "It looks like you have specified a `kubeflow_hostname` but it doesn't seem to "
+                    "end with `/pipeline`, which is not usual. Please ensure this is intended."
+                )
+
+            if (
+                values.get("client_username") is None
+                and values.get("client_password") is None
+                and values.get("client_args") is None
+            ):
+                logger.warning(
```
schustmi, line 150: This assumes that setting a kubeflow hostname necessarily means multi-tenant deployment. Is that actually the case?

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -116,6 +121,37 @@ def _migrate_pod_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
             values["node_affinity"] = {}
             values["node_selectors"] = {}
 
+        client_creds_error = AssertionError(
+            "`client_username` and `client_password` both need to be set."
+        )
+        if (
```
schustmi, line 127: There are a lot of `values.get("client_username")` and `values.get("client_password")` down below. I would suggest calling it once up here and then using the variables:
```python
username = values.get("client_username")
password = values.get("client_password")

if username and password is None:
  ...
```

schustmi (changes requested):

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -116,6 +121,37 @@ def _migrate_pod_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+        ):
+            raise client_creds_error
+        if (
+            values.get("client_password")
+            and values.get("client_username") is None
+        ):
+            raise client_creds_error
+
+        if values.get("kubeflow_hostname"):
+            if "/pipeline" not in values.get("kubeflow_hostname"):
```
schustmi, line 139: This should probably also be `hostname.endswith("/pipeline")`?

htahir1: Yeah but it might also be `/pipeline/` so thats why I just made it easier. WDYT?

schustmi: Good point, makes sense 👍

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
htahir1: I did not want to get into the area of which root validator is called first, so I have for now just renamed the method

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
htahir1: You're right. I replaced this with just the check that the hostname needs to exist

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
htahir1: Yes this is why its just a warning and it says "You see to be...". However, I have changed the error message now to make it not tied to multi-tenant

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -52,13 +57,29 @@ class KubeflowOrchestratorSettings(BaseSettings):
...
+
+        Args:
+            values: Dict representing user-specified runtime settings.
+
+        Returns:
+            dict: Validated settings.
+            
+        Raises:
+            AssertionError: If old and new settings are used together.
+            client_creds_error: If username and password are not specified together.
```
schustmi, line 81:
```suggestion
            ValueError: If username and password are not specified together.
```

src/zenml/integrations/kubeflow/flavors/kubeflow_orchestrator_flavor.py
```diff
@@ -52,13 +57,29 @@ class KubeflowOrchestratorSettings(BaseSettings):
...
+    def _validate_and_migrate_pod_settings(
+        cls, values: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        """Validates settings and migrates pod settings from older version.
+
+        Args:
+            values: Dict representing user-specified runtime settings.
+
+        Returns:
+            dict: Validated settings.
```
schustmi, line 77:
```suggestion
            Validated settings.
```

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +868,72 @@ def _get_kfp_ui_daemon_port(self) -> int:
...
 
+    def _get_session_cookie(self, username: str, password: str) -> str:
+        """Gets session cookie from username and password.
+
+        Args:
+            username: Username for kubeflow host.
+            password: Password for kubeflow host.
+
+        Raises:
+            ValueError: Raised when trying to fetch cookie.
```
schustmi, line 879:
```suggestion
            RuntimeError: If the cookie fetching failed.
```

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +868,72 @@ def _get_kfp_ui_daemon_port(self) -> int:
...
+            RuntimeError: Raised when trying to fetch cookie.
+            RuntimeError: Raised when trying to fetch cookie.
+            ProvisioningError: Raised when trying to fetch cookie.
+            ProvisioningError: Raised when trying to fetch cookie.
```
schustmi, line 883:
```suggestion
```

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +868,72 @@ def _get_kfp_ui_daemon_port(self) -> int:
...
+
+        Returns:
+            str: Cookie with the prefix `authsession=`.
+        """
+        if self.config.kubeflow_hostname is None:
+            raise ValueError(
```
schustmi, line 889:
```suggestion
            raise RuntimeError(
```

So the function just raises a single error type, easier in the docstring and also to catch outside

src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py
```diff
@@ -844,6 +868,72 @@ def _get_kfp_ui_daemon_port(self) -> int:
...
+
+        Raises:
+            ValueError: Raised when trying to fetch cookie.
+            RuntimeError: Raised when trying to fetch cookie.
+            RuntimeError: Raised when trying to fetch cookie.
+            ProvisioningError: Raised when trying to fetch cookie.
+            ProvisioningError: Raised when trying to fetch cookie.
+
+        Returns:
+            str: Cookie with the prefix `authsession=`.
```
schustmi, line 886:
```suggestion
            Cookie with the prefix `authsession=`.
```

schustmi (changes requested):
Nice, looks much cleaner now. Just a few minor comments (mostly related to stuff that will fail in the CI anyway)

docs/book/component-gallery/orchestrators/kubeflow.md
```diff
@@ -291,7 +291,7 @@ zenml orchestrator register <NAME> \
     --kubeflow_hostname=<KUBEFLOW_HOSTNAME> # e.g. https://mykubeflow.example.com/pipeline
 ```
 
-Then, ensure that you use the pass the right settings before triggerling a pipeline run. The following snipper will prove useful:
+Then, ensure that you use the pass the right settings before triggerling a pipeline run. The following snippet will prove useful:
```
bcdurak, line 294:
```suggestion
Then, ensure that you use the pass the right settings before triggering a pipeline run. The following snippet will prove useful:
```

bcdurak (changes requested):

schustmi (approved):

PR #1125, opened 2022-12-01
Bugfix/fix endpoints for dashboard development

AlexejPenner:
## Describe changes
Talha ran into some issues with our endpoints during Dashboard development:
This PR does the following:

1) Add a Pipeline runs endpoint at the project scope
2) Fix the user/<ID>/roles endpoint that had faulty typing
3) Take some fixes from Stefans test branch to make sure they get merged into this release (namely the fixes on the role-assignment endpoints)

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/routers/projects_endpoints.py
+++ b/src/zenml/zen_server/routers/projects_endpoints.py
@@ -511,48 +511,6 @@ def list_project_pipelines(
     )
 
 
-@router.get(
-    "/{project_name_or_id}" + PIPELINES,
-    response_model=List[PipelineRunResponseModel],
-    responses={401: error_response, 404: error_response, 422: error_response},
-)
-@handle_exceptions
-def list_runs(
-    project_name_or_id: Union[str, UUID],
-    stack_id: Optional[UUID] = None,
-    run_name: Optional[str] = None,
-    user_name_or_id: Optional[Union[str, UUID]] = None,
-    component_id: Optional[UUID] = None,
-    pipeline_id: Optional[UUID] = None,
-    unlisted: bool = False,
-    _: AuthContext = Security(authorize, scopes=[PermissionType.READ]),
-) -> List[PipelineRunResponseModel]:
-    """Get pipeline runs according to query filters.
-
-    Args:
-        project_name_or_id: Name or ID of the project for which to filter runs.
-        stack_id: ID of the stack for which to filter runs.
-        run_name: Filter by run name if provided
-        user_name_or_id: If provided, only return runs for this user.
-        component_id: Filter by ID of a component that was used in the run.
-        pipeline_id: ID of the pipeline for which to filter runs.
-        unlisted: If True, only return unlisted runs that are not
-            associated with any pipeline.
-
-    Returns:
-        The pipeline runs according to query filters.
-    """
-    return zen_store().list_runs(
-        project_name_or_id=project_name_or_id,
-        name=run_name,
-        stack_id=stack_id,
-        component_id=component_id,
-        user_name_or_id=user_name_or_id,
-        pipeline_id=pipeline_id,
-        unlisted=unlisted,
-    )
-
-
 @router.post(
     "/{project_name_or_id}" + PIPELINES,
     response_model=PipelineResponseModel,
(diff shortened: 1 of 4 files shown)


src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -511,6 +511,48 @@ def list_project_pipelines(
     )
 
 
+@router.get(
+    "/{project_name_or_id}" + PIPELINES,
```
schustmi, line 515: Is this really the correct endpoint to list runs?

src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -511,6 +511,48 @@ def list_project_pipelines(
...
+    response_model=List[PipelineRunResponseModel],
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+@handle_exceptions
+def list_runs(
+    project_name_or_id: Union[str, UUID],
+    stack_id: Optional[UUID] = None,
+    run_name: Optional[str] = None,
```
schustmi, line 523: I think we had a problem on develop this week where we realized that our list endpoint need to have an argument called `name` (instead of the `run_name` here. I'm guessing that's also relevant here @bcdurak?

schustmi (changes requested):

src/zenml/zen_server/routers/projects_endpoints.py
AlexejPenner: good catch

src/zenml/zen_server/routers/projects_endpoints.py
AlexejPenner: Would make sense because else our serialization at the rest_zen_store will try filtering by name but the endpoint has run_name as the keyword arg. Will include this as well

src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -555,6 +555,49 @@ def create_pipeline(
...
 
 
+
+@router.get(
+    "/{project_name_or_id}" + RUNS,
+    response_model=List[PipelineRunResponseModel],
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+@handle_exceptions
+def list_runs(
```
fa9r, line 565: I don't get why this needs to exists, we already have an endpoint that does exactly the same

fa9r (changes requested):

src/zenml/zen_server/routers/projects_endpoints.py
AlexejPenner: basically we have all list and creates on resources twice, once on the project resource and once on its own. In general now that we are going for project scoping on the dashboard side, all the list queries are explicitly moving to the project scoped endpoints. Thats where we realized that the runs endpoint was missing. This change just makes runs behave like all other resources.

fa9r: instead of calling `/projects/{project_id}/runs` he can just use `/runs?project_name_or_id={project_id}`

src/zenml/zen_server/routers/projects_endpoints.py
AlexejPenner: done

src/zenml/zen_server/routers/projects_endpoints.py
fa9r: the issue I see is that we have those functions duplicated now which opens the door to a lot of bugs and inconsistencies IMO

AlexejPenner: theoretically yes, difference being 
`/projects/{project_id}/runs` is basically `project_id: UUID` 
while
`/runs?project_name_or_id={project_id}` equates to  `project_id: Optional[UUID] = None`

As we are slowly transitioning to real role based access control I believe it makes sense to enforce the scope now already to make the transition earlier, then we can let Fastapi do the heavy lifting when the time comes.

the /project endpoints can then be scoped to users who have the right permissions in that particular project

fa9r: i.e., the RestZenStore methods hit a different endpoint than the dashboard now so the CLI and dashboard might show different results for the same request

AlexejPenner: My proposal would be to have duplicates for the time being until projects are firmly established as the scoping mechanism in frontend and clientside, then I would propose to deprecate the duplicates unscoped endpoints

fa9r (approved):

schustmi (approved):

PR #1130, opened 2022-12-02
Fix Neptune run cleanup

safoinme:
## Describe changes
Fix Cleanup so Neptune can initialize a run at step level and stop it once the the step run is finished instead of waiting until the last step of the pipeline.

## Pre-requisites

## Types of changes


src/zenml/integrations/neptune/experiment_trackers/neptune_experiment_tracker.py
```diff
@@ -99,12 +99,11 @@ def prepare_step_run(self, info: "StepRunInfo") -> None:
         self.run_state.tags = list(settings.tags)
 
     def cleanup_step_run(self, info: "StepRunInfo", step_failed: bool) -> None:
-        """If the current step is the last step of the pipeline, stop the Neptune run.
+        """stop the Neptune run.
```
schustmi, line 102:
```suggestion
        """Stop the Neptune run.
```

src/zenml/integrations/neptune/experiment_trackers/neptune_experiment_tracker.py
```diff
@@ -99,12 +99,11 @@ def prepare_step_run(self, info: "StepRunInfo") -> None:
...
 
         Args:
             info: Info about the step that was executed.
             step_failed: Whether the step failed or not.
         """
-        if self._is_last_step(info):
-            self.run_state.active_run.sync()
-            self.run_state.active_run.stop()
+        self.run_state.active_run.sync()
```
schustmi, line 108: I think now that the `_is_last_step()` method is not called here we can probably remove it?

src/zenml/integrations/neptune/experiment_trackers/run_state.py
```diff
@@ -115,20 +115,22 @@ def tags(self, tags: List[str]) -> None:
...
-        if self._active_run is None:
-            run = neptune.init_run(
-                project=self.project,
-                api_token=self.token,
-                custom_run_id=hashlib.md5(self.run_name.encode()).hexdigest(),  # type: ignore
-                tags=self.tags,
-            )
-            run[_INTEGRATION_VERSION_KEY] = zenml.__version__
-            self._active_run = run
+        run = neptune.init_run(
```
schustmi, line 126: Can you explain why this change was necessary and we couldn't re-use the previously created run?

I'm asking because users will also call `run_state.active_run` in their step code (and might do so many times), so recreating the run each time might not be optimal.

Maybe as an alternative suggestion (not sure if this might work):
- The logic in `active_run` stays as before, only creating it if it doesn't exists
- In the experiment tracker `cleanup_step_run(...)`, we call `run_state.reset_run()` which just sets `self._active_run = None` to clean up the run and create a new one next time its needed?

schustmi (changes requested):

src/zenml/integrations/neptune/experiment_trackers/run_state.py
safoinme: There was a problem with stopping a Neptune run and we try to use it again to log with it:
```
neptune.new.exceptions.InactiveRunException: 

----InactiveRunException----------------------------------------

It seems you are trying to log metadata to (or fetch it from) a run that was stopped (ZEN-20).

Here's what you can do:
    - Resume the run to continue logging to it:
    https://docs.neptune.ai/logging/to_existing_object/
    - Don't invoke `stop()` on a run that you want to access. If you want to stop monitoring only,
    you can resume a run in read-only mode:
    https://docs.neptune.ai/api/connection_modes/#read-only-mode

You may also want to check the following docs pages:
    - https://docs.neptune.ai/logging/to_existing_object/
    - https://docs.neptune.ai/usage/querying_metadata/

Need help?-> https://docs.neptune.ai/getting_help
```

schustmi: Do you think my suggestion (resetting the `active_run` once we stopped it) solves this without always creating it?

schustmi (approved):
LGTM!

PR #1136, opened 2022-12-05
Ensure model string fields have a max length

strickvl:
I added `max_length` into the string fields in our models where it made sense. Also tests to validate / confirm the changes.

## Pre-requisites

## Types of changes


src/zenml/models/constants.py
```diff
@@ -19,6 +19,12 @@
 MODEL_DESCRIPTIVE_FIELD_MAX_LENGTH = 300
 # The maximum length of the config schema of a flavor.
 MODEL_CONFIG_SCHEMA_MAX_LENGTH = 10000
+# The maximum length of an URI string field in models.
+MODEL_URI_FIELD_MAX_LENGTH = 1000
+# The maximum length of a metadata field in models.
+MODEL_METADATA_FIELD_MAX_LENGTH = 200
```
htahir1, line 25: Is this enough for metadata? Are these 'estimated' by gut these numbers?

htahir1 (commented):
One comment/question otherwise LGTM

src/zenml/models/constants.py
strickvl: Metadata maybe was the wrong name. These are things like 'data_type' or 'materializer' name. They're longer than the actual 'name' max_length. Beyond that logic, no special estimations in there...

Do you think I should increase this value?

htahir1: I would increase it to 512 or something (powers of 2 are somehow common in these, maybe because of the byte allocation at the end). regarding name Maybe you can say `MODEL_PROPERTY_FIELD_MAX_LENGTH`?

strickvl: Updated following this feedback / following a conversation with @AlexejPenner

fa9r: Also, instead of defining a separate constants for each field you could perhaps just have a `TEXT_FIELD_MAX_LENGTH = 65,535` and `STR_FIELD_MAX_LENGTH = 255` and then assign those where applicable

fa9r (changes requested):
Nice, looks great! Love that you even added so many new tests for these!

The only suggestion I have is to tie the max length in the models to the DB constraint, as Alexej already suggested, and to (perhaps) share some of the constants then.

src/zenml/models/constants.py
strickvl: Ok I'm probably being dumb here but are you just saying that we should have the same number referenced in both places, @fa9r? AFAICS there is only one place where we explicitly say 4096 as the length and then elsewhere it's implicit max length of 65K because it's a `TEXT` value.

AlexejPenner: I think whats meant is that we should have one constant for lenght = 65K and set this as the max lenght for all fields that will be TEXT in the DB and 255 as max lenght for all other str fields (except where strictly necessary to be different

strickvl: Ok I think I addressed all this now in the changes I just made :)

fa9r (approved):
LGTM :)

AlexejPenner (approved):
:seal:

htahir1 (approved):
Ill approve this now as the others have as well and I dont know more than them about this part of the codebase :-)

Issue #1135, opened 2022-12-05
[BUG]: HuggingFace dataset materializer problem


ghpu:
### System Information

zenml v0.30rc0 (applies also to v0.22 and v0.23)

ZenML version: 0.30.0rc0
Install path: /opt/anaconda3/lib/python3.9/site-packages/zenml
Python version: 3.9.7
Platform information: {'os': 'linux', 'linux_distro': 'ubuntu', 'linux_distro_like': 'debian', 'linux_distro_version': '22.10'}
Environment: native
Integrations: ['github', 'huggingface', 'pillow', 'plotly', 'pytorch', 'pytorch_lightning', 'scipy', 'sklearn']

The current user is: 'default'
The active project is: 'default' (global)
The active stack is: 'default' (global)

        Stack Configuration        
┏━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━┓
┃ COMPONENT_TYPE │ COMPONENT_NAME ┃
┠────────────────┼────────────────┨
┃ ARTIFACT_STORE │ default        ┃
┠────────────────┼────────────────┨
┃ ORCHESTRATOR   │ default        ┃
┗━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━┛

### What happened?

While passing a arrow dataset between two steps through HuggingFace dataset materializer, the dataset cannot be used in the second step, because its cache_file folder doesn't exist anymore.

The problem arises from the use of TemporaryDirectory in handle_input , which is destroyed at return, whereas this directory is still referenced in the dataset info. The folder must survive as long as the dataset is in use, so a simple work-around would be to use tempfile.mkdtemp instead of TemporaryDirectory.

But we should find how to clean-up properly after use.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dnth:
Hi @ghpu thank you for reporting this issue. Is there any code snippet to help me to replicate this issue?
It would be helpful if you can post the full error traceback here too.

Also, would you please post the output of 

`python -c "import zenml.environment; print(zenml.environment.get_system_details())" ` 

`zenml status`

and

`zenml stack describe` ?

morganveyret:
Here is  a minimal example producing the issue:
```python
from zenml.pipelines import pipeline
from zenml.steps import step
from zenml.steps import BaseParameters, Output

import datasets, transformers
from transformers import AutoTokenizer
from transformers import DataCollatorWithPadding
[...]

PR #1142, opened 2022-12-05
Fix Huggingface dataset materializer

safoinme:
## Describe changes
This PR is a fix to #1135 which changes the input handler `TemporaryDirectory` with `mkdtemp` to prevent automatic delete of the temp directory

## Pre-requisites

## Types of changes


src/zenml/integrations/huggingface/materializers/huggingface_datasets_materializer.py
```diff
@@ -43,12 +44,12 @@ def handle_input(self, data_type: Type[Any]) -> Dataset:
...
         super().handle_input(data_type)
-        temp_dir = TemporaryDirectory()
+        temp_dir = mkdtemp()
         io_utils.copy_dir(
             os.path.join(self.artifact.uri, DEFAULT_DATASET_DIR),
-            temp_dir.name,
+            temp_dir,
         )
-        return load_from_disk(temp_dir.name)
+        return load_from_disk(temp_dir)
```
schustmi, line 52: This works, but it is a little unclean as we don't cleanup this directory after it's not needed anymore (= the user step code has finished). Not necessary for this PR, but I'll create a ticket for this as it's something that we also do in other materializers

schustmi (approved):

src/zenml/integrations/huggingface/materializers/huggingface_datasets_materializer.py
safoinme: @schustmi how can we do that while making sure it still works when a step is cached? because the other function used before to create temp dir was able to clean up but this fails when the user uses a cached step.

schustmi: Not within this function, but we could have a `cleanup()` method that get's called on all materializers after a step is finished. https://zenml.atlassian.net/browse/ENG-1538

PR #1143, opened 2022-12-05
Move pandas and numpy as integrations

htahir1:
## Describe changes
I have attempted to refactor numpy and pandas as integrations rather than explicitly installing them/

## Pre-requisites

## Types of changes


src/zenml/integrations/numpy/__init__.py
```diff
@@ -0,0 +1,32 @@
+#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
schustmi, line 1:
```suggestion
#  Copyright (c) ZenML GmbH 2022. All Rights Reserved.
```

:P

src/zenml/integrations/numpy/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Initialization of the NumPy integration."""
+
+from zenml.integrations.constants import NUMPY
+from zenml.integrations.integration import Integration
+
+
+class NumpyIntegration(Integration):
+    """Definition of NUMPY integration for ZenML."""
```
schustmi, line 21:
```suggestion
    """Definition of NumPy integration for ZenML."""
```

schustmi (changes requested):
One other thing that is missing here is adding `numpy/pandas` to the `required_integrations` for all examples/blog posts that require them

fa9r (changes requested):
Implementation so far looks good.

One thing that is still missing is to adjust the requirements of all other integrations that depend on numpy or pandas being installed. E.g., almost all data validator, feature store, and visualizer integrations do `import pandas as pd`, and some of the model deployers do `import numpy as np`, which might or might not work depending on whether the respective tool itself has numpy/pandas as a requirement. To be save, I would just add numpy/pandas as a requirement in the respective integration init.

htahir1:
Closing because of internal discussions that we keep numpy and pandas as core dependencies for now

PR #1144, opened 2022-12-06
Check if all ZenML server dependencies are installed for local zenml deployment using `zenml up`

dnth:
## Describe changes

## Pre-requisites

## Types of changes


dnth:
> ```
> fastapi = { version = "~0.75.0", optional = true }
> uvicorn = { extras = ["standard"], version = "~0.17.5",  optional = true}
> python-multipart = { version = "~0.0.5", optional = true}
> python-jose = { extras = ["cryptography"], version = "~3.3.0", optional = true}
> fastapi-utils = { version = "~0.2.1", optional = true}
> ```
> 
> These are the dependencies we additionally install when someone runs `pip install zenml[server]` instead of the base zenml package. My question for this PR: do we need to actually verify that all of these are installed instead of just `fastapi_utils`? It might be that right now `fastapi_utils` gets imported first which is why this solution works, but the import order could easily swap in the future

I see.. thanks for the feedback @schustmi! I didn't see that earlier but now it's clear to me. I will also check for the other packages in this PR.

dnth:
On the other hand, I just found this function already does the dependency check for local deployment 

https://github.com/zenml-io/zenml/blob/561689c2b2f927badaa43dfef3adbda5da67a9bd/src/zenml/zen_server/deploy/local/local_provider.py#L55

Is this PR even necessary?

schustmi:
@dnth That is a very clean solution now! This `LocalProvider` is used when running `zenml up` I assume and therefore shows the correct error message?

dnth:
> @dnth That is a very clean solution now! This `LocalProvider` is used when running `zenml up` I assume and therefore shows the correct error message?

Yes that's right.

stefannica (approved):

bcdurak (approved):

Issue #1145, opened 2022-12-06
[BUG] pandas materializer: fails when serializing pd.Series


jordandelbar:
https://github.com/zenml-io/zenml/blob/561689c2b2f927badaa43dfef3adbda5da67a9bd/src/zenml/materializers/pandas_materializer.py#L134

This lead to an error when you have `pandas.Series` artifacts in 0.23.0 as there is no `to_parquet` or `to_csv` method for `pandas.Series`

Restoring that bit of code would solve the issue:
```
if isinstance(df, pd.Series):
   df = df.to_frame(name="series")
```
https://github.com/zenml-io/zenml/commit/1f5e8152bc8c4bcd70773d9b05f05c00099639a3#diff-44dcbf73d9e24d2d858d9e1a2f3cdf305de697459cbcb3781cc7e559e9702f6bL84

PR #1146, opened 2022-12-06
Fix pandas Series materializer

jordandelbar:
## Describe changes
Fix the #1145 issue

## Pre-requisites

## Types of changes


schustmi:
Thanks for the PR @jordandelbar! It would be great if you could rebase your changes on the latest develop state and add a quick test for this change in `tests/unit/materializers/test_pandas_materializer.py`. We have an utility function to quickly test materializers (check [this](https://github.com/zenml-io/zenml/blob/develop/tests/unit/materializers/test_built_in_materializer.py) out for some inspiration on how to use it) that should hopefully make this very easy

safoinme (approved):
LGTM! Thank you for fixing this 🙌

jordandelbar:
The bug only appears when `pyarrow` is also installed in the environment.
The unit test I added must be run in a `venv` with `pyarrow` and one without `pyarrow` (`pyarrow` that could come from another dependency). 
It could be done by adding a dimension in your matrix strategy of your `.github/workflows/unit-test.yml`

tests/unit/materializers/test_pandas_materializer.py
```diff
@@ -0,0 +1,33 @@
...
+#  permissions and limitations under the License.
+
+import pandas
+
+from tests.unit.test_general import _test_materializer
+from zenml.materializers.pandas_materializer import PandasMaterializer
+
+
+def test_pandas_materializer():
+
```
safoinme, line 22:
```suggestion
"""Test the pandas materializer."""
```
Adding the docstring to the test function, can you please also try running the `scripts/format.sh` so the code can be in the same format

safoinme (changes requested):

tests/unit/materializers/test_pandas_materializer.py
jordandelbar: My bad, it's done

safoinme (approved):

schustmi:
We're currently developing a new integration testing framework which can also be used to run specific unit tests in special environments, this might be a candidate for that to check that the materializer works with and without `pyarrow`.

PR #1147, opened 2022-12-06
Replace User Password with Token in docker images

safoinme:
## Describe changes
Replacing the username and password in the docker images to a token that is not as clear as a text user and password

## Pre-requisites

## Types of changes


schustmi (commented):
Minor comment, but overall looks good!

schustmi (commented):
Maybe as an additional question: Is there any expiration date on this token? It might be stored in scheduled runs running for days/weeks/months

safoinme:
@schustmi we don't have any expiration date for these api_tokens as of now, so I think this should work till we implement a token generator within the dashboard

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -1424,23 +1457,32 @@ def _get_auth_token(self) -> str:
...
+            if self.config.api_token:
+                self._api_token = self.config.api_token
+            # Otherwise, try to get the token from the server
+            else:
+                response = self._handle_response(
+                    requests.post(
+                        self.url + API + VERSION_1 + LOGIN,
+                        data={
+                            "username": self.config.username,
+                            "password": self.config.password,
```
stefannica, line 1470: same here: I would recommend throwing an exception if the username or password are None, before calling the request.

stefannica (changes requested):
Two minor things that you can do to make this more robust. Your changes look great otherwise.

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -149,11 +150,32 @@ class RestZenStoreConfiguration(StoreConfiguration):
...
+        Args:
+            values: A dictionary containing the values to be validated.
+
+        Raises:
+            ValueError: If neither api_token nor username is set.
+
+        Returns:
+            The values dictionary.
+        """
+        # Check if the values dictionary contains either an api_token or a username
```
schustmi, line 172: I had a few cases where this `values` dict actually contains lots of key-value pairs, but all the values are `None`.

If we want the token or username to be configured to an actual value, I'd recommend something like this:
```python
if values.get("api_token") is None and values.get("username") is None:
  ...

# the previous version allows empty strings, the requires non-empty strings
if values.get("api_token") and values.get("username"):
  ...
```

schustmi (changes requested):

src/zenml/zen_stores/rest_zen_store.py
safoinme: @schustmi  sorry, i'm a bit confused about this,
We should check if at least one of them is not an empty string. does that what the following code do?
```if values.get("api_token") and values.get("username"):```

schustmi: Yes exactly!

stefannica (approved):
Looks good, thank you !

schustmi (approved):

PR #1149, opened 2022-12-06
Context manager to track events

bcdurak:
## Describe changes
I have implemented a context manager to be used when tracking events. It enables us to track whether the selected event finished successfully or not (and if not, what type of error it fails with). 

I have changed most of the functions which used the direct `track_event` call to use the context manager instead. The `@track` decorator also uses this context manager by default now.

## Pre-requisites

## Types of changes


src/zenml/cli/base.py
```diff
@@ -219,54 +219,59 @@ def go() -> None:
...
-        with tempfile.TemporaryDirectory() as tmpdirname:
-            tmp_cloned_dir = os.path.join(tmpdirname, "zenml_repo")
-            with console.status(
-                "Cloning tutorial. This sometimes takes a minute..."
-            ):
-                Repo.clone_from(
-                    TUTORIAL_REPO,
-                    tmp_cloned_dir,
-                    branch=f"release/{zenml_version}",
+    with event_handler(AnalyticsEvent.RUN_ZENML_GO) as handler:
```
schustmi, line 222: I would suggest naming this `analytics_handler` or something similar, `handler` lacks any meaning in the code further down, especially for these long indents

src/zenml/pipelines/base_pipeline.py
```diff
@@ -432,107 +432,110 @@ def run(
...
-            # pipelines. This is done to prevent execution of module-level
-            # pipeline.run() calls inside docker containers which should only
-            # run a single step.
-            logger.info(
-                "Preventing execution of pipeline '%s'. If this is not "
-                "intended behavior, make sure to unset the environment "
-                "variable '%s'.",
-                self.name,
-                constants.ENV_ZENML_PREVENT_PIPELINE_EXECUTION,
+        with event_handler(AnalyticsEvent.RUN_PIPELINE) as handler:
```
schustmi, line 435: I think this handler should only start after the if the checks if the pipeline gets executed.

Otherwise all our docker images that run steps and don't have `if __name__ == '__main__':` around the `pipeline.run()` call will log pipeline run events

schustmi (changes requested):
Very cool, just a few comments :)

src/zenml/utils/analytics_utils.py
```diff
@@ -565,24 +545,77 @@ def inner_func(*args: Any, **kwargs: Any) -> Any:
...
+            return result
+
+    inner_func.__doc__ = func.__doc__
     return inner_func
+
+
+class event_handler(object):
+    """Context handler to enable tracking the success status of an event."""
+
+    def __init__(self, event: AnalyticsEvent):
```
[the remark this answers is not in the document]
bcdurak: I thought about this at first. But there are some operations, where the metadata of an event gets generated either during the operation or right at the end of it. In these cases, it would be impossible to feed it through the constructor.

We can still add it though for the example patterns that you have written.

src/zenml/pipelines/base_pipeline.py
```diff
@@ -362,16 +362,19 @@ def connect(self, *args: BaseStep, **kwargs: BaseStep) -> None:
         """
         raise NotImplementedError
 
-    def _track_pipeline_deployment(
+    def _track_pipeline_metadata(
```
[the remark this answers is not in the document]
bcdurak: Good catch! Renaming it accordingly.

src/zenml/utils/analytics_utils.py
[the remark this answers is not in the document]
bcdurak: Got it. I adjusted it accordingly.

src/zenml/pipelines/base_pipeline.py
bcdurak: Nice catch. Fixing it!

stefannica (approved):
Awesome use of a context manager !

src/zenml/cli/base.py
bcdurak: In the cases, where the metadata is already ready, I have put the metadata in the constructor and removed the `handler`. In all the other cases where the metadata is computed halfway through the operation, I have renamed the `handler` to `analytics_handler` following your suggestion.

PR #1151, opened 2022-12-07
Custom Metadata Tracking

fa9r:
## Describe changes
This PR introduces custom metadata tracking for pipeline runs, step runs, and artifacts.

https://www.loom.com/share/46d4d5d7511242d08670b0786e5289e7

(4min Loom video giving a short overview of the PR and how it works)

### Pipeline and Step Run Metadata Tracking

Docs page: https://app.gitbook.com/o/-MCl1Hlw9oU4xibZ-ymf/s/vlf6HVLGPRknfyMwMtR2/advanced-guide/stacks/component-metadata

Stack components now have two new methods `get_pipeline_run_metadata()` and `get_step_run_metadata()` which can be overwritten to track custom metadata about a pipeline run or a specific step of the run respectively.

This PR implements metadata tracking for the following components/flavours:
- Orchestrator:
    - Airflow:  orchestrator URI
    - Kubeflow: orchestrator URI
    - Vertex: orchestrator URI
    - SageMaker: pipeline execution ARN
- Experiment Trackers:
    - MLFlow: experiment tracker URL, MLFlow run ID, MLFlow experiment ID
     - WandB: experiment tracker URL, WandB run name
     - Neptune: experiment tracker URL
 - Model Deployer: -> deployed model URI

 #### Not Included (Future TODO)
- GitHubActions orchestrator URL
- SageMaker full orchestrator URL instead of ARN
- Step Operator URLs
- Container Registry Image name
- Secrets Manager: List of secrets used

### Artifact Metadata Tracking

Docs page: https://app.gitbook.com/o/-MCl1Hlw9oU4xibZ-ymf/s/vlf6HVLGPRknfyMwMtR2/advanced-guide/pipelines/materializers#optional-which-metadata-to-extract-for-the-artifact

Materializers now have a method `extract_metadata()` which can be overwritten to track custom metadata about the artifacts that a materializer saves.

By default, this will now always extract the `storage_size` of each artifact. 

Additionally, the following metadata will be extracted depending on the data type:
- Service Artifacts (e.g. deployed models):
    - `zenml.BaseService`: endpoint URI
- Data Artifacts:
    - `str` / `int` / `float` / `bool`: string representation
   - `list` / `dict` / `tuple` / `set` / `tf.Dataset`: length
    - `np.ndarray` / `pd.DataFrame` / `pd.Series` / `deepchecks.Dataset` / (HuggingFace) `datasets.Dataset`: shape, dtype, min, max, mean, std
    - `lightgbm.Dataset` / `spark.DataFrame` / `xgboost.DMatrix`: `shape`
    - `PIL.Image`: `width` / `height` / `mode`
    - `torch.DataLoader`: `num_samples` / `batch_size` / `num_batches`
    - `scipy.spmatrix`: `shape` / `dtype` / `nnz`
    - `bentoml.Bento`: tag name, info name, bentoml version
- Model Artifacts:
    - `torch.Module`: `num_params` / `num_trainable_params`
    - `tensorflow.keras.Model` / (HuggingFace) `transformers.TFPreTrainedModel`: `num_layers` / `num_params` / `num_trainable_params`
    - (HuggingFace) `transformers.PreTrainedModel`: `num_params` / `num_trainable_params` / `dtype` / `device`
- Data Analysis Artifacts:
    - `deepchecks.CheckResult`: check name, check passed
    - `deepchecks.SuiteResult`: suite name, suite passed
    - `great_expectations.CheckpointResult`: checkpoint result name, checkpoint result passed
    - `great_expectations.ExpectationSuite`: suite name

To disable artifact metadata tracking altogether, one can set `enable_artifact_metadata=False` at either pipeline or step level, similar to how `enable_cache` is used at the moment.

#### DB Migration
- All metadata is saved in a new table `run_metadata`
- `enable_artifact_metadata` is now saved as a separate DB column for both pipeline runs and step runs.

#### Other Changes
- All artifact stores / filesystems now have a `size()` method which is used to find the storage size of an artifact. Custom artifact stores that don't overwrite this will see a warning.
- Modified `post_execution.LineageGraph` and `DashVisualizer` to include metadata in the run visualizations
- Changed artifact type of `whylogs.DatasetProfileView` from `ArtifactType.STATISTICS` to `ArtifactType.DATA_ANALYSIS`
- Changed artifact type of HuggingFace `datasets.Dataset` from `ArtifactType.DATA_ANALYSIS` to `ArtifactType.DATA`

## Pre-requisites

## Types of changes


src/zenml/post_execution/pipeline_run.py
```diff
@@ -156,6 +156,18 @@ def enable_cache(self) -> bool:
         enable_cache = self.pipeline_configuration["enable_cache"]
         return cast(bool, enable_cache)
 
+    @property
+    def enable_artifact_metadata(self) -> bool:
```
AlexejPenner, line 160: a bit of a nitpick, but as a property I would expect it to not start with the verb, maybe `artifact_metadata_enabled` ?

src/zenml/post_execution/step.py
```diff
@@ -171,6 +171,15 @@ def enable_cache(self) -> bool:
         """
         return self.step_configuration.enable_cache
 
+    @property
+    def enable_artifact_metadata(self) -> bool:
```
AlexejPenner, line 175: see above

AlexejPenner (commented):
Two tiny nitpicks, looks great in general. I couldn't find the place where the string representation of the data is passed in as metadata (You mentioned this in the PR description) I was gonna say, we should onl do this if the string representation is sufficiently small (like <256 Charecters?)

fa9r:
@AlexejPenner the artifact metadata is handled similarly as, e.g., the step config right now: in the `ArtifactBaseModel` it is defined as `Dict[str, str]`, and in the `ArtifactSchema` we JSON-dump this into a column of type `TEXT`, so `65535` is the supported max length at the moment.

src/zenml/post_execution/pipeline_run.py
fa9r: Somewhat reluctant to change this since the underlying attribute in the model is also called `enable_artifact_metadata`

fa9r: And we also have `enable_cache` as a property which would be the same issue I guess

fa9r:
@schustmi @AlexejPenner @bcdurak would love some reviews on this before I implement the remaining stack components

src/zenml/metadata/metadata_types.py
```diff
@@ -0,0 +1,144 @@
...
+    bool,
+    dict,
+    list,
+    set,
+    tuple,  # type: ignore[type-arg]
+    Uri,
+    Path,
+    DType,
+    StorageSize,
+]
```
fa9r, line 69: I noticed this can't be used with `isinstance` checks anyways in older Python versions 😕

@schustmi @bcdurak in particular, how should we type the container types here then?
- `dict`?
- `Dict["MetadataType", "MetadataType"]`?
- `Dict[Any, Any]`?
- ???

AlexejPenner: With this many potential classes I feel moving to Any is probably best?

src/zenml/client.py
```diff
@@ -2386,6 +2392,113 @@ def _delete_artifact_metadata(
         self.zen_store.delete_artifact(artifact.id)
         logger.info(f"Deleted metadata of artifact '{artifact.uri}'.")
 
+    # ----------------
+    # - Run Metadata -
+    # ----------------
+
+    def create_run_metadata(
```
AlexejPenner, line 2399: This is on the client so that it can potentially be called manually within pipeline steps? How would that be done? Do we have a way to acces the current pipeline ID in the step at runtime? We'll definitely need docs around this.

src/zenml/client.py
```diff
@@ -2386,6 +2392,113 @@ def _delete_artifact_metadata(
...
+        pipeline_run_id: Optional[UUID] = None,
+        step_run_id: Optional[UUID] = None,
+        artifact_id: Optional[UUID] = None,
+        stack_component_id: Optional[UUID] = None,
+    ) -> List[RunMetadataResponseModel]:
+        """List run metadata.
+
+        Args:
+            project_id: If provided, only return metadata for this project.
+            user_id: If provided, only return metadata for this user.
```
AlexejPenner, line 2482: Why is this scoped by user or project. a pipeline_run already is scoped by this.

AlexejPenner (commented):
Some early thoughts

src/zenml/config/pipeline_configurations.py
```diff
@@ -42,6 +43,25 @@ class PipelineConfiguration(PipelineConfigurationUpdate):
 
     name: str
     enable_cache: bool
+    enable_artifact_metadata: bool
+
+    @root_validator(pre=True)
+    def _fill_missing_values(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
schustmi, line 49: Can we just give it a default value of `True` or `False` and remove this method entirely?

src/zenml/integrations/airflow/orchestrators/airflow_orchestrator.py
```diff
@@ -597,3 +603,22 @@ def _log_webserver_credentials(self) -> None:
...
+            run_id: The ID of the pipeline run.
+
+        Returns:
+            A dictionary of metadata.
+        """
+        run_metadata = {}
+        if self._dag_path:
+            run_metadata["airflow_dag_path"] = Path(self._dag_path)
+        if self.config.local:
+            run_metadata["airflow_log_file"] = Path(self.log_file)
```
schustmi, line 622: This is actually the server log file, I don't think that's particularly relevant for the run?

src/zenml/materializers/base_materializer.py
```diff
@@ -260,6 +262,38 @@ def save(self, data: Any) -> None:
...
+        data_type = type(data)
+        if not self._can_handle_type(data_type):
+            raise TypeError(
+                f"Unable to extract metadata from {data_type}. "
+                f"{self.__class__.__name__} can only write the following "
+                f"types: {self.ASSOCIATED_TYPES}."
+            )
+
+        metadata: Dict[str, "MetadataType"] = {
+            "runtime_data_type": DType(type(data))
```
schustmi, line 290:
```suggestion
            "runtime_data_type": DType(data_type)
```

Can we also just name this property `data_type`, why is the runtime prefix necessary?

src/zenml/metadata/metadata_types.py
schustmi: Any seems fine to me as well, plus some checks later on that make sure that those `Any` elements actually only contain the expected elements

src/zenml/orchestrators/step_launcher.py
```diff
@@ -215,13 +220,16 @@ def _create_or_reuse_run(self) -> PipelineRunResponseModel:
...
+            enable_artifact_metadata=self._deployment.pipeline.enable_artifact_metadata,
             status=ExecutionStatus.RUNNING,
             pipeline_configuration=self._deployment.pipeline.dict(),
             num_steps=len(self._deployment.steps),
             client_environment=self._deployment.client_environment,
             orchestrator_environment=get_run_environment_dict(),
         )
-        return client.zen_store.get_or_create_run(pipeline_run)
+        run_response = client.zen_store.get_or_create_run(pipeline_run)
+        run_was_created = run_response.name == run_name
```
schustmi, line 231: This logic is flawed I think, what if the run name is the same, e.g. a fixed run name specified by the user?

src/zenml/orchestrators/step_runner.py
```diff
@@ -354,12 +367,19 @@ def _store_output_artifacts(
...
+        output_artifact_metadata: Dict[str, Dict[str, "MetadataType"]] = {}
         for output_name, return_value in output_data.items():
             materializer_class = output_materializers[output_name]
             materializer_source = self.configuration.outputs[
                 output_name
             ].materializer_source
             uri = output_artifact_uris[output_name]
+            materializer = materializer_class(uri)
+            materializer.save(return_value)
+            # TODO: or self._pipeline.enable_artifact_metadata
```
schustmi, line 379: todo

src/zenml/pipelines/base_pipeline.py
```diff
@@ -164,6 +170,15 @@ def enable_cache(self) -> bool:
         """
         return self.configuration.enable_cache
 
+    @property
```
schustmi, line 173: I would remove this property tbh, it's already accessible via `pipeline.configuration.enable_artifact_metadata`. The only reason the other shortcuts exist is because of backwards compatability reasons

src/zenml/stack/stack.py
```diff
@@ -751,6 +751,42 @@ def prepare_step_run(self, info: "StepRunInfo") -> None:
         ).values():
             component.prepare_step_run(info=info)
 
+    def publish_pipeline_run_metadata(self, run_id: UUID) -> None:
```
schustmi, line 754: Any reason these methods can't be inside the step launcher/runner to keep them in the same place as all the other client calls related to pipeline/step runs? They seem a little out of place here in my opinion

schustmi (commented):
Didn't go into too much detail as the PR isn't complete yet, but generally speaking I really like the solution and think we can continue like this with the remaining components

src/zenml/materializers/built_in_materializer.py
```diff
@@ -66,7 +72,7 @@ def load(self, data_type: Type[Any]) -> Any:
             )
         return contents
 
-    def save(self, data: Any) -> None:
+    def save(self, data: Union[bool, float, int, str]) -> None:
```
AlexejPenner, line 75: Now that this is getting specified, why not also support dicts?

src/zenml/zen_stores/schemas/run_metadata_schemas.py
```diff
@@ -0,0 +1,155 @@
...
+            id=self.id,
+            pipeline_run_id=self.pipeline_run_id,
+            step_run_id=self.step_run_id,
+            artifact_id=self.artifact_id,
+            stack_component_id=self.stack_component_id,
+            key=self.key,
+            value=json.loads(self.value),
+            type=self.type,
+            project=self.project.to_model(),
+            user=self.user.to_model(),
```
AlexejPenner, line 128: Should take into account that users could be deleted while the metadata stays

```suggestion
            user=self.user.to_model() if self.user else None,
```

src/zenml/zen_stores/schemas/run_metadata_schemas.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Returns:
+            The created `RunMetadataResponseModel`.
+        """
+        return RunMetadataResponseModel(
+            id=self.id,
+            pipeline_run_id=self.pipeline_run_id,
+            step_run_id=self.step_run_id,
+            artifact_id=self.artifact_id,
+            stack_component_id=self.stack_component_id,
```
AlexejPenner, line 123:
```suggestion
            stack_component_id=self.stack_component_id if self.stack_component_id else None,
```

Stack components can be deleted

AlexejPenner (commented):
Two more small things I noticed, other than that, really like the abstraction and the overall strategy for metadata tracking!

src/zenml/materializers/built_in_materializer.py
```diff
@@ -75,6 +80,23 @@ def save(self, data: Any) -> None:
...
+
+        Args:
+            data: The built-in container object to extract metadata from.
+
+        Returns:
+            The extracted metadata as a dictionary.
+        """
+        base_metadata = super().extract_metadata(data)
+        builtin_metadata = {
+            "string_representation": str(data),
```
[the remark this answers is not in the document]
fa9r: Yeah the idea was if you have a basic data type it's probably most useful to just show the value. 

For `int`/`float`/`bool` this is always reasonably small. For `str` this could become very big though, I even encountered that already with the evidently materializer that returns the entire dashboard visualization as a 5M char `str`. That's why I included a manual check somewhere to filter such metadata out.

src/zenml/client.py
fa9r: Yeah it's in the client so users can log custom stuff if they really want/need to. I think you could get both the pipeline run ID and step run ID via `StepContext`. But I haven't experimented with that too much yet. Not even sure whether we should let users know this functionality exists.

src/zenml/client.py
fa9r: I imagine it could come handy if you want to see all metadata that a given user created, or all the metadata in a certain project. Now, how we resolve this in the backend is another question, I guess we could do it by joining the metadata schema with the respective entity that it is linked to (pipeline run / step run / artifact) and get the user/project from there, but I figured just adding it as foreign key directly makes that a lot easier

src/zenml/materializers/built_in_materializer.py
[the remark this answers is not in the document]
AlexejPenner: maybe we could be more restrictive, numbers and booleans seem safe

src/zenml/integrations/airflow/orchestrators/airflow_orchestrator.py
fa9r: Yeah makes sense. I didn't know which attributes make sense to log for Airflow so I just took everything that looked reasonable at first glance 😅

src/zenml/client.py
AlexejPenner: alright makes sense, i mean it will move into the FIlter class from my PR anyways

src/zenml/orchestrators/step_launcher.py
fa9r: Mhhh yeah true. Any suggestions how this could be done cleaner?

src/zenml/orchestrators/step_runner.py
fa9r: almost forgot, thanks for reminding 🙏

src/zenml/pipelines/base_pipeline.py
fa9r: Ah I see. Sure, let's get rid of it then

src/zenml/stack/stack.py
fa9r: I agree. Thinking about it again, I would say the cleanest design would be to have the stack collect the entire metadata as a dict and then have the client call to publish it somewhere in the launcher/runner. Will look into this again 👍

src/zenml/materializers/built_in_materializer.py
fa9r: dicts are handled by `BuiltInContainerMaterializer` below

src/zenml/zen_stores/schemas/run_metadata_schemas.py
fa9r: Good catch!

src/zenml/zen_stores/schemas/run_metadata_schemas.py
fa9r: Another good catch 👍

src/zenml/materializers/built_in_materializer.py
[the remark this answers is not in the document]
fa9r: Yeah I changed this to only include numbers and booleans

tests/unit/io/test_fileio.py
```diff
@@ -266,6 +266,23 @@ def test_stat_raises_error_when_file_doesnt_exist(tmp_path) -> None:
...
+def test_size_returns_int_for_file(tmp_path):
+    """Test that size returns an int for a file input"""
+    with NamedTemporaryFile(dir=tmp_path) as temp_file:
+        assert isinstance(fileio.size(temp_file.name), int)
+
+
+def test_size_returns_int_for_dir(tmp_path):
+    """Test that size returns an int for a directory input"""
+    with NamedTemporaryFile(dir=tmp_path):
+        assert isinstance(fileio.size(str(tmp_path)), int)
```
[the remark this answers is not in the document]
fa9r: Add a test for empty dirs and added asserts to the existing tests that compare the values to `os.path.getsize`.

src/zenml/materializers/base_materializer.py
fa9r: Still undecided about this

src/zenml/orchestrators/step_runner.py
fa9r: Will include in the upcoming `enable_artifact_metadata` rework

fa9r: Fixed

fa9r:
@schustmi @AlexejPenner do you think splitting the `run_metadata` table, model, and endpoint into `run_metadata` and `artifact_metadata` makes sense? Which would you find more intuitive?

AlexejPenner:
> @schustmi @AlexejPenner do you think splitting the `run_metadata` table, model, and endpoint into `run_metadata` and `artifact_metadata` makes sense? Which would you find more intuitive?

I don't really have an opinion on this, whatever is easier to maintain

src/zenml/integrations/bentoml/model_deployers/bentoml_model_deployer.py
```diff
@@ -455,3 +457,27 @@ def delete_model_server(
             self._clean_up_existing_service(
                 existing_service=service, timeout=timeout, force=force
             )
+
+    def get_step_run_metadata(
```
fa9r, line 461: Oh nice if this logic is the same for all deployer flavors we can even move it into the base deployer class 👍

fa9r: moved it here 👍

src/zenml/post_execution/pipeline_run.py
fa9r: ok, ignored :D

src/zenml/materializers/base_materializer.py
fa9r: Will remove it since it is duplicated from `data_type`

src/zenml/pipelines/base_pipeline.py
fa9r: discussion: just leave them for now since we haven't even deprecated them

src/zenml/orchestrators/step_launcher.py
fa9r: I added a dedicated `get_or_create_run` endpoint now that also returns a boolean of whether the run was created or not and adjusted all corresponding ZenStore methods accordingly.

src/zenml/steps/base_step.py
```diff
@@ -707,6 +721,15 @@ def enable_cache(self) -> Optional[bool]:
         """
         return self.configuration.enable_cache
 
+    @property
+    def enable_artifact_metadata(self) -> Optional[bool]:
```
schustmi, line 725: Can probably be removed same as the pipeline property?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -228,3 +230,23 @@ def prepare_or_run_pipeline(
...
+            run_id: The ID of the pipeline run.
+
+        Returns:
+            A dictionary of metadata.
+        """
+        # TODO: Add this once we can get the region
+        # run_url = (
+        #     f"https://{region}.console.aws.amazon.com/sagemaker/"
+        #     f"home?region={region}"
+        # )
```
AlexejPenner, line 249: Just a reminder in case this should make it into this PR

src/zenml/post_execution/pipeline_run.py
```diff
@@ -140,6 +140,21 @@ def enable_cache(self) -> Optional[bool]:
 
         return self.model.pipeline_configuration.get(PARAM_ENABLE_CACHE)
 
+    @property
+    def enable_artifact_metadata(self) -> Optional[bool]:
```
AlexejPenner, line 144: I think I already commented on this in teh past, i feel the phrasing just feels wrong, can't we rename all the new methods to show that they refer to a state `artifact_metadata_enabled` instead of making it sound like you can use this to set something to enabled

src/zenml/post_execution/step.py
```diff
@@ -129,6 +129,15 @@ def enable_cache(self) -> Optional[bool]:
         """
         return self.step_configuration.enable_cache
 
+    @property
+    def enable_artifact_metadata(self) -> Optional[bool]:
```
AlexejPenner, line 133: same as above

src/zenml/post_execution/pipeline_run.py
AlexejPenner: For example the param in the step environment is called `step_enabled`

schustmi (changes requested):
Have a few more suggestions, and also a request for some tests that would make sense IMO:
- Test the `Stack.get_pipeline_run_metadata` and `Stack.get_step_run_metadata` methods
- Test the `SQLZenStore.get_or_create_run` method
- Test the new `publish_utils` functions

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -1354,7 +1360,7 @@ def get_run(
 
     def get_or_create_run(
         self, pipeline_run: PipelineRunRequestModel
-    ) -> PipelineRunResponseModel:
+    ) -> Tuple[PipelineRunResponseModel, bool]:
```
AlexejPenner, line 1363: I feel this breaks with a lot assumptions we have about how the rest zen store methods work. Why is it important here to not return the response model if it already exists

src/zenml/zen_stores/schemas/run_metadata_schemas.py
AlexejPenner: Isn't it? 

    stack_component_id: Optional[UUID] = build_foreign_key_field(
        source=__tablename__,
        target=StackComponentSchema.__tablename__,
        source_column="stack_component_id",
        target_column="id",
        ondelete="SET NULL",
        nullable=True,
    )

AlexejPenner (approved):
:seal: i got some very nitty nitpicks - but in general, this lookgs good to me :seal:

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
fa9r: This TODO will probably stay open till we have `zenml pipeline deploy` implemented

src/zenml/post_execution/pipeline_run.py
fa9r: `caching_enabled` in the step environment is the processed version of this setting though, i.e., it takes both the pipeline setting and step setting into account, whereas these `enable_...` properties merely reflect the setting that the user specified on the respective object. @schustmi and I discussed the other day whether we should add `caching_enabled` into the step config as well but decided against it since we thought having it together with `enable_cache` might be too confusing.

fa9r: IMO it's also not worth spending a lot of effort on the post exec class property naming, I'd rather think about how we can further abstract those away or potentially even deprecate the post exec classes entirely

src/zenml/zen_stores/rest_zen_store.py
fa9r: We need to know whether it was created or not in `step_launcher.py`, because we only want to do the pipeline run metadata registration once per pipeline run.

fa9r: @schustmi and I discussed this for quite a while and, at the moment, having a separate `get_or_create` endpoint that gives us this bool is the best solution we could come up with. The alternatives would have been to 
- add `was_created_in_this_step` into the run model (which feels even more wrong) or
- have run metadata logging called multiple times per run (which would be unnecessary overhead and require race condition handling the back)

src/zenml/zen_stores/schemas/run_metadata_schemas.py
fa9r: sorry, wrong wording, it is a reference, but not a `Relationship`.

I.e., if `self.stack_component_id` is `None`, then `self.stack_component_id if self.stack_component_id else None` will return `None`, which is the same as `self.stack_component_id` would give you anyways.

src/zenml/client.py
```diff
@@ -1471,7 +1478,7 @@ def active_stack_model(self) -> "StackResponseModel":
         if not stack:
             raise RuntimeError(
                 "No active stack is configured. Run "
-                "`zenml stack set PROJECT_NAME` to set the active "
+                "`zenml stack set WORKSPACE_NAME` to set the active "
```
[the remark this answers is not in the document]
fa9r: good catch

src/zenml/client.py
```diff
@@ -3003,6 +3010,157 @@ def _delete_artifact_metadata(
...
+        pipeline_run_id: Optional[UUID] = None,
+        step_run_id: Optional[UUID] = None,
+        artifact_id: Optional[UUID] = None,
+        stack_component_id: Optional[UUID] = None,
+    ) -> List[RunMetadataResponseModel]:
+        """Create run metadata.
+
+        Args:
+            metadata: The metadata to create as a dictionary of key-value pairs.
+            pipeline_run_id: The ID of the pipeline run during which the
```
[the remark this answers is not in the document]
fa9r: Makes sense, added this into the arg descriptions

src/zenml/io/fileio.py
```diff
@@ -246,6 +235,44 @@ def stat(path: "PathType") -> Any:
...
+    if not exists(path):
+        return 0
+
+    # If the path is a file, return its size.
+    if not file_system.isdir(path):
+        return file_system.size(path)
+
+    # If the path is a directory, recursively sum the sizes of everything in it.
+    files = file_system.listdir(path)
+    file_sizes = [size(os.path.join(str(path), str(file))) for file in files]
```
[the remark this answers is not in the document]
fa9r: should be fine since `size` is called recursively

src/zenml/client.py
```diff
@@ -3003,6 +3010,157 @@ def _delete_artifact_metadata(
...
+                step_run_id=step_run_id,
+                artifact_id=artifact_id,
+                stack_component_id=stack_component_id,
+                key=key,
+                value=value,
+                type=metadata_type,
+            )
+            created_metadata = self.zen_store.create_run_metadata(run_metadata)
+            created_metadata_list.append(created_metadata)
+        return created_metadata_list
```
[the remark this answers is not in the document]
fa9r: I agree, changed it to dict 👍

src/zenml/zen_stores/rest_zen_store.py
AlexejPenner: Thanks, makes sense

tests/unit/stack/test_stack.py
```diff
@@ -456,3 +456,145 @@ def test_deployment_server_validation(mocker, stack_with_mock_components):
...
+    )
+    run_metadata = stack.get_pipeline_run_metadata(run_id=uuid4())
+    assert len(run_metadata) == 2
+    assert run_metadata[local_orchestrator.id] == orchestrator_metadata
+    assert run_metadata[local_artifact_store.id] == artifact_store_metadata
+    assert orchestrator_get_pipeline_run_mock.call_count == 1
+    assert artifact_store_get_pipeline_run_mock.call_count == 1
+
+
+def test_get_pipeline_never_raises_errors(
```
schustmi, line 494:
```suggestion
def test_get_pipeline_run_metadata_never_raises_errors(
```

tests/unit/stack/test_stack.py
```diff
@@ -456,3 +456,145 @@ def test_deployment_server_validation(mocker, stack_with_mock_components):
...
+    )
+    run_metadata = stack.get_step_run_metadata(info=MockStepInfo())
+    assert len(run_metadata) == 2
+    assert run_metadata[local_orchestrator.id] == orchestrator_metadata
+    assert run_metadata[local_artifact_store.id] == artifact_store_metadata
+    assert orchestrator_get_step_run_mock.call_count == 1
+    assert artifact_store_get_step_run_mock.call_count == 1
+
+
+def test_get_step_run_never_raises_errors(
```
schustmi, line 565:
```suggestion
def test_get_step_run_metadata_never_raises_errors(
```

schustmi (approved):

AlexejPenner (approved):
:seal:

PR #1154, opened 2022-12-07
Deprecate all the things

fa9r:
## Describe changes

I majorly refactored materializers:
- Renamed `materializer.handle_return()` -> `materializer.save()`
- Renamed `materializer.handle_input()` -> `materializer.load()`
- Deprecated `materializer.ASSOCIATED_ARTIFACT_TYPES` in favor of new `materializer.ASSOCIATED_ARTIFACT_TYPE` which is now of type `ArtifactType` instead of `Tuple[BaseArtifact]`
- Deprecated the materializer init arg `Materializer(artifact=...)` in favor of new `Materializer(uri=...)` which now only requires the artifact URI instead of the whole `BaseArtifact`
- Deprecated old way of skipping materialization by defining an output of type `BaseArtifact`. Instead, materialization is now skipped if the output type is of type `ArtifactResponseModel`
- Deprecated `@step(output_artifacts=...)` and removed `artifact_source` from all config classes
- Deleted the type registry
- Deleted all step interface classes

# TODO:
- update custom materializer docs

## Pre-requisites

## Types of changes


docs/book/advanced-guide/pipelines/materializers.md
```diff
@@ -386,35 +375,21 @@ returned and consumed from pipeline steps, you might sometimes need to have a
...
 
 ```python
-from zenml.artifacts import DataArtifact
+from zenml.models.artifact_models import ArtifactResponseModel
 from zenml.steps import step
 
 
 @step
-def my_step(my_artifact: DataArtifact):  # rather than pd.DataFrame
+def my_step(my_artifact: ArtifactResponseModel):  # rather than pd.DataFrame
```
schustmi, line 389: I don't like this too much, why would I as the user care (or even know) that this is somehow related to a response from something? I think this needs some rethinking and maybe some kind of public class that hides this information

src/zenml/config/step_configurations.py
```diff
@@ -26,14 +26,12 @@
 class PartialArtifactConfiguration(StrictBaseModel):
     """Class representing a partial input/output artifact configuration."""
 
-    artifact_source: Optional[str] = None
```
schustmi, line 29: This is not backwards compatible and will crash when loading an old config from the zen store in e.g. post execution

src/zenml/integrations/deepchecks/materializers/deepchecks_dataset_materializer.py
```diff
@@ -39,23 +39,23 @@ def handle_input(self, data_type: Type[Any]) -> Dataset:
         Returns:
             A Deepchecks Dataset.
         """
-        super().handle_input(data_type)
+        super().load(data_type)
 
         # Outsource to pandas
-        pandas_materializer = PandasMaterializer(self.artifact)
-        df = pandas_materializer.handle_input(data_type)
+        pandas_materializer = PandasMaterializer(self.uri)
```
schustmi, line 45: Wait what, why does this materializer not simply subclass the pandas one? :D

src/zenml/integrations/deepchecks/materializers/deepchecks_results_materializer.py
```diff
@@ -34,11 +34,9 @@ class DeepchecksResultMaterializer(BaseMaterializer):
         CheckResult,
         SuiteResult,
     )
-    ASSOCIATED_ARTIFACT_TYPES = (DataAnalysisArtifact,)
+    ASSOCIATED_ARTIFACT_TYPE = ArtifactType.DATAANALYSIS
```
schustmi, line 37: Maybe `DATA_ANALYSIS` is nicer to read?

src/zenml/steps/step_interfaces/__init__.py
```diff
@@ -1,42 +0,0 @@
-#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
schustmi, line 1: I'm personally very happy to see these gone, but maybe @bcdurak or @htahir1 who wanted these have some good reasons to keep them?

src/zenml/utils/materializer_utils.py
```diff
@@ -102,7 +95,7 @@ def load_model_from_metadata(model_uri: str) -> Any:
     return model
 
 
-def model_from_model_artifact(model_artifact: ModelArtifact) -> Any:
+def model_from_model_artifact(model_artifact: ArtifactResponseModel) -> Any:
```
schustmi, line 98: Should this check that the artifact type is actually `model` in this case? Otherwise it seems like it's just a `load_artifact` function

src/zenml/materializers/base_materializer.py
```diff
@@ -51,61 +52,116 @@ def __new__(
...
-                    f"type in its ASSOCIATED_TYPES class variable.",
+                    f"ASSOCIATED_ARTIFACT_TYPES value '{artifact_class}' for "
+                    f"materializer '{name}' is not a `BaseArtifact` "
+                    f"subclass.",
                     url="https://docs.zenml.io/advanced-guide/pipelines/materializers",
                 )
+            artifact_type = artifact_class.TYPE_NAME
+
+        # Validate associated artifact type.
+        is_invalid_artifact_type = True
```
schustmi, line 91: I think this variable isn't needed, the `MaterializerInterfaceError` can simply be raised in the `except ValueError` clause right?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -313,14 +310,15 @@ def _run_step(
                     step_run_info=step_run_info,
                 )
             else:
-                self._run_step_without_step_operator(
+                runner = StepRunner(step=self._step, stack=self._stack)
```
schustmi, line 313: I would still keep this in a separate method, it's a nice separation and prevents messy code in the future by adding additional code in this else condition

src/zenml/orchestrators/step_runner.py
```diff
@@ -177,12 +187,19 @@ def run(
                     output_name
                 ].materializer_source
 
-                self._store_output_artifact(
-                    materializer_class=materializer_class,
-                    materializer_source=materializer_source,
-                    artifact=output_artifacts[output_name],
-                    data=return_value,
+                uri = output_artifact_uris[output_name]
```
schustmi, line 190: Again, I would suggest we try to create additional small methods instead of adding all code into (already far too huge) methods.

schustmi (changes requested):
Few minor comments, looks good otherwise!

bcdurak (commented):
Not much to add on top of @schustmi 's comments. Everything looks good! Really happy to see a way cleaner implementation!

docs/book/advanced-guide/pipelines/materializers.md
bcdurak: It also looked a bit out of place for me as well.

src/zenml/integrations/deepchecks/materializers/deepchecks_results_materializer.py
bcdurak: +1

src/zenml/steps/step_interfaces/__init__.py
bcdurak: Same here. I do not see any reason why we should keep them.

htahir1: lets get rid of em

docs/book/advanced-guide/pipelines/materializers.md
fa9r: Good point, will change this to use a new class `UnmaterializedArtifact` that subclasses `ArtifactResponseModel`.

src/zenml/integrations/deepchecks/materializers/deepchecks_dataset_materializer.py
fa9r: good point, let me rewrite that

fa9r: Ah, the issue is that `save(x)` checks `if not self._can_handle_type(type(x)): ...`. If we subclass the pandas materializer and call `super().save(df.data)`, this would then assert that `df.data` is of type `Dataset` which it is not.

src/zenml/orchestrators/step_runner.py
fa9r: Not as easy here as for the other one. I tried to quickly refactor some code into a util method similar as we had before but ended up refactoring the entire `StepRunner` 😄

src/zenml/integrations/deepchecks/materializers/deepchecks_dataset_materializer.py
fa9r: This means materializer inheritance does not work as expected at the moment. Any suggestions how to handle this? Is it even worth fixing?

Personally, I feel like the proper solution would be to have `save()` and `load()` call two abstract functions `_save()` and `_load()` that perform just the data saving/loading without additional validations but that would be a major breaking change.

fa9r: I made the main changes on a separate branch, what do you guys think @htahir1 @schustmi @bcdurak?

https://github.com/zenml-io/zenml/pull/1175

fa9r: Another thing, I just realized the current PR would break custom materializers anyways since overwriting `handle_return()` and `handle_input()` will not do anything once our code calls `load()` and `save()`.

src/zenml/config/step_configurations.py
fa9r: I guess we could explicitly pop `artifact_source` before we create the `Step` in `StepRunSchema.to_model()`. What do you think about that?

fa9r: Added a pydantic root validator to remove the `artifact_source`

src/zenml/integrations/deepchecks/materializers/deepchecks_dataset_materializer.py
fa9r: For now we decided to keep the inheritance as it is. In order to support deprecated custom materializers, I added some logic that routes `save()` and `load()` to those legacy methods if they were overridden.

strickvl (approved):
I mainly looked at the tests + how they were / are affected by this. That all looks good to me. For the rest, a quick review doesn't raise any big questions for me. Thank you for doing this; I think this makes it much nicer for the user to deal with materializers.

src/zenml/orchestrators/step_runner.py
```diff
@@ -235,49 +231,138 @@ def _load_input_artifact(
...
         Returns:
             The artifact value.
-
-        Raises:
-            RuntimeError: If the artifact has no materializer.
         """
-        # Skip materialization for BaseArtifact and its subtypes.
+        # Skip materialization for `UnmaterializedArtifact`.
+        if data_type == UnmaterializedArtifact:
+            return UnmaterializedArtifact.parse_obj(artifact.dict())
```
schustmi, line 237:
```suggestion
            return UnmaterializedArtifact.parse_obj(artifact)
```

This is a pydantic shorthand for the same thing, they call `model.dict()` internally

src/zenml/config/step_configurations.py
```diff
@@ -26,14 +28,30 @@
...
+        Args:
+            values: The values dict used to instantiate the model.
+
+        Returns:
+            The values dict without deprecated attributes.
+        """
+        deprecated_attributes = ["artifact_source"]
+        for deprecated_attribute in deprecated_attributes:
+            if deprecated_attribute in values:
+                values.pop(deprecated_attribute)
```
schustmi, line 48: I prefer this one-liner here, up to you what you think is clearer

```suggestion
            values.pop(deprecated_attribute, None)
```

schustmi (approved):
Great job! Two tiny suggestions, up to you if you want to apply them or leave as is

src/zenml/config/step_configurations.py
fa9r: I'll stay with the explicit version for this one

PR #1158, opened 2022-12-08
Scheduling documentation

strickvl:
I added a page describing how to use ZenML's schedule object in pipeline runs.

## Pre-requisites

## Types of changes


docs/book/toc.md
```diff
@@ -44,6 +44,7 @@
 * [Practical MLOps](./advanced-guide/practical/practical-mlops.md)
   * [Stack Recipes](./advanced-guide/practical/stack-recipes.md)
   * [Switching Orchestration](./advanced-guide/practical/switching-orchestration.md)
+  * [Scheduling Pipeline Runs](./advanced-guide/practical/scheduling-runs.md)
```
htahir1, line 47: I would put this rather here: https://docs.zenml.io/advanced-guide/pipelines (In depth configuration)

docs/book/advanced-guide/practical/scheduling-runs.md
```diff
@@ -0,0 +1,54 @@
...
+from zenml.config.schedule import Schedule
+
+schedule = Schedule(start_time=datetime.now(), interval_second=1800)
+pipeline.run(schedule=schedule)
+```
+
+You can also optionally specify an `end_time` for the schedule if you want it not
+to run after a certain moment. Similarly, the boolean `catchup` parameter can be
+used to specify whether a recurring run should catch up (i.e. backfill pipeline
+runs) if it is behind schedule, if you paused the schedule, for example.
```
htahir1, line 37: Maybe link to what backfill is? Also make a note that backfill isnt supported for all orchestrators

docs/book/advanced-guide/practical/scheduling-runs.md
```diff
@@ -0,0 +1,54 @@
...
+
+ZenML pipelines can also be used for scheduling orchestration jobs, but there are some limitations to consider. ZenML-powered orchestrators only support scheduling in certain orchestrators and setup (see below for details on what works and what doesn't), as they require something running in the background to trigger the pipeline runs. Despite these limitations, using scheduling with ZenML can be a useful way to automate your MLOps workflow and save time and effort.
+
+# How to schedule a pipeline run
+
+ZenML's scheduling functionality rests on the use of a `Schedule` object that
+you pass in when calling `pipeline.run()`. There are two ways to create a
+schedule with the `Schedule` object:
+
+You could write a cron expression to describe the pipeline schedule in terms
```
htahir1, line 15: We should write here that not all parameters are supported for all orchestrators. for example cron_Expression is only supported by Vertex, which therefore ignores start_time etc

docs/book/advanced-guide/practical/scheduling-runs.md
```diff
@@ -0,0 +1,54 @@
...
+pipeline.run(schedule=schedule)
+```
+
+Alternatively, you could manually specify start and end times and the interval between runs. For example, if you wanted your pipeline to run once every 30 minutes starting right now, you could use the following:
+
+```python
+from zenml.config.schedule import Schedule
+
+schedule = Schedule(start_time=datetime.now(), interval_second=1800)
+pipeline.run(schedule=schedule)
```
htahir1, line 31: make clear this is pipeline_instance by showing more code

docs/book/advanced-guide/practical/scheduling-runs.md
```diff
@@ -0,0 +1,54 @@
...
+| Orchestrator | Scheduling Support |
+| ------------ | ------------------ |
+| LocalOrchestrator | ⛔️ |
+| LocalDockerOrchestrator | ⛔️ |
+| KubernetesOrchestrator | ✅ |
+| KubeflowOrchestrator | ✅ |
+| VertexOrchestrator | ✅ |
+| TektonOrchestrator | ⛔️ |
+| AirflowOrchestrator | ✅ |
+| GiHubActionsOrchestrator | ✅ |
```
htahir1, line 54: After this, maybe add that the first step of a pipeline should usually not contain static data. E.g. Usually you pull from an API (You can use the zenml gitflow steps as examples (https://github.com/zenml-io/zenml-gitflow/blob/main/steps/data_loaders.py#L69)

Also cache should be disabled in these pipelines/steps usually

Perhaps a full example would help here. Copy paste a minimal example with a step loading data from API and cache being disabled

docs/book/advanced-guide/practical/scheduling-runs.md
```diff
@@ -0,0 +1,54 @@
...
+# Supported Orchestrators
+
+| Orchestrator | Scheduling Support |
+| ------------ | ------------------ |
+| LocalOrchestrator | ⛔️ |
+| LocalDockerOrchestrator | ⛔️ |
+| KubernetesOrchestrator | ✅ |
+| KubeflowOrchestrator | ✅ |
+| VertexOrchestrator | ✅ |
+| TektonOrchestrator | ⛔️ |
```
htahir1, line 52: Maybe add a link to the roadmap here that if somebody wants to implement it they are welcome

htahir1 (changes requested):
Great! I just left some comments as I think its an important page and common things ive seen people struggle with should be addressed

fa9r (changes requested):
The page looks good and describes well how schedules can be used in ZenML right now.

However, I am wondering whether we should write about this at all at this point since schedules have several major design flaws:
- Creating a schedule via `pipeline.run()` feels unintuitive. Personally, I would expect that this would run the pipeline once and then create a schedule afterwards.
- After creating a schedule, there is no way in ZenML to ever interact with it again or to even find out that it exists. This feels very bad from a UX perspective and might even incur unexpected costs.
- The schedule parameters have unexpected behavior. As a user, I would expect that I can set either start time + interval time or CRON expression and that it will work on all orchestrators that support scheduling. This should indeed easily be possible, we just haven't implemented it.

Personally, I would rather not mention schedules at all at this point until we figured out a better UX around them.

docs/book/advanced-guide/practical/scheduling-runs.md
schustmi: It deson't necessarily have to be the first step, just one of the steps in the pipeline should load dynamic data, could also be a later step in a batch inference pipeline for example.

schustmi (approved):
Apart from the mentioned comments looks good to me!

schustmi:
@fa9r These points aren't as simple:

- If you use two different methods for running/scheduling, you lose the option to change the schedule configuration with a configuration file. I might want to run the same `run.py` file locally without a schedule to test, and with a schedule in my CI system after it merged to main. This functionality is lost if you need to change code to go from one-off runs to scheduled runs. We thought about this a few times, and maybe the option is to have `pipeline.run()` for one-off runs, `pipeline.schedule()` for scheduled runs and `pipeline.deploy()` for either of the two. In any case, something we should put serious thought in as it's the main user-facing interaction.
- This will **never** be possible in the case in for example Airflow. I don't think this should skip us from having and documenting this feature.
- I disagree. Converting from cron to intervals is not something that is generally possible I think, and we therefore can't allow all options for all orchestrators. (Even the reverse `start_time` and `interval` -> `cron` conversion isn't trivial or even impossible in some cases). As long as there's an error I don't see any reason why only supporting one is a huge issue

htahir1:
I agree with @schustmi tbh @fa9r . We should get this out there and not block ourselves

strickvl:
@htahir1 updated now with all your suggestions handled.

htahir1:
@fa9r Check out this question: https://zenml.slack.com/archives/C01FWQ5D0TT/p1670530355778149 :-P

docs/book/advanced-guide/pipelines/scheduling-runs.md
```diff
@@ -0,0 +1,114 @@
...
+
+schedule = Schedule(start_time=datetime.now(), interval_second=1800)
+pipeline_instance.run(schedule=schedule)
+```
+
+You can also optionally specify an `end_time` for the schedule if you want it not
+to run after a certain moment. Similarly, the boolean `catchup` parameter can be
+used to specify whether a recurring run should catch up (i.e. backfill pipeline
+runs) if it is behind schedule, if you paused the schedule, for example.
+
```
htahir1, line 46: Rephrasing to make a bit clearer IMO:

```suggestion
You can specify an optional `end_time` for the schedule to prevent it from running after a certain time. The `catchup` parameter, which is a boolean, can be used to specify whether a recurring run should catch up (i.e. backfill pipeline
runs) on missed runs if it has fallen behind schedule. This can happen, for example, if you paused the schedule.
```

docs/book/advanced-guide/pipelines/scheduling-runs.md
```diff
@@ -0,0 +1,114 @@
...
+misses a scheduled run at 12:00 PM, backfilling can be used to run the pipeline
+for the 12:00 PM time slot, in order to collect the missing data. This helps
+ensure that the pipeline is up-to-date and that downstream jobs have the
+necessary data to run correctly. Backfilling is a useful technique for catching
+up on missed runs and filling in gaps in scheduled jobs, and can help ensure
+that pipelines and cron schedules are running smoothly. Usually, if your
+pipeline handles backfill internally, you should turn catchup off to avoid
+duplicate backfill. (Here's [a handy
+guide](https://medium.com/nerd-for-tech/airflow-catchup-backfill-demystified-355def1b6f92)
+in the context of Airflow.) Note that the `catchup` parameter enabling
```
htahir1, line 58: I think the note should be in a `{% hint style="warning" %}{% endhint %}` in the next line

docs/book/advanced-guide/pipelines/scheduling-runs.md
```diff
@@ -0,0 +1,114 @@
...
+backfilling is not supported in all orchestrators.
+
+# How to stop or pause a scheduled run
+
+The way pipelines are scheduled depends on the orchestrator you are using. For example, if you are using Kubeflow, you can use the Kubeflow UI to stop or pause a scheduled run. If you are using Airflow, you can use the Airflow UI to do the same. However, the exact steps for stopping or pausing a scheduled run may vary depending on the orchestrator you are using. We recommend consulting the documentation for your orchestrator to learn the current method for stopping or pausing a scheduled run.
```
htahir1, line 63: Ok you should add here that currently ZenML only gets involved to initially schedule a job, and to maintain the lifecycle of the schedule is still on the user (with the methods you described).

Maybe explain what happens if you run a scheduled pipeline twice. @schustmi actually what happens in thsi case? Does it just create two pipelines with different names?

docs/book/advanced-guide/pipelines/scheduling-runs.md
```diff
@@ -0,0 +1,114 @@
+---
+description: Schedule runs to add automation to your pipelines
+---
+
+In the world of MLOps, scheduling orchestration jobs is an important aspect of automating the processes involved in deploying and maintaining machine learning (ML) models in production environments. Scheduling orchestration jobs allows you to automate tasks such as training and evaluating ML models, deploying models to production, or running periodic checks to ensure that the models are functioning as expected. This can help ensure that your ML pipelines are executed at the right time and in the right order, and can save you time and effort by eliminating the need to manually trigger these tasks.
+
+ZenML pipelines can also be used for scheduling orchestration jobs, but there are some limitations to consider. ZenML-powered orchestrators only support scheduling in certain orchestrators and setup (see below for details on what works and what doesn't), as they require something running in the background to trigger the pipeline runs. Despite these limitations, using scheduling with ZenML can be a useful way to automate your MLOps workflow and save time and effort.
+
+# How to schedule a pipeline run
```
htahir1, line 9: Do we have a single `#` in all of our docs like this? I somehow remember we do ## ?

htahir1 (changes requested):
Almost there! A few more comments

htahir1 (approved):

PR #1161, opened 2022-12-08
Moved the active_user property onto the client, implemented get_myself as zenstore method

AlexejPenner:
## Describe changes
I implemented a zenstore method: get_myself() that returns the currently active user including private information (email) 

On the RestZenstore this method hits the /current-user endpoint which gets its user model from the authentication context. The authentication context in turn uses the get_myself(user_id) method of the sqlzenstore

On the SQLZenStore, this method defaults to return the default user if no user is specified. 

## Pre-requisites

## Types of changes


src/zenml/zen_server/auth.py
```diff
@@ -94,7 +94,7 @@ def authenticate_credentials(
     if user_name_or_id:
         try:
             user = zen_store().get_auth_user(user_name_or_id)
-            user_model = zen_store().get_user(user_name_or_id=user_name_or_id)
+            user_model = zen_store().get_myself(user_name_or_id=user_name_or_id)
```
safoinme, line 97: why would we still need to specify `user_name_or_id` if it's get_myself shouldn't the function be getting the current active user by default?

src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -439,6 +431,19 @@ def get_user(self, user_name_or_id: Union[str, UUID]) -> UserResponseModel:
             KeyError: If no user with the given name or ID exists.
         """
 
+    @abstractmethod
+    def get_myself(
+        self, user_name_or_id: Optional[Union[str, UUID]] = None
```
schustmi, line 436: Without reviewing this in detail, I have two questions regarding this method:
1. I think safoine for the token inside docker image recently implemented a `current_user` call, is this now duplicated?
2. Why can I pass arguments like the name or ID to this method, doesn't that defeat the purpose of returning "me"?

src/zenml/zen_stores/schemas/user_schemas.py
```diff
@@ -105,11 +105,16 @@ def update(self, user_update: UserUpdateModel) -> "UserSchema":
...
+    def to_model(
+        self, _block_recursion: bool = False, include_private: bool = False
+    ) -> UserResponseModel:
         """Convert a `UserSchema` to a `UserResponseModel`.
 
         Args:
             _block_recursion: Don't recursively fill attributes
+            include_private: Whether to include the user private information
+                             this is to limit the amount of data one can get
+                             about other users
```
fa9r, line 117: Not formatted correctly

src/zenml/zen_server/auth.py
fa9r: Ah so the difference between `get_myself() and `get_user()` is not that `get_myself()` returns a fixed user but rather that the returned model has more fields than the default user response.

fa9r (changes requested):
Love that you cleaned up those unnecessary ZenStore properties 🙌 

Regarding `get_myself()`, I agree with the others that it is a bit unintuitive that this has a `user_name_or_id` arg. I would expect that this hits the `current-user` endpoint when going through REST and simply returns the default user for the SQLZenStore, so I think this arg shouldn't be needed either. Except for that arg, I think the PR looks great 👍

src/zenml/zen_stores/zen_store_interface.py
AlexejPenner: 2. Answer see above, how would the SQLZenStore know what the active user is. We have that information on the endpoint in the auth_context, but the SQLZenStore does not and should not have this information

AlexejPenner: I'm definitely open to suggestions though @fa9r @schustmi @safoinme

safoinme:
I think the implementation actually makes sense, but I only find a problem with the name of the function. as the first thing that I understand from the name `get_myself` is to get the current user. 
Is there a possibility to include all in the `get_user()` function by changing it to something like this:

The abstract  `get_user` method gets optional `user_name_or_id` and a new by default False boolean `include_private`.
then the _rest_zen_store_ implementation would be:
```suggestion:
    def get_user(
        self,
        user_name_or_id: Optional[Union[str, UUID]] = None,
        include_private: bool = False
    ) -> UserResponseModel:
        """Gets a specific user.

        Args:
            user_name_or_id: The name or ID of the user to get.

        Returns:
            The requested user, if it was found.
        """
        if not user_name_or_id:
            body = self.get(MYSELF)
            return UserResponseModel.parse_obj(body)
        else:
            return self._get_resource(
                resource_id=user_name_or_id,
                route=USERS,
                response_model=UserResponseModel,
            )
```
on the other side the `_sql_zen_store_` would be:

```suggestion
    def get_user(
        self,
        user_name_or_id: Optional[Union[str, UUID]] = None,
        include_private: bool = False
    ) -> UserResponseModel:
    ...
    ...
        return user.to_model(include_private=include_private)
```
Changing every time `get_myself` to `get_user(include_private=True)`.  In every function, we always check if `user_name_or_id` exists before calling the `get_user` which means it will be empty only when we want actually the current user or default one, and for same for `include_private` since it will include private on the same places where `get_myself` called them and won't be applied in r`est_zen_store` since not used within the definition.

src/zenml/zen_stores/zen_store_interface.py
schustmi: Ah I see, I think my confusion came mostly from the naming of the method then. If this just gives me any arbitrary logged in used, can we rename it to reflect that? `get_logged_in_user`, `get_full_user`, ...?

schustmi: Actually `get_logged_in_user` isn't any better, I think it needs to somehow describe the difference between users and "full" users

AlexejPenner: I have now applied a suggestion I got from safoine, I removed the get_myself() method and moved everything onto get user

if you don't explicitely set the user_name_or_id you get the active user

if you get the active user, you can then opt to get the private information as well (this happens implicitely)

schustmi: Oh I like that, looks great now!

safoinme (approved):
LGTM!

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -667,29 +667,36 @@ def create_user(self, user: UserRequestModel) -> UserResponseModel:
...
-
-        Returns:
-            The requested user, if it was found.
-        """
-        body = self.get(f"{CURRENT_USER}")
-        return UserResponseModel.parse_obj(body)
+    def get_user(
+        self,
+        user_name_or_id: Optional[Union[str, UUID]] = None,
+        include_private: bool = False,
```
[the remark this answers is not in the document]
AlexejPenner: Yes, it would be ignored anyway, see the docstring

PR #1162, opened 2022-12-08
Adding component conditions to experiment tracker examples and adding to the environmental variable docs

bcdurak:
## Describe changes
- I have added a condition to some of our examples to include a specific type of experiment tracker. This way we can avoid the `NoneType has no attribute 'name'` error. 
- All the examples with step operators which mirror this problem were already taken care of.
- I have also added one-liner to the list of environmental variables on our docs.

## Pre-requisites

## Types of changes


fa9r (approved):
Good catch 👍

schustmi (approved):

PR #1171, opened 2022-12-12
Fixing the error messages when fetching entities

bcdurak:
## Describe changes
I have improved the error messages when `get`ting entities by name, ID, or prefix.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/project.py
+++ b/src/zenml/cli/project.py
@@ -127,10 +127,14 @@ def describe_project(project_name_or_id: Optional[str] = None) -> None:
             [active_project], exclude_columns=["created", "updated"]
         )
     else:
-        project = client.get_project(project_name_or_id)
-        cli_utils.print_pydantic_models(
-            [project], exclude_columns=["created", "updated"]
-        )
+        try:
+            project_ = client.get_project(project_name_or_id)
+        except KeyError as err:
+            cli_utils.error(str(err))
+        else:
+            cli_utils.print_pydantic_models(
+                [project_], exclude_columns=["created", "updated"]
+            )
 
 
 @project.command("set", help="Set the active project.", hidden=True)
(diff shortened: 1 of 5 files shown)


strickvl (changes requested):
The changes so far are looking good, though to bring it in line with other parts of the CLI could you make sure that the raised errors are handled in the CLI? I.e. like what happens with `zenml stack delete ...` for a non-existent stack. Specifically, I don't believe the user should be exposed to a full stacktrace when updating a non-existent component.

Also, could you add a test for this scenario in `test_client.py` (it is covered already in the corresponding CLI test). We don't have a test for it yet.

bcdurak:
You are absolutely right. The only reason why we have not invested more time into the tests is the upcoming `list, prefix matching, and pagination` branch. With those changes, these two functions will completely change. Until then I would propose keeping it as it is in terms of tests.

As for the stack trace, I think I have a solution. But, it is a generalized solution. It would help us remove all the 
```
try:
    something()
except ThisError as e:
    cli_utils.error(str(e))
except ThatError as e:
    ... 
    cli_utils.error(e)
```
blocks and clean up the CLI code. However, it might make sense then to do it in a separate PR as it would be significantly larger in size.

strickvl:
I leave it to your better judgement about the tests, but for now I'd just make the non-generic fix for the CLI probably. I did like the sample / suggestion you sent to me over discord, but I would not invest all the effort to create the generic fix that's applicable everywhere, since that seems like it'd take a bit of time / effort to implement..

bcdurak:
@strickvl I tried to fix as many occurrences as I can find.

strickvl (approved):
Thanks. Looks good to go now!

AlexejPenner (approved):
:seal:

PR #1173, opened 2022-12-12
Persist the server ID in the database

stefannica:
## Describe changes
The global configuration `user_id` attribute is now persisted as a row in a new `identity` table in the database, either when the database is initialized or when it is upgraded from a previous version. This identity field is then used to identify any and all servers that connect to the same database.

This change allows us to have a more consistent view into the lifecycle of various ZenML server instances, independent of whether they are re-installed from scratch or not.

UPDATE: additional testing also revealed that the previous way of setting the `user_id` global configuration attribute through the use of `ZENML_USER_ID` environment variable and the `serverId` helm chart configuration field **did not work** because this field is immutable in the global configuration pydatic class. This in turn caused the analytics to report a new server ID for every different zenml server instance, even when connected to the same database. This is also corrected by this PR, the `ZENML_USER_ID` environment variable mentions are removed from the docs and the `serverId` variable is removed from the helm chart.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/server.py
+++ b/src/zenml/cli/server.py
@@ -484,9 +484,6 @@ def deploy(
         metadata = {
             "server_deployment": str(server.config.provider),
         }
-        from zenml.zen_server.deploy.terraform.terraform_zen_server import (
-            TerraformServerDeploymentConfig,
-        )
 
         analytics_handler.metadata = metadata
 
@@ -527,10 +524,6 @@ def destroy() -> None:
         metadata = {
             "server_deployment": str(server.config.provider),
         }
-
-        from zenml.zen_server.deploy.terraform.terraform_zen_server import (
-            TerraformServerDeploymentConfig,
-        )
 
         analytics_handler.metadata = metadata
 
--- a/src/zenml/cli/server.py
+++ b/src/zenml/cli/server.py
@@ -488,11 +488,6 @@ def deploy(
             TerraformServerDeploymentConfig,
         )
 
-        if isinstance(server.config, TerraformServerDeploymentConfig):
-            # TODO: maybe move the server ID into the ServerDeploymentConfig
-            #   class
-            metadata["server_id"] = str(server.config.server_id)
-
         analytics_handler.metadata = metadata
 
         if server.status and server.status.url:
@@ -536,9 +531,6 @@ def destroy() -> None:
         from zenml.zen_server.deploy.terraform.terraform_zen_server import (
             TerraformServerDeploymentConfig,
         )
-
-        if isinstance(server.config, TerraformServerDeploymentConfig):
-            metadata["server_id"] = str(server.config.server_id)
 
         analytics_handler.metadata = metadata
 
--- a/src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
+++ b/src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
@@ -1,7 +1,7 @@
 """Add identity table [7e4a481d17f7].
 
 Revision ID: 7e4a481d17f7
-Revises: 7834208cc3f6
+Revises: 0.30.0
 Create Date: 2022-12-08 16:27:48.909015
 
 """
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "7e4a481d17f7"
-down_revision = "7834208cc3f6"
+down_revision = "0.30.0"
 branch_labels = None
 depends_on = None
 
--- a/src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
+++ b/src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
@@ -1,0 +1,52 @@
+"""Add identity table [7e4a481d17f7].
+
+Revision ID: 7e4a481d17f7
+Revises: 7834208cc3f6
+Create Date: 2022-12-08 16:27:48.909015
+
+"""
+import sqlmodel
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "7e4a481d17f7"
+down_revision = "7834208cc3f6"
+branch_labels = None
+depends_on = None
+
+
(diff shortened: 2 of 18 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -726,10 +727,25 @@ def get_store_info(self) -> ServerModel:
...
         model = super().get_store_info()
         sql_url = make_url(self.config.url)
         model.database_type = ServerDatabaseType(sql_url.drivername)
+
+        # Fetch the deployment ID from the database and use it to replace the one
+        # fetched from the global configuration
+        with Session(self.engine) as session:
+            identity = session.exec(select(IdentitySchema)).first()
+
+            if identity is None:
```
htahir1, line 744: Are you sure this should be an exception? Wouldnt it suck to have this happen? What would be a way to recover

src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
```diff
@@ -0,0 +1,51 @@
...
+    # Prefill the identity table with a single row that contains the deployment
+    # id extracted from the global configuration
+    op.bulk_insert(
+        sa.Table(
+            "identity",
+            meta,
+        ),
+        [
+            {
+                "id": str(GlobalConfiguration().user_id).replace("-", ""),
```
htahir1, line 43: maybe i misunderstood this but shouldnt this be the server_id?

htahir1 (commented):
I had a couple of questions otherwise looks good

src/zenml/zen_stores/sql_zen_store.py
fa9r: I guess you could insert `str(GlobalConfiguration().user_id).replace("-", "")` again. However, I am not recommending this since it might hide unintended bugs that screw with our analytics. Overall, `identity is None` should never happen unless the user intentionally screwed with the DB using SQL which we can never fully recover from.

src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
```diff
@@ -0,0 +1,51 @@
...
+# revision identifiers, used by Alembic.
+revision = "7e4a481d17f7"
+down_revision = "7834208cc3f6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    from zenml.config.global_config import GlobalConfiguration
```
fa9r, line 21: This is somewhat dangerous since it makes the migration dependent on other modules, i.e., after we add this migration we will need to retroactively modify it every time we rename/delete/modify `GlobalConfiguration().user_id`.

However, the only other solution I see is to simply reset all server IDs, which we probably want even less. So I guess there is no way around importing this here.

src/zenml/cli/server.py
```diff
@@ -488,11 +488,6 @@ def deploy(
             TerraformServerDeploymentConfig,
         )
 
-        if isinstance(server.config, TerraformServerDeploymentConfig):
-            # TODO: maybe move the server ID into the ServerDeploymentConfig
-            #   class
-            metadata["server_id"] = str(server.config.server_id)
```
fa9r, line 494: we don't need this in the analytics anymore?

fa9r (commented):
Schema and migration look good. Only thing I am wondering is why we don't need to add the server ID into the analytics metadata anymore.

src/zenml/zen_stores/sql_zen_store.py
stefannica: Yes, I had my doubts about making this an exception, but the I realized that if this row is missing from the database, something terribly wrong must've happened during the database migration, which we absolutely want to catch early.

src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
stefannica: the server_id is just the external interpretation of this value (i.e. from the PoV of the client). From this point on, the server will always report the value in the database when queried through the `/api/v1/info` endpoint instead of the `GlobalConfiguration().user_id` value. This way, all clients will always get the value in the database.

src/zenml/zen_stores/migrations/versions/7e4a481d17f7_add_identity_table.py
stefannica: That's a very good point. I don't know of a better way to do this either. Perhaps we can remove the user_id from the global configuration in a future release and just populate the database with a random ID here in its place.

src/zenml/cli/server.py
stefannica: After my changes, it's impossible to know the server ID in advance without connecting to it, which is another analytics event in itself. For this reason, I prefer to just not include it in the deployment events.

fa9r (approved):

src/zenml/zen_stores/sql_zen_store.py
htahir1: in this case, i would tell the user to create a new database or do something. this exception would be incredibly difficult to deal with otherwise

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -726,10 +727,25 @@ def get_store_info(self) -> ServerModel:
...
+                raise KeyError(
+                    "The deployment ID could not be loaded from the database."
```
htahir1, line 746: As mention in the comment above, it would be good to see a 'way out' for the user in this case. Mayebe something like, please consider dropping the database or something

htahir1 (commented):
One last comment but I do somehow have to understand this - will the server_id not be available anymore in the telemetry?

stefannica:
> One last comment but I do somehow have to understand this - will the server_id not be available anymore in the telemetry?

@htahir1 of course the `server_id` will be available. Nothing in this PR is breaking the previous telemetry assumptions about fields and collected events. The difference is that `server_id` will be now read from the server's database instead of the server's global configuration file.

htahir1 (approved):

PR #1176, opened 2022-12-13
Add cloud CI environments

stefannica:
## Describe changes
Adds cloud test environments in AWS and GCP to be used locally as well as in the CI.

The PR also includes some changes required to allow using the AWS/GCP test environments and running the existing integration test cases successfully:
* update some integration requirements to reduce or remove conflicts during installation
* allow test environment secrets to not be resolved locally (i.e. cloud environments are not available everywhere)
* add step operator example to integration tests 
* fixes for minor bugs in the test framework

## Pre-requisites

## Types of changes


.dockerignore
```diff
@@ -0,0 +1,3 @@
+.venv
```
schustmi, line 1: I've personally used this dockerignore for a long time, do you see any issue with ignoring more files like this?

```
*
!/README.md
!/pyproject.toml
!/src/**
```

.github/workflows/integration-test.yml
```diff
@@ -95,13 +121,21 @@ jobs:
       # for more information: http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html)
       OBJC_DISABLE_INITIALIZE_FORK_SAFETY: 'YES'
 
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_US_EAST_1_ENV_ACCESS_KEY_ID }}
```
schustmi, line 124: Not necessary for this PR, but maybe we can optimize these regions to be as close as possible to our GH runner regions (if those are even fixed/we can find out about them)?

src/zenml/integrations/airflow/__init__.py
```diff
@@ -30,7 +30,7 @@ class AirflowIntegration(Integration):
     """Definition of Airflow Integration for ZenML."""
 
     NAME = AIRFLOW
-    REQUIREMENTS = ["apache-airflow~=2.4.0"]
+    REQUIREMENTS = ["apache-airflow~=2.5.0"]
```
schustmi, line 33: Why was this change necessary? This version is for example not supported on some managed Airflow platforms, not tested if that is going to be an issue though

schustmi (changes requested):

.github/workflows/integration-test.yml
stefannica: I did some digging. Apparently it's not possible to predict where the runners are located and they're basically all over the place (EU, US, Asia).

stefannica: From an older issue: ```
GitHub hosts runners in the following Azure regions:

East US (eastus)
East US 2 (eastus2)
West US 2 (westus2)
Central US (centralus)
South Central US (southcentralus)```

So definitely US based is better. Which means I should switch the GCP region to something in the US as well.

src/zenml/integrations/airflow/__init__.py
stefannica: If that's an issue, I'll switch it back. I think I was trying to solve a conflict with other packages but forgot which one exactly (there were so many...)

stefannica: done

.github/workflows/integration-test.yml
stefannica: done

schustmi (approved):
Can't wait to run this! 🎉

strickvl (changes requested):
Looks great to me barring one copypaste typo. Looking forward to having these all available to us!

src/zenml/integrations/aws/__init__.py
```diff
@@ -33,13 +33,9 @@ class AWSIntegration(Integration):
...
 
     NAME = AWS
-    # We need to pin protobuf to a version <=4 here, as this sagemaker release
-    # does not pin it. They fixed this in a later version, so we can probably
-    # remove this once we update the sagemaker version.
     REQUIREMENTS = [
         "boto3==1.21.0",
-        "sagemaker==2.82.2",
-        "protobuf>=3.1,<4.0.0",
+        "sagemaker==2.117.0",
```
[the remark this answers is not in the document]
stefannica: those look ok to me @strickvl , but we won't know for sure until they are all installed together with the other integrations. My reason for bumping up these dependencies was to avoid conflicts with other integrations.

strickvl (approved):

PR #1177, opened 2022-12-14
Sagemaker orchestrator integration

strickvl:
[ScreenShot 2022-12-14 at 12 08 03](https://www.loom.com/share/fdd35c829c86463f939ffe6cd53778ee)

(An overview video showcasing the main landmarks of the integration code / configuration). Note that it fails at the end of the video, but that was just my boto credentials having expired. It's working again now after I renewed my credentials.

**Breaking Change**: We're bumping the Sagemaker version we support such that the old versions specified for the AWS integration aren't compatible with the new ones.

## To be added:

- [x] Docs
- [ ] Example (will write the README but won't make an example following a discussion on merging our examples)
- [ ] Integration test(s)
- [ ] Update integration table and corresponding website section

## Pre-requisites

## Types of changes


src/zenml/integrations/aws/__init__.py
```diff
@@ -27,6 +27,7 @@
 AWS_SECRET_MANAGER_FLAVOR = "aws"
 AWS_CONTAINER_REGISTRY_FLAVOR = "aws"
 AWS_SAGEMAKER_STEP_OPERATOR_FLAVOR = "sagemaker"
+AWS_SAGEMAKER_ORCHESTRATOR_FLAVOR = "sagemaker"
 
 
 class AWSIntegration(Integration):
```
htahir1, line 33: Honestly, why dont we have a seperate integration for this? It seems like we are overloading the dependencies. I also had the same feeling with Vertex

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -0,0 +1,90 @@
...
+    """Config for the Sagemaker orchestrator.
+    Attributes:
+        instance_type: The instance type to use for the processing job.
+        bucket: Name of the S3 bucket to use for storing artifacts
+            from the job run. If not provided, a default bucket will be created
+            based on the following format: "sagemaker-{region}-{aws-account-id}".
+    """
+
+    # 'ml.t3.medium' is the default instance for Sagemaker processing jobs
+    instance_type: str = "ml.t3.medium"
```
htahir1, line 43: Rename to default_instance_type as its used in every step by default unless overriden by settings?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,157 @@
...
+
+            processor = sagemaker.processing.Processor(
+                role=sagemaker.get_execution_role(),
+                image_uri=image_name,
+                instance_count=1,
+                instance_type=self.config.instance_type,
+                entrypoint=entrypoint,
+                sagemaker_session=session,
+                base_job_name=deployment.run_name,
+                env={
```
htahir1, line 131: We can give the ability to pass in more things in the env here via settings and config

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,157 @@
...
+            arguments = StepEntrypointConfiguration.get_entrypoint_arguments(
+                step_name=step_name
+            )
+            entrypoint = command + arguments
+
+            processor = sagemaker.processing.Processor(
+                role=sagemaker.get_execution_role(),
+                image_uri=image_name,
+                instance_count=1,
+                instance_type=self.config.instance_type,
```
htahir1, line 127: We can give more control to the user for instance type

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,157 @@
...
+        sagemaker_steps = []
+        for step_name, step in deployment.steps.items():
+            command = StepEntrypointConfiguration.get_entrypoint_command()
+            arguments = StepEntrypointConfiguration.get_entrypoint_arguments(
+                step_name=step_name
+            )
+            entrypoint = command + arguments
+
+            processor = sagemaker.processing.Processor(
+                role=sagemaker.get_execution_role(),
```
htahir1, line 124: The user should be able to define this role as well and we should only default to this default execution role if not specified?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,157 @@
...
+        ).replace("_", "-")
+
+        # construct the pipeline from the sagemaker_steps
+        pipeline = Pipeline(
+            name=orchestrator_run_name,
+            steps=sagemaker_steps,
+            sagemaker_session=session,
+        )
+
+        pipeline.create(role_arn=sagemaker.get_execution_role())
```
htahir1, line 156: Again, the role should be something the user has more control over iMO

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -0,0 +1,90 @@
...
+from zenml.config.base_settings import BaseSettings
+from zenml.integrations.aws import AWS_SAGEMAKER_STEP_OPERATOR_FLAVOR
+from zenml.orchestrators.base_orchestrator import BaseOrchestratorFlavor
+from zenml.step_operators.base_step_operator import BaseStepOperatorConfig
+
+if TYPE_CHECKING:
+    from zenml.integrations.aws.orchestrators import SagemakerOrchestrator
+
+
+class SagemakerOrchestratorSettings(BaseSettings):
```
htahir1, line 27: Here is where my biggest feedback comes in for this implementation. We can provide way more  control to the user on how the pipeline is executed IMO in Settings or config. I have made individual comments below but here is a summary

* Give them the chance to specify the resources per step and translate to an approapriate instance_type
* Give them the chance to attach a GPU to the instance (Same as above)
* Give them the chance to override the role
* Give the chance to pass env variables
* Potentially let them override the `Processor` or `ProcessingStep` (Not sure if possible tbh)
* Explore other options that are 'normal' to override in a sagemaker pipeline and expose them via these settings or config params


Basically, we have to think from the users perspective: What are the constraints they have usually when running pipelines? What regions are accessible to them? what buckets? What sort of control would they like on an infrastructure level?

htahir1 (changes requested):
Left an initial review

src/zenml/integrations/aws/__init__.py
strickvl: We used to, I think, but then we decided to merge them on the assumption that if someone was in the AWS ecosystem then they'd probably want it all.

src/zenml/integrations/aws/__init__.py
```diff
@@ -37,8 +38,9 @@ class AWSIntegration(Integration):
     # does not pin it. They fixed this in a later version, so we can probably
     # remove this once we update the sagemaker version.
     REQUIREMENTS = [
-        "boto3==1.21.0",
-        "sagemaker==2.82.2",
+        "boto3==1.26.26",
```
schustmi, line 41: Just FYI, this makes the PR a breaking change. I still think we should do this, but as far as I know updating the sagemaker version also allows us to remove the explicit `protobuf>=3.1,<4.0.0` dependency (and the comment above this list). This needs some testing though

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
schustmi: I agree with most of the points by @htahir1 above, but I wouldn't worry about letting them switch from `Processor` for now.

Also needs some docstrings to explain the various attributes.

src/zenml/integrations/s3/__init__.py
```diff
@@ -30,7 +30,8 @@ class S3Integration(Integration):
     """Definition of S3 integration for ZenML."""
 
     NAME = S3
-    REQUIREMENTS = ["s3fs==2022.3.0"]
+    # REQUIREMENTS = ["s3fs==2022.3.0"]
```
schustmi, line 33: This can probably be removed or reverted?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,186 @@
...
+        """Prepares or runs a pipeline on Sagemaker.
+
+        Args:
+            deployment: The deployment to prepare or run.
+            stack: The stack to run on.
+
+        Returns:
+            The result of the pipeline run.
+        """
+        session = sagemaker.Session(default_bucket=self.config.bucket)
```
schustmi, line 124: A warning that the sagemaker orchestrator doesn't support schedules would be nice at the beginning of this method. Example from tekton:
```python
if deployment.schedule:
  logger.warning(
    "The Tekton Orchestrator currently does not support the "
    "use of schedules. The `schedule` will be ignored "
    "and the pipeline will be run immediately."
  )
```

schustmi (changes requested):

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,170 @@
...
+            execution_role = step_settings.execution_role or execution_role
+
+            processor = sagemaker.processing.Processor(
+                role=execution_role,
+                image_uri=image_name,
+                instance_count=1,
+                sagemaker_session=session,
+                instance_type=step_settings.instance_type,
+                entrypoint=entrypoint,
+                base_job_name=deployment.run_name,
```
schustmi, line 143: One other small thing: This `deployment.run_name` looks like `pipeline_name_{date}_{time}`, not sure we should use this here. What is the base job name used for anyway? I guess if it's a prefix that is added to the step name, then using the `orchestrator_run_name` defined below is probably best

src/zenml/integrations/s3/__init__.py
```diff
@@ -30,7 +30,7 @@ class S3Integration(Integration):
     """Definition of S3 integration for ZenML."""
 
     NAME = S3
-    REQUIREMENTS = ["s3fs==2022.3.0"]
+    REQUIREMENTS = ["s3fs<=2022.11.0"]
```
schustmi, line 33: Can this also have a lower bound, e.g. bigger than the version we previously required? An open range is always dangerous, we might install some 2+ years old dependency like this

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
+---
+description: How to orchestrate pipelines with Amazon Sagemaker
+---
+
+The Sagemaker orchestrator is an [orchestrator](./orchestrators.md) flavor provided
+with the ZenML `aws` integration that uses [Amazon Sagemaker](https://aws.amazon.com/sagemaker)
+to run your pipelines.
```
htahir1, line 7:
```suggestion
with the ZenML `aws` integration that uses [Amazon Sagemaker Pipelines](https://sagemaker-examples.readthedocs.io/en/latest/sagemaker-pipelines/index.html)
to run your pipelines.
```

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
...
+will help you set up the infrastructure with one click.
+
+## How to use it
+
+To use the Sagemaker orchestrator, we need:
+
+* The ZenML `aws` integration installed. If you haven't done so, run 
+    ```shell
+    zenml integration install aws
+    ```
```
htahir1, line 47:
```suggestion
* The ZenML `aws` integration installed. If you haven't done so, run 

```shell
zenml integration install aws
```
```

File: docs/book/component-gallery/orchestrators/sagemaker.md [hunk 15]
```diff
@@ -0,0 +1,121 @@
...
+* [Docker](https://www.docker.com) installed and running.
+* [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) installed.
```
htahir1, line 49: Why do we need kubectl?

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
...
+* A [remote artifact store](../artifact-stores/artifact-stores.md) as part of 
+your stack (configured with an `authentication_secret` attribute) along with [a remote secrets manager](../secrets-managers/secrets-managers.md) to store that
+authentication secret.
+* A [remote container registry](../container-registries/container-registries.md) 
+as part of your stack.
+* An IAM role or user with an `AmazonSageMakerFullAccess` policy applied to it
```
htahir1, line 55: Isnt this a bit too open? Perhaps we can find a better policy to suggest?

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
...
+  as well as `sagemaker.amazonaws.com` added as a Principal Service. Full
+  details on these permissions can be found
+  [here](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-roles.html)
+  or use the ZenML recipe (when available) which will set up the necessary
+    permissions for you.
```
htahir1, line 60: I think this part needs to be fleshed out. How do you actually enable these pipelines and see the UI? I remember we went through so many hoops to do it, and id imagine it would be nice to understand how to do it for our users too. Also I think you need to create a Sagemaker role for the pipeline as you do it

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
...
+* A [remote artifact store](../artifact-stores/artifact-stores.md) as part of 
+your stack (configured with an `authentication_secret` attribute) along with [a remote secrets manager](../secrets-managers/secrets-managers.md) to store that
+authentication secret.
+* A [remote container registry](../container-registries/container-registries.md) 
+as part of your stack.
+* An IAM role or user with an `AmazonSageMakerFullAccess` policy applied to it
+  as well as `sagemaker.amazonaws.com` added as a Principal Service. Full
+  details on these permissions can be found
+  [here](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-roles.html)
+  or use the ZenML recipe (when available) which will set up the necessary
```
htahir1, line 59: Correct me if im wrong, but doesnt the role also need access to the container registry, secrets manager, and the artifact store? How else would it be interacting with these things? Maybe mention here that you need to add the neccessary permissions to the sagemaker role

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
...
+know](https://zenml.io/slack-invite)!
+
+### Additional configuration
+
+For additional configuration of the Sagemaker orchestrator, you can pass
+`SagemakerOrchestratorSettings` which allows you to configure (among others) the following attributes:
+
+* `instance_type`: The instance type to use for the Sagemaker training job.
+  (Defaults to `ml.t3.medium`.)
+* `execution_role: The IAM role to use for the Sagemaker training job.
```
htahir1, line 100:
```suggestion
* `execution_role`: The IAM role to use for the Sagemaker training job.
```

src/zenml/integrations/aws/__init__.py
htahir1: Did we decide to now keep it as it is?

src/zenml/integrations/aws/__init__.py
htahir1: Did we verify this?

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -0,0 +1,121 @@
...
+    ```shell
+    zenml integration install aws
+    ```
+* [Docker](https://www.docker.com) installed and running.
+* [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) installed.
+* A [remote artifact store](../artifact-stores/artifact-stores.md) as part of 
+your stack (configured with an `authentication_secret` attribute) along with [a remote secrets manager](../secrets-managers/secrets-managers.md) to store that
+authentication secret.
+* A [remote container registry](../container-registries/container-registries.md) 
+as part of your stack.
```
htahir1, line 54: We also need the local client of the runner (the person who does python run.py) to have a role that can launch jobs in Sagemaker

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # construct the pipeline from the sagemaker_steps
+        pipeline = Pipeline(
+            name=orchestrator_run_name,
+            steps=sagemaker_steps,
+            sagemaker_session=session,
+        )
+
+        pipeline.create(role_arn=execution_role)
+        pipeline.start()
```
htahir1, line 170: Given that @stefannica suggested we need a "synchronous" flag here to also do a `pipeline.wait()` 

Source: https://sagemaker.readthedocs.io/en/stable/workflows/pipelines/sagemaker.workflow.pipelines.html#sagemaker.workflow.pipeline._PipelineExecution.wait

Another cool add-on would be to stream the logs but thats maybe too far for now (unless you find it in the sagemaker sdk)

htahir1 (changes requested):
Ok wow we're almost there! Great so far, left comments

docs/book/component-gallery/orchestrators/sagemaker.md
strickvl: Not sure. I started with the Vertex docs. Do we not need it? It would make sense that we don't need it, but I'm not sure if it's used somewhere that I didn't touch / know about in the codebase.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
strickvl: A super simple pipeline of 2 steps adding two integers together takes about 17 or 18 minutes to run. Not sure it's a good experience for a ZenML user to add this. The maximum for that `.wait()` method would be 60*30 seconds (so 30 mins) which would not be enough for many real workflows, I think.

docs/book/component-gallery/orchestrators/sagemaker.md
htahir1: It shouldnt even be in the Vertex docs, so if its there maybe remove it there too :-)

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
htahir1: Btw you could also not wait by setting the `synchronous` flag to be false when registering the stack component

schustmi: I agree, the synchronous flag should definitely be `False` by default and it should mostly be enabled by tests, not too many other reasonable use-cases for that

strickvl: It seems the slowness is just a thing with these Processing jobs. [This](https://stackoverflow.com/questions/71256518/delay-in-executing-sagemaker-sklearn-processing-sklearnprocessor-run-job-in-sag) suggests that the two things to reduce instance load time is to reduce the image (not really possible for us) and reduce data import. I'll see if adding a dockerignore file for my test pipeline improves things significantly, but I think the 5 minute coldstart time to provision the processor instance may be something we're stuck with for this orchestrator. I'll keep looking, though.

chemsnabti: Hello,
the processor class of Sagemaker is not for me the reason, it maybe related to your AWS region, instance type, etc.
but not 15 minutes to start the instance for any reason.

src/zenml/integrations/aws/__init__.py
strickvl: Yes. New versions are good to go.

docs/book/component-gallery/orchestrators/sagemaker.md
strickvl: This contains everything we need except for access to the secrets manager. We could probably find a more selective subset of the permissions contained in `AmazonSagemakerFullAccess` but it is the one Amazon recommends when people create roles created for using Sagemaker. I think it's not an unreasonable set of permissions to use.

docs/book/component-gallery/orchestrators/sagemaker.md
strickvl: https://docs.aws.amazon.com/sagemaker/latest/dg/security-iam-awsmanpol.html has more on their managed policies.

docs/book/component-gallery/orchestrators/sagemaker.md
strickvl: All (apart from secrets manager access) are all included with the `AmazonSageMakerFullAccess`. I tested this in a fresh account yesterday.

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -0,0 +1,110 @@
...
+        max_runtime_in_seconds: The maximum runtime in seconds for the
+            processing job.
+        processor_tags: Tags to apply to the processor assigned to the step.
+    """
+
+    instance_type: str = "ml.t3.medium"
+    execution_role: Optional[str] = None
+    volume_size_in_gb: Optional[int] = 30
+    max_runtime_in_seconds: Optional[int] = 86400
+    processor_tags: Optional[Dict[str, str]] = None
```
schustmi, line 44: I think all of these can be non-optional? For the tags you could simply make it have a default value of `{}`

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,186 @@
...
+                step_name=step_name
+            )
+            entrypoint = command + arguments
+
+            step_settings = cast(
+                SagemakerOrchestratorSettings, self.get_settings(step)
+            )
+            execution_role = step_settings.execution_role or execution_role
+            # accessing value via getattr to avoid mypy error
+            tags = getattr(step_settings, "processor_tags", None)
```
schustmi, line 143: What was the mypy issue here? Maybe it's solved by making the attribute non-optional as described in my other comment?

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -0,0 +1,110 @@
...
+        execution_role: The IAM role to use for the processing job.
+        volume_size_in_gb: The size of the EBS volume to use for the processing
+            job.
+        max_runtime_in_seconds: The maximum runtime in seconds for the
+            processing job.
+        processor_tags: Tags to apply to the processor assigned to the step.
+    """
+
+    instance_type: str = "ml.t3.medium"
+    execution_role: Optional[str] = None
```
schustmi, line 41: I'm a little confused still why we need this role and the `processor_role` on the config? They seem to overwrite each other in the code?

examples/sagemaker_orchestration/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+Once everything is done on the AWS side, we will need to configure a
+stack with all of these components.
+
+* The **artifact store** to store step outputs in an S3 Bucket.
+* The Docker images that are created to run your pipeline are stored in ECR
+  **container registry**.
+* The **Sagemaker orchestrator** is responsible for running your ZenML pipeline
+  in Sagemaker.
+* The **secrets manager** contains the secrets to allow access to the metadata
```
schustmi, line 60: Metadata store revival? :P

examples/sagemaker_orchestration/README.md
```diff
@@ -0,0 +1,159 @@
...
+for standby compute.
+
+## 📄 Pre-requisites
+
+In order to run on Sagemaker you will need to do some configuration
+within AWS to create all the required resources and permissions. In total, you will need to have:
+
+- [ECR container registry](https://docs.zenml.io/component-gallery/container-registries/amazon-ecr)
+- [S3 artifact store](https://docs.zenml.io/component-gallery/artifact-stores/amazon-s3)
+- [AWS Secrets Manager](https://docs.zenml.io/component-gallery/secrets-managers/aws).
```
schustmi, line 16: I think the secrets manager isn't even needed in this example?

schustmi (changes requested):

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
strickvl: Solved with the non-optional attributes changes mentioned above. Problem was Python 3.7 and 3.8 were failing to initialise tests with a dot-notation for `step_settings.processor_tags`, yet when I changed to `step_settings['processor_tags']` then mypy complained that the `SagemakerOrchestratorSettings` object was not indexable. (`Value of type "SagemakerOrchestratorSettings" is not indexable  [index]`). The fix above was the only thing that appeased all parties, but now it's a moot point.

examples/sagemaker_orchestration/README.md
strickvl: Needed for the artifact store authentication.

schustmi: I guess in theory could be solved using AWS roles as well, but doesn't hurt to have it here

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
strickvl: You're right. I think I have this the wrong way round. The main thing is that we want to enable people to have a separate pipeline execution role from the role used for the processor. I've made a fix for that now + will test first.

htahir1 (approved):
From my side the docs look good now... the code i think is also good but might be worth writing tests for it with the new framwork

strickvl:
> From my side the docs look good now... the code i think is also good but might be worth writing tests for it with the new framwork

Spoke with @stefannica about it just now and we'll add tests once #1176 is merged in, which will support an AWS environment that can be spun up specifically.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -0,0 +1,199 @@
...
+            entrypoint = command + arguments
+
+            step_settings = cast(
+                SagemakerOrchestratorSettings, self.get_settings(step)
+            )
+            processor_role = (
+                step_settings.processor_role or self.config.execution_role
+            )
+
+            if step_settings.processor_tags:
```
schustmi, line 142: This if-else looks identical apart from the tags right? Can we reduce the duplication by removing it and just optionally including the tags?
```python
tags = [step_settings.processor_tags] if step_settings.processor_tags else None
Processor(..., tags =tags)
```

strickvl: Fixed.

schustmi (approved):
Great job!

PR #1178, opened 2022-12-14
Fix `PyTorchDataLoaderMaterializer` for older torch versions

fa9r:
## Describe changes

I relaxed the typing of the `PyTorchDataLoaderMaterializer` to use `Any` instead of `DataLoader[Any]` since `DataLoader` is not subscriptable under `torch<1.7`.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/pytorch/__init__.py
+++ b/src/zenml/integrations/pytorch/__init__.py
@@ -22,7 +22,7 @@ class PytorchIntegration(Integration):
     """Definition of PyTorch integration for ZenML."""
 
     NAME = PYTORCH
-    REQUIREMENTS = ["torch"]
+    REQUIREMENTS = ["torch>=1.7"]
 
     @classmethod
     def activate(cls) -> None:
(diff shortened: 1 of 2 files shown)


htahir1 (approved):
I suppose this is fine but wouldnt it be nicer to throw a warning for the lower versions in our materializer?

fa9r:
@htahir1 the problem is we can't throw a warning since the type annotation would already raise a syntax error before anything is even executed.

I changed the typing to `Any` instead of `DataLoader[Any]` now instead of restricting the torch version.

htahir1 (commented):
Isnt this a bit of a  bad pattern tbh? Why cant we stick to a typed example with the requirements change you made

fa9r:
We could, but IMO supporting torch < 1.7 seems more useful to me than having the materializer typed correctly.

@schustmi any opinions on this? Which would you choose?

schustmi:
I'm not too familiar with the usage of different torch versions, but if this is still actively used then I'd prefer the solution with the `Any` type annotation

schustmi (approved):

htahir1 (approved):
Then LGTM

PR #1180, opened 2022-12-16
Feature/dashboard 2.0

AlexejPenner:
## Describe changes
Changes to accommodate the new dashboard/ fix past oversights. This can only be merged when the new dashboard is ready

So far this branch contains the following changes:

* User models contain roles now.
* No longer assign default roles to new users (this was only implemented temporarily until the dashboard allows for assigning roles to users)
* The Pipeline Model was returning all runs, this has been reduced to 3 by default (as it used to be pre-0.20.0)
* Some models didn't have `_block_recursion` implemented correctly

## Pre-requisites

## Types of changes


fa9r (commented):
Two conceptual questions / feedback:

1) Since role assignments are project-scoped, shouldn't the user have role assignments instead of roles? What would it even mean if a user has role "admin" for instance? That he assumes this role globally for all projects?

2) Personally, I am really not a fan of `_block_recursion` and was already thinking to get rid of this on the pipeline run models. In my opinion, this was just a hack to get the optimization branch merged quickly and not a standard that should be implemented elsewhere. The main issue I have with it is that it introduces inconsistency into the models, e.g., the same run sometimes has a model in its run and sometimes not. Also, not all references can be implemented with `_block_recursion` (e.g. project can never be None) and it also introduces a lot of duplicate code within the `to_model()` functions. In my opinion, the proper way to solve this would be to define a hierarchy of which model needs to hydrate which.

AlexejPenner:
> Two conceptual questions / feedback:
> 
> 1. Since role assignments are project-scoped, shouldn't the user have role assignments instead of roles? What would it even mean if a user has role "admin" for instance? That he assumes this role globally for all projects?
> 2. Personally, I am really not a fan of `_block_recursion` and was already thinking to get rid of this on the pipeline run models. In my opinion, this was just a hack to get the optimization branch merged quickly and not a standard that should be implemented elsewhere. The main issue I have with it is that it introduces inconsistency into the models, e.g., the same run sometimes has a model in its run and sometimes not. Also, not all references can be implemented with `_block_recursion` (e.g. project can never be None) and it also introduces a lot of duplicate code within the `to_model()` functions. In my opinion, the proper way to solve this would be to define a hierarchy of which model needs to hydrate which.

1. These are global roles as we currently do not do anything with project scoped roles. But good point. Only issue is the dashboard is now built around the assumption of roles on users, not sure we can pivot that fast enough still.

2. Block recursion should never be used on a native call (i.e. get_pipelines will always give you the same Model. What block_recursion does is make sure that when you populate run_models onto the pipeline_model, they don't in turn populate their pipeline_model again.

src/zenml/zen_stores/schemas/component_schemas.py
```diff
@@ -93,23 +93,43 @@ def update(
...
         return self
 
-    def to_model(self) -> "ComponentResponseModel":
+    def to_model(
+        self, _block_recursion: bool = False
+    ) -> "ComponentResponseModel":
         """Creates a `ComponentModel` from an instance of a `StackSchema`.
 
+        Args:
+            _block_recursion: If other models should be recursively filled
```
schustmi, line 102: This docstring could be a little more clear, it isn't actually other models being recursively filled but it even prevents the returned `ComponentResponseModel` to have a user in one case and `None` in the other?

src/zenml/zen_stores/schemas/role_schemas.py
```diff
@@ -78,19 +78,31 @@ def update(self, role_update: RoleUpdateModel) -> "RoleSchema":
...
             The converted `RoleResponseModel`.
         """
-        return RoleResponseModel(
-            id=self.id,
-            name=self.name,
-            created=self.created,
-            updated=self.updated,
-            permissions=[PermissionType(p.name) for p in self.permissions],
-        )
+        if _block_recursion:
```
schustmi, line 90: Why do we need this if-condition if both branches are exactly the same?

src/zenml/zen_stores/schemas/user_schemas.py
```diff
@@ -129,6 +129,7 @@ def to_model(
                 full_name=self.full_name,
                 created=self.created,
                 updated=self.updated,
+                roles=[ra.role.to_model(_block_recursion=True) for ra in self.assigned_roles]
```
schustmi, line 132: Can you explain why `roles` get set if blocking recursion, but e.g. teams are not?

schustmi (changes requested):

fa9r:
> 2. Block recursion should never be used on a native call (i.e. get_pipelines will always give you the same Model. What block_recursion does is make sure that when you populate run_models onto the pipeline_model, they don't in turn populate their pipeline_model again.

That is exactly the first kind of inconsistency I meant, e.g., `pipeline.runs[0].pipeline != pipeline` seems not very intuitive to me if both `pipeline` and `pipeline.runs[0].pipeline` are of the same type and are referring to the same entity. @schustmi 's comments above also illustrate the other inconsistency quite well which is that it feels arbitrary which references are blocked from being hydrated and which are not.

As an alternative to defining a clear model hierarchy, we could also interlink models that really need to link to each other recursively such that, e.g., `pipeline.runs[0].pipeline == pipeline`. I implemented a prototype of this here: https://github.com/zenml-io/zenml/pull/1191. Unfortunately, this doesn't really make sense for N-M relationships like user-team either since it could lead to chain reactions that potentially hydrate all users/teams in the entire DB.

Overall, I still think the proper way to solve this would be to define a clear hierarchy. E.g., do we really need `runs.pipeline` to return a full `PipelineRequestModel`?. AFAIK, this is currently only used to get the pipeline name of a run. If so, couldn't we just add `pipeline_name` as a field into the `PipelineRunResponseModel` instead?

src/zenml/zen_stores/schemas/user_schemas.py
AlexejPenner: removed this inconsistency

src/zenml/zen_stores/schemas/component_schemas.py
AlexejPenner: removed now

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -161,9 +164,7 @@ def to_model(
             return PipelineRunResponseModel(
                 id=self.id,
                 name=self.name,
-                stack=self.stack.to_model() if self.stack else None,
                 project=self.project.to_model(),
-                user=self.user.to_model() if self.user else None,
```
fa9r, line 166: why were these removed?

src/zenml/zen_stores/schemas/pipeline_schemas.py
```diff
@@ -89,20 +89,18 @@ def to_model(
                 id=self.id,
                 name=self.name,
                 project=self.project.to_model(),
-                user=self.user.to_model() if self.user else None,
                 docstring=self.docstring,
                 spec=PipelineSpec.parse_raw(self.spec),
                 created=self.created,
                 updated=self.updated,
-                status=status_last_x_runs,
```
fa9r, line 97: same here, why are those gone?

fa9r (approved):
First three changes all look good. The `_block_recursion`-related changes I don't fully understand (as in I have no clue what these changes do) so I'll approve under the premise that you have a quick look at the `_block_recursion` changes yourself again and make sure they do what you wanted them to ^^

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
AlexejPenner: done

src/zenml/zen_stores/schemas/pipeline_schemas.py
AlexejPenner: done

PR #1187, opened 2022-12-20
Move inbuilt Flavors into the Database

AlexejPenner:
## Describe changes
The single source of truth for flavors has moved into the database. The flavor registry has a method FlavorRegistry().register_flavors() responsible for uploading all flavors into the DB.

The sync happens in parallel with every alembic migration within the sql_zen_store.initialize() method.

This PR also adds two fields to flavors. A logo url and a docs url.

I have also created a convenience cli function to help during development. This function will help verify that all logos are valid. 

## Verify logo urls
`./zen-dev flavors verify logos`  # This will verify all logo urls

## Verify doc urls
This will verify all doc urls

`./zen-dev flavors verify docs`  

When there are changes on the docs, this can be simulated using out gitbook testpage

`./zen-dev flavors verify docs -r https://zenml-io.gitbook.io/test-space/`  

## Sync flavors
Using the following command, everyone developing a new flavor can manually trigger the sync to store the flavor in the db during local development
`./zen-dev flavors sync`

TODO:
[] Tests
[] Documentation 

## TODOs before this can be merged
[] Ensure Migration script does not break previously registered custom flavors

## Pre-requisites

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1434,6 +1434,10 @@ def create_flavor(self, flavor: FlavorRequestModel) -> FlavorResponseModel:
                 integration=flavor.integration,
                 project_id=flavor.project,
                 user_id=flavor.user,
+                logo_url=flavor.logo_url,
+                configuration=json.dumps(
```
htahir1, line 1438: Maybe handle exceptions here more gracefully?

src/zenml/models/flavor_models.py
```diff
@@ -49,14 +88,17 @@ class FlavorBaseModel(BaseModel):
         title="The name of the integration that the Flavor belongs to.",
         max_length=STR_FIELD_MAX_LENGTH,
     )
+    logo_url: str = Field(default="https://tinyurl.com/m4xab3yj")
```
htahir1, line 91: Will this tinyurl expire at some point? It seems like it takes me to https://github.com/zenml-io. Is that the intended behavior?

src/zenml/enums.py
```diff
@@ -151,3 +151,15 @@ class PermissionType(StrEnum):
     WRITE = "write"  # allows the user to create, update, delete everything
     READ = "read"  # allows the user to read everything
     ME = "me"  # allows the user to self administrate (change name, password...)
+
+
+class FlavorFieldDataType(StrEnum):
```
htahir1, line 156: Dont we want a more generic name as @fa9r might reuse this for the metadata story?

htahir1 (changes requested):
Left some comments but id ask @fa9r  and @stefannica to take a look at this as well to get some more opinions into this debate

src/zenml/models/flavor_models.py
AlexejPenner: will be replaced before merge once we decide on a default logo

src/zenml/enums.py
AlexejPenner: I would argue, that the Metadata story will use very different types.

src/zenml/zen_stores/schemas/flavor_schemas.py
```diff
@@ -68,6 +68,8 @@ class FlavorSchema(NamedSchema, table=True):
     )
     user: Optional["UserSchema"] = Relationship(back_populates="flavors")
 
+    logo_url: str
```
bcdurak, line 71: Can we perhaps make this an optional parameter?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1426,19 +1427,29 @@ def create_flavor(self, flavor: FlavorRequestModel) -> FlavorResponseModel:
...
-                type=flavor.type,
-                source=flavor.source,
-                config_schema=flavor.config_schema,
-                integration=flavor.integration,
-                project_id=flavor.project,
-                user_id=flavor.user,
-            )
-            session.add(new_flavor)
-            session.commit()
+            config_schema = json.dumps(flavor.config_schema)
```
bcdurak, line 1430: Just out of curiosity, is there a reason why we moved the `json.dumps` here? The `RequestModel` also has a dict instead of a str for the config schema.

bcdurak (commented):
The changes look really good overall. I left a few small comments.

I may have missed the answer in the code, but I have one question here. Where does the registration of built-in and integrated flavors to our DB happen? Previously, these flavors were registered when you create an instance of the flavor registry. However, the imports are removed, and unless there is no migration from a different version, these flavors will not be registered to our DB,

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: Having this a json in the FlavorResponseModel makes the serialization of that model very ugly, as you now have a string representation of a json within another json. This looks much cleaner and makes the dict immediatyl useable within python.

src/zenml/zen_stores/schemas/flavor_schemas.py
AlexejPenner: sure

AlexejPenner:
> The changes look really good overall. I left a few small comments.
> 
> I may have missed the answer in the code, but I have one question here. Where does the registration of built-in and integrated flavors to our DB happen? Previously, these flavors were registered when you create an instance of the flavor registry. However, the imports are removed, and unless there is no migration from a different version, these flavors will not be registered to our DB,

This all happens as part of the migration. I generated a json file with all of our flavors, and those then get added to the DB

htahir1 (approved):

bcdurak (approved):

AlexejPenner:
Thanks for the approvals, however just for stability i would not merge this into the pre-Xmas release as It its untested how this interacts with preexisting custom flavors.

AlexejPenner:
Also remember to add the new sagemaker orchestrator to the migration json-file before merging this after the release.

schustmi:
Just a general question from my side: Does that mean once this PR is merged we need to write migrations for each new flavor?

AlexejPenner:
> Just a general question from my side: Does that mean once this PR is merged we need to write migrations for each new flavor?

Yes, it would mean that. Also all significant changes to current flavors (like path changes or naming changes) would need to be migrated (with appropriate downgrade) 

Additionally, the flavor registry might need to be deprecated somewhere down the line. 

I believe this might be a halfway point to eventually removing the in-built flavors out of the package and installing them as needed from separate packages, but I think that remains to be debated.

schustmi:
> > Just a general question from my side: Does that mean once this PR is merged we need to write migrations for each new flavor?
> 
> Yes, it would mean that. Also all significant changes to current flavors (like path changes or naming changes) would need to be migrated (with appropriate downgrade)
> 
> Additionally, the flavor registry might need to be deprecated somewhere down the line.
> 
> I believe this might be a halfway point to eventually removing the in-built flavors out of the package and installing them as needed from separate packages, but I think that remains to be debated.

Thanks!

Good point regarding the removal, I see moving the flavors into separate packages as a necessity before going to 1.0 as otherwise breaking changes will occur much more often.
Regarding the flavor registry: Can't we already remove it in this PR? Or is there any use for it still?

schustmi (changes requested):
Just as a reminder: If this get's merged without the automatic registry <-> db sync, we need to add the local and kaniko image builder in the migration script

src/zenml/artifact_stores/local_artifact_store.py
```diff
@@ -149,6 +149,15 @@ def name(self) -> str:
...
         return "local"
 
+    @property
+    def logo_url(self) -> str:
+        """A url to represent the flavor in the dashboard.
+
+        Returns:
+            The flavor logo.
+        """
+        return "https://public-flavor-logos.s3.eu-central-1.amazonaws.com/artifact_Store/local.svg"
```
schustmi, line 159: artifact_Store with capital `S`?

src/zenml/zen_stores/migrations/versions/bea8a6ce3015_port_flavors_into_database.py
```diff
@@ -0,0 +1,50 @@
...
+    """Upgrade database schema and/or data, creating a new revision."""
+    with op.batch_alter_table("flavor", schema=None) as batch_op:
+
+        batch_op.add_column(sa.Column("logo_url", sa.TEXT(), nullable=True))
+        batch_op.add_column(
+            sa.Column("flavor_docs_url", sa.TEXT(), nullable=True)
+        )
+
+        batch_op.alter_column(
+            "project_id", existing_type=sa.CHAR(length=32), nullable=True
```
schustmi, line 29: workspace_id?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1597,6 +1624,16 @@ def delete_flavor(self, flavor_id: UUID) -> None:
             except NoResultFound as error:
                 raise KeyError from error
 
+    def _purge_non_custom_flavors(self) -> None:
```
schustmi, line 1627: maybe `internal` or `builtin` instead of `non_custom`?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -850,6 +852,19 @@ def migrate_database(self) -> None:
                 self.alembic.stamp(ZENML_ALEMBIC_START_REVISION)
                 self.alembic.upgrade()
 
+        # If an alembic migration took place, all non-custom flavors are purged
+        #  and the FlavorRegistry recreates all in-built and integration
+        #  flavors in the db.
+        revisions_afterwards = self.alembic.current_revisions()
```
schustmi, line 858: I'm not sure how complicated it is to test this, but I think it would be quite critical to make sure this works as intended?

src/zenml/zen_stores/schemas/flavor_schemas.py
```diff
@@ -75,10 +79,14 @@ def to_model(self) -> FlavorResponseModel:
...
             source=self.source,
-            config_schema=self.config_schema,
+            config_schema=json.loads(self.config_schema),
             integration=self.integration,
             user=self.user.to_model() if self.user else None,
-            workspace=self.workspace.to_model(),
+            workspace=self.workspace.to_model() if self.workspace else None,
             created=self.created,
             updated=self.updated,
+            logo_url=self.logo_url if self.logo_url else None,
```
schustmi, line 88:
```suggestion
            logo_url=self.logo_url,
```

src/zenml/zen_stores/schemas/flavor_schemas.py
```diff
@@ -75,10 +79,14 @@ def to_model(self) -> FlavorResponseModel:
...
+            flavor_docs_url=self.flavor_docs_url
+            if self.flavor_docs_url
+            else None,
```
schustmi, line 91:
```suggestion
            flavor_docs_url=self.flavor_docs_url,
```

src/zenml/stack/flavor_registry.py
```diff
@@ -70,102 +76,31 @@ def register_default_flavors(self) -> None:
...
+    def integration_flavors(self) -> List[Type[Flavor]]:
+        """A list of all default integration flavors."""
+        integrated_flavors = []
+        for _, integration in integration_registry.integrations.items():
+            for flavor in integration.flavors():
+                integrated_flavors.append(flavor)
+
+        return integrated_flavors
+
+    def register_default_flavors(self, store: BaseZenStore) -> None:
```
schustmi, line 92: `register_builtin_flavors` to keep it consistent with the property?

src/zenml/stack/flavor.py
```diff
@@ -34,6 +34,31 @@ def name(self) -> str:
...
+        component_type = self.type.plural.replace("_", "-")
+        name = self.name.replace("_", "-")
+        base = f"https://docs.zenml.io/v/{__version__}"
+        url = f"{base}/component-gallery/{component_type}/{name}"
+
+        return url
+
+    @property
+    @abstractmethod
+    def logo_url(self) -> str:
```
schustmi, line 55: Maybe we could make this optional, seems like quite a hassle for users getting started writing a custom integration? The dashboard could just display some placeholder image I guess?

src/zenml/stack/flavor.py
```diff
@@ -34,6 +34,31 @@ def name(self) -> str:
             The flavor name.
         """
 
+    @property
+    def flavor_docs_url(self) -> str:
```
schustmi, line 38: Again, this should probably be optional for custom integrations? And maybe shouldn't by default be implemented like this which is wrong for custom flavors?

I guess returning `None` here and then when registering it in the flavor registry, we could do
`flavor.flavor_docs_url or compute_default_docs_url()`?

Also, do we need the `flavor_` prefix or is `docs_url` fine as well?

scripts/verify_flavor_url_valid.py
```diff
@@ -0,0 +1,199 @@
...
+import requests
+import click
+
+from zenml import __version__
+from zenml.zen_stores.sql_zen_store import SqlZenStore
+
+
+@click.group()
+def cli() -> None:
+    """CLI base command for the ZenML test framework."""
```
schustmi, line 35: No :P

scripts/verify_flavor_url_valid.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@cli.group()
+def flavors() -> None:
+    """View and manage test deployments."""
```
schustmi, line 40: Also no :P

Same for the rest of the commands

schustmi (changes requested):
This looks great now with the sync, also really like the dev scripts/CLI!

Have some minor questions/renaming suggestions but nothing too wild

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: built_in are our non-integration flavors, internal feels like it doesn't capture the integrations, basically this nukes everything that is not called 'custom' - feels like the most fitting name, even if it is a bit ugly

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: My worry as well, this also relies on the fact, that with every release we create an alembic revision

scripts/verify_flavor_url_valid.py
AlexejPenner: Can't I just copy-paste in peace? :D :P

AlexejPenner: Can't I just copy-paste in peace? :D :P

src/zenml/stack/flavor.py
AlexejPenner: Yeah, I intended this to be optional, will make sure it will be.

src/zenml/stack/flavor.py
AlexejPenner: docs_url is fine

Yeah you're right, we can generate this elsewhere, stil makes sense to have it as a filed though, so that users can link to their own docs.

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: I am not sure how to write these tests here, I think in general we need to write some testts, that our alembic migrations work as intended. It would then slot in well there

src/zenml/zen_stores/sql_zen_store.py
schustmi: Good point, I was somehow confusing this with our actual builtin flavors. Not a big fan of `non_custom` but can't think of anything better either 👍

src/zenml/artifact_stores/local_artifact_store.py
schustmi: should still be `s` instead of `S` right?

src/zenml/zen_stores/migrations/versions/bea8a6ce3015_port_flavors_into_database.py
```diff
@@ -0,0 +1,51 @@
...
+            "user_id", existing_type=sa.CHAR(length=32), nullable=True
+        )
+
+
+def downgrade() -> None:
+    """Downgrade database schema and/or data back to the previous revision."""
+    with op.batch_alter_table("flavor", schema=None) as batch_op:
+
+        batch_op.drop_column("logo_url")
+
```
schustmi, line 43: docs_url also needs to be dropped I guess?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1608,6 +1635,14 @@ def delete_flavor(self, flavor_id: UUID) -> None:
             except NoResultFound as error:
                 raise KeyError from error
 
+    def _purge_non_custom_flavors(self) -> None:
+        """Delete all non-custom flavors."""
+        with Session(self.engine) as session:
+            non_custom_flavors = session.exec(
+                select(FlavorSchema).where(FlavorSchema.is_custom is False)
```
schustmi, line 1642: I think we had a similar problem with the filter model queries where we used `is` instead of sqlalchemy `.is_(...)`. This might also be a problem here?

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -484,98 +480,6 @@ def create_stack_component(
     return zen_store().create_stack_component(component=component)
 
 
-@router.get(
```
schustmi, line 487: Aren't those still workspace scoped and need an endpoint here?

schustmi (changes requested):

src/zenml/artifact_stores/local_artifact_store.py
AlexejPenner: ah i fixed it in the s3 bucket but not here, done now

src/zenml/zen_server/routers/workspaces_endpoints.py
AlexejPenner: Flavors as a concept can't be workspace scoped, as such workspace_id needs to stay optional. However, for custom flavors we can still always ensure that they stay scoped.

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: just to be safe I will switch to .is_()

AlexejPenner: So apparently you need to go with is_( when you are using the following API:

`getattr(table, "workspace_id").is_(None)`

PR #1189, opened 2022-12-21
add validator to endpoint url to replace hostname with k3d or docker …

safoinme:
## Describe changes
Add validator on endpoint URL to change localhost to k3d or docker hostname so it can be recognized by other services when running inside docker or k3d

## Pre-requisites

## Types of changes


src/zenml/integrations/s3/flavors/s3_artifact_store_flavor.py
```diff
@@ -101,6 +105,41 @@ def _convert_json_string(
...
+
+        Returns:
+            The validated value.
+        """
+        if value is None:
+            return value
+
+        if "endpoint_url" in value and value["endpoint_url"]:
+            url = value["endpoint_url"].rstrip("/")
+            scheme = re.search("^([a-z0-9]+://)", url)
```
fa9r, line 128: instead of extracting the scheme with a custom regex, I would recommend using a standard library like `urllib.parse.urlparse` for this, see https://docs.python.org/3/library/urllib.parse.html
```suggestion
            scheme = urlparse(url).scheme
```

fa9r (approved):
LGTM! General question, does this only apply to S3 or should we also do this for GCP and Azure?

stefannica (approved):

PR #1193, opened 2022-12-21
Fix Pandas Materializer Index

safoinme:
## Describe changes
Fix pandas materializer indexing by forcing the save of the index as an extra column and loading it back when reading CSV.

## Pre-requisites

## Types of changes


htahir1 (commented):
Is it possible to write some tests for this please with various dataframes and cases?

htahir1 (changes requested):
Great that looks good now. I just have one wish to test this with a datetime index just to check it out!

htahir1 (commented):
Ok and now last question: What happens to users who have materialized the dataframes with older versions of ZenML? :D

safoinme:
@htahir1 it depends on what the older version means if it's pre-0.30 which was done only using pyarrow it will raise an import error telling them to `pip install pyarrow` if it's post-0.30 I think nothing huge changes on the `CSV` save to cause errors! the save with pyarrow is still the best way as it is more efficient and save the whole state of the frame so they still get a warning that they must install it to get the best way

htahir1 (approved):
As long as it works on all versions then im happy to go!

schustmi (approved):

PR #1199, opened 2022-12-22
Tests for `io_utils` and removal of duplicate code

strickvl:
Tests for the methods inside `io_utils` and a couple of places where I thought the original source needed small changes.

Found some duplicate code which existed inside fileio as well as in io_utils so I removed it, leaving `fileio.py` as the canonical place for most of those kinds of functions.

## Pre-requisites

## Types of changes


tests/unit/io/test_fileio.py
```diff
@@ -89,23 +86,7 @@ def test_find_files_when_file_absent(tmp_path):
...
-    some_random_path = os.path.join(filesystem + "some_directory")
-    assert io_utils.is_remote(some_random_path)
-
-
-@given(text())
-def test_is_remote_when_using_non_remote_prefix(filesystem):
-    """is_remote returns False when path doesn't start with
-    a remote prefix"""
-    some_random_path = os.path.join(filesystem + "some_directory")
-    assert io_utils.is_remote(some_random_path) is False
```
fa9r, line 108: why were those tests removed?

tests/unit/utils/test_io_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    fileio.create_dir_if_not_exists(new_dir_path)
+    file_path = os.path.join(new_dir_path, "test.txt")
+    fileio.create_file_if_not_exists(file_path, "some_content_about_blupus")
+
+    io_utils.copy_dir(dir_path, new_dir_path, overwrite=True)
+    assert os.path.exists(new_dir_path)
+    assert os.path.isdir(new_dir_path)
+    assert os.path.exists(os.path.join(new_dir_path, "test.txt"))
+    assert os.path.isfile(os.path.join(new_dir_path, "test.txt"))
+    with open(os.path.join(new_dir_path, "test.txt"), "r") as f:
```
fa9r, line 111:
```suggestion
    assert os.path.exists(file_path)
    assert os.path.isfile(file_path)
    with open(file_path, "r") as f:
```

tests/unit/utils/test_io_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    file_path = os.path.join(dir_path, "test.txt")
+    fileio.create_file_if_not_exists(file_path, "some_content_about_aria")
+
+    new_dir_path = os.path.join(tmp_path, "test2")
+    fileio.create_dir_if_not_exists(new_dir_path)
+    file_path = os.path.join(new_dir_path, "test.txt")
+    fileio.create_file_if_not_exists(file_path, "some_content_about_blupus")
+
+    with pytest.raises(FileExistsError):
+        io_utils.copy_dir(dir_path, new_dir_path, overwrite=False)
```
fa9r, line 128: could perhaps add another assert at the end to make sure that the content is still "some_content_about_aria"

fa9r (changes requested):
Awesome, always happy to get rid of unused/duplicate code 😊 

New tests all look good. Only question: what happened to the `is_remote` tests?

tests/unit/io/test_fileio.py
strickvl: Mistake. Added back in now.

tests/unit/utils/test_io_utils.py
strickvl: Handled.

tests/unit/utils/test_io_utils.py
fa9r: ah sorry I mean to add an assert that ensures that the old file was not overwritten, so in this case we would need to assert the contents of `file_path2`

tests/unit/utils/test_io_utils.py
```diff
@@ -95,34 +95,38 @@ def test_copy_dir_overwriting_works(tmp_path):
...
-    fileio.create_file_if_not_exists(file_path, "some_content_about_blupus")
+    file_path2 = os.path.join(new_dir_path, "test.txt")
+    fileio.create_file_if_not_exists(file_path2, "some_content_about_blupus")
 
     with pytest.raises(FileExistsError):
         io_utils.copy_dir(dir_path, new_dir_path, overwrite=False)
+        assert (
+            io_utils.read_file_contents_as_string(file_path1)
+            == "some_content_about_aria"
+        )
```
fa9r, line 132: I guess it doesn't hurt to assert both contents

```suggestion
        assert (
            io_utils.read_file_contents_as_string(file_path1)
            == "some_content_about_aria"
        )
        assert (
            io_utils.read_file_contents_as_string(file_path2)
            == "some_content_about_blupus"
        )
```

fa9r (approved):
🚀

strickvl:
@bcdurak As discussed, I moved things back such that fileio is the core selection of filesystem functions, and `io_utils` is where everything else lives. I left `convert_to_str` in fileio on account of circular import issues.

tests/unit/io/test_fileio.py
strickvl: Actually, if the tests were for something in fileio, then I moved the tests over to fileio tests instead of having them here.

bcdurak (approved):
@strickvl Looks great! Amazing job with the tests as well!

PR #1200, opened 2022-12-22
Tests for `uuid_utils`

strickvl:
Tests for `uuid_utils` code.

## Pre-requisites

## Types of changes

### DIFF
--- a/tests/unit/utils/test_uuid_utils.py
+++ b/tests/unit/utils/test_uuid_utils.py
@@ -1,0 +1,13 @@
+#  Copyright (c) ZenML GmbH 2022. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at:
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_uuid_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    assert isinstance(parsed_generated, UUID)
+
+
+def test_generate_uuid_from_string_works():
+    """Test the generate_uuid_from_string function."""
+    assert isinstance(uuid_utils.generate_uuid_from_string("abc"), UUID)
+    assert uuid_utils.generate_uuid_from_string("abc") != "abc"
+    assert uuid_utils.generate_uuid_from_string("abc") == UUID(
+        "90015098-3cd2-4fb0-9696-3f7d28e17f72"
+    )
```
fa9r, line 64: This assert seems a bit too restrictive since we don't really care about the value, we just want it to be unique UUID right? Perhaps we should instead check that:
- running the function twice with same input leads to same result
- running the function with two different strings leads to different results

strickvl: Good point. Fixed in latest push.

safoinme (approved):
LGTM!

fa9r (approved):
🚀 🚀

PR #1207, opened 2023-01-02
Tests for `pydantic_utils`

strickvl:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/tests/unit/utils/test_pydantic_utils.py
+++ b/tests/unit/utils/test_pydantic_utils.py
@@ -102,9 +102,13 @@ def test_update_model_works_with_none_exclusion():
 def test_template_generator_works():
     """Tests that the template generator works."""
 
+    class ChildTestModel(BaseModel):
+        c: int
+        d: str
+
     class TestModel(BaseModel):
         a: int
-        b: int
+        b: ChildTestModel
 
     template = pydantic_utils.TemplateGenerator(TestModel).run()
-    assert template == {"a": "int", "b": "int"}
+    assert template == {"a": "int", "b": {"c": "int", "d": "str"}}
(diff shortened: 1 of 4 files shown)


bcdurak (approved):
LGTM!

tests/unit/utils/test_pydantic_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    """Tests that updating a Pydantic model using our util function works."""
+
+    class TestModel(BaseModel):
+        a: int
+        b: int
+
+    original = TestModel(a=1, b=2)
+    update = TestModel(a=3, b=4)
+    updated = pydantic_utils.update_model(original, update)
+    assert updated.a == 3
```
schustmi, line 32: I guess this could include something that tests the `exclude_unset` functionality. For example, make b an `Optional[int]` and don't set it on the update model. In that case, the original value should be preserved instead of setting the value to `None`. Or give `b` a default value that doesn't get set, tests the same thing.

tests/unit/utils/test_pydantic_utils.py
```diff
@@ -0,0 +1,96 @@
...
+        original2, update2, exclude_none=False
+    )
+    assert updated2.a == 3
+    assert updated2.b is None
+
+
+def test_template_generator_works():
+    """Tests that the template generator works."""
+
+    class TestModel(BaseModel):
```
schustmi, line 91: I think this model could be a little bit more complex, instead of two `int` properties maybe a submodel to verify that it gets resolved recursively?

schustmi (changes requested):

schustmi (approved):
LGTM!

PR #1211, opened 2023-01-02
Use `ruff` instead of `flake8`

strickvl:
As originally suggested by @schustmi, I tested out [`ruff`](https://github.com/charliermarsh/ruff) as an alternative to `flake8`. It indeed runs **much** faster.

This PR doesn't do too deep a surgery, but potentially we can also remove:

- `pydocstyle`
- `isort`
- `autoflake`

...since their functionality is covered by `ruff`.

## Pre-requisites

## Types of changes

### DIFF
--- a/examples/huggingface/steps/training/token_training_step.py
+++ b/examples/huggingface/steps/training/token_training_step.py
@@ -42,8 +42,12 @@ def token_trainer(
     )
 
     # Update label2id lookup
-    model.config.label2id = {l: i for i, l in enumerate(label_list)}
-    model.config.id2label = {i: l for i, l in enumerate(label_list)}
+    model.config.label2id = {
+        list_label: list_id for list_id, list_label in enumerate(label_list)
+    }
+    model.config.id2label = {
+        list_id: list_label for list_id, list_label in enumerate(label_list)
+    }
 
     # Prepare optimizer
     num_train_steps = (
--- a/examples/spark_distributed_programming/steps/__init__.py
+++ b/examples/spark_distributed_programming/steps/__init__.py
@@ -19,18 +19,20 @@ from zenml.integrations.spark.step_operators.spark_step_operator import (
     SparkStepOperator,
 )
 
+from .importer_step.importer_step import ImporterParameters, importer_step
+from .split_step.split_step import SplitParameters, split_step
+from .statistics_step.statistics_step import statistics_step
+from .trainer_step.trainer_step import trainer_step
+from .transformer_step.transformer_step import transformer_step
+
 step_operator = Client().active_stack.step_operator
+
 if not step_operator or not isinstance(step_operator, SparkStepOperator):
     raise RuntimeError(
         "Your active stack needs to contain a Spark step operator for this "
         "example to work."
     )
 
-from .importer_step.importer_step import ImporterParameters, importer_step
-from .split_step.split_step import SplitParameters, split_step
-from .statistics_step.statistics_step import statistics_step
-from .trainer_step.trainer_step import trainer_step
-from .transformer_step.transformer_step import transformer_step
 
 __all__ = [
     "importer_step",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -206,9 +206,9 @@ exclude = [
     "venv",
 ]
 per-file-ignores = {}
-# Never enforce `E501` (line length violations).
 ignore = ["E501", "F401"]
 src = ["src", "test"]
+# use Python 3.7 as the minimum version for autofixing
 target-version = "py37"
 ignore-init-module-imports = true
 # Disable autofix for unused imports (`F401`).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ server = ["fastapi", "uvicorn", "python-multipart", "python-jose", "fastapi-util
 black = "^22.3.0"
 pytest = "^6.2.4"
 mypy = "^0.991"
-flake8 = "^3.9.2"
+ruff = "^0.0.206"
 interrogate = "^1.4.0"
 coverage = { extras = ["toml"], version = "^5.5" }
 isort = "^5.9.3"
@@ -180,6 +180,50 @@ exclude_lines = [
     'if __name__ == "__main__":',
     "if TYPE_CHECKING:",
 ]
+
+[tool.ruff]
+line-length = 80
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".bzr",
+    ".direnv",
+    ".eggs",
+    ".git",
+    ".hg",
+    ".mypy_cache",
+    ".nox",
+    ".pants.d",
+    ".ruff_cache",
+    ".svn",
+    ".tox",
+    ".venv",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "venv",
+]
+per-file-ignores = {}
+# Never enforce `E501` (line length violations).
+ignore = ["E501", "F401"]
+src = ["src", "test"]
+target-version = "py37"
+ignore-init-module-imports = true
+# Disable autofix for unused imports (`F401`).
+unfixable = ["F401"]
+
+[tool.ruff.flake8-import-conventions.aliases]
+altair = "alt"
+"matplotlib.pyplot" = "plt"
+numpy = "np"
+pandas = "pd"
+seaborn = "sns"
+
+[tool.ruff.pydocstyle]
+# Use Google-style docstrings.
+convention = "google"
 
 [tool.isort]
 profile = "black"
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -8,7 +8,7 @@ SRC_NO_TESTS=${1:-"src/zenml tests/harness"}
 
 export ZENML_DEBUG=1
 export ZENML_ANALYTICS_OPT_IN=false
-flake8 $SRC
+ruff $SRC
 autoflake --remove-all-unused-imports --recursive --remove-unused-variables --in-place $SRC --exclude=__init__.py --check | ( grep -v "No issues detected" || true )
 isort $SRC scripts --check-only
 black $SRC  --check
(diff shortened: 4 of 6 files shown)


fa9r (commented):
Changes LGTM, however I am unsure how we can test this. Perhaps we could checkout some old commits that had flake8/pydocstyle/... issues and check whether ruff finds them as expected. Other than that we can't test much here since the CI only runs with ruff once we merge to main I believe. I also just killed the GH Action of this PR for that reason.

htahir1:
@fa9r @strickvl to test you can temporarily  change the github actions to trigger for this branch right?

strickvl:
Updated this to remove basically all of our linting / checking dependencies apart from `autoflake` (which can be removed in a few weeks), `pyspelling` and `interrogate`.

I think this update might mean that people start allowing the pre-commit scripts to run again since it's now fast enough.

As for checking, I did a bunch of manual tests on my machine to see whether things were caught or not. (Also did some fixes along the way).

strickvl:
Hmmm ruff worked, but mypy failed... Will look into this in a down moment tomorrow.

CONTRIBUTING.md
```diff
@@ -148,8 +148,6 @@ poetry run bash scripts/format.sh
 poetry run bash scripts/run-ci-checks.sh
 ```
 
-If the spell checker catches errors, you will have to add the words to `.pyspelling-ignore-words`. (Note that if you wish to run the spellchecks manually on your local machine, first install `aspell` (using `brew` or `apt-get` or whatever package manager Windows uses).)
```
schustmi, line 151: Is this not relevant anymore?

scripts/docstring.sh
```diff
@@ -7,5 +7,3 @@ DOCSTRING_SRC=${1:-"src/zenml tests/harness"}
 export ZENML_DEBUG=1
 export ZENML_ANALYTICS_OPT_IN=false
 interrogate $DOCSTRING_SRC -c pyproject.toml --color
-pydocstyle $DOCSTRING_SRC -e --count --convention=google --add-ignore=D403
```
schustmi, line 10: How about we get rid of interrogate as well? I think the other docstrings checks (pydocstyle, darglint) already covered similar/better metrics which should now be handled by ruff?

schustmi (changes requested):

schustmi:
Apart from the two comments, if you ran the formatting script locally and it doesn't do any reformatting (and passes the linting), this would be good to go from my side

scripts/docstring.sh
strickvl: Good point. Yes I think we can get rid of it completely as well.

CONTRIBUTING.md
strickvl: Good point / catch. I think when I removed that, I had thought that the spellchecks were covered there as well. Adding it back in now.

.github/workflows/ci.yml
```diff
@@ -77,11 +77,5 @@ jobs:
       test_environment: ${{ matrix.test_environment }}
     secrets: inherit
 
-  spell-check:
-    uses: ./.github/workflows/spellcheck.yml
-
```
stefannica, line 82: In spite of how happy this makes me, don't we still need the spellcheck ?

stefannica (changes requested):
Great initiative ! Can't wait to see this in action !

.github/workflows/ci.yml
strickvl: You're right. This was from when it seemed like maybe ruff handled that too. Forgot to revert the changes. Thanks for catching!

stefannica:
> Changes LGTM, however I am unsure how we can test this. Perhaps we could checkout some old commits that had flake8/pydocstyle/... issues and check whether ruff finds them as expected. Other than that we can't test much here since the CI only runs with ruff once we merge to main I believe. I also just killed the GH Action of this PR for that reason.

Just to set the record straight, changes to existing workflows (usually) don't need to be propagated to main to take effect. Adding new workflows and sometimes changing the triggers of an existing workflow is when you need that.

stefannica (approved):

fa9r:
> Just to set the record straight, changes to existing workflows (usually) don't need to be propagated to main to take effect. Adding new workflows and sometimes changing the triggers of an existing workflow is when you need that.

Oh really, I thought workflows are always defined by whatever is on main, good to know!

.pyspelling-ignore-words
```diff
@@ -495,6 +495,8 @@ PYTHONPATH
 PYxx
 PandasAnalyzer
 PandasDatasource
+PandasMaterializer
+PandasMaterializer
```
fa9r:
```suggestion
PandasMaterializer
```

.pyspelling-ignore-words
```diff
@@ -1108,6 +1111,8 @@ gritty
 grpc
 gserviceaccount
 gsutil
+gzip
+gzip
```
fa9r:
```suggestion
gzip
```

.pyspelling-ignore-words
```diff
@@ -1158,6 +1163,8 @@ ipynb
 ir
 isdir
 isinstance
+issubclass
+issubclass
```
fa9r:
```suggestion
issubclass
```

.pyspelling-ignore-words
```diff
@@ -1237,6 +1244,8 @@ mixin
 mixins
 mkdir
 mkdocs
+mkdtemp
+mkdtemp
```
fa9r:
```suggestion
mkdtemp
```

src/zenml/cli/secret.py
```diff
@@ -121,7 +121,8 @@ def register_secret(
         will interpret the value of the field `password` as the literal string
         `@password`.
 
-        Examples:
+        Examples:-
```
fa9r, line 124:
```suggestion
        Examples:
```

src/zenml/cli/secret.py
```diff
@@ -353,7 +352,8 @@ def update_secret(
         will interpret the value of the field `password` as the literal string
         `@password`.
         
-        Examples:
+        Examples:-
```
fa9r, line 355:
```suggestion
        Examples:
```

src/zenml/cli/secret.py
fa9r: or was that intentional for some reason? I don't get this formatting change

src/zenml/integrations/airflow/orchestrators/airflow_orchestrator.py
```diff
@@ -44,7 +44,9 @@
 from zenml.orchestrators.utils import get_orchestrator_run_name
 from zenml.stack import StackValidator
 from zenml.utils import daemon, io_utils
-from zenml.utils.pipeline_docker_image_builder import PipelineDockerImageBuilder
+from zenml.utils.pipeline_docker_image_builder import (
+    PipelineDockerImageBuilder,
+)
```
fa9r, line 49: This might indicate that something in the ruff config is off, the old line was exactly 80 chars so it should not have been reformatted.

fa9r: Ah I think this changed the max length to 79 now, which actually makes sense since that is PEP8 standard anyways

src/zenml/integrations/evidently/data_validators/evidently_data_validator.py
```diff
@@ -131,7 +131,7 @@ def _unpack_options(
         class with Evidently options
         * a dictionary with kwargs used as parameters for the option instance
 
-        Example:
+        Example:-
```
fa9r, line 134:
```suggestion
        Example:
```

?

src/zenml/integrations/slack/__init__.py
```diff
@@ -15,7 +15,6 @@
 
 from typing import List, Type
 
-from zenml.enums import StackComponentType
```
fa9r, line 18: Nice these are gone but why were these not flagged as unused imports previously?

src/zenml/services/container/container_service.py
```diff
@@ -115,7 +115,7 @@ class ContainerService(BaseService):
     `run` method. Upon `start`, the service will spawn a container that
     ends up calling the `run` method.
 
-    Example:
+    Example:-
```
fa9r, line 118: ?

src/zenml/services/local/local_service.py
```diff
@@ -191,7 +191,7 @@ class LocalDaemonService(BaseService):
     `run` method. Upon `start`, the service will spawn a daemon process that
     ends up calling the `run` method.
 
-    Example:
+    Example:-
```
fa9r, line 194: ?

src/zenml/utils/daemon.py
```diff
@@ -48,7 +48,7 @@ def daemonize(
     Use this decorator to easily transform any function into a daemon
     process.
 
-    Example:
+    Example:-
```
fa9r, line 51: ?

fa9r (approved):
Can't wait to have this live and finally use precommit again 😄 

One implication of this that you haven't mentioned is that the max line length was changed to 79 from 80. At least that's what it looks like from some of the code changes where 80-char lines have now been split. IMO this is fine since 79 is the PEP8 standard anyways, but we should be aware of that and make sure it's consistent with black and potential other tools we use.

schustmi (approved):

strickvl:
@fa9r we were previously inconsistent. Some places we said set line length to 80, other config files we set to 79. I'm happy to unify to 79 if that's the standard.

There are some other things where we'd set values (cyclomatic complexity is the most prominent) but that weren't actually being factored into the linting tests. As a result, we have many functions/methods with really high cyclomatic complexity but I left those to one side for now. I'll address the rest of your review in a bit.

src/zenml/integrations/slack/__init__.py
strickvl: We ignore checking `init` files. And actually I think I should add these back in because no way to know how / where they're being used. (See my question in #dumb-questions).

src/zenml/cli/secret.py
strickvl: The google style guide for docstrings (which we follow), states that a line followed by a colon like this is a section header, but then that means we fail some other rules relating to how a section should be formatted. I've made a fix for this now.

src/zenml/integrations/airflow/orchestrators/airflow_orchestrator.py
strickvl: Everything is 79 now.

PR #1212, opened 2023-01-03
Change pandas materializer to tempfile copy instead of fileio.open

safoinme:
## Describe changes
To allow external custom artifact stores, we change the way pandas materializes loads and saves files by copying the file into a temporary file instead of using fileio.open directly.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/materializers/pandas_materializer.py
+++ b/src/zenml/materializers/pandas_materializer.py
@@ -14,6 +14,7 @@
 """Materializer for Pandas."""
 
 import os
+import tempfile
 from typing import Any, Type, Union
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


htahir1 (commented):
Are we sure we want to do this just for this case? Why don't we just ask people to create a custom materializer? I think copying is a big overhead tbh and it would be better to have a switch case or something at least

safoinme:
@htahir1 that was exactly the same logic behind using fileio.open instead of copying twice just to run this! while technically even creating a new materialize is sort of additional step that can be last solution since `fileio.open` should be working just fine if implemented right in custom artifact store! 
do you think we just close this one and just suggest to people one of the 2 proper ways of doing this?

fa9r (commented):
The changes look fine but I don't understand yet why we need this. Could you give some more background on when exactly the current implementation is not working?

safoinme:
@fa9r "The current implementation requires that any custom artifact store used must have an open function implemented and that the `fileio.open` function can recognize which artifact store to use based on the schema of the given path. If either of these conditions is not met, the materialize function will raise an error because it is unable to open the file."

fa9r:
Sounds to me like "If a custom artifact store is implemented incorrectly it cannot open the file", isn't that intended behaviour? 😄

htahir1:
@fa9r As seen in the comments above, I think we already have a way forward where @safoinme is going to mention this in the docs now and will use his materialzier as an example implementation only :-)

src/zenml/materializers/pandas_materializer.py
```diff
@@ -125,9 +138,20 @@ def save(self, df: Union[pd.DataFrame, pd.Series]) -> None:
...
-            with fileio.open(self.csv_path, mode="wb") as f:
-                df.to_csv(f, index=True)
+            with tempfile.NamedTemporaryFile(
+                mode="wb", suffix=".csv", delete=False
+            ) as f:
+                df.to_csv(f.name, index=True)
+                fileio.copy(f.name, self.csv_path)
+
+        # Close and remove the temporary file
+        f.close()
```
schustmi, line 156: I think this isn't needed, the `with ...` context manager should already close the file automatically?

PR #1213, opened 2023-01-03
expose run time in step context

Amitg1:
## Describe changes
What is the request?

to have something similar to Airflow {{ ds }} : https://airflow.apache.org/docs/apache-airflow/2.0.0/macros-ref.html
a run date based on the scheduled time.

where it's needed? 
in a sensor ( first step of a pipeline) and second step of a pipeline
usually, have something like 
select * from source_tabe where date=='{{ ds }}', and then the pipeline starts.

the impl here is a result of a talk with @schustmi and @hamzamaiot regarding this feature.

I implemented a call to PipelineRunResponseModel in StepContext class in order to expose the context start time in a given run.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/steps/step_context.py
+++ b/src/zenml/steps/step_context.py
@@ -56,10 +56,10 @@ class StepContext:
     """
 
     def __init__(
-        self,
-        step_name: str,
-        output_materializers: Dict[str, Type["BaseMaterializer"]],
-        output_artifact_uris: Dict[str, str],
+            self,
+            step_name: str,
+            output_materializers: Dict[str, Type["BaseMaterializer"]],
+            output_artifact_uris: Dict[str, str],
     ):
         """Initializes a StepContext instance.
 
@@ -92,7 +92,7 @@ class StepContext:
         self._stack = Client().active_stack
 
     def _get_output(
-        self, output_name: Optional[str] = None
+            self, output_name: Optional[str] = None
     ) -> StepContextOutput:
         """Returns the materializer and artifact URI for a given step output.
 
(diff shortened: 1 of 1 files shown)


src/zenml/steps/step_context.py
```diff
@@ -56,10 +56,10 @@ def my_step(context: StepContext, ...)
     """
 
     def __init__(
-        self,
-        step_name: str,
-        output_materializers: Dict[str, Type["BaseMaterializer"]],
-        output_artifact_uris: Dict[str, str],
+            self,
```
fa9r, line 59: The formatting is off here, running `scripts/format.sh` should fix this

src/zenml/steps/step_context.py
```diff
@@ -184,3 +184,22 @@ def get_output_artifact_uri(self, output_name: Optional[str] = None) -> str:
...
         return self._get_output(output_name).artifact_uri
+
+    @staticmethod
+    def _get_context_run_time() -> PipelineRunResponseModel:
+        """
+        Returns: PipelineRunResponseModel
+        """
+
+        pipeline_run_id = Environment().step_environment.pipeline_run_id
+        pipeline_run = Client().get_pipeline_run(pipeline_run_id)
```
fa9r, line 195: So far we only called `Client` in the `__init__` function which is always called in the orchestrator environment and is guaranteed to have a connection to the ZenML server. I'm not sure if this will work in all step environments, e.g., when running with a Vertex step operator. Perhaps @schustmi or @bcdurak might know this?

fa9r (commented):
Currently it feels hard for me to review this since I don't fully understand why this change is needed in the first place. So my main request for now would be to have a more detailed PR description that explains the user story behind this change in a bit more detail.

src/zenml/steps/step_context.py
schustmi: Should not be an issue IMO

src/zenml/steps/step_context.py
```diff
@@ -184,3 +184,22 @@ def get_output_artifact_uri(self, output_name: Optional[str] = None) -> str:
...
+        return pipeline_run
+
+    def get_context_run_time(self) -> datetime.datetime:
```
schustmi, line 198: I think a more fitting name would be something like `get_run_start_time`? Run time (at least to me) seems more like the duration of the run, which this method does not return

schustmi (changes requested):

PR #1214, opened 2023-01-04
Fix `alembic_start` migration if tables exist

fa9r:
## Describe changes
Users migrating from ZenML versions 0.20.0-0.20.5 might already have the new ZenML tables in which case the `alembic_start` migration will fail with error `(pymysql.err.OperationalError) (1050, "Table 'projectschema' already exists")
`. 

To prevent this, I added a simple check that skips the `alembic_start` table creation statement if the `projectschema` table already exists.

## Pre-requisites

## Types of changes


AlexejPenner:
It may be the vacation fog, but I don't fully understand why the projectschema would raise this error but none of the other tables raise this. Is this related to our renaming of the project table to workspaces somewhere down the line?

fa9r:
@AlexejPenner any of the tables would raise this, `projectschema` is just the first one we create

fa9r:
Implementation-wise I just did it lazily in that we simply finish the migration if any of the tables already exist (which I check using `projectschema`). I guess we could have `if X in tables: op.create_table(X)` instead but it shouldn't make a difference IMO.

AlexejPenner (approved):
:seal:

stefannica (approved):

PR #1215, opened 2023-01-05
Track System Info and Python Version of Pipeline Runs

fa9r:
## Describe changes
I added two dicts `client_environment` and `orchestrator_environment` as attributes into pipeline run models and schemas to track system info and python versions of runs.

E.g., runs on my local machine now have the following environment data associated with it (similar for both client and orchestrator since I'm running locally):

```
{
    "environment": "native",
    "os": "mac", 
    "mac_version": "13.0.1", 
    "python_version": "3.10.0"
}
```

## Pre-requisites

## Types of changes


strickvl (changes requested):
LGTM, though is there a test we ca add to verify that these things are being stored?

htahir1:
@schustmi I think the answer is both

fa9r:
Okay then let's track both. Now we have `client_environment` as well as `orchestrator_environment`.

htahir1:
I think canonically we call this environment `runner` not `client`. if you'd  like to change this wording please also search through the docs for the mention of the word runner

htahir1 (approved):
Apart from the runner/client comment above LGTM

fa9r:
@htahir1 renamed "runner" to "client" in the docs (was only the vertex page actually)

fa9r:
@strickvl I added an integration test that checks whether the post execution class has both environments correctly set. Also added a unit test for the new `get_run_environment_dict` function.

strickvl (approved):

schustmi (approved):

PR #1216, opened 2023-01-05
Track Source Code of Steps

fa9r:
## Describe changes
We now track the source code of each step (so we can show it in the dashboard). If the source code is longer than 1000 chars, we truncate it to reduce storage space. 

Similarly, the `docstring` field of steps now also gets truncated if it is longer than 1000 chars and it is no longer contained in `config.Step`.

## Pre-requisites

## Types of changes


strickvl (commented):
This looks good / clean to me. I think we've set the maximum text size that can be entered to the absolute maximum, so we should be good here.

Is there a post-execution-style test we could add in here to make sure that this new feature works as hoped, and perhaps also a test to denote the limit of how much code you can stuff into the database?

Also, I'm thinking this can / should go somewhere in the docs that this is a feature of what we're keeping track of, and what can be accessed on the dashboard.

src/zenml/steps/step_decorator.py
```diff
@@ -146,6 +148,7 @@ def inner_decorator(func: F) -> Type[BaseStep]:
                     PARAM_OUTPUT_MATERIALIZERS: output_materializers,
                     PARAM_SETTINGS: settings,
                     PARAM_EXTRA_OPTIONS: extra,
+                    PARAM_SOURCE_CODE: inspect.getsource(func),
```
htahir1, line 151: Do we want to encode this to compress it / make safer?

htahir1 (changes requested):

src/zenml/steps/base_step.py
```diff
@@ -260,6 +261,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         kwargs = {**self.INSTANCE_CONFIGURATION, **kwargs}
         name = kwargs.pop(PARAM_STEP_NAME, None) or self.__class__.__name__
+        source_code = kwargs.pop(PARAM_SOURCE_CODE, None)
```
schustmi, line 264: This does not work for class-based steps now right? You can take a look at `BaseStep.caching_parameters` to see how we handle it for both

schustmi (changes requested):
Some general comments:
- I agree with @strickvl that we should have some way to catch overflowing fields. It should fail with a reasonable error message or even discard fields like the docstring/code to make it fit into the database limits.
- This also unnecessarily "bloats" the StepConfiguration which is part of the intermediate representation that we serialize. We could instead also get this information when writing the database entry, at which point I think we re-create the `BaseStep` instance anyway
- This makes the `docstring` field superfluous I think. The `inspect.getsource` call includes the docstring, are we going to visualize both?

fa9r:
`docstring` field is redundant now, yes. But it is still useful I think, e.g., if you want to display the docstring specifically as a description in the UI or similar. Not trivial to get that from the full source code, probably you'd need to write some custom regex for that

fa9r:
Regarding DB overflow, I wanted to add memory checks but it's not so trivial since the code is saved as part of the step config. I like your idea of dropping code/docstring though, will add some logic that does that in case the step config becomes too big

src/zenml/steps/step_decorator.py
fa9r: Good idea, added `zlib` encoding for the `zenml.config.Step` object in general.

src/zenml/steps/base_step.py
fa9r: Changed it to also work for class-based steps. This makes it much cleaner in general, great suggestion 👍

fa9r:
@schustmi @strickvl @htahir1 when writing a step to DB, we now encode the step configuration to reduce storage space. If that is not either, we truncate the `docstring` and `source_code` fields.

I also added two post-execution tests that ensure that `source_code` is set correctly and that it gets truncated if its too long.

tests/integration/functional/post_execution/test_pipeline_run.py
```diff
@@ -1,4 +1,4 @@
-#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
+#  Copyright (c) ZenML GmbH 2023. All Rights Reserved.
```
htahir1, line 1: 👀

htahir1 (changes requested):
Great! About compressing before database insertion I looked around on second thought as  imagine the underlying databases are already compressing the texts.. And I found this thread: https://stackoverflow.com/questions/8228950/compressing-text-before-storing-it-in-the-database

Apparently there is no real need for compressing them before entry as it makes the database less searchable. However if one does compress, then dont use TEXT to store it, rather BLOG, as the bytes might get mangled

fa9r:
Will wait for more reviews before removing the encoding again

src/zenml/config/step_configurations.py
```diff
@@ -187,3 +194,48 @@ class Step(StrictBaseModel):
...
+
+    def encode(self) -> bytes:
+        """Encodes the step into a byte string.
+
+        Returns:
+            The encoded step.
+        """
+        step_dict = self.dict(exclude_none=True)
+        step_json = json.dumps(step_dict, sort_keys=True)
+        encoded = zlib.compress(step_json.encode())
```
strickvl, line 206: Wonder whether it's worth being explicit about the level of compression here? Not quite sure where we are on the scale of speed:compression, but I don't think it's bang in the middle... https://docs.python.org/3/library/zlib.html#zlib.compress

strickvl (approved):
This looks good to me now. I had one question around the extent to which we compress, but feel free to ignore that. Perhaps defaults just are best.

schustmi (commented):
Code-wise this looks fine to me, however I still have my original concern of this bloating the intermediate representation. The `Step` class was never intended to be a catch-all class to add random stuff to, but to contain all the necessary information needed to run a step. Now that it gets added to the database, we abuse it by adding stuff like the code and docstring.
I'd also suggest doing some tests how much benefit the encoding/compression really gives us, and whether readability/searchability isn't actually more useful?

src/zenml/steps/step_decorator.py
fa9r: Removed encoding/decoding again following reviewer consensus

src/zenml/config/step_configurations.py
fa9r: Removed encoding again according to reviewer consensus

fa9r:
@schustmi I removed both `docstring` and `source_code` from the step config as discussed.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -186,6 +189,28 @@ def launch(self) -> None:
...
+
+        Returns:
+            The docstring and source code of the step.
+        """
+        from zenml.steps.base_step import BaseStep
+
+        step_instance = BaseStep.load_from_source(self._step.spec.source)
+
+        docstring = step_instance.docstring
+        if docstring and len(docstring) > 1000:
```
schustmi, line 205: Shouldn't this be `TEXT_FIELD_MAX_LENGTH<3` or something? 1000 seems quite a low amount of characters for code/docstring?

src/zenml/steps/base_step.py
```diff
@@ -322,6 +323,21 @@ def _created_by_functional_api(cls) -> bool:
             PARAM_CREATED_BY_FUNCTIONAL_API, False
         )
 
+    @classmethod
```
schustmi, line 326: Love this!

schustmi (changes requested):

src/zenml/orchestrators/step_launcher.py
fa9r: I thought about that but intentionally kept it at 1000. The use case of this is only to have something to display in the UI, I don't think it's worth potentially saving 65KB of data per step for that

schustmi (approved):

PR #1217, opened 2023-01-05
Enable cache for all steps that have explicit `enable_cache=True`

fa9r:
## Describe changes
This PR changes the precedence of `enable_cache` configs from `run > pipeline > step` to `run > step > pipeline`.

### Human Readable Explanation
Previously, if `enable_cache=False` was set on pipeline level, caching would always be disabled, no matter what was set at step level. 

Now, if `enable_cache=True` is explicitly set for a step, caching is enabled for that step even if `enable_cache=False` is set for the pipeline in general.

## Pre-requisites

## Types of changes


src/zenml/orchestrators/cache_utils.py
```diff
@@ -31,6 +31,30 @@
...
+        pipeline_enable_cache: The enable cache parameter of the pipeline.
+
+    Returns:
+        True if caching is enabled, False otherwise.
+    """
+    if step_enable_cache is not None:
+        return step_enable_cache
+    if pipeline_enable_cache is not None:
+        return pipeline_enable_cache
+    return True
```
strickvl, line 55:
```suggestion
    return step_enable_cache or pipeline_enable_cache or True
```

src/zenml/post_execution/pipeline_run.py
```diff
@@ -147,12 +147,14 @@ def extra(self) -> Dict[str, Any]:
...
 
     @property
-    def enable_cache(self) -> bool:
+    def enable_cache(self) -> Optional[bool]:
         """Returns whether caching is enabled for this pipeline run.
 
         Returns:
             True if caching is enabled for this pipeline run.
         """
+        if "enable_cache" not in self.pipeline_configuration:
```
strickvl, line 156: Worth refactoring this out to a constant somewhere? We've been burned by these strings in the past.

strickvl (approved):
I like it! + the beautification you did along the way.

Two minor suggestions, but otherwise looks good to me.

src/zenml/orchestrators/cache_utils.py
fa9r: That's certainly not equivalent since `x or y or True` is always `True` :D

src/zenml/post_execution/pipeline_run.py
fa9r: We even have this as a constant in `base_pipeline` already, I imported and used that one instead of the strings 👍

fa9r:
@strickvl requesting review again since I added additional logic with corresponding test that ensures that `enable_cache` is also overridden for steps if set at run level.

strickvl (approved):
Approved pending / assuming CI fixes.

schustmi (commented):
Maybe this is due to me being familiar/having used the old approach for some time, but this somehow feels wrong to me?
Imagine we have a step zoo in the future (or teams are sharing steps using a github repo), and people upload their steps with caching explicitly enabled/disabled. I would want to disable caching for this step anyway in my pipeline, and I want it to happen anytime not just in any specific run? I guess a similar argument could be made for settings, but this somehow feels different to me?

src/zenml/post_execution/pipeline_run.py
```diff
@@ -147,13 +147,17 @@ def extra(self) -> Dict[str, Any]:
...
+    def enable_cache(self) -> Optional[bool]:
         """Returns whether caching is enabled for this pipeline run.
 
         Returns:
             True if caching is enabled for this pipeline run.
         """
-        enable_cache = self.pipeline_configuration["enable_cache"]
+        from zenml.pipelines.base_pipeline import PARAM_ENABLE_CACHE
+
+        if PARAM_ENABLE_CACHE not in self.pipeline_configuration:
```
schustmi, line 158: `return self.pipeline_configuration.get(PARAM_ENABLE_CACHE)` isn't this the same?

src/zenml/pipelines/base_pipeline.py
```diff
@@ -477,6 +477,19 @@ def run(
                 pipeline=self, stack=stack, run_configuration=run_config
             )
 
+            # Override `enable_cache` of all steps if set at run level
+            if enable_cache is not None:
```
schustmi, line 481: I think this can be improved:
1) It doesn't respect values for `enable_cache` set from a config file. The combined value from config file and code would be `run_config.enable_cache`
2) The `Compiler` class currently takes care of doing all the merging of pipeline config/step config/run config, I suggest we move this there

schustmi (changes requested):

src/zenml/pipelines/base_pipeline.py
schustmi: `Compiler._apply_run_configuration` would probably the best place to do it

src/zenml/post_execution/pipeline_run.py
fa9r: mhh, yes, no clue why I wrote this convoluted garbage 😅

src/zenml/pipelines/base_pipeline.py
fa9r: Makes sense, moved that to `Compiler._apply_run_configuration` now and cleaned it up, feels much better 👍

src/zenml/config/compiler.py
```diff
@@ -134,6 +139,13 @@ def _apply_run_configuration(
                 raise KeyError(f"No step with name {step_name}.")
             pipeline.steps[step_name]._apply_configuration(step_config)
 
+        # Override `enable_cache` of all steps if set at run level
+        if config.enable_cache is not None:
+            for step_ in pipeline.steps.values():
+                step_._apply_configuration(
```
schustmi, line 145: `step.configure(enable_cache=config.enable_cache)` is cleaner I think?

src/zenml/pipelines/base_pipeline.py
```diff
@@ -158,7 +158,7 @@ def name(self) -> str:
         return self.configuration.name
 
     @property
-    def enable_cache(self) -> bool:
+    def enable_cache(self) -> Optional[bool]:
```
schustmi, line 161: Actually now looking at this again, can't we make this non-optional still? I mean, not specifying anything in the config is essentially the same as having the cache enabled right?

schustmi: Same for `BaseStep` and post execution `PipelineRun`

schustmi (changes requested):

src/zenml/pipelines/base_pipeline.py
fa9r: I would suggest to leave it as is. The current properties are all called `enable_cache`, not `cache_enabled` or similar, so I would personally expect that the property simply reflects the setting of the user.

schustmi (approved):

PR #1221, opened 2023-01-06
Test `materializer` utilities

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


tests/unit/utils/test_materializer_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    load_model_from_metadata, 
+    save_model_metadata,
+    _load_artifact,
+)
+from zenml.models import ArtifactResponseModel
+
+@pytest.fixture
+def model_artifact(mocker):
+    return mocker.Mock(spec=ArtifactResponseModel,
+                id="123",
```
fa9r, line 35: formatting is off, running the format script should fix that

tests/unit/utils/test_materializer_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    load_artifact, 
+    load_model_from_metadata, 
+    save_model_metadata,
+    _load_artifact,
+)
+from zenml.models import ArtifactResponseModel
+
+@pytest.fixture
+def model_artifact(mocker):
+    return mocker.Mock(spec=ArtifactResponseModel,
```
fa9r, line 34: what does returning a Mock instead of the actual artifact do here?

tests/unit/utils/test_materializer_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    mocker_load_artifact = mocker.patch("zenml.utils.materializer_utils._load_artifact", return_value=model)
+
+    # Load the model from the metadata file
+    model = load_model_from_metadata(model_metadata_dir)
+
+    # Ensure that the model object is returned
+    mocker_load_artifact.assert_called_once()
+    assert model is not None
+    assert isinstance(model, mocker.MagicMock)
+    assert model == model
```
fa9r, line 89: `A == A` is always True, I guess you wanted to compare to the original model from line 77?

fa9r (changes requested):
Nice unit tests, I love the effort you put into making the fixtures clean up after themselves. 

One thing that I would still like to see is a test that ensures that `load_artifact` can successfully load an actual artifact (since `test__load_artifact` currently only tests failure cases). To do so, you could take a materializer, save some random data with it, define the corresponding artifact, and then ensure that `load_artifact(artifact)` returns the original data. Strictly speaking, this won't be a unit test, but I feel like that is the main functionality of `materializer_utils` which is currently not covered with these tests.

tests/unit/utils/test_materializer_utils.py
safoinme: I apologize if I did not fully understand the question. Is returning a mock object representing an `ArtifactResponseModel` an appropriate way to implement this?

fa9r (approved):

fa9r:
wow I love the new test you came up with, that's as close to a unit test as possible I would say, very nice 👍

strickvl (approved):
Looks really nice, and a much needed test!

tests/unit/utils/test_materializer_utils.py
```diff
@@ -86,39 +98,77 @@ def test_load_model_from_metadata(mocker,model_metadata_dir):
...
+    # Test with valid materializer and data type and ensure that the artifact
+    # is loaded correctly
+    data_type = "numpy.ndarray"
+    try:
+        artifact = _load_artifact(materializer, data_type, numpy_file_uri)
+    except Exception as e:
+        assert False, f"Unexpected error: {e}"
+    else:
+        assert artifact is not None
+        assert isinstance(artifact, np.ndarray)
```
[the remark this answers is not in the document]
strickvl: The canonical way to check whether exceptions are raised is the `with pytest.raises(EXCEPTION_GOES_HERE):` pattern as described in the docs. https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions. Maybe appropriate?

PR #1225, opened 2023-01-09
Make shared stacks visible again

AlexejPenner:
## Describe changes
I synced the /stacks and /projects/<ID>/stacks endpoints to use the same logic. Also made it so that the user filter can not be used to get someone elses private stacks.

## Pre-requisites

## Types of changes


AlexejPenner:
@fa9r I know this is yet another reason to have one endpoint instead of two for each functionality, looking forward to when we get there :)

src/zenml/zen_server/routers/stacks_endpoints.py
```diff
@@ -64,10 +64,14 @@ def list_stacks(
     stacks: List[StackResponseModel] = []
 
     # Get private stacks unless `is_shared` is set to True
-    if is_shared is None or not is_shared:
+    if (
+        is_shared is None
+        or not is_shared
+        and user_name_or_id == auth_context.user.id
```
fa9r, line 70: This won't work for `list_stacks(user_name_or_id=None)` which is the default `list_stacks()` call right?
```suggestion
        and (not user_name_or_id or user_name_or_id == auth_context.user.id)
```

src/zenml/zen_server/routers/stack_components_endpoints.py
```diff
@@ -73,10 +73,14 @@ def list_stack_components(
     components: List[ComponentResponseModel] = []
 
     # Get private stack components unless `is_shared` is set to True
-    if is_shared is None or not is_shared:
+    if (
+        is_shared is None
+        or not is_shared
+        and user_name_or_id == auth_context.user.id
```
fa9r, line 79:
```suggestion
        and (not user_name_or_id or user_name_or_id == auth_context.user.id)
```
same as below

src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -329,15 +340,25 @@ def list_project_stack_components(
...
-        is_shared=False,
-    )
-    # In case the user didn't explicitly filter for is shared == False
+    components: List[ComponentResponseModel] = []
+
+    # Get private stack components unless `is_shared` is set to True
+    if (
+        is_shared is None
+        or not is_shared
+        and user_name_or_id == auth_context.user.id
```
fa9r, line 349: same as below

```suggestion
        and (not user_name_or_id or user_name_or_id == auth_context.user.id)
```

src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -231,18 +231,29 @@ def list_project_stacks(
...
-    )
-    # In case the user didn't explicitly filter for is shared == False
+    stacks: List[StackResponseModel] = []
+
+    # Get private stacks unless `is_shared` is set to True
+    # Get private stack components unless `is_shared` is set to True
+    if (
+        is_shared is None
+        or not is_shared
+        and user_name_or_id == auth_context.user.id
```
fa9r, line 241: same as below

```suggestion
        and (not user_name_or_id or user_name_or_id == auth_context.user.id)
```

fa9r (changes requested):
Looks good apart from one thing that I marked 4 times which is that the default `list_stacks()` might not contain any private stacks anymore with the new if condition check

src/zenml/zen_server/routers/stacks_endpoints.py
AlexejPenner: good point, changed it everywhere

src/zenml/zen_server/routers/stack_components_endpoints.py
AlexejPenner: good point, changed it everywhere

src/zenml/zen_server/routers/projects_endpoints.py
AlexejPenner: good point, changed it everywhere

src/zenml/zen_server/routers/projects_endpoints.py
```diff
@@ -234,20 +234,17 @@ def list_project_stacks(
...
-            project_name_or_id=project_name_or_id,
-            user_name_or_id=auth_context.user.id,
-            component_id=component_id,
-            is_shared=False,
-            name=name,
-        )
-        stacks += own_stacks
+    if is_shared is None or not is_shared:
+        # only private stacks of the authenticated user can be accessed
+        if user_name_or_id and user_name_or_id == auth_context.user.id:
```
fa9r, line 239: Still won't work (I think) 😅 

```suggestion
        if not user_name_or_id or user_name_or_id == auth_context.user.id:
```

fa9r: same below ^^

fa9r (changes requested):

fa9r (approved):
MSSVA!

bcdurak (approved):
Good catch and fix! Everything looks good.

PR #1227, opened 2023-01-11
Track Pipeline Run Schedules

fa9r:
## Describe changes
ZenML now keeps track of your pipeline run schedules. Whenever you call `pipeline.run(schedule=...)`, we now save this information in our DB and you can get a list of all your schedules via `zenml pipeline schedule list`:

```
┏━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━┯━━━━━━━━━━┯━━━━━━━━━━━━━━━━━┯━━━━━━━━━┯━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━┓
┃        NAME        │ CRON_EXPRESSION │ START_TIME │ END_TIME │ INTERVAL_SECOND │ CATCHUP │ ACTIVE │ ORCHESTRATOR_ID    │ PIPELINE_ID       ┃
┠────────────────────┼─────────────────┼────────────┼──────────┼─────────────────┼─────────┼────────┼────────────────────┼───────────────────┨
┃ first_pipeline-202 │ */5 * * * *     │ None       │ None     │ None            │ False   │ True   │ 8f79e086-3eb4-4297 │ 9b14a487-93d6-47f ┃
┃ 2_11_17-23_50_55_4 │                 │            │          │                 │         │        │ -b4dc-34df1050fe88 │ 6-ba40-87591d5a46 ┃
┃       84621        │                 │            │          │                 │         │        │                    │ 6f                ┃
┗━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━┷━━━━━━━━━━┷━━━━━━━━━━━━━━━━━┷━━━━━━━━━┷━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━┛
```

To delete a schedule from ZenML, one can use the new CLI command `zenml pipeline schedule delete`. Note, however, that this will only delete the reference saved in ZenML and does NOT stop the schedule in the respective orchestrator. This still needs to be done manually.

### Detailed changes
- New table `schedule` (`ScheduleSchema`)
- 3 new models `ScheduleRequestModel`, `ScheduleResponseModel`, `ScheduleUpdateModel`
- 5 new API endpoints (create, get, list, update, delete)
- `zenml.config.schedule.Schedule` now has a `name`
- pipeline runs now have an optional `schedule_id` field

## Pre-requisites

## Types of changes


fa9r:
Comment from @schustmi on old PR:

The schedule probably needs to be connected not only to the ID of an orchestrator, but also to the ID of the job/recurring run of the orchestrator. This only exists once the orchestrator has submitted the request (or in the case of Airflow, never).
So the generic way to handle this would maybe be a return value from orchestrator.run(...) which returns this ID, after which we could register the schedule.

I'm not really sure this is something we want to do/give users access to anyway, because even if we might be able to stop schedules for some orchestrators like KFP, others like Airflow won't support this and the schedule will be in there forever.
Additionally, there probably won't be an easy way if the user cancels the schedule on the orchestrator directly

fa9r:
@schustmi regarding above, I would suggest to just keep it as simple as possible for now and neither track the orchestrator schedule ID nor allow modifications/deletion of the schedules in the orchestrator. Tracking schedules could still be useful then to at least make users aware that there is some schedule they created in a certain orchestrator.

Also, notice how my stupid simple implementation in this PR even writes a schedule to DB for orchestrators that don't support scheduling. I intentionally did that because:
- it's much easier to implement
- it's not much of a drawback (if you accidentally create a garbage schedule you can just delete it in the CLi now; also, no one uses schedules anyways)
- it allows us to test the new schedule tracking functionality without having to use Kubeflow/Airflow/...

schustmi:
I'd actually say that displaying wrong information is much worse and more confusing than not showing anything, especially if there is way to view the correct information in the orchestrator UI. This just smells like technical debt to me.

htahir1 (changes requested):
I cant comment on the code but I see no docs being updated to reflect this new functionality :-) Some docs would be cool

src/zenml/pipelines/base_pipeline.py
```diff
@@ -499,6 +502,42 @@ def run(
                     update={"pipeline_id": pipeline_id}
                 )
 
+            if schedule:
```
schustmi, line 505: Not a huge fan of registering the schedule here without any checks whether it's possible (and if it will even get until the orchestrator, there are lots of failure points even before getting there). I'd say definitely add a comment here that this is to be changed in the future if reworking it and making it nicer doesn't work for now

schustmi (changes requested):

src/zenml/pipelines/base_pipeline.py
fa9r: Yeah me neither, but I would keep it for now because it's actually quite useful for testing. Added a "TODO" nonetheless.

schustmi (approved):

src/zenml/cli/__init__.py
```diff
@@ -744,6 +744,23 @@
...
+
+To delete a schedule, use:
+
+```bash
+zenml pipeline schedule delete <SCHEDULE_NAME_OR_ID>
+```
+
+Note, however, that this will only delete the reference saved in ZenML and does
+NOT stop/delete the schedule in the respective orchestrator. This still needs to
+be done manually.
```
strickvl, line 762:
```suggestion
NOT stop/delete the schedule in the respective orchestrator. This still needs to
be done manually. For example, using the Airflow orchestrator you would have 
to open the web UI to manually click to stop the schedule from executing.
```

src/zenml/cli/pipeline.py
```diff
@@ -102,6 +102,80 @@ def delete_pipeline(pipeline_name_or_id: str, yes: bool = False) -> None:
...
+        if not confirmation:
+            cli_utils.declare("Schedule deletion canceled.")
+            return
+
+    try:
+        Client().delete_schedule(name_id_or_prefix=schedule_name_or_id)
+    except KeyError as e:
+        cli_utils.error(str(e))
+    else:
+        cli_utils.declare(f"Deleted schedule '{schedule_name_or_id}'.")
```
strickvl, line 176: I think probably the 'completion' message of this should still always tell the user that they need to go delete it on the orchestrator itself. It's so easy to forget this otherwise, and just having it in the docs isn't enough IMO.

strickvl: Ok I see that the client is giving this message.

PR #1228, opened 2023-01-11
Tests for analytics

bcdurak:
## Describe changes
I implemented tests for analytics.

## Pre-requisites

## Types of changes


tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,173 @@
...
+    assert "version" in metadata
+    assert "event_success" in metadata
+
+    if event == AnalyticsEvent.REGISTERED_STACK:
+        assert StackComponentType.ARTIFACT_STORE in metadata
+        assert StackComponentType.ORCHESTRATOR in metadata
+
+        assert "project_id" in metadata
+        assert "entity_id" in metadata
+        assert "user_id" in metadata
```
htahir1, line 116: We can also check if the user_id is the same as the user_id in the database?

tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,173 @@
...
+    assert "environment" in metadata
+    assert "python_version" in metadata
+    assert "version" in metadata
+    assert "event_success" in metadata
+
+    if event == AnalyticsEvent.REGISTERED_STACK:
+        assert StackComponentType.ARTIFACT_STORE in metadata
+        assert StackComponentType.ORCHESTRATOR in metadata
+
+        assert "project_id" in metadata
```
htahir1, line 114: We can check if the project_id is the right project that should be tracked via querying the db

tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,173 @@
...
+        assert "entity_id" in metadata
+        assert "user_id" in metadata
+        assert "is_shared" in metadata
+
+    if event == AnalyticsEvent.REGISTERED_STACK_COMPONENT:
+        assert "type" in metadata
+        assert "flavor" in metadata
+        assert "entity_id" in metadata
+        assert "user_id" in metadata
+        assert "is_shared" in metadata
```
htahir1, line 124: We can check whether the stack is shared or not in reality

tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,173 @@
...
+
+    if event == AnalyticsEvent.RUN_PIPELINE:
+        assert "store_type" in metadata
+        assert "artifact_store" in metadata
+        assert "orchestrator" in metadata
```
htahir1, line 129: We can check that the orchestrator is the right type

htahir1 (changes requested):

tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,173 @@
...
+from zenml.config.global_config import GlobalConfiguration
+from zenml.enums import StackComponentType
 from zenml.utils.analytics_utils import (
     AnalyticsEvent,
     get_segment_key,
     track_event,
 )
 
 
+class environment_handler(object):
```
schustmi, line 31: Just did a quick scan and saw this, and was wondering if we can also use the `mocker` to patch the os.environ dict like described here: https://docs.python.org/dev/library/unittest.mock.html#unittest.mock.patch.dict

bcdurak: My so-called "quick trick" had an even quicker trick. Thanks for the tip. I replaced it and it is working!

tests/unit/utils/test_analytics_utils.py
bcdurak: Yes. Fixed this.

tests/unit/utils/test_analytics_utils.py
bcdurak: Yes. Fixed this.

tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,173 @@
...
+
+    if event == AnalyticsEvent.REGISTERED_STACK_COMPONENT:
+        assert "type" in metadata
+        assert "flavor" in metadata
+        assert "entity_id" in metadata
+        assert "user_id" in metadata
+        assert "is_shared" in metadata
+
+    if event == AnalyticsEvent.RUN_PIPELINE:
+        assert "store_type" in metadata
```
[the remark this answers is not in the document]
bcdurak: Great idea. Fixed it!

tests/unit/utils/test_analytics_utils.py
bcdurak: I have used the active stack of the client to ensure that the flavor value is correctly set in the corresponding stack component fields. The reason behind it is to avoid hardcoding the values so we can expand this test in the future or execute it with different environments/stacks.

tests/unit/utils/test_analytics_utils.py
bcdurak: I would like to avoid hard-coding this. We might want to expand the tests in the future with different types of stacks.

tests/unit/utils/test_analytics_utils.py
```diff
@@ -11,16 +11,157 @@
...
+    """
+
+
+@patch.dict(
+    "os.environ", {"ZENML_DEBUG": "true", "ZENML_ANALYTICS_OPT_IN": "true"}
+)
+def test_analytics_event(
+    mocker: MockFixture, clean_client, one_step_pipeline, empty_step
+) -> None:
+    """Checks whether the event sent for analytics has the right properties."""
```
AlexejPenner, line 135: As this departs from our other tests it might make sense to describe a bit how this test works and that the assertions take place in the mocked event_context_method

tests/unit/utils/test_analytics_utils.py
AlexejPenner: why not just use Constants at the top of the file, that way its easy to change in the future

tests/unit/utils/test_analytics_utils.py
AlexejPenner: same comment as below, why not use constants at the top of the file to keep track of these things

tests/unit/utils/test_analytics_utils.py
bcdurak: Added a comprehensive docstring. Hope it helps.

AlexejPenner (approved):
:seal:

PR #1238, opened 2023-01-16
Introduce `post_execution.BaseView`

fa9r:
## Describe changes

I added an abstract base class `BaseView` for post-execution classes.

This base class has several purposes:
- Calling `XView.y_attr` now automatically gets the value of `y_attr` from the corresponding response model, so it is no longer necessary to define a `@property` in the views for every single field of the respective underlying models.
- Post execution classes are now strongly linked to their respective models and trying to initialize a post execution class with a different model (e.g. `StepView(ArtifactResponseModel)` will now fail with a corresponding custom error message.
- `__repr__` and `__eq__` no longer need to be defined on each post-execution class. Which attributes to include in `__repr__` is now instead defined via a `REPR_KEYS` class var that is also validated during the `__init__`.

## Pre-requisites

## Types of changes


src/zenml/post_execution/base_view.py
```diff
@@ -0,0 +1,142 @@
...
+        The string representation is of the form
+        `__qualname__(<key1>=<value1>, <key2>=<value2>, ...)` where the keys
+        are the ones specified in `REPR_KEYS`.
+
+        Returns:
+            A string representation of this artifact.
+        """
+        repr = self.__class__.__qualname__
+        repr_key_strs = [
+            f"{key}={self.__getattribute__(key)}" for key in self.REPR_KEYS
```
schustmi, line 124:
```suggestion
            f"{key}={getattr(self, key)}" for key in self.REPR_KEYS
```

src/zenml/post_execution/base_view.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Returns:
+            The attribute with the given name.
+        """
+        # Handle special cases that are required by the `dir` call below
+        if __name in ("__dict__", "__class__"):
+            return super().__getattribute__(__name)
+
+        # Check the custom view properties first in case of overwrites
+        if __name in [attr for attr in dir(self) if not attr.startswith("__")]:
```
schustmi, line 102: Could be a set instead of list which is better for lookups

schustmi (changes requested):
Few minor comments, in general I think some tests for the base view would be nice to make sure it works as intended

src/zenml/post_execution/base_view.py
fa9r: Sure, done.

fa9r:
@schustmi good point on the tests, I just added a bunch of unit tests that should ensure that `BaseView` does what it's supposed to 👍

schustmi (approved):

PR #1241, opened 2023-01-17
[ANALYTICS] Allow server to override client's analytics opt-in

wjayesh:
## Describe changes

We want the client analytics opt in to be overridden by the value from the server. This would allow more control over the analytics in the case of managed ZenML servers, both by ZenML and any other organization that creates and hosts them.

### Scope
One way of ensuring this use case is:
- as soon as a server is connected, we call the /info endpoint on the server and get the value for the analytics opt-in and update the client's global config.

### Challenges
- What happens if we change the value in the server after the client is connected?
**Solution:** Should we hit the /info endpoint periodically to check if the value has changed?
or can we ignore that case now, since the only use of this feature is in our managed servers and we will not be changing the value of the analytics opt in in the server, once it is set.

- The client can simply override the value in its global config again
**Solution:** One way of solving this would be to check the `gc.store.config.analytics_opt_in` value (if it is set) instead of the value in gc, everytime an analytics call gets made. This can be done in the `__init__` function of the `AnalyticsContext` class.

### Questions

- Should I be overriding the `get_store_info()` function on the `SQLZenStore` to return an analytics variable or should I enable this extra variable for all server types.
- Adding the this would mean more complexity as the user might expect the analytics opt-in value in the server to always reflect in the global config which in its current scope, it won't.
- Is the `RestZenStoreConfig` a good place to add the analytics opt in variable?

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/global_config.py
+++ b/src/zenml/config/global_config.py
@@ -130,6 +130,8 @@ class GlobalConfiguration(BaseModel, metaclass=GlobalConfigMetaClass):
         user_email: Email address associated with this client.
         user_email_opt_in: Whether the user has opted in to email communication.
         analytics_opt_in: If a user agreed to sending analytics or not.
+        override_client_analytics_opt_in: Whether the server should override
+            the client's analytics opt-in setting using this value.
         version: Version of ZenML that was last used to create or update the
             global config.
         store: Store configuration.
@@ -145,6 +147,7 @@ class GlobalConfiguration(BaseModel, metaclass=GlobalConfigMetaClass):
     user_email: Optional[str] = None
     user_email_opt_in: Optional[bool] = None
     analytics_opt_in: bool = True
+    override_client_analytics_opt_in: Optional[bool] = None
     version: Optional[str]
     store: Optional[StoreConfiguration]
     active_stack_id: Optional[uuid.UUID]
--- a/src/zenml/config/global_config.py
+++ b/src/zenml/config/global_config.py
@@ -45,7 +45,6 @@ from zenml.utils.analytics_utils import (
     identify_user,
     track_event,
 )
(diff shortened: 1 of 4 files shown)


htahir1 (commented):
Apart from the failing tests, I would ask the question: How would this affect our analytics in Mixpanel? Would we now get 'server' events as well as client events, if both are opted in? i.e. will there be duplicate events? What ID would the server have

wjayesh:
@htahir1 That's a good point. I added a new variable in the global config whose only purpose is to determine if the client analytics setting should be overridden. While creating the servers, we can now just set this to true to serve our use case. In this scenario, only the client-side analytics will be activated, and the server won't send any metrics like usual.
As for the tests, I'll get to them once this idea and implementation is validated..

htahir1:
@wjayesh

> While creating the servers, we can now just set this to true to serve our use case. In this scenario, only the client-side analytics will be activated, and the server won't send any metrics like usual.

Do you mean when its set to false, only client-side analytics is activated?

wjayesh:
@htahir1  Yes, when it is set to false, the client-side analytics will follow whatever value the user has set there. But when set to true, the server will override the client-side analytics opt-in value and set it to true.

wjayesh:
Nope, we won't get double the metrics because the server-side analytics (decided by the value of `analytics_opt_in` in gc) is always set to false. The override is done based on the value of a new variable (`override_client_analytics_opt_in`) in the global config that has no relation to server-side analytics and only acts as a switch.

htahir1:
I would like to see a demonstration of this with tests before we merge it if thats okay

wjayesh:
Sure, I'll set some testing up and let you know.

PR #1243, opened 2023-01-17
Fix copy function to copyfile in registered filesystem

safoinme:
## Describe changes
This issue was found with the `fileio.copy` function in the code. It looks like when using it with Minio as an artifact store through the S3 integration, it's calling `src_fs.copyfile()` which was previously being referenced as `copy` when the filesystem is registered. This is the first time we've seen this error and it's a bit strange since the same step has been used successfully with three other storage systems even tho the implementation seems to be wrong.

## Pre-requisites

## Types of changes


schustmi (changes requested):
I have no idea how this happened, as the original file includes the correct `copy` method call so it works (https://github.com/tensorflow/tfx/blob/master/tfx/dsl/io/fileio.py#L46). I would however suggest doing it as in the above link and in `fileio.copy` call `filesystem.copy` instead of the change in this PR. The solution here is a little confusing because the abstract `Filesystem` interface does not define a copyfile method

schustmi (approved):

PR #1245, opened 2023-01-19
Add more examples to integration tests

schustmi:
## Describe changes
This PR adds most of our remaining examples to the integration test suite.
During this process I encountered many issues in the examples (especially when running with a containerized orchestrator) which were also fixed.

## Pre-requisites

## Types of changes


tests/harness/cfg/cloud/aws/environments.yaml
```diff
@@ -15,6 +15,9 @@ environments:
     deployment: aws-us-east-1-server
     requirements:
       - external-data-validators
+      - external-neptune-experiment-tracker
+      - external-wandb-experiment-tracker
+      - external-slack-alerter
```
stefannica, line 20: Does this mean that the AWS CI test environment now also has all these provisioned ? If yes, then wow, this is beautiful ! If no, then these shouldn't be here, because using this environment to run tests will fail from now on.

tests/harness/cfg/cloud/aws/environments.yaml
```diff
@@ -124,11 +145,34 @@ environments:
...
+      - aws-us-east-1-sagemaker-op
       - aws-us-east-1-mlflow
       - aws-us-east-1-seldon
       - aws-us-east-1-kserve
     mandatory_requirements:
       - aws-us-east-1-cloud
       - aws-us-east-1-ecr
       - aws-us-east-1-tekton
+
+  - name: aws-cloud-sagemaker-orchestrator
```
stefannica, line 160: Same here, if it is possible to use the sagemaker orchestrator to run tests now in AWS, beautiful !

.github/workflows/integration-test.yml
```diff
@@ -74,6 +74,7 @@ on:
           - aws-cloud-kubeflow-orchestrator
           - aws-cloud-kubernetes-orchestrator
           - aws-cloud-tekton-orchestrator
+          - aws-cloud-sagemaker-orchestrator
```
stefannica, line 77: Wohoo ! Does this actually work ? Did you try it ?

tests/harness/cfg/tests.yaml
```diff
@@ -178,3 +163,100 @@ tests:
           - type: step_operator
       - capabilities:
           synchronized: true
+
+  - module: tests.integration.examples.test_quickstart
```
stefannica, line 167: I'm so happy you made sense of all these !

tests/integration/examples/conftest.py
```diff
@@ -21,7 +21,7 @@
 from zenml.stack.stack import Stack
 
 
-@pytest.fixture(scope="module", autouse=True)
+@pytest.fixture(scope="function", autouse=True)
```
stefannica, line 24: I don't know if this works as you think it does, because you can't define test requirements at function level in the test harness configuration. The only thing that this change gives you is a clean stack at the level of each function that uses the fixture, but you'll still get the same exact stack configuration for all functions inside a module.

tests/integration/examples/utils.py
```diff
@@ -111,24 +120,23 @@ def run_example(
 
     yield example, runs
 
+    # Cleanup registered pipelines so they don't cause trouble in future
+    # example runs
+    for pipeline in client.depaginate(list_method=client.list_pipelines):
+        if pipeline.id not in existing_pipeline_ids:
+            client.delete_pipeline(pipeline.id)
```
stefannica, line 127: This will end up doing bad things when running tests in parallel or when running multiple CI batches at the same time on the same environment. Why don't you just use the `runs` yielded at the previous step instead ? Not 100% fail proof, but better than hacking off all pipelines created in the interval.

tests/integration/examples/utils.py
```diff
@@ -141,19 +149,17 @@ def run_example(
...
-                docker_client.images.prune()
-            except ImageNotFound:
-                pass
+            docker_client.containers.prune()
+            docker_client.volumes.prune()
+            for image_name in image_names:
+                try:
+                    logging.debug(f"Removing Docker image {image_name}")
+                    image = docker_client.images.get(image_name)
+                    docker_client.images.remove(image.id, force=True)
```
stefannica, line 158: Why use the force here ? Assuming there is no container using the image, the image should want to die voluntarily.

tests/integration/examples/utils.py
```diff
@@ -111,24 +120,23 @@ def run_example(
...
-                StackComponentType.CONTAINER_REGISTRY
-            ][0]
-            image_name = f"{container_registry.configuration['uri']}/zenml:{pipeline_name}"
-        else:
-            # For orchestrators that don't need a container registry (e.g. local
-            # Docker orchestrator and local Airflow orchestrator) and for step
-            # operators, we leverage the default convention used to name the
-            # image `zenml:<pipeline_name>`.
-            image_name = f"zenml:{pipeline_name}"
+            image_names.update(item.image for item in build.images.values())
```
stefannica, line 139: So good to see the build power in action here !

Same as before though, can you determine the list of builds and images from the list of pipeline runs returned by wait_and_validate_pipeline_run ? That should help focus this deadly laser beam a bit.

tests/integration/examples/conftest.py
stefannica: I think I understand now that it's exactly what you need: a clean stack for each function in a module, even if it has the same configuration. But why ?

tests/integration/examples/test_mlflow_deployment.py
```diff
@@ -0,0 +1,38 @@
...
+from tests.integration.examples.utils import run_example
+from zenml.post_execution.pipeline import get_pipeline
+
+
+def test_example(request: pytest.FixtureRequest) -> None:
+    """Runs the mlflow_deployment example."""
+
+    with run_example(
+        request=request,
+        name="mlflow_deployment",
```
stefannica, line 26: The mlflow deployment example starts a local daemon service to serve the model. At the end of the test, the daemon will become a zombie, destined to haunt us in the forever after. Can you do something about it ?

tests/integration/examples/example_validations.py
```diff
@@ -1,107 +0,0 @@
...
-    )
-    assert len(mlflow_runs) == 1
-    mlflow_run = mlflow_runs[0]
-
-    client = MlflowClient()
-    # fetch the MLflow artifacts logged during the first pipeline run
-    artifacts = client.list_artifacts(mlflow_run.info.run_id)
-    assert len(artifacts) == 3
-
-    from zenml.integrations.mlflow.services import MLFlowDeploymentService
```
stefannica, line 99: I kept this around for so long in the hope that we could reuse some of this to test the mlflow deployer example. Is it really un-salvageable ?

stefannica (changes requested):

tests/integration/examples/conftest.py
schustmi: Some examples have the option to run with various configurations (actually right now only the huggingface one, for which one of them is disabled) and having this a module scope messed with defining both of those test cases in the same file. Is there any reason not to have this function scoped, seems like the safer solution that works in all scenarios?

tests/integration/examples/utils.py
schustmi: Lots of examples run multiple pipelines (quickstart, mlflow deployment) in which case going by the runs doesn't work. Actually I'd argue the functionality to fetch/wait for those runs has the same issue when running in parallel? The only real solution to that would be clean workspaces for all those tests?

tests/integration/examples/utils.py
schustmi: If the same image ID is tagged twice (e.g. the training and inference pipeline of the quickstart), Docker refuses to delete the image even if no container is using it

tests/harness/cfg/cloud/aws/environments.yaml
schustmi: Yes, at least I did that at some point, not sure if it got lots during some upgrade?

tests/harness/cfg/cloud/aws/environments.yaml
schustmi: Same as above, it existed at some point

tests/integration/examples/example_validations.py
schustmi: No, it was just my lazy brain that decided to ignore it :P
I added the parts that made sense to the new validation

tests/integration/examples/test_mlflow_deployment.py
schustmi: Absolutely, I added a `service.stop()` call in case the service is still running for both the quickstart and the mlflow deployment example

tests/integration/examples/conftest.py
stefannica: I can't think of a reason not to have it function scoped.

tests/integration/examples/utils.py
stefannica: Alright, I agree that lots of examples that you added in this PR run multiple pipelines, and that you need a solution that accommodates that, but I'm still not ok with hacking off *all* of the pipelines that are registered during the time that an example is running.

Currently, it is possible to run all of the examples in parallel on a single machine (the `run_example` utility takes in a pipeline name and only looks at runs from that pipeline). Your strategy makes that impossible.

Given that you do need to run and monitor multiple pipelines in some of the examples, I would suggest the following alternatives:

1. find a way to run one pipeline at a time (example command line arguments)
2. change the `run_example` signature to allow for a list of pipeline names to be supplied, then just use those pipeline names for monitoring and cleanup. These should help narrow the search and keeping the deletion confined to the running example.

stefannica (changes requested):
Everything else looks great, but I still don't agree to the brute pipeline cleanup you implemented. I would even be inclined to overlook this, but I think there's a simple solution to it that's worth pursuing.

tests/integration/examples/utils.py
```diff
@@ -111,24 +123,26 @@ def run_example(
 
     yield example, runs
 
+    # Cleanup registered pipelines so they don't cause trouble in future
+    # example runs
+    for pipeline in client.depaginate(list_method=client.list_pipelines):
+        if (
+            pipeline.id not in existing_pipeline_ids
+            and pipeline_name in pipelines
+        ):
```
stefannica, line 132: Yes, thank you ! The senseless pipeline bloodshed is averted !

stefannica (approved):
Looks great now, thank you !

stefannica:
The evidently example is still failing on docker, but unrelated to your changes. I fixed it in #1348 .

PR #1247, opened 2023-01-20
Add support for modular recipes to the recipe CLI

wjayesh:
## Describe changes
I implemented support for modular recipes for the `zenml stack recipe` CLI commands.
Also included are additions to the example READMEs specifying how users can set up local stacks with components like MLflow, Kubeflow, Seldon, KServe and Tekton.

## User flow

In addition to making the existing recipe CLI commands work with the new modular architecture, the user can now specify what components to install using flags in the CLI too.

For example, if a user wishes to deploy a local stack with Seldon installed, they can simply run:
```
zenml stack recipe deploy k3d-modular --install seldon
```

and so on.

## Pre-requisites

## Types of changes

### DIFF
--- a/examples/kserve_deployment/README.md
+++ b/examples/kserve_deployment/README.md
@@ -84,6 +84,35 @@ deployment server so that the new model is being served instead of the old one.
 The inference pipeline loads the image from the local filesystem and performs 
 online predictions on the running KServe inference service.
 
+# 🏠 Local Stack
+
+## 📄 Infrastructure Requirements (Pre-requisites)
+
+You don't need to set up any infrastructure to run your pipelines with KServe, locally. However, you need the following tools installed:
+  * Docker must be installed on your local machine.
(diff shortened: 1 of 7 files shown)


examples/kserve_deployment/README.md
```diff
@@ -84,6 +84,35 @@ deployment server so that the new model is being served instead of the old one.
...
+> This recipe comes with MLflow, Kubeflow and Minio enabled by default. If you want any other components like Seldon or Tekton, you can specify that using the `--install/-i` flag.
+
+This will deploy a local Kubernetes cluster with KServe installed. 
+It will also generate a stack YAML file that you can import as a ZenML stack by running 
+
+```bash
+zenml stack import -f <path-to-stack-yaml>
+```
+Once the stack is set, you can then simply proceed to running your pipelines.
+
```
safoinme, line 115: I think we may want to ignore KServe as of now since it is have not been tested within the recipe

PR #1249, opened 2023-01-23
update mlflow to verion greater than 2.0

safoinme:
## Describe changes
Update Mlflow to latest version 

## Pre-requisites

## Types of changes


src/zenml/integrations/mlflow/__init__.py
```diff
@@ -34,10 +34,9 @@ class MlflowIntegration(Integration):
     # does not pin it. They fixed this in a later version, so we can probably
     # remove this once we update the mlflow version.
     REQUIREMENTS = [
-        "mlflow>=1.2.0,<1.26.0",
-        "mlserver>=0.5.3",
-        "mlserver-mlflow>=0.5.3",
-        "protobuf>=3.7.0,<4.0.0",
+        "mlflow>=1.26.0",
+        "mlserver>=1.0.0",
+        "mlserver-mlflow>=1.0.0",
```
fa9r, line 39: Did you check whether these are really the minimum compatible versions? E.g., `mlflow==1.25.0` will not work with ZenML anymore, right?

examples/mlflow_deployment/steps/predictor/predictor_step.py
```diff
@@ -26,6 +26,6 @@ def predictor(
     """Run a inference request against a prediction service."""
     service.start(timeout=10)  # should be a NOP if already started
     prediction = service.predict(data)
-    prediction = prediction.argmax(axis=-1)
+    prediction = np.apply_along_axis(lambda x: x.argmax(), -1, prediction)
```
fa9r, line 29: Doesn't this do the same as the old code?

.github/workflows/integration-test.yml
```diff
@@ -247,6 +247,7 @@ jobs:
           # re-install zenml to be sure
           poetry install --extras server
           # install airflow requirements
+          pip install numpy==1.23.5
```
fa9r, line 250: Could you explain why this is needed?

fa9r (approved):
PR changes look great. 

However, while you're working on this, I would ask you to have a quick look at a few other mlflow-deployer-related things:
- Make sure the versions specified in the integration requirements really are the maximum supported range.
- When running integration tests I get a bunch of deprecation warnings like this: `WARNING:py.warnings:/Users/felix/.pyenv/versions/zenml310/lib/python3.10/site-packages/mlflow/protos/databricks_pb2.py:473: DeprecationWarning: Call to deprecated create function Descriptor(). Note: Create unlinked descriptors is going to go away. Please use get/find descriptors from generated code or query the descriptor_pool.` I'm still using 1.25.1 locally, so maybe this will be fixed already, but would be great if you could check again whether this is related to our integration code somehow.
- Briefly think about what metadata we could extract for MLflow model deployers as part of #1151. Ideally, we would like to get the URL of the deployed model service, but perhaps there are other run-specific metadata that we might want to keep track of? (You don't need to implement this here, just think about it briefly so I don't have to 😛 )

examples/mlflow_deployment/steps/predictor/predictor_step.py
safoinme: Yes it's the same, i changed this just as sort of debugging something but forgot to revert it back

fa9r: Then I would prefer the old syntax

.github/workflows/integration-test.yml
fa9r: Ah, so this is an issue now because we use newer MLflow versions, which, in turn, use newer NumPy? In that case, how about we add `numpy<1.24.0` as an integration requirement explicitly into the MLflow integration?

The change above made seem as if this was related to Airflow somehow, that's why I was confused

AlexejPenner (approved):
Nothing to add to Felixs comments :)

.github/workflows/integration-test.yml
safoinme: I tried to fix the NumPy version on the MLFlow requirements but it didn't work! exactly as you said the upgrading of `numpy`happnes at stage 3 of the test which is related to `airflow` but it only start happening after a larger range of `mlflow`versions was supported! Should we leave it in the integration-test.yml for now? @fa9r

PR #1254, opened 2023-01-24
Renaming Project to Workspace

AlexejPenner:
## Describe changes
Renamed Project -> Workspace

Still todo:
* create some deprecated interfaces to catch people who still use old ENV vars
* migrations

## Pre-requisites

## Types of changes


docs/book/starter-guide/stacks/managing-stacks.md
```diff
@@ -163,18 +163,18 @@ Running with active project: 'default' (global)
 
 </details>
 
-With collaboration being the key part of ZenML, the 0.20.0 release puts the
-concepts of Users and Projects front and center and introduces
+With collaboration being the key part of ZenML, the 
+concepts of Users and Workspaces  is front and center, giving
```
fa9r, line 167:
```suggestion
concepts of Users and Workspaces are front and center, giving 
```

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -70,135 +70,137 @@
...
     _: AuthContext = Security(authorize, scopes=[PermissionType.READ]),
-) -> Page[ProjectResponseModel]:
-    """Lists all projects in the organization.
+) -> Page[WorkspaceResponseModel]:
+    """Lists all workspaces in the organization.
 
     Args:
-        project_filter_model: Filter model used for pagination, sorting,
+        workspace_filter_model: Filter model used for pagination, sorting,
                               filtering
```
fa9r, line 95: hanging docstring

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -207,28 +209,28 @@ def list_user_role_assignments_for_project(
     """Returns a list of all roles that are assigned to a team.
 
     Args:
-        project_name_or_id: Name or ID of the project.
+        workspace_name_or_id: Name or ID of the workspace.
         user_role_assignment_filter_model: Filter model used for pagination, sorting,
                                     filtering
```
fa9r, line 214: hanging docstring

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -237,88 +239,88 @@ def list_team_role_assignments_for_project(
     """Returns a list of all roles that are assigned to a team.
 
     Args:
-        project_name_or_id: Name or ID of the project.
+        workspace_name_or_id: Name or ID of the workspace.
         team_role_assignment_filter_model: Filter model used for pagination, sorting,
                                     filtering
```
fa9r, line 244: hanging docstring

src/zenml/zen_stores/migrations/versions/3944116bbd56_rename_project_to_workspace.py
```diff
@@ -0,0 +1,96 @@
...
+        with op.batch_alter_table(table, schema=None) as batch_op:
+            old_fk = _fk_constraint_name(table=table, column="project_id")
+            batch_op.drop_constraint(old_fk, type_="foreignkey")
+
+            batch_op.alter_column(
+                column_name="project_id", new_column_name="workspace_id"
+            )
+
+        # Once all data is moved to the new colum, create new fk
+        with op.batch_alter_table(table, schema=None) as batch_op:
```
fa9r, line 64: This could be done in the same `batch_op` I guess
```suggestion
```

src/zenml/zen_stores/migrations/versions/3944116bbd56_rename_project_to_workspace.py
```diff
@@ -0,0 +1,96 @@
...
+    for table in tables:
+        with op.batch_alter_table(table, schema=None) as batch_op:
+            old_fk = _fk_constraint_name(table=table, column="workspace_id")
+            batch_op.drop_constraint(old_fk, type_="foreignkey")
+
+            batch_op.alter_column(
+                column_name="workspace_id", new_column_name="project_id"
+            )
+
+        with op.batch_alter_table(table, schema=None) as batch_op:
```
fa9r, line 88:
```suggestion
```

src/zenml/zen_stores/migrations/versions/3944116bbd56_rename_project_to_workspace.py
```diff
@@ -0,0 +1,96 @@
...
+            )
+
+
+def downgrade() -> None:
+    """Downgrade database schema and/or data back to the previous revision."""
+    tables = _get_changed_tables()
+
+    for table in tables:
+        with op.batch_alter_table(table, schema=None) as batch_op:
+            old_fk = _fk_constraint_name(table=table, column="workspace_id")
```
fa9r, line 81: a bit nit-picky, but I'd call this one `new_fk` and the one we create `old_fk` in the downgrade

fa9r (approved):
Wow, how did you get this done so quickly? I am amazed. So much for my 5 story point estimate 😄 

I looked over all changes and also went through all "project" mentions in the entire repo manually and I think you have correctly changed it everywhere it matters. 

It's not so easy to tell which project mentions are outdated, but I believe I still saw a few in the following places:
- tests/README.md
- docs/getting-started/deploying-zenml/docker.md
- docs/book/guidelines/migration-zero-twenty.md -> I guess we want to keep `project` here since that's what it was called in 0.20.0, but perhaps we could add a disclaimer somewhere that says this was renamed since then

None of those are really important but perhaps have a quick a look at these again and adjust as you see fit.

I also checked the alembic migration, seems all good to me, I think this should work. You could verify that by manually creating another temp revision on top of this one (e.g. `alembic revision --autogenerate -m "test"`) which shouldn't do anything if the current migration code is correct.

Overall, I'd say this looks pretty much done. Any minor bugs introduced through the renaming should be spotted by the CI, so once the CI passes this can be merged I'd say 🚀

src/zenml/zen_server/routers/workspaces_endpoints.py
fa9r: Resolving this since it's technically not in the scope of this PR I guess

fa9r:
Tested migration manually with SQLite, seems all good. All create a temp migration which included no changes, so the migration should be good to go.

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -69,15 +69,25 @@
...
 )
 
+# This is a workaround to slowly deprecate the projects routes. This along with
+#  all the decorators using it, can be removed in a few releases.
+PROJECTS = "/projects"
+
+
 router = APIRouter(
-    prefix=API + VERSION_1 + WORKSPACES,
+    prefix=API + VERSION_1,
```
fa9r, line 78: how come this changed?

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -69,15 +69,25 @@
...
 
 
 @router.get(
-    "",
+    WORKSPACES,
+    response_model=Page[WorkspaceResponseModel],
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+@router.get(
+    PROJECTS,
```
fa9r, line 90: putting two `@router...` decorators on top of each other simply uses this method for two endpoints? Awesome if it works like this, but would recommend testing manually once to make sure.

src/zenml/cli/workspace.py
```diff
@@ -189,3 +189,21 @@ def delete_workspace(workspace_name_or_id: str) -> None:
...
+def project() -> None:
+    """Deprecated commands for project management."""
+    cli_utils.warning(
+        "DEPRECATION WARNING: `project` has been renamed to"
+        "`workspace`. Please switch to using "
+        "`zenml workspace ...`."
+    )
+
+
+project.add_command(list_workspaces)
```
fa9r, line 204: Also here not sure whether you can add commands from other groups like this. Awesome if it works but please check manually to make sure both versions work with this.

fa9r (approved):

src/zenml/zen_stores/migrations/versions/3944116bbd56_rename_project_to_workspace.py
AlexejPenner: that didn't work wehn I tried, I think within one context it might port only a subset of the data, and then setting the contraint won't be possible.

src/zenml/zen_server/routers/workspaces_endpoints.py
AlexejPenner: https://github.com/tiangolo/fastapi/discussions/4299 

Got it there, it works in my tests (dashboard also still works, and it heavily relies on the project routes)

src/zenml/zen_stores/migrations/versions/3944116bbd56_rename_project_to_workspace.py
fa9r: ok nevermind then

src/zenml/zen_stores/migrations/versions/3944116bbd56_rename_project_to_workspace.py
AlexejPenner: done

src/zenml/cli/workspace.py
AlexejPenner: Found it here: https://stackoverflow.com/a/45514541 - also works in my tests

src/zenml/zen_server/routers/workspaces_endpoints.py
fa9r: Cool. In the answer you linked I saw you can even do `@router.post(..., deprecated=True)`, should we also do this for all the project endpoints?

AlexejPenner: lets do it

src/zenml/cli/workspace.py
fa9r: Awesome

src/zenml/zen_server/routers/workspaces_endpoints.py
fa9r: 🔥

PR #1255, opened 2023-01-24
Integrations conditional requirements

safoinme:
## Describe changes
Check system type if needed to differentiate between dependency versions.

## Pre-requisites

## Types of changes


src/zenml/integrations/integration.py
```diff
@@ -82,6 +83,11 @@ def check_installation(cls) -> bool:
             )
             return False
 
+    @classmethod
+    def check_system(cls) -> None:
```
fa9r, line 87: Looking at this again, perhaps `update_system_specific_requirements` or something similar might be a more descriptive name? @schustmi any suggestions?

src/zenml/integrations/integration.py
```diff
@@ -61,6 +61,7 @@ def check_installation(cls) -> bool:
         Returns:
             True if all required packages are installed, False otherwise.
         """
+        cls.check_system()
```
fa9r, line 64: I'm not sure if it's enough to call this here. I think `zenml integration install tensorflow -f` for instance would never reach this because `src.zenml.cli.integration.install` does this:
```
if force or not integration_registry.is_installed(
    integration_name
):
    ... # install integration
```

fa9r: perhaps we could call this in `IntegrationMeta.__new__` instead so the requirements are overridden right when the class is created?

src/zenml/integrations/integration.py
```diff
@@ -82,6 +83,11 @@ def check_installation(cls) -> bool:
             )
             return False
 
+    @classmethod
+    def check_system(cls) -> None:
+        """Abstract method to check the system for the integration."""
```
fa9r, line 88: Also, this needs a bit longer docstring explaining what this method is supposed to do, when you would need to implement it, and when it is called. Perhaps we could even include an example.

src/zenml/integrations/integration.py
```diff
@@ -82,6 +83,11 @@ def check_installation(cls) -> bool:
...
+        pass
```
fa9r, line 89: I was thinking a bit about how we could simplify the code that subclasses need to define. How about we design the method like so:
```suggestion
    @classmethod
    def define_platform_specific_requirements(cls, platform: str) -> List[str]:
        return []
```

and then do the following in `IntegrationMeta.__new__`:

```
platform_specific_requirements = cls.define_platform_specific_requirements(platform.system)
cls.REQUIREMENTS = cls.REQUIREMENTS + platform_specific_requirements
```

Then integration subclasses would neither need `platform.system()` nor override `cls.REQUIREMENTS` every time, which would make the implementation a bit easier perhaps

fa9r (commented):
Cool functionality but not 100% sure yet how we should design it. Perhaps we should schedule a short call with @schustmi sometime so we can brainstorm a really clean solution together?

schustmi (commented):
I'm going to make a completely different suggestion that makes this much cleaner IMO:

How about we just have a `get_requirements(cls) -> List[str]` class function that specific integrations have to override? If they need to do any OS/python version specific stuff, they can do that. Otherwise they can just return a static list.

To not be breaking, we could default that to `cls.REQUIREMENTS` in the base integration class.


One additional point that is not considered in both my suggestion or your implementation:
These requirements are also queried when figuring out which packages to install in a Docker image. That means if I'm now running on MacOS and building a Docker image running ubuntu, it would install `tensorflow-macos` which would break.

I guess we could solve that by having arguments on `get_requirements(cls, target_os: Optional[str], ...)`?

fa9r:
I love @schustmi's suggestion, that's the cleanest approach so far I'd say. 

In that case, I'd say we should also change all integrations to use the new class method instead of `REQUIREMENTS`.

Regarding backward compatibility, how about we do something like the following in the meta class:

```
if getattr(cls, "REQUIREMENTS"):
    # some deprecation warning

get_requirements = getattr(cls, "get_requirements", None)
if get_requirements and callable(get_requirements):
    cls.REQUIREMENTS = cls.get_requirements()
```

safoinme:
@fa9r can you please confirm if the new implementation covers what was suggested in the reviews?

src/zenml/integrations/integration.py
```diff
@@ -43,6 +43,17 @@ def __new__(
         cls = cast(Type["Integration"], super().__new__(mcs, name, bases, dct))
         if name != "Integration":
             integration_registry.register_integration(cls.NAME, cls)
+
+        if getattr(cls, "REQUIREMENTS"):
+            logger.warning(
+                f"Integration {cls.NAME} has a REQUIREMENTS attribute. "
+                f"Please use the get_requirements() method instead."
+                f"REQUIREMENTS will be deprecated in the future."
+            )
```
fa9r, line 52: I would now even suggest to keep the current `REQUIREMENTS` syntax as well (and to use it for airflow and pytorch again).

The reason why I changed my mind on this is that the integrations might move into the Hub in the somewhat-near future at which point we will have to deprecate/adjust all of this anyways. Thus, I'd suggest to not spend too much effort on this PR right now and just go with the minimal viable solution that makes TensorFlow work on Mac.

```suggestion
```

fa9r (approved):
Nice, love it 👍 

To finish up the PR, you would still need to do one of the following:
1. implement `get_requirements()` for all integrations
2. OR: allow both `get_requirements` and `REQUIREMENTS` and remove the corresponding deprecation warning.

I'd suggest 2) since it's less effort, but up to you :P

src/zenml/integrations/integration.py
schustmi: I agree, the old way should still be possible without a warning and should simply be used as the default returned by `get_requirements` (which you already implemented here.

src/zenml/integrations/integration.py
```diff
@@ -43,6 +43,17 @@ def __new__(
...
+
+        get_requirements = getattr(cls, "get_requirements", None)
+        if get_requirements and callable(get_requirements):
+            cls.REQUIREMENTS = cls.get_requirements()
```
schustmi, line 56: The entire target os only makes sense if we then do not statically set something once. Instead:
- `get_requirements(...) by default returns `cls.REQUIREMENTS`
- In all places where we use `SomeIntegration.REQUIREMENTS` you should now use `SomeIntegration.get_requirements(...)` and optionally specify a target os if necessary (this is for example for the docker builds)

schustmi (changes requested):
This still needs changes to work with Docker builds, see `zenml.utils.pipeline_docker_image_builder.PipelineDockerImageBuilder._gather_requirements_files(...)`

src/zenml/integrations/integration.py
safoinme: @schustmi should the target_os be defaulted into specific target_os, `Linux` for example?

schustmi: I think the default should be your local OS (e.g. if you want to do `zenml integration install ...` you want the packages for your local os. Only docker builds need to explictly request packages for linux

src/zenml/config/docker_settings.py
```diff
@@ -169,6 +177,8 @@ class DockerSettings(BaseSettings):
     copy_global_config: bool = True
     user: Optional[str] = None
 
+    target_os: Optional[str] = "Linux"
```
AlexejPenner, line 180: Any reason we shouldn't use an enum for this?

AlexejPenner: Especially useful as most people would not write "Darwin" instead of "MacOS"

AlexejPenner (commented):
One question, besides that would be good to go from my side

src/zenml/config/docker_settings.py
schustmi: Actually I would suggest to remove this from the Docker settings entirely. All our Docker images are linux based, so there's really no reason to expose this.

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -426,7 +426,8 @@ def _gather_requirements_files(
         integration_requirements = set(
             itertools.chain.from_iterable(
                 integration_registry.select_integration_requirements(
-                    integration
+                    integration_name=integration,
+                    target_os=docker_settings.target_os,
```
schustmi, line 430: As mentioned on the other comment, I would hardcode this instead of exposing it on the docker settings.

schustmi (changes requested):

PR #1266, opened 2023-01-26
Rename page attributes

schustmi:
## Describe changes
Rename page attributes and include property for the actual page size.

## Pre-requisites

## Types of changes


fa9r (approved):
LGTM but might require some changes in the dashboard @AlexejPenner

tests/unit/orchestrators/test_cache_utils.py
```diff
@@ -242,8 +242,8 @@ def test_fetching_cached_step_run_queries_cache_candidates(
     mock_list_run_steps = mocker.patch(
         "zenml.client.Client.list_run_steps",
         return_value=Page(
-            page=1,
-            size=1,
+            index=1,
+            max_size=1,
```
AlexejPenner, line 246: Don't we need both? How many items are on this page, and whats the maximum size allowed per page?

AlexejPenner, line 246: Don't we need both? How many items are on this page, and whats the maximum size allowed per page? Or do you think its enough to implicitely have the current size as len(items)

schustmi: There is `page.size` as a property, but it's calculated dynamically based on `len(page.items)`

AlexejPenner (approved):
LGTM, lets not merge yet, I would like to first communicate with the Dashboard team and see how much this impacts them. :seal:

tests/unit/orchestrators/test_cache_utils.py
AlexejPenner: ok, i remember now why this might be an issue, the problem with having it as a property is, that that does not get serialized, so if the frontend needs the information, it should be a field

schustmi: But can't they simply compute it from the size of the items list as well?

schustmi: Otherwise we can also store it as a field if necessary I guess, just stores/sends "useless" info IMO

AlexejPenner: I will ask and get back

PR #1270, opened 2023-01-28
GCP Image Builder

gabrielmbmb:
## Describe changes

Hey guys, I love the Image Builder component addition! Hope I'm not late to the party 😜

* I've implemented an image builder using [Google Cloud Build](https://cloud.google.com/build).
* I've moved the functionality in `KanikoImageBuilder._upload_build_context` to a new util class called `ImageBuilderContextUploader` as the exact same functionality is used by the `GCPImageBuilder`. Maybe it's a bit overkill to create a class just for this and it can be a method in the `ImageBuilder` base class component.
* I've removed the `project` attribute from `VertexOrchestratorConfig` and `VertexStepOperatorConfig` and moved it to the `GoogleCredentialsMixinConfig`. Both, the orchestrator and step operator, were checking that the value of this `project` attribute was the same as the one obtained by the environment (gcloud default auth credentials), so I moved this logic to the base method `GoogleCredentialsMixin._get_default_authentication`.

I haven't forced the GCP Image Builder to be used with the GCP Container Registry, but if using this image builder with another container registry that is not the GCP Container Registry, then the registry has to be public, otherwise, the image cannot be pushed. I tried to find a way to authenticate to non-GCP container registries from Google Cloud Build, but as we're not saving the credentials of the container registries in their config classes, it has been impossible.

## Pre-requisites

## Types of changes


gabrielmbmb:
Where can I add the `GCPImageBuilder` so it appears in https://zenml.io/integrations?

src/zenml/utils/image_builder_context_uploader.py
```diff
@@ -0,0 +1,67 @@
...
+
+            filename = f"{hash_.hexdigest()}.tar.gz"
+            filepath = os.path.join(parent_path, filename)
+            if not fileio.exists(filepath):
+                logger.info("Uploading build context to `%s`.", filepath)
+                fileio.copy(f.name, filepath)
+            else:
+                logger.info("Build context already exists, not uploading.")
+
+        os.unlink(f.name)
```
gabrielmbmb, line 66: Not sure why, but if the `NamedTemporaryFile` is not deleted manually, it won't work in Windows (PermissionError).

AlexejPenner:
Glad you like the container image implementation and happy to see your PR

> Where can I add the `GCPImageBuilder` so it appears in https://zenml.io/integrations?

Don't worry about that, we will need to do that manually once this PR is merged.

src/zenml/utils/image_builder_context_uploader.py
AlexejPenner: We have run into exactly this issue at some other point before. I'm not too sure where, but I think we also solved it like this as well.

tests/unit/utils/test_image_builder_context_uploader.py
```diff
@@ -0,0 +1,38 @@
...
+def _get_build_context() -> BuildContext:
+    """Get a build context for testing.
+
+    Returns:
+        A build context.
+    """
+    return BuildContext(root=".", dockerignore_file=None)
+
+
+def test_upload_build_context() -> None:
```
schustmi, line 32: This should probably use the pytest `tmp_path` fixture as a temp directory

src/zenml/utils/image_builder_context_uploader.py
```diff
@@ -0,0 +1,67 @@
...
+            build_context.write_archive(f, gzip=True)
+
+            while True:
+                data = f.read(64 * 1024)
+                if not data:
+                    break
+                hash_.update(data)
+
+            filename = f"{hash_.hexdigest()}.tar.gz"
+            filepath = os.path.join(parent_path, filename)
```
schustmi, line 59: The `parent_path` in the image builders is created using a slash as path separator even on windows, but the `ImageBuilderContextUploader` here uses `os.path.join(...)` to add another path component to it. On windows that would lead to a path like `s3://my_bucket/kaniko\\build_context.tar.gz`. I think the image build components would probably fail in that case, which is why we should probably use explicit slashes here as well?

src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
```diff
@@ -0,0 +1,229 @@
...
+        Cloud Build can access it.
+
+        Returns:
+            Stack validator.
+        """
+
+        def _validate_remote_components(stack: "Stack") -> Tuple[bool, str]:
+            assert stack.container_registry
+
+            if stack.container_registry.config.is_local:
```
schustmi, line 70: I'm wondering if we need additional checks to make sure the container registry is also in GCP? Or is there some way to let cloud build know about the registry credentials to make this work if the registry is outside the GCP project?

docs/book/component-gallery/image-builders/gcloud-build.md
```diff
@@ -0,0 +1,43 @@
...
+The Google Cloud image builder is an [image builder](./image-builders.md) flavor provided
+with the ZenML `gcp` integration that uses [Google Cloud Build](https://cloud.google.com/build)
+to build container images.
+
+## When to use it
+
+You should use the Google Cloud image builder if:
+* you're **unable** to install or use [Docker](https://www.docker.com) on your client machine.
+* you're already using GCP.
+* your stack is conformed mainly by other Google Cloud components such as the [GCS Artifact Store](../artifact-stores/gcloud-gcs.md) or the [Vertex Orchestrator](../orchestrators/gcloud-vertexai.md).
```
schustmi, line 14:
```suggestion
* your stack is mainly composed of other Google Cloud components such as the [GCS Artifact Store](../artifact-stores/gcloud-gcs.md) or the [Vertex Orchestrator](../orchestrators/gcloud-vertexai.md).
```

docs/book/component-gallery/image-builders/gcloud-build.md
```diff
@@ -0,0 +1,43 @@
...
+
+## How to deploy it
+
+In order to use the ZenML Google Cloud image builder is enabling Google Cloud Build relevant APIs on the Google Cloud project.
```
schustmi, line 18:
```suggestion
In order to use the ZenML Google Cloud image builder you need to enable Google Cloud Build relevant APIs on the Google Cloud project.
```

schustmi (changes requested):
Another great PR, thanks @gabrielmbmb! Glad you like the image builder abstraction, the cloud build component will be a great addition to a hopefully growing list of implementations :)

I have some minor comments but overall looks really great already.
- The `tests/unit/integrations/gcp/image_builders/__init__.py` is still missing the license
- I like your idea of adding the `_upload_build_context()` method to the `BaseImageBuilder` class

src/zenml/utils/image_builder_context_uploader.py
gabrielmbmb: yes! I'd changed it because I got an error with the permissions on Windows, and I thought that it was because of the slashes.

src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
gabrielmbmb: My idea was that you could provide the authentication credentials when registering a container registry. Then, using these credentials and knowing the type of container registry, we could have some conditions to add an additional step to login to the container registry, so the image could be pushed from Google Cloud Build to external and private registries such as Github, ECR, etc.

Having that said, I think it's unlikely that anyone would use a stack in which the container registry is non-GCP and the image builder is GCP.

schustmi: Definitely agree, we'll be working on authentication credentials soon anyway after which we can hopefully lift this restriction and provide cloud build with access to any container registry registered in ZenML

schustmi:
Thanks for applying all the suggestions @gabrielmbmb! If we can move the `_upload_build_context` to the `BaseImageBuilder` class this is good to go IMO and can make it into the release on friday!

gabrielmbmb:
@schustmi All finished on my side!

schustmi:
Awesome, will merge once the tests pass 🎉 Thanks again for the amazing work, looking forward to use this!

PR #1271, opened 2023-01-30
Store run start time

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/pipelines/base_pipeline.py
+++ b/src/zenml/pipelines/base_pipeline.py
@@ -513,8 +513,8 @@ class BasePipeline(metaclass=BasePipelineMeta):
             # registering the schedule
             if schedule:
                 if not schedule.name:
-                    date = datetime.now().strftime("%Y_%m_%d")
-                    time = datetime.now().strftime("%H_%M_%S_%f")
+                    date = datetime.utcnow().strftime("%Y_%m_%d")
+                    time = datetime.utcnow().strftime("%H_%M_%S_%f")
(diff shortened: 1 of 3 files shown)


src/zenml/pipelines/base_pipeline.py
```diff
@@ -513,8 +513,8 @@ def run(
             # registering the schedule
             if schedule:
                 if not schedule.name:
-                    date = datetime.now().strftime("%Y_%m_%d")
-                    time = datetime.now().strftime("%H_%M_%S_%f")
+                    date = datetime.utcnow().strftime("%Y_%m_%d")
```
schustmi, line 516: @fa9r Not sure I agree with changing this one. This is displayed as-is in our database, somehow it would seem weird to show a date in the future/past there. The others are database entries which can be adjusted to the frontend time zone once we allow configuring something like that in the dashboard. wdyt?

fa9r: unless we want to change that back to local time as well

fa9r (approved):
Your changes look good, what about mine? :D

fa9r:
@schustmi unfortunately I can't request you to review your own PR ^^

src/zenml/pipelines/base_pipeline.py
schustmi: Ah I completely missed that, I thought this was a change in the `StepLauncher` and was arguing against changing that. But if that already happened then I guess this should be fine as well

PR #1273, opened 2023-01-30
Configure the zenml-server docker image and helm chart to run as non-privileged user

stefannica:
## Describe changes
The ZenML server deployed by our helm chart cannot currently run in Kubernetes clusters with more restrictive security policies that enforce the following rules:

* the containers cannot run as root (UID=0)
* the containerized processes are not allowed to listen on port 80, which is a system port value that only privileged processes are allowed to listen on

This PR adapts the zenml-server container image to configure a regular user and then updates the helm chart to run the containers with that user instead of root. It also switches the init container to use the same zenml-server container image instead of the base image to avoid having to update the base image, although the same restrictions are applicable to Kubernetes based orchestrators like Kubeflow, Tekton etc. 

## Pre-requisites

## Types of changes


docker/zenml-dev.Dockerfile
```diff
@@ -10,13 +10,10 @@ ENV PYTHONFAULTHANDLER=1 \
     ZENML_LOGGING_VERBOSITY=INFO \
     ZENML_CONTAINER=1
 
-
-WORKDIR /zenml
```
schustmi, line 14: Why is copying to `/` preferable to having a separate directory where the ZenML files are located?

docker/base.Dockerfile
```diff
@@ -14,15 +14,25 @@ ARG ZENML_VERSION
...
 
 WORKDIR /zenml
 
-RUN mkdir -p .zenconfig/local_stores/default_zen_store
-
 ENV ZENML_CONFIG_PATH=/zenml/.zenconfig \
     ZENML_DEBUG=true \
     ZENML_ANALYTICS_OPT_IN=false
 
+ARG USERNAME=zenml
```
schustmi, line 26: These changes are intentionally added only for the server part of the image, correct?

schustmi (commented):
Can't comment much on the terraform/helm changes apart from the fact that I like that the image/init_image distinction is gone :)

docker/base.Dockerfile
stefannica: Correct ! Adding them to the base image breaks a lot of things, especially when running local orchestrators, because the localhost user and the docker user may have different UIDs/GIDs. Tackling that side of the story is not something I wanted to do in this PR.

docker/zenml-dev.Dockerfile
stefannica: It's not, you're right. I went too far trying to make the dev image resemble the official base image, but the official base image doesn't copy any files, which is why it doesn't need a workdir.

stefannica: Done

schustmi (approved):

wjayesh (approved):
LGTM!

PR #1277, opened 2023-01-31
Rework params / artifact docs

strickvl:
Following user feedback, I added some extra context to two sections of our docs. I also shifted the location of the main params doc up a bit in the hierarchy as it seems to be more important than the others.

## Pre-requisites

## Types of changes


docs/book/starter-guide/pipelines/parameters-and-caching.md
```diff
@@ -4,24 +4,25 @@ description: Iteration is native to ZenML.
...
+this with `BaseSettings`.
+{% endhint %}
 
-You can parameterize a step by creating a subclass of the `BaseParameters`. When such a 
-config object is passed to a step, it is not treated like other artifacts. Instead, it 
-gets passed into the step when the pipeline is instantiated.
+You can parameterize a step by creating a subclass of the `BaseParameters`. When
+an object like this is passed to a step, it is not treated like other artifacts
+within ZenML. (Artifacts are the inputs and outputs of steps, but these
+parameters objects are handled differently.) Instead, it gets passed into the
```
fa9r, line 24: (...) feels a bit weird, perhaps we should just have a short section on artifacts somewhere and then simply link to it?

docs/book/starter-guide/pipelines/pipelines.md
```diff
@@ -47,6 +47,17 @@ As this step has multiple outputs, we need to use the
 These names can be used to directly access the outputs of steps after running
 a pipeline, as we will see [in a later chapter](./fetching-pipelines.md).
 
+{% hint style="info" %}
+The inputs and outputs of a step are *artifacts* that are automatically tracked
+and stored by ZenML in the artifact store. Artifacts are produced by steps
+whenever your step returns an object or a value. If a step returns only a single
```
fa9r, line 53: These 3 lines are basically the artifact section I mentioned above, perhaps just extract this into a section, put it somewhere where it makes sense (`pipelines.md` is not a bad place I guess) and then we can just link to it anytime "artifact" is mentioned instead of having to explain that term again

fa9r (changes requested):
Would prefer to move some of the info you added into a dedicated `### Artifacts` section as suggested in my comments below. I think that would make it easier to maintain in the future. Otherwise looks good 👍

docs/book/starter-guide/pipelines/parameters-and-caching.md
```diff
@@ -4,24 +4,25 @@ description: Iteration is native to ZenML.
...
 # Quickly iterating in ZenML
 
-Machine learning pipelines are rerun many times over throughout their development lifecycle. 
+Machine learning pipelines are rerun many times over throughout their development lifecycle.
 
 ## Parameterizing steps
 
-In order to iterate quickly, one must be able to quickly tweak pipeline runs by changing various 
-parameters for the steps that make up your pipeline.
+In order to iterate quickly, one must be able to quickly tweak pipeline runs by
```
stefannica, line 11: Somewhere in this section, you could show an example of a step that could benefit from parameterization, then show the parameterized version. For instance, you could have a step that uses some value hard-coded in the step implementation, that could be extracted as a parameter. 

I would also use the example to justify why parameterization is a good thing to have or generally why separation of code from configuration is a good practice.

docs/book/starter-guide/pipelines/parameters-and-caching.md
stefannica: I agree, this doesn't quite explain the difference between artifacts and parameters. I think it needs to be more clear in pointing out that artifacts are information circulated between steps during the pipeline execution, while the parameters are only used in the initial phase, to configure the steps, way before the pipeline is even launched.

strickvl:
@fa9r @stefannica reworked these following your feedback.

docs/book/starter-guide/pipelines/parameters-and-caching.md
```diff
@@ -19,10 +19,11 @@ this with `BaseSettings`.
...
 
 You can parameterize a step by creating a subclass of the `BaseParameters`. When
-an object like this is passed to a step, it is not treated like other artifacts
-within ZenML. (Artifacts are the inputs and outputs of steps, but these
-parameters objects are handled differently.) Instead, it gets passed into the
-step when the pipeline is instantiated.
+an object like this is passed to a step, it is not handled like other artifacts
+within ZenML. ([Visit this
+page](../../starter-guide/pipelines/pipelines.md#artifacts) to learn more about
+the difference between artifacts and parameters in ZenML.) Instead, it gets
```
fa9r, line 25:
```suggestion
an object like this is passed to a step, it is not handled like other [Artifacts](../../starter-guide/pipelines/pipelines.md#artifacts) within ZenML. Instead, it gets
```

fa9r (approved):

docs/book/starter-guide/pipelines/parameters-and-caching.md
```diff
@@ -71,7 +75,76 @@ is also supported as an attribute type in the `BaseParameters`.
...
 In essence, each pipeline can be viewed as an experiment, and each run is a trial of 
 the experiment, defined by the `BaseParameters`. You can always get the parameters again 
-when you [fetch pipeline runs](./fetching-pipelines.md), to compare various runs.
+when you [fetch pipeline runs](./fetching-pipelines.md), to compare various
+runs, and of course all this information is also available in the ZenML
+Dashboard.
+
+<details>
+<summary>How-To: Parameterization of a step</summary>
+A practical example of how you might parameterize a step is shown below. We can start with a version of a step that has yet to be parameterized. You can see how arguments for `gamma`, `C` and `kernel` are passed in and are hard-coded in the step definition.
```
stefannica, line 84: Beautiful ! Thank you for taking my suggestion !

stefannica (approved):

PR #1283, opened 2023-02-02
Update evidently to include reports and tests

wjayesh:
## Describe changes
I updated our evidently integration to v0.2.2 which replaces Profiles and Dashboard with the Report object and introduces the concept of TestSuites.

### Data Profiling
Profiling is now achieved through the use of the Report object.
It takes in a list of metrics to evaluate the data on and can be viewed as a JSON and an HTML object. This functionality is now reflected in the `evidently_report_step` which replaces the `evidently_profile_step` we had earlier.

### Data Validation
There's also a new step now that performs data validation using Evidently's TestSuites.
The `evidently_test_step` has a similar API to the report step but deals with generating test reports that indicate if the metrics pass/fail with respect to some threshold.

### Example
The example in its current form is only good for validating that the report and test additions function. I want to update it to make more sense and also add in some of the newer features like text profiling. Use the example only to get an idea of what the new flow will look like.

### Documentation
No documentation on the newer features exist and will be added soon.

## Pre-requisites

## Types of changes


src/zenml/integrations/evidently/data_validators/evidently_data_validator.py
```diff
@@ -52,60 +49,132 @@
...
+    #      ...
+    # }
     try:
-        return (
-            [profile_mapper[profile]() for profile in profile_list],
-            [
-                dashboard_mapper[profile](verbose_level=verbose_level)
-                for profile in profile_list
-            ],
+        # if metric contains preset in name, return an object of the class of same name from metric_preset
```
wjayesh, line 76: I need advice here. Is this the right way to handle loading the classes from the string values? Earlier, we used a map to link the strings and their respective classes.

src/zenml/integrations/evidently/data_validators/evidently_data_validator.py
```diff
@@ -52,60 +49,132 @@
...
+        metric: Name of the metric or metric preset.
 
     Returns:
-        A tuple of two lists of profile sections and tabs.
+        The Evidently metric or metric preset class.
 
     Raises:
-        ValueError: if the profile_section is not supported.
+        ValueError: If the metric or metric preset is not a valid Evidently metric or metric preset.
     """
```
wjayesh, line 63: I need advice here. Is this the right way to handle loading the classes from the string values? Earlier, we used a map to link the strings and their respective classes.

src/zenml/integrations/evidently/data_validators/evidently_data_validator.py
```diff
@@ -52,60 +49,132 @@
...
-    profile_list = profile_list or list(profile_mapper.keys())
+    # metric_mapping = {
```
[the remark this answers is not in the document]
htahir1: I think it makes sense to me, and is better than hard-coded list

htahir1 (commented):
I have not checked this thoroughly but my big question is what happens to all current usages of evidently with ZenML? How are you handling deprecation here?

wjayesh:
@htahir1 I have added support for the existing `evidently_profile_step` so folks using that should not be affected.

htahir1 (changes requested):
Please also update quickstart and other examples in PR's in ZenML projects and ZenBytes as this is one of the most widely used integratins we have to careful

PR #1286, opened 2023-02-02
Always add active user to analytics

stefannica:
## Describe changes
Always add active user to analytics

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/analytics_utils.py
+++ b/src/zenml/utils/analytics_utils.py
@@ -335,6 +335,8 @@ class AnalyticsContext:
             zen_store = gc.zen_store
             user = zen_store.get_user()
 
+            if "client_id" not in properties:
+                properties["client_id"] = self.user_id
             if "user_id" not in properties:
                 properties["user_id"] = str(user.id)
 
@@ -555,7 +557,6 @@ def track(
             Result of the function.
         """
         with event_handler(event) as handler:
-
             try:
                 if len(args) and isinstance(args[0], AnalyticsTrackerMixin):
                     handler.tracker = args[0]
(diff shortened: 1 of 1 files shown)


schustmi (commented):
Not crucial for this PR as it needs to make it into the release, but for such an important part we should probably have a unit test that ensures this is always included

src/zenml/utils/analytics_utils.py
```diff
@@ -333,11 +333,15 @@ def track(
         # infinite loop
         if gc._zen_store is not None:
             zen_store = gc.zen_store
+            user = zen_store.get_user()
```
schustmi, line 336: I haven't looked at this code in a while, so excuse the stupid question: This will now be (in most local cases) the ID of the default user of the local store, right? So deleting this store will give me a new user ID, but my client ID (which is also always tracked) will stay the same?

stefannica: Yes, that's exactly right !

htahir1 (approved):

src/zenml/utils/analytics_utils.py
```diff
@@ -333,11 +333,15 @@ def track(
...
+            if "user_id" not in properties:
+                properties["user_id"] = str(user.id)
+
             if (
                 zen_store.type == StoreType.REST
                 and "server_id" not in properties
             ):
-                user = zen_store.get_user()
                 server_info = zen_store.get_store_info()
                 properties.update(
```
htahir1, line 346: Can we have `client_id: self.user_id` in here please (after the discussion today)

stefannica: Done

stefannica: @schustmi yes, he just wants it duplicated as an event metadata field, to help with grouping and filtering (I think).

htahir1 (approved):
Great now im happy to merge!

PR #1296, opened 2023-02-07
increase timeout on quickstart example

safoinme:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/examples/quickstart/steps/model_deployers.py
+++ b/examples/quickstart/steps/model_deployers.py
@@ -13,5 +13,12 @@
 #  permissions and limitations under the License.
 
 from zenml.integrations.mlflow.steps import mlflow_model_deployer_step
+from zenml.integrations.mlflow.steps.mlflow_deployer import (
+    MLFlowDeployerParameters,
+)
 
-model_deployer = mlflow_model_deployer_step()
+model_deployer = mlflow_model_deployer_step(
+    params=MLFlowDeployerParameters(
+        timeout=60,
+    )
+)


htahir1 (approved):

examples/quickstart/steps/model_deployers.py
```diff
@@ -13,5 +13,12 @@
 #  permissions and limitations under the License.
 
 from zenml.integrations.mlflow.steps import mlflow_model_deployer_step
+from zenml.integrations.mlflow.steps.mlflow_deployer import (
+    MLFlowDeployerParameters,
+)
 
-model_deployer = mlflow_model_deployer_step()
+model_deployer = mlflow_model_deployer_step(
```
schustmi, line 20: Why don't we change the default value on the params instead, won't this fail for anyone using this step?

schustmi (changes requested):

examples/quickstart/steps/model_deployers.py
safoinme: Because if we change that we will need to make a new release, we have already done that on a different PR https://github.com/zenml-io/zenml/pull/1294

fa9r (approved):
Great fix that can be backported to 0.33.0

examples/quickstart/steps/model_deployers.py
schustmi: Should probably create a ticket to revert the change after the next release then right?

schustmi: + the mlflow deployment example might need the same change?

safoinme: Yes, that's actually true, will create a ticket to revert this back. I think yes MLFlow example may need a similar fix

PR #1299, opened 2023-02-07
Add warning and docs update for `label_studio` installation issue

strickvl:
Added a warning post-install, and a docs update that reflects the same message. These are interim fixes until we figure out the right version combination that works without breaking ZenML's Pydantic version.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-gallery/annotators/label-studio.md
+++ b/docs/book/component-gallery/annotators/label-studio.md
@@ -48,6 +48,14 @@ and add it to your stack:
 zenml integration install label_studio
 ```
 
+{% hint style="warning" %}
+There is a known issue with Label Studio installations via `zenml integration
+install...`. You might find that the Label Studio installation breaks the ZenML
+CLI. In this case, please run `pip install 'pydantic==1.10.4'` to fix the issue
+or [message us on Slack](https://zenml.io/slack-invite) if you need more help with
+this. We are working on a more definitive fix.
+{% endhint %}
+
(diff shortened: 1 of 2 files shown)


docs/book/component-gallery/annotators/label-studio.md
```diff
@@ -48,6 +48,14 @@ and add it to your stack:
 zenml integration install label_studio
 ```
 
+{% hint style="warning" %}
+There is a known issue with Label Studio installations via `zenml integration
+install...`. You might find that the Label Studio installation breaks the ZenML
+CLI. In this case, please run `pip install 'pydantic==1.10.4'` to fix the issue
```
schustmi, line 54: Can we maybe include the full range of supported pydantic versions in this instruction as well? We don't actually need 1.10.4 for this to work I assume

src/zenml/cli/integration.py
```diff
@@ -253,6 +253,10 @@ def install(
     ):
         with console.status("Installing integrations..."):
             install_packages(requirements)
+            if "label_studio" in integrations:
+                warning(
+                    "There is a known issue with Label Studio installations via zenml. You might find that the Label Studio installation breaks the ZenML CLI. In this case, please run `pip install 'pydantic==1.10.4'` to fix the issue or message us on Slack if you need help with this. We are working on a more definitive fix."
```
schustmi, line 258: Same as above

schustmi (changes requested):

PR #1308, opened 2023-02-08
Increase timeout for mlflow deployment example

strickvl:
#1294 was not enough to fix this particular example, since the timeout values are hardcoded in the example. I updated them as per the main change we made yesterday.

## Pre-requisites
Please ensure you have done the following:
- [ ] I have read the **CONTRIBUTING.md** document.
- [ ] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #1310, opened 2023-02-09
Update step name assignment with the parameter name

strickvl:
Possibly I went too far in changing the assignment of `step_name` inside the `mlflow_deployer` step, but otherwise this PR addresses the ticket relating to which version of the two 'step names' (i.e. what the parameter is called vs what the actual (step) function that's passed in is called).

Potentially I could add a test for this, but if it's changing anyway soon with @schustmi's pipeline work then maybe not needed?

## Pre-requisites

## Types of changes


src/zenml/integrations/mlflow/steps/mlflow_deployer.py
```diff
@@ -102,7 +102,7 @@ def mlflow_model_deployer_step(
     step_env = cast(StepEnvironment, Environment()[STEP_ENVIRONMENT_NAME])
     pipeline_name = step_env.pipeline_name
     run_name = step_env.run_name
-    step_name = step_env.step_name
+    step_name = step_env.step_run_info.pipeline_step_name
```
strickvl, line 105: @schustmi I think this change was probably too zealous. It'd be a breaking change for anyone using our `mlflow_deployer`, and in any case we don't need to change the name here, right?

schustmi: I'm guessing because it won't find the previous services deployed with the same step?

Was the reason for this name change to make it more consistent/easier to reach with the post execution workflow?

examples/mlflow_deployment/README.md
```diff
@@ -142,7 +142,7 @@ def predictor(
 ) -> Output(predictions=np.ndarray):
     """Run a inference request against a prediction service"""
 
-    service.start(timeout=10)  # should be a NOP if already started
+    service.start(timeout=60)  # should be a NOP if already started
```
schustmi, line 145: I'm wondering why we even need to supply a custom timeout value here, doesn't that indicate that our default value is somehow wrong?

schustmi (changes requested):

examples/mlflow_deployment/README.md
strickvl: We've already made this change in a separate PR (#1294). I just needed it in here as well for the example to work while working on the code.

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
strickvl: Yes. That was [the ticket](https://zenml.atlassian.net/jira/software/projects/FRW/boards/5?assignee=61710083892c4200720ca56d&selectedIssue=FRW-1654) which more or less made sense to me, though tbh I'm not sure which of those values make for a better step name for the post-execution workflow. Your argument this morning that you'd want it to be the param name value seemed to make sense, but from what I understood of the ticket the need was for this value to be the one used in post-execution as well.

@stefannica maybe you can clarify? I don't have a strong opinion one way or another which of the two should be used.

examples/mlflow_deployment/README.md
strickvl: #1308 is just waiting for a review before it gets merged in. Same changes here...

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
stefannica: This is definitely what's needed to fix the bug: the previous step name is worthless, because it doesn't give me any lineage information I can use in the post-execution workflow. This new step name value is what should be "stamped" on the served model.

I would even go one step further and change `StepEnvironment.step_name` to be set to `step_run_info.pipeline_step_name` instead of what's currently being set to. The Seldon and KServe model deployers suffer from the same problem, and instead of fixing them individually, you could just correct the `step_name` value reported by the step environment. The justification is the same: the previous `StepEnvironment.step_name` value is worthless, because I can't use it anywhere in the post-execution workflow.

stefannica (approved):

PR #1312, opened 2023-02-09
Add user environment `zenml info` command to CLI for debugging

strickvl:
ScreenShot 2023-02-09 at 16 46 49

I added a (hidden) CLI command to get information about the user's setup to help with Slack queries and other remote debugging scenarios.

`zenml info` is the base command, but add:

- the `-a` flag to get version information and names of all locally-installed packages
- the `-f` or `--file` option where you can pass in a path where you want the information to be written to.
- the `-p [PACKAGE_NAME]` option (takes multiple, e.g. `-p numpy -p mlflow`) to get pypi package version information about specific packages i.e. instead of getting *all* packages which can be a bit unwieldy.

## Pre-requisites

## Types of changes


src/zenml/cli/utils.py
```diff
@@ -1230,3 +1230,42 @@ def wrapper(function: F) -> F:
...
+
+    Returns:
+        A dictionary of all locally installed packages and their versions.
+    """
+    import pkg_resources
+
+    return {pkg.key: pkg.version for pkg in pkg_resources.working_set}
+
+
+def get_package_information(package_names: List[str]) -> Dict[str, str]:
```
schustmi, line 1246: Tiny nit, but this and the function above are very similar and we could probably unify them? E.g. make the parameter an optional `include` list, and if given only those packages get returned. If left empty, all packages get returned

src/zenml/cli/base.py
```diff
@@ -433,3 +437,80 @@ def _prompt_email(event_source: AnalyticsEventSource) -> bool:
...
+@click.option(
+    "--all",
+    "-a",
+    is_flag=True,
+    default=False,
+    help="Output information about all installed packages.",
+    type=bool,
+)
+@click.option(
+    "--file",
```
schustmi, line 461: Should we let users pass a string here for the actual file path?

src/zenml/cli/base.py
```diff
@@ -433,3 +437,80 @@ def _prompt_email(event_source: AnalyticsEventSource) -> bool:
...
+    "-f",
+    is_flag=True,
+    help="Output to a file.",
+    type=bool,
+)
+def info(packages, all, file) -> None:
```
schustmi, line 467:
```suggestion
def info(packages: Tuple[str], all: bool = False, file: bool = False) -> None:
```

I hope this is correct, might also need some imports

src/zenml/cli/base.py
```diff
@@ -433,3 +437,80 @@ def _prompt_email(event_source: AnalyticsEventSource) -> bool:
...
+        "telemetry_status": "enabled" if gc.analytics_opt_in else "disabled",
+        "analytics_client_id": str(gc.user_id),
+        "analytics_user_id": str(client.active_user.id),
+        "analytics_server_id": str(client.zen_store.get_store_info().id),
+        "integrations": integration_registry.get_installed_integrations(),
+    }
+
+    if all:
+        user_info["packages"] = cli_utils.get_installed_local_packages()
+    if packages:
```
schustmi, line 506: Could this be an `elif`? If all packages are already included above, I probably don't need this extra subset here?

schustmi (changes requested):

src/zenml/cli/base.py
strickvl: There are so many packages. I can imagine wanting to have them all, but also wanting to have a special place just to see the subset I care about right now. At least that was what I was going for here.

src/zenml/cli/utils.py
strickvl: Done

src/zenml/cli/base.py
strickvl: Refactored to allow this. I was in two minds about it, since it will generally always be us who are asking users to use this command (given that it is hidden) but I think it can't hurt.

schustmi (approved):

src/zenml/cli/base.py
```diff
@@ -433,3 +437,89 @@ def _prompt_email(event_source: AnalyticsEventSource) -> bool:
...
+        "python_version": environment.python_version(),
+        "environment": get_environment(),
+        "system_info": environment.get_system_info(),
+        "active_workspace": client.active_workspace.name,
+        "active_stack": client.active_stack_model.name,
+        "active_user": client.active_user.name,
+        "telemetry_status": "enabled" if gc.analytics_opt_in else "disabled",
+        "analytics_client_id": str(gc.user_id),
+        "analytics_user_id": str(client.active_user.id),
+        "analytics_server_id": str(client.zen_store.get_store_info().id),
```
stefannica, line 500: I know @htahir1 would love this

stefannica (approved):

htahir1 (changes requested):
This is good but be aware that that the `-p` flag might become redundant when Hub is introduced. Also, could you add changes to the GitHub bug report template (in the .github folder) to ask people to do this before submitting a bug? And maybe add this to the debugging guide in the docs

wjayesh (approved):

schustmi (approved):

PR #1313, opened 2023-02-09
Secrets store with SQL back-end

stefannica:
## Describe changes
Implements *the Secrets Store* as an alternative to using Secrets Manager stack components to store and manage secrets in a centralized service or database that doesn't require an active stack.

### Design considerations

The Secrets Store is an optional runtime extension to the Zen Store:

* it is an extension because it's not an integral part of the Zen Store interface. The Secrets Store interface is defined separately and Secrets Store implementations only need to implement the Secrets Store interface rather than the entire Zen Store API.
* it is a *runtime* extension because the Zen Store can be dynamically extended at runtime with a Secrets Store according to how it is configured
* it is optional because ZenML can even be configured to not implement a Secrets Store

This modular design allows us to support multiple implementations of Secrets Store and to decide at runtime through configuration means which one to use:
* the REST Secrets Store (included in this PR) is implemented to use the REST Zen Store and its ZenML server HTTP connection as a back-end
* the SQL Secrets Store (included in this PR) is implemented to use the SQL Zen Store and its SQL database as a back-end
* other Secrets Stores (not included) will use the existing Secrets Manager implementations and should, in theory, be usable with any of the Zen Store flavors, although it only makes sense to use them in combination with the SQL Zen Store.

I considered various design patterns to support this, but there was one that was clearly superior in terms of ease of implementation and long term maintenance: the Proxy, or a slight variation of it. The ZenStore singleton instance acts as a Proxy or a Facade, meaning that it exposes a single API that includes its existing concerns (CRUD for Stacks, Stack Components, Pipelines etc.) as well as the Secrets Store concerns (CRUD for Secrets). I.e. the consumer code can just run `zen_store.create_secret(...)` without having to be aware that the Secrets Store is an optional extension. The ZenStore is responsible for proxying/relaying those calls to the Secrets Store instance configured at runtime, or raising a `NotImplementedError` exception if a Secrets Store wasn't configured.

Here's how this is implemented:
* the BaseZenStore class inherits the Secrets Store abstract interface, so the exposed API includes secrets management
* the BaseZenStore also has a BaseSecretsStore field - the object to which secrets management calls will be proxied
* finally, the BaseZenStore implements the Secrets Store abstract interface by simply relaying all calls to the Secrets Store

To avoid writing a lot of boilerplate code required to proxy all secrets management method calls in the BaseZenStore, I wrote a reusable class decorator that generates those methods dynamically.

### What's in the box ?

Included in this PR:

* Secrets (domain) Models
* Secrets Store interface definition - slightly resembles the Base Secrets Manager abstraction
* Base Secrets Store and Secrets Store configuration classes 
* SQL Secrets Store implementation - built as an extension of the SQL Zen Store (in fact, it uses some of it's attributes and methods). Optionally encrypts and stores secrets in a new table in the SQL database.
* REST Secrets Store implementation - built as an extension of the REST Zen Store (in fact, it uses some of it's attributes and methods)
* ZenML Client API for centralized secrets management
* ZenML CLI for centralized secrets management
* Changes to existing secrets management consumer code (i.e. secret reference resolution) to first attempt to resolve secret references through the Secrets Store and then fall back to the current Secrets Manager stack components.

### Secrets Scoping
Secrets Managers currently support different scopes for managing secrets. The Secrets Store being a global entity, it also implements secrets scoping and namespacing but it uses the existing ZenML concepts to that purpose:

* secrets are by default scoped to a Workspace
* in addition to that, secrets can also be scoped to a User
* resolving a secret name is implicitly done first at User level, then at Workspace level
* there is currently no enforcing of secrets visibility server-side through RBAC or authorization, just like it is currently done with Secrets Managers: any user can access any secret in any scope. Scoping is about managing different secrets namespaces.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/secrets_stores/sql_secrets_store.py
```diff
@@ -0,0 +1,438 @@
...
+            KeyError: if the secret doesn't exist.
+        """
+        with Session(self.engine) as session:
+            try:
+                secret_in_db = session.exec(
+                    select(SecretSchema).where(SecretSchema.id == secret_id)
+                ).one()
+                session.delete(secret_in_db)
+            except NoResultFound as error:
+                raise KeyError from error
```
schustmi, line 434: I think the error message isn't really nice when doing this, it'll just be an empty KeyError right

src/zenml/zen_stores/secrets_stores/sql_secrets_store.py
```diff
@@ -0,0 +1,438 @@
...
+
+    # =======================
```
schustmi, line 436: Not needed?

src/zenml/client.py
```diff
@@ -3140,6 +3147,331 @@ def list_run_metadata(
...
+
+        Args:
+            name_id_or_prefix: The name or ID of the secret.
+            scope: The scope of the secret to delete.
+        """
+        secret = self.get_secret(
+            name_id_or_prefix=name_id_or_prefix,
+            scope=scope,
+            # Don't allow partial name matches, but allow partial ID matches
+            allow_partial_name_match=False,
```
schustmi, line 3468: Is this something that you think makes sense for other delete methods as well, or is it more important in the case of secrets to avoid accidently deleting something?

src/zenml/client.py
```diff
@@ -3140,6 +3147,331 @@ def list_run_metadata(
...
+            allow_partial_id_match=True,
+        )
+
+        secret = self.get_secret(name_id_or_prefix=name_id_or_prefix)
```
schustmi, line 3472: This line should probably go, currently it just overrides the logic above?

src/zenml/client.py
```diff
@@ -3140,6 +3147,331 @@ def list_run_metadata(
...
+        if new_scope:
+            secret_update.scope = new_scope
+        values: Dict[str, Optional[str]] = {}
+        if add_or_update_values:
+            values.update(add_or_update_values)
+        if remove_values:
+            for key in remove_values:
+                values[key] = None
+        if values:
+            secret_update.values = cast(Dict[str, Optional[SecretStr]], values)
```
schustmi, line 3449: This cast is incorrect right, it's not actually a dict containing secret strings? Do we run into some trouble of calling `SecretStr.some_method()` somewhere down the line by doing this?

schustmi (changes requested):
Great PR, I especially love the proxy decorator :) 
Only some minor comments regarding the implementation from my side, however I think there are quite a few areas that could benefit from additional unit tests:
- The secret resolving order of first using the secrets store and then falling back to the secrets manager
- The secrets stores themselves
- Configuring the secrets store config using env variables
- proxy utils

src/zenml/client.py
stefannica: Yes, this is something that I think may be confusing in a lot of other places. Prefix matching is fine when you retrieve information with get/list, but in the case of update/delete operations, I feel like prefix matching on names should be turned off, otherwise you may end up updating/deleting the wrong object. This is how Docker does it, for reference. I have a ticket already filed for this.

src/zenml/client.py
stefannica: Oops, good catch, thank you !

src/zenml/zen_stores/migrations/versions/03742aa7fdd7_add_secrets.py
```diff
@@ -0,0 +1,51 @@
...
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    op.create_table(
+        "secret",
+        sa.Column("values", sa.TEXT(), nullable=False),
+        sa.Column(
+            "workspace_id", sqlmodel.sql.sqltypes.GUID(), nullable=False
+        ),
+        sa.Column("user_id", sqlmodel.sql.sqltypes.GUID(), nullable=False),
```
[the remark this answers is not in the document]
stefannica: Secrets can be user-scoped. Leaving those around when a user is deleted is not useful for anything. Particularly for secrets, you don't want sensitive data associated with your account to be left behind after you delete your account.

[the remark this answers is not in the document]
stefannica: It may make sense to leave the user nullable and set the user to None on creation for workspace scoped secrets, but that is something that breaks a lot of assumptions in the ZenML model, so I'd rather leave it like this until someone actually complains that this is a limitation.

src/zenml/zen_stores/secrets_stores/sql_secrets_store.py
```diff
@@ -0,0 +1,438 @@
...
+
+        Raises:
+            KeyError: if the secret doesn't exist.
+        """
+        with Session(self.engine) as session:
+            try:
+                secret_in_db = session.exec(
+                    select(SecretSchema).where(SecretSchema.id == secret_id)
+                ).one()
+                session.delete(secret_in_db)
```
strickvl, line 432: I think we're missing a `session.commit()` at some point after this line.

src/zenml/models/page_model.py
```diff
@@ -94,3 +95,12 @@ def __contains__(self, item: B) -> bool:
             Whether the item is in the page.
         """
         return item in self.items
+
+    class Config:
+        """Pydantic configuration class."""
+
+        # This is needed to allow the REST API server to unpack SecretStr
+        # values correctly before sending them to the client.
+        json_encoders = {
```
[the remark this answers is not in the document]
stefannica: I grouped those under base_models.py into a BaseZenModel class, but the page model is too divergent, I'll leave it as it is.

src/zenml/client.py
stefannica: Quite right, yes. You found another bug that likely would've been revealed by the integration tests I need to write :)

src/zenml/zen_stores/secrets_stores/sql_secrets_store.py
stefannica: HA ! I copied this from `delete_flavors`, where we're also missing a commit :)

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -801,6 +803,55 @@ def create_run_metadata(
...
+            f"of this endpoint `{workspace_name_or_id}` is "
+            f"not supported."
+        )
+    if secret.user != auth_context.user.id:
+        raise IllegalOperationError(
+            "Creating pipeline runs for a user other than yourself "
+            "is not supported."
+        )
+    return zen_store().create_secret(secret=secret)
+
```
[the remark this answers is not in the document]
schustmi: I agree, but I'm not sure if those endpoints are maybe something the dashboard is calling and is therefore necessary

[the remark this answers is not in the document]
schustmi: @AlexejPenner Do you know which endpoints they're calling?

schustmi (approved):
Can't wait to see all the features that we're going to build on top of this!

stefannica:
Should be good to go when the CI passes.

docs/book/advanced-guide/practical/secrets-management.md
```diff
@@ -5,11 +5,144 @@ description: How to register and use secrets
...
+[Secrets Managers](../../component-gallery/secrets-managers/secrets-managers.md)
+are ZenML stack components that allow you to register and access secrets when
+used as part of your active stack.
+
+{% hint style="warning" %}
+We are slowly deprecating secrets managers in favor of the [centralized ZenML secrets store](#centralized-secrets-store).
+Going forward, we recommend using the secrets store instead of secrets managers
+to configure and store secrets.
+
+Managing secrets through a secrets manager stack components suffers from a
```
strickvl, line 127:
```suggestion
Managing secrets through a secrets manager stack component suffers from a
```

docs/book/advanced-guide/practical/secrets-management.md
```diff
@@ -105,13 +245,48 @@ their values.
...
 well as the existence of the exact key-value pair.
 
+
+{% hint style="warning" %}
+If you have secrets registered through both the [centralized secrets management](#centralized-secrets-store)
+and [a secrets manager](#secrets-management-with-secrets-managers), ZenML will
+first try to fetch the secret from the centralized secrets management and only
+fall back to the secrets manager if the secret is not found. This means that if
+you have a secret registered with the same name in both centralized secrets
+store
```
strickvl, line 255: I think this thought needs completing?

docs/book/advanced-guide/practical/secrets-management.md
```diff
@@ -105,13 +245,48 @@ their values.
...
+        username = secret.secret_values["username"],
+        password = secret.secret_values["password"],
+    )
+    ...
+```
+
+If you are using a Secrets Manager to manage secrets, you can access the secrets
+manager directly from within your steps through the `StepContext`. This allows
+you to use your secrets for querying APIs from  within your step without
+hard-coding your access keys. Don't forget to  make the appropriate decision
```
strickvl, line 287:
```suggestion
hard-coding your access keys. Don't forget to make the appropriate decision
```

docs/book/component-gallery/secrets-managers/secrets-managers.md
```diff
@@ -9,6 +9,35 @@ those services. Instead of storing these credentials in code or files,
 ZenML secrets managers can be used to store and retrieve these values
 in a secure manner.
 
+{% hint style="warning" %}
+We are slowly deprecating secrets managers in favor of the
+[centralized ZenML secrets store](../../advanced-guide/practical/secrets-management.md#centralized-secrets-store).
+Going forward, we recommend using the secrets store instead of secrets managers
+to configure and store secrets.
+
+Managing secrets through a secrets manager stack components suffers from a
```
strickvl, line 18:
```suggestion
Managing secrets through a secrets manager stack component suffers from a
```

docs/book/component-gallery/secrets-managers/secrets-managers.md
```diff
@@ -9,6 +9,35 @@ those services. Instead of storing these credentials in code or files,
...
+number of limitations, some of which are:
+
+* you need to configure Secrets Manager stack component
```
strickvl, line 21:
```suggestion
* you need to configure a Secrets Manager stack component
```

docs/book/component-gallery/secrets-managers/secrets-managers.md
```diff
@@ -9,6 +9,35 @@ those services. Instead of storing these credentials in code or files,
...
+and add it to your active stack before you can register and access secrets. With
+centralized secrets management, you don't need to configure anything, your ZenML
```
strickvl, line 23:
```suggestion
centralized secrets management, you don't need to configure anything; your ZenML
```

docs/book/advanced-guide/practical/secrets-management.md
stefannica: Thanks for noticing, this got away from me.

stefannica:
Thanks for the last-minute docs review @strickvl ! Should be good now.

PR #1314, opened 2023-02-09
Add warning when attempting server connection without user permissions

strickvl:
Added a warning that catches the `IllegalOperationError` that we previously handed to the user without context/explanation.

Following discussion with @AlexejPenner, we agreed that there aren't really any scenarios we could imagine where a user might want to be logged into a ZenML server without any permissions attached to their role.

## Pre-requisites

## Types of changes


schustmi:
Is this something that we can add a test for or is that unreasonably complicated?

AlexejPenner:
> Is this something that we can add a test for or is that unreasonably complicated?

I don't think the test harness currently allows for attempting to connect to a new deployed zenml server. It also feels like unproportional effort to create this scenario for a niche exception like this one.

schustmi:
What I had in mind, but might not be possible:
When the test are running with a rest zen server, create a new user without role. Then run CLI `connect` command to connect to that already connected server (or disconnect first if necessary) which should trigger this? But I now realize we only log a warning which isn't testable anyway I guess

strickvl:
> What I had in mind, but might not be possible: When the test are running with a rest zen server, create a new user without role. Then run CLI `connect` command to connect to that already connected server (or disconnect first if necessary) which should trigger this? But I now realize we only log a warning which isn't testable anyway I guess

Technically we could test that the `IllegalOperationError` doesn't get raised...

schustmi:
Probably not worth it though, maybe if it works after 5 mins it's a nice test but not that important to have I guess

strickvl:
@AlexejPenner @schustmi added a test. Not sure if it'll work, but let's see.

src/zenml/cli/server.py
```diff
@@ -785,7 +786,15 @@ def connect(
...
-    GlobalConfiguration().set_store(store_config)
+    try:
+        GlobalConfiguration().set_store(store_config)
+    except IllegalOperationError:
+        cli_utils.warning(
+            f"User '{username}' has no permissions attached to its role with "
+            f"which to access the server at '{url}'. Please ask the server "
+            f"administrator to assign a role with permissions to your "
+            f"username."
+        )
```
stefannica, line 797:
```suggestion
    except IllegalOperationError as e:
        cli_utils.warning(
            f"User '{username}' does not have sufficient permissions to "
            f"to access the server at '{url}'. Please ask the server "
            f"administrator to assign a role with permissions to your "
            f"username: {str(e)}"
        )
```

Or something similar. The reason being that the IllegalOperationError can be thrown in cases where the user only has read permissions or doesn't have permissions to modify the current project, or other situations such as these, so the message should be more generic and also include the original error.

stefannica (changes requested):

tests/integration/functional/cli/test_server.py
```diff
@@ -0,0 +1,64 @@
...
+    with does_not_raise():
+        result = runner.invoke(
+            cli.commands["connect"],
+            [
+                f"--url={server_url}",
+                f"--user={new_user_name}",
+                f"--password={new_password}",
+            ],
+        )
+        assert result.exit_code == 0
```
stefannica, line 64: Forgot to mention this the last time, but could you please delete the user at the end of the test, to not clutter the external servers with left-over users ?

stefannica (approved):

PR #1323, opened 2023-02-14
GCP Image Builder network 

gabrielmbmb:
## Describe changes

I encountered an authentication problem while trying to build a pipeline Docker image that uses a dependency from a private GCP Artifact Registry using the GCP Image Builder.

As described in this [Google Cloud documentation page](https://cloud.google.com/build/docs/build-config-file-schema#network) about Google Cloud Build, when cloud build runs a step the container gets attached to a network called `cloudbuild` that can be used to get the Application Default Credentials.

In order to install a dependency from a private GCP Artifact Registry, the build container needs to be attached to this network using `--network=cloudbuild` option. This way, the container will be automatically authenticated and the dependency can be installed from the private GCP Artifact Registry. For it to work, you will also need to use a custom base `Dockerimage` like this one:

```dockerfile
FROM zenmldocker/zenml:0.33.0-py3.10

RUN pip install keyrings.google-artifactregistry-auth
```

This PR adds the `GCPImageBuilder.network` attribute that can be used to pass the name of the network to which the build container has to be attached. By default, I gave it the value of `cloudbuild` as it's likely that someone building a Dockerimage using the Google Cloud Build will need to install a dependency from a private GCP Artifact Registry.

Let me know if the instructions described above need to be included in the docs!

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-gallery/image-builders/gcp.md
+++ b/docs/book/component-gallery/image-builders/gcp.md
@@ -41,3 +41,19 @@ zenml image-builder register <IMAGE_BUILDER_NAME> \
 # Register and activate a stack with the new image builder
 zenml stack register <STACK_NAME> -i <IMAGE_BUILDER_NAME> ... --set
 ```
+
+## Caveats
+
+As described in this [Google Cloud Build documentation page](https://cloud.google.com/build/docs/build-config-file-schema#network), Google Cloud Build uses containers to execute the build steps which are automatically attached to a network called `cloudbuild` that provides some Application Default Credentials (ADC), that allows the container to be authenticated and therefore use other GCP services.
+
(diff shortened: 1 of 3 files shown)


schustmi (commented):
Code looks good to me! I don't think this needs an explicit mention in the docs, but maybe in the docstring of the `network` attribute we could add the link to the google docs page that you also linked in the PR?

gabrielmbmb:
What about the base `Dockerfile` that is needed with the `keyrings.google-artifactregistry-auth` dependency? Should we mention it in the GCP Image Builder documentation page as a caveat?

schustmi:
@gabrielmbmb Good point, seems like that isn't mentioned anywhere on their page and would be helpful for our users

gabrielmbmb:
@schustmi I've added it to the GCP Image Builder documentation page

schustmi (approved):
Thanks so much, will merge once the checks pass!

PR #1328, opened 2023-02-16
Add individual stack component deployment through recipes

wjayesh:
## Help Needed

I want to figure out a nice and clean way to make users aware of what flavors can be chosen for each stack component, when using the CLI.

I also want to communicate what configuration options can be set for each stack component flavor.

More details on what can be done is in [`docs/book/component-gallery/deploying-components.md`](https://github.com/zenml-io/zenml/blob/935bc8ee2e109c8cea5839c7ca5a148de3de7741/docs/book/component-gallery/deploying%20components.md).

## Describe changes
I implemented new functionality to allow individual stack components to be deployed using the stack recipe CLI and introduced a new CLI for stack component deployment.

Two changes
- The `zenml stack recipe` CLI now supports conditional deployment and destruction of individual stack components.
- Every stack component can now be deployed using the `deploy` subcommand in the stack-component CLI.

## Workflow for Recipes

In order to deploy any component, first choose the recipe corresponding to your cloud provider (let's say gcp this case)

### Deployment 

Run the deploy command with options for each stack component you want to deploy. You can choose either one or a series of 
components at a time.

- Deploy one component
  ```
  zenml stack recipe deploy gcp-modular -o "kubernetes"
  ```
  This would create a Kubernetes cluster on GCP and return the outputs that can be used to register a Kubernetes orchestrator. 

- Deploy more components: If you wish to create more stack components, you can re-run the deploy command and that would add new components on top of already existing ones. For example, with the command below, you can get an mlflow experiment tracker and a seldon model deployer in addtion to the kubernetes orchestrator we created earlier.

  ```
  zenml stack recipe deploy gcp-modular -e "mlflow" -d "seldon"
  ```
  This would then output the right values for both these components.

- You can pass a config file along with the deploy command now to include any variables you might want to set for the recipe.
  
  ```
  zenml stack recipe deploy gcp-modular -e "mlflow" --config="{"mlflow-bucket": "zenfiles"}"
  ```

### Deleting resources

If you wish to disable any of the stack components, use the destroy CLI command.

- Destroy individual components: You can delete one or more components using a similar experience to the deploy command.

  ```
  zenml stack recipe destroy gcp-modular -d "seldon" -e "mlflow"
  ```
  This will delete those deployments and empty their outputs. 

- Destroy all resources: If you wish to delete all resources deployed through a recipe, run the destroy command with no options.
  
  ```
  zenml stack recipe destroy gcp-modular
  ```
  This will clean all resources and remove the recipe from active recipes in your ZenML config.

### Deployment

You can deploy any stack component using the deploy subcommand.
For example, in order to deploy a GCS artifact store, you would run

  ```
  zenml artifact-store deploy my_store --flavor=gcp
  ```

This automatically registers the newly created store as an artifact store on your ZenML server.

You can also configure details about this stack component by passing a config value to the command directly.

  ```
  zenml artifact-store deploy my_store --flavor=gcp --bucket_name=unique_name
  ```

### Destroying resources

Run the following to destroy the GCS bucket from before.

  ```
  zenml artifact-store destroy my_store
  ```

## Other changes

- Get outputs from your recipes using the new `output` CLI command. You can either get one output or all depending on whether you pass an argument or not.
  
  ```
  zenml stack recipe output gcp-modular
  ```
  
  or
  
  ```
  zenml stack recipe output gcp-modular --output "mlflow-bucket"
  ```

- All terraform commands are now executed with `refresh=False` set by default.

## Pre-requisites

## Types of changes


src/zenml/cli/stack_recipes.py
```diff
@@ -251,7 +250,9 @@ def clone(self) -> None:
 
             logger.info(f"Downloading recipes to {self.cloning_path}")
             self.repo = Repo.clone_from(
-                STACK_RECIPES_GITHUB_REPO, self.cloning_path, branch="main"
+                STACK_RECIPES_GITHUB_REPO,
+                self.cloning_path,
+                branch="feature/add-sagemaker",
```
safoinme, line 255: Is this left-over or do we want to get it from a different branch?

src/zenml/cli/stack_recipes.py
```diff
@@ -633,7 +634,7 @@ def pull(
         path: The path at which you want to install the stack_recipe(s).
     """
     cli_utils.warning(ALPHA_MESSAGE)
-    git_stack_recipes_handler.pull(branch="main", force=force)
+    git_stack_recipes_handler.pull(branch="feature/add-sagemaker", force=force)
```
safoinme, line 637: same as previous comment

safoinme (changes requested):

src/zenml/cli/stack_recipes.py
```diff
@@ -1212,3 +1451,114 @@ def destroy(
...
+
+    Args:
+        git_stack_recipes_handler: The GitStackRecipesHandler instance.
+        stack_recipe_name: The name of the stack_recipe.
+        path: The path of the stack recipe you want to get the outputs from.
+        output: The name of the output you want to get the value of. If none is given,
+            all outputs are returned.
+
+    Returns:
+        A dictionary with the outputs of the stack recipe.
```
strickvl, line 1497: Do we want to offer different options for this output? i.e. YAML format or JSON? Maybe as an `--export` option along with the format?

src/zenml/client.py
```diff
@@ -2229,6 +2247,213 @@ def delete_stack_component(
...
+        )
+        # Create the new component model
+        result = ctx.invoke(
+            deploy,
+            path=path,
+            stack_recipe_name=f"{cloud}-modular",
+            config=configuration,
+            no_server=True,
+            **component,
+        )
```
strickvl, line 2304: Not sure this feels right to be invoking a CLI command within the `client` code.

src/zenml/client.py
```diff
@@ -2229,6 +2247,213 @@ def delete_stack_component(
...
+            metadata=metadata,
+        )
+
+        # if the component is an experiment tracker of flavor mlflow, then
+        # output the name of the mlflow bucket if it exists
+        if (
+            component_type == StackComponentType.EXPERIMENT_TRACKER
+            and flavor == "mlflow"
+        ):
+            # TODO check if key in outputs
```
strickvl, line 2347: reminder of TODO

src/zenml/client.py
```diff
@@ -2229,6 +2247,213 @@ def delete_stack_component(
...
+        # set the stack component and flavor
+        component_flavor = {component_type.value: component.flavor}
+
+        # Invoke the destroy command
+        ctx.invoke(
+            destroy,
+            path=path,
+            stack_recipe_name=f"{component.metadata['cloud']}-modular",
+            **component_flavor,
+        )
```
strickvl, line 2444: again, not sure this is the right behavior to be invoking CLI commands within the `client`

src/zenml/zen_stores/migrations/versions/f32bde26b014_metadata.py
```diff
@@ -0,0 +1,35 @@
...
+        batch_op.add_column(
+            sa.Column("metadata_values", sa.LargeBinary(), nullable=True)
+        )
+
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    """Downgrade database schema and/or data back to the previous revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
```
strickvl, line 31:
```suggestion
```
If you've checked what it is doing is appropriate.

src/zenml/client.py
strickvl: Thinking about this more, we should probably rather be invoking whatever commands / class etc that the CLI is invoking here within its implementation rather than invoking the click command itself.

src/zenml/cli/stack_components.py
```diff
@@ -1017,6 +1064,157 @@ def delete_stack_component_flavor_command(name_or_id: str) -> None:
...
+    @click.argument("args", nargs=-1, type=click.UNPROCESSED)
+    @click.pass_context
+    def deploy_stack_component_command(
+        ctx: click.Context,
+        name: str,
+        flavor: str,
+        cloud: str,
+        args: List[str],
+    ) -> None:
+        """Deploy a stack component.
```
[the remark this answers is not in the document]
wjayesh: what examples would you suggest? something like this?
```bash
zenml experiment-tracker deploy mlflow_tracker --flavor=mlflow --mlflow_bucket=gs://my_bucket
```

I will add that the stack component is registered.

src/zenml/cli/stack_components.py
```diff
@@ -1017,6 +1064,157 @@ def delete_stack_component_flavor_command(name_or_id: str) -> None:
...
+            "step_operator": ["sagemaker"],
+            "secrets_manager": ["aws", "gcp"],
+        }
+
+        # if the flavor is not allowed for the given component type
+        # raise an error
+        if flavor not in allowed_flavors[component_type.value]:
+            raise cli_utils.error(
+                f"Flavor '{flavor}' is not supported for "
+                f"{_component_display_name(component_type, True)}."
```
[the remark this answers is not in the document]
wjayesh: good idea, adding that

src/zenml/cli/stack_components.py
```diff
@@ -1017,6 +1064,157 @@ def delete_stack_component_flavor_command(name_or_id: str) -> None:
...
+        "name_id_or_prefix",
+        type=str,
+        required=True,
+    )
+    @click.pass_context
+    def destroy_stack_component_command(
+        ctx: click.Context,
+        name_id_or_prefix: str,
+    ) -> None:
+        """Deploy a stack component.
```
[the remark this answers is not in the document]
wjayesh: thanks for catching this!

src/zenml/cli/stack_components.py
```diff
@@ -1017,6 +1064,157 @@ def delete_stack_component_flavor_command(name_or_id: str) -> None:
...
+    """
+    display_name = _component_display_name(component_type)
+
+    @click.argument(
+        "name_id_or_prefix",
+        type=str,
+        required=True,
+    )
+    @click.pass_context
+    def destroy_stack_component_command(
```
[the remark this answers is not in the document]
wjayesh: I agree. will add those

src/zenml/cli/stack_recipes.py
```diff
@@ -825,15 +871,51 @@ def deploy(
...
+        experiment_tracker: The flavor of experiment tracker to deploy.
+        secrets_manager: The flavor of secrets manager to deploy. In the case of
+            the secrets manager, it doesn't matter what you specify here, as
+            there's only one flavor per cloud provider and that will be deployed.
+        step_operator: The flavor of step operator to deploy.
+        config: Use a YAML or JSON configuration or configuration file to pass
+            variables to the stack recipe.
+
+    Returns:
+        The exit code of the command. 1 if deployment successful, 0 otherwise.
```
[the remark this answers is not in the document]
wjayesh: the use case was to know if a deployment was cancelled from a prompt and then exit the program without raising an error (since it is not one, technically)/ But I agree that it looks out of place and so I have removed it in favour of using `cli_utils.error()`

src/zenml/cli/stack_recipes.py
```diff
@@ -1049,10 +1173,13 @@ def deploy(
                             "metadata-db-password' to get the "
                             "value in the command-line."
                         )
+                else:
+                    logger.info("Deployment cancelled.")
+                    return 1
```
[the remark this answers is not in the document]
wjayesh: yup, this case happens when someone says no to the confirmation prompt for checking the values. I think this will go away soon when I start work on refining the logs

src/zenml/client.py
wjayesh: I agree that it doesn't feel right but there's a lot of code that needs to be executed when you are about to deploy the recipe. Stuff like checking prerequisites, checking if existing service exists, not to mention error handling. In other words, everything in that deploy CLI function would have to be rewriten inside the client function which I think is not very good. Still open to changes however

strickvl: I think the idea is that you'd have some util function(s) / or class which deals with deployment that you can call from both places. I think you'd want to have all that functionality in one place that can be called independently of the CLI. Where that code lives I guess is an open question. (Maybe @schustmi you have thoughts on whether it should be on the client or separately as a different chunk of functionality.) But I think it's an opportunity to a) not have CLI code in the client as that's not a good pattern and b) have an abstract way to do everything that needs handling (everything you mention above).

src/zenml/cli/stack_recipes.py
wjayesh: done!

src/zenml/client.py
```diff
@@ -2229,6 +2247,213 @@ def delete_stack_component(
...
+            and flavor == "mlflow"
+        ):
+            # TODO check if key in outputs
+            mlflow_bucket = outputs.get("mlflow-bucket")
+            if mlflow_bucket:
+                logger.info(
+                    f"The bucket used for MLflow is: %s {mlflow_bucket}"
+                    "You can use this bucket as an artifact store to "
+                    "avoid having to create a new one."
+                )
```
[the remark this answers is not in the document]
wjayesh: thanks! I've removed the f.

src/zenml/client.py
```diff
@@ -2229,6 +2247,213 @@ def delete_stack_component(
...
+                io_utils.get_global_config_directory(),
+                STACK_COMPONENT_RECIPE_DIR,
+            )
+        )
+
+        # set the stack component and flavor
+        component_flavor = {component_type.value: component.flavor}
+
+        # Invoke the destroy command
+        ctx.invoke(
```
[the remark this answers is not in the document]
wjayesh: the `ModuleNotFoundError` is an exception that can be thrown from this method and I've added a catch for it.

src/zenml/client.py
```diff
@@ -2275,6 +2276,228 @@ def delete_stack_component(
...
+        path = Path(
+            os.path.join(
+                io_utils.get_global_config_directory(),
+                STACK_COMPONENT_RECIPE_DIR,
+            )
+        )
+
+        try:
+            # Create the new component model
+            ctx.invoke(
```
strickvl, line 2329: Still think this a bit of an anti-pattern to invoke the CLI from within the client code. Could you extract this out somewhere else, so that the deployment is handled by some function that we invoke here (and that can then be used by the CLI deploy command as well as here in the client code?

src/zenml/cli/utils.py
```diff
@@ -1437,15 +1437,28 @@ def inner_decorator(func: F) -> F:
...
+                # if k is metadata, add option with multiple values
+                if k == "metadata_values":
+                    options.append(
+                        click.option(
+                            "--metadata",
+                            "-m",
+                            type=str,
+                            multiple=True,
+                            required=False,
+                            help=create_filter_help_text(filter_model, k),
```
stefannica, line 1449: The field name should be `labels`, not `metadata`.

Second thing: this is a generic decorator, so adding exceptions for attribute names here is not going to scale well. Just imagine 10 different if clauses piling up here, one for every special case.

The correct way to do it is as follows:
1. add the `labels` CLI argument to the "list stack component" command, like I did here with `label`: https://github.com/zenml-io/zenml/blob/af927f2bc1399b27fdbd2438e4aa6df64511442b/src/zenml/cli/service_connectors.py#L372-L379
2. inhibit the `label` field in the StackComponentFilterModel, like I did with the ServiceConnectorFilterModel here: https://github.com/zenml-io/zenml/blob/af927f2bc1399b27fdbd2438e4aa6df64511442b/src/zenml/models/service_connector_models.py#L938 . This tells the generic decorator to not process this field.
3. parse and add the `label` values to the filter model in the CLI command implementation, like I did here with labels: https://github.com/zenml-io/zenml/blob/af927f2bc1399b27fdbd2438e4aa6df64511442b/src/zenml/cli/service_connectors.py#L849-L854

You should maybe run some tests to make sure this works.

src/zenml/cli/stack_components.py
```diff
@@ -1056,6 +1056,225 @@ def delete_stack_component_flavor_command(name_or_id: str) -> None:
...
+                )
+
+            client.deploy_stack_component(
+                name=name,
+                flavor=flavor,
+                cloud=cloud,
+                configuration=parsed_args,
+                component_type=component_type,
+                labels={"cloud": cloud, "created_by": "recipe"},
+            )
```
stefannica, line 1197: Loving this ! All the complexity is hidden behind the Client instead of being layed out here in the CLI. This is good design !

stefannica (changes requested):

src/zenml/cli/stack_recipes.py
```diff
@@ -776,21 +776,67 @@ def pull(
...
+    "-e",
+    help="The flavor of experiment tracker to use. "
+    "If not specified, no experiment tracker will be deployed.",
+)
+@click.option(
+    "--secrets-manager",
+    "-x",
+    help="The flavor of secrets manager to use. "
+    "If not specified, no secrets manager will be deployed.",
+)
```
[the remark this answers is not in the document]
wjayesh: served me really well to test this feature out quickly. May it rest in peace!

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -562,6 +562,7 @@ def _create_default_stack(
                 type=StackComponentType.ORCHESTRATOR,
                 flavor="local",
                 configuration={},
+                metadata={"cloud": "local"},
```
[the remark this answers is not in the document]
wjayesh: You're right and thanks for the explanation! I had removed them from the metadata PR following your review earlier but it somehow lingered around in this one. Thanks for catching it, yet again :)

src/zenml/cli/utils.py
wjayesh: This is beautiful. Thanks for the fix, I would however, remove this altogether like we did with the other PR. This is because we currently don't support listing by labels due to the fact that we store labels as bytes and filtering by that is a bit complex.
I'd however come back to this later and use this design when I do.

src/zenml/cli/stack_components.py
wjayesh: Thanks, happy to hear that you approve of this :)

src/zenml/cli/utils.py
stefannica: Yeah, label filtering is a pain, I agree. I'm on board with leaving it out until we get it right in the back-end.

PR #1333, opened 2023-02-17
Compile dev dependencies for custom requirements.txt file

strickvl:
I've made the most minimal version of this story. We retain the `pyproject.toml` as our preferred way of tracking dev and mainstream dependencies for `zenml`, but there would be an additional compilation of these two `requirements.txt` files, one for dev and one for mainstream.

(I'd propose adding a stage to the [Make a Release](https://www.notion.so/zenml/Make-A-Release-adb5e415107e4189bbb894a4c08c00f1?pvs=4#01a91b6527a542b485fb3a88d3c4d7ae) internal page where someone runs the `scripts/compile-dependencies.sh` script and commits it to develop. Optionally, I can also add / update the CONTRIBUTING.md document for external contributors to have them using this file as well instead of requiring them to use `poetry install`).

You'll have to `pip install pip-tools` first, but otherwise this is a fairly easy way of removing the requirement to install dev dependencies via Poetry.

## Pre-requisites

## Types of changes


schustmi:
I'll just throw another idea in the ring which might be even simpler (it doesn't require `pip compile`):
- We track our normal dependencies (including the extras for the server, recipes) in the `pyproject.toml` as we did before. This already enables us to install these dependencies using `pip install -e .`. I also think the toml is very convenient to keep the options for all the other tools, so we would keep it anyway
- Remove the dev dependencies from the toml and put it in `dev-requirements.txt`

Additionally (whether we use my solution or the one currently implemented), I'd suggest we use `pip install` in the CI instead of poetry just to speed that up a little. I guess that leaves poetry still in place for the release process, but nothing else.

strickvl:
> I'll just throw another idea in the ring which might be even simpler (it doesn't require `pip compile`):
> 
> * We track our normal dependencies (including the extras for the server, recipes) in the `pyproject.toml` as we did before. This already enables us to install these dependencies using `pip install -e .`. I also think the toml is very convenient to keep the options for all the other tools, so we would keep it anyway
> * Remove the dev dependencies from the toml and put it in `dev-requirements.txt`
> 
> Additionally (whether we use my solution or the one currently implemented), I'd suggest we use `pip install` in the CI instead of poetry just to speed that up a little. I guess that leaves poetry still in place for the release process, but nothing else.

I discussed this a bit with @AlexejPenner earlier this week. The tricky thing about removing poetry from the CI is that our package caching is currently all built around us using Poetry. It's [possible](https://github.com/actions/cache/blob/main/examples.md#python---pip) to have it work with `pip` since the caching action supports that, but I would def need some eyes over my shoulder while removing that since the potential of breaking something is fairly high I think.

strickvl:
I [listed some of the way](https://www.notion.so/zenml/Removing-Poetry-74ba723c0a3d46a3a43158eb512e8a69) we currently use Poetry here.

schustmi:
I think we don't even need any major changes to the caching:
- it caches the entire virtualenv directory, which will still be used by pip to install the packages
- the cache key (which is used to determine whether the cache should be loaded) only contains the pyproject.toml, not the poetry.lock file. We could simply add the `dev-requirements.txt` to the cache key and would achieve the same behaviour

The other points on the notion page should also be very quick IMO, happy to go over them next week

strickvl:
Closing this as it has been implemented over on #1346.

PR #1336, opened 2023-02-20
Update README for test environment provisioning

strickvl:
- The `docker-compose` pip package is required for some environments.
- Also removed the reference to the `pytest-xdist`package / flag for running in parallel since this doesn't reliably work and is probably more confusion than guidance. We're also removing the package in #1334 for this reason.

## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -83,7 +83,10 @@ with the test framework CLI, e.g.:
 
 2. Provision the test environment. This will configure and start a local ZenML
 deployment and will register and provision the stack components that are
-configured for the environment. The test framework will also take care of
+configured for the environment. (Note that some environments require the 
+`docker-compose` pip package to be installed, so make sure to do that.
+
+The test framework will also take care of
 rebuilding the ZenML container images that are required for the server to run
 or for the pipelines to be executed. E.g.:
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -144,15 +147,6 @@ volumes and images that might be created during the test run. The `--no-provisio
 flag is used to skip provisioning the environment, which is already done in
 step 2.
 
-With an environment provisioned separately, you could even run the tests in
```
stefannica, line 147: I would keep this, because more and more integration tests are heading in this direction and some can already be run in parallel (examples).

stefannica: If you plan on removing the `pytest-xdist` package from the development requirements, you should document here that you need it instead.

stefannica (changes requested):

tests/README.md
strickvl: `pytest-xdist` is already gone in #1334.

strickvl: I'll add something abuot the need to install it if you want to do that.

stefannica (approved):

PR #1337, opened 2023-02-20
Add dependency resolution docs

strickvl:
Added an initial attempt at docs on dependency resolution. Welcome suggestions on how this could be improved.

## Pre-requisites

## Types of changes


docs/book/misc/handling-dependencies.md
```diff
@@ -0,0 +1,42 @@
...
+flexibility comes the possibility of dependency conflicts.
+
+## Suggestions for Resolving Dependency Conflicts
+
+### Use a tool like `pip-compile` for reproduciblity
+
+Consider using a tool like `pip-compile` (available through [the `pip-tools`
+package](https://pip-tools.readthedocs.io/)) to compile your dependencies into a
+static `requirements.txt` file that can be used across environments.
+
```
stefannica, line 23: You should link the gitflow section on dealing with dependencies with pip-compile here: https://github.com/zenml-io/zenml-gitflow#-software-requirements-management

docs/book/misc/handling-dependencies.md
```diff
@@ -0,0 +1,42 @@
...
+### Well-known dependency resolution issues
```
stefannica, line 24: You should add another section that explains how they can use `pip check` to show possible dependency conflicts. Same idea here: they might be harmless, but users should be aware of them in case problems do occur when they run pipelines.

stefannica: Here's one more idea: mention what happens if the dependencies required by zenml integrations are not respected: some components might stop working or throwing errors. A good way to check if these are met is to use `zenml integration list` to see if an integration is active.

stefannica (changes requested):
Some suggestions for improvement, but overall looks good

htahir1 (commented):
Maybe you can also add the fact that one can skip doing `zenml integration install...` and directly do pip install (and subsequently put these requirements in DockerSettings in case of a remote orchestrator) at their own risk

stefannica (approved):

docs/book/misc/handling-dependencies.md
```diff
@@ -0,0 +1,80 @@
...
+  account of another dependency of ZenML. Using versions of `click` in your own
+  project that are greater than 8.0.3 may cause unanticipated behaviors.
+
+### Manually bypassing ZenML's integration installation
+
+It is possible to skip ZenML's integration installation process and install
+dependencies manually. This is not recommended, but it is possible and can be
+run at your own risk.
+
+To do this, you will need to install the dependencies for the integration you
```
htahir1, line 73: Maybe add here that `zenml integration install` essentially runs a `pip install ..` in the background and give them an example of an integration install like `zenml integration install gcp` vs a `pip install gcp-lib-1 gcp-lib-2` etc

htahir1 (changes requested):
Left one more comment

htahir1 (approved):

PR #1345, opened 2023-02-21
Disable name prefix matching when updating/deleting entities

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -2235,11 +2235,13 @@ class Client(metaclass=ClientMetaClass):
     def create_flavor(
         self,
         source: str,
+        component_type: StackComponentType,
     ) -> "FlavorResponseModel":
         """Creates a new flavor.
 
         Args:
             source: The flavor to create.
+            component_type: The type of the flavor.
 
         Returns:
             The created flavor (in model form).
@@ -2250,7 +2252,7 @@ class Client(metaclass=ClientMetaClass):
         from zenml.utils.source_utils import validate_flavor_source
 
         flavor = validate_flavor_source(
-            source=source,
+            source=source, component_type=component_type
         )()
 
         if len(flavor.config_schema) > TEXT_FIELD_MAX_LENGTH:
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -646,12 +646,15 @@ class Client(metaclass=ClientMetaClass):
         return created_user
 
     def get_user(
-        self, name_id_or_prefix: Union[str, UUID]
+        self,
+        name_id_or_prefix: Union[str, UUID],
+        allow_name_prefix_match: bool = True,
     ) -> UserResponseModel:
         """Gets a user.
 
         Args:
             name_id_or_prefix: The name or ID of the user.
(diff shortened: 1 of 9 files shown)


src/zenml/client.py
```diff
@@ -3822,23 +3880,29 @@ def _get_entity_by_prefix(
...
-            logical_operator=LogicalOperators.OR,
-            name=f"contains:{partial_id_or_name}",
-            id=f"startswith:{partial_id_or_name}",
-        )
+        list_method_args: Dict[str, Any] = {
+            "logical_operator": LogicalOperators.OR,
+            "id": f"startswith:{partial_id_or_name}",
+        }
+        if allow_name_prefix_match:
+            list_method_args["name"] = f"startswith:{partial_id_or_name}"
```
stefannica, line 3888:
```suggestion
        if allow_name_prefix_match:
            list_method_args["name"] = f"startswith:{partial_id_or_name}"
        else:
            list_method_args["name"] = partial_id_or_name        
```

I think you missed this

tests/integration/functional/test_client.py
```diff
@@ -964,3 +965,134 @@ def test_deleting_deployments(clean_client):
...
+    get_args: Dict[str, Any] = {}
+    update_args: Dict[str, Any] = {}
+    delete_args: Dict[str, Any] = {}
+
+
+crud_test_configs = [
+    ClientCrudTestConfig(
+        entity_name="user",
+        create_args={"name": "user_name"},
+        update_args={"updated_name": "updated_user_name"},
```
stefannica, line 982: I love this generalized way of testing the client !

Could you please use some way to randomize the names of the created resources, similar to what we do in the zen_store generic CRUD tests ? As these tests will also be run on shared CI environments, they will trample over one-another. Using random names also makes it easier to run these tests in parallel in the future, or in another dimension.

tests/integration/functional/test_client.py
```diff
@@ -964,3 +965,134 @@ def test_deleting_deployments(clean_client):
...
+    """Tests basic CRUD method on the client."""
+    create_method = getattr(
+        clean_client, f"create_{crud_test_config.entity_name}"
+    )
+    get_method = getattr(clean_client, f"get_{crud_test_config.entity_name}")
+    delete_method = getattr(
+        clean_client, f"delete_{crud_test_config.entity_name}"
+    )
+
+    entity = create_method(**crud_test_config.create_args)
```
stefannica, line 1052: Could you please wrap this into a nice try/finally block and clean up any remaining resources at the end using the generated random name? If the test fails mid-way, those will be left on the test environment.

stefannica (changes requested):
Thanks for making this a better UX, I know it couldn't have been easy going through all those client API functions.

src/zenml/client.py
schustmi: As far as I understood, that case is already handled in the method that calls this, which covers both the exact ID and exact name match

stefannica: Ah, of course. You're right.

stefannica (commented):
LGTM

PR #1346, opened 2023-02-22
Remove poetry from CI

schustmi:
## Describe changes

This PR switches to `pip` for installing ZenML in development environments and the CI.
`poetry` is removed everywhere except the release publishing.

## Pre-requisites

## Types of changes


CONTRIBUTING.md
```diff
@@ -117,20 +117,18 @@ the ["fork-and-pull" Git workflow](https://github.com/susam/gitpr)
...
 ### 🧐 Linting, formatting, and tests
 
-ZenML is mainly developed using [Poetry](https://python-poetry.org/) as the
-dependency management system. In order to
-install all core dev-dependencies, do:
+To install ZenML from your local checked out files includin all core dev-dependencies, run:
 
 ```
-poetry install
+pip install -e ".[dev]"
```
strickvl, line 123: Wouldn't they want the server as well for developing with ZenML?

pyproject.toml
```diff
@@ -85,55 +85,50 @@ jinja2-time = { version = "^0.2.0", optional = true }
...
+types-requests = { version = "^2.27.11", optional = true }
+types-setuptools = { version = "^57.4.2", optional = true }
+types-six = { version = "^1.16.2", optional = true }
+types-termcolor = { version = "^1.1.2", optional = true }
+types-psutil = { version = "^5.8.13", optional = true }
 # test framework dependencies
-docker-compose = "^1.28"
+docker-compose = { version = "^1.28", optional = true }
+
+[tool.poetry.extras]
```
strickvl, line 128: why is this still referring to Poetry?

src/zenml/README.md
```diff
@@ -18,43 +15,41 @@ git clone git@github.com:zenml-io/zenml.git
...
-poetry shell
+pip install --upgrade pip
 ```
 
 ## Install dependencies
+
 Then from the root of the package:
 ```bash
-poetry install
+pip install -e ".[dev]"
```
strickvl, line 40: wouldn't they also want the server dependencies?

.github/workflows/ci.yml
```diff
@@ -16,23 +16,23 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  poetry-install:
+  install:
```
strickvl, line 19: `install` sounds a bit close to a reserved word; perhaps worth making less generic just in case?

strickvl (commented):
This looks good to me and assume it has been tested alongside the CI / local installs etc. Some questions only...

CONTRIBUTING.md
schustmi: Potentially but not necessarily? I just copied the previous behaviour tbh, which also didn't install the server extra. I'll change it

strickvl: Yeah either add both by default, or explain the different combinations they might want to use depending on what they're doing?

pyproject.toml
schustmi: This is for the publishing to pypi which still uses poetry. However this poetry value is a synonym that also works for pip (See https://peps.python.org/pep-0621/)

.github/workflows/ci.yml
schustmi: Mhm it seems to work without any issues, what other name would you suggest?

strickvl: install-environment? install-python-environment?

stefannica (approved):
YES !

strickvl (approved):

AlexejPenner (approved):
:seal:

PR #1348, opened 2023-02-22
Add Cloud Secrets Store back-ends

stefannica:
## Describe changes

Here we implement 4 different ZenML Secret Stores to replace the 4 Secrets Manager stack components:

* AWS Secrets Store
* GCP Secrets Store
* Azure Secrets Store
* HashiCorp Vault Secrets Store

#### Implementation highlights

* the name/ID of an AWS secret is derived from the ZenML secret UUID and a
`zenml` prefix in the form `zenml/{zenml_secret_uuid}`. This clearly
identifies a secret as being managed by ZenML in the AWS console.

* the Secrets Store also makes heavy use of AWS secret tags to store all the
metadata associated with a ZenML secret (e.g. the secret name, scope, user
and workspace) and to filter secrets by these metadata. The `zenml` tag in
particular is used to identify and group all secrets that belong to the same
ZenML deployment.

* all secret key-values configured in a ZenML secret are stored as a single
JSON string value in the AWS secret value.

* when a user or workspace is deleted, the secrets associated with it are
deleted automatically via registered event handlers.

#### Known challenges and limitations:

* there is a known problem with the AWS Secrets Manager API that can cause
the `list_secrets` method to return stale data for a long time (seconds)
after a secret is created or updated. The AWS secrets store tries to
mitigate this problem by waiting for a maximum configurable number of
seconds after creating or updating a secret until the changes are reflected
in the `list_secrets` AWS content. However, this is not a perfect solution,
because it blocks ZenML server API threads while waiting. This can be
disabled by setting the `secret_list_refresh_timeout` configuration
parameter to zero.

* only updating the secret values is reflected in the secret's `updated`
timestamp. Updating the secret metadata (e.g. name, scope, user or
workspace) does not update the secret's `updated` timestamp. This is a
limitation of the AWS Secrets Manager API (updating AWS tags does not update
the secret's `updated` timestamp).

* the AWS Secrets Store API does not natively support some of the
filtering, sorting or pagination options that ZenML supports. The
implementation of the `list_secrets` method is therefore a bit
involved. We try to make use of the AWS filtering API as much as
possible to reduce the number of secrets that we need to fetch, then
we apply the rest of filtering, sorting and pagination on
the client side.

#### Implementation highlights

* the name/ID of an Azure secret is derived from the ZenML secret UUID and a
`zenml` prefix in the form `zenml-{zenml_secret_uuid}`. This clearly
identifies a secret as being managed by ZenML in the Azure console.

* the Secrets Store also makes heavy use of Azure Key Vault secret tags to
store all the metadata associated with a ZenML secret (e.g. the secret name,
scope, user and workspace) and to filter secrets by these metadata. The
`zenml` tag in particular is used to identify and group all secrets that
belong to the same ZenML deployment.

* all secret key-values configured in a ZenML secret are stored as a single
JSON string value in the Azure Key Vault secret value.

* when a user or workspace is deleted, the secrets associated with it are
deleted automatically via registered event handlers.

#### Known challenges and limitations

* every Azure Key Vault secret has one or more versions. Every update to a
secret creates a new version. The created_on and updated_on timestamps
returned by the Secrets Store API are the timestamps of the latest version
of the secret. This means that we need to fetch the first version of the
secret to get the created_on timestamp. This is not ideal, as we'd need to
fetch all versions for every secret to get the created_on timestamp during
a list operation. So instead we manage the `created` and `updated`
timestamps ourselves and save them as tags in the Azure Key Vault secret.

* the Azure Key Vault API does not natively support any of the
filtering, sorting or pagination options that ZenML supports. The
implementation of the `list_secrets` method therefore has to fetch all secrets from
the Key Vault, then apply the filtering, sorting and pagination on
the client side.

#### Implementation highlights

* the name/ID of an HashiCorp Vault secret is derived from the ZenML secret
UUID and a `zenml` prefix in the form `zenml/{zenml_secret_uuid}`. This
clearly identifies a secret as being managed by ZenML in the HashiCorp Vault
server. This also allows use to reduce the scope of `list_secrets` to cover
only secrets managed by ZenML by using `zenml/` as the path prefix.

* given that HashiCorp Vault secrets do not support attaching arbitrary
metadata in the form of label or tags, we store the entire ZenML secret
metadata (e.g. name, scope, etc.) alongside the secret values in the
HashiCorp Vault secret value.

* when a user or workspace is deleted, the secrets associated with it are
deleted automatically via registered event handlers.

#### Known challenges and limitations

* HashiCorp Vault secrets do not support filtering secrets by metadata
attached to secrets in the form of label or tags. This means that we cannot
filter secrets server-side based on their metadata (e.g. name, scope, etc.).
Instead, we have to retrieve all ZenML managed secrets and filter them
client-side.

* HashiCorp Vault secrets are versioned. This means that when a secret is
updated, a new version is created which has its own creation timestamp.
Furthermore, older secret versions are deleted automatically after a certain
configurable number of versions is reached. To work around this, we also
manage `created` and `updated` timestamps here and store them in the secret
value itself.

## Pre-requisites

## Types of changes


src/zenml/integrations/aws/secrets_stores/aws_secrets_store.py
```diff
@@ -0,0 +1,1047 @@
...
+            The newly created secret.
+
+        Raises:
+            EntityExistsError: If a secret with the same name already exists
+                in the same scope.
+            RuntimeError: If the AWS Secrets Manager API returns an unexpected
+                error.
+        """
+        self._validate_aws_secret_name(secret.name)
+        user, workspace = self._validate_user_and_workspace(
```
schustmi, line 590: I know that we usually also verify that this user is the authenticated user (-> you can't create a secret for a user other than yourself). This is already covered by the endpoint logic and we don't really care about the local case?

src/zenml/integrations/aws/secrets_stores/aws_secrets_store.py
```diff
@@ -0,0 +1,1047 @@
...
+                error.
+        """
+        aws_secret_id = self._get_aws_secret_id(secret_id)
+
+        try:
+            get_secret_value_response = self.client.get_secret_value(
+                SecretId=aws_secret_id
+            )
+            # We need a separate AWS API call to get the AWS secret tags which
+            # contain the ZenML secret metadata, since the create_secret API
```
schustmi, line 705:
```suggestion
            # contain the ZenML secret metadata, since the get_secret_ value API
```

src/zenml/integrations/aws/secrets_stores/aws_secrets_store.py
```diff
@@ -0,0 +1,1047 @@
...
+                SecretId=aws_secret_id
+            )
+        except ClientError as e:
+            raise RuntimeError(f"Error updating secret: {e}")
+
+        logger.debug("Updated AWS secret: %s", aws_secret_id)
+
+        # We wait for the secret update to be reflected in the `list_secrets`
+        # API.
+        for _ in range(self.config.secret_list_refresh_timeout):
```
schustmi, line 966: Can we extract that into a separate method that also handles it in the `create_secret` case?

schustmi (changes requested):
Great stuff, I can imagine how painful it was to implement the sorting/filtering manually now that we get that functionality for free for our SQL/Rest stores. Have a few questions but those are mostly for my understanding.

docker/base.Dockerfile
```diff
@@ -17,6 +17,9 @@ FROM base AS server
 
 RUN pip install zenml${ZENML_VERSION:+==$ZENML_VERSION}[server]
 
+# Required for the AWS secrets store
```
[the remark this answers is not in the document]
stefannica: As we agreed offline, I'll change this to be another zenml package extra that only installs the minimum necessary for the AWS secrets store to function.

src/zenml/integrations/aws/secrets_stores/aws_secrets_store.py
stefannica: Yes, it's already covered.

src/zenml/integrations/aws/secrets_stores/aws_secrets_store.py
stefannica: done

docs/book/advanced-guide/practical/secrets-management.md
```diff
@@ -4,38 +4,44 @@ description: How to register and use secrets
...
+connected to a remote ZenML server, the secrets are stored in the secrets
+management back-end that the server is configured to use, but all access to the
+secrets is done through the ZenML server API.
 
-Currently, the only back-end supported for the centralized secrets store is
-the ZenML SQL database, but more back-ends will be added in the near future to
-replace the existing [secrets manager](../../component-gallery/secrets-managers/secrets-managers.md)
-stack components:
+Currently, the ZenML server can be configured to use one of the following
+supported secret store back-ends:
```
strickvl, line 32:
```suggestion
supported secret store back-ends:
```
```suggestion
supported secrets store back-ends:
```

docs/book/advanced-guide/practical/secrets-management.md
```diff
@@ -312,3 +240,174 @@ This will only work if the environment that your orchestrator uses to execute
...
+provided `zenml secrets-manager secrets migrate` CLI command to migrate your
+secrets to the centralized secrets store.
+
+Managing secrets through a secrets manager stack component suffers from a
+number of limitations, some of which are:
+
+* you need to configure [a Secrets Manager stack component](../../component-gallery/secrets-managers/secrets-managers.md)
+and add it to your active stack before you can register and access secrets. With
+centralized secrets management, you don't need to configure anything, your ZenML
+local deployment or ZenML server replaces the secrets manager role.
```
strickvl, line 266:
```suggestion
local deployment or ZenML server takes on the secrets manager role.
```

docs/book/component-gallery/secrets-managers/secrets-managers.md
```diff
@@ -31,11 +35,25 @@ is also a security risk, because it basically represents a large attack surface.
...
+* AWS Secrets Manager - replaces the `aws` secrets manager flavor.
+* GCP Secret Manager - replaces the `gcp` secrets manager flavor.
+* Azure Key Vault - replaces the `azure` secrets manager flavor.
+* HashiCorp Vault - replaces the `vault` secrets manager flavor.
+
+The centralized secrets store also supports using a custom back-end
+implementation.
+
+There is no direct migration path planned for the GitHub secrets manager flavor,
+given that it can only be used inside a GitHub Action workflow and thus is not
```
strickvl, line 53:
```suggestion
given that it can only be used inside a GitHub Actions workflow and thus is not
```

docs/book/component-gallery/secrets-managers/secrets-managers.md
```diff
@@ -31,11 +35,25 @@ is also a security risk, because it basically represents a large attack surface.
...
+a service that can be used as a centralized secrets store.
+If you are using the GitHub secrets manager flavor, you have the option of
+migrating your secrets to one of the other supported secrets store back-ends.
```
strickvl, line 56:
```suggestion
manually transferring your secrets to one of the other supported secrets store back-ends.
```
'migrating' seems a bit weasly, given that we're telling them that we do not help them with that...

docs/book/getting-started/deploying-zenml/helm.md
```diff
@@ -390,6 +441,210 @@ you can map the hostname `zenml.<subdomain>` to the Ingress controller's IP
...
+
+    # The type of the secrets store
+    type: gcp
+
+    # Configuration for the GCP Secrets Manager secrets store
+    gcp:
+
+      # The GCP project ID to use. This must be set to the project ID where the
+      # GCP Secrets Manager service that you want to use is located.
+      project_id: us-east-1
```
strickvl, line 511:
```suggestion
      project_id: my-zenml-project
```

Confusing to use a region-name value as a project id, even if technically valid...

docs/book/getting-started/deploying-zenml/helm.md
```diff
@@ -390,6 +441,210 @@ you can map the hostname `zenml.<subdomain>` to the Ingress controller's IP
...
+      # and the path configured here needs to be relative to the root of the
+      # helm chart.
+      google_application_credentials: cloud-credentials.json
+
+serviceAccount:
+
+  # If you're using workload identity, you need to annotate the service
+  # account with the GCP service account name (see https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)
+  annotations:
+    iam.gke.io/gcp-service-account: <service-account-name>@<project-name>.iam.gserviceaccount.com
```
strickvl, line 527:
```suggestion
    iam.gke.io/gcp-service-account: <SERVICE_ACCOUNT_NAME>@<PROJECT_NAME>.iam.gserviceaccount.com
```

docs/book/getting-started/deploying-zenml/upgrades.md
```diff
@@ -0,0 +1,4 @@
+---
+description: How to upgrade ZenML
```
strickvl, line 2: What is this?

examples/secrets_management/steps/secret_loader/secret_loader_step.py
```diff
@@ -12,32 +12,32 @@
...
+            f"Could not find secret `{SECRET_NAME}`. "
+            f"Please run `zenml secrets create` to create it."
         )
+        return
 
     # Load specific secret value from the secret using the secret key
     print(
         f"The secret `{SECRET_NAME}` contains the following key-value pair "
-        f"{{{SECRET_KEY}: {retrieved_secret.content[SECRET_KEY]}}}"
+        f"{{{SECRET_KEY}: {secret.secret_values[SECRET_KEY]}}}"
```
schustmi, line 41: We can just print `secret.secret_values` here that actually lists the KV-pairs in case someone plays around with the example and adds more?

pyproject.toml
```diff
@@ -128,6 +141,10 @@ docker-compose = { version = "^1.28", optional = true }
 [tool.poetry.extras]
 server = ["fastapi", "uvicorn", "python-multipart", "python-jose", "fastapi-utils"]
 templates = ["copier", "jinja2-time", "black", "ruff"]
+secrets-aws = ["boto3"]
+secrets-gcp = ["google-cloud-secret-manager"]
+secrets-azure = ["azure-identity", "azure-keyvault-secrets"]
+secrets-hashicorp = ["hashicorp"]
```
schustmi, line 147: Shouldn't this be `hvac`, the dependency specified above?

src/zenml/config/secrets_store_config.py
```diff
@@ -31,9 +33,51 @@ class SecretsStoreConfiguration(BaseModel):
...
+            if values["class_path"] is None:
+                raise ValueError(
+                    "A class_path must be set when using a custom secrets "
+                    "store implementation."
+                )
+        elif (
+            values["type"]
+            in [
+                SecretsStoreType.SQL,
+                SecretsStoreType.REST,
```
schustmi, line 71: Is the class path allowed for aws/gcp/etc?

src/zenml/config/secrets_store_config.py
```diff
@@ -31,9 +33,51 @@ class SecretsStoreConfiguration(BaseModel):
...
+            a subclass of `BaseSecretsStore`. This is optional and only
+            required if the store backend is not one of the built-in
+            implementations.
     """
 
     type: SecretsStoreType
+    class_path: Optional[str] = None
+
+    @root_validator
+    def validate_external(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
schustmi, line 46: Just from a naming perspective, this should probably be `custom` instead of `external`? It switches between the two which is a little confusing

src/zenml/zen_stores/secrets_stores/gcp_secrets_store.py
```diff
@@ -0,0 +1,612 @@
...
+            returned secrets do not include any secret values, only metadata. To
+            fetch the secret values, use `get_secret` individually with each
+            secret.
+
+        Raises:
+            ValueError: If the filter contains an out-of-bounds page number.
+            RuntimeError: If the Azure Key Vault API returns an unexpected
+                error.
+        """
+        # convert the secret_filter_model to a GCP filter string
```
strickvl, line 412:
```suggestion
        # TODO: implement filter method for server-side filtering
        # convert the secret_filter_model to a GCP filter string
```

src/zenml/zen_stores/secrets_stores/base_secrets_store.py
```diff
@@ -75,60 +106,102 @@ def __init__(
...
+        Raises:
+            ValueError: If the configured class path cannot be imported or is
+                not a subclass of `BaseSecretsStore`.
+        """
+        # Ensured through Pydantic root validation
+        assert store_config.class_path is not None
+
+        # Import the class dynamically
+        try:
+            store_class = import_class_by_path(store_config.class_path)
```
schustmi, line 146: There is `source_utils.load_and_validate_class` which I think generalizes what you implemented here

src/zenml/zen_stores/secrets_stores/base_secrets_store.py
```diff
@@ -75,60 +106,102 @@ def __init__(
...
+        if store_config.type == SecretsStoreType.REST:
             from zenml.zen_stores.secrets_stores.rest_secrets_store import (
                 RestSecretsStore,
             )
 
             return RestSecretsStore
-        else:
+
+        if store_config.type == SecretsStoreType.AWS:
+            store_config.class_path = "zenml.zen_stores.secrets_stores.aws_secrets_store.AWSSecretsStore"
```
schustmi, line 191: Why is this handled differently than the SQL/Rest cast and the classes aren't simply imported here?

schustmi (changes requested):

src/zenml/zen_stores/secrets_stores/gcp_secrets_store.py
```diff
@@ -0,0 +1,612 @@
...
+            workspace=secret.workspace,
+            user=secret.user,
+        )
+        if secret_exists:
+            raise EntityExistsError(msg)
+
+        secret_id = uuid.uuid4()
+        secret_value = json.dumps(secret.secret_values)
+
+        created = datetime.now(timezone.utc).replace(
```
schustmi, line 298: Does `datetime.utcnow()` achieve something similar but less confusing to read?

src/zenml/zen_stores/secrets_stores/azure_secrets_store.py
```diff
@@ -0,0 +1,633 @@
...
+        # required to exclude everything but Azure secrets that belong to the
+        # current ZenML deployment.
+        results: List[SecretResponseModel] = []
+
+        try:
+            all_secrets = self.client.list_properties_of_secrets()
+            for secret_property in all_secrets:
+                try:
+                    # NOTE: we do not include the secret values in the
+                    # response. We would need a separate API call to fetch
```
schustmi, line 456: It seems like this returns values for some implementations and not for others? This is hella confusing IMO, that should be consistent.

src/zenml/zen_stores/secrets_stores/azure_secrets_store.py
```diff
@@ -0,0 +1,633 @@
...
+            if secret_exists:
+                raise EntityExistsError(msg)
+
+        azure_secret_id = self._get_azure_secret_id(secret_id)
+        secret_value = json.dumps(secret.secret_values)
+
+        # Convert the ZenML secret metadata to Azure tags
+        metadata = self._get_secret_metadata_for_secret(secret)
+
+        # We manage the created and updated times ourselves, so we need to
```
schustmi, line 581:
```suggestion
        # We manage the created and updated times ourselves, so we don't need to
```

src/zenml/zen_stores/secrets_stores/hashicorp_secrets_store.py
```diff
@@ -0,0 +1,634 @@
...
+        return self._convert_vault_secret(
+            vault_secret,
+        )
+
+    def list_secrets(
+        self, secret_filter_model: SecretFilterModel
+    ) -> Page[SecretResponseModel]:
+        """List all secrets matching the given filter criteria.
+
+        Note that returned secrets do not include any secret values. To fetch
```
schustmi, line 394: As far as I can see, the secrets actually do contain the values right?

src/zenml/enums.py
```diff
@@ -103,6 +103,11 @@ class SecretsStoreType(StrEnum):
     NONE = "none"  # indicates that the secrets store is disabled
```
schustmi, line 103: Can one actually do this? I thought we always fall back to SQL/Rest depending on the configuration of the active zen store? In `BaseSecretsStore.get_store_class`, the `NONE` case isn't covered and it instead raises a TypeError

schustmi (changes requested):
Amazing job @stefannica and @strickvl!

This was a lot to review, but the unified structure of all the methods like `_validate_{PROVIDER}_secret_name` and the well-designed helper methods made this really enjoyable to review. I only have some tiny questions and requests but overall this looks great!

Edit: And sorry for the late review, hopefully some of my requests can still make it in, otherwise most of them are actually more of a suggestion and/or something we can add in an upcoming release.

docs/book/getting-started/deploying-zenml/helm.md
stefannica: You're right, it was a copy-paste mishap.

docs/book/getting-started/deploying-zenml/upgrades.md
stefannica: Would you believe this was dragged in by the proverbial cat ? :)

src/zenml/cli/__init__.py
```diff
@@ -197,6 +197,176 @@
 zenml integration uninstall INTEGRATION_NAME
 ```
 
+Secrets Management
+------------------
+
+ZenML offers a way to securely store secrets associated with your other
+stack components and infrastructure. A ZenML Secret is a collection or grouping
+of key-value pairs securely stored by the ZenML secrets store.
```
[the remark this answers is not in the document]
stefannica: Ok, not securely then

pyproject.toml
stefannica: I was counting on pip install to catch something like this, but apparently I can put whatever I want here and it will just ignore it.

stefannica: This is really weird, it works on my machine in its current form:

```
pip install -e .[secrets-hashicorp]

...

  Attempting uninstall: zenml
    Found existing installation: zenml 0.34.0
    Uninstalling zenml-0.34.0:
      Successfully uninstalled zenml-0.34.0
Successfully installed hvac-1.0.2 zenml-0.34.0
```

I can't explain this.

stefannica: Ok, apparently, if you have optional packages in your pyproject.toml that are not tied to any "extras", they are installed by default when you ask for any extra, real or fake.

src/zenml/config/secrets_store_config.py
stefannica: This is a left-over from another iteration during which I kept all secrets store implementations inside integrations, but now this can go away.

src/zenml/zen_stores/secrets_stores/base_secrets_store.py
stefannica: Same as the root validator, a left-over from another time when secrets stores were located inside integrations. This can be unified now.

src/zenml/zen_stores/secrets_stores/hashicorp_secrets_store.py
```diff
@@ -0,0 +1,634 @@
...
+                error.
+        """
+        vault_secret_id = self._get_vault_secret_id(secret_id)
+
+        try:
+            vault_secret = (
+                self.client.secrets.kv.v2.read_secret(
+                    path=vault_secret_id,
+                )
+                .get("data", {})
```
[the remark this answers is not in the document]
stefannica: yup, they really have this double layer :)

src/zenml/zen_stores/secrets_stores/gcp_secrets_store.py
```diff
@@ -0,0 +1,612 @@
...
+            tzinfo=None, microsecond=0
+        )
+        labels = self._get_secret_metadata_for_secret(
+            secret=secret, secret_id=secret_id
+        )
+        labels[ZENML_GCP_SECRET_CREATED_KEY] = created.strftime(
```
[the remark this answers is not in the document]
stefannica: yes, GCP has extra restrictions around label values: we can't use any uppercase letters or semicolons in the value. We had to settle for a custom format because of this.

src/zenml/zen_stores/secrets_stores/gcp_secrets_store.py
stefannica: yes

src/zenml/enums.py
stefannica: Yes, this is possible, but you have to explicitly set the secrets store type to NONE in the configuration. Leaving the configuration empty defaults to whatever is used for the Zen store. It never gets as far as `get_store_class` when this is set to NONE because of this code in the `BaseZenStore.create_store` initialization method:

```        # Initialize the secrets store
        if (
            secrets_store_config
            and secrets_store_config.type != SecretsStoreType.NONE
        ):
            ...
```

src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
```diff
@@ -0,0 +1,816 @@
...
+            )
+
+            for page in pages:
+                for secret in page["SecretList"]:
+                    try:
+                        # NOTE: we do not include the secret values in the
+                        # response. We would need a separate API call to fetch
+                        # them for each secret, which would be very inefficient
+                        # anyway.
+                        secret_model = self._convert_aws_secret(
```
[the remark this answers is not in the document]
stefannica: Yes and yes.

src/zenml/zen_stores/secrets_stores/hashicorp_secrets_store.py
stefannica: No, they dont. The implementation removes them explicilty:

```
                # Remove the secret values from the response
                secret_model.values = {}
                results.append(secret_model)
```

src/zenml/zen_stores/secrets_stores/azure_secrets_store.py
stefannica: No, absoluely not. `list_secrets` never returns any secret values. This is not only documented in the API and Client methods and implemented consistently the same across all back-ends, it is also tested in the integration tests, and we ran and passed all tests on all of the back-ends.

strickvl (approved):
🤫🚦🥳

PR #1349, opened 2023-02-23
Fix broken docs URLs and add SDK docs url

strickvl:
This reimplements #1342, with the proper understanding of the task.

- Added an `sdk_docs_url` property method to all flavours plus a default way of generating these. (This doesn't solve for people using versions of ZenML prior to the existence of certain integrations, but I don't think worth the effort of identifying + catching all those).
- Updated the `docs_url` properties for integrations whose docs page names were renamed in #1187. At a certain point we'll want to retire this check, but for now it seems like a reasonable fix.

## Pre-requisites

## Types of changes


src/zenml/integrations/aws/flavors/aws_container_registry_flavor.py
```diff
@@ -73,7 +73,24 @@ def docs_url(self) -> Optional[str]:
...
+        from zenml import __version__
+
+        if parse(__version__) >= parse("0.34.0"):
+            return self.generate_default_docs_url()
+
+        old_docs_name = "amazon-ecr"
+        return self.generate_default_docs_url(component_name=old_docs_name)
+
+    @property
+    def sdk_docs_url(self) -> Optional[str]:
```
htahir1, line 87: Maybe im missing something but if this function exists in all flavors then why cant we just put it in the BaseFlavor class?

src/zenml/stack/flavor.py
```diff
@@ -43,6 +43,15 @@ def docs_url(self) -> Optional[str]:
...
         return None
 
+    @property
+    def sdk_docs_url(self) -> Optional[str]:
+        """A url to point at SDK docs explaining this flavor.
+
+        Returns:
+            A flavor SDK docs url.
+        """
+        return None
```
htahir1, line 53: Why dont we just do `return self.generate_default_sdk_docs_url()` here?

src/zenml/stack/flavor.py
```diff
@@ -139,20 +148,37 @@ def to_model(
...
         """
         from zenml import __version__
 
         component_type = self.type.plural.replace("_", "-")
         name = self.name.replace("_", "-")
+        docs_component_name = component_name or name
         base = f"https://docs.zenml.io/v/{__version__}"
-        url = f"{base}/component-gallery/{component_type}/{name}"
+        return (
+            f"{base}/component-gallery/{component_type}/{docs_component_name}"
```
htahir1, line 171: Isnt the URL directly generated from the file name rather than the `component_type` and `docs_component_name`? However, I do see the problem that the files are not accessible to the package... Therefore, we must make sure the filename conform to this

htahir1 (commented):
Looks good! Just left some potentially silly questions

src/zenml/integrations/aws/flavors/aws_container_registry_flavor.py
strickvl: Same was true for the `docs_url`, but that also wasn't in the `BaseFlavor`. You're right it probably should be, though.

src/zenml/stack/flavor.py
strickvl: I was basing what I did here on what was implemented for the `docs_url`. Presumed that there was maybe something about the frontend that required this? I'll check with @AlexejPenner when he's back on whether that's the case. Otherwise, I can update it for both here and just return a string and not an optional string.

src/zenml/stack/flavor.py
strickvl: I think #1187 addressed most of this renaming. I don't think there was anything that happened that would affect the API docs in this way, however.

src/zenml/integrations/aws/flavors/aws_container_registry_flavor.py
schustmi: This is true for our flavors, but not custom user flavors which subclass the same`Flavor` base class

src/zenml/stack/flavor.py
schustmi: Same as above, this breaks custom user flavors. I guess you could implement some check here to see if this is a zenml flavor and only then return it? However it will soon change with the hub anyway so I wouldn't spend too much time on it,

src/zenml/stack/flavor.py
htahir1: ok

schustmi (approved):

src/zenml/container_registries/gcp_container_registry.py
```diff
@@ -39,7 +39,24 @@ def docs_url(self) -> Optional[str]:
...
-        return self.generate_default_docs_url()
+        from packaging.version import parse
+
+        from zenml import __version__
+
+        if parse(__version__) >= parse("0.34.0"):
+            return self.generate_default_docs_url()
+
+        old_docs_name = "gcloud"
+        return self.generate_default_docs_url(component_name=old_docs_name)
```
stefannica, line 50: This code will never be executed. Might as well remove it.

src/zenml/integrations/aws/flavors/aws_container_registry_flavor.py
```diff
@@ -73,7 +73,24 @@ def docs_url(self) -> Optional[str]:
...
-        return self.generate_default_docs_url()
+        from packaging.version import parse
+
+        from zenml import __version__
+
+        if parse(__version__) >= parse("0.34.0"):
+            return self.generate_default_docs_url()
+
+        old_docs_name = "amazon-ecr"
+        return self.generate_default_docs_url(component_name=old_docs_name)
```
stefannica, line 84: Same here, this code will never be reached, because the changes will be released in 0.35.0.

src/zenml/integrations/aws/flavors/sagemaker_step_operator_flavor.py
```diff
@@ -91,7 +91,24 @@ def docs_url(self) -> Optional[str]:
...
-        return self.generate_default_docs_url()
+        from packaging.version import parse
+
+        from zenml import __version__
+
+        if parse(__version__) >= parse("0.34.0"):
+            return self.generate_default_docs_url()
+
+        old_docs_name = "amazon-sagemaker"
+        return self.generate_default_docs_url(component_name=old_docs_name)
```
stefannica, line 102: Code for the bit bucket.

src/zenml/integrations/azure/flavors/azure_artifact_store_flavor.py
```diff
@@ -53,7 +53,24 @@ def docs_url(self) -> Optional[str]:
...
-        return self.generate_default_docs_url()
+        from packaging.version import parse
+
+        from zenml import __version__
+
+        if parse(__version__) >= parse("0.34.0"):
+            return self.generate_default_docs_url()
+
+        old_docs_name = "azure-blob-storage"
+        return self.generate_default_docs_url(component_name=old_docs_name)
```
stefannica, line 64: Could be thrown away.

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -147,6 +147,15 @@ def docs_url(self) -> Optional[str]:
...
         return self.generate_default_docs_url()
 
+    @property
+    def sdk_docs_url(self) -> Optional[str]:
+        """A url to point at SDK docs explaining this flavor.
+
+        Returns:
+            A flavor SDK docs url.
+        """
+        return self.generate_default_sdk_docs_url()
```
stefannica, line 157: After seeing what seems like 100 copies of this, it seems very likely that this could be the base class implementation that checks whether a class is "builtin" or not and returns the default docs if it is.

stefannica (approved):
Good to merge, but could definitely have been done with fewer changes in the base class.

PR #1352, opened 2023-02-23
ZenML Hub Client & CLI Commands

fa9r:
## Describe changes
This PR adds several CLI commands to interact with the ZenML Hub:
- `zenml hub list`: List all plugins available on the Hub.
- `zenml hub install` Install a plugin from the Hub. Installed plugins can be imported via `from zenml.hub.<plugin_name> import ...`. Docker building then requires `required_hub_plugins` to be set in the `DockerSettings` of the pipeline.
- `zenml hub uninstall`: Uninstall an installed plugin.
- `zenml hub clone`: Clone the repository of a plugin (without installing it). Useful, e.g., for extending an existing plugin or for getting the examples of a plugin.
- `zenml hub submit`: Contribute a plugin to the Hub. Performs extensive client-side validation before submitting. Can be run in `--interactive` mode to ask and validate inputs one at a time. Requires login.
- `zenml hub login`: Login to the ZenML Hub. This sets `hub_token` of the active user in the DB.
- `zenml hub logout`: Logout by unsetting `hub_auth_token`.
- `zenml hub logs`: Get the build logs of a plugin.

All those commands interact with the hub via a new `zenml.hub.client.HubClient` class. The `__init__.py` files across the repos are configured such that hub plugins can be imported via `from zenml.hub.<plugin_name> import ...` but `zenml.hub.client` cannot be overridden.

## Pre-requisites

## Types of changes


fa9r:
@schustmi only requesting review of the `docker_settings` changes for now

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -493,6 +571,10 @@ def _generate_zenml_pipeline_dockerfile(
                 f"RUN pip install --default-timeout=60 --no-cache-dir -r {file}"
             )
 
+        for index, packages in extra_index_requirements.items():
```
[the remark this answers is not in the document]
schustmi: I think in general having the hub plugins in a requirements.txt (maybe some lines of which have an index specifier) would be much nicer as you can then directly look at that file from within the container

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -454,13 +474,68 @@ def _gather_requirements_files(
...
+            if isinstance(required_plugin, str):
+                plugin_name = required_plugin
+                plugin_version = "latest"
+            else:
+                plugin_name, plugin_version = required_plugin
+            plugin = _get_plugin(plugin_name, plugin_version)
+            if plugin and plugin.index_url and plugin.wheel_name:
+                hub_packages[plugin.index_url].append(plugin.wheel_name)
+                if plugin.requirements:
+                    hub_requirements.extend(plugin.requirements)
```
schustmi, line 522: What if the plugin requires another plugin?

schustmi (changes requested):

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -326,32 +329,49 @@ def _add_requirements_files(
...
-        requirements_file_names = []
+        requirements_file_names: List[str] = []
+        extra_index_reqs: Dict[str, List[str]] = {}
+        hub_requirements: List[str] = []
+
+        if docker_settings.required_hub_plugins:
+            hub_packages, hub_requirements = cls._get_hub_requirements(
+                docker_settings.required_hub_plugins
+            )
+            extra_index_reqs = {**extra_index_reqs, **hub_packages}
```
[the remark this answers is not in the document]
fa9r: cool didn't know `dict.update` exists, neat 👍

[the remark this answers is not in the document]
fa9r: as for the second suggestion, I think updating is more extensible than overriding

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -326,32 +329,49 @@ def _add_requirements_files(
...
             build_context.add_file(source=file_content, destination=filename)
             requirements_file_names.append(filename)
 
-        return requirements_file_names
+        return requirements_file_names, extra_index_reqs
 
     @staticmethod
     def _gather_requirements_files(
-        docker_settings: DockerSettings, stack: "Stack", log: bool = True
+        docker_settings: DockerSettings,
```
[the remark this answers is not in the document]
fa9r: Mhhh a bit odd that it's a `_...` method then which is usually used to indicate internal methods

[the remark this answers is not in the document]
fa9r: Also, `hub_requirements` is optional and the method won't do anything differently if that arg is not provided, so why would this break any of the step operators?

src/zenml/utils/pipeline_docker_image_builder.py
fa9r: this is not possible at the moment in general, we do have an open ticket about supporting this but it's probably not high enough prio to make it into the MVP release

src/zenml/utils/pipeline_docker_image_builder.py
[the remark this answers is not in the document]
fa9r: Unfortunately `--extra-index-url` is unsafe since there is no way to tell pip which index to prefer, i.e., if we use a private repo, someone could register the same package name in PyPi and then `zenml hub install ...` would install that package by default if the version is higher than the one on the hub. It's really stupid actually. See https://github.com/pypa/pip/issues/8606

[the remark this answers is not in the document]
fa9r: I also don't see how you can specify `--index-url` inside a requirements file from the link you sent, all I see is 
> "The following options have an effect on the entire pip install run, and must be specified on their individual lines. [-i, --index-url](https://pip.pypa.io/en/stable/cli/pip_install/#install-index-url) ..."
which doesn't sound requirement-specific. I guess you could specify the full URL to the wheel but I'm not a big fan of that.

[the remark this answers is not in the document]
fa9r: I guess we could write all the plugins themselves into another requirements file and then install that via `pip install -r ... --index-url ...`. Would you prefer that option?

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -311,12 +314,12 @@ def _get_target_image_name(
         return target_image_name
 
     @classmethod
-    def _add_requirements_files(
+    def _get_requirements(
```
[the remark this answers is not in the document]
fa9r: Alright, changed it back

src/zenml/config/docker_settings.py
```diff
@@ -149,20 +155,16 @@ class DockerSettings(BaseSettings):
...
     skip_build: bool = False
-
     target_repository: str = "zenml"
-
     replicate_local_python_environment: Optional[
         Union[List[str], PythonEnvironmentExportMethod]
     ] = None
     requirements: Union[None, str, List[str]] = None
     required_integrations: List[str] = []
+    required_hub_plugins: List[Union[str, Tuple[str, str]]] = []
```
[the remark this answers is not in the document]
fa9r: Yeah makes sense, changed it accordingly 👍

src/zenml/utils/pipeline_docker_image_builder.py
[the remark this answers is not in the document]
schustmi: AzureML uses a separate build system where they build the image, this means that we need to pass them the hub requirements to install which will now be missing I guess?

src/zenml/utils/pipeline_docker_image_builder.py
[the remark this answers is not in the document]
schustmi: Mhm you're right, apparently it's a per-file settings, my bad. I'd still prefer two requirements files tbh, and the hub one could even have the `--index-url` inside the file so we can treat them the same from the outside?

src/zenml/constants.py
```diff
@@ -63,6 +63,7 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 ENV_ZENML_LOGGING_VERBOSITY = "ZENML_LOGGING_VERBOSITY"
 ENV_ABSL_LOGGING_VERBOSITY = "ZENML_ABSL_LOGGING_VERBOSITY"
 ENV_ZENML_REPOSITORY_PATH = "ZENML_REPOSITORY_PATH"
+ENV_ZENML_HUB_URL = "ZENML_HUB_URL"
```
fa9r, line 66:
```suggestion
```

src/zenml/utils/pipeline_docker_image_builder.py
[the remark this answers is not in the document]
fa9r: Sounds good to me, I'll give that a try

src/zenml/utils/pipeline_docker_image_builder.py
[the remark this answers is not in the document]
fa9r: Ah okay so it doesn't break step operators in general, the issue is just that they might not have some required hub plugins making the run fail if you try to use some code from the hub within a step operator?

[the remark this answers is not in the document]
fa9r: Anyways, I'll have a look how we could refactor this to keep this particular method unchanged, should work fine to just move all the plugin install logic into here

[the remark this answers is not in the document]
fa9r: I also renamed the method to `gather_requirements_files` to indicate that this is not just an internal method of the `DockerImageBuilder`.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+
+    Args:
+        plugin_name: Name of the plugin.
+        version: Version of the plugin. If not specified, the latest version
+            will be used.
+    """
+    with event_handler(
+        event=AnalyticsEvent.ZENML_HUB_PLUGIN_UNINSTALL,
+        metadata={
+            "plugin_name": plugin_name,
```
bcdurak, line 275: Same as above.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+    "--version",
+    "-v",
+    type=str,
+    help="Version of the plugin to uninstall.",
+)
+def uninstall_plugin(
+    plugin_name: str,
+    version: Optional[str] = None,
+) -> None:
+    """Uninstall a plugin from the hub.
```
bcdurak, line 265: I think the docstring is a bit misleading here. This command uninstalls it from your environment, not the hub.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+        repository_commit: Commit to checkout from the repository. Overrides
+            `repository_branch`.
+        tags: Tags to add to the plugin.
+        interactive: Whether to run the command in interactive mode, asking for
+            missing or invalid parameters.
+    """
+    with event_handler(
+        event=AnalyticsEvent.ZENML_HUB_PLUGIN_SUBMIT,
+        metadata={
+            "plugin_name": plugin_name,
```
bcdurak, line 632: Same as above.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+        version: Version of the plugin. If not specified, the latest version
+            will be used.
+        output_dir: Output directory to clone the plugin to. If not specified,
+            the plugin will be cloned to a directory with the same name as the
+            plugin in the current working directory.
+    """
+    with event_handler(
+        event=AnalyticsEvent.ZENML_HUB_PLUGIN_CLONE,
+        metadata={
+            "plugin_name": plugin_name,
```
bcdurak, line 340: Same as above.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+
+    Args:
+        url: The URL to the repository to clone.
+        commit: The commit to checkout.
+        branch: The branch to checkout. Will be ignored if commit is provided.
+        subdir: The subdirectory in which the plugin is located.
+        interactive: Whether to run in interactive mode.
+
+    Returns:
+        - The validated URL to the repository,
```
bcdurak, line 792: One of the returns is missing.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+
+    Args:
+        plugin_name: Name of the plugin.
+        version: Version of the plugin. If not provided, the latest version
+            will be used.
+    """
+    with event_handler(
+        event=AnalyticsEvent.ZENML_HUB_PLUGIN_LOGS,
+        metadata={
+            "plugin_name": plugin_name,
```
bcdurak, line 1038: Same as above.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+    module_name = _get_plugin_module(plugin)
+    try:
+        spec = find_spec(module_name)
+        return spec is not None
+    except ModuleNotFoundError:
+        return False
+
+
+def _get_plugin_module(plugin: HubPluginResponseModel) -> str:
+    """Helper function to get the module name of a plugin."""
```
bcdurak, line 1102: Missing args and returns in the docsting.

bcdurak (commented):
Great additions to our CLI. Really looking forward to using the new commands myself. I have left a few small comments.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1164 @@
...
+            "MODULE": _get_plugin_module(plugin),
+            "PACKAGE NAME": plugin.package_name or "",
+            "REPOSITORY URL": plugin.repository_url,
+        }
+        plugins_table.append(display_data)
+    return plugins_table
+
+
+@hub.command("install")
+@click.argument("plugin_name", type=str, required=True)
```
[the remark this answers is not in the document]
fa9r: Yep makes sense, that's also how you would specify it in the `required_hub_plugins` Docker settings, so changing it as you proposed also make that more consistent 👍

src/zenml/__init__.py
```diff
@@ -23,3 +23,8 @@
 from zenml.logger import init_logging  # noqa
 
 init_logging()
+
+# The following code is needed for `zenml.hub` subpackages to be found
+from pkgutil import extend_path
+
+__path__ = extend_path(__path__, __name__)
```
schustmi, line 30: Is this necessary both here and in the `hub/__init__.py`?

src/zenml/utils/analytics_utils.py
```diff
@@ -136,6 +136,17 @@ class AnalyticsEvent(str, Enum):
     ZENML_SERVER_DEPLOYED = "ZenML server deployed"
     ZENML_SERVER_DESTROYED = "ZenML server destroyed"
 
+    # ZenML Hub events
+    ZENML_HUB_PLUGIN_LIST = "ZenML Hub plugins listed"
```
schustmi, line 140: I'll be the devils advocate and ask: Why do we track this? We removed all the `list_...` events recently if I remember correctly because no one ever took a look at them in our analytics?

src/zenml/models/hub_plugin_models.py
```diff
@@ -0,0 +1,76 @@
...
+from datetime import datetime
+from typing import TYPE_CHECKING, List, Optional
+from uuid import UUID
+
+from pydantic import BaseModel
+
+from zenml.utils.enum_utils import StrEnum
+
+if TYPE_CHECKING:
+    pass
```
schustmi, line 25:
```suggestion
```

src/zenml/models/hub_plugin_models.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class PluginStatus(StrEnum):
+    """Enum that represents the status of a plugin."""
+
+    PENDING = "pending"  # Plugin is being built
```
schustmi, line 31: I think those comments might be nice in the docstring so they should up in the API docs?

src/zenml/models/hub_plugin_models.py
```diff
@@ -0,0 +1,76 @@
...
+class HubPluginBaseModel(BaseModel):
+    """Base model for a ZenML Hub plugin."""
+
+    name: str
+    description: Optional[str]
+    version: Optional[str]
+    release_notes: Optional[str]
+    repository_url: str
+    repository_subdirectory: Optional[str]
+    repository_branch: Optional[str]
```
schustmi, line 54: With the recent mypy changes, I think these should all be `Optional[...] = None`.

Also: can both `commit` and `branch` be set at the same time? What does branch even mean in this case, I would expect it to build a new plugin if the branch gets updated.

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -465,8 +482,106 @@ def _gather_requirements_files(
                     ", ".join(f"`{r}`" for r in integration_requirements_list),
                 )
 
+        # Generate requirements files for all ZenML Hub plugins
+        if docker_settings.required_hub_plugins:
+            (
+                hub_packages,
```
schustmi, line 488: This code was really confusing to me because I couldn't really get what `hub_packages` and `hub_requirements` refer to until I finished reading all the code. Are there some more intuitive names we could use, e.g. `hub_requirements` and `pypi_requirements` or something along those lines?

src/zenml/hub/client/client.py
```diff
@@ -0,0 +1,275 @@
...
+        self.auth_token = Client().active_user.hub_token
+
+    @staticmethod
+    def get_default_url() -> str:
+        """Get the default URL of the ZenML Hub.
+
+        Returns:
+            The default URL of the ZenML Hub.
+        """
+        return os.getenv(ENV_ZENML_HUB_URL, default="https://hub.zenml.io/")
```
schustmi, line 60: Maybe useful to have a constant for this as well?

src/zenml/hub/client/client.py
```diff
@@ -0,0 +1,275 @@
...
+        raise HubAPIError(f"Unexpected response: {str(response)}")
+
+    def get_me(self) -> Optional[HubUserResponseModel]:
+        """Get the current user.
+
+        Returns:
+            The user response model or None if the user does not exist.
+        """
+        try:
+            response = self._request("GET", "/user/users/me")
```
schustmi, line 217: `/user/users/me`?

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+    is_flag=True,
+    help="List only plugins that you own.",
+)
+@click.option(
+    "--installed",
+    "-i",
+    is_flag=True,
+    help="List only plugins that are installed.",
+)
+def list_plugins(all: bool, mine: bool, installed: bool) -> None:
```
schustmi, line 69: A default value for all those flags would increase the readability here I think

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+    Returns:
+        The username of the plugin author or "" if no username was found.
+    """
+    if plugin.user:
+        if plugin.user.username:
+            return plugin.user.username
+    return ""
+
+
+def _parse_plugin_name(plugin_name: str) -> Tuple[str, Optional[str]]:
```
schustmi, line 1165: Something similar is done in the `PipelineDockerImageBuilder` (including the version), can we re-use the code somehow? In general this CLI file contains a lot of utility functions which I assume are useful in different parts of our code, it probably makes sense to have those in some separate file?

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+        if not plugin.build_logs:
+            logger.info(
+                f"Plugin '{display_name}' finished building, but no logs "
+                "were found."
+            )
+            return
+
+        for line in plugin.build_logs.splitlines():
+            if line.startswith("DEBUG"):
+                pass
```
schustmi, line 1104: Why is this not logged as a debug log here? Shouldn't we allow users to set the verbosity of what they want to see in these logs?

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+    if not tags:
+        if not interactive:
+            return []
+
+        # In interactive mode, ask for tags if none were provided.
+        return _ask_for_tags()
+
+    # If tags were provided, print a warning if any of them is invalid.
+    for tag in tags:
+        if tag.startswith(ZENML_HUB_INTERNAL_TAG_PREFIX):
```
schustmi, line 1014: I'm assuming that this check also exists serverside to make sure anyone modifying/bypassing our CLI can't use that tag?

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+        if output_dir is None:
+            output_dir = os.path.join(os.getcwd(), plugin_name)
+        logger.info(f"Cloning plugin '{display_name}' to {output_dir}...")
+        # If no subdir is set, we can clone directly into output_dir
+        if not subdir:
+            repo_path = plugin_dir = output_dir
+        # Otherwise, we need to clone into a random dir and then move the subdir
+        else:
+            random_repo_name = f"_{uuid4()}"
+            repo_path = os.path.abspath(
```
schustmi, line 392: This seems like a tempdir to me, not a random string in the current working directory

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+        except GitCommandError:
+            error(
+                f"Could not find commit '{commit}' in repository '{repo_url}' "
+                f"of plugin '{display_name}'. This might happen if the owner "
+                "of the plugin has force-pushed to the plugin repository. "
+                "Please report this plugin version in the ZenML Hub or via "
+                "Slack."
+            )
+        finally:
+            shutil.rmtree(repo_path, ignore_errors=True)
```
schustmi, line 411: If `plugin.subdirectory` is `None`, this will delete the just-cloned repo right?

src/zenml/utils/analytics_utils.py
```diff
@@ -136,6 +136,17 @@ class AnalyticsEvent(str, Enum):
...
+    ZENML_HUB_LOGIN = "ZenML Hub login"
+    ZENML_HUB_LOGIN_GH = "ZenML Hub login via GitHub"
```
schustmi, line 142: Along the same lines: Login/Logout? If anything, these should be tracked serverside. But why would I even care if someone logs out, I really don't get it.

src/zenml/utils/analytics_utils.py
```diff
@@ -136,6 +136,17 @@ class AnalyticsEvent(str, Enum):
...
+    ZENML_HUB_LOGOUT = "ZenML Hub logout"
+    ZENML_HUB_PLUGIN_INSTALL = "ZenML Hub plugin installed"
+    ZENML_HUB_PLUGIN_UNINSTALL = "ZenML Hub plugin uninstalled"
+    ZENML_HUB_PLUGIN_CLONE = "ZenML Hub plugin pulled"
+    ZENML_HUB_PLUGIN_SUBMIT = "ZenML Hub plugin pushed"
+    ZENML_HUB_PLUGIN_LOGS = "ZenML Hub plugin logs viewed"
```
schustmi, line 148: Same as the others: Why? Does this give us any useful information? It's not like we can get rid of the functionality to view the build logs if they happen serverside, it's mandatory to debug. Tracking stuff like `% of failed builds` which is interesting again should happen serverside where the actual build is happening.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+        me = client.get_me()
+        if not me:
+            error("Could not retrieve user information from the ZenML Hub.")
+        if not me.username:
+            error(
+                "Your ZenML Hub account does not have a username yet. Please "
+                "set a username in your account settings and try again."
+            )
+        username = me.username
+        analytics_handler.metadata["plugin_author"] = username
```
schustmi, line 687: https://docs.zenml.io/misc/usage-analytics

In order to help us better understand how the community uses ZenML, the pip package reports **anonymized** usage statistics.

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+
+        # Validate the plugin name and check if it exists
+        plugin_name, plugin_exists = _validate_plugin_name(
+            client=client,
+            plugin_name=plugin_name,
+            username=username,
+            interactive=interactive,
+        )
+        analytics_handler.metadata["plugin_name"] = plugin_name
```
schustmi, line 696: Same as above, this is not tracking anonymized data.

schustmi (changes requested):

src/zenml/models/hub_plugin_models.py
```diff
@@ -0,0 +1,76 @@
...
+    pass
+
+
+class PluginStatus(StrEnum):
+    """Enum that represents the status of a plugin."""
+
+    PENDING = "pending"  # Plugin is being built
+    FAILED = "failed"  # Plugin build failed
+    AVAILABLE = "available"  # Plugin is available for installation
+    YANKED = "yanked"  # Plugin was yanked and is no longer available
```
schustmi, line 34: Can users (and if yes, how? I didn't see a CLI command for that) yank a plugin? Have you thought about the implications of that? Me as the user of a popular plugin can potentially mess with production pipelines of others using that plugin?

schustmi (changes requested):

schustmi (changes requested):
In addition to my other comments, I think there are lots of functions that can and should be tested. Especially the verification of repository structures and CLI functions (probably need to mock the `HubClient` for that) seem like good candidates as it's often very complex code with lots of `if` conditions.

src/zenml/__init__.py
fa9r: apparently yes but don't ask me why 😄

src/zenml/cli/hub.py
fa9r: yes it is, how would you like it to be changed then? Are you suggesting I use some existing temp dir utils we have or should I simply change the variable name?

src/zenml/cli/hub.py
fa9r: It should always delete the repo, the idea here is we:
- clone the repo to a temp dir
- move the plugin dir (either repo_root or repo_root/subdir) to the output_dir
- delete the temp dir

src/zenml/cli/hub.py
fa9r: oh yes, fair enough, I'll remove the author tracking for plugin create. 

The other CLI commands also have the plugin author in the analytics metadata (since the plugin name alone is not unique), I guess that is fine since it is information of the plugin that is interacted with and not information of the active user at that point?

src/zenml/cli/hub.py
fa9r: good catch 👍

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+                    commit = click.prompt("Repository commit")
+                    branch = None
+                else:
+                    logger.info("Please enter the name of a branch.")
+                    branch = click.prompt("Repository branch")
+                    commit = None
+
+        try:
+            # Check if the URL/branch/commit are valid.
+            repo_path = f"_{uuid4()}"
```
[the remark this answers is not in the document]
fa9r: same as above, what would you like me to do here instead?

src/zenml/__init__.py
schustmi: Why?

src/zenml/cli/hub.py
schustmi: Yep, I guess the usage of a plugin (e.g. downloads/installs) is even something that's publicly displayed right?

src/zenml/cli/hub.py
schustmi: But if `plugin.subdirectory` is `None`, this will also delete the `output_dir` right? Which I'm assuming will be used downstream?

A little above:
```python
if not subdir:
  repo_path = plugin_dir = output_dir
```

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+            _validate_repository_structure(plugin_path)
+            return repository_subdir
+        except ValueError as e:
+            msg = (
+                f"Plugin code structure at {repository_subdir} is invalid: "
+                f"{str(e)}"
+            )
+            if not interactive:
+                error(str(e))
+            logger.info(msg)
```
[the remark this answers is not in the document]
fa9r: good point, I replaced all `logger....` calls with cli `declare` / ...

src/zenml/cli/hub.py
fa9r: yep we have the same checks server-side 👍

src/zenml/cli/hub.py
schustmi:
```python
import tempfile

with tempfile.TemporaryDirectory(...):
  ...
```

schustmi: While not likely, your current approach could still crash if it doesn't get to the finally clause (which might just be a `CTRL-C` from the user at an unfortunate moment.

src/zenml/cli/hub.py
fa9r: I replaced these with cli utils declare / warning / error already. Actually, we won't need the "DEBUG" and "WARNING" checks right now since we currently only log info/error there

src/zenml/cli/hub.py
```diff
@@ -0,0 +1,1203 @@
...
+def _get_plugin_author_username(plugin: HubPluginResponseModel) -> str:
+    """Helper function to get the username of the author of a plugin.
+
+    Args:
+        plugin: The plugin.
+
+    Returns:
+        The username of the plugin author or "" if no username was found.
+    """
+    if plugin.user:
```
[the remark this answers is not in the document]
fa9r: That's what we thought initially, but actually, it can never be `None` now since ZenML official plugins also have an admin user account associated with it. This still needs to be fixed both here and in the hub, thanks for pointing out 👍

src/zenml/cli/hub.py
fa9r: yes already moved this and the util function below into a new `zenml.hub.internal.utils` module

src/zenml/hub/client/client.py
fa9r: already fixed

src/zenml/models/hub_plugin_models.py
fa9r: makes sense, done

src/zenml/cli/hub.py
schustmi: Btw, what if my username is `internal` and my plugin called `utils`? Should we maybe have a separate namespace `zenml._hub` where we have our internal implementations to avoid any potential conflicts?

src/zenml/models/hub_plugin_models.py
fa9r: at the moment, yanking is not possible but we will support this in the future. The idea here would be that yanking simply "hides" this version when listing/browsing plugins, but the wheel itself will still exist

src/zenml/models/hub_plugin_models.py
fa9r: yeah, already changed the types.

As for `commit` and `branch`, `branch` can only be set at creation time will be ignored if `commit` is set. When the plugin is being built, we also override `commit` according to the actual repo commit that we built the plugin from, so `commit` will always be set for plugins that were built correctly, but might be missing for plugins where the build failed or is still in progress.

src/zenml/utils/analytics_utils.py
fa9r: yeah fair enough, I just added analytics to every single command for now. I wouldn't mind removing list plugin / login / logs. @bcdurak any opinions on this?

src/zenml/utils/pipeline_docker_image_builder.py
fa9r: I renamed them to `hub_pypi_requirements` and `hub_internal_requirements`, is that more clear?

src/zenml/__init__.py
fa9r: Well because that is how package resolving works in Python, see [the offical Python docs]( https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) for reference

src/zenml/cli/hub.py
fa9r: `internal` and `client` won't get resolved since we already have them as packages in `src/zenml/hub` right now, i.e., if you give yourself username `internal`, your plugins can never be imported.

fa9r: but yes makes sense to rename our `zenml.hub` to `zenml._hub` to avoid headaches down the line

fa9r: done

src/zenml/utils/analytics_utils.py
bcdurak: Unless any of us can think of a meaningful Insights table regarding the usage of this CLI command, I think, we can remove the event.

src/zenml/cli/hub.py
fa9r: ah cool, thanks, changed it accordingly 👍

src/zenml/cli/hub.py
fa9r: ah yes, I missed the part you highlighted. Should also be fixed in the new version now.

fa9r:
As for tests, I would suggest waiting till the Hub API itself is stable. Not much point in testing the interaction with an unstable untested API IMO.

src/zenml/cli/hub.py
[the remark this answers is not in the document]
fa9r: The model now no longer has `plugin.user` on it, but instead a non-optional `plugin.author` which is the username of the user that created the plugin.

src/zenml/utils/analytics_utils.py
fa9r: okay then let's get rid of them

src/zenml/__init__.py
schustmi: Hahah 😄

htahir1: Alt Text

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -465,8 +483,99 @@ def _gather_requirements_files(
...
+                package names.
+            - A list of all unique dependencies of the required hub plugins
+                (which can be installed from PyPI).
+        """
+        from zenml._hub.client import HubClient
+        from zenml._hub.utils import parse_plugin_name, plugin_display_name
+
+        client = HubClient()
+
+        _DICT_TYPE = Union[Dict[str, List[str]], DefaultDict[str, List[str]]]
```
schustmi, line 545: Wow that is ugly :D

Why is this necessary, so you can return a `Dict` instead of `DefaultDict`? Maybe if we just do `return dict(hub_internal_requirements)` instead of re-assigning `hub_internal_requirements=dict(hub_internal_requirements)` we can avoid that?

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -334,27 +345,34 @@ def _add_requirements_files(
...
-    def _gather_requirements_files(
+    def gather_requirements_files(
         docker_settings: DockerSettings,
         stack: "Stack",
         code_repository: Optional["BaseCodeRepository"] = None,
         log: bool = True,
     ) -> List[Tuple[str, str]]:
         """Gathers and/or generates pip requirements files.
 
+        This method is used during `DockerImageBuilder.build_docker_image` but
```
schustmi, line 371:
```suggestion
        This method is used during `PipelineDockerImageBuilder.build_docker_image` but
```

src/zenml/models/hub_plugin_models.py
schustmi: Wouldn't it be cleaner to have the `branch` only on the request model in that case?

schustmi (changes requested):

schustmi:
Regarding the tests: I still think it would be useful them as lots of them are not really related to the Hub API I think, e.g. verifying that the repository structure check works both with and without a subdirectory supplied. However I think that should not be a blocker to get this out for the release, we can do them in some follow up PR.
We should just be aware that our manual tests and reviews probably don't catch all the bugs/corner cases.

src/zenml/models/hub_plugin_models.py
fa9r: Client-side yes, but the models that we have here are identical to the models we have server-side and I would like to keep them in sync to avoid unnecessary complexity.

Server-side we need the branch on the response model as well because it currently works like this:
- client submits plugin -> validate plugin -> write plugin to DB with status "pending" -> return corresponding response model
- plugin build is triggered with response model as input -> plugin is built -> plugin is updated in DB with status "failed" or "available" (and only now the branch is not needed anymore)

src/zenml/utils/pipeline_docker_image_builder.py
fa9r: Yeah I agree, changed it as suggested 👍

schustmi (approved):
Can't wait to start using this!

PR #1361, opened 2023-02-27
Add failure (and success hooks)

htahir1:
## Describe changes
This PR adds hooks to ZenML pipelines that can be defined like this:

```
def on_failure_hook(context: StepContext, params: MyParams):  # both args optional
    print("Call hook")


@step(on_failure=on_failure_hook)
def get_first_num() -> Output(first_num=int):
    """Returns an integer."""
    raise ValueError("Error")
    return 10
```

Here is a video to describe the feature

https://www.loom.com/share/7989ab86940646969600e1a27bcf1c29

## Pre-requisites

## Types of changes


bcdurak (approved):
LGTM!

fa9r (commented):
Design looks good, just needs a bit of polish from @schustmi then it should be good to go 👍 💯

src/zenml/orchestrators/step_runner.py
```diff
@@ -120,6 +120,28 @@ def run(
...
+                            args=hook_spec.args,
+                            annotations=hook_spec.annotations,
+                            input_artifacts=input_artifacts,
+                            output_artifact_uris=output_artifact_uris,
+                            output_materializers=output_materializers,
+                        )
+                        on_failure(**function_params)
+                    except ImportError as e:
+                        raise ValueError(
+                            f"ZenML can not import the config class '{on_fail_source}': {e}"
```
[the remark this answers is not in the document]
htahir1: Raise original exception as well

src/zenml/orchestrators/step_runner.py
```diff
@@ -120,6 +120,28 @@ def run(
                 return_values = step_entrypoint(**function_params)
             except:  # noqa: E722
                 step_failed = True
+                on_fail_source = self.configuration.on_failure
+                if on_fail_source:
+                    try:
+                        # For now, abusing _parse_inputs to get function_params
```
htahir1, line 126: Fix this and refactor

htahir1 (commented):
Review from Michael:

* <del> Add Exception to the on_success hook</del>
* <del> Unit test that it gets called </del>
* <del> Potetially add success hook </del>
* <del> Rename to hooks everywhere </del>
* <del> Add standard alerter implementation </del>
* <del> Add pipeline name and run name into the Context </del>

* Validation: If its a function type, no need to validate it. If its a string, we need to check its a function type. We should also check if the argument space is good
* Add docs and example
* Make it a union of string and function type types.functiontype instead of callable
* Add docstring that explains the signature of the hook in the step decorator and the configure method

src/zenml/config/step_configurations.py
```diff
@@ -69,6 +69,7 @@ class StepConfigurationUpdate(StrictBaseModel):
     parameters: Dict[str, Any] = {}
     settings: Dict[str, BaseSettings] = {}
     extra: Dict[str, Any] = {}
+    on_failure: Optional[str] = None
```
schustmi, line 72:
```suggestion
    failure_hook_source: Optional[str] = None
```
seems better to me, also needs some changing elsewhere

examples/failure_and_success_hooks/run.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class HookParams(BaseParameters):
+    """Parameters to define which hook to trigger"""
+
+    fail: bool = True
+
+
+@step(on_failure=alerter_failure_hook, on_success=alerter_success_hook)
+def get_first_num(params: HookParams) -> int:
```
schustmi, line 33: All our examples follow the recommended structure of having a `steps` and `pipelines` subdirectory which define those classes, I think we should stay consistent and also do that for this example

src/zenml/pipelines/pipeline_decorator.py
```diff
@@ -80,6 +87,15 @@ def pipeline(
         enable_artifact_metadata: Whether to enable artifact metadata or not.
         settings: Settings for this pipeline.
         extra: Extra configurations for this pipeline.
+        on_failure: Callback function in event of failure of the step. Can be
+            a function with three possible parameters,
+            `StepContext`, `BaseParameters`, and `Exception`,
+            or a UDF path to a function of the same specifications
```
schustmi, line 93: What is a UDF path?

src/zenml/hooks/hook_validators.py
```diff
@@ -0,0 +1,64 @@
...
+        # This means hook is a callable
+        func = hook_func
+
+    if not callable(func):
+        raise ValueError(f"{func} is not a valid function.")
+
+    from zenml.steps.base_parameters import BaseParameters
+    from zenml.steps.step_context import StepContext
+
+    if func.__annotations__:
```
schustmi, line 50: I think this validation could go one step further:
- There must be annotations for each argument (otherwise we will just not pass them and it will fail at runtime)
- There can only be these three arguments, each type exactly once. Again, this would otherwise fail during runtime which we want to avoid, specifically if our success hook crashes the pipeline.

src/zenml/orchestrators/step_runner.py
```diff
@@ -415,3 +509,38 @@ def _store_output_artifacts(
...
+                args=hook_spec.args,
+                annotations=hook_spec.annotations,
+                step_exception=step_exception,
+                output_artifact_uris=output_artifact_uris,
+                output_materializers=output_materializers,
+            )
+            logger.debug(f"Running hook {hook} with params: {function_params}")
+            hook(**function_params)
+        except (ImportError, TypeError, ValueError) as e:
+            raise ValueError(
```
schustmi, line 544: This `try...except` should only be around the actual `load_source_path` call if it is to catch the loading exceptions.
IMO we want a general
```python
try:
  load_and_run_hook()
except:
  logger.error("Failed to run hook")
```

here instead though. Failure to run a hook should never actually lead to a failure of the pipeline

src/zenml/hooks/alerter_hooks.py
```diff
@@ -0,0 +1,83 @@
...
+    message += f"Run name: `{context.run_name}`" + "\n"
+    message += f"Step name: `{context.step_name}`" + "\n"
+    message += f"Exception: `({type(exception)}) {exception}`" + "\n\n"
+    message += (
+        f"Step Cache Enabled: `{'True' if context.cache_enabled else 'False'}`"
+        + "\n"
+    )
+
+    if context.stack and context.stack.alerter:
+        context.stack.alerter.post(message)
```
schustmi, line 50: This if check should be at the top to even avoid computing the message if no alerter exists.

src/zenml/hooks/alerter_hooks.py
```diff
@@ -0,0 +1,83 @@
...
+    else:
+        logger.warning(
+            "Specified standard failure hook but no alerter configured in the stack!. Skipping.."
```
schustmi, line 53:
```suggestion
            "Specified standard failure hook but no alerter configured in the stack. Skipping.."
```

src/zenml/hooks/alerter_hooks.py
```diff
@@ -0,0 +1,83 @@
...
+    message += f"Step name: `{context.step_name}`" + "\n"
+    message += (
+        f"Step Cache Enabled: `{'True' if context.cache_enabled else 'False'}`"
+        + "\n"
+    )
+    if context.stack and context.stack.alerter:
+        context.stack.alerter.post(message)
+    else:
+        logger.warning(
+            "Specified standard success hook but no alerter configured in the stack!. Skipping.."
```
schustmi, line 82:
```suggestion
            "Specified standard success hook but no alerter configured in the stack. Skipping.."
```

src/zenml/hooks/alerter_hooks.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    message += f"Pipeline name: `{context.pipeline_name}`" + "\n"
+    message += f"Run name: `{context.run_name}`" + "\n"
+    message += f"Step name: `{context.step_name}`" + "\n"
+    message += (
+        f"Step Cache Enabled: `{'True' if context.cache_enabled else 'False'}`"
+        + "\n"
+    )
+    if context.stack and context.stack.alerter:
+        context.stack.alerter.post(message)
```
schustmi, line 79: Again the check should be above IMO

src/zenml/hooks/alerter_hooks.py
```diff
@@ -0,0 +1,83 @@
...
+def alerter_failure_hook(
+    context: "StepContext", params: "BaseParameters", exception: Exception
+) -> None:
+    """Standard failure hook that executes after step fails.
+
+    This hook uses any `BaseAlerter` that is configured within the active stack to post a message.
+
+    Args:
+        context: Context of the step.
+        params: Parameters used in the step.
```
schustmi, line 36: Should we print them as well or remove the argument?

schustmi: Same for the success hook

schustmi (changes requested):
Great feature! Had a few comments and I think this could benefit from an additional test that verifies the pipeline hook feature and merging

src/zenml/pipelines/pipeline_decorator.py
htahir1: user-defined function: https://en.wikipedia.org/wiki/User-defined_function

htahir1:
Thanks for the review @schustmi ! I actually wanted to write that test but I dont know how. Could you help me out here? Will work on your other comments in the meanwhile

src/zenml/pipelines/pipeline_decorator.py
schustmi: Ah I see, yes that needs more explanation. To be consistent with other places, we might also just want to call it `source path` or `source`

schustmi:
`tests/unit/config/test_compiler.py::test_extra_merging` should be a good example that you can almost copy-paste to test the merging.

As for running a pipeline with a hook, I guess your existing tests could be copied and just use `pipeline.configure` instead of `step.configure` and check whether the hook got called?

htahir1:
@schustmi I have taken all your suggestions :-) Ready for review again. Thanks!

src/zenml/pipelines/pipeline_decorator.py
```diff
@@ -80,6 +87,15 @@ def pipeline(
         enable_artifact_metadata: Whether to enable artifact metadata or not.
         settings: Settings for this pipeline.
         extra: Extra configurations for this pipeline.
+        on_failure: Callback function in event of failure of the step. Can be
+            a function with three possible parameters,
+            `StepContext`, `BaseParameters`, and `Exception`,
+            or a source path to a function of the same specifications
+            (e.g. `module.my_file.my_function`).
```
schustmi, line 94: Tiny nit, but I would suggest removing the `my_file` part of the path for all of these, as file==module in this case

src/zenml/orchestrators/step_runner.py
```diff
@@ -118,8 +130,17 @@ def run(
             step_failed = False
             try:
                 return_values = step_entrypoint(**function_params)
-            except:  # noqa: E722
+            except Exception as step_exception:  # noqa: E722
```
schustmi, line 133:
```suggestion
            except BaseException as step_exception:  # noqa: E722
```

I just realized that we actually want this here instead (this is equivalent to the empty `except` and also catches KeyboardInterrupt etc. The hook function then also needs to accept this type I think

src/zenml/hooks/hook_validators.py
```diff
@@ -0,0 +1,91 @@
...
+
+    from zenml.steps.base_parameters import BaseParameters
+    from zenml.steps.step_context import StepContext
+
+    sig = inspect.getfullargspec(inspect.unwrap(func))
+    sig_annotations = sig.annotations
+    if "return" in sig_annotations:
+        sig_annotations.pop("return")
+
+    if sig.args and len(sig.args) != len(sig_annotations):
```
schustmi, line 55: If `sig.args` enough here or do we also need `sig.kwonlyargs` or something like that?

schustmi (changes requested):
Some last tiny requests, and also one more thing I remembered: Now that our CI is able to nicely run integration tests for examples, it would probably be nice to also add this one to the test suite

examples/failure_and_success_hooks/README.md
```diff
@@ -0,0 +1,110 @@
...
+In this example, we define two hooks: `on_failure` and `on_success`, which print a message when the step fails or succeeds, respectively. We then use these hooks with two steps, `my_failing_step` and `my_successful_step`. When `my_failing_step` is executed, it raises a `ValueError`, which triggers the `on_failure` hook. Similarly, when `my_successful_step` is executed, it returns an integer successfully, which triggers the on_success hook.
+
+A step can also be specified as a local user-defined function
+path (of the form `mymodule.myfile.my_function`). This is
+particularly useful when defining the hooks via
+a [YAML Config](../pipelines/settings.md).
+
+## Defining steps on a pipeline level
+
+In some cases, there is a need to define a hook on all steps of a given pipeline. Rather than having to define it on all steps individually, you can also specify any hook on the pipeline level.
```
strickvl, line 42:
```suggestion
In some cases, there is a need to define a hook on all steps of a given pipeline rather than just on individual steps. 
```

examples/failure_and_success_hooks/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```python
+@pipeline(on_failure=on_failure, on_success=on_success)
+def my_pipeline(...):
+    ...
+```
+
+## Accessing step information inside a hook
+
+A hook function signature can optionally take three type annotated arguments of
```
strickvl, line 55:
```suggestion
A hook function signature can optionally take three type-annotated arguments of
```

examples/failure_and_success_hooks/README.md
```diff
@@ -0,0 +1,110 @@
...
+`Exception`.
+
+```python
+from zenml.steps import BaseParameters, StepContext, step
+
+# Use one or any of these in the signature
+def on_failure(context: StepContext, params: BaseParameters, exception: Exception):
+    print(context.step_name)  # Output will be `my_step`
+    print(type(params))  # Of type MyParameters
+    print(type(exception))  # Of type value error
```
strickvl, line 77:
```suggestion
    print(type(exception))  # Of type ValueError
```

examples/failure_and_success_hooks/hooks/success_hooks.py
```diff
@@ -0,0 +1,19 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+
+from zenml.steps import StepContext
+
+
+def pipeline_success_hook(context: StepContext) -> None:
+    print(f"Step {context.step_name} succeeded!")
```
strickvl, line 19: Should this maybe be logging instead of printing? Maybe encouraging antipatterns somehow?

src/zenml/hooks/hook_validators.py
```diff
@@ -0,0 +1,91 @@
...
+    Returns:
+        str: Resolved source path of `hook_func`.
+
+    Raises:
+        ValueError: If `hook_func` is not a valid callable.
+    """
+    if type(hook_func) is str:
+        func = source_utils.load_source_path(hook_func)
+    else:
+        # This means hook is a callable
```
strickvl, line 41:
```suggestion
```

Maybe remove this comment, since you're still checking below for whether it is actually a callable.

src/zenml/hooks/hook_validators.py
htahir1: Its enough, I think. See image attached?

image

src/zenml/hooks/hook_validators.py
```diff
@@ -0,0 +1,91 @@
...
+        func = hook_func
+
+    if not callable(func):
+        raise ValueError(f"{func} is not a valid function.")
+
+    from zenml.steps.base_parameters import BaseParameters
+    from zenml.steps.step_context import StepContext
```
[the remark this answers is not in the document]
htahir1: Yes, otherwise its a circular import

examples/failure_and_success_hooks/hooks/success_hooks.py
htahir1: Makes sense!

htahir1:
Alright guys @schustmi and @strickvl , I implemented the last of the review comments. Would appreciate  a final scan

src/zenml/orchestrators/step_runner.py
```diff
@@ -243,6 +276,67 @@ def _parse_inputs(
...
+            elif issubclass(arg_type, StepContext):
+                step_name = self.configuration.name
+                context = arg_type(
+                    step_name=step_name,
+                    output_materializers=output_materializers,
+                    output_artifact_uris=output_artifact_uris,
+                )
+                function_params[arg] = context
+
+            elif issubclass(arg_type, Exception):
```
schustmi, line 328: This needs to be BaseException now

src/zenml/hooks/hook_validators.py
```diff
@@ -0,0 +1,90 @@
...
+            "and/or `StepContext`."
+        )
+
+    if sig_annotations:
+        annotations = sig_annotations.values()
+        seen_annotations = set()
+        for annotation in annotations:
+            if annotation:
+                if annotation not in (
+                    Exception,
```
schustmi, line 67: This needs to be BaseException now

src/zenml/hooks/alerter_hooks.py
```diff
@@ -0,0 +1,84 @@
...
+
+logger = get_logger(__name__)
+
+
+if TYPE_CHECKING:
+    from zenml.steps import BaseParameters, StepContext
+
+
+def alerter_failure_hook(
+    context: "StepContext", params: "BaseParameters", exception: Exception
```
schustmi, line 28: This should be BaseException now

schustmi (changes requested):

htahir1:
Thanks @strickvl ! @schustmi i understood it now! Fixed, but would appreciate another scan

src/zenml/orchestrators/step_runner.py
```diff
@@ -415,3 +509,35 @@ def _store_output_artifacts(
...
+            function_params = self._parse_hook_inputs(
+                args=hook_spec.args,
+                annotations=hook_spec.annotations,
+                step_exception=step_exception,
+                output_artifact_uris=output_artifact_uris,
+                output_materializers=output_materializers,
+            )
+            logger.debug(f"Running hook {hook} with params: {function_params}")
+            hook(**function_params)
+        except BaseException as e:
```
schustmi, line 540:
```suggestion
        except Exception as e:
```

tests/integration/functional/cli/test_utils.py
```diff
@@ -39,7 +39,7 @@ def test_temporarily_setting_the_active_stack():
 
 def test_error_raises_exception():
     """Tests that the error method raises an exception."""
-    with pytest.raises(Exception):
+    with pytest.raises(BaseException):
```
schustmi, line 42:
```suggestion
    with pytest.raises(Exception):
```

schustmi (changes requested):

schustmi (changes requested):

schustmi (approved):

htahir1:
🚀 🚀 🚀

PR #1367, opened 2023-03-01
Load artifact store before loading artifact to register filesystem

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/materializer_utils.py
+++ b/src/zenml/utils/materializer_utils.py
@@ -124,7 +124,8 @@ def load_artifact(artifact: "ArtifactResponseModel") -> Any:
         logger.warning(
             "Unable to restore artifact store while trying to load artifact "
             "`%s`. If this artifact is stored in a remote artifact store, "
-            "this might lead to issues when trying to load the artifact."
+            "this might lead to issues when trying to load the artifact.",
+            artifact.id,
         )
 
     return _load_artifact(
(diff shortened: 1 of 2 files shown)


safoinme (approved):

htahir1 (commented):
@schustmi Would you mind adding a note in the docs about the behavior Jiri experienced on the slack thread as well? https://zenml.slack.com/archives/C01FWQ5D0TT/p1677661528488819?thread_ts=1677596537.697169&cid=C01FWQ5D0TT

schustmi:
@htahir1 This PR fixes that issue, what would you additionally expect in the docs?

htahir1:
@schustmi I thought maybe if there is a way to still get this error it might make sense to talk about managing multiple active stacks and how it affects the post execution workflow, But if thats sovled feel free to merge

schustmi:
I think this completely eliminates that tbh, as we always load the correct artifact store now in which the artifact is stored. The only issue is if they somehow deleted the artifact store, but at that point there's also nothing we can do to help them (we can't simply activate a random artifact store and hope it works)

htahir1:
I think the real question I have is: Why does reading the artifact from the artifact store require you to activate the right stack in the first place in post-execution? Isnt that something ZenML should figure out

schustmi:
But isn't this exactly what this PR is implementing? Yes I agree ZenML should figure this out. This PR implements the figuring it out. It needs to load the artifact store to register the filesystem with the correct credentials to access the e.g. S3 bucket?

Btw this PR doesn't activate any stack, it simply loads the artifact store (= registers the file system with the configured credentials) that was used to store the artifact. We're not storing the credentials and source code of the filesystem for every artifact, so I don't see any way around that.

htahir1:
Ok now i understood it! Sorry! LGTM

PR #1369, opened 2023-03-01
Update huggingface token classification example

strickvl:
Updated the example to use some default values. (The example wasn't actually broken if you use the command listed in the `README.md`, but it fails when you try running the pipeline without specifying all the default values.) It now includes some core defaults (as per the documented suggestions) so it should run without errors for anyone not referring to the documentation.

Uncommenting the integration test will have to wait until other changes are made elsewhere.

## Pre-requisites

## Types of changes


examples/huggingface/run.py
```diff
@@ -103,13 +103,15 @@ def main(
...
         token_classification_config = HuggingfaceParameters(
             label_all_tokens=True,
-            pretrained_model=pretrained_model,
-            epochs=epochs,
-            batch_size=batch_size,
+            pretrained_model=pretrained_model or "distilbert-base-uncased",
+            epochs=epochs or 1,
+            batch_size=batch_size or 16,
             dummy_run=not full_set,
+            dataset_name=kwargs.get("dataset_name") or "conll2003",
```
schustmi, line 112: I just checked the code and all of these already contain defaults from the click option? Wouldn't this `or` never be called?

Same for text_column and label_column

examples/huggingface/run.py
```diff
@@ -103,13 +103,15 @@ def main(
...
 ):
     if nlp_task == "token-classification":
-        # Run Pipeline
         token_classification_config = HuggingfaceParameters(
             label_all_tokens=True,
-            pretrained_model=pretrained_model,
-            epochs=epochs,
-            batch_size=batch_size,
+            pretrained_model=pretrained_model or "distilbert-base-uncased",
+            epochs=epochs or 1,
```
schustmi, line 109: `epochs` is an int and will therefore never be optional, the `or` won't be called (unless someone passes 0 but that's just a user error that should crash IMO)

examples/huggingface/run.py
```diff
@@ -103,13 +103,15 @@ def main(
...
+            batch_size=batch_size or 16,
```
schustmi, line 110: Same as epochs I think

examples/huggingface/run.py
```diff
@@ -103,13 +103,15 @@ def main(
     **kwargs
 ):
     if nlp_task == "token-classification":
-        # Run Pipeline
         token_classification_config = HuggingfaceParameters(
             label_all_tokens=True,
-            pretrained_model=pretrained_model,
-            epochs=epochs,
-            batch_size=batch_size,
+            pretrained_model=pretrained_model or "distilbert-base-uncased",
```
schustmi, line 108: Similar to the others, this already has a default value.

schustmi (changes requested):
All of my comments relate to the same issue: These values already get default values by the click options defined on the function. Maybe some of those (especially the ones which need to be different for token vs sequence classification, e.g. `dataset_name`) should not have a default value and then set the correct default dynamically?

safoinme (approved):

examples/huggingface/run.py
```diff
@@ -100,17 +97,33 @@ def main(
     batch_size: int,
     epochs: int,
     full_set: bool,
-    **kwargs
+    max_seq_length: int,
+    init_lr: float,
+    weight_decay_rate: float,
+    text_column: str,
+    label_column: str,
+    dataset_name: str,
```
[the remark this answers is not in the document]
strickvl: Yep you're right.

schustmi (approved):

PR #1371, opened 2023-03-01
Add pydantic materializer

htahir1:
## Describe changes
I added a pydantic materializer. To check:

```
import random

from zenml.pipelines import pipeline
from zenml.steps import Output, step

class Lol(BaseModel):
    a: int = 1

@step(enable_cache=False)
def get_first_num() -> Lol:
    """Returns an integer."""

    return Lol(a=5)


@step(enable_cache=False)
def get_random_int() -> Output(random_num=int):
    """Get a random integer between 0 and 10"""
    return random.randint(0, 10)


@step
def subtract_numbers(first_num: int, lol: Lol) -> Output(result=int):
    """Subtract random_num from first_num."""
    print(lol.a)
    return first_num - lol.a


@pipeline
def hamzas_example_pipeline2(get_first_num, get_random_int, subtract_numbers):
    # Link all the steps artifacts together
    first_num = get_first_num()
    random_num = get_random_int()
    subtract_numbers(random_num, first_num )


if __name__ == "__main__":
    # Initialize a new pipeline run
    aep = hamzas_example_pipeline2(
        get_first_num=get_first_num(),
        get_random_int=get_random_int(),
        subtract_numbers=subtract_numbers(),
    )
    DAG = aep.run()
    
```

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/materializers/pydantic_materializer.py
+++ b/src/zenml/materializers/pydantic_materializer.py
@@ -32,7 +32,7 @@ class PydanticMaterializer(BuiltInContainerMaterializer):
     )
 
     def load(self, data_type: Type[BaseModel]) -> Any:
-        """Reads basic primitive types from JSON.
+        """Reads BaseModel from JSON.
 
         Args:
             data_type: The type of the data to read.
@@ -44,7 +44,7 @@ class PydanticMaterializer(BuiltInContainerMaterializer):
         return data_type(**contents)
 
     def save(self, data: BaseModel) -> None:
-        """Serialize a basic type to JSON.
+        """Serialize a BaseModel to JSON.
 
         Args:
             data: The data to store.
--- a/src/zenml/materializers/pydantic_materializer.py
+++ b/src/zenml/materializers/pydantic_materializer.py
@@ -13,26 +13,26 @@
 #  permissions and limitations under the License.
 """Implementation of ZenML's pydantic materializer."""
 
+import os
 from typing import TYPE_CHECKING, Any, Dict, Type
 
 from pydantic import BaseModel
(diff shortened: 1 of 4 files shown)


src/zenml/materializers/pydantic_materializer.py
```diff
@@ -0,0 +1,70 @@
...
+        contents = super().load(dict)
+        return data_type(**contents)
+
+    def save(self, data: BaseModel) -> None:
+        """Serialize a BaseModel to JSON.
+
+        Args:
+            data: The data to store.
+        """
+        super().save(data.dict())
```
schustmi, line 55: This does not work unfortunately, as pydantic supports more types than the default json serializable ones (e.g. datetime, UUID).

We need to either
- store `data.json()` in a file (I don't see a reason not to do this)
- use `super().save(json.loads(data.json()))

htahir1: But `data.json` basically does `data.dict` in the background right?

schustmi: And in the process converts classes like UUID and datetime which the current solution won't cover. We have this pattern in our code a few times for exactly that reason, see e.g. https://github.com/zenml-io/zenml/blob/main/src/zenml/utils/pydantic_utils.py#L193-L194

htahir1: Ah I got it. I now did it so that it goes via `data.json()`

schustmi (approved):
🦭

AlexejPenner (approved):
:seal:

PR #1374, opened 2023-03-02
Add debug mode flag for `zenml info`

strickvl:
Added a `-d` or `--debug` flag to go with `zenml info` that will output information about a user's current / active stack (and the components included as part of it. In addition to whatever we usually have as part of the `zenml info` command, you get this kind of information, to be used potentially as part of debugging cloud stacks on Slack:

ScreenShot 2023-03-02 at 12 24 37

## Pre-requisites

## Types of changes


src/zenml/cli/utils.py
```diff
@@ -970,6 +970,42 @@ def print_components_table(
     print_table(configurations)
 
 
+def print_debug_stack(
+    stack: "StackResponseModel", components: "List[ComponentResponseModel]"
+) -> None:
+    """Print stack and components for debugging purposes.
```
schustmi, line 976: I think it would be nicer if this function extracted the components itself instead of them being passed as a separate argument

src/zenml/cli/utils.py
```diff
@@ -970,6 +970,42 @@ def print_components_table(
...
+    console.print(f"User: {stack.user.name} / {str(stack.user.id)}")
+    console.print(
+        f"Workspace: {stack.workspace.name} / {str(stack.workspace.id)}"
+    )
+
+    for component in components:
+        declare(
+            f"\n{component.type.value.upper()}: {component.name}\n", bold=True
+        )
+        console.print(f"Name: {component.name}")
```
schustmi, line 995: Is this essentially a better looking version of `zenml stack-component describe`? If yes, can we repurpose it for that somehow?

schustmi (changes requested):

schustmi:
Also on a more general note, the `--debug` flag name doesn't really capture what this does IMO, maybe something like `--include-stack-info` or some shortened version of that?

src/zenml/cli/utils.py
strickvl: Not quite. I explicitly didn't want all the rich tabular formatting so that it was copy-pastable, so not sure how much crossover there is any more then.

schustmi (approved):

stefannica (approved):
Looks good, thanks for scrubbing the secrets !

PR #1375, opened 2023-03-02
Add orchestrator logs link for Sagemaker

strickvl:
Added a metadata URL for sagemaker orchestrator logs. The link takes you to a CloudWatch logs panel where you can click through to logs for the specific step you're interested in, as in:

ScreenShot 2023-03-02 at 16 20 51

Needs testing on a Sagemaker stack to confirm that it works as expected.

## Pre-requisites

## Types of changes


schustmi (approved):

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -215,6 +218,23 @@ def prepare_or_run_pipeline(
...
+
+        Returns:
+            The region name.
+
+        Raises:
+            RuntimeError: If the region name cannot be retrieved.
+        """
+        try:
+            return cast(str, sagemaker.Session().boto_region_name)
+        except Exception as e:
```
htahir1, line 232: Should this be such a hard error? Its only used to get the orchestrator link right? Maybe we can raise a warning instead of maybe catch it in get_pipeline_run_metadata in line 249 to not fail `hard`

strickvl: You're right. I'll fix this.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -215,6 +218,23 @@ def prepare_or_run_pipeline(
...
+            raise RuntimeError(
```
htahir1, line 233: it should be a `logger.error(...)` now right? We dont want it to have an exception just because we couldnt get the region

strickvl: I'm catching the exception where it's called.

htahir1: You are catching it and then raising another exception so rather catch it and do a logger.error so that the program doesnt fail

strickvl: Hadn't pushed my code 🙈. In any case, I think we're agreed / aligned on what needs to happen here.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -226,11 +246,20 @@ def get_pipeline_run_metadata(
...
-        #     f"home?region={region}"
-        # )
+        try:
+            region_name = self._get_region_name()
+        except RuntimeError:
+            region_name = ""
+
+        aws_run_id = os.environ[ENV_ZENML_SAGEMAKER_RUN_ID].split("/")[-1]
+        orchestrator_logs_url = (
+            f"https://{region_name}.console.aws.amazon.com/"
```
htahir1, line 256: in case `region_name` is "" then `orchestrator_logs_url` would be misformatted.. maybe return in the except above

htahir1 (approved):
LGTM 🚀

PR #1377, opened 2023-03-02
Fix label studio `dataset delete` command

strickvl:
- updated README which had some wrong commands
- implemented the `zenml annotator dataset delete` command which was waiting for an update to the label_studio sdk package
- bumped our version of the label_studio sdk package accordingly

(Tested this manually to ensure it works.)

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/label_studio/__init__.py
+++ b/src/zenml/integrations/label_studio/__init__.py
@@ -26,7 +26,7 @@ class LabelStudioIntegration(Integration):
     """Definition of Label Studio integration for ZenML."""
 
     NAME = LABEL_STUDIO
-    REQUIREMENTS = ["label-studio==1.6.0", "label-studio-sdk==0.0.15"]
+    REQUIREMENTS = ["label-studio==1.6.0", "label-studio-sdk>=0.0.17"]
 
     @classmethod
     def flavors(cls) -> List[Type[Flavor]]:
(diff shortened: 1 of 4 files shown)


src/zenml/integrations/label_studio/__init__.py
```diff
@@ -26,7 +26,7 @@ class LabelStudioIntegration(Integration):
     """Definition of Label Studio integration for ZenML."""
 
     NAME = LABEL_STUDIO
-    REQUIREMENTS = ["label-studio==1.6.0", "label-studio-sdk==0.0.15"]
+    REQUIREMENTS = ["label-studio==1.6.0", "label-studio-sdk>=0.0.17"]
```
safoinme, line 29: Would we want to limit this to the current version instead of fully open range?

safoinme (approved):

src/zenml/integrations/label_studio/__init__.py
strickvl: In general we're trying to open up the ranges of the versions we support for all integrations, so the minimum version to be installed is 0.0.17. Label Studio is one of our integrations with the most dependency conflicts, so I thought it's probably better to open it up?

schustmi (approved):

src/zenml/integrations/label_studio/__init__.py
safoinme: Ah yes, i guess just brought it up because just had an issue with this in mlflow

PR #1380, opened 2023-03-03
Fix order of runs and order of pipelines in post-execution

stefannica:
## Describe changes
As a follow-up to #1332 , where we fixed the order of pipeline runs fetched in the post-execution phase, this PR updates the docs and all our examples to use the correct indexing `runs[0]` instead of `runs[-1]` to fetch the last run of a pipeline.

This PR also applies the same order change to pipelines fetched by `get_pipelines()`, which suffers from the same problem (returns the oldest X pipelines instead of the newest X). Here we also use the index `[0]` instead of `[-1]`.

## Pre-requisites

## Types of changes


htahir1 (commented):
Oh I suppose we need to do this for Zenbytes and other examples to like gitflow?

schustmi (approved):
Thanks so much for fixing the mess I introduced!

PR #1384, opened 2023-03-06
CI speed improvements

stefannica:
## Describe changes
This PR attempts to speed up the CI using various tricks:
 * install all python requirements, including (non-problematic) ZenML integrations, as part of a separate job that caches virtualenvs and pip downloads to speed up the CI on re-runs
 * run linting, docstring checks, spell checker, broken link checker and unit-tests as a single job
 * condition running integration tests on passing the linter/unit-tests first to avoid blocking runners unnecessarily. Justification: PRs need to pass both unit tests and integration tests to be merged, so running integration tests when linter checks are failing is a waste of resources, given that more changes are needed to fix the linter errors, which will re-trigger the CI. In some cases the linter also reveals errors that also impact the integration tests. As an extra reason, this discourages developers to push changes without running the linters locally first.
 * use different CI "lanes" for different OS distros to avoid the MacOS runner bottleneck and maximize runner time.

Screenshot from 2023-03-07 22-25-12

The PR also unifies all CI workflows to look and feel like the integration workflow, updates some actions and uses other popular actions instead of custom steps to setup and cache Python virtual environments.

The key take-away is that PRs that modify the pyproject.toml file may take longer to run, but outside of this, jobs will be overall faster, especially on incremental PR CI runs, for the following reasons:
1. the number of github runners required for a complete CI run has been reduced
2. the jobs are now faster because we cache not only the python virtualenv but also the pip download cache and we don't invalidate the pip download cache on code file changes, but rather periodically.
3. we also cache the zenml integrations (and we invalidate the cache if their requirements change)
4. we no longer engage github runners unnecesarily: integration tests are no longer executed if linter or unit-tests are failing

Some more details on caching:

* every (os, python-version) pair uses two caches:
  * a cache for the pip downloaded packages (the pip cache). This cache is updated on a weekly basis.
  * a cache for the python virtualenv where all packages (zenml, all its extra requirements and all basic integrations) are installed. This cache is updated when one of the following happens:
    * the `pyproject.toml` file is changed
    * the cache is invalidated on a weekly basis
    * the ZenML integrations requirements change
* the caches are updated at the end of each job and only if the cache is not hit and the job run succeeds. For this reason, we continue to use a separate job to create these caches at the beginning of the CI run instead of creating them as part of the linter-and-unit-test job for example. This way we maximize the cache re-use over multiple CI re-runs, which is the most common pattern with PRs. This still requires an extra of 12 github runners at the beginning of the CI run, but the alternative is worse.
* finally, we still have the means to manually invalidate all caches via a github secret, if they become broken somehow

## Pre-requisites

## Types of changes


.github/actions/setup_environment/action.yml
```diff
@@ -32,43 +38,59 @@ runs:
...
+      if: steps.cache-virtualenv.outputs.cache-hit == 'true' || inputs.install_integrations == 'yes'
       shell: bash
-      run: echo "VENV=.venv/bin/activate" >> "$GITHUB_ENV"
-      if: runner.os != 'Windows'
+      run: |
+        pip install -e .[server,templates,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,dev]
+
+        # install airflow requirements
+        # pip install --upgrade "numpy>=1.21.0,<1.24.0"
+        # pip install apache-airflow-providers-docker
```
strickvl, line 89: What are these here for?

.github/workflows/ci.yml
```diff
@@ -16,64 +16,103 @@ concurrency:
...
-  integration-test:
-    needs: setup-python-environment
-    # Exit if its a commit from Gitbook
-    if: ${{ ! startsWith(github.event.head_commit.message, 'GitBook:') }}
+  lint-unit-test-ubuntu:
+    needs: setup-python-environment-ubuntu
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
strickvl, line 63: is it possible to declare these versions in one place somewhere? Just thinking about when we have to update this + needing to change in multiple places.

strickvl:
ScreenShot 2023-03-08 at 09 30 39

@stefannica, would it be possible to put the OS' name first in the identification? It's impossible to tell from this (at a glance) which you're looking at.

.github/actions/setup_environment/action.yml
stefannica: Left overs from the old CI, already cleaned up.

.github/workflows/ci.yml
stefannica: I tried a bunch of things, including YAML anchors, but hit this 3+ years old open request to GitHub to support them that is still in limbo and then I just gave up: https://github.com/actions/runner/issues/1182

AlexejPenner:
> > It looks like lint-unit-test- does not actually make any changes to the pip environment. Couldn't we also perform the setup-python-environment tasks within the same worker as lint-unit-test? During "rush-hour" the provisioning of workers seems to be a large contributor to our long wait times. So not relinquishing workers between setup-python-environment and lint-unit-test- could be beneficial.
> 
> I thought about it and this was even one of the first implementations I tried, but I realized there is one thing that prevents us from using it: if the job fails, the cache is not saved (and there's nothing you can do about it). But where we really save time with caches is when we re-use them across subsequent re-runs of failed jobs, which happens a lot with linter/unit tests, especially during the first few iterations of a PR.

Good point :)

stefannica:
> ScreenShot 2023-03-08 at 09 30 39
> 
> @stefannica, would it be possible to put the OS' name first in the identification? It's impossible to tell from this (at a glance) which you're looking at.

You're right, this looks better.
Screenshot from 2023-03-08 10-32-23

strickvl (approved):
Let's get this on the road!

safoinme (approved):
Really amazing work put on this! This will save a lot of time. Thank you

PR #1390, opened 2023-03-07
Change `pipeline_run_id` to `run_name`

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


docs/book/advanced-guide/practical/deploying-models.md
```diff
@@ -83,7 +83,7 @@ def seldon_model_deployer_step(
         implementation="TENSORFLOW_SERVER",
         secret_name="seldon-secret",
         pipeline_name = step_env.pipeline_name,
-        pipeline_run_id = step_env.pipeline_run_id,
+        rune_name = step_env.rune_name,
```
strickvl, line 86:
```suggestion
        run_name = step_env.run_name,
```

src/zenml/cli/model.py
```diff
@@ -630,15 +631,18 @@ def register_model_version(
...
         # Parse metadata
         metadata = dict(metadata) if metadata else {}
         registerted_metadata = ModelRegistryModelMetadata(**dict(metadata))
         registerted_metadata.zenml_version = zenml_version
-        registerted_metadata.zenml_pipeline_run_id = zenml_pipeline_run_id
+        registerted_metadata.zenml_run_name = (
+            zenml_run_name or zenml_pipeline_run_id
+        )
         registerted_metadata.zenml_pipeline_name = zenml_pipeline_name
         registerted_metadata.zenml_step_name = zenml_step_name
```
strickvl, line 647:
```suggestion
        registered_metadata.zenml_run_name = (
            zenml_run_name or zenml_pipeline_run_id
        )
        registered_metadata.zenml_pipeline_name = zenml_pipeline_name
        registered_metadata.zenml_step_name = zenml_step_name
```
Needs fixing elsewhere as well in the model, (i.e. above). I think.

strickvl (changes requested):
Some small things. If I'm not mistaken, you've changed the model so this will also require an alembic migration script as well? @AlexejPenner is that correct?

AlexejPenner:
> Some small things. If I'm not mistaken, you've changed the model so this will also require an alembic migration script as well? @AlexejPenner is that correct?

Couldn't find any of our models/schemas that changed, so loogs good to me

src/zenml/model_registries/base_model_registry.py
```diff
@@ -63,7 +63,7 @@ class ModelRegistryModelMetadata(BaseModel):
     """
 
     zenml_version: Optional[str] = None
-    zenml_pipeline_run_id: Optional[str] = None
+    zenml_run_name: Optional[str] = None
     zenml_pipeline_name: Optional[str] = None
     zenml_pipeline_uuid: Optional[str] = None
     zenml_pipeline_run_uuid: Optional[str] = None
```
schustmi, line 69: I guess this is what we call `pipeline_run_id` everywhere, does it make sense to change that as well?

src/zenml/integrations/seldon/model_deployers/seldon_model_deployer.py
```diff
@@ -398,7 +398,8 @@ def find_model_server(
         # Use a Seldon deployment service configuration to compute the labels
         config = SeldonDeploymentConfig(
             pipeline_name=pipeline_name or "",
-            pipeline_run_id=pipeline_run_id or "",
+            run_name=run_name or "",
+            pipeline_run_id=run_name or "",
```
schustmi, line 402: Can you explain why we still need to pass it for the deprecated/hopefully unused `pipeline_run_id` here?

src/zenml/services/service.py
```diff
@@ -42,15 +42,51 @@ class ServiceConfig(BaseTypedModel):
...
+        """Setter for the run_name attribute.
+
+        If the run_name attribute is set to an empty string, it will be
+        automatically set to the pipeline_run_id value.
+
+        Args:
+            name: name of the attribute.
+            value: value of the attribute.
+        """
+        if name == "run_name" and not value:
```
schustmi, line 91: Can you explain when this case would happen, someone setting the run name to `None`/empty string but still setting the old `pipeline_run_id`?

src/zenml/integrations/seldon/services/seldon_deployment.py
```diff
@@ -84,8 +84,10 @@ def get_seldon_deployment_labels(self) -> Dict[str, str]:
         labels = {}
         if self.pipeline_name:
             labels["zenml.pipeline_name"] = self.pipeline_name
+        if self.run_name:
+            labels["zenml.run_name"] = self.run_name
         if self.pipeline_run_id:
-            labels["zenml.pipeline_run_id"] = self.pipeline_run_id
+            labels["zenml.pipeline_run_id"] = self.run_name
```
schustmi, line 90: I don't really understand the logic here, why do we still need to add it to the labels for "pipeline_run_id"?

schustmi (changes requested):

src/zenml/integrations/seldon/model_deployers/seldon_model_deployer.py
safoinme: Because this is used to also get the already deployed model within a Seldon cluster, because the function that finds the deployed model tries to search for a label with the name `pipeline_run_id` within the deployed resource which may return that no deployed model exists if that doesn't match

src/zenml/services/service.py
safoinme: i thought maybe, if we try to load back a service of a model that was already deployed

src/zenml/integrations/bentoml/model_deployers/bentoml_model_deployer.py
```diff
@@ -388,9 +389,8 @@ def _matches_search_criteria(
                 == config.pipeline_step_name
             )
             and (
-                not config.pipeline_run_id
-                or existing_service_config.pipeline_run_id
-                == config.pipeline_run_id
+                not config.run_name
```
schustmi, line 392: Just a clarification for me: This now works because of the logic you implemented in the `ServiceConfig` class, where you set the value of `run_name` to the old `pipeline_run_id` value, correct? Otherwise the legacy deployments would not have a value for `run_name`.

schustmi (changes requested):

src/zenml/integrations/bentoml/model_deployers/bentoml_model_deployer.py
safoinme: Yes, that should be the case here, even tho if `run_name` isn't provider here it would still work using other parameters.

schustmi (approved):

strickvl (approved):

src/zenml/services/service.py
```diff
@@ -42,15 +42,55 @@ class ServiceConfig(BaseTypedModel):
...
         pipeline_step_name: name of the pipeline step that spun up the service
+        run_name: name of the pipeline run that spun up the service.
     """
 
     name: str = ""
     description: str = ""
     pipeline_name: str = ""
-    pipeline_run_id: str = ""
     pipeline_step_name: str = ""
+    run_name: str = ""
```
stefannica, line 53: This change needs to be coupled with a root validator that copies all data from the existing `pipeline_run_id` field into the new `run_name` field, otherwise this information is lost on de-serialization, e.g.:

```
>>> from zenml.services.service import ServiceConfig
>>> d = {"name":"test", "pipeline_run_id": "pipeline_1234"}
>>> ServiceConfig.parse_obj(d)
ServiceConfig(type='zenml.services.service.ServiceConfig', name='test', description='', pipeline_name='', pipeline_step_name='', run_name='')```

File: examples/great_expectations_data_validation/great_expectations.ipynb [hunk 9]
```diff
@@ -357,9 +357,9 @@
     "    whether it passed or failed.\"\"\"\n",
     "    step_env = cast(StepEnvironment, Environment()[STEP_ENVIRONMENT_NAME])\n",
     "    pipeline_name = step_env.pipeline_name\n",
-    "    pipeline_run_id = step_env.pipeline_run_id\n",
+    "    run_name = step_env.run_name\n",
     "    step_name = step_env.step_name\n",
-    "    pipeline_context = f\"Pipeline {pipeline_name}, with run {pipeline_run_id}, in step {step_name} produced the following output:\\n\\n\"\n",
+    "    pipeline_context = f\"Pipeline {pipeline_name}, with run {run_name}, in step {step_name} produced the following output:\\n\\n\"\n",
```
stefannica, line 362: This change is not needed/incorrect.

stefannica (changes requested):

stefannica (approved):

PR #1392, opened 2023-03-07
Numpy materializer metadata for arrays with strings

safoinme:
## Describe changes
This checks what type of array and tried to extract some metadata based on the type of the array.

## Pre-requisites

## Types of changes


src/zenml/materializers/numpy_materializer.py
```diff
@@ -145,4 +144,53 @@ def extract_metadata(
...
+            The extracted metadata as a dictionary.
+        """
+        base_metadata = super().extract_metadata(arr)
+        if np.issubdtype(arr.dtype, np.number):
+            return {**base_metadata,**self.extract_numeric_metadata(arr)}
+        elif np.issubdtype(arr.dtype, np.unicode_) or np.issubdtype(
+            arr.dtype, np.object_):
+            return {**base_metadata,**self.extract_text_metadata(arr)}
+        else:
+            logger.info(
```
fa9r, line 193: Would suggest to change this to debug or take it out since I think users will have no clue what this message could mean

src/zenml/materializers/numpy_materializer.py
```diff
@@ -145,4 +144,53 @@ def extract_metadata(
...
+        """Extracts text metadata from a numpy array.
+
+        Args:
+            arr: The numpy array to extract metadata from.
+
+        Returns:
+            A dictionary of metadata.
+        """
+        text = " ".join(arr)
+        words = text.split()
```
fa9r, line 161: This I don't get, you join the whole array into a giant string but then immediately split it again, shouldn't there be an easier way to do that such as `arr.tolist()`? Or is it because each array element could have multiple words in it? (I'm not familiar with text numpy arrays at all)

fa9r (approved):
Wow I didn't even know text numpy arrays were a thing. Good catch. Implementation looks good to me 👍

schustmi (commented):
Agree with @fa9r's points, also a test would be nice. Otherwise looks good!

src/zenml/materializers/numpy_materializer.py
```diff
@@ -145,4 +144,53 @@ def extract_metadata(
...
+        """
+        base_metadata = super().extract_metadata(arr)
+        if np.issubdtype(arr.dtype, np.number):
+            return {**base_metadata,**self.extract_numeric_metadata(arr)}
+        elif np.issubdtype(arr.dtype, np.unicode_) or np.issubdtype(
+            arr.dtype, np.object_):
+            return {**base_metadata,**self.extract_text_metadata(arr)}
+        else:
+            logger.info(
+                f"No Additional metadata extraction for dtype {arr.dtype}"
```
strickvl, line 194:
```suggestion
                f"No additional metadata extraction for dtype {arr.dtype}"
```

src/zenml/materializers/numpy_materializer.py
safoinme: because in first join its series of phrases, so we join every phrase to give us one big text that we then split based on words? would it make a different if we iterate and split every phrase?

fa9r: Nah it's fine then I guess, no point in performance optimization for something like this

tests/unit/materializers/test_numpy_materializer.py
```diff
@@ -0,0 +1,102 @@
...
+            )  # something was written
+
+            # Assert that metadata extraction returns a dict
+            metadata = materializer.extract_metadata(step_output)
+            assert isinstance(metadata, dict)
+            for key, value in metadata.items():
+                assert isinstance(key, str)
+                assert isinstance(value, MetadataTypeTuple)
+
+            return metadata
```
fa9r, line 65: This seems duplicated with `_test_materializer()` except for the return. I'd suggest to change `_test_materializer()` instead to make it work with your test. The easiest solutions that come to my mind:
- `return loaded_data, metadata`
- or, take an optional arg `metadata_validation_function` that is then called on the metadata, similar to what the `validation_function` arg does for the loaded data

fa9r (approved):
Nice, now we also finally have unit tests for the numpy materializer :D

The only suggestion I have is to refactor the test a bit to reduce code duplication, otherwise ready to go 👍

schustmi (approved):
+1 to what felix said, rest looks great!

tests/unit/materializers/test_numpy_materializer.py
safoinme: Good idea, i have gone with ading a condition to the `_test_materializer()` to return or not return metadata with loaded data

Issue #1391, opened 2023-03-07
[BUG]: Server Error when materializing single row Pandas DataFrame


Christopher-Norman:
### Contact Details [Optional]

chris@fuzzylabs.ai

### System Information

ZENML_LOCAL_VERSION: 0.34.0
ZENML_SERVER_VERSION: 0.34.0
ZENML_SERVER_DATABASE: sqlite
ZENML_SERVER_DEPLOYMENT_TYPE: local
ZENML_CONFIG_DIR: /Users/christophernorman/Library/Application Support/zenml
ZENML_LOCAL_STORE_DIR: /Users/christophernorman/Library/Application Support/zenml/local_stores
ZENML_SERVER_URL: REMOVED
ZENML_ACTIVE_REPOSITORY_ROOT: /Users/christophernorman/reproduce-zenml-bug
PYTHON_VERSION: 3.10.9
ENVIRONMENT: native
SYSTEM_INFO: {'os': 'mac', 'mac_version': '10.16'}
ACTIVE_WORKSPACE: default
ACTIVE_STACK: default
ACTIVE_USER: default
TELEMETRY_STATUS: enabled
ANALYTICS_CLIENT_ID: d66ee04d-9e06-4057-b159-89991a575308
ANALYTICS_USER_ID: 8f970f2d-2027-48fe-b98f-cc3be62b23ec
ANALYTICS_SERVER_ID: d66ee04d-9e06-4057-b159-89991a575308
INTEGRATIONS: ['kaniko']
PACKAGES: {'certifi': '2022.12.7', 'pytz': '2022.7.1', 'setuptools': '65.6.3', 'cryptography': '39.0.2', 'pyzmq': '25.0.0', 'pip': '23.0.1', 'packaging': '23.0', 'attrs': '22.2.0', 'argon2-cffi': 
'21.3.0', 'argon2-cffi-bindings': '21.2.0', 'isoduration': '20.11.0', 'rich': '12.6.0', 'websockets': '10.4', 'ipython': '8.11.0', 'click': '8.1.3', 'jupyter-client': '8.0.3', 'ipywidgets': '7.7.3', 
'nbconvert': '7.2.9', 'ipykernel': '6.21.3', 'notebook': '6.5.3', 'tornado': '6.2', 'docker': '6.0.1', 'bleach': '6.0.0', 'psutil': '5.9.4', 'traitlets': '5.9.0', 'nbformat': '5.7.3', 'pyyaml': '5.4.1', 
[...]

PR #1395, opened 2023-03-08
Switch FastAPI response class to orjson so `NaN` values don't break the server

fa9r:
Saving any `NaN` values in our DB currently breaks the server since FastAPI does not know how to serialize it, see https://github.com/zenml-io/zenml/issues/1391

To fix this, I switched the default response class to `orjson` as suggested here: https://github.com/tiangolo/fastapi/issues/459

This also introduces [orjson](https://github.com/ijl/orjson) as new server dependency.

## Pre-requisites

## Types of changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,7 +140,7 @@ types-psutil = { version = "^5.8.13", optional = true }
 docker-compose = { version = "^1.28", optional = true }
 
 [tool.poetry.extras]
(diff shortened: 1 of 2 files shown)


wjayesh (approved):
Thanks for the quick fix! LGTM

stefannica (approved):
Nice one ! As I understand it, orjson also claims to be a lot faster than other JSON serializer implementations (the exact words are "the fastest Python library for JSON"). This should also boost our server performance somewhat, especially at scale, when lots of objects are returned in `list` calls.

PR #1400, opened 2023-03-09
Replace `pyspelling` with `typos`

strickvl:
(Context: I'd [opened an issue](https://github.com/charliermarsh/ruff/issues/1628) on the `ruff` project wondering if they could replace `pyspelling` but they suggested that [the `typos` crate](https://crates.io/crates/typos) was already 'blazingly fast' and thus it probably wouldn't be worth their time to implement something new. I'd forgotten about this conversation until @schustmi reminded me of it today.)

[`typos`](https://crates.io/crates/typos) has two main advantages:
- it is indeed 'blazingly fast', so our CI and local checks will be much faster
- it's far less likely to throw false positives, so we no longer need our `.pyspelling-ignored-words` exception list or to bother with that process of always adding new exceptions

The main disadvantage that I can see right now is that it's not available as a pip package, so for the CI we would use the github actions available for us, and locally you can download a binary (for those using linux) or install via `brew` (Mac).

IMO the benefits outweigh the downsides.

## Pre-requisites

## Types of changes


pyproject.toml
```diff
@@ -138,8 +138,6 @@ types-termcolor = { version = "^1.1.2", optional = true }
 types-psutil = { version = "^5.8.13", optional = true }
 # test framework dependencies
 docker-compose = { version = "^1.28", optional = true }
-# spell checker dependencies
-pyspelling = { version = ">=2.8.2", optional = true }
```
fa9r, line 142: wouldn't this need to contain `typos` now instead?

fa9r (approved):
Awesome!! Finally pyspelling ignore is gone 😄

fa9r:
Also kudos for fixing all these spelling issues that previously slipped through @alex-zenml 🙏💯

.typos.toml
```diff
@@ -0,0 +1,27 @@
...
+K_Scatch = "K_Scatch"
+MCAGA1UECgwZQW1hem9uIFdlYiBTZXJ2aWNlcywgSW5jLjETMBEGA1UECwwKQW1h = "MCAGA1UECgwZQW1hem9uIFdlYiBTZXJ2aWNlcywgSW5jLjETMBEGA1UECwwKQW1h"
+VQQGEwJVUzEQMA4GA1UEBwwHU2VhdHRsZTETMBEGA1UECAwKV2FzaGluZ3RvbjEi = "VQQGEwJVUzEQMA4GA1UEBwwHU2VhdHRsZTETMBEGA1UECAwKV2FzaGluZ3RvbjEi"
+
+
+[default.extend-words]
+# Don't correct the surname "Teh"
+aks = "aks"
+hashi = "hashi"
+womens = "womens"
```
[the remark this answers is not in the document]
strickvl: No easy other way round it.

pyproject.toml
strickvl: No, because it isn't a pip package. You have to install it as a binary locally, and we use the GH Actions version in the CI.

schustmi (approved):
🙏🙏🙏

PR #1402, opened 2023-03-12
Improve Slack Alerter by adding message blocks

soubenz:
## Describe changes
I added the support of blocks in the slack alerter component. Blocks allow further customization of the slack message, the current implementation will generate a slack message in the following format :

zenml_slack_bot

## Pre-requisites

## Types of changes


htahir1:
@soubenz Thanks for the contribution! Before we begin reviewing, could you make sure you resolve the conflicts? :-) Then we'll leave comments!

Again, super happy for your contribution!

htahir1:
@soubenz Just a soft ping on this one :-)!

soubenz:
> @soubenz Just a soft ping on this one :-)!

@htahir1  thanks! 
Conflicts are resolved now :)

htahir1:
@soubenz Nice! Is it still a draft PR or would you mark it as ready for review now?

src/zenml/integrations/slack/alerters/slack_alerter.py
```diff
@@ -131,6 +141,59 @@ def _get_disapprove_msg_options(
...
+        self, message: str, params: Optional[BaseAlerterStepParameters]
+    ) -> List[Dict]:  # type: ignore
+        """Define which messages will lead to disapproval during ask().
+
+        Args:
+            message: message
+            params: Optional parameters.
+
+        Returns:
+            Set of messages that lead to disapproval in alerter.ask().
```
strickvl, line 154: As far as I can see, this block creation has nothing to do with disapproval during `ask()`. The docstring probably needs updating.

strickvl (changes requested):
Some initial feedback following a quick review just now. Also noticed some linting errors showing up in the CI. Could you take a look at those?

Thank you for the contribution, though! This makes the alerter messages much easier on the eyes! 🎨

fa9r (changes requested):
Awesome, love it!

The only change I'd request is to make the new formatting optional so other users can still configure their step to get the plain message version in case they prefer that.

Once that is possible I'd be happy to approve this and see it go live, super useful feature 👍

fa9r:
Oh, also, if you like you could also add the same logic to the ask step. I think having the environment info is super useful as context for all messages actually.

src/zenml/integrations/slack/alerters/slack_alerter.py
soubenz: done !

src/zenml/integrations/slack/alerters/slack_alerter.py
```diff
@@ -44,6 +53,8 @@ class SlackAlerterParameters(BaseAlerterStepParameters):
 
     # Set of messages that lead to disapproval in alerter.ask()
     disapprove_msg_options: Optional[List[str]] = None
+    payload: Optional[SlackAlerterPayload]
```
fa9r, line 56: I guess this needs to have a default None to be optional

```suggestion
    payload: Optional[SlackAlerterPayload] = None
```

fa9r (approved):
Awesome, that was quick. Looks good to me, thanks again for contributing 🙏 🚀

strickvl:
Thank you for the updates and pushed changes. I'll just let the CI tests run through but otherwise it looks good to go from my side. 🙏🏻🙏🏻

PR #1406, opened 2023-03-13
Changed to debug log

htahir1:
## Describe changes
I wanted to clean up log output

## Pre-requisites

## Types of changes


schustmi (commented):
I don't like this change tbh, we should reduce the amount but this is something security relevant which users should not do

htahir1:
@schustmi I understand but this should only be given once and it shouldnt be in the constructor.. Otherwise its spammy as hell. I think we can raise it when people are registering a component

schustmi:
Registering seems like a good compromise to me!

schustmi:
Maybe we can add a parameter like `warn_about_plain_text_secrets` to `StackComponentConfig.__init__` which defaults to `False`, and then in `Client.create_stack_component(...)` we explicitly set it to `True`? This would only show the warning when registering a component, but the warning will not be shown in the dashboard (which should probably handle it differently anyway, e.g. using a dynamic warning message or automatically creating a secret)

src/zenml/config/secret_reference_mixin.py
```diff
@@ -51,8 +54,11 @@ def __init__(self, **kwargs: Any) -> None:
                 continue
 
             if not secret_utils.is_secret_reference(value):
-                if secret_utils.is_secret_field(field):
-                    logger.warning(
+                if (
+                    secret_utils.is_secret_field(field)
+                    and warn_about_plain_text_secrets
+                ):
+                    logger.debug(
```
schustmi, line 61:
```suggestion
                    logger.warning(
```

schustmi (changes requested):

schustmi (approved):

PR #1415, opened 2023-03-14
Add tests for the model registry

safoinme:
## Describe changes
Adding some test for model registry 
## Pre-requisites

## Types of changes


tests/integration/functional/cli/test_model.py
```diff
@@ -0,0 +1,326 @@
...
+        flavor="test",
+        config={},
+        type=StackComponentType.MODEL_REGISTRY,
+        user=uuid4(),
+        workspace=uuid4(),
+        created=datetime.now(),
+        updated=datetime.now(),
+    )
+
+
```
strickvl, line 192: Could you add a test that list works for model registries as well? (one that shows it's working when we have a model registry, and another that shows the alternative cases of no model registry, and also trying to use / interact with a model registry that doesn't exist?

tests/integration/functional/cli/test_model.py
```diff
@@ -0,0 +1,326 @@
...
+    )
+    # Run the command
+    result = runner.invoke(
+        update_registered_model,
+        ["test_model", "--description", "new_test_description"],
+        obj=concrete_model_registry,
+    )
+
+    # Check the result
+    assert result.exit_code == 0
```
strickvl, line 242: I think you also now need to check that the model was updated. Just the `exit_code` isn't enough.

tests/integration/functional/cli/test_model.py
```diff
@@ -0,0 +1,326 @@
...
+        "get_model",
+        return_value=concrete_registered_models[0],
+    )
+    # Run the command
+    result = runner.invoke(
+        list_registered_model, ["test_model"], obj=concrete_model_registry
+    )
+
+    # Check the result
+    assert result.exit_code == 0
```
strickvl, line 268: I'd add a test to confirm the expected behavior when you try to get a model that doesn't exist.

tests/integration/functional/cli/test_model.py
```diff
@@ -0,0 +1,326 @@
...
+        "list_model_versions",
+        return_value=concrete_registered_model_version,
+    )
+    # Run the command
+    result = runner.invoke(
+        list_model_versions, ["test_model"], obj=concrete_model_registry
+    )
+
+    # Check the result
+    assert result.exit_code == 0
```
strickvl, line 296: Same here. Confirm listing model versions when there aren't any. Also for this test, check the `result.output` to ensure that our model version is mentioned.

tests/integration/functional/cli/test_model.py
```diff
@@ -0,0 +1,326 @@
...
+    )
+
+    # Run the command
+    result = runner.invoke(
+        get_model_version,
+        ["test_model", "--version", "test_version"],
+        obj=concrete_model_registry,
+    )
+    # Check the result
+    assert result.exit_code == 0
```
strickvl, line 326: same here. test that our model version is in `result.output`, and add another test to confirm the failure scenario (when no version exists).

tests/unit/model_registries/test_base_model_registry.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_custom_attributes(self):
+        metadata = ModelRegistryModelMetadata(
+            zenml_version="1.0",
+            custom_attr_1="foo",
+            custom_attr_2="bar",
+        )
+        expected = {"custom_attr_1": "foo", "custom_attr_2": "bar"}
+        assert metadata.custom_attributes == expected
+
+    def test_dict(self):
```
strickvl, line 31: Could maybe add a couple more parts to this where you test the functionality of `exclude_unset` and `exclude_none` too?

tests/unit/model_registries/test_base_model_registry.py
```diff
@@ -0,0 +1,38 @@
...
+
+from zenml.model_registries.base_model_registry import (
+    ModelRegistryModelMetadata,
+)
+
+
+class TestModelRegistryModelMetadata:
+    def test_custom_attributes(self):
+        metadata = ModelRegistryModelMetadata(
+            zenml_version="1.0",
```
strickvl, line 24: Could also maybe just test / show that this is also accepting non-string inputs, as in:

```python
m = ModelRegistryModelMetadata(zenml_version=1.55) # apparently this works fine
```

But then you can confirm / test that the float value I inputted is actually being converted to a string on the other end (`isinstance(m.zenml_version, str)` or something).

strickvl (changes requested):
Looking good. Could probably be improved a bit. Some suggestions in my review.

tests/integration/functional/cli/test_model.py
```diff
@@ -240,6 +240,7 @@ def test_update_model(
 
     # Check the result
     assert result.exit_code == 0
+    assert "updated successfully." in result.output
```
strickvl, line 243: I meant more to test the actual model itself (get it via the `Client()` etc), and not just the output.

PR #1416, opened 2023-03-15
Add metadata to stack components

wjayesh:
## Describe changes
I implemented the addition of metadata to stack components. This PR includes the following capabilities:

- Add metadata at registration time
  ```
  zenml orchestrator register new_orch --flavor=local -m cloud=local
  ```

- Update metadata for existing orchestrator
  ```
  zenml orchestrator update old_orch -m cloud=gcp
  ```

- Remove metadata from a component
  ```
  zenml orchestrator remove-attribute new_orch -m cloud
  ```

## Help Needed

I want that we can list stack components by metadata keys but as of now, this isn't possible because
  a) metadata is currently stored as a byte string in the DB.
  b) filtering can't be performed on bytes.

I have explored creating a HEX type decorator that converts the bytes to and from hex but it fails with some error. Worth exploring but I don't have enough time atm.

## Pre-requisites

## Types of changes


src/zenml/cli/stack_components.py
```diff
@@ -195,6 +195,13 @@ def generate_stack_component_register_command(
         required=True,
         type=str,
     )
+    @click.option(
+        "--metadata",
+        "-m",
+        "metadata",
+        help="Metadata to be associated with the component.",
+        multiple=True,
+    )
```
AlexejPenner, line 204: Is there any place where the expectations for what to input here are communicated to the user? Maybe you should add this to the xamples in src/zenml/cli/__init__.py so it is added to the api_docs. Also maybe the help messages should contain an example.

src/zenml/cli/stack_components.py
```diff
@@ -228,6 +237,14 @@ def register_stack_component_command(
...
         )
 
+        parsed_metadata = None
+        # split metadata by "=" and create a dict
+        if metadata:
+            parsed_metadata = {}
+            for m in metadata:
+                key, value = m.split("=")
+                parsed_metadata[key] = value
+
```
AlexejPenner, line 247: This probably could be extracted into a utils function

src/zenml/models/component_models.py
```diff
@@ -126,6 +130,10 @@ class ComponentFilterModel(ShareableWorkspaceScopedFilterModel):
         default=None, description="Workspace of the stack component"
     )
     user_id: Union[UUID, str] = Field(None, description="User of the stack")
+    metadata_values: bytes = Field(
```
schustmi, line 133: I think having the metadata as `bytes` on the filter model makes it almost unusable?

I would assume I need to pass in the same bytes of all the component metadata to make it work as a filter? Much more useful would be if I could just check for a single (or multiple) key-value pairs, e.g. if the metadata attribute would also be `Dict[str, str]` here. The filter logic implementation would be much more complicated though, but useful to have also for future features.

src/zenml/client.py
```diff
@@ -2022,8 +2033,12 @@ def list_stack_components(
             id=id,
             created=created,
             updated=updated,
+            metadata_values=base64.b64encode(
+                json.dumps(parsed_metadata).encode("utf-8")
+            ) if metadata else None,
         )
         component_filter_model.set_scope_workspace(self.active_workspace.id)
+        breakpoint()
```
schustmi, line 2041:
```suggestion
```

src/zenml/client.py
```diff
@@ -1987,6 +1988,7 @@ def list_stack_components(
         type: Optional[str] = None,
         workspace_id: Optional[Union[str, UUID]] = None,
         user_id: Optional[Union[str, UUID]] = None,
+        metadata: Optional[List[str]] = None,
```
schustmi, line 1991: This should probably be `Dict[str, str]` as in the other places?

src/zenml/cli/stack_components.py
```diff
@@ -404,6 +449,7 @@ def remove_attribute_stack_component_command(
                     name_id_or_prefix=name_id_or_prefix,
                     component_type=component_type,
                     configuration={k: None for k in args},
+                    metadata={k: None for k in metadata},
```
AlexejPenner, line 452: Did you test that this works?

src/zenml/zen_stores/schemas/component_schemas.py
```diff
@@ -48,6 +67,7 @@ class StackComponentSchema(ShareableSchema, table=True):
     type: StackComponentType
     flavor: str
     configuration: bytes
+    metadata_values: Optional[bytes] = Field(sa_column=Column(HexByteString))
```
schustmi, line 70: I would call this simply `metadata` to use the same name as the model, which is how we usually handle these.

src/zenml/zen_stores/schemas/component_schemas.py
```diff
@@ -40,6 +40,25 @@
...
+    impl = String
+
+    def process_bind_param(self, value, dialect):
+        breakpoint()
+        if not isinstance(value, bytes):
+            raise TypeError("HexByteString columns support only bytes values.")
+        return value.hex()
+
+    def process_result_value(self, value, dialect):
+        breakpoint()
```
schustmi, line 58:
```suggestion
```

src/zenml/zen_stores/schemas/component_schemas.py
```diff
@@ -40,6 +40,25 @@
...
+from sqlalchemy.types import String, TypeDecorator
+
+
+class HexByteString(TypeDecorator):
+    """Convert Python bytestring to string with hexadecimal digits and back for storage."""
+
+    impl = String
+
+    def process_bind_param(self, value, dialect):
+        breakpoint()
```
schustmi, line 52:
```suggestion
```

src/zenml/zen_stores/schemas/component_schemas.py
```diff
@@ -40,6 +40,25 @@
     from zenml.zen_stores.schemas import ScheduleSchema
 
 
+from sqlalchemy.types import String, TypeDecorator
+
+
+class HexByteString(TypeDecorator):
```
schustmi, line 46: Can you explain why this was necessary? We had byte columns in the DB before right?

src/zenml/cli/utils.py
```diff
@@ -1388,15 +1419,28 @@ def inner_decorator(func: F) -> F:
...
             if k not in filter_model.CLI_EXCLUDE_FIELDS:
-                options.append(
-                    click.option(
-                        f"--{k}",
-                        type=str,
-                        default=v.default,
-                        required=False,
-                        help=create_filter_help_text(filter_model, k),
+                # if k is metadata, add option with multiple values
+                if k == "metadata_values":
```
schustmi, line 1423: Can we make this more generic? I envision something that adds a `multiple=True` option in case the filter model property is a collection of some sort (Dict, List)

schustmi (changes requested):

AlexejPenner (changes requested):
I think in the long run we probably should solve this with a table for arbitrary metadata that is connected to the stack component table. This would also allow for filtering and searching. FOr the time being, here are a few smaller comments to consider.

htahir1 (commented):
Dont we have the metadata concept implemented by @fa9r that can be reused here?

src/zenml/zen_stores/schemas/component_schemas.py
wjayesh: Calling it `metadata` would cause a clash with a field by the same name in `SQLModel` and so I added the extra `_values` to it.

src/zenml/client.py
wjayesh: You're right. For now, I have removed it altogether from the filtering logic.

src/zenml/cli/stack_components.py
wjayesh: Yup, I did.

src/zenml/cli/stack_components.py
wjayesh: Good catch, I've added it as a fn.

src/zenml/cli/stack_components.py
wjayesh: Good idea. I have added the expectation to both the `__init__` file and the `help` message.

src/zenml/models/component_models.py
wjayesh: You're right, having the metdata as bytes doesn't allow filtering. I agree that this should be a Dict instead; with logic implemented for filtering. For now, I have removed the option to filter by metadata altogether and will revisit this problem if I have time.

src/zenml/zen_stores/schemas/component_schemas.py
wjayesh: Yes, I was hoping to experiment with storing hex strings instead but yes, doesn't achieve anything significant towards the goal of easy filtering. I have removed it.

src/zenml/cli/utils.py
schustmi: Can we also remove this filtering  logic here in the CLI then?

schustmi (changes requested):
One small request, otherwise looks good to me!

AlexejPenner (commented):
Looks good to me, one open question for me is more on the naming front. Shouldn't we call this something like `labels` which is already a known concept from the gloud providers?

schustmi:
@wjayesh Ping me once it's updated to `labels` and I'll give it another look

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -562,6 +562,7 @@ def _create_default_stack(
                 type=StackComponentType.ORCHESTRATOR,
                 flavor="local",
                 configuration={},
+                metadata={"cloud": "local"},
```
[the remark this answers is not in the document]
wjayesh: okay, I can remove it since it isn't necessary.

wjayesh:
@schustmi done!

src/zenml/cli/__init__.py
```diff
@@ -283,6 +289,12 @@
 zenml orchestrator register ORCHESTRATOR_NAME --flavor=ORCHESTRATOR_FLAVOR [--ORCHESTRATOR_OPTIONS]
 ```
 
+You can also add any label to your stack component using the `--label` or `-m` flag:
```
schustmi, line 292:
```suggestion
You can also add any label to your stack component using the `--label` or `-l` flag:
```

src/zenml/cli/__init__.py
```diff
@@ -283,6 +289,12 @@
...
+
+```bash
+zenml orchestrator register ORCHESTRATOR_NAME --flavor=ORCHESTRATOR_FLAVOR -m key1=value1 -m key2=value2
```
schustmi, line 295:
```suggestion
zenml orchestrator register ORCHESTRATOR_NAME --flavor=ORCHESTRATOR_FLAVOR -l key1=value1 -l key2=value2
```

schustmi (changes requested):

schustmi (approved):

strickvl (changes requested):
@wjayesh one thing I think this PR is really missing is tests. Also docs should be updated accordingly to reflect the new changes. Basically CLI docs, and then also anywhere someone would deploy a component we can probably add an *info* box or something saying that they can add/register this kind of metadata?

PR #1417, opened 2023-03-15
Add option to get deployment outputs

wjayesh:
## Describe changes
I implemented a new function for the server deployer to get outputs related to Zen Server deployments.
Also included is the implementation for the Terraform provider.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/server.py
+++ b/src/zenml/cli/server.py
@@ -505,6 +505,32 @@ def deploy(
                 )
 
 
+# get outputs
+@cli.command(
+    "get-deploy-outputs",
+    help="Get the outputs of a ZenML server deployment.",
+)
+@click.argument(
+    "name",
+    required=True,
+    type=click.STRING,
+)
+@click.option(
+    "--output",
+    help="The name of the output to get.",
+    required=False,
+)
+def get_outputs(name: str, output: Optional[str] = None) -> Dict[str, str]:
+    """Get the outputs of the ZenML server deployment."""
+    from zenml.zen_server.deploy.deployer import ServerDeployer
+
+    deployer = ServerDeployer()
+    outputs = deployer.get_deployment_outputs(name, output)
+
+    cli_utils.declare("Outputs:")
+    cli_utils.declare(outputs)
+
+
 @cli.command(
     "destroy", help="Tear down and clean up the cloud ZenML deployment."
 )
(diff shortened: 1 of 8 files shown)


src/zenml/zen_server/deploy/deployer.py
```diff
@@ -418,3 +418,34 @@ def get_server_logs(
...
+        # does not exist
+        server = self.get_server(server_name)
+
+        provider_name = server.config.provider.value
+        provider = self.get_provider(server.config.provider)
+
+        # if the provider does not implement this method, we return an empty
+        # dict
+        if not hasattr(provider, "get_deployment_outputs"):
+            return {}
```
safoinme, line 445: Would it be worth to change this try/except and do some sort of logging that provider doesn't have `get_deployment_outputs`
```suggestion
        if not hasattr(provider, "get_deployment_outputs"):
            return {}
```

src/zenml/cli/server.py
```diff
@@ -505,6 +505,32 @@ def deploy(
...
+    "name",
+    required=True,
+    type=click.STRING,
+)
+@click.option(
+    "--output",
+    help="The name of the output to get.",
+    required=False,
+)
+def get_outputs(name: str, output: Optional[str] = None) -> Dict[str, str]:
```
strickvl, line 523: What if someone wants multiple outputs? I think you'll need a `multiple=True` here, and then a way to pass them into the `ServerDeployer` method to get multiple outputs.

src/zenml/services/terraform/terraform_service.py
```diff
@@ -356,6 +356,25 @@ def get_logs(
             "This method is not available for Terraform services."
         )
 
+    def get_outputs(self, output: Optional[str] = None) -> Dict[str, Any]:
```
strickvl, line 359: This probably needs to account for the fact that people will want to get multiple outputs at any one time. (As mentioned above). So it could accept an optional list of outputs that people want, and then output a dictionary containing whatever outputs were requested (or just all of them as the default)

strickvl (changes requested):
This looks good overall. The main thing I think it should accommodate is the option to get multiple outputs at the same time vs how it's currently implemented (which is to only allow getting a single output for each CLI command executed).

Was also trying to think about which parts of this can be tested, but that's a bit harder to implement, I think.

strickvl:
@wjayesh is this still needed?

PR #1418, opened 2023-03-15
Add cloud orchestrator warning message

strickvl:
Added a warning message for when running cloud orchestrators to warn them that pipeline runs will take a while to show up on the ZenML dashboard.

## Pre-requisites

## Types of changes


src/zenml/pipelines/base_pipeline.py
```diff
@@ -562,6 +563,17 @@ def run(
             finally:
                 constants.SHOULD_PREVENT_PIPELINE_EXECUTION = False
 
+            if stack.orchestrator.config.is_remote:
+                cli_utils.warning(
```
schustmi, line 567: I don't think we should use `cli_utils` outside of the CLI, `logger.warning` should do the same thing right?

Some other points:
- The same applies to the `LocalDockerOrchestrator`, this is actually the case for all orchestrators except the local one.
- For some orchestrators, we have a `synchronous` config attribute which would also make this warning message unnecessary if set.
- The `up to twenty minutes` seems quite scary and AFAIK only applies to the Sagemaker orchestrator, maybe we can leave that out and potentially add a separate log in the `SagemakerOrchestrator.prepare_or_run_pipeline(...)` method?


I have one suggestion that could cover both these cases and could even be used to improve the `dashboard_utils.get_run_url` function (Maybe we could even include this message in the `get_run_url` function?). It uses the fact that we create a `PipelineDeployment` just a few lines above which the run will be associated with (it will be the first run of that deployment).

```python
runs = Client().list_runs(deployment_id=deployment_model.id, ...) # Add more filter options here, we only want the first run if it exists

if runs.items:
  # The run is already registered in the ZenML server and will therefore be shown in the dashboard
  # We can even use runs[0].id for a better dashboard URL
else:
  # Only here the warning message is relevant
```

schustmi (changes requested):

src/zenml/pipelines/base_pipeline.py
```diff
@@ -660,11 +660,32 @@ def run(
...
-            dashboard_utils.print_run_url(
-                run_name=deployment.run_name_template,
-                pipeline_id=pipeline_id,
-            )
+            if (
+                deployment_model
+                and deployment_model.pipeline
+                and deployment_model.stack
+            ):
+                runs = Client().list_runs(
```
schustmi, line 668: I'm guessing since the deployment has a unique UUID, we could make this simpler and just filter by deployment ID (skipping the stack/workspace/pipeline)?
There won't be any functional difference to the current solution however, so I'll leave that up to you.

strickvl: Yeah that was my thought going in, but you did mention that we'd want to filter by other params as well so I thought maybe you were thinking of something that I wasn't. I think it's cleaner without the stack / workspace + pipeline tbh.

strickvl: Ah yes good point. `desc:created` does what we want it to (i.e. gets the most recent run)

schustmi: Not sure if it's relevant (probably there is no possibility that multiple runs exist at that point), but we actually want the first one to get the run that was triggered by the `pipeline.run()` call.

strickvl: AFAICS what the code now does get that first one. Actually it should get that just by specifying the `deployment_model.id`, unless I'm missing something?

PR #1425, opened 2023-03-20
Update `pytorch-lightning` to support `2.0`

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/integrations/pytorch_lightning/__init__.py
```diff
@@ -21,7 +21,7 @@ class PytorchLightningIntegration(Integration):
     """Definition of PyTorch Lightning integration for ZenML."""
 
     NAME = PYTORCH_L
-    REQUIREMENTS = ["pytorch_lightning<2.0.0"]
+    REQUIREMENTS = ["pytorch_lightning"]
```
strickvl, line 24: Does this mean that our integration works for both pre-2.0 as well as post-2.0 releases?

src/zenml/integrations/pytorch_lightning/materializers/pytorch_lightning_materializer.py
```diff
@@ -40,15 +42,16 @@ def load(self, data_type: Type[Any]) -> Trainer:
             A PyTorch Lightning trainer object.
         """
         super().load(data_type)
-        return Trainer(
```
strickvl, line 43: Was this even working before? 🤔

safoinme: I honestly doubt it, I don't see how the trainer will be able to load the trainer without the class definition

src/zenml/integrations/pytorch_lightning/materializers/pytorch_lightning_materializer.py
```diff
@@ -27,28 +29,29 @@
 class PyTorchLightningMaterializer(BaseMaterializer):
     """Materializer to read/write PyTorch models."""
 
-    ASSOCIATED_TYPES = (Trainer,)
+    ASSOCIATED_TYPES = (Module,)
```
stefannica, line 32: This is a breaking change, isn't it ? All existing steps that still return `Trainer` will break now. Is this something you would like to avoid by keeping the old materializer around ? Might also be fine to just drop it, given that it doesn't look like it was working to begin with.

This also makes me wonder if our example doesn't also need fixing.

stefannica: Oh, wait, we don't have a pytorch lighting example.

stefannica (changes requested):

src/zenml/integrations/pytorch_lightning/materializers/pytorch_lightning_materializer.py
safoinme: We don't actually have an example for it, that's why I don't it has been ever used. For that I don't think we need to keep the old one I guess

src/zenml/integrations/pytorch_lightning/materializers/pytorch_lightning_materializer.py
```diff
@@ -27,28 +29,29 @@
...
 
         Returns:
-            A PyTorch Lightning trainer object.
+            A PyTorch Lightning model object.
         """
         super().load(data_type)
-        return Trainer(
-            resume_from_checkpoint=os.path.join(self.uri, CHECKPOINT_NAME)
-        )
+        super().load(data_type)
```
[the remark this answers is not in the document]
safoinme: Sorry just an error, fixed now

stefannica (approved):
Looks good, I don't think we need to keep the previous implementation either.

src/zenml/integrations/pytorch_lightning/__init__.py
htahir1, line 24: Would this change also work in `pytorch_lightning < 2.0.0`?`

safoinme: ~~Yes, it should work, cause we now are defaulting the `pytorch-lightning` materializer to the way the original native `pytorch` materializer work.~~
Well we can actually just delete this whole materializer, we already have `Pytorch` one

pyproject.toml
```diff
@@ -49,7 +49,7 @@ include = [
 zenml = "zenml.cli.cli:cli"
 
 [tool.poetry.dependencies]
-pandas = "^1.1.5"
+pandas = "^>=1.1.5"
```
stefannica, line 52:
```suggestion
pandas = ">=1.1.5"
```

stefannica (changes requested):

PR #1426, opened 2023-03-20
Fix some pipeline bugs

schustmi:
## Describe changes
This fixes:
- not being able to use the same artifact for multiple inputs of a step
- having to call all steps inside the pipeline function
 
## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/728c6369cfaa_add_name_column_to_input_artifact_pk.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    engine_name = op.get_bind().engine.name
+    if engine_name == "sqlite":
+        constraint_name = "pk_step_run_input_artifact"
+    elif engine_name == "mysql":
+        constraint_name = "PRIMARY"
```
fa9r, line 24: Wow this is much easier than for the foreign keys, nice. I assume you tested this works, right?

fa9r (approved):
That migration looks very painful. I wanted to suggest introducing constraint names for ALL pks in the DB while you're at it, but nevermind 😄 

Changes LGTM, very nice that you even added tests for the use cases that this enables 👍

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -301,7 +301,7 @@ class StepRunInputArtifactSchema(SQLModel, table=True):
     artifact: ArtifactSchema = Relationship(
         back_populates="input_to_step_runs"
     )
-    name: str
+    name: str = Field(nullable=False, primary_key=True)
```
stefannica, line 304: Haven't fully grasped the meaning of this yet, but just wanted to point out that there is also the `StepRunOutputArtifactSchema` that might require a similar change.

stefannica: Had more time to think about this and find an answer by myself. There is no way to use the same name for multiple output artifacts generated by the same step run, is there ? First, because you can't define a function to return e.g. `Output[first = dict, second = str, first = dict]` and second, the whole idea makes no practical sense (why would you even want to return the same artifact multiple times ?).

So forget I asked :)

src/zenml/config/compiler.py
```diff
@@ -99,6 +99,7 @@ def compile(
                 pipeline_success_hook_source=pipeline.configuration.success_hook_source,
             )
             for name, step in self._get_sorted_steps(steps=pipeline.steps)
+            if step._has_been_called
```
stefannica, line 102: This was surprisingly simple to fix :+1:

stefannica (approved):
Thanks ! These bugs were like lego blocks: seemingly small and insignificant... and then you step on one and it hurts like a sob.

src/zenml/zen_stores/migrations/versions/728c6369cfaa_add_name_column_to_input_artifact_pk.py
schustmi: I tested `upgrade -> downgrade -> re-upgrade` on both MySQL and SQLite which both seemed to work (after an hour or so of trying to get this right for both to be fair 😄)

src/zenml/zen_stores/schemas/step_run_schemas.py
schustmi: You got it exactly right 👍 

Maybe some more explanation what my change does for the input table: Previously we used just the `step_run_id` and `artifact_id` as primary keys for the table that relates steps with their input artifacts. If a step now had two inputs (with different names) that refer to the same artifact, only one of those would be stored in the database, and when then trying to resolve the inputs for the step we would get the `KeyError` that no artifact for input X exists. This PR simply includes the `name` column in the primary key.

PR #1428, opened 2023-03-21
Fix hook parser

strickvl:
Noticed an error when running the hooks example (also on the video). Looked into it and seems that we're passing a `str` type as the `arg_type` always and so were always hitting the `TypeError` exception. Not sure if this is *the* solution, but proposing it as a possible solution.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/hooks/hook_validators.py
+++ b/src/zenml/hooks/hook_validators.py
@@ -67,10 +67,6 @@ def resolve_and_validate_hook(hook_func: "HookSpecification") -> str:
                     BaseException,
                     BaseParameters,
                     StepContext,
-                    # Have to do string version for TYPE_CHECKING
-                    "BaseException",
-                    "BaseParameters",
-                    "StepContext",
                 ):
                     raise ValueError(
                         "Hook parameters must be of type `BaseException`, `BaseParameters`, "
--- a/src/zenml/orchestrators/step_runner.py
+++ b/src/zenml/orchestrators/step_runner.py
@@ -315,6 +315,11 @@ class StepRunner:
             arg_type = annotations.get(arg, None)
             arg_type = resolve_type_annotation(arg_type)
 
+            if isinstance(arg_type, str):
+                arg_type = globals().get(
+                    arg_type, locals().get(arg_type, None)
+                )
+
             # Parse the parameters
             if issubclass(arg_type, BaseParameters):
                 step_params = arg_type.parse_obj(self.configuration.parameters)
--- a/src/zenml/orchestrators/step_runner.py
+++ b/src/zenml/orchestrators/step_runner.py
@@ -531,7 +531,7 @@ class StepRunner:
             output_materializers: The output materializers of the step.
         """
         try:
-            hook = source_utils.import_class_by_path(hook_source)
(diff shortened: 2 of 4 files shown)


src/zenml/orchestrators/step_runner.py
```diff
@@ -315,6 +315,11 @@ def _parse_hook_inputs(
             arg_type = annotations.get(arg, None)
             arg_type = resolve_type_annotation(arg_type)
 
+            if isinstance(arg_type, str):
```
schustmi, line 318: We don't allow string annotations in any other place for probably the same reason, is there any specific reason we require them for hooks?

This lookup in globals/locals seems pretty unreliable to me, what if a the class that is annotated is not in one of those dicts?

strickvl:
> Thanks for catching this! Could you let me know what the error is exactly (Is it an exception or a warning or something?).

ScreenShot 2023-03-21 at 12 38 43

@htahir1 This is the message that is output when you run it without the fix.

@schustmi I didn't dive deep into the hooks code so not sure where the string annotations come from, but I'll take a look and see if we can just improve it at the source.

htahir1:
So the hooks never worked at all if i understood this correctly?

strickvl:
> So the hooks never worked at all if i understood this correctly?

I think not.

src/zenml/orchestrators/step_runner.py
```diff
@@ -531,8 +531,9 @@ def load_and_run_hook(
             output_materializers: The output materializers of the step.
         """
         try:
-            hook = source_utils.load_source_path(hook_source)
+            hook = source_utils.import_class_by_path(hook_source)
```
schustmi, line 534: Just out of curiosity, can you explain why this change necessary here?

schustmi (approved):

src/zenml/orchestrators/step_runner.py
strickvl: Good catch. Actually it might not be any more. I can probably just revert it back. I changed this early on when we were grabbing objects from globals and locals. I'll revert it and rerun the tests.

strickvl: Indeed the change was unnecessary with the later fix we made.

src/zenml/hooks/hook_validators.py
```diff
@@ -67,10 +67,6 @@ def resolve_and_validate_hook(hook_func: "HookSpecification") -> str:
                     BaseException,
                     BaseParameters,
                     StepContext,
-                    # Have to do string version for TYPE_CHECKING
-                    "BaseException",
```
htahir1, line 71: But I added these just for the case.. Im very unsure about why we revert this now but if @schustmi and @strickvl are confident id say merge this PR in...

schustmi: That case doesn't really work in a nice way though: What if I pass my custom `BaseParameters` subclass as the type annotation there, it would already break down. I think we should just not allow these for now until we implement an actual solution to support string annotations in our steps/hooks.

htahir1 (approved):

PR #1429, opened 2023-03-21
Move token to the db

AlexejPenner:
## Describe changes
The token moves from global config to the db. What do you think, merge into your branch?

## Pre-requisites

## Types of changes


fa9r (approved):
Great change, just tried it out, works like a charm 😍

The only question I'd have is whether it is an issue that we store the token as plain text in the DB now. IMO it should be fine since it's a temporary token and also only used for the hub itself, not GitHub or other apps. Maybe @stefannica has some suggestions on this?

Anyways, this is strictly better than the previous solution of storing the token as plain text in the global config so let's just merge it in 👍

PR #1430, opened 2023-03-21
ZenML Failure Hook for OpenAI ChatGPT fixes

strickvl:
I implemented a failure hook that passes the stacktrace and exception to ChatGPT (`gpt-3.5-turbo` model) to output suggestions on how to fix the failing step. It uses `rich`'s traceback to get more context on the failing code and passes those suggestions through to a registered alerter component. Output looks something like this:

failure_alerter

Users have to pass in their OpenAI API key as a secret. It's a bit clunky, but hard to validate or pass down configuration any other way since this failure hook behaviour isn't represented by a stack component.

## Pre-requisites

## Types of changes


src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
```diff
@@ -0,0 +1,83 @@
...
+        params: Parameters used in the step.
+        exception: Original exception that lead to step failing.
+    """
+    if context.stack and context.stack.alerter:
+        output_captured = io.StringIO()
+        original_stdout = sys.stdout
+        sys.stdout = output_captured
+        console = Console()
+        # show_locals set to False to limit tokens sent to OpenAI
+        console.print_exception(show_locals=False)
```
htahir1, line 50: Do we need to use console here? Not sure why its neccessary and if it is we should use it in the other standard hook as well i suppose?

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
```diff
@@ -0,0 +1,83 @@
...
+
+        sys.stdout = original_stdout
+        rich_traceback = output_captured.getvalue()
+
+        response = openai.ChatCompletion.create(  # type: ignore[no-untyped-call]
+            model="gpt-3.5-turbo",
```
htahir1, line 56: I wonder if there is a way to parameterize this... not sure there is an easy way

htahir1 (changes requested):
What a beautiful feature! I really love this sort of innovation. Kudos!

Other than my comments, some minor testing might be worth taking a look?

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
strickvl: It's the rich output console which gives a much more detailed traceback. Currently the standard hook is printing just the exception type and not the traceback. Not sure if we'd want to output the full traceback? In any case, we only need the `rich` console if we're getting that full traceback.

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
strickvl: I thought about using a `partial` here, but I think no way then to validate the hook function without either ignoring some annotations. I think we can just offer it as an example of how you could do this, and then people can of course make their own.

Perhaps `hook` functions would be in the hub as well at some point? For now, I'll make a standard and a gpt-4 version of the hook.

safoinme (changes requested):

docs/book/advanced-guide/pipelines/hooks.md
```diff
@@ -116,3 +116,41 @@ from zenml.hooks import alerter_success_hook, alerter_failure_hook
...
+export OPENAI_API_KEY='<YOUR_KEY>'
+```
+
+Then, you can use the hook in your pipeline:
+
+```python
+from zenml.integration.openai.hooks import openai_chatgpt_alerter_failure_hook
+
+@step(on_failure=openai_alerter_failure_hook)
+def my_step(...):
```
safoinme, line 145:
```suggestion
from zenml.integration.openai.hooks import openai_chatgpt_alerter_failure_hook

@step(on_failure=openai_chatgpt_alerter_failure_hook)
def my_step(...):
```

safoinme (changes requested):

safoinme (approved):
LGTM

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
htahir1: Id say we should be consistent, so if we do this here, then lets update the other one

htahir1 (commented):
Apart from my one comment above, how about adding in these blocks that @soubenz has added in this PR? https://github.com/zenml-io/zenml/pull/1402/files

They might fit in well in all the standard hooks. Just a small suggestion

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
strickvl: Ok so we'll have the full `rich` exception now instead of the standard Python one. Will make the change.

strickvl: Done

strickvl:
> Apart from my one comment above, how about adding in these blocks that @soubenz has added in this PR? https://github.com/zenml-io/zenml/pull/1402/files
> 
> They might fit in well in all the standard hooks. Just a small suggestion

We can probably just wait until #1402 has been merged and then update as appropriate?

htahir1 (changes requested):
Ah caught one more thing. Lets enforce that the `OPENAI_API_KEY` has to be set for these hooks?

Also we should merge in your other hooks fix PR before merging this one in to avoid conflcits (You have approvals so should be good to go the other one)

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
```diff
@@ -39,7 +40,9 @@ def openai_alerter_failure_hook_helper(
         exception: The exception that was raised.
         model_name: The OpenAI model to use for the chatbot.
     """
-    if context.stack and context.stack.alerter:
+    openai_api_key = os.environ.get("OPENAI_API_KEY")
+
+    if context.stack and context.stack.alerter and openai_api_key:
```
htahir1, line 45: do we need to check the alerter?

strickvl: If we don't check it and you don't have an alerter, later on when we run `context.stack.alerter.post(message)` this will raise an error.

htahir1 (approved):

PR #1432, opened 2023-03-22
Update `mypy>=1.1.1` and fix new errors

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


tests/harness/model/deployment.py
```diff
@@ -51,7 +51,7 @@ class Config:
         """Pydantic configuration class."""
 
         validate_assignment = True
-        extra = Extra.allow
+        extra = "allow"
```
schustmi, line 54: Why is this change necessary?

src/zenml/client.py
```diff
@@ -1040,7 +1034,7 @@ def update_role(
             name_id_or_prefix=name_id_or_prefix, allow_name_prefix_match=False
         )
 
-        role_update = RoleUpdateModel()
+        role_update = RoleUpdateModel(name=new_name or role.name)  # type: ignore
```
schustmi, line 1037: Why do we need these type ignores for some of the update model creation now?

schustmi (changes requested):

tests/harness/model/deployment.py
safoinme: I think `extra` requires a str instead of enum. that change was actually responsible for fixing some of `mypy` errors

src/zenml/client.py
safoinme: There are some other non `optional` attributes, and because `mypy` seems unable to get the `@update_model`  decorator that makes every attribute actually optional

schustmi: But the `UserUpdateModel` should be the same right, it also uses the `@update_model` decorator and we seem to be able to create it without the ignore?

Btw not trying to say that you did anything wrong here, I'm just trying to figure out why mypy/the plugin is working so inconsistently for these update models (and also the `extra="allow"` thing).

tests/harness/model/deployment.py
safoinme: Or maybe I changed something else before this and I thought it was this causing those issues haha😅

schustmi (approved):

src/zenml/zen_server/routers/users_endpoints.py
```diff
@@ -315,6 +318,7 @@ def email_opt_in_response(
 
     if str(auth_context.user.id) == str(user_name_or_id):
         user_update = UserUpdateModel(
+            name=user_name_or_id,
```
stefannica, line 321:
```suggestion
            name=user.name,
```

stefannica (changes requested):
You have my sympathies for being tasked with cleaning up our collective mess and my respect for having done it so diligently !

A few minor tweaks here and there, but otherwise looks good. Please double-check my suggestions before applying them, I'm not 100% that they are correct in 2-3 places.

src/zenml/models/team_role_assignment_models.py
```diff
@@ -84,8 +86,8 @@ class TeamRoleAssignmentRequestModel(
     """Request model for role assignments using UUIDs for all entities."""
 
     workspace: Optional[UUID] = Field(
-        None, title="The workspace that the role is limited to."
+        title="The workspace that the role is limited to."
     )
-    team: UUID = Field(None, title="The user that the role is assigned to.")
+    team: UUID = Field(title="The user that the role is assigned to.")
```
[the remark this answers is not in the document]
safoinme: I think this should kept required, as the `TeamRoleAssignmentRequestModel` is about requesting role assignments for a team which means that must be provided and also there is a function that expect that to be always set so if this changes to optional it complains:

```
error: Argument 1 to "_get_team_schema" of "SqlZenStore" has incompatible type "Optional[UUID]"; expected "Union[str, UUID]"  [arg-type]
```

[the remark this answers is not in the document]
safoinme: By the way the same story goes for other comment about `user`

src/zenml/models/user_role_assignment_models.py
```diff
@@ -84,8 +86,8 @@ class UserRoleAssignmentRequestModel(
     """Request model for role assignments using UUIDs for all entities."""
 
     workspace: Optional[UUID] = Field(
-        None, title="The workspace that the role is limited to."
+        title="The workspace that the role is limited to."
     )
-    user: UUID = Field(None, title="The user that the role is assigned to.")
+    user: UUID = Field(title="The user that the role is assigned to.")
```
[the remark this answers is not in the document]
safoinme: I think just like team, this should be require, also here:

```
error: Argument 1 to "_get_user_schema" of "SqlZenStore" has incompatible type "Optional[UUID]"; expected "Union[str, UUID]"  [arg-type]
```

stefannica (approved):

PR #1459, opened 2023-04-03
chore: update teams.yml

Cahllagerfeld:


strickvl (approved):
Looks good, pending license etc

PR #1460, opened 2023-04-04
Kubernetes Orchestrator Improvements

fa9r:
## Describe changes
Various minor improvements to the k8s orchestrator:
- `service_account_name` can now be specified as a setting
- The orchestrator can now be started from within other k8s pods
- [Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) can now be assigned to pods using `KubernetesPodSettings.annotations`

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubeflow/utils.py
+++ b/src/zenml/integrations/kubeflow/utils.py
@@ -61,3 +61,7 @@ def apply_pod_settings(
     resource_limits = settings.resources.get("limits") or {}
     for name, value in resource_limits.items():
         container_op.add_resource_limit(name, value)
+
+    annotations = settings.annotations or {}
+    for name, value in annotations.items():
+        container_op.add_pod_annotation(name, value)
--- a/src/zenml/integrations/kubeflow/utils.py
+++ b/src/zenml/integrations/kubeflow/utils.py
@@ -62,6 +62,5 @@ def apply_pod_settings(
     for name, value in resource_limits.items():
         container_op.add_resource_limit(name, value)
 
-    annotations = settings.annotations or {}
-    for name, value in annotations.items():
+    for name, value in settings.annotations.items():
         container_op.add_pod_annotation(name, value)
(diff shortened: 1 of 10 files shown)


src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -36,12 +36,16 @@ class KubernetesOrchestratorSettings(BaseSettings):
...
+        service_account_name: Name of the service account to use for the
+            orchestrator pod. If not provided, a new service account with "edit"
+            permissions will be created.
         pod_settings: Pod settings to apply.
     """
 
     synchronous: bool = False
     timeout: int = 0
 
+    service_account_name: Optional[str] = None
```
schustmi, line 48: Is this something relevant for other Kubernetes-based orchestrators and should therefore be in the `KubernetesPodSettings` instead?

fa9r: I just checked and couldn't find any flavors that have a service account as a setting. However, the following flavors have a service account as a config value:
- `KserveModelDeployer`
- `KanikoImageBuilder`
- `VertexOrchestrator`
- `KubernetesSparkStepOperator`

Since these are all different stack component types, I don't really see a way to unify this.

I'm also not sure how exactly it works for kfp or other k8s-based orchestrators. My guess would be that you already need to have the service account set up to deploy the tool in the first place. Perhaps you or @stefannica know this better?

src/zenml/integrations/kubeflow/utils.py
```diff
@@ -61,3 +61,7 @@ def apply_pod_settings(
     resource_limits = settings.resources.get("limits") or {}
     for name, value in resource_limits.items():
         container_op.add_resource_limit(name, value)
+
+    annotations = settings.annotations or {}
+    for name, value in annotations.items():
```
schustmi, line 66:
```suggestion
    for name, value in settings.annotations.items():
```
Seems to be a non-optional Dict and this should work as well right?

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
schustmi: Yep I just checked and it seems not possible to set this for KFP (and therefore our other kubenertes-based orchestrators), so the solution you implemented looks good 👍

schustmi (changes requested):

stefannica (approved):
LGTM

schustmi (approved):

schustmi:
Love the tests :)

PR #1462, opened 2023-04-04
Apply formatting fixes to `/scripts`

strickvl:
If you run our `format.sh` script without passing in arguments, it ignores the `scripts/` folder, but if you run it passing in a `.` as the folder to check, it always tries to autofix this file. It often ends up accidentally in PRs, having to be removed later etc, so I've just applied the fixes now so this doesn't happen any more.

## Pre-requisites

## Types of changes

### DIFF
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 set -x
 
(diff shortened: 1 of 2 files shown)


fa9r (approved):
Nice, IMO these should be formatted anyways, should we even add `scripts` to the formatting SRC?

strickvl:
@fa9r good point. added it in.

PR #1464, opened 2023-04-04
Fix `BuiltInContainerMaterializer` for subtypes and non-built-in types

fa9r:
## Describe changes
I fixed two issues of the `BuiltInContainerMaterializer`:
- Elements would previously be loaded as the data type that the corresponding materializer is registered with. E.g., a pydantic model `MyModel(BaseModel)` would be loaded as `BaseModel` since only `BaseModel` is in the `ASSOCIATED_TYPES` of the `PydanticMaterializer`. This is obviously problematic since some functionality might get lost.
- Elements that are not handled by built-in materializers could previously sometimes not be loaded because the default materializer registry might no longer have the respective materializer in it. To fix this, we now save the path of each materializer as part of the list metadata and load the materializer class directly from source.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/materializers/built_in_materializer.py
+++ b/src/zenml/materializers/built_in_materializer.py
@@ -306,7 +306,7 @@ class BuiltInContainerMaterializer(BaseMaterializer):
                     element = materializer.load(type_)
                     outputs.append(element)
 
-            # New format for zenml >= 0.37.1
+            # New format for zenml > 0.37.0
             elif isinstance(metadata, list):
                 for entry in metadata:
                     path_ = entry["path"]
--- a/src/zenml/materializers/built_in_materializer.py
+++ b/src/zenml/materializers/built_in_materializer.py
@@ -23,7 +23,7 @@ from zenml.materializers.base_materializer import BaseMaterializer
 from zenml.materializers.default_materializer_registry import (
     default_materializer_registry,
 )
-from zenml.utils import yaml_utils
+from zenml.utils import source_utils, yaml_utils
 
 if TYPE_CHECKING:
     from zenml.metadata.metadata_types import MetadataType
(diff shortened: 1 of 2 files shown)


src/zenml/materializers/built_in_materializer.py
```diff
@@ -294,20 +294,40 @@ def load(self, data_type: Type[Any]) -> Any:
...
+                for path_, type_str in zip(
+                    metadata["paths"], metadata["types"]
+                ):
+                    type_ = find_type_by_str(type_str)
+                    materializer_class = default_materializer_registry[type_]
+                    materializer = materializer_class(uri=path_)
+                    element = materializer.load(type_)
+                    outputs.append(element)
+
+            # New format for zenml >= 0.37.1
```
schustmi, line 309:
```suggestion
            # New format for zenml > 0.37.0
```

Who knows, it might be 0.38.0 😉

tests/unit/materializers/test_built_in_materializer.py
```diff
@@ -129,3 +135,81 @@ def test_none_values():
...
+        default_materializer_registry,
+    )
+
+    example = [CustomType(), CustomSubType()]
+    with TemporaryDirectory() as artifact_uri:
+        materializer = BuiltInContainerMaterializer(uri=artifact_uri)
+
+        # Container materializer should find materializer for both elements in
+        # the default materializer registry.
+        mocker.patch.object(
```
schustmi, line 193: Shouldn't this be in the registry anyway as we inherit from `BaseMaterializer` with the associated type set?

fa9r: Yeah, true, it also works without this.

schustmi (approved):

strickvl (approved):
LGTM!

PR #1471, opened 2023-04-05
Allow secret values replacement in REST API PUT

stefannica:
## Describe changes
Change the default behavior of `PUT /api/v1/secrets/{secret_id}` to replace all values of an existing secret with those passed in the request body, to facilitate a simpler dashboard implementation for updating/deleting keys.

The ZenML python client is still using the previous behavior that only patches the existing secret. The differentiation is made through a `patch_values` request parameter.

This PR also fixes a bug that reported null values instead of empty string values for secret values that were set to the empty string.

## Pre-requisites

## Types of changes


AlexejPenner (approved):
:seal:

PR #1473, opened 2023-04-05
Fix two steps race condition

safoinme:
## Describe changes
This is just an initial fix on how to avoid the race condition since we already catching `EntityExistsError` which should raise if the pipeline run we try to add to the DB exists, adding `IntegrityError` can also catch the error causing the race condition since it tries to create new raw with a run id that already exists which breaks the unique constraints. 

##Question for reviewers :
Adding an exception for `IntegrityError` may not be the best approach, since it could be raised for other broken constraints should it be also verified that it matches `IntegrityError: UNIQUE constraint failed: pipeline_run.id` or we want different approach were we have some pre-verification than try-catch 

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -40,7 +40,12 @@ import pymysql
 from pydantic import root_validator, validator
 from sqlalchemy import asc, desc, func, text
 from sqlalchemy.engine import URL, Engine, make_url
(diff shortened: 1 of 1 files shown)


fa9r (approved):
LGTM. IMO catching general `IntegrityError` is fine.

schustmi (approved):
Unless there are more specific error subclasses, this LGTM as well

PR #1492, opened 2023-04-14
Validate slack alerter credentials on registration

yasht-twri2k1:
Two new options slack-token and slack-channel are added to the CLI, which are used to prompt the user to enter the Slack bot user token and channel ID during registration time.

## Describe changes

## Pre-requisites

## Types of changes

 - [ x] New feature (non-breaking change which adds functionality)
- [x ] Breaking change (fix or feature that would cause existing functionality to change)


examples/slack_alert/run.py
```diff
@@ -52,15 +65,15 @@ def main(config: str):
             trainer=svc_trainer(),
             evaluator=evaluator(),
             formatter=test_acc_post_formatter(),
-            alerter=slack_alerter_post_step(),
+            alerter=slack_alerter_post_step(slack_token, slack_channel),
```
fa9r, line 68: Great idea, but this will require a few more changes. If you look at the definition of `slack_alerter_post_step`, it doesn't actually accept those two arguments yet:

```python
def slack_alerter_post_step(
    params: SlackAlerterParameters, context: StepContext, message: str
) -> bool:
```

To set the channel ID, we could pass it as part of the params argument:

```suggestion
            alerter=slack_alerter_post_step(params=SlackAlerterParameters(slack_channel_id=slack_channel)),
```

The token is a bit more challenging to parametrize though. Currently, it can only be set in the stack component configuration, i.e., you have to provide it when running `zenml alerter register ...`. We could also add it to the `SlackAlerterParameter` class, but then we will need to implement some logic for determining the correct token at run time, similar to `_get_channel_id` on line 72 of `src/zenml/integrations/slack/alerters/slack_alerter.py`.

fa9r (changes requested):

schustmi:
Closing this in favour of the more general solution in #1609 that does not only work for our example but all slack alerter stack components.

PR #1495, opened 2023-04-17
ZenML Hub CLI Improvements

fa9r:
## Describe changes
Misc `zenml hub ...` CLI command improvements:
- Added `zenml hub submit-batch` that can be used to batch submit/update multiple plugins at once
- Adjusted `zenml hub list` to sort plugins by author > name > version
- Fixed `zenml hub logs` (calling `error` before would cause the execution to end and not show the stacktrace)
- Moved `_clone_repo()` util function to `zenml.utils.git_utils.clone_git_repository()`

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/hub.py
+++ b/src/zenml/cli/hub.py
@@ -39,6 +39,7 @@ from zenml.models.hub_plugin_models import (
     PluginStatus,
 )
 from zenml.utils.analytics_utils import AnalyticsEvent, event_handler
+from zenml.utils.git_utils import clone_git_repository
 
 logger = get_logger(__name__)
 
@@ -70,7 +71,7 @@ def hub() -> None:
 def list_plugins(
     all: bool = False, mine: bool = False, installed: bool = False
 ) -> None:
-    """List all plugins available on the hub.
+    """List all plugins available on the ZenML Hub.
 
     Args:
         all: Whether to list all plugins, including ones that are not available.
@@ -109,7 +110,7 @@ def _format_plugins_table(
         The formatted table.
     """
     plugins_table = []
-    for plugin in plugins:
+    for plugin in sorted(plugins, key=_sort_plugin_key_fn):
         if _is_plugin_installed(plugin):
             installed_icon = ":white_check_mark:"
         else:
@@ -122,7 +123,7 @@ def _format_plugins_table(
             status_icon = ":x:"
 
         display_name = plugin_display_name(
-            plugin_name=plugin.name,
+            name=plugin.name,
             version=plugin.version,
             author=plugin.author,
         )
@@ -136,6 +137,22 @@ def _format_plugins_table(
         }
         plugins_table.append(display_data)
     return plugins_table
+
+
+def _sort_plugin_key_fn(
+    plugin: HubPluginResponseModel,
+) -> Tuple[str, str, str]:
+    """Helper function to sort plugins by name, version and author.
+
+    Args:
+        plugin: The plugin to sort.
+
+    Returns:
+        A tuple of the plugin's author, name and version.
+    """
+    if plugin.author == ZENML_HUB_ADMIN_USERNAME:
+        return "0", plugin.name, plugin.version  # Sort admin plugins first
+    return plugin.author, plugin.name, plugin.version
 
 
 @hub.command("install")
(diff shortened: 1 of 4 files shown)


src/zenml/utils/git_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            ignored.
+
+    Returns:
+        The cloned repository.
+
+    Raises:
+        RuntimeError: If the repository could not be cloned.
+    """
+    from git import GitCommandError
+    from git.repo import Repo
```
schustmi, line 44: Already imported above

src/zenml/cli/hub.py
```diff
@@ -995,10 +974,68 @@ def _ask_for_tags() -> List[str]:
...
+            - ...
+            ```
+    """
+    from pydantic import ValidationError
+
+    from zenml.utils.yaml_utils import read_yaml
+
+    config = read_yaml(config)
+    client = HubClient()
+    declare(f"Submitting {len(config)} plugins to the hub...")
```
schustmi, line 1016: I think we should move this below the check that makes sure that `config` is a dict, otherwise the `len(...)` call might lead to some hard to understand error message?

schustmi (changes requested):

src/zenml/cli/hub.py
fa9r: good point 👍

src/zenml/utils/git_utils.py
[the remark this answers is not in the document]
fa9r: Yes there is a little bit of overlap with `ExamplesRepo.clone`, but I don't think it's worth refactoring this since `zenml example ...` will hopefully soon be deprecated anyways

strickvl (approved):

fa9r:
@schustmi I also added tests for everything that does not involve cloning git repos.

Testing the repo structure validation seems difficult without, how would you go about that? IMO it's also not that important since the main check will be made server-side anyways.

Any other suggestions for things I should/could test?

tests/integration/functional/cli/test_hub.py
```diff
@@ -0,0 +1,75 @@
...
+
+EXAMPLE_PLUGIN_NAME = "langchain_qa_example"
+
+
+def test_hub_list():
+    """Test that zenml hub list does not fail."""
+    runner = CliRunner()
+    list_command = cli.commands["hub"].commands["list"]
+    result = runner.invoke(list_command)
+    assert result.exit_code == 0
```
strickvl, line 28: maybe also just check that the `langchain_qa_example` is inside the output string as well? or something else that we know will always be in the hub?

tests/unit/_hub/test_client.py
```diff
@@ -0,0 +1,61 @@
...
+    client = HubClient()
+    plugins = client.list_plugins()
+    assert len(plugins) > 0
+
+
+def test_get_plugin():
+    """Test getting a plugin."""
+    plugin_name = "langchain_qa_example"
+    client = HubClient()
+    plugin = client.get_plugin(plugin_name)
```
strickvl, line 48: In the past we removed all the tests of examples because they were doing these repo clones etc and it was making our tests really slow. I think the eventual idea was that we would move them out to be integration tests somehow. Anyway, maybe this is fine, but I think at some point we need a better way to handle these kinds of tests where we're pulling things or cloning repos that works for hub, examples and stack recipes.

tests/unit/_hub/test_client.py
```diff
@@ -0,0 +1,61 @@
...
+    assert plugin.name == plugin_name
+
+    # Test getting a specific version.
+    version = "0.1"
+    plugin = client.get_plugin(plugin_name, version=version)
+    assert plugin.name == plugin_name
+    assert plugin.version == version
+
+    # Test getting a non-existent plugin.
+    plugin_name = "non_existent_plugin"
```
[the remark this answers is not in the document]
fa9r: good point, we should take our established naming conventions seriously, my bad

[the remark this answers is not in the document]
strickvl: haha more to the point, if the hub takes off at some point someone will actually create a plugin with that name...

tests/unit/_hub/test_client.py
schustmi: I would suggest mocking this kind of functionality instead if that's at all possible

schustmi (approved):
LGTM. There are some more tests that could be written but all of those would require complicated mocking, not sure if that's possible at this moment?

PR #1499, opened 2023-04-20
Allow filtering runs by code repo id

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/models/pipeline_run_models.py
```diff
@@ -212,19 +220,39 @@ def generate_filter(
...
+
+        if self.code_repository_id:
+            from zenml.zen_stores.schemas import (
+                CodeReferenceSchema,
+                PipelineDeploymentSchema,
+                PipelineRunSchema,
+            )
+
+            code_repo_filter = and_(
+                True,
```
AlexejPenner, line 247: Why is it necessary to explicitely set the first Value to true here?

src/zenml/models/pipeline_run_models.py
```diff
@@ -212,19 +220,39 @@ def generate_filter(
...
+                PipelineRunSchema.deployment_id == PipelineDeploymentSchema.id,
+                PipelineDeploymentSchema.code_reference_id
+                == CodeReferenceSchema.id,
+                CodeReferenceSchema.code_repository_id
+                == self.code_repository_id,
+            )
+
+            base_filter = operator(base_filter, code_repo_filter)
```
AlexejPenner, line 255: This looks easier than I thought it would be :D

AlexejPenner (approved):
:seal:

src/zenml/models/pipeline_run_models.py
schustmi: That was a try for mypy, but didn't fix the issue either so I just added an ignore

PR #1503, opened 2023-04-24
Increase step run field lengths

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -95,8 +97,22 @@ class StepRunSchema(NamedSchema, table=True):
     end_time: Optional[datetime] = Field(nullable=True)
     status: ExecutionStatus
     entrypoint_name: str
-    parameters: str = Field(sa_column=Column(TEXT, nullable=False))
-    step_configuration: str = Field(sa_column=Column(TEXT, nullable=False))
+    parameters: str = Field(
+        sa_column=Column(
+            String(length=MEDIUMTEXT_MAX_LENGTH).with_variant(
+                MEDIUMTEXT, "mysql"
```
fa9r, line 103: I just saw we do this in some schemas already, but are we sure this works with other SQL variants as well?

schustmi: From what I found, SQLite has a single string type with a fixed length so it doesn't affect it at all. So the only case we need to care about for now is mysql which is solved by this `with_variant(...)` call

strickvl: If you want other SQL variants you have to just chain more `.with_variant` calls together. Discovered this with @stefannica a few weeks back. For the moment we only support MySQL, but in the future if we had a postgres variant this probably needs revisiting.

Issue #1502, opened 2023-04-24
[BUG]: SageMaker Step Operator - too long trainingJobNames


christianversloot:
### System Information

ZENML_LOCAL_VERSION: 0.38.0
ZENML_SERVER_VERSION: 0.38.0

### What happened?

In the case of long experiment names and step names, the `trainingJobName` as generated by ZenML when using the SageMaker Step Operator violates the <= 63 length requirement by the SageMaker Estimator:

```
ClientError: An error occurred (ValidationException) when calling the CreateTrainingJob operation: 2 validation errors detected: Value 'basic-zenml-project-pipeline-2023-04-24-12-22-59-886794-step-show-sagemaker-s3-link' at
'trainingJobName' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[a-zA-Z0-9](-*[a-zA-Z0-9]){0,62}; Value 'basic-zenml-project-pipeline-2023-04-24-12-22-59-886794-step-show-sagemaker-s3-link' at
'trainingJobName' failed to satisfy constraint: Member must have length less than or equal to 63
```

### Reproduction steps

1.
2.
3.
...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christianversloot:
Multiple options:

* Remove the step name from the `trainingJobName`, so `basic-zenml-project-pipeline-2023-04-24-12-22-59-886794`.
* Cut everything at length 60, providing three dots as a suffix, so `basic-zenml-project-pipeline-2023-04-24-12-22-59-886794-step...`
* Maybe as an addition, allow people to override `info.run_name` with their custom name, combined with any of the two above. If people would get stuck by e.g. naming their pipelines wrongly, they could quickly get back to work by e.g. specifying a name themselves.

I can add this with a PR; would love to hear your thoughts re/preferred options :)

@htahir1 @schustmi @strickvl

strickvl:
Thanks for the fix PR!

PR #1505, opened 2023-04-24
Generate unique SageMaker training job name based on pipeline and ste…

christianversloot:
…p name

## Describe changes
Attempts fixing https://github.com/zenml-io/zenml/issues/1502.

## Pre-requisites

## Types of changes


christianversloot:
On Slack, we agreed to the following approach:

```
Suggested approach:
Compose name composed_name as {pipeline_name}-{step-name}, cut it off to 63-5 = 58 chars if necessary.
Then add the suffix as {composed_name}-{random_suffix}, where random suffic is a length-4 random suffix.
```

...with the slight change that I am using 60 characters as a safety margin; let me know if this must be omitted.

Using the `StepRunInfo` object available to the step operator, and the `PipelineConfiguration` and `StepConfiguration` objects available in there, this PR first attempts to construct the `{pipeline_name}-{step_name}` composition as discussed above.

In doing so, I need to use `Client()` (I borrowed that idea from `get_image(...)` in the `StepRunInfo` object), but I am not yet sure if that's the right approach. Feel free to make other suggestions.

Then, as the suffix plus the `-` character will be 5 characters, the composed name is cut off to the first 55 characters. To that, the generated (pseudo)random suffix is appended as well as the `-` separator. This is sanitized (as was done previously) and subsequently used to fit the `estimator`. 

Looking forward to your thoughts!

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
```diff
@@ -179,9 +181,14 @@ def launch(
...
+        # SageMaker allows 63 characters at maximum for job name - ZenML uses 60 for safety margin.
+        step_name = Client().get_run_step(info.step_run_id).name
+        training_job_name = f"{info.pipeline.name}-{step_name}"[:55]
+        suffix = "".join(random.choices(string.ascii_lowercase, k=4))
+        unique_training_job_name = f"{training_job_name}-{suffix}"
+
         # Sagemaker doesn't allow any underscores in job/experiment/trial names
-        job_name = f"{info.run_name}-{info.pipeline_step_name}"
-        job_name = job_name.replace("_", "-")
+        sanitized_training_job_name = unique_training_job_name.replace("_", "-")
```
strickvl: I'm thinking this could be abstracted out of the main `launch` functionality as a util outside the integration, which would have the advantage of allowing for a unit test to be written as well. Maybe to live in `src/zenml/utils/string_utils.py`? WDYT? I imagine that there will be other places where we'll hit these external string length limitations.

strickvl (commented):
Looks good, but I had one query about whether to take some of the string construction logic outside the main `launch` method..

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
christianversloot: That would be fine with me. Most of the functionality is very specific to SageMaker however, as it's meant to reduce length of the training job name. The only truly generic part is the construction of `suffix` i.e. `suffix = "".join(random.choices(string.ascii_lowercase, k=4))`. I can move that to a string util (and try write a unit test). Anything else you'd consider generic and suitable for moving?

christianversloot: https://github.com/zenml-io/zenml/pull/1505/commits/4e1fe2441ede2e85bf001cc300fb1fc02541b845

strickvl: I was thinking a function that takes a string, max_length and a length of random chars might be the reusable part, but you're right outside that specific little piece, everything else is quite Sagemaker-specific.

christianversloot:
Agreed, I don't know either. Without it the build, lint tests etc would break.

strickvl (approved):
Ok I think this looks good now. I would like to fix the formatting issue, however, before merging this, and will let the CI / tests run through as well.

christianversloot:
I have reverted the mass formatting commit in https://github.com/zenml-io/zenml/pull/1505/commits/af51e839f12d86ef0acc399b2297e2d3e323deba and ran `ruff` and `black` manually on the `src/zenml/integrations/aws/step_operators` containing the only changes in this PR in https://github.com/zenml-io/zenml/pull/1505/commits/5daed8b22c32fa857b873fd8226cf4ec14f05587. This should mean that the Files changed section is ok again while lint checks etc should pass.

PR #1507, opened 2023-04-25
Default `cloudpickle` Materializer & Materializer Inheritance

fa9r:
## Describe changes
I added a `DefaultMaterializer` that uses `cloudpickle` to save artifacts that cannot be handled by any other materializers.

While doing so, I also refactored all existing materializers to enable materializer inheritance:
- input data validation was split to a separate method so the public materializer methods `load()`, `save()` and `extract_metadata()` no longer need to call `super()....`
- `BaseMaterializer.extract_metadata()` no longer extracts anything so forgetting to call `super().extract_metadata` doesn't matter. Added new `BaseMaterializer.extract_base_metadata()` method to replace that. I also adjusted all materializer tests to check that the expected number of metadata was extracted for each materializer.

The following materializer inheritance was introduced
- `DefaultMaterializer`
    - `SklearnMaterializer`
    - `LangchainOpenaiEmbeddingMaterializer`
    - `LangchainVectorStoreMaterializer`
- (new) `BasePytorchMaterializer`
    - `NeuralProphetMaterializer`
    - `PyTorchDataLoaderMaterializer`
    - `PyTorchModuleMaterializer`
    - `PyTorchLightningMaterializer` 
- `PandasMaterializer`
    - `DeepchecksDatasetMaterializer` 
- `PydanticMaterializer`
    - `LangchainDocumentMaterializer`
        - `LlamaIndexDocumentMaterializer`

## Pre-requisites

## Types of changes


schustmi (changes requested):

src/zenml/materializers/default_materializer.py
```diff
@@ -0,0 +1,80 @@
...
+        Args:
+            data_type: The data type of the artifact.
+
+        Returns:
+            The loaded artifact data.
+        """
+        super().load(data_type)
+        filepath = os.path.join(self.uri, DEFAULT_FILENAME)
+        with fileio.open(filepath, "rb") as fid:
+            data = cloudpickle.load(fid)
```
schustmi, line 57: Does the error message when running this in a different python version than it was saved in give a reasonable explanation? If not, is it worth it to query our DB to find out the environment in which the artifact was saved and compare the python versions?

src/zenml/materializers/default_materializer.py
```diff
@@ -0,0 +1,80 @@
...
+    def load(self, data_type: Type[Any]) -> Any:
+        """Reads an artifact from a cloudpickle file.
+
+        Args:
+            data_type: The data type of the artifact.
+
+        Returns:
+            The loaded artifact data.
+        """
+        super().load(data_type)
```
strickvl, line 54: For the ones I recently implemented, I also did a bit of validation that would log another warning if the Python versions were different. Possibly worth adding in? https://github.com/zenml-io/zenml/blob/develop/src/zenml/integrations/langchain/materializers/vector_store_materializer.py#L79

tests/unit/materializers/test_default_materializer.py
```diff
@@ -0,0 +1,28 @@
...
+from tests.unit.test_general import _test_materializer
+
+
+class Unmaterializable:
+    """A class that has no materializer."""
+
+    cat = "aria"
+
+
+def test_default_materializer(clean_client):
```
strickvl, line 25: If you also get around to adding in the validation for mismatches of Python versions, would be nice to have a test here that confirms that's also happening.

src/zenml/materializers/default_materializer.py
fa9r: Yeah makes sense, we should check this. I copied @strickvl's implementation that saves the python version to another file to check that. Do you think that's fine or would you prefer loading this info from the DB instead? How would you query the DB inside the materializer even? `list_artifacts(uri=...)[0]`?

src/zenml/steps/base_step.py
```diff
@@ -963,10 +961,8 @@ def _finalize_configuration(
...
 
             if not output.materializer_source:
-                if default_materializer_registry.is_registered(output_class):
-                    materializer_class = default_materializer_registry[
-                        output_class
-                    ]
+                if materializer_registry.is_registered(output_class):
+                    materializer_class = materializer_registry[output_class]
                 else:
                     raise StepInterfaceError(
```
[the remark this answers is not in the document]
fa9r: Makes sense, adjusted accordingly. You can now do `materializer_registry.default_materializer = MyClass` to set which default materializer to use.

src/zenml/materializers/default_materializer.py
schustmi: IMO the cleanest solution would be: All materializers get the `ArtifactResponseModel` passed in their `__init__` instead of just the URI. The pickle materializer stores the python version in the artifact metadata (useful anyway I guess?) and can then read that python version from there in `load()`. Wdyt?

src/zenml/materializers/default_materializer.py
```diff
@@ -0,0 +1,117 @@
...
+    write_file_contents_as_string,
+)
+
+logger = get_logger(__name__)
+
+DEFAULT_FILENAME = "artifact.pkl"
+DEFAULT_PYTHON_VERSION_FILENAME = "python_version.txt"
+
+
+class DefaultMaterializer(BaseMaterializer):
```
schustmi, line 41: I would rather call this `CloudpickleMaterializer` as that is what it really does. In the future this might not be our default anymore, and at that point we run into a mess keeping the old class available for backwards compatibility.

src/zenml/integrations/langchain/materializers/openai_embedding_materializer.py
```diff
@@ -13,130 +13,21 @@
...
-DEFAULT_PICKLE_FILENAME = "embedding.pkl"
-DEFAULT_PYTHON_VERSION_FILENAME = "python_version.txt"
-DEFAULT_LANGCHAIN_VERSION_FILENAME = "langchain_version.txt"
-LANGCHAIN_PACKAGE_NAME = "langchain"
-
-logger = get_logger(__name__)
 
-
-class LangchainOpenaiEmbeddingMaterializer(BaseMaterializer):
+class LangchainOpenaiEmbeddingMaterializer(DefaultMaterializer):
```
schustmi, line 28: Can this still read the old files written by `pickle` using `cloudpickle`?

src/zenml/steps/base_step.py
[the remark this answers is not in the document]
schustmi: Does this work now though? It seems to me that `is_registered` still does not consider the default materializer and will return `False` and throw an error? I guess the entire if clause here could be removed and just always use `materializer_class = materializer_registry[output_class]`?

schustmi (changes requested):

src/zenml/steps/base_step.py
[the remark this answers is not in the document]
fa9r: Ah yes you're right, this check in `base_step` is now redundant too, adjusted it as you proposed 👍

src/zenml/materializers/default_materializer.py
fa9r: Makes sense 👍

src/zenml/materializers/default_materializer.py
fa9r: I agree that would be cleaner in this case, but IMO it's not worth changing the materializer interface over this (which would require adjusting the `__init__` of every materializer that implements it, deprecating the `uri` argument, adding corresponding backwards compatibility code, ...).

Generally, I'm also not sure whether coupling the materializer to the artifact is cleaner overall. Intuitively it seems good because each URI is currently assigned to a unique artifact, so we might as well pass the whole artifact. However, I do like the simplicity of the current interface, IMO it's very obvious what it's supposed to do without requiring any other classes for context.

src/zenml/integrations/langchain/materializers/openai_embedding_materializer.py
fa9r: Great question. Previously that didn't work because the Python file was not there but I fixed it now. Fortunately, it seems `pickle` and `cloudpickle` are fully compatible otherwise (I tried it with several classes incl. sklearn manually and also added a unit test for this).

src/zenml/materializers/default_materializer.py
schustmi: Depends, I guess we might want to make it clean at some point right? Unfortunately it doesn't even make sense to have it in the `__init__` as a materializer that just gets initialized to save some data does not have the response model available at that time I guess.

Ideally I think we would have `classmethods` for both save/load, where the `load` method would get the model. But you'd say this PR is not the time to adjust that right?

src/zenml/integrations/deepchecks/materializers/deepchecks_dataset_materializer.py
```diff
@@ -42,13 +39,7 @@ def load(self, data_type: Type[Any]) -> Dataset:
...
-
-        # Outsource to pandas
-        pandas_materializer = PandasMaterializer(self.uri)
-        df = pandas_materializer.load(data_type)
-
-        # Recreate from pandas dataframe
+        df = super().load(data_type)
         return Dataset(df)
 
     def save(self, df: Dataset) -> None:
```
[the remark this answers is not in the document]
fa9r: yes yes I already changed it on the other PR now :D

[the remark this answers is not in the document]
fa9r: Yeah already changed on the other PR, will do the same here 👍

src/zenml/materializers/default_materializer.py
fa9r: Yeah I would not do it now since I'm not even convinced that coupling materializers to artifact classes is necessarily cleaner.

PR #1381, opened 2023-03-06
SageMaker step operator: expand input arguments and add support for S3 input data

christianversloot:
## Describe changes
Hi, we are seeing great potential in ZenML to serve as the core of our MLOps strategy. However, we're also missing certain functionalities, at the moment esp. in the SageMaker step operator. We plan to actively contribute to ZenML development in order to fulfill our requirements while making them available for the broader public too.

This PR does the following:
* Expand `SagemakerStepOperatorSettings` with arguments available in `sagemaker.estimator.Estimator` to give ZenML based SageMaker training jobs more control over SageMaker behavior. It also adds `input_data_s3_uri` to the settings class, see next bullet point.
* Add support for S3-based training (or training, validation, testing) data imports into the SageMaker training container via `TrainingInput` objects (as described here - https://docs.aws.amazon.com/sagemaker/latest/dg/model-access-training-data.html). We've recently had a use case where we trained with approx. 300GB of data, which currently is unsupported with ZenML SageMaker step operator, but will be important for us in the near future.

What I am unsure about at the moment:
* There is a second class named `SagemakerStepOperatorConfig`, too, and it can be that certain arguments added in this PR belong there instead of in `SagemakerStepOperatorSettings`. Let me know the threshold for when specific things must be in Settings and in Config, and I will adapt the PR.
* There is no **test suite** for the SageMaker step operator atm; I omitted it here, too.

## Pre-requisites
Please ensure you have done the following:
- [X] I have read the **CONTRIBUTING.md** document.
- [X] If my change requires a change to docs, I have updated the documentation accordingly. (Only changes in API docs).
- [X] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #1509, opened 2023-04-25
Extend SageMakerOrchestratorSettings with processor_args enabling step level configuration

christianversloot:
## Describe changes
In https://github.com/zenml-io/zenml/pull/1381, we expanded the SageMaker Step Operator with `estimator_args` which could be passed to the Step Operator either via the component configuration or passing them directly to the step.

It works beautifully and especially the two following items are useful to us:

1. Access to data in S3 without having to copy things manually.
2. Managed Spot training for cost savings.

We have set up the SageMaker Orchestrator too because some processes take too long and we want them to run in the cloud - something we've discovered only after using ZenML for a bit.

For this reason, this PR adapts the SageMaker Orchestrator in the following way:

1. Add `processor_args` to the `SagemakerOrchestratorSettings` class. Typing wise it is exactly equal to the `estimator_args` in `SagemakerStepOperatorSettings` and behaves similarly.
2. Add `input_data_s3_uri` and `input_data_s3_mode` to the `SagemakerOrchestratorSettings`. Typing wise it is exactly equal to adding S3 data in the Step Operator.
3. In the SageMaker Orchestrator's `prepare_or_run_pipeline` def, start with the provided `processor_args_for_step` as default (or an empty dictionary if not set). Then, add all the default values for the settings configured when adding the component (e.g. instance type, role, volume size, et cetera) as default values, in order to maintain backwards compatibility (I'm not a fan of deprecating those, as we want to provide a default config when they are not provided via individual step settings - and configuring them for the whole component is indeed a good way of doing so).
4. That way, when no processor args are provided, the default values are used, but they can be overwritten by providing them in the `processor_args` if wanted (e.g. to run most steps in a specific instance type, while running another e.g. training in a heavy instance). Other `Processor` arguments can be provided as well https://sagemaker.readthedocs.io/en/stable/api/training/processing.html#sagemaker.processing.Processor.
5. Also, specific steps cannot be overwritten, and they are fixed to the values provided either by ZenML or by the configuration (e.g. SageMaker session).
6. Then, finally, `input_data_s3_uri` and `_mode` are used to construct a list of `inputs`, which are `ProcessingInput` objects that can be used to pass S3 data into a pipeline step when running it with SageMaker orchestrator.

How this can be used is as follows:

1. If you don't want to change anything and run with the settings provided when configuring the component - do nothing! Things keep working as they do.
2. If you want to provide specific settings for some steps, configure them at step level:

```python
@step(settings={ "orchestrator.sagemaker": { "processor_args": { "instance_type": "..." }} })
def train_model(
```

Hope this makes sense! Looking forward to hearing your thoughts.

Now, for the future - all steps under the hood are `ProcessingStep`s. This means that Managed Spot Training is currently not supported. We may consider allowing to use `TrainingStep`s as well (see e.g. https://docs.aws.amazon.com/sagemaker/latest/dg/build-and-manage-steps.html), but I'd say this is something for a later PR - adding S3 support into the orchestrator adds more value as it enables training with large datasets via AWS.

## Pre-requisites

## Types of changes


htahir1:
@christianversloot Looks amazing but we still have this problem with the file changes as it looks like you started from main again?

christianversloot:
Thanks @htahir1! I used the same fork as the other PR and although I think I started off `develop`, I am not sure as I branched when I had little time (and built the rest the day after). So will make sure to double check next time 😅

For now, I can suggest that we temporarily revert commit https://github.com/zenml-io/zenml/pull/1509/commits/16d6912ffa2112e80fecae59c63c9a40ccbdbdb7 which contains the mass formatting changes. This would help you review the PR. Then, when review is done, we reformat again after which we can merge with `develop`. Next time, then, this problem should not occur again.

Also, over the course of last week, we realized that we may also need the `outputs` argument in `processor.run(...)`. 

```
outputs (list[[ProcessingOutput](https://sagemaker.readthedocs.io/en/stable/api/training/processing.html#sagemaker.processing.ProcessingOutput)]) – Outputs for the processing job. These can be specified as either path strings or [ProcessingOutput](https://sagemaker.readthedocs.io/en/stable/api/training/processing.html#sagemaker.processing.ProcessingOutput) objects (default: None).
```

I will add this into the PR.

strickvl:
I think if you can just remove the commit that contains the mass formatting changes we might not have to add it back in. (Or else we can / we'll have to make a PR on our end to revert those changes back since develop requires no formatting changes currently). But thank you for doing that. Makes it much easier to review.

christianversloot:
In x support for `ProcessingOutput` was added.

This allows you to write outputs from your training process to S3 from a local folder. For example, this config:

```
sagemaker_orchestrator_settings = SagemakerOrchestratorSettings(
    processor_args = {
        'volume_size_in_gb': 30
    },
    input_data_s3_uri = {
        'data_path': 's3://some-zenml-basic-project'
    },
    output_data_s3_uri = {
        "data_path": "s3://some-zenml-basic-project-output"
    }
)
```

...will tell SageMaker to first copy all data from `s3://some-zenml-basic-project` to `/opt/ml/processing/input/data-path`. When done, the contents of `/opt/ml/processing/output/data_path` will be copied to S3 (and via `step_settings.output_data_s3_mode`, it is possible to configure it as 'Continuous' as well, e.g. to save model checkpoints or save other things). 

* Backwards compatible with the current SageMaker settings.
* In structure similar to `ProcessingInput`.
* Can be used to save training artifacts and other things.
* Big one for us: allows for running data preprocessing pipelines (batch style instead of on the fly) through ZenML as well.

Just ran a pipeline configured above with the local changes and it completes successfully: input from S3 becomes available and output is written to the bucket.

This would finalize my work on the PR for now so review can start :)

strickvl (changes requested):
This is looking good to me. The main missing piece from this PR is the documentation. Could you update https://docs.zenml.io/component-gallery/orchestrators/sagemaker (from within `/docs/book/component-gallery/...`) with the new attributes on the `Settings` object as well as a version of the use case (basically as you describe it in the intro to this PR). I think that would be useful at surfacing this functionality for users who won't necessarily be diving into the docstrings to understand what the Sagemaker orchestrator can do.

Thanks!

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -169,38 +170,105 @@ def prepare_or_run_pipeline(
...
+
+            # Construct S3 outputs from container for step
+            outputs = None
+
+            if step_settings.output_data_s3_uri is None:
+                pass
+            elif isinstance(step_settings.output_data_s3_uri, str):
+                outputs = [
+                    ProcessingOutput(
+                        source="/opt/ml/processing/output",
```
schustmi, line 243: Is this something we want configurable as well?

christianversloot: We could do that, but then we would deviate a bit from what SageMaker discusses here: https://docs.aws.amazon.com/sagemaker/latest/dg/model-access-training-data.html - in the image, the location in the filesystem specifies `/opt/ml/input/data/<channel>` (I assumed the same for output).
To be highly precise, I would probably need to add `./data` to all `ProcessingInput` and `ProcessingOutput` references (let me do that right now), but I would suggest to keep the hardcoded ref as-is? What do you think?

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -19,6 +19,9 @@ You should use the Sagemaker orchestrator if:
 * you're looking for a managed solution for running your pipelines.
 * you're looking for a serverless solution for running your pipelines.
 
+## How it works
+The Sagemaker orchestrator works with Sagemaker Pipelines, which can be used to construct machine learning pipelines. Under the hood, for each ZenML pipeline step, it creates a SageMaker `PipelineStep`, which contains a Sagemaker Processing job. Currently, other step types are not supported.
```
strickvl, line 23:
```suggestion
The ZenML Sagemaker orchestrator works with [Sagemaker Pipelines](https://aws.amazon.com/sagemaker/pipelines), which can be used to construct machine learning pipelines. Under the hood, for each ZenML pipeline step, it creates a SageMaker `PipelineStep`, which contains a Sagemaker Processing job. Currently, other step types are not supported.
```

docs/book/component-gallery/orchestrators/sagemaker.md
```diff
@@ -94,30 +97,106 @@ a schedule. We maintain a public roadmap for ZenML, which you can find
 want to enable scheduling for Sagemaker, please [do let us
 know](https://zenml.io/slack-invite)!
 
-### Additional configuration
+### Configuration at pipeline or step level
+When running your ZenML pipeline with the Sagemaker orchestrator, the configuration set when configuring the orchestrator as a ZenML component will be used by default. However, it is possible to provide additional configuration at [pipeline or step level](../../advanced-guide/pipelines/settings#how-to-use-settings). This allows you to run whole pipelines or individual steps with alternative configurations. For example, this allows you to run the training process with a heavier, GPU-enabled instance type, while running other steps with lighter instances.
```
strickvl, line 101:
```suggestion
When running your ZenML pipeline with the Sagemaker orchestrator, the configuration set when configuring the orchestrator as a ZenML component will be used by default. However, it is possible to provide additional configuration at the [pipeline or step level](../../advanced-guide/pipelines/settings#how-to-use-settings). This allows you to run whole pipelines or individual steps with alternative configurations. For example, this allows you to run the training process with a heavier, GPU-enabled instance type, while running other steps with lighter instances.
```

strickvl (changes requested):
A couple of tiny changes, but otherwise this looks good to me and ready to merge.

strickvl (approved):
🙌

stefannica (approved):
I truly hope we can formalize these specialized mechanisms of inputting and outputting data into a Sagemaker powered step and make them part of the Artifact Store abstraction, or something that has to do with the interaction between the Artifact Store and the input/output step artifacts.

PR #1510, opened 2023-04-27
Fix post execution `get_pipeline()` and `pipeline.get_runs()`

fa9r:
## Describe changes
Getting the last run of a pipeline was broken because both `get_pipeline(...).runs` and `pipeline.get_runs()` would return the last run of the most recent pipeline version, but older versions might have newer runs.

This was very problematic because such basic behavior as the following was not working correctly:
```python
pipeline_instance.run()
run_view = pipeline_instance.get_runs()[0]  # was not guaranteed to be the run that was just executed
```

See the new integration test for a more detailed example.

To fix it, I changed the following behaviors:
- `pipeline_instance.get_runs()`: now returns the last 50 runs of this exact pipeline version (instead of the last 50 runs of the last version of the pipeline with the same name)
- `pipeline_class.get_runs()` / `get_pipeline(pipeline_name).runs ` / `get_pipeline(pipeline_class).runs`: now return the last 50 runs of all pipeline versions with the given pipeline name (instead of the last 50 runs of the last version only)

### Technical Details
- I introduced a new `PipelineClassView` post execution class that is just a namespace and has a list of all versions associated to that pipeline name. Calling `get_pipeline("pipeline_name")` or `get_pipeline(pipeline_class)` will now return such a `PipelineClassView` instead of the `PipelineView`.
- I reimplemented `BasePipeline.get_runs()` as a Descriptor that has different behavior based on whether the method is called on a pipeline class or a pipeline instance.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/pipelines/base_pipeline.py
+++ b/src/zenml/pipelines/base_pipeline.py
@@ -177,8 +177,8 @@ class GetRunsDescriptor:
         if pipeline_view:
             return lambda: pipeline_view.runs
         raise RuntimeError(
-            "The pipeline view for this pipeline was not found. Are you sure "
-            "this pipeline has been run already?"
+            "The pipeline view for this pipeline was not found. Please check "
+            "that the pipeline has been run already."
         )
 
 
--- a/src/zenml/pipelines/base_pipeline.py
+++ b/src/zenml/pipelines/base_pipeline.py
@@ -85,7 +85,7 @@ from zenml.utils.analytics_utils import AnalyticsEvent, event_handler
 if TYPE_CHECKING:
     from zenml.config.base_settings import SettingsOrDict
     from zenml.config.source import Source
-    from zenml.post_execution import PipelineRunView
+    from zenml.post_execution import PipelineRunView, PipelineView
 
     StepConfigurationUpdateOrDict = Union[
         Dict[str, Any], StepConfigurationUpdate
@@ -141,6 +141,45 @@ class BasePipelineMeta(type):
 
 
 T = TypeVar("T", bound="BasePipeline")
+
+
+class GetRunsDescriptor:
+    """Descriptor to define the `BasePipeline.get_runs`.
+
+    Descriptors (https://docs.python.org/3/reference/datamodel.html#implementing-descriptors)
+    allow us to define different behaviors for pipeline classes and instances.
+    """
+
+    def __get__(
+        self, instance: Optional["BasePipeline"], cls: Type["BasePipeline"]
+    ) -> Callable[[], List["PipelineRunView"]]:
+        """Get all runs of this pipeline instance or class.
+
+        Args:
+            instance: The pipeline instance if called on an instance else None.
+            cls: The pipeline class.
+
+        Returns:
+            A list of all runs of this pipeline instance or class.
+
+        Raises:
+            RuntimeError: If the method is called on a pipeline instance that
+                has not been run yet.
+        """
+        from zenml.post_execution import get_pipeline
+
+        pipeline_view: PipelineView
+        if instance is None:
+            pipeline_view = get_pipeline(cls)
+        else:
+            pipeline_view = get_pipeline(instance)
+
+        if pipeline_view:
+            return lambda: pipeline_view.runs
+        raise RuntimeError(
+            "The pipeline view for this pipeline was not found. Are you sure "
+            "this pipeline has been run already?"
+        )
 
 
 class BasePipeline(metaclass=BasePipelineMeta):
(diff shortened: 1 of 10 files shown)


schustmi:
On a general note, wouldn't `PipelineView` and `PipelineVersionView` be more appropriate names?

src/zenml/pipelines/base_pipeline.py
```diff
@@ -143,6 +143,45 @@ def __new__(
...
+        if instance is None:
+            pipeline_view = get_pipeline(cls)
+        else:
+            pipeline_view = get_pipeline(instance)
+
+        if pipeline_view:
+            return lambda: pipeline_view.runs
+        raise RuntimeError(
+            "The pipeline view for this pipeline was not found. Are you sure "
+            "this pipeline has been run already?"
```
strickvl, line 181: Rather than asking a question, I'd maybe rephrase this as "Please check that..."

src/zenml/post_execution/pipeline.py
```diff
@@ -74,56 +70,51 @@ def get_pipeline(
...
+    # Pipeline instance: find the corresponding pipeline version in the DB.
+    if isinstance(pipeline, BasePipeline):
         pipeline_model = pipeline._get_registered_model()
         if pipeline_model:
             return PipelineView(model=pipeline_model)
         else:
             return None
+
+    # Otherwise, for pipeline name or class, determine the name first.
+    if isinstance(pipeline, str):
```
strickvl, line 98:
```suggestion
    elif isinstance(pipeline, str):
```

Maybe, just to preserve the logic of this conditional, even if the `return` would bypass the rest?

src/zenml/post_execution/pipeline.py
```diff
@@ -134,11 +125,78 @@ def get_pipeline(
...
+    @property
+    def num_runs(self) -> int:
+        """Returns the number of runs of this pipeline name/class.
+
+        This is the sum of all runs of all versions of this pipeline.
+
+        Returns:
+            The number of runs of this pipeline name/class.
+        """
+        return sum([version.num_runs for version in self.versions])
```
strickvl, line 163:
```suggestion
        return sum(version.num_runs for version in self.versions)
```

Save a bit of memory with a generator expression instead of a list comprehension?

strickvl (changes requested):

src/zenml/post_execution/pipeline.py
fa9r: I think `if` is preferable over `elif` here since certain linters like pylint even complain if you do `elif` since the "else" part is effectively redundant

src/zenml/post_execution/pipeline.py
```diff
@@ -134,11 +125,78 @@ def get_pipeline(
...
+
+    @property
+    def versions(self) -> List["PipelineView"]:
+        """Returns all versions/instances of this pipeline name/class.
+
+        Returns:
+            A list of all versions of this pipeline.
+        """
+        client = Client()
+        pipelines = client.list_pipelines(
```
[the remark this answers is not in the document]
fa9r: Hmm yeah probably. Pagination + post execution don't go along well in general. I'll depaginate it for now.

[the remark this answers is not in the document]
htahir1: @fa9r  Wouldnt it make sense to pass the pagination params as first class parameters in the post execution method?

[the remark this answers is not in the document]
AlexejPenner: in the long run i agree with you Hamza, but I think that takes a bit of thought and care to figure out to what extent to do this (will we include the `contains` language?). As this ticket was supposed to be a quickfix.

tests/integration/functional/post_execution/test_pipeline.py
```diff
@@ -87,3 +93,98 @@ def test_get_pipeline_fails_for_wrong_type(clean_client, wrong_pipeline_type):
     """Test that `get_pipeline` fails for wrong input types."""
     with pytest.raises(RuntimeError):
         get_pipeline(wrong_pipeline_type)
+
+
+def test_get_runs(clean_client, connected_two_step_pipeline):
```
AlexejPenner, line 98: Love the thorough tests!

AlexejPenner (approved):
:seal:

src/zenml/post_execution/pipeline.py
```diff
@@ -144,12 +146,16 @@ def versions(self) -> List["PipelineView"]:
             A list of all versions of this pipeline.
         """
         client = Client()
-        pipelines = client.list_pipelines(
-            workspace_id=client.active_workspace.id,
-            name=self.name,
-            sort_by="desc:created",
+
+        pipelines = depaginate(
+            partial(
```
strickvl, line 151: cool

strickvl (approved):
🙌

PR #1511, opened 2023-04-28
Simplify Dashboard UX via `zenml.show()`

fa9r:
## Describe changes

I added a `zenml.show()` Python function and corresponding `zenml show` CLI command that can be used to open the ZenML dashboard in an environment-specific way:
- In the native environment, the dashboard is opened in a new browser tab (similar as `zenml up` does right now)
- In Colab and other notebooks, the dashboard is embedded as an IFrame
- To work with Colab, one can also provide a `zenml.show(NGROK_TOKEN)` arg to automatically tunnel the dashboard via ngrok.

### New UX in notebooks:
Screenshot 2023-04-28 at 14 32 01

### Old UX in notebooks:
Screenshot 2023-04-28 at 14 14 29

## Pre-requisites

## Types of changes


fa9r:
Some inspiration from [gradio docs](https://gradio.app/docs/#interface-launch):

Screenshot 2023-05-02 at 12 04 58

htahir1 (commented):
In general, I like this UX a lot! Now sure what the value of having zenml.up() zenml.down() would be though? Do you mean they wouldnt have to do `!zenml up` in the notebook?

src/zenml/cli/server.py
```diff
@@ -260,26 +226,29 @@ def up(
                 )
 
             if server.status and server.status.url:
-                cli_utils.declare(
-                    f"The local ZenML dashboard is available at "
+                logger.info(
+                    f"The ZenML dashboard is available at "
```
fa9r, line 230:
```suggestion
                cli_utils.declare(
                    f"The local ZenML dashboard is available at "
```

src/zenml/cli/server.py
```diff
@@ -260,26 +226,29 @@ def up(
...
                     f"'{server.status.url}'. You can connect to it using the "
                     f"'{DEFAULT_USERNAME}' username and an empty password. "
-                    f"To open the dashboard in a browser automatically, "
-                    f"set the env variable AUTO_OPEN_DASHBOARD=true."
                 )
+                zenml.show()
```
fa9r, line 234:
```suggestion
                zenml.show(ngrok=ngrok)
```

src/zenml/utils/dashboard_utils.py
```diff
@@ -81,3 +83,42 @@ def print_run_url(run_name: str, pipeline_id: Optional[UUID] = None) -> None:
...
+        from IPython.display import IFrame
+
+        display(IFrame(src=url, width="100%", height=720))
+
+    elif environment == EnvironmentType.NATIVE:
+        if handle_bool_env_var(ENV_AUTO_OPEN_DASHBOARD, default=True):
+            try:
+                import webbrowser
+
+                webbrowser.open(url)
```
htahir1, line 109: This webbrowser.open doesnt work on WSL. I wanted to fix that for a while, but there  is no real solution except here: https://github.com/python/cpython/issues/89752#issuecomment-1254943311 . Maybe we can try this suggestion in a try block?

```python
import os
import shutil


## maybe add these where we detect the OS to add another WSL type
from platform import uname

# source: https://www.scivision.dev/python-detect-wsl/
def in_wsl() -> bool:
    return 'microsoft-standard' in uname().release
    
# probably wrap in a try catch
if in_wsl():
    webbrowser.get("wslview %s").open(url)
```

htahir1 (changes requested):
Just left one comment, for the other part I am not the best person to review

src/zenml/cli/server.py
```diff
@@ -128,6 +85,12 @@ def get_active_deployment(local: bool = False) -> Optional["ServerDeployment"]:
     help="Use a custom Docker image for the ZenML server. Only used when "
     "`--docker` is set.",
 )
+@click.option(
+    "--ngrok",
+    type=str,
+    default=None,
+    help="Specify an ngrok auth token to use for exposing the ZenML server.",
+)
```
AlexejPenner, line 93: How about we call this auth-token or something more obvious to the user?

AlexejPenner (approved):
Generally looks good, lets get this in as well to the next release. I guess the cahnges on the quickstart notebook are probably deprecated though

src/zenml/cli/server.py
fa9r: Makes sense I renamed this to `ngrok-token` and also renamed `ngrok` args to `ngrok_token` everywhere

src/zenml/utils/dashboard_utils.py
fa9r: Ok dashboard opening should work under WSL, could you give it a try and see if it works?

htahir1 (approved):
Ok makes sense!

src/zenml/utils/dashboard_utils.py
fa9r: Update: worked! 🎉

PR #1512, opened 2023-05-01
Basic PyCaret integration and materializer

christianversloot:
## Describe changes
I was building a PyCaret example for our internal examples and needed a PyCaret materializer. This PR implements a (basic) PyCaret integration (`zenml integration install pycaret`) and makes available a materializer for PyCaret models. 

As PyCaret implements models using a variety of frameworks/libraries under the hood (Scikit-learn and others), the integration requires that they are installed alongside PyCaret itself.

Because of this generic implementation and the way ZenML is built, it was difficult to make a catch-all implementation that allows a list of PyCaret models to be written away (which can be the output of PyCaret's `compare_models` defs). Currently, they must be written individually as individual outputs. Although this can be cumbersome, it seems to be the only way of doing so at the moment:

```
@step(output_materializers=PyCaretMaterializer)
def find_best_model(
    dataset: pd.DataFrame
) -> Output(
    model_1=object,
    model_2=object,
    model_3=object
):
```

Let me know what you think. If you're interested in having this within ZenML, then I will try and write some docs (prerequisites below).

## Pre-requisites

## Types of changes


fa9r (approved):
Awesome, thanks for submitting the PR! Code looks good to me!

The only thing I am wondering is whether we could decouple the integrations somehow. It is indeed not ideal that all of the underlying modeling libraries need be installed. However, currently this is required since `ASSOCIATED_TYPES` needs to list all the types it can handle. The only clean change I can think of right now is to allow specifying import paths as string in `ASSOCIATED_TYPES` by changing it to `Tuple[Union[str, Type[Any]], ...]`. Not fully sure about that one though. @schustmi any ideas?

Regardless of what we decide, I think the decoupling would be a larger refactoring that shouldn't be part of this PR, so IMO we can merge this right away 👍

christianversloot:
Great, thanks for your quick review!

Any need for some docs - e.g. here https://zenml.io/integrations? Or is this added in a different way?

fa9r:
Thanks for asking, but no, no need for any docs, we only require those for integrations that contain stack components flavors. The website integrations page is maintained separately and we'll add PyCaret there manually once the PR is merged.

fa9r:
As for the failing CI checks:
- the `macos-setup-python-environment` is failing because catboost didn't build correctly on the Mac runner. Do you have a Mac you can test this on? If not, I can help debug this on my M1 Mac.
- the `ubuntu-integration-test` is also failing on other PRs so this should be unrelated to your changes, perhaps a new mlflow version was released, we'll need to investigate that separately.

christianversloot:
@fa9r I do not have a Mac unfortunately, so if you could help debug this on yours that would be great!

christianversloot:
@fa9r quite recently other people have experienced this issue as well, so this may be useful for your debugging work: https://github.com/catboost/catboost/issues/2371.

So far they have only fixed this by hardcoding an older version of `catboost`, but I would not prefer to do that within the ZenML integration unless really necessary (nothing is as permanent as a temporary solution).

The `cmake` error in the ZenML GitHub action is the same as reported in the `catboost` issue:

```
      subprocess.CalledProcessError: Command '['cmake', '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-euzeiln9/catboost_ba2c03e7117b4895a5cf423b82ab3989/catboost_all_src', '-B', '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-euzeiln9/catboost_ba2c03e7117b4895a5cf423b82ab3989/build/temp.macosx-10.15-x86_64-cpython-38', '-G', 'Ninja', '-DCMAKE_BUILD_TYPE=Release', '-DCMAKE_TOOLCHAIN_FILE=/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-euzeiln9/catboost_ba2c03e7117b4895a5cf423b82ab3989/catboost_all_src/build/toolchains/clang.toolchain', '--log-level=VERBOSE', '-DCMAKE_POSITION_INDEPENDENT_CODE=On', '-DCATBOOST_COMPONENTS=python-package', '-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0', '-DHAVE_CUDA=no', '-DPython3_ROOT_DIR=/Users/runner/.virtualenvs/.venv']' returned non-zero exit status 1.
```

So probably [Ninja](https://ninja-build.org/) needs to be added to the build environment of the Action.

Hope this helps! 🤞🏻

src/zenml/integrations/pycaret/__init__.py
```diff
@@ -0,0 +1,38 @@
+#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
strickvl, line 1:
```suggestion
#  Copyright (c) ZenML GmbH 2023. All Rights Reserved.
```

src/zenml/integrations/pycaret/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+from zenml.integrations.constants import PYCARET
+from zenml.integrations.integration import Integration
+
+
+class PyCaretIntegration(Integration):
+    """Definition of PyCaret integration for ZenML."""
+
+    NAME = PYCARET
+    REQUIREMENTS = [
+        "pycaret",
```
strickvl, line 25: Wonder whether it's worth specifying >3.0.0 here? @fa9r thoughts?

src/zenml/integrations/pycaret/materializers/__init__.py
```diff
@@ -0,0 +1,18 @@
+#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
strickvl, line 1:
```suggestion
#  Copyright (c) ZenML GmbH 2023. All Rights Reserved.
```

src/zenml/integrations/pycaret/materializers/model_materializer.py
```diff
@@ -0,0 +1,166 @@
+#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
```
strickvl, line 1:
```suggestion
#  Copyright (c) ZenML GmbH 2023. All Rights Reserved.
```

src/zenml/integrations/pycaret/materializers/model_materializer.py
```diff
@@ -0,0 +1,166 @@
...
+    ASSOCIATED_ARTIFACT_TYPE = ArtifactType.MODEL
+
+    def load(self, data_type: Type[Any]) -> Any:
+        """Reads and returns a PyCaret model after copying it to temporary path.
+
+        Args:
+            data_type: The type of the data to read.
+
+        Returns:
+            A PyCaret  model.
```
strickvl, line 134:
```suggestion
            A PyCaret model.
```

strickvl (changes requested):
Looks good to me. Just a few nits, but otherwise once the CI is addressed this looks ready to merge.

src/zenml/integrations/pycaret/__init__.py
fa9r: If it only works for `pycared>3.0.0` then yes otherwise no 😄

strickvl: @christianversloot I'm not familiar enough with `pycaret`'s API to have an opinion on this. From a brief perusal of their release notes it seems that 3.0 is quite a bit different from <3.0 and we might have some compatibility issues if we support pre-3.0 versions. Do you have an opinion / better sense of this?

christianversloot: The ecosystem is evolving fast. In fact I can't even find <3.0 docs anymore (although I did a quick search only). Maybe it would be good to force `>=3.0.0` (not `> 3.0.0` as `3.0.0` is newest? or maybe I'm wrong and `> 3.0.0` includes 3.0.0): people will be using this integration primarily for _new models_, I presume, not to convert their existing scripts/pipelines.

strickvl: Yep. `>=3.0.0` would be good I think.

strickvl: If people ask for a wider range we can always consider it as and when people use the integration. And of course there's always the option of having a custom materializer etc.

strickvl, line 25:
```suggestion
        "pycaret>=3.0.0",
```

christianversloot: Done :)

strickvl (approved):
🙌

fa9r:
> @fa9r I do not have a Mac unfortunately, so if you could help debug this on yours that would be great!

This indeed seems to be just a GitHub runner issue. I just tried installing and using the materializer on my M1 Mac under Python 3.8 and it worked fine. I would suggest we simply ignore the integration in the CI for now since we don't need it for linting/tests anyways.

In particular, I'd suggest you add `--ignore-integration pycaret` at the end of line 76 of `.github/actions/setup_environment/action.yml`, that should fix it

fa9r:
As for the other CI issues, they are again not related to your PR, llama index just released a new breaking version that is no longer compatible with our integration, we're fixing that in another PR right now

christianversloot:
Ok, thanks! 😄

fa9r:
Awesome, let's get this merged, thanks again for contributing 🙌

PR #1472, opened 2023-04-05
Artifact Visualization

fa9r:
## Describe changes

This PR adds functionality to visualize artifacts in the dashboard and deprecates all existing visualizer classes.

Visualizations can now be created by overriding the `save_visualization` method of the materializer that handles an artifact. To disable artifact visualization, steps or pipelines can be configured with `enable_artifact_visualization=False`.

Alternatively, artifact visualizations can now also be displayed in Jupyter notebooks using the new `visualize()` post execution methods.

### TODO
- Test on GCP/Azure
- Delete all visualizers directly (?)

### Supported Visualization Types
- **HTML:** used for embedded HTML visualizations; can be activated by returning an `zenml.types.HTMLString` from a step 
    -> data validation results: evidently reports, deepchecks results, great expectation suites, whylogs profiles, facets comparisons
- **Image:** used for image data 
    -> Pillow images or certain numeric numpy arrays
- **CSV:** Used for tables 
    -> pandas DataFrame `.describe()` output
- **Markdown:** currently unused but can be activated by returning a `zenml.types.MarkdownString` from a step.

### Implementation Details
- Materializers save visualizations of the artifact in the artifact store.
- For each visualization, we write an entry in the new `artifact_visualization` table.
- HTML and Markdown visualizations are defined by a new `HTMLMarkdownMaterializer`.
- A new `FacetsMaterializer` and several Facets standard steps were added to replace the Facets visualizer.
- Artifact store dependencies are now installed in the ZenML base image so the server can load visualizations from all artifact stores.
- A new endpoint `GET artifacts/{artifact_id}/visualize` can be used to fetch the visualization of an artifact.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [x] If my change requires a change to docs, I have updated the documentation accordingly.
- [x] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [x] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #1522, opened 2023-05-08
Raise 501 for Unauthenticated Artifact Stores

fa9r:
### Main Change
**artifact_utils.py Lines 325-332:** Bugfix following #1472. If the artifact store holding the visualization does not have authentication credentials built in, it would previously raise an implementation-specific 500 error, but we want to raise 501 so we can catch it in the frontend.

### Minor Changes
While at it, I also refactored the code a bit:
- renamed `materializer_utils.py` > `artifact_utils.py`
- split artifact visualization logic into two new helper functions `_load_artifact_store_of_artifact` and `_load_uri_from_artifact_store` that can be reused to load other things from the artifact store in the future.
- Moved type checking imports where they belong

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/artifact_utils.py
+++ b/src/zenml/utils/artifact_utils.py
@@ -11,30 +11,31 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 #  or implied. See the License for the specific language governing
 #  permissions and limitations under the License.
-"""Util functions for models and materializers."""
+"""Util functions for artifact handling."""
 
 import base64
 import os
 import tempfile
 from typing import TYPE_CHECKING, Any, Optional, Union, cast
 
-from zenml.artifact_stores.base_artifact_store import BaseArtifactStore
 from zenml.client import Client
-from zenml.config.source import Source
 from zenml.constants import MODEL_METADATA_YAML_FILE_NAME
 from zenml.enums import StackComponentType, VisualizationType
 from zenml.exceptions import DoesNotExistException
 from zenml.io import fileio
 from zenml.logger import get_logger
-from zenml.materializers.base_materializer import BaseMaterializer
 from zenml.models.visualization_models import LoadedVisualizationModel
 from zenml.stack import StackComponent
 from zenml.utils import source_utils
 from zenml.utils.yaml_utils import read_yaml, write_yaml
 
 if TYPE_CHECKING:
+    from zenml.artifact_stores.base_artifact_store import BaseArtifactStore
+    from zenml.config.source import Source
+    from zenml.materializers.base_materializer import BaseMaterializer
     from zenml.models import ArtifactResponseModel
     from zenml.zen_stores.base_zen_store import BaseZenStore
+
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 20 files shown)


src/zenml/utils/artifact_utils.py
```diff
@@ -243,30 +280,53 @@ def load_artifact_visualization(
...
     except FileNotFoundError:
         raise DoesNotExistException(
-            f"Artifact '{artifact.id}' could not be visualized because the "
-            f"visualization file '{visualization.uri}' was not be found in "
-            f"the underlying artifact store '{artifact_store_model.name}'."
+            f"File '{uri}' does not exist in artifact store "
+            f"'{artifact_store.name}'."
+        )
+    except Exception:
+        link = "https://docs.zenml.io/component-gallery/artifact-stores/custom#enabling-artifact-visualizations-with-custom-artifact-stores"
```
stefannica, line 326: Could you also please do a logger.exception() here so we don't lose the error info ? catching generic exceptions and then ignoring their source can be devastating for maintainability.

stefannica (changes requested):

src/zenml/utils/artifact_utils.py
```diff
@@ -215,24 +216,60 @@ def load_artifact_visualization(
...
+
+
+def _load_artifact_store_of_artifact(
+    artifact: "ArtifactResponseModel",
+    zen_store: Optional["BaseZenStore"] = None,
+) -> "BaseArtifactStore":
+    """Load the artifact store of the given artifact.
+
+    Args:
+        artifact: The artifact to load the artifact store of.
```
strickvl, line 257:
```suggestion
        artifact: The artifact for which to load the artifact store.
```

;)

src/zenml/utils/artifact_utils.py
```diff
@@ -215,24 +216,60 @@ def load_artifact_visualization(
...
+        zen_store: The ZenStore to use for finding the artifact store. If not
+            provided, the ZenStore of the client will be used.
```
strickvl, line 259:
```suggestion
            provided, the client's ZenStore will be used.
```

src/zenml/utils/artifact_utils.py
```diff
@@ -243,30 +280,53 @@ def load_artifact_visualization(
...
+    uri: str,
+    artifact_store: "BaseArtifactStore",
+    mode: str = "rb",
+) -> Any:
+    """Load the given uri from the given artifact store.
+
+    Args:
+        uri: The uri of the file to load.
+        artifact_store: The artifact store to load the file from.
+        mode: The mode to open the file in.
```
strickvl, line 308:
```suggestion
        artifact_store: The artifact store from which to load the file.
        mode: The mode in which to open the file.
```

src/zenml/utils/artifact_utils.py
```diff
@@ -243,30 +280,53 @@ def load_artifact_visualization(
...
+
+    Returns:
+        The loaded file.
```
strickvl, line 311: are we returning a loaded file, or a URI?

strickvl (changes requested):

src/zenml/utils/artifact_utils.py
fa9r: Done 👍

src/zenml/utils/artifact_utils.py
fa9r: A file. I guess the function name was confusing, I adjusted it accordingly.

strickvl (approved):
🚀

htahir1 (approved):

PR #1532, opened 2023-05-10
Define `@pipeline` and `@step` decorators as public API

fa9r:
## Describe changes

The pipeline and step decorators are now part of the public API and can be imported via `zenml.pipeline` and `zenml.step` directly.

### Detailed Changes
- moved both decorator definitions to a new `zenml.api` package
- adjusted `zenml.__init__.py` to import the decorators
- optimized imports so we don't need to import anything except for `typing` to load the decorators
- ~~adjusted all examples/... imports to use `from zenml import pipeline, step` syntax instead (thus all the file changes)~~

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/__init__.py
+++ b/src/zenml/__init__.py
@@ -34,5 +34,7 @@ __path__ = extend_path(__path__, __name__)
 
 # Define public Python API
 from zenml.api import show
+from zenml.new.pipelines.pipeline_decorator import pipeline
+from zenml.new.steps.step_decorator import step
 
-__all__ = ["show"]
+__all__ = ["show", "pipeline", "step"]
--- a/src/zenml/new/pipelines/pipeline_decorator.py
+++ b/src/zenml/new/pipelines/pipeline_decorator.py
@@ -24,18 +24,16 @@ from typing import (
     overload,
 )
 
-from zenml.new.pipelines.pipeline import Pipeline
-
 if TYPE_CHECKING:
     from zenml.config.base_settings import SettingsOrDict
+    from zenml.new.pipelines.pipeline import Pipeline
 
     HookSpecification = Union[str, FunctionType]
-
-F = TypeVar("F", bound=Callable[..., None])
+    F = TypeVar("F", bound=Callable[..., None])
 
 
 @overload
-def pipeline(_func: F) -> Pipeline:
+def pipeline(_func: "F") -> "Pipeline":
     ...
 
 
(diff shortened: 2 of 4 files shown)


strickvl (approved):
The core changes look good to me. Will look through the rest of the changes when they're re-reverted in case anything slips through, though testing should hopefully be the biggest barrier to pass with all that.

fa9r:
@alex-zenml all the integration tests already passed before I reverted the example changes, so I think those should be good. IMO not worth reviewing even, looking over 250 files that all just do the same two import changes

schustmi (commented):
I would highly suggest waiting with this until my new pipeline/step interface is merged. Reason for that is: At that point we will have multiple pipeline/step decorators (new and old for backwards compatibility) and we want to expose the new ones as our public API. However if we release this, we need to keep this backwards compatible and can't expose the new API here

fa9r:
@schustmi I agree, we should certainly wait with merging this till the new interface is out.

The main point was to get early feedback on how/where we could define that API and how to make it fast

schustmi:
What are the benefits of having it in a separate `api` module instead of just importing it in the root `__init__.py` directly?

But in general I'd love to have this, we should include even more classes though: `Client`, `StepContext`, `Output`, `ResourceSettings`, `DockerSettings`, `Schedule`, the post execution classes come to my mind

fa9r:
> What are the benefits of having it in a separate api module instead of just importing it in the root __init__.py directly?

To keep imports as small as possible. E.g., for the `@pipeline` decorator, if we would keep it in `zenml.pipelines.pipeline_decorator` and just import it in the root `__init__.py`, then that would import everything that `pipelines.__init__.py` imports. 

So I saw three options how to handle this:
- move public api to a separate `api` module that is highly import optimized (current solution)
- define all the public api in the root `__init__.py` directly 
-> not maintainable as our API grows
- keep functions where they are and remove submodule imports (e.g. delete all the imports in `pipelines.__init__.py`) 
-> not backwards compatible since we have been using, e.g., `from zenml.pipelines import ...` for years in all our docs/examples/... now

fa9r:
> When you import `from zenml.steps.step_decorator import step`

Yeah but if you import `from zenml import step` directly this will not import `zenml.steps` at all which is the main point

fa9r:
> 



> When you import `from zenml.steps.step_decorator import step`, this one will also import the `zenml/steps/__init__.py` module

Exactly, that's why I moved it to `zenml.api.step_decorator`, now only `zenml/api/__init__.py` gets imported (which is empty)

Issue #620, opened 2022-05-23
[FEATURE]: Allow secrets to be created by passing dicts or JSON values


strickvl:
### Contact Details [Optional]

support@zenml.io

### Describe the feature you'd like

We currently allow for the registration of secrets via an interactive CLI process, or by passing one or more `--secret_key=secret_value` as command line arguments:

```
$ zenml secret create -i area51
Entering interactive mode:
Please enter a secret key: location
Please enter the secret value for the key [location]:
Do you want to add another key-value pair to this secret? [y/n]: y
Please enter a secret key: aliens
Please enter the secret value for the key [aliens]:
Do you want to add another key-value pair to this secret? [y/n]: n
The following secret will be registered.
┏━━━━━━━━━━━━┯━━━━━━━━━━━━━━┓
┃ SECRET_KEY │ SECRET_VALUE ┃
┠────────────┼──────────────┨
┃  location  │ ***          ┃
┠────────────┼──────────────┨
┃   aliens   │ ***          ┃
┗━━━━━━━━━━━━┷━━━━━━━━━━━━━━┛
⠙ Saving secret `area51`...
Secret 'area51' successfully created.
```

```
$ zenml secret create area51 --location=Nevada --aliens=many
The following secret will be registered.
┏━━━━━━━━━━━━┯━━━━━━━━━━━━━━┓
┃ SECRET_KEY │ SECRET_VALUE ┃
┠────────────┼──────────────┨
┃  location  │ ***          ┃
┠────────────┼──────────────┨
┃   aliens   │ ***          ┃
┗━━━━━━━━━━━━┷━━━━━━━━━━━━━━┛
⠙ Saving secret `area51`...
Secret 'area51' successfully created.
```

We would like to enable users to pass in a `dict` or perhaps a JSON/YAML object containing multiple key-value pairs. The dict/JSON/YAML object could either be supplied as a command line argument, or by loading values from a file, e.g.:

```
$ zenml secret create area51 --values='{"location": "Nevada", "aliens":"many"}'
The following secret will be registered.
┏━━━━━━━━━━━━┯━━━━━━━━━━━━━━┓
┃ SECRET_KEY │ SECRET_VALUE ┃
┠────────────┼──────────────┨
┃  location  │ ***          ┃
┠────────────┼──────────────┨
┃   aliens   │ ***          ┃
┗━━━━━━━━━━━━┷━━━━━━━━━━━━━━┛
⠙ Saving secret `area51`...
[...]

PR #1547, opened 2023-05-19
Secrets input structure change method

bhatt-priyadutt:
## Describe changes
I have created Inital PR with logic added for **create_secret** function to accept arguments in the format specified in issue: https://github.com/zenml-io/zenml/issues/620

Currently, these is working for `zenml create secret` commands. I would like to have a feedback if i am doing things right so that i can move ahead with rest of the changes

## Pre-requisites

## Types of changes


src/zenml/cli/secret.py
```diff
@@ -840,9 +841,8 @@ def create_secret(
         interactive: Whether to use interactive mode to enter the secret values.
         args: The arguments to pass to the secret.
     """
-    name, parsed_args = parse_name_and_extra_arguments(  # type: ignore[assignment]
-        list(args) + [name], expand_args=True
-    )
+    name, parsed_args = parse_secret_name_and_arguements(  # type: ignore[assignment]
```
stefannica, line 844: May I suggest a slightly different approach with a somewhat better UX?

Instead of expecting a `--values` or `--values-file` argument here, you could do the following:
* add a single `--values` string click option to the CLI command (just like `--scope` or `--interactive`)
* keep this line unchanged, because now the `values` option will be parsed independently of arguments by click
* parse the `values` option value using the same logic we do with regular arguments: the value could be an inline JSON/YAML encoded string, or it could be a filepath starting with the special character `@`

For example:

```
zenml secret create my-secret --values='{'location': 'Nevada', 'aliens': 'many'}'

zenml secret create my-secret --values=@path/to/my/file.yaml
```

src/zenml/cli/utils.py
```diff
@@ -16,6 +16,9 @@
 import os
 import subprocess
 import sys
+import json
+import yaml
+import ast
```
stefannica, line 21: you shouldn't need to use an additional library like ast to evaluate the JSON/YAML contents. Just use `yaml.safe_load` and try/catch any errors that it may throw during parsing.

src/zenml/cli/secret.py
stefannica: You can use this utility to expand the `values` option value into the file contents, if it points to a file: https://github.com/zenml-io/zenml/blob/develop/src/zenml/cli/utils.py#L533

stefannica (changes requested):
Many thanks for your contribution! Overall I would say you are on the right track. I made some suggestions that will help you achieve a better UX.

src/zenml/cli/secret.py
bhatt-priyadutt: agreed, but I got this format structure based on @alex-zenml 's description in below issue:
https://github.com/zenml-io/zenml/issues/620

src/zenml/cli/utils.py
```diff
@@ -555,6 +558,83 @@ def expand_argument_value_from_file(name: str, value: str) -> str:
...
+    try:
+        dict_ = yaml.safe_load(string)
+        return dict_
+    except yaml.YAMLError:
+        pass
+
+    error(f"Invalid argument: '{string}'. Please provide the value in Json/Dict/Yaml format")
+
+
+def read_args_from_file(value: str) -> str:
```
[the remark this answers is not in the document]
bhatt-priyadutt: Below is the code that is not required for new secret format, so either i can get a logic and fit in with that old function itself to avoid as i cant remove this logic as its used by other functions as well.
https://github.com/zenml-io/zenml/blob/e055d318b326cdc305301d5130b2f2fa46d613e0/src/zenml/cli/utils.py#LL551C1-L554C21

src/zenml/cli/utils.py
stefannica: I understand. Perhaps we've been too broad with the #620 issue description and included `dict` in the list of possible format this new value can take. We should have limited that to JSON/YAML. Using `dict` is not advisable because using `eval` on user-entered input is not a "safe" operation and could have damaging results. We don't use that anywhere else in ZenML for that same reason.

src/zenml/cli/secret.py
stefannica: I see what you mean, but that issue is not set in stone, more like a set of guidelines. You can improvise when it makes sense.

src/zenml/cli/utils.py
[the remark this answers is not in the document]
stefannica: If you take my suggestion and allow `--values=@path/to/file`, then this code will also required for the new secret format.

bhatt-priyadutt:
I have added test cases for the logic, also I see a conflict that needs to be resolved, before merge

strickvl:
Feel free to resolve the conflict on your end, btw. Just merge develop back into this branch and you'll be able to make that fix. Thanks!

bhatt-priyadutt:
@strickvl  I have resolved conflicts, I am almost done with test cases and code changes and before I go ahead with documentation changes, I have also applied the same format of specifying secret key,value pairs to the `secret update` command thinking that it will be useful there as well as users will be able to bulk update their credentials. Please let me know if this is fine, thanks.
CC: @stefannica

bhatt-priyadutt:
I have added the test cases as well as updated the docs, can someone please review, thanks

docs/book/platform-guide/set-up-your-mlops-platform/use-the-secret-store/interact-with-secrets.md
```diff
@@ -28,12 +27,11 @@ Alternatively, you can create the secret in an interactive session (in which Zen
 zenml secret create <SECRET_NAME> -i
 ```
 
-For secret values that are too big to pass as a command line argument, or have special characters, you can also use the special `@` syntax to indicate to ZenML that the value needs to be read from a file:
+For secret key-value pairs that are too big to pass as a command line argument, or have special characters, you can also use the special `@` syntax to indicate to ZenML that the key-value pair (Json/Yaml) needs to be read from a file:
```
strickvl, line 30:
```suggestion
For secret key-value pairs that are too big to pass as a command line argument, or have special characters, you can also use the special `@` syntax to indicate to ZenML that the key-value pair (JSON or YAML) needs to be read from a file:
```

src/zenml/cli/utils.py
```diff
@@ -655,6 +710,70 @@ def parse_name_and_extra_arguments(
     return name, args_dict
 
 
+def parse_secret_name_and_arguements(
```
strickvl, line 713:
```suggestion
def parse_secret_name_and_arguments(
```

strickvl: And any associated changes

src/zenml/cli/utils.py
```diff
@@ -578,6 +584,71 @@ def expand_argument_value_from_file(name: str, value: str) -> str:
         )
 
 
+def convert_str_to_dict(string) -> dict:
+    """
+    Converts String into dict type
```
strickvl, line 589:
```suggestion
def convert_structured_str_to_dict(string) -> dict:
    """Convert a structured string (JSON or YAML) into a dict.
```

strickvl: and any associated changes

src/zenml/cli/utils.py
```diff
@@ -578,6 +584,71 @@ def expand_argument_value_from_file(name: str, value: str) -> str:
...
+    error(
+        f"Invalid argument: '{string}'. Please provide the value in Json/Yaml format"
+    )
+
+
+def extract_name_from_args(args: List[str], name_mandatory: bool):
+    """
+    If the name was not supplied as the first argument, we have to
+    search the other arguments for the name.
+    Args:
```
strickvl, line 627:
```suggestion
    """Extract name from arguments.
    
    If the name was not supplied as the first argument, we have to
    search the other arguments for the name.
    
    Args:
```

src/zenml/cli/utils.py
```diff
@@ -578,6 +584,71 @@ def expand_argument_value_from_file(name: str, value: str) -> str:
...
+        args: A list of command line arguments from the CLI.
+        name_mandatory: Whether the name argument is mandatory.
+
+    Returns:
+        name: returns name from the list of arguments
+        args: returns args that contains only the secret key value paris
```
strickvl, line 633:
```suggestion
        name: name from the list of arguments
        args: args that contains only the secret key value pairs
```

src/zenml/cli/utils.py
```diff
@@ -655,6 +710,70 @@ def parse_name_and_extra_arguments(
...
+        name_mandatory: Whether the name argument is mandatory.
+
+    Returns:
+        The name and a dict of parsed args.
+    """
+    name, args = extract_name_from_args(args, name_mandatory)
+    message = (
+        "Please provide args with a proper "
+        "identifier as the key and the value with the following structure: "
+        "--values='value' (Json/Yaml) or --values=@file_path"
```
strickvl, line 749:
```suggestion
        "--values='value' (JSON/YAML) or --values=@file_path"
```

strickvl (commented):
Just some high-level comments from a quick review, and also there are some docstring errors in the CI.

docs/book/platform-guide/set-up-your-mlops-platform/use-the-secret-store/interact-with-secrets.md
```diff
@@ -76,8 +74,7 @@ By default, all created secrets are scoped to the active workspace. To create a
 ```shell
 zenml secret create <SECRET_NAME> \
     --scope user \
-    --<KEY_1>=<VALUE_1> \
-    --<KEY_2>=<VALUE_2>
+    --values='{"key1":"value2","key2":"value2"}'
```
stefannica, line 77: same as before: document the new option _in addition to the new one_, not by replacing it.

docs/book/platform-guide/set-up-your-mlops-platform/use-the-secret-store/interact-with-secrets.md
```diff
@@ -104,8 +101,7 @@ In the dashboard values that are considered `Secret` can be set using registered
 
 # Using central secrets management
 zenml secret create mlflow_secret \
-    --username=admin \
-    --password=abc123
+    --values='{"username":"admin","password":"abc123"}'
```
stefannica, line 104: same here

docs/book/user-guide/advanced-guide/use-failure-success-hooks.md
```diff
@@ -119,7 +119,7 @@ First, ensure that you have the OpenAI integration installed and have stored you
 
 ```shell
 zenml integration install openai
-zenml secret create openai --api_key=<YOUR_API_KEY>
+zenml secret create openai --values='{"api_key":"<YOUR_API_KEY>"}'
```
stefannica, line 122: in this case it isn't even about keeping the documentation about the current CLI option unchanged. You should just look at the two forms and ask yourself "which of the two would I rather use ?". I, for one, would definitely not like having to type the extra brackets and quotes needed for JSON when the previous option is so much cleaner and easier. Wouldn't you agree ?

docs/book/user-guide/component-guide/model-deployers/seldon.md
```diff
@@ -153,50 +153,50 @@ the [`rclone` documentation for various providers](https://rclone.org/).
 Example of configuring a Seldon Core secret for AWS S3:
 
 ```shell
-zenml secret create s3-seldon-secret \
---rclone_config_s3_type="s3" \ # set to 's3' for S3 storage.
---rclone_config_s3_provider="aws" \ # the S3 provider (e.g. aws, Ceph, Minio).
---rclone_config_s3_env_auth=False \ # set to true to use implicit AWS authentication from EC2/ECS meta data
+zenml secret create s3-seldon-secret --values= \
```
stefannica, line 156: This is one example where storing the values as a YAML file and passing it using the `--values=@path/to/file.yaml` would be preferable to writing this whole complicated, multi-line string in the command line.

You could try the same exercise as before here: ask yourself what you would choose to do here to make your life easier:

* use multiple `--key=value` arguments
* use the inline `--values=<JSON>` option
* use the `--values=@path/to/file.yaml` option

Putting yourself in the user's shoes is the best way to understand how your work will impact everyone else.

tests/integration/functional/cli/test_secret.py
```diff
@@ -35,7 +35,10 @@ def test_create_secret():
     with cleanup_secrets() as secret_name:
         result = runner.invoke(
             secret_create_command,
-            [secret_name, "--test_value=aria", "--test_value2=axl"],
+            [
+                secret_name,
+                '--values={"test_value":"aria","test_value2":"axl"}',
+            ],
```
stefannica, line 41: You need to test the new option _in addition_ to the current one. Both options are supported.

tests/integration/functional/cli/test_secret.py
```diff
@@ -51,7 +54,11 @@ def test_create_secret_with_scope():
     with cleanup_secrets() as secret_name:
         result = runner.invoke(
             secret_create_command,
-            [secret_name, "--test_value=aria", f"--scope={SecretScope.USER}"],
+            [
+                secret_name,
+                '--values={"test_value":"aria"}',
+                f"--scope={SecretScope.USER}",
+            ],
```
stefannica, line 61: same here and everywhere else in the tests. You don't need to edit all tests though. Just writing a few new tests to target the new option during create/update is enough.

src/zenml/cli/secret.py
```diff
@@ -914,8 +915,8 @@ def update_secret(
         interactive: Whether to use interactive mode to update the secret.
         remove_keys: The keys to remove from the secret.
     """
-    name, parsed_args = parse_name_and_extra_arguments(
-        list(extra_args) + [name_or_id], expand_args=True
+    name, parsed_args = parse_secret_name_and_arguements(
```
stefannica, line 918: here same as for the create command: you should add a new click.option instead.

stefannica (changes requested):
As a general comment, please remember you are not replacing the current `--key=value` option with the new `--values=json/yaml` option. Both need to be supported, documented and tested.

I wrote more around my suggestion to avoid code duplication, hopefully this time around it is explained more clearly.

bhatt-priyadutt:
Ok got it, sorry from what I knew, i thought I need to replace the existing way we specify the key-value pairs. thanks for the comments will look into and work accordingly.

src/zenml/cli/secret.py
bhatt-priyadutt: I think we should also stop user from specifying both at the same time in the command like below - 
`zenml secret create foo --values=JSON --"a"="b"`

bhatt-priyadutt:
I have updated the changes according to the comments and below are my points regarding the changes:-

1. I was under the impression that we need to replace the existing specification of key-value pairs, now that its not the case...still we should not allow from user to input both formats in one command at the same time like below -
`zenml secret create --values='{"a":"b"}' --c=f`
`zenml secret update --values='{"a":"c"}' --a=d`
please acknowledge that this behaviour is not allowed and i will add this constraint.

2. In the docs, I have added at places for alternative suggestions of relevant formats only, for example at a place where there are multiple key-value pairs, I have added to use file format and where there are fewer key-value pairs, I have kept the same format using the default arguments as its fine there.

3. I found a bug where in while inputting the key-value pairs in arguments, we are checking if a **key** is a valid string by using the python inbuilt function .isidentifier() (https://github.com/zenml-io/zenml/blob/bfebc60dd400227b3626d56968ce7af33bd9dc3a/src/zenml/cli/utils.py#LL645C1-L645C1) . But, we are not checking in case of taking input in **interactive** mode and now as well as in **values** option. Maybe, this can be taken care of in different PR.

stefannica (approved):
This looks great now, thanks for following my suggestions ! You just need to fix the linter errors and you should be good to go.

stefannica:
> I have updated the changes according to the comments and below are my points regarding the changes:-
> 
> 1. I was under the impression that we need to replace the existing specification of key-value pairs, now that its not the case...still we should not allow from user to input both formats in one command at the same time like below -
>    `zenml secret create --values='{"a":"b"}' --c=f`
>    `zenml secret update --values='{"a":"c"}' --a=d`
>    please acknowledge that this behaviour is not allowed and i will add this constraint.

I see no harm in allowing it. It gives more flexibility.

> 2. In the docs, I have added at places for alternative suggestions of relevant formats only, for example at a place where there are multiple key-value pairs, I have added to use file format and where there are fewer key-value pairs, I have kept the same format using the default arguments as its fine there.

Wonderful, thank you !

> 3. I found a bug where in while inputting the key-value pairs in arguments, we are checking if a **key** is a valid string by using the python inbuilt function .isidentifier() (https://github.com/zenml-io/zenml/blob/bfebc60dd400227b3626d56968ce7af33bd9dc3a/src/zenml/cli/utils.py#LL645C1-L645C1) . But, we are not checking in case of taking input in **interactive** mode and now as well as in **values** option. Maybe, this can be taken care of in different PR.

Sure, different PR will do just fine.

bhatt-priyadutt:
I am done with fixing linting and annotation errors, i see the last 3 remaining checks were maybe aborted or cancelled, please look into it or let me know if i need to do anything else, thanks.

stefannica:
> I am done with fixing linting and annotation errors, i see the last 3 remaining checks were maybe aborted or cancelled, please look into it or let me know if i need to do anything else, thanks.

We've been having some problems with long running windows CI tests lately. I re-triggered those, but I would say this is good to go otherwise.

PR #1556, opened 2023-05-23
Easy CSV Visualization

fa9r:
I added a `zenml.types.CSVString` special string class that can be returned from steps to automatically visualize the corresponding string as CSV in the dashboard.

### Usage Example:

```python
from zenml.types import CSVString

@step
def my_step() -> CSVString:
    some_csv = "a,b,c\n1,2,3"
    return CSVString(some_csv)
```

### Resulting Visualization in Dashboard:

Screenshot 2023-05-23 at 22 50 27

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/materializers/special_string_materializer.py
+++ b/src/zenml/materializers/special_string_materializer.py
@@ -20,24 +20,25 @@ from zenml.enums import ArtifactType, VisualizationType
 from zenml.io import fileio
 from zenml.logger import get_logger
 from zenml.materializers.base_materializer import BaseMaterializer
-from zenml.types import HTMLString, MarkdownString
+from zenml.types import CSVString, HTMLString, MarkdownString
 
 logger = get_logger(__name__)
 
 
-HTML_OR_MARKDOWN = Union[HTMLString, MarkdownString]
+SPECIAL_STRINGS = Union[CSVString, HTMLString, MarkdownString]
 
 HTML_FILENAME = "output.html"
 MARKDOWN_FILENAME = "output.md"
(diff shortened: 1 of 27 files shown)


src/zenml/materializers/special_string_materializer.py
```diff
@@ -20,24 +20,25 @@
...
 from zenml.logger import get_logger
 from zenml.materializers.base_materializer import BaseMaterializer
-from zenml.types import HTMLString, MarkdownString
+from zenml.types import CSVString, HTMLString, MarkdownString
 
 logger = get_logger(__name__)
 
 
-HTML_OR_MARKDOWN = Union[HTMLString, MarkdownString]
+SPECIAL_STRINGS = Union[CSVString, HTMLString, MarkdownString]
```
strickvl, line 28: How about 'STRUCTURED_STRINGS' instead? Somehow SPECIAL could be lots of things, like those containing special characters etc.

strickvl (commented):
Broadly fine with this, though wondering a bit about the name. Also will probably want a Docs Update alongside, otherwise people won't use these materialisers, no?

htahir1 (commented):
I dont have any idea how this works so im just maybe leave a question: Is this bakcwards- compatible? :-D

fa9r:
@htahir1 for existing visualizations yes. However, loading artifacts saved by the materializer is no longer possible since the code reference to the materializer has changed by renaming it. That will always happen if you rename/move materializers though.

strickvl (approved):

AlexejPenner (approved):

PR #1567, opened 2023-05-26
Fixing neural prophet test case error

bhatt-priyadutt:
## Describe changes
image

Downgrading holidays package version from 0.25 to 0.24 fixes this issue. This issue is coming everywhere for that particular version, I was getting the same using greykite package as it also uses holidays.

## Pre-requisites

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)


htahir1:
@bhatt-priyadutt Thanks for the PR but we dont pin integration specific packages in the pyproject.toml (thats only for zenml dependencies, and zenml does not reply on `holidays`. Instead we pin them in the integration itself. In this case, we have already fixed this bug [here](https://github.com/zenml-io/zenml/blob/main/src/zenml/integrations/neural_prophet/__init__.py#L24)

bhatt-priyadutt:
Oh Okay got it, maybe we can close this then.

PR #1573, opened 2023-05-30
Add option to list all resources when verifying service connector config

stefannica:
## Describe changes
Adds support to list all resources when verifying a service connector config. This is to make it possible for the dashboard to fetch all the information it needs to render the list of possible resources that can be accessed by a multi-type service connector (a service connector that can access multiple types of resources, like an AWS connector that can access S3 buckets, EKS clusters and ECR registries).

The core change of this PR is allowing resources from multiple types to be listed in the same `ServiceConnectorResourcesModel` object:

* previously:
```json
{
  "id": null,
  "name": null,
  "connector_type": {},
  "resource_type": "aws-generic",
  "resource_ids": [
        "us-east-1"
      ],
  "error": null,
   ...
}
```

* currently:
```json
{
  "id": null,
  "name": null,
  "connector_type": {},
  "resources": [
    {
      "resource_type": "aws-generic",
      "resource_ids": [
        "us-east-1"
      ],
      "error": null
    },
    {
      "resource_type": "s3-bucket",
      "resource_ids": [
        "s3://sagemaker-studio-907999144431-m11qlsdyqr8",
        "s3://sagemaker-studio-d8a14tvjsmb"
      ],
      "error": null
    },
   ...
}
```

This is an example of how the dashboard uses this new format returned in the `/verify` endpoint:

Register Connector Configuration - Stage 2

The same is reflected in the CLI now:

```
$ zenml service-connector verify aws-connectors
Service connector 'aws-connectors' is correctly configured with valid credentials and has access to the following resources:
┏━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃     RESOURCE TYPE     │ RESOURCE NAMES                               ┃
┠───────────────────────┼──────────────────────────────────────────────┨
┃    🔶 aws-generic     │ us-east-1                                    ┃
┠───────────────────────┼──────────────────────────────────────────────┨
┃     📦 s3-bucket      │ s3://aws-ia-mwaa-715803424590                ┃
┃                       │ s3://zenbytes-bucket                         ┃
┃                       │ s3://zenfiles                                ┃
┃                       │ s3://zenml-demos                             ┃
┃                       │ s3://zenml-generative-chat                   ┃
┃                       │ s3://zenml-public-datasets                   ┃
┃                       │ s3://zenml-public-swagger-spec               ┃
┃                       │ s3://zenml-sandbox-infra                     ┃
┃                       │ s3://zenml-terraform-ci                      ┃
┠───────────────────────┼──────────────────────────────────────────────┨
┃ 🌀 kubernetes-cluster │ zenhacks-cluster                             ┃
┠───────────────────────┼──────────────────────────────────────────────┨
┃  🐳 docker-registry   │ 715803424590.dkr.ecr.us-east-1.amazonaws.com ┃
┗━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -4475,6 +4475,7 @@ class Client(metaclass=ClientMetaClass):
         labels: Optional[Dict[str, str]] = None,
         auto_configure: bool = False,
         verify: bool = True,
+        list_resources: bool = True,
         register: bool = True,
     ) -> Tuple[
         Optional[
@@ -4505,6 +4506,8 @@ class Client(metaclass=ClientMetaClass):
                 connector from the local environment.
             verify: Whether to verify that the service connector configuration
                 and credentials can be used to gain access to the resource.
+            list_resources: Whether to also list the resources that the service
+                connector can give access to (if verify is True).
             register: Whether to register the service connector or not.
 
         Returns:
@@ -4538,8 +4541,9 @@ class Client(metaclass=ClientMetaClass):
                 "Python packages and ZenML integrations and try again."
             )
 
-        if not resource_type and len(connector.resource_types) == 1:
-            resource_type = connector.resource_types[0].resource_type
+        if not resource_type:
+            if len(connector.resource_types) == 1:
+                resource_type = connector.resource_types[0].resource_type
 
         # If auto_configure is set, we will try to automatically configure the
         # service connector from the local environment
@@ -4565,7 +4569,6 @@ class Client(metaclass=ClientMetaClass):
                 auth_method=auth_method,
                 resource_id=resource_id,
             )
-
             assert connector_instance is not None
             connector_request = connector_instance.to_model(
                 name=name,
@@ -4585,11 +4588,18 @@ class Client(metaclass=ClientMetaClass):
                 if connector.remote:
                     connector_resources = (
                         self.zen_store.verify_service_connector_config(
-                            connector_request
+                            connector_request,
+                            list_resources=list_resources,
                         )
                     )
                 else:
-                    connector_resources = connector_instance.verify()
+                    connector_resources = connector_instance.verify(
+                        list_resources=list_resources,
+                    )
+
+                if connector_resources.error:
+                    # Raise an exception if the connector verification failed
+                    raise AuthorizationException(connector_resources.error)
 
         else:
             if not auth_method:
@@ -4631,7 +4641,8 @@ class Client(metaclass=ClientMetaClass):
                 if connector.remote:
                     connector_resources = (
                         self.zen_store.verify_service_connector_config(
-                            connector_request
+                            connector_request,
+                            list_resources=list_resources,
                         )
                     )
                 else:
@@ -4640,23 +4651,25 @@ class Client(metaclass=ClientMetaClass):
                             model=connector_request
                         )
                     )
-                    connector_resources = connector_instance.verify()
+                    connector_resources = connector_instance.verify(
+                        list_resources=list_resources,
+                    )
 
                 if connector_resources.error:
+                    # Raise an exception if the connector verification failed
                     raise AuthorizationException(connector_resources.error)
+
+                # For resource types that don't support multi-instances, it's
+                # better to save the default resource ID in the connector, if
+                # available. Otherwise, we'll need to instantiate the connector
+                # again to get the default resource ID.
+                connector_request.resource_id = (
+                    connector_request.resource_id
+                    or connector_resources.get_default_resource_id()
+                )
 
         if not register:
             return connector_request, connector_resources
-
-        # For resource types that don't support multi-instances, it's
-        # better to save the default resource ID in the connector, if
-        # available. Otherwise, we'll need to instantiate the connector
-        # again to get the default resource ID.
-        if connector_resources:
-            connector_request.resource_id = (
-                connector_request.resource_id
-                or connector_resources.get_default_resource_id()
-            )
 
         # Register the new model
         connector_response = self.zen_store.create_service_connector(
@@ -4685,6 +4698,7 @@ class Client(metaclass=ClientMetaClass):
         is_shared: Optional[bool] = None,
         labels: Optional[Dict[str, Optional[str]]] = None,
         verify: bool = True,
+        list_resources: bool = True,
         update: bool = True,
     ) -> Tuple[
         Optional[
@@ -4730,6 +4744,8 @@ class Client(metaclass=ClientMetaClass):
                 is set to None, the label will be removed.
             verify: Whether to verify that the service connector configuration
                 and credentials can be used to gain access to the resource.
+            list_resources: Whether to also list the resources that the service
+                connector can give access to (if verify is True).
             update: Whether to update the service connector or not.
 
         Returns:
@@ -4768,8 +4784,9 @@ class Client(metaclass=ClientMetaClass):
         else:
             resource_types = resource_type
 
-        if not resource_type and len(connector.resource_types) == 1:
-            resource_types = connector.resource_types[0].resource_type
+        if not resource_type:
+            if len(connector.resource_types) == 1:
+                resource_types = connector.resource_types[0].resource_type
 
         if resource_id == "":
             resource_id = None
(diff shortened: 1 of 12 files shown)


src/zenml/client.py
```diff
@@ -4538,8 +4541,9 @@ def create_service_connector(
                 "Python packages and ZenML integrations and try again."
             )
 
-        if not resource_type and len(connector.resource_types) == 1:
-            resource_type = connector.resource_types[0].resource_type
+        if not resource_type:
+            if len(connector.resource_types) == 1:
+                resource_type = connector.resource_types[0].resource_type
```
fa9r, line 4546: Why was this changed? Old syntax seemed cleaner IMO

src/zenml/client.py
```diff
@@ -4768,8 +4784,9 @@ def update_service_connector(
         else:
             resource_types = resource_type
 
-        if not resource_type and len(connector.resource_types) == 1:
-            resource_types = connector.resource_types[0].resource_type
+        if not resource_type:
+            if len(connector.resource_types) == 1:
+                resource_types = connector.resource_types[0].resource_type
```
fa9r, line 4789: same here

src/zenml/models/service_connector_models.py
```diff
@@ -857,6 +857,39 @@ def is_satisfied_by(
...
+        title="The type of resource that the service connector instance can "
+        "be used to access.",
+        max_length=STR_FIELD_MAX_LENGTH,
+    )
+
+    resource_ids: Optional[List[str]] = Field(
+        default=None,
+        title="The resource IDs of all resource instances that the service "
+        "connector instance can be used to access. Omitted (set to None) for "
+        "multi-type service connectors that didn't explicitly requested to "
```
fa9r, line 877:
```suggestion
        "multi-type service connectors that didn't explicitly request to "
```

src/zenml/models/service_connector_models.py
```diff
@@ -882,30 +915,92 @@ class ServiceConnectorResourcesModel(BaseModel):
...
-        "connector instance can be used to access. Omitted (set to None) for "
-        "multi-type service connectors. Also omitted if an error occurred "
-        "while listing the resource instances or if no resources are listed "
-        "due to authorization issues or lack of permissions (in both cases the "
-        "'error' field is set to an error message). For resource types that "
-        "do not support multiple instances, a single resource ID is listed.",
+    resources: List[ServiceConnectorTypedResourcesModel] = Field(
+        default_factory=list,
+        title="The list of resources that the service connector instance can "
+        "be used to access. Contains one entry for every resource type "
```
fa9r, line 921:
```suggestion
        "give access to. Contains one entry for every resource type "
```

src/zenml/models/service_connector_models.py
```diff
@@ -882,30 +915,92 @@ class ServiceConnectorResourcesModel(BaseModel):
...
+            resource.error = error
+            resource.resource_ids = None
+            if len(self.resources) == 1:
+                # If there is only one resource type involved, set the global
+                # error message as well.
+                self.error = error
+        else:
+            self.error = error
+            for resource in self.resources:
+                resource.error = error
```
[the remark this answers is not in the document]
fa9r: If it does need to happen, maybe consider splitting the function into `set_global_error` and `set_resource_error` or similar

src/zenml/models/service_connector_models.py
```diff
@@ -929,46 +1024,36 @@ def emojified_connector_type(self) -> str:
 
         return self.connector_type
 
-    @property
-    def emojified_resource_types(self) -> Optional[List[str]]:
-        """Get the emojified resource types.
+    def get_emojified_resource_types(
```
fa9r, line 1027: **emojified**, love it ^^  🌺🌈🎉🌞🤗

src/zenml/service_connectors/service_connector.py
```diff
@@ -1128,98 +1129,149 @@ def verify(
...
+            except Exception as exc:
+                error = (
+                    f"connector{name_msg} verification failed with "
+                    f"unexpected error: {exc}"
+                )
+                # Log an exception if debug logging is enabled
+                if logger.isEnabledFor(logging.DEBUG):
+                    logger.exception(error)
+                else:
+                    logger.warning(error)
```
fa9r, line 1229: Seems odd, why?

fa9r (approved):
Looks good to me, UX-wise this is a great improvement for multi-type connectors, looking forward to getting this live 👍

fa9r:
Oh, also, I'd love if we could add a unit test for the `list_resources` behavior, and maybe another one for `print_service_connector_resource_table` that checks that the result looks reasonable for both options of `show_resources_only`.

src/zenml/client.py
stefannica: probably the result of going through several stages to get this right. I'll change it back.

src/zenml/client.py
stefannica: I'll change it back

src/zenml/client.py
```diff
@@ -4849,24 +4867,25 @@ def update_service_connector(
...
-
-        # For resource types that don't support multi-instances, it's
-        # better to save the default resource ID in the connector, if
-        # available. Otherwise, we'll need to instantiate the connector
-        # again to get the default resource ID.
-        if connector_resources:
+            # For resource types that don't support multi-instances, it's
+            # better to save the default resource ID in the connector, if
+            # available. Otherwise, we'll need to instantiate the connector
+            # again to get the default resource ID.
```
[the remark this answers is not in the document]
stefannica: yes, you're correct, the "create service connector" and "update service connector" methods contain code that is very similar. However, the differences are subtle enough and spread out across several lines of code, which would make the code unreadable if I were to extract the common pieces out into a separate parameterized function.

src/zenml/service_connectors/service_connector.py
stefannica: This is a choice I had to make to keep the CLI output cleaner. Basically, if you have a lot of connectors that spit out errors, you end up with screens full of exception traces and error messages before you get the final result, which can be really annoying. With this in place, you only get the final messages as warnings.

stefannica:
> Oh, also, I'd love if we could add a unit test for the `list_resources` behavior, and maybe another one for `print_service_connector_resource_table` that checks that the result looks reasonable for both options of `show_resources_only`.

I'll see what I can do. It's not easy to implement unit-testing for any of this, given that we rely on real-life credentials and external services for the authentication part. I need to build something like an entirely fake service connector implementation. Definitely worth doing, but something that will take time and should be the subject of a dedicated PR.

schustmi (changes requested):

src/zenml/service_connectors/service_connector.py
```diff
@@ -1128,98 +1129,149 @@ def verify(
...
-                logger.warning(resources.error)
-            return resources
+            for rt in resource_types
+        ]
 
         if self.has_expired():
-            resources.error = (
-                "the connector's authentication credentials have expired."
-            )
+            error = "the connector's authentication credentials have expired."
```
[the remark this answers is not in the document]
stefannica: Yes, it's intended behavior: this flag only controls whether the `resource_ids` fields are populated or not, but the `resources` list is always included. I changed the method docstring to make that a bit more clear.

src/zenml/service_connectors/service_connector.py
```diff
@@ -1128,98 +1129,149 @@ def verify(
...
-            resources.error = (
-                "an unexpected error occurred while verifying the connector."
-            )
-            return resources
+                resource_ids = self._verify(
+                    resource_type=resource_type,
+                    resource_id=resource_id,
+                )
+            except ValueError as exc:
+                logger.debug(f"connector verification failed: {exc}")
```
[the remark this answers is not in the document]
stefannica: you're right, we don't need it

schustmi (approved):

PR #1578, opened 2023-06-01
Fix TensorFlow Requirements for Rosetta Mac

fa9r:
## Describe changes
If you run on an M1/M2 Mac under Rosetta, the TF requirements are currently set incorrectly because `platform.machine() == 'x86_64'` under Rosetta. 

I guess `tensorflow-macos` should work under all Mac versions so I simply removed the platform requirement, but I would be happy to implement this differently if there are any better suggestions.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/tensorflow/__init__.py
+++ b/src/zenml/integrations/tensorflow/__init__.py
@@ -30,10 +30,7 @@ class TensorflowIntegration(Integration):
         """Activates the integration."""
         # need to import this explicitly to load the Tensorflow file IO support
         # for S3 and other file systems
-        if (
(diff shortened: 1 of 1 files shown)


fa9r:
Hmm apparently this breaks the CI runners. Any ideas how we can distinguish Rosetta M1 Macs from regular Intel Macs?

fa9r:
Apparently, `tensorflow-macos` does not work under all Mac versions, at least the CI runners break. Thus, the proposed change doesn't make sense and since it's a niche issue it is probably not worth further investigation.

PR #1579, opened 2023-06-01
Add orchestrator outputs for sandbox examples

strickvl:
Added outputs for the Kubeflow orchestrator if used to the two examples loaded in the sandbox.

## Pre-requisites

## Types of changes

### DIFF
--- a/examples/great_expectations_data_validation/run.py
+++ b/examples/great_expectations_data_validation/run.py
@@ -12,8 +12,18 @@
 #  or implied. See the License for the specific language governing
 #  permissions and limitations under the License.
 
-
 from pipelines.validation import validation_pipeline
+
+from zenml.client import Client
 
 if __name__ == "__main__":
     validation_pipeline()
+
+    c = Client()
+    orchestrator = c.active_stack.orchestrator
+    orchestrator_config = orchestrator.config
+    if orchestrator_config.get("kubeflow_hostname"):
+        print("Kubeflow URL: `{orchestrator_config.kubeflow_hostname}`")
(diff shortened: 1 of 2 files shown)


examples/great_expectations_data_validation/run.py
```diff
@@ -12,8 +12,15 @@
...
+from zenml.client import Client
+
 if __name__ == "__main__":
     validation_pipeline()
+
+    c = Client()
+    orchestrator = c.active_stack.orchestrator
+    orchestrator_config = orchestrator.config
+    if orchestrator_config.get("kubeflow_hostname"):
+        print("Kubeflow URL: `{orchestrator_config.kubeflow_hostname}`")
```
htahir1, line 26: image
via https://docs.zenml.io/user-guide/starter-guide/fetch-runs-after-execution

I think we can use the post execution for this

htahir1, line 26: Maybe make it visually split up like print("\n\n************See kubeflow run***********\n")

htahir1, line 26: Similarly we can use `experiment_tracker_url` to get the experiment tracker run 
image

htahir1 (changes requested):
Thanks , this helped me formulate what I wanted and I left comments

examples/great_expectations_data_validation/run.py
strickvl: This I don't quite understand. The great expectations pipeline has no experiment tracker.

PR #1589, opened 2023-06-02
Implement Azure service connector

stefannica:
## Describe changes
Implement the Azure service connector.

```
$ zenml service-connector list-types --type azure
┏━━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━┯━━━━━━━┯━━━━━━━━┓
┃          NAME           │ TYPE     │ RESOURCE TYPES        │ AUTH METHODS      │ LOCAL │ REMOTE ┃
┠─────────────────────────┼──────────┼───────────────────────┼───────────────────┼───────┼────────┨
┃ Azure Service Connector │ 🇦 azure │ 🇦 azure-generic      │ implicit          │ ✅    │ ✅     ┃
┃                         │          │ 📦 blob-container     │ service-principal │       │        ┃
┃                         │          │ 🌀 kubernetes-cluster │ access-token      │       │        ┃
┃                         │          │ 🐳 docker-registry    │                   │       │        ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━┷━━━━━━━┷━━━━━━━━┛

$ zenml service-connector list
┏━━━━━━━━┯━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━┯━━━━━━━━┯━━━━━━━━━┯━━━━━━━━━━━━┯━━━━━━━━┓
┃ ACTIVE │ NAME        │ ID                                   │ TYPE     │ RESOURCE TYPES        │ RESOURCE NAME │ SHARED │ OWNER   │ EXPIRES IN │ LABELS ┃
┠────────┼─────────────┼──────────────────────────────────────┼──────────┼───────────────────────┼───────────────┼────────┼─────────┼────────────┼────────┨
┃   👉   │ azure-multi │ 057ba070-c3d7-436b-87bc-f6334c35b240 │ 🇦 azure │ 🇦 azure-generic      │ <multiple>    │ ➖     │ default │            │        ┃
┃        │             │                                      │          │ 📦 blob-container     │               │        │         │            │        ┃
┃        │             │                                      │          │ 🌀 kubernetes-cluster │               │        │         │            │        ┃
┃        │             │                                      │          │ 🐳 docker-registry    │               │        │         │            │        ┃
┗━━━━━━━━┷━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━┷━━━━━━━━┷━━━━━━━━━┷━━━━━━━━━━━━┷━━━━━━━━┛

$ zenml service-connector verify azure-multi
⠸ Verifying service connector 'azure-multi'...
Service connector 'azure-multi' is correctly configured with valid credentials and has access to the following resources:
┏━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃     RESOURCE TYPE     │ RESOURCE NAMES                                ┃
┠───────────────────────┼───────────────────────────────────────────────┨
┃   🇦 azure-generic    │ ZenML Subscription                            ┃
┠───────────────────────┼───────────────────────────────────────────────┨
┃   📦 blob-container   │ az://demo-zenmlartifactstore                  ┃
┠───────────────────────┼───────────────────────────────────────────────┨
┃ 🌀 kubernetes-cluster │ demo-zenml-demos/demo-zenml-terraform-cluster ┃
┠───────────────────────┼───────────────────────────────────────────────┨
┃  🐳 docker-registry   │ demozenmlcontainerregistry.azurecr.io         ┃
┗━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

## Pre-requisites

## Types of changes


AlexejPenner (approved):
Looks pretty good to me, that 1700 line file took some time to get through. Not sure I understood every smallest aspect of it so might make sense to also get a review from someone who reviewd the original connectors PR

src/zenml/integrations/kubernetes/service_connectors/kubernetes_service_connector.py
```diff
@@ -255,29 +255,44 @@ def _connect_to_resource(
...
-                with tempfile.NamedTemporaryFile(delete=False) as fp:
+                fd, temp_path = tempfile.mkstemp()
+                fp = os.fdopen(fd, "wb")
+                try:
                     fp.write(client_cert_bs)
-                    k8s_conf.cert_file = fp.name
+                    fp.flush()
+                finally:
+                    fp.close()
+                k8s_conf.cert_file = temp_path
```
fa9r, line 265: I see this logic duplicated several times below, should we move this to a separate util function perhaps? TBH I also don't fully understand what it does so having this in one place with corresponding docstrings would be helpful as well.

src/zenml/integrations/azure/service_connectors/azure_service_connector.py
```diff
@@ -0,0 +1,1778 @@
...
+the results depend on how ZenML is deployed and the environment where it is used
+and is thus not fully reproducible:
+
+- when used with the default local ZenML deployment or a local ZenML server, the
+credentials are the same as those used by the Azure CLI or extracted from local
+environment variables.
+- when connected to a ZenML server, this method only works if the ZenML server
+is deployed in Azure and will use the workload identity attached to the Azure
+resource where the ZenML server is running (e.g. an AKS cluster). The
+permissions of the managed identity may need to be adjusted to allows listing
```
fa9r, line 253:
```suggestion
permissions of the managed identity may need to be adjusted to allow listing
```

src/zenml/integrations/azure/service_connectors/azure_service_connector.py
```diff
@@ -0,0 +1,1778 @@
...
+            name="Generic Azure resource",
+            resource_type=AZURE_RESOURCE_TYPE,
+            description="""
+Multi-purpose Azure resource type. It allows Stack Components to use the
+connector to connect to any Azure service. When used by Stack Components, they
+are provided generic azure-identity credentials that can be used to create
+Azure python clients for any particular Azure service.
+
+This generic Azure resource type is meant to be used with Stack Components that
+are not represented by other, more specific resource type, like Azure blob
```
fa9r, line 319:
```suggestion
are not represented by other, more specific resource types, like Azure blob
```

src/zenml/integrations/azure/service_connectors/azure_service_connector.py
```diff
@@ -0,0 +1,1778 @@
...
+        Raises:
+            ValueError: If the provided resource ID is not a valid Azure blob
+                resource ID.
+        """
+        # The resource ID could mean different things:
+        #
+        # - Azure blob container URI: `{az|abfs}://{container-name}`
+        # - Azure blob container name: `{container-name}`
+        #
+        # We need to extract the container name from the provided resource ID
```
fa9r, line 709: include in docstrings above?

fa9r (commented):
Overall this looks good to me, However, similar to @AlexejPenner, I found it quite hard to review this in detail without having reviewed the original connectors PR, so I would suggest waiting for a review from @schustmi (in addition to docs and tests) before merging this. 

High-level I was wondering why Azure/AWS/GCP all require k8s as a dependency now. In a sense this is not ideal since I would now need to install k8s even if I just want to use, e.g., the GCP artifact store. Do you think it might make sense to decouple the connectors from the existing integrations for that reason? Or perhaps to move the GCP/Azure artifact store implementations to separate integrations similar to how we have `aws` and `s3`?

Also, I think long texts like the `description` of your new models would be much easier to consume, review, and maintain if they lived in the docs. Thus, I'd propose to turn them into docs and to replace them with short one-line descriptions that include links to the relevant pages instead.

stefannica:
> High-level I was wondering why Azure/AWS/GCP all require k8s as a dependency now. In a sense this is not ideal since I would now need to install k8s even if I just want to use, e.g., the GCP artifact store. Do you think it might make sense to decouple the connectors from the existing integrations for that reason? Or perhaps to move the GCP/Azure artifact store implementations to separate integrations similar to how we have `aws` and `s3`?

@fa9r  The kubernetes library is used in so many of our stack components that it's almost as fundamental to ZenML as the docker library. But more to the point, all these integrations are cloud provider integrations that include some form of managed Kubernetes and the connectors rely on the k8s library to generate kubernetes specific API credentials.

Decoupling connectors from integrations make it even more complicated to deal with dependencies, because the pipeline code now also needs to be aware of which connectors are linked to which components in order to track and install the necessary packages (just like we do with stack components and integrations). The compromise of including kubernetes as a dependency for all integrations that deal with cloud providers is the simplest way out of a situation that only exacerbates the dependency hell.

> 
> Also, I think long texts like the `description` of your new models would be much easier to consume, review, and maintain if they lived in the docs. Thus, I'd propose to turn them into docs and to replace them with short one-line descriptions that include links to the relevant pages instead.

I disagree. In fact, I think in a future where connectors and integrations exist separately from the ZenML core repository (i.e. Hub plugins), some form of documentation needs to live closer to the descriptors that encode what each plugin contains and how to use it. But more specifically for service connectors, the documentation embedded in the code is extensively used to provide a more interactive CLI experience and soon in more interactive dashboard experience as well.

You do have a point however that we do need to cover these service connectors in the ZenML docs and right now the real problem is that these same docs are duplicated there as well. A situation for which I don't have a solution at the moment, but maybe we can think of one as part of the future "zenml integrations as Hub plugins" movement.

fa9r (approved):
Looks good to me, I also don't have strong opinions about the general suggestions above, so happy to keep those as they are if you prefer that. 

Duplication between docs and src is a general problem we have so I don't expect that to be solved as part of this PR. Ideally, we should find a way to "import" text/code in our docs from other places but I think gitbook doesn't allow that, so nothing we can do about it right now I would say.

Thus, happy to approve, but, as I mentioned above, please wait for @schustmi's review before merging :)

src/zenml/integrations/azure/service_connectors/azure_service_connector.py
```diff
@@ -0,0 +1,1777 @@
...
+                    "storage accounts."
+                )
+
+            if self.config.resource_group:
+                # Keep only the storage accounts in the specified resource
+                # group.
+                accounts = [
+                    account
+                    for account in accounts
+                    if account.id.split("/")[4] == self.config.resource_group
```
schustmi, line 1166: I've seen this `split("/")[4]` in a few places now, should we extract it in some function in case it needs to be updated in the future?

schustmi (changes requested):
Love it, had a quite detailed look but only have some minor comments! 🎉

(Apart from unit tests, but I completely understand that mocking the entire Azure framework would be a huge effort)

src/zenml/container_registries/azure_container_registry.py
```diff
@@ -32,6 +33,25 @@ def name(self) -> str:
...
+
+        Specifies resource requirements that are used to filter the available
+        service connector types that are compatible with this flavor.
+
+        Returns:
+            Requirements for compatible service connectors, if a service
+            connector is required for this flavor.
+        """
+        return ServiceConnectorRequirements(
+            connector_type="azure",
```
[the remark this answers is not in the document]
stefannica: I did it for the `docker-registry` part, but I can't import here constants from the Azure integration. Not sure how to solve this without polluting the global constants namespace with integration specific constants, which I'd rather not do.

src/zenml/integrations/azure/service_connectors/azure_service_connector.py
```diff
@@ -0,0 +1,1777 @@
...
+                    f"configuration: {e}"
+                )
+
+            return credential, None
+
+        if auth_method == AzureAuthenticationMethods.ACCESS_TOKEN:
+            assert isinstance(cfg, AzureAccessTokenConfig)
+
+            expires_at = self.expires_at
+            assert expires_at is not None
```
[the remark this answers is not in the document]
stefannica: Yes, there are provisions in the ServiceConnector base class that validate the configuration before this function is even called: https://github.com/zenml-io/zenml/blob/main/src/zenml/models/service_connector_models.py#L193

src/zenml/container_registries/azure_container_registry.py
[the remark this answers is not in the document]
schustmi: Ah I see, because this file is still in core ZenML, makes sense

schustmi (approved):

PR #1593, opened 2023-06-05
make the `Github` repo token optional

safoinme:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/github/code_repositories/github_code_repository.py
+++ b/src/zenml/integrations/github/code_repositories/github_code_repository.py
@@ -48,7 +48,7 @@ class GitHubCodeRepositoryConfig(BaseCodeRepositoryConfig):
     owner: str
     repository: str
     host: Optional[str] = "github.com"
-    token: str = SecretField()
+    token: Optional[str] = SecretField()
 
 
 class GitHubCodeRepository(BaseCodeRepository):
@@ -70,9 +70,15 @@ class GitHubCodeRepository(BaseCodeRepository):
         Returns:
             The GitHub repository.
         """
-        return self._github_session.get_repo(
-            f"{self.config.owner}/{self.config.repository}"
-        )
+        try:
+            github_repository = self._github_session.get_repo(
+                f"{self.config.owner}/{self.config.repository}"
+            )
+        except GithubException as e:
+            raise RuntimeError(
+                f"An error occurred while getting the repository: {str(e)}"
+            )
+        return github_repository
 
     def login(
         self,
@@ -84,8 +90,13 @@ class GitHubCodeRepository(BaseCodeRepository):
         """
         try:
             self._github_session = Github(self.config.token)
-            user = self._github_session.get_user().login
-            logger.debug(f"Logged in as {user}")
+            if self.config.token:
+                user = self._github_session.get_user().login
+                logger.debug(f"Logged in as {user}")
+            else:
+                logger.info(
+                    "No token provided. Access is possible to public repositories only."
+                )
         except Exception as e:
             raise RuntimeError(f"An error occurred while logging in: {str(e)}")
 


htahir1 (commented):
Should we somehow check that in the case token is none that the repo is public or not?

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -69,10 +70,43 @@ def github_repo(self) -> Repository:
...
+
+        Raises:
+            RuntimeError: If the repository is not public.
+        """
+        url = f"https://api.github.com/repos/{owner}/{repo}"
+        response = requests.get(url)
+
+        try:
+            if response.status_code == 200:
+                response.json()
```
schustmi, line 102: Why is this necessary if we don't need the return value?

htahir1, line 102:
```suggestion
```

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -69,10 +70,43 @@ def github_repo(self) -> Repository:
...
+                logger.info(f"Repository {repo} is public")
```
schustmi, line 103: I think this isn't necessarily something that the user needs to know about right?

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -69,10 +70,43 @@ def github_repo(self) -> Repository:
...
+            else:
+                raise RuntimeError(f"Repository {repo} is not public")
```
schustmi, line 105: Maybe we can make the error message a little more clear by stating that a public repository is necessary when not providing a token?

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -69,10 +70,43 @@ def github_repo(self) -> Repository:
...
+            github_repository = self._github_session.get_repo(
+                f"{self.config.owner}/{self.config.repository}"
+            )
+        except GithubException as e:
+            raise RuntimeError(
+                f"An error occurred while getting the repository: {str(e)}"
+            )
+        return github_repository
+
+    def check_github_repo_public(self, owner, repo) -> None:
```
schustmi, line 87: Missing type annotations for the CI

schustmi (changes requested):

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -84,8 +118,16 @@ def login(
...
-            user = self._github_session.get_user().login
-            logger.debug(f"Logged in as {user}")
+            if self.config.token:
+                user = self._github_session.get_user().login
+                logger.debug(f"Logged in as {user}")
+            else:
+                logger.info(
+                    "No token provided. Access is possible to public repositories only."
+                )
+                self.check_github_repo_public(
```
htahir1, line 128: Tbh I am not sure this is the right workflow. When the traceback hits, the log "No token provided. Access is possible to public repositories only." will be lost. Better to throw an error like "It is not possible to access this repository as it does not appear to be public. Access to private repositories is only possible when a token is provided. Please provide a token and try again"

htahir1 (changes requested):
Sorry, some nits

htahir1 (approved):
I personally would not have implemented it like this now the `check_github_repo_public` is throwing an error about token which is not even part of its functionality. Rather i would have simply returned a boolean from that function and raised the error inside the `login` function. but it is too bothersome now to refactor a relatively straightforward function. Lets merge it and release

PR #1599, opened 2023-06-06
Adding the ability to tag the source of an event for the analytics

bcdurak:
## Describe changes

This PR focuses on the ability to track the source of any event in ZenML. This is achieved by introducing a new context variable called the `source_context` which can be set to `python`, `cli`, `dashboard`, `api`, or `unknown`.

- The ZenML server, analytics module, and the CLI app have been adjusted accordingly.

## Pre-requisites

## Types of changes


src/zenml/__init__.py
```diff
@@ -27,6 +27,14 @@
 
 init_logging()
 
+# Initialize source context
```
schustmi, line 30: Is this really something that needs to be in the public top level `__init__`? AFAIK this isn't a user-facing feature right?

bcdurak: Fixed it and moved it to the analytics module.

src/zenml/zen_server/zen_server_api.py
```diff
@@ -104,6 +106,38 @@ def validation_exception_handler(
...
+    It extracts the source context from the header of incoming requests
+    and applies it to the ZenML source context on the API side. This way, the
+    outgoing analytics request can append it as an additional field.
+
+    Args:
+        request: the incoming request object.
+        call_next: a function that will receive the request as a parameter and
+            pass it to the corresponding path operation.
+
+    Return:
```
bcdurak, line 122:
```suggestion
    Returns:
```

PR #1607, opened 2023-06-09
Add `zenml secret export` CLI command

fa9r:
## Describe changes
I added a `zenml secret export` CLI command that can be used to export a ZenML secret to a YAML file. The resulting YAML file can then be imported via `zenml secret create <name> -v @<filename>`.

## Pre-requisites

## Types of changes


src/zenml/cli/secret.py
```diff
@@ -1118,3 +1132,51 @@ def delete_secret(name_or_id: str, yes: bool = False) -> None:
...
+    "--filename",
+    "-f",
+    type=click.STRING,
+    default=None,
+    help=(
+        "The name of the file to export the secret to. Defaults to "
+        "<secret_name>.yaml."
+    ),
+)
+def export_secret(
```
strickvl, line 1158: was going to ask to have an `--all` flag to export all secrets, but we don't actually have that anywhere. Maybe FutureTeam could implement some sort of general export command which would export all your stacks and secrets etc all as a single file. In any case, no need to implement all that here :)

src/zenml/cli/secret.py
```diff
@@ -1118,3 +1132,51 @@ def delete_secret(name_or_id: str, yes: bool = False) -> None:
...
+        name_id_or_prefix: The name of the secret to export.
+        scope: The scope of the secret to export.
+        filename: The name of the file to export the secret to.
+    """
+    from zenml.utils.yaml_utils import write_yaml
+
+    secret = _get_secret(name_id_or_prefix=name_id_or_prefix, scope=scope)
+    if not secret.secret_values:
+        warning(f"Secret with name `{name_id_or_prefix}` is empty.")
+        return
```
strickvl, line 1178: Is this how we want to do it? If someone makes some script to backup their secrets, and we only create secret.yaml files if there are values contained inside, then potentially they lose those secrets (at least the knowledge that they had set one up etc). Would it be better practice to just output a file with an empty secret value instead?

strickvl (approved):
With a couple of small comments, but otherwise SGFMA

src/zenml/cli/secret.py
AlexejPenner: agreed

src/zenml/cli/secret.py
fa9r: Yeah agreed, ideally, you should at least be able to export your stacks together with the respective secrets somehow. Or export ALL ZenML data via `zenml export` (I wanted to build this a while ago after I added `zenml pipeline runs export ...` but noone used that command and it was a lot of code to maintain so we ditched that idea).

Anyways, a more comprehensive solution would be much more effort so, as you said, I'd keep that ticket for FutureTeam 😉

Issue #1014, opened 2022-10-26
[FEATURE]: Validate slack alerter credentials on registration


strickvl:
We currently allow the registration of the slack alerter (via the CLI) but we don’t validate that the credentials are sound at that exact moment. If someone uses the wrong credentials, they will get an error message when using the alerter (e.g. when running [the Slack Alerter example](https://github.com/zenml-io/zenml/tree/develop/examples/slack_alert)).

The experience would be cleaner if the user knew at component registration that the credentials supplied were invalid. This issue therefore includes making the relevant checks for credentials at registration time.

bhatt-priyadutt:
@strickvl at the registration time, I see we will have to get these validation checks in the stack_components.py register logic for specifically alerter - slack. below is the suspected place where this logic likely to be placed if I understood correctly. But this will be like adding logic in a common register logic for a very specific alerter integration.

https://github.com/zenml-io/zenml/blob/e055d318b326cdc305301d5130b2f2fa46d613e0/src/zenml/cli/stack_components.py#LL243C1-L243C1

PR #1609, opened 2023-06-10
Validating slack alerter

bhatt-priyadutt:
## Describe changes
Added code to validate **slack** stack component configurations at the registration time. #1014 

## Pre-requisites

## Types of changes


bhatt-priyadutt:
I have added the logic,
please review, if all good then I can add the doc and test cases, thanks

schustmi (changes requested):
There's still a missing docstring in the `SlackAlerterConfig`, otherwise looks good to me.

src/zenml/integrations/slack/flavors/slack_alerter_flavor.py
```diff
@@ -15,10 +15,16 @@
 
 from typing import TYPE_CHECKING, Optional, Type
 
+from slack_sdk import WebClient
```
schustmi, line 18: Actually I just realised another issue: ZenML imports this flavor file (and calls `StackComponentConfig.is_valid`) when the slack alerter is registered. However, this can also happen server side or you can even do it locally without having the slack integration installed. In those cases, it would now crash because the `slack_sdk` library is not installed. To avoid this, I suggest moving the import into the `is_valid` property and handling the case when it's not installed:
```python
@property
def is_valid(self) -> bool:
  try:
    from slack_sdk import WebClient
    from slack_sdk.errors import SlackApiError
  except ImportError:
    logger.warning("Unable to validate Slack alerter credentials because the Slack integration is not installed.")
    return
  
  ...  # do the actual validation here
```

schustmi (changes requested):

bhatt-priyadutt:
Thanks for the suggestions @schustmi , I have made the changes accordingly and also have updated the validation error message to cater to installation case, please review, thanks

schustmi (changes requested):

src/zenml/integrations/slack/flavors/slack_alerter_flavor.py
```diff
@@ -36,6 +39,41 @@ class SlackAlerterConfig(BaseAlerterConfig):
...
+        """
+        try:
+
+            from slack_sdk import WebClient
+            from slack_sdk.errors import SlackApiError
+        except ImportError:
+            logger.warning(
+                "Unable to validate Slack alerter credentials because the Slack integration is not installed."
+            )
+            return False
```
fa9r, line 57: Long-term I would suggest we generalize the import error handling and move it into `Client._validate_stack_component_configuration`

Doesn't necessarily need to happen as part of this PR though

src/zenml/client.py
```diff
@@ -2564,6 +2565,12 @@ def _validate_stack_component_configuration(
                 "other users. You should consider using a non-local stack "
                 "component alternative instead."
             )
+        if not configuration.is_valid:
+            raise StackComponentValidationError(
+                f"Invalid stack component configuration or uninstalled integration. please verify "
+                f"the configurations set for {component_type} "
+                f"and Please ensure that the necessary integration is installed."
```
fa9r, line 2572: If we return `True` in the import error case, this message would need to be adjusted again

fa9r (approved):
Awesome, great PR! Some of our core developers were also planning to add component-specific validation for some stack components but haven't decided on a consistent way of doing it yet. I like your approach, feels fairly clean. Perhaps we can generalize some of the code to other components in the future. Happy to get it merged once the import error default is adjusted 👍

src/zenml/integrations/slack/flavors/slack_alerter_flavor.py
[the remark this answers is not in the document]
bhatt-priyadutt: oh okay got it, I have made the changes.

src/zenml/client.py
bhatt-priyadutt: Yep, adjusted

bhatt-priyadutt:
the lint file is running without the error shown in my local, but here its giving some format error.

AlexejPenner:
@bhatt-priyadutt this should be fixable by running ./scripts/format.sh

I think it has something to do with the empty line 50 in the slack_alerter_flavor.py - removing it made the error go away on my end

src/zenml/integrations/slack/flavors/slack_alerter_flavor.py
```diff
@@ -36,6 +39,41 @@ class SlackAlerterConfig(BaseAlerterConfig):
...
+    @property
+    def is_valid(self) -> bool:
+        """Check if the stack component is valid.
+
+        Returns:
+            True if the stack component is valid, False otherwise.
+        """
+        try:
+
+            from slack_sdk import WebClient
```
AlexejPenner, line 51:
```suggestion
        try:
            from slack_sdk import WebClient
```

AlexejPenner:
Just FYI, the failure of the Mac runner is a known issue that you may ignore

bhatt-priyadutt:
Guess only the mac error is remaining, looks like these can be merged now!
@AlexejPenner

schustmi:
Thanks @bhatt-priyadutt!

PR #1611, opened 2023-06-12
Make terraform an optional dependency

schustmi:
## Describe changes
This PR removes `python-terraform` as a required dependency and makes it available as an extra instead.

## Pre-requisites

## Types of changes


AlexejPenner (approved):
LGTM - did you manually verify that the stack recipes all still work with the larger changes of moving code around? If I am not mistaken, we don't really have good test coverage of these functions.

@wjayesh I think your review is definitely needed here.

htahir1 (changes requested):
We should at least update the docs and examples in this PR that are affected and not let the exception manage it

wjayesh (approved):
Thank you for not just removing python_terraform as permanent dependency but also structuring the classes into logical files. It looks much cleaner now, LGTM!
P.S. I verified that the deploy still works.

strickvl:
@schustmi is this still planned for after mlstacks happens? I didn't remove the logic for `zenml deploy` so there's still a lot of the 'old way' still inside the codebase.

fa9r:
Closing in favor of #1855

PR #1604, opened 2023-06-08
Fix for bentoml integration/example

strickvl:
Fixes for bentoml example. Requires:

- some daemon checks to be commented out from `local_service.py` 
- as well as pinning BentoML to 1.0.10 (whereas the current version is 1.0.21).

## Pre-requisites
Please ensure you have done the following:
- [ ] I have read the **CONTRIBUTING.md** document.
- [ ] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #1614, opened 2023-06-12
Adjust BentoML Steps and Example to New Pipeline Interface

fa9r:
## Describe changes

See title. Also includes fixes from https://github.com/zenml-io/zenml/pull/1604.

## Pre-requisites

## Types of changes


src/zenml/services/local/local_service.py
safoinme, line 1: Does this have any impact on other local services?

src/zenml/services/local/local_service.py
```diff
@@ -167,20 +167,21 @@ def pid(self) -> Optional[int]:
...
+                        daemon_entrypoint.__name__ not in cmd_line[0]
+                        or config_file not in cmd_line[0]
+                    )
                 ):
-                    logger.debug(
+                    logger.warning(
                         f"Process with PID {pid} is not a ZenML local daemon "
                         f"service."
                     )
-                    return None
```
safoinme, line 183: Does this change impact other services built on top of `local_service`?

fa9r: Not sure either, I implemented it how I understood @stefannica's review comment on #1604. @stefannica do you think it makes sense how it is implemented now? Also, do you have an answer to @safoinme's question?

stefannica: This is an extra validation to ensure that the PID of a ZenML daemon process is indeed a daemon process and not some other process that happened to reuse the PID of a ZenML daemon process that was not properly cleaned up. It doesn't work in bentoml's case, so we can just make it a warning.

stefannica (approved):
LGTM

PR #1615, opened 2023-06-12
Fix Azure Integration Imports and Improve Flavor Registration Error Handling

fa9r:
## Describe changes
The Azure artifact store flavor imported the service connector which required. the Azure library.

This is very problematic since it would lead to an error during flavor registration, which would mean that subsequent integrations would be missing, e.g., `zenml model-deployer flavor list` would return an empty list even if all deployer integration requirements were installed.

To prevent integration issues like this to affect other integrations in the future, I also wrapped the flavor registration logic in our flavor registry in a try-catch block.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/azure/flavors/azure_artifact_store_flavor.py
+++ b/src/zenml/integrations/azure/flavors/azure_artifact_store_flavor.py
@@ -20,15 +20,16 @@ from zenml.artifact_stores import (
     BaseArtifactStoreFlavor,
 )
 from zenml.integrations.azure import AZURE_ARTIFACT_STORE_FLAVOR
-from zenml.integrations.azure.service_connectors.azure_service_connector import (
-    AZURE_CONNECTOR_TYPE,
-    BLOB_RESOURCE_TYPE,
-)
 from zenml.models.service_connector_models import ServiceConnectorRequirements
 from zenml.stack.authentication_mixin import AuthenticationConfigMixin
 
 if TYPE_CHECKING:
     from zenml.integrations.azure.artifact_stores import AzureArtifactStore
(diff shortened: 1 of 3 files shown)


fa9r:
[fa9r](https://github.com/fa9r) prevented a review from [schustmi](https://github.com/schustmi) now

strickvl (approved):
Looks good to me, and I'd hit this exact issue earlier in the week in a fresh environment.

bcdurak:
I guess, I was a bit too late to press the comment button 😄 In any case, some thoughts that I wanted to share anyhow:

The changes look good to me as well. However, I am confused about the implications of this issue and am unsure if this is the ultimate solution.

When we initialize the store, we always try to register all the integration flavors. From what I understood, the Azure artifact store flavor itself imported constants from the connectors module which required the Azure library to be installed, thus it failed. However, this should not only just make the flavor registration fail but cause a store initialization error as well, meaning the first time you initialize the SQLZenStore [_it is gonna fail_](https://github.com/zenml-io/zenml/blob/cdde7ea2a2af7d1406cae178ad1a2f4d8a77fb06/src/zenml/zen_stores/base_zen_store.py#L145). The confusing part is: since this is the last step of the initialization process, you were able to keep using ZenML even though the main ZenStore has failed to migrate the flavors with an error during the initialization process. 

It means, this was a relatively silent failure and these changes save other integrations but make the failing integration fail even more silently now. I think, we really need to implement some sort of testing mechanism to make sure this does not happen to our integrated flavors now.

bcdurak:
also, #unblock-@schustmi 😄

PR #1629, opened 2023-06-14
Fix output of dashboard url on pipeline run

strickvl:
The wrong (broken) URL was being output when running a pipeline. It's fixed now.

Refactored the URL printing logic based on new items being passed in. Some big assumptions there, and maybe the helper function is less useful down the line, but it's only called this one time in our codebase so I figured it was acceptable.

## Pre-requisites

## Types of changes


schustmi (approved):

fa9r (changes requested):
Good catch, but the overall logic here still doesn't really make sense in my opinion:
- we list all runs of the deployment
- we pass the run name (that we already knew) and the pipeline id (which was already in the deployment) into `print_run_url` which forwards this to `get_run_url`, 
- `get_run_url` again lists all runs for the given run name just to get the run ID
- The pipeline id is passed unnecessarily as it is also contained in the run response model

Some (mutually exclusive) ideas how we could refactor this:
- refactor the `print_run_url` and `get_run_url` functions to expect a run ID only: `print_run_url(run_id: UUID)`
- refactor the `print_run_url` and `get_run_url` functions to expect a run model only: `print_run_url(run: PipelineRunResponseModel)`
- remove the list call from `Pipeline._run` and pass `deployment.pipeline.id` instead of `runs[0].pipeline.id`

strickvl:
@fa9r @schustmi Refactored the URL printing logic based on updated item being passed in. Some big assumptions there, and maybe the helper function is less useful down the line, but it's only called this one time in our codebase so I figured it was acceptable.

src/zenml/utils/dashboard_utils.py
```diff
@@ -12,69 +12,52 @@
...
-    if pipeline_id:
-        url += f"/workspaces/{client.active_workspace.name}/pipelines/{str(pipeline_id)}/runs"
-    elif runs:
-        url += "/runs"
-    else:
-        url += "/pipelines/all-runs"
-
-    if runs:
-        url += f"/{runs[0].id}/dag"
+    url += f"/workspaces/{client.active_workspace.name}/pipelines/{pipeline_id}/runs/{run_id}/dag"
```
strickvl, line 46: Someone will generally want to have the view that ties their run back to the pipeline, I figured, so that's why I chose this URL vs the disembodied 'one run of many runs' option.

strickvl: My assumption is that if we're passing a 'run' object into the function, we'll always have a run ID and a pipeline ID available to us.

src/zenml/utils/dashboard_utils.py
```diff
@@ -12,69 +12,52 @@
...
 
     return url
```
fa9r, line 48: See above

```suggestion
    if run.pipeline:
        pipeline_id = str(run.pipeline.id)
        url += f"/workspaces/{client.active_workspace.name}/pipelines/{pipeline_id}/runs"
    else:
        url += "/runs"
    url += f"/{run_id}/dag"
    return url
```

fa9r (changes requested):
Run might not always have a pipeline, otherwise LGTM!

src/zenml/utils/dashboard_utils.py
schustmi: We thought about this when implementing the pipeline versioning and back then the consensus was to keep the unlisted runs. I guess it doesn't really hurt us to keep the feature and maybe some users don't want to "clutter" their pipeline overview page with random local experiments that never make it any further?

Don't have a strong opinion on this though, if we find a good reason for it (e.g. it would make the code much easier) I'm also open to deprecating and removing the feature at some point.

src/zenml/new/pipelines/pipeline.py
```diff
@@ -677,11 +677,7 @@ def _run(
                 )
 
                 if runs.items:
-                    # Log the dashboard URL
-                    dashboard_utils.print_run_url(
-                        run_name=deployment.run_name_template,
-                        pipeline_id=runs[0].id,
-                    )
+                    dashboard_utils.print_run_url(runs[0])
```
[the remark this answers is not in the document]
schustmi: In most cases with remote orchestrators: `runs.items` will be empty as the run hasn't started yet. I think in some old release we at least provided a dashboard link to the pipeline (if the run isn't unlisted), is that maybe something we want to re-add or is it more confusing if we provide a link which doesn't contain the run for now?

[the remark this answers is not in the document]
schustmi: I don't see any good way to implement this so I think the warning message below is fine.

fa9r (approved):

PR #1633, opened 2023-06-18
Seldon Core V2 

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


htahir1 (commented):
What a machine, over the weekend implementation!!! <3 

Amazing! I have not seen the code yet but i see two things id like to give quick feedback on:

1. Tests and docs are missing but i assume youd take that into the sprint
2. Maybe we should let @edshee take a look already for early feedback. For that, a nice example might be nice

edshee:
Awesome work @safoinme . If you create some basic docs for it I'm happy to build an end-to-end example based on the previous one (for core v1).

safoinme:
Thank you so much for the feedback @edshee, That would be awesome! right now, this branch has basically the same functionalities supported in V1. However, we are exploring the possibility of supporting more than just the `model` deployment but maybe also `Pipelines` and `Experiments`.

edshee:
If you're able to add support for pipelines that'd be absolutely awesome. Seems like there would be an ideal 1:1 mapping between steps:models and pipelines:pipelines.

Johnyz21:
Heya 😊
Really love the work that's done here! Is there a timeline for getting this change merged in as we're very eager to test out Seldon Core V2 with ZenML

PR #1647, opened 2023-06-22
Fix BentoML deployer

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


examples/bentoml_deployment/steps/predictor.py
```diff
@@ -33,7 +33,11 @@ def predictor(
     """
     service.start(timeout=10)  # should be a NOP if already started
     for img, data in inference_data.items():
-        prediction = service.predict("predict_ndarray", np.array(data))
+        grayscale_image = np.dot(
+            np.array(data)[..., :3], [0.2989, 0.5870, 0.1140]
+        )
+        grayscale_image = grayscale_image.astype(np.uint8)
+        prediction = service.predict("predict_ndarray", grayscale_image)
```
fa9r, line 40: why can't it predict on the original data directly anymore?

safoinme: Actually, I was surprised by how it was working from the start! the image we were using for inference where (28, 28, 3), and in the bentoml service validation we were asserting that it was greyscale (28, 28) and it was working somehow

safoinme:
@fa9r  I'm not so sure what exactly was blocking this, but I just went through Bentoml code and update the serve function based on what they have now (deprecating the development function and setting all these variables) and it worked after that!

examples/bentoml_deployment/steps/predictor.py
fa9r: Ah so this is turning it to greyscale? In that case, just add a docstring please so we will remember this in the future 😄 

```
       grayscale_image = np.dot(
            np.array(data)[..., :3], [0.2989, 0.5870, 0.1140]
        )
```

fa9r (approved):
Ok cool, if it works it works, let's go to main! 😄 🚀

examples/bentoml_deployment/steps/predictor.py
strickvl: @safoinme will you add this?

strickvl (approved):
LGTM! We should reach out to the people who were having problems with this in Slack once it's merged / released.

PR #1651, opened 2023-06-26
Fix the `zenml deploy` story

wjayesh:
## Describe changes
I implemented
- bug fixes on the zenml deploy recipes
- made high-level permission objects optional
- updated docs on the changes

## Pre-requisites

## Types of changes


src/zenml/cli/stack_recipes.py
```diff
@@ -473,6 +472,10 @@ def deploy(
             variables to the stack recipe.
     """
     cli_utils.warning(ALPHA_MESSAGE)
+    logger.info(
+        "No server is deployed by default anymore. "
+        "Please use `zenml deploy` to create a ZenML server."
```
strickvl, line 477:
```suggestion
        "Servers are no longer deployed by default. Please use the "
        "`zenml deploy` command to deploy a ZenML server."
```

src/zenml/cli/stack_recipes.py
```diff
@@ -615,58 +618,7 @@ def deploy(
...
-                else:
-                    from zenml.cli import server
-
-                    ctx.invoke(
-                        server.deploy,
-                        config=stack_recipe_service.get_deployment_info(),
-                        connect=True,
-                    )
-
+            logger.warning("The flag `--no-server` has been deprecated. ")
```
strickvl, line 621:
```suggestion
            logger.warning("The `--no-server` flag has been deprecated. ")
```

src/zenml/zen_server/deploy/terraform/recipes/aws/variables.tf
```diff
@@ -144,7 +139,7 @@ variable "create_ingress_controller" {
 }
 
 # if you already have an ingress controller, supply it's URL
-variable "ingress_controller_hostname" {
+variable "ingress_controller_ip" {
```
strickvl, line 142: do deletions / updates like this potentially break pre-existing stack recipe deployments for anyone? Do we need to support the old variables as well? i.e. how is this usually handled?

src/zenml/zen_server/deploy/terraform/recipes/gcp/variables.tf
```diff
@@ -125,10 +125,13 @@ variable "database_ssl_verify_server_cert" {
   type        = bool
 }
 
-variable "ingress_path" {
-  description = "The path on the Ingress URL to expose ZenML at"
-  default     = "zenml"
-  type        = string
+
+# Enable secrets manager API. Listing services might need elevated permissions.
+# Disable this if you don't have the ListServices permission.
```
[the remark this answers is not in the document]
strickvl: Given that we should be deprecating ZenML secrets managers anyway, is this even necessary for the secret store?

src/zenml/zen_server/deploy/terraform/recipes/gcp/variables.tf
```diff
@@ -125,10 +125,13 @@ variable "database_ssl_verify_server_cert" {
...
+variable "enable_secrets_manager_api" {
+  description = "Should the recipe enable the secrets manager API?"
```
strickvl, line 132:
```suggestion
  description = "Enable the secrets manager API"
```

strickvl (changes requested):
Some comments etc, but I'll also test that it works on my end as expected.

src/zenml/zen_server/deploy/terraform/recipes/aws/variables.tf
wjayesh: Usually this is handled by the CLI. If somebody pulls/deploys a stack recipe, they have a local copy of the recipe files that is used for deployment. When we push an update to the repo and the user does a `zenml stack recipe pull` they get a warning saying that an existing copy exists and it would not be overwritten so their state is safe. If you want to force update you also get a warning and a prompt that confirms that you want to update the files and this could break existing deployments.

strickvl: And what about people who just use the recipes for the raw TF scripts?

src/zenml/zen_server/deploy/terraform/recipes/gcp/variables.tf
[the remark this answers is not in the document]
wjayesh: This API is used for the new secret store in the ZenServer and not for the deprecated secrets manager. Disabling this would mean that the project that the user is deploying to, should already have this API enabled (probably by someone who has those permissions). 
Keeping it optional is important because not everybody running this recipe has elevated permissions.

src/zenml/zen_server/deploy/terraform/recipes/aws/variables.tf
wjayesh: It could potentially break for people using the raw TF files if they pull a new version of the recipes to their local. It is unlikely to cause big problems because most of the time, terraform will naturally just update the deployment to match the new configuration but we don't offer guarantees.

strickvl: I guess what I'm getting at: can we deprecate these somehow? Does that even make sense?

wjayesh: Ah, I get it, yes we can deprecate this if we keep the old variable intact. However, I feel that keeping both options would add to the confusion in an already complex story. Plus, on the technical side, we don't want people to continue using the hostname since it might interfere with the ingress set up through the stack recipes (assuming that the user used them to create the stack components).

strickvl (approved):
Approved pending someone testing that it actually works as hoped.

PR #1653, opened 2023-06-27
New step output annotations

schustmi:
## Describe changes
This PR implements a new way to annotate the outputs of ZenML steps using `Tuple` and `Annotated` annotations:
```python
from typing import Tuple, Annotated

# Single output step with default output name 'output'
@step
def single_output_step() -> int:

# Single output step with custom output name
@step
def single_output_step() -> Annotated[int, "my_output"]:

# Multi output step with default output names 'output_0' and 'output_1'
@step
def multi_output_step() -> Tuple[int, str]:

# Multi output step with custom output names 'my_output' and 'output_1'
@step
def multi_output_step() -> Tuple[Annotated[int, "my_output"], str]:
```

Additionally, we now allow steps without return annotations and dynamically load materializers at runtime in that case.

## Pre-requisites

## Types of changes


src/zenml/steps/utils.py
```diff
@@ -38,57 +131,183 @@ def resolve_type_annotation(obj: Any) -> Any:
...
-def get_args(obj: Any) -> Tuple[Any, ...]:
-    """Get arguments of a Union type annotation.
+    return origin, None
+
+
+def validate_annotation_metadata(annotation: Any) -> str:
+    """Validates annotation metadata.
+
+    Args:
+        annotation: The type annotation, must be of type `Annotated[...]`
```
fa9r, line 156: This first part of the docstring could be clearer on what the function does

fa9r: Maybe also rename to `get_output_name_from_annotation_metadata()`?

src/zenml/steps/utils.py
```diff
@@ -38,57 +131,183 @@ def resolve_type_annotation(obj: Any) -> Any:
...
+        raise ValueError(
+            "Annotation metadata must be a string which will be used as the "
+            "output name."
+        )
+
+    return output_name
+
+
+class ReturnVisitor(ast.NodeVisitor):
+    """AST visitor class that can be subclasses to visit function returns."""
```
fa9r, line 185:
```suggestion
    """AST visitor class that can be subclassed to visit function returns."""
```

src/zenml/steps/utils.py
```diff
@@ -38,57 +131,183 @@ def resolve_type_annotation(obj: Any) -> Any:
...
 
 
-def parse_return_type_annotations(return_annotation: Any) -> Dict[str, Any]:
-    """Parse the returns of a step function into a dict of resolved types.
+def has_only_none_returns(func: Callable[..., Any]) -> bool:
+    """Checks whether a function contains only `None` returns.
 
-    Called within `BaseStepMeta.__new__()` to define `cls.OUTPUT_SIGNATURE`.
+    A `None` return could be either an explicit `return None` or an empty
+    `return` statement.
```
fa9r, line 299: Maybe also here an example like above could make this clearer

src/zenml/steps/utils.py
```diff
@@ -38,57 +131,183 @@ def resolve_type_annotation(obj: Any) -> Any:
...
+    if not isinstance(output_name, str):
+        raise ValueError(
+            "Annotation metadata must be a string which will be used as the "
+            "output name."
+        )
+
+    return output_name
+
+
+class ReturnVisitor(ast.NodeVisitor):
```
fa9r, line 184: 😮 👍 🎉

fa9r (changes requested):
Nice!!! 🔥 🔥 🔥

The only thing I don't fully understand here is why we need to write the default materializer into the step config now, that feels like a lot of useless info to me. Otherwise not much to add, the implementation seems clean, docs look great, everything is well tested, ... Let's get it merged 😄 👍

src/zenml/config/step_configurations.py
```diff
@@ -44,6 +44,7 @@ class PartialArtifactConfiguration(StrictBaseModel):
     """Class representing a partial input/output artifact configuration."""
 
     materializer_source: Optional[Tuple[Source, ...]] = None
+    default_materializer_source: Optional[Source] = None
```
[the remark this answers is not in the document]
schustmi: If the type annotation is Any, we now set the `default_materializer_source` to the default materializer of the registry (usually cloudpickle, but can be overwritten by users). Now when executing a step and it returns an `int`, we do the following:
- Check if an explicit materializer should be used (this is the case if the step has a type annotation or a materializer was manually configured)
- If not, query the materializer registry to find a materializer at runtime (this will only work if a materializer was registered for the type)
- If not, use the materializer defined by `default_materializer_source`

The second step would never happen if we were to store the default materializer as part of the `materializer_source` attribute. I hope that explains it?

tests/unit/steps/test_utils.py
```diff
@@ -11,18 +11,119 @@
...
+
+
+def func_with_no_output_annoation_and_no_return():
+    return
+    return None
+
+
+def func_with_no_output_annoation_and_return():
+    return 1
+    return
```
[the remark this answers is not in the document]
schustmi: This is just for the AST parsing which doesn't care about that

src/zenml/steps/utils.py
schustmi: Renamed it

src/zenml/steps/utils.py
schustmi: Added one

docs/book/user-guide/starter-guide/fetch-runs-after-execution.md
```diff
@@ -245,18 +245,8 @@ Similarly, you can use the `inputs` and `input` properties to get the input
...
-```python
-from zenml import step
-from zenml.steps import Output
-
-@step
-def some_step(input_name: int) -> Output(output_name=int):
-    ...
-```
+Check out [this page](../advanced-guide/configure-steps-pipelines.md#step-output-names)
+for more information on default step output names and how to define custom names for your step outputs.
```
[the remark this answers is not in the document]
schustmi: I'd rather not have all the info duplicated again, how would you want me to tell them to check out the other page? Isn't that exactly what the referenced line is doing? :D

docs/book/user-guide/advanced-guide/configure-steps-pipelines.md
```diff
@@ -2,6 +2,73 @@
...
+def square_root(number: int) -> Annotated[float, "custom_output_name"]:
+    return number ** 0.5
+
+@step
+def divide(a: int, b: int) -> Tuple[
+    Annotated[int, "quotient"],
+    Annotated[int, "remainder"]
+]:
+    return a // b, a % b
+```
```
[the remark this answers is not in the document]
schustmi: Why would that be helpful in this context? If someone is looking at this, they're looking for information on how to define a step. When they want to fetch information later on, they'll head to the corresponding page and find the line of code relevant to what they want to do (+ a link to this section in case they want to change the output name).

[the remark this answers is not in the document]
schustmi: Discussion: the link above is enough

docs/book/user-guide/starter-guide/fetch-runs-after-execution.md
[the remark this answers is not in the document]
schustmi: Updated

fa9r (approved):
SGFMA 🤓

fa9r (approved):
SINGFMA

PR #1654, opened 2023-06-27
fix: use k8s orchestrator service account in step pod's manifest

francoisserra:
## Describe changes
I implemented/fixed the Kubernetes orchestrator so that the entry point will use the service account name defined in the settings when building the Pod Spec.
This fixes the main orchestrator Pod is created with the service account name in its Spec, but when it spins steps Pods, it doesn't include it in their Specs.
As a consequence, it is now possible to leverage k8s<->Cloud provider service account bindings (like GKE Workload Identity) in the steps  to access cloud resources (buckets, ...)

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
@@ -117,6 +117,7 @@ def main() -> None:
             args=step_args,
             env=env,
             settings=settings,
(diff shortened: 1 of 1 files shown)


fa9r (approved):
Great suggestion, I don't see any reason to not do that. 

I'll do a few more manual tests before merging but I think this looks good!

schustmi (approved):

fa9r:
Manual tests look good, let's get this merged! Thanks again for the PR @francoisserra 🙏

PR #1665, opened 2023-06-30
Fix/set env variables before installing packages

lopezco:
## Describe changes
While building a Docker file, I  need to set the environment variables before installing the packages as some of the ENV vars change the way the packages are installed.

## Pre-requisites
- {x] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).

## Types of changes

### DIFF
--- a/src/zenml/utils/pipeline_docker_image_builder.py
+++ b/src/zenml/utils/pipeline_docker_image_builder.py
@@ -606,6 +606,17 @@ class PipelineDockerImageBuilder:
         """
         lines = [f"FROM {parent_image}", f"WORKDIR {DOCKER_IMAGE_WORKDIR}"]
 
+        lines.append(f"ENV {ENV_ZENML_ENABLE_REPO_INIT_WARNINGS}=False")
+        if download_files:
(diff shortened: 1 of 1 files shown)


schustmi (changes requested):
Changes look good to me, you'd just need to rebase on develop and adjust the target branch of this PR so we can merge it!

lopezco:
@schustmi It is ok now?

schustmi (changes requested):
Yes looks good now. I thought of one more change that we could make to this:
I would suggest to keep the following lines
```python
if download_files:
  lines.append(f"ENV {ENV_ZENML_REQUIRES_CODE_DOWNLOAD}=True")
```
below the apt/pip installs. The reason for this (and also why we set the environment variables as late as possible in the first place) is to improve the Docker caching for repeated builds.

lopezco:
@schustmi that is true, I just pushed the changes to leave only user-set environment variables before apt/pip installs

schustmi (approved):
Perfect, thanks!

fa9r (approved):

PR #1666, opened 2023-06-30
Integrate core azureml orchestrator logic

bhatt-priyadutt:
## Describe changes
Creating PR to see the changes

## Pre-requisites

## Types of changes


examples/azure_ml_orchestrator_flavor/orchestrator/azure_ml_pipelines_orchestrator.py
```diff
@@ -73,356 +69,337 @@ def settings_class(self) -> Optional[Type["BaseSettings"]]:
...
+        # each step can have its own environment, or they can all use the same one
+        # Here you specify your dependencies etc.
+        # This can be a Docker image.
+        custom_env_name = "zenml-test-env"
+        source_directory = get_source_root()
+        orchestrator_run_name = get_orchestrator_run_name(
+            pipeline_name=deployment.pipeline.name
+        ).replace("_", "-")
+        docker_image_builder = PipelineDockerImageBuilder()
+        image_name_or_digest,_,_ = docker_image_builder.build_docker_image(docker_settings=deployment.pipeline_configuration.docker_settings,
```
schustmi, line 356: Instead of having to manually build the image here, your orchestrator already inherits from `ContainerizedOrchestrator` which takes care of building this image. To get the image digest, you can simply call `self.get_image(...)`. See the other orchestrators for reference: https://github.com/zenml-io/zenml/blob/58cfdbd857cde809aa32f0bc24964101a1ac0175/src/zenml/integrations/kubeflow/orchestrators/kubeflow_orchestrator.py#L539C1-L541

examples/azure_ml_orchestrator_flavor/orchestrator/azure_ml_pipelines_orchestrator.py
```diff
@@ -73,356 +69,337 @@ def settings_class(self) -> Optional[Type["BaseSettings"]]:
...
+            The orchestrator run id.
 
         Raises:
-            RuntimeError: If you try to run the pipelines in a notebook environment.
+            RuntimeError: If the run id cannot be read from the environment.
         """
-        # First check whether the code running in a notebook
-        if Environment.in_notebook():
+        try:
+            return os.environ[ENV_ZENML_AZUREML_RUN_ID]
```
schustmi, line 85: For your pipeline to actually run on AzureML, you'd need to make sure this environment variable is set.

Best would be to use an environment variable that already exists and is getting set by AzureML itself, e.g. if they have a unique ID for the pipeline run that is happening on their side. The value that you return from this method needs to be
- different for each pipeline run
- same for all steps of a single pipeline run

Maybe it's also possible to use the `Run` class (https://learn.microsoft.com/en-us/python/api/azureml-core/azureml.core.run%28class%29?view=azure-ml-py) of the AzureML SDK to get the current run. Note that the `get_run_id()` method will only be called inside the Docker image running on AzureML, not on your local machine.

schustmi (changes requested):
I think I might have found the issues with your flavor registration:
When you're trying to run `zenml orchestrator flavor register ...`, ZenML will verify the following things:
- That it's able to import the flavor class. This should work in your case, because you're specifying `examples.azure_ml_orchestrator.orchestrator. ...`
- That it's able to import the `implementation_class`: This is where I think it fails. If you take a look at the `implementation_class` property of your `AzureMLPipelinesOrchestratorFlavor`, it tries to import the orchestrator class like this: `from orchestrator.azure_ml_pipelines_orchestrator import AzureMLPipelinesOrchestrator`. This is missing the `examples.azure_ml_orchestrator.` prefix which causes the import to fail.

After additional checks: There are multiple imports in the code which are relative to the `examples/azure_ml_orchestrator` directory. Therefore the easiest solution is to
- change your working directory to `zenml/examples/azure_ml_orchestrator_flavor`
- Run `zenml init`
- Register the flavor like this: `zenml orchestrator flavor register orchestrator.azure_ml_pipelines_orchestrator_flavor.AzureMLPipelinesOrchestratorFlavor`

strickvl:
Closing this in favor of https://github.com/zenml-io/zenml/pull/1743 (which I believe supersedes it)

PR #1680, opened 2023-07-07
Improve yaml config docs page

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/user-guide/advanced-guide/configure-steps-pipelines.md
+++ b/docs/book/user-guide/advanced-guide/configure-steps-pipelines.md
@@ -396,13 +396,11 @@ steps:
   ...
 ```
 
-You can also use the following method to generate a config template for your pipelines:
+You can also use the following method to generate a config template (at path `/local/path/to/config.yaml`) that
+includes all configuration options for this specific pipeline and your active stack:
 ```python
 my_pipeline.write_run_configuration_template(path='/local/path/to/config.yaml')
 ```
-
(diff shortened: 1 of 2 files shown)


docs/book/user-guide/advanced-guide/configure-steps-pipelines.md
```diff
@@ -387,20 +389,20 @@ run_name: my_run
...
     settings: { }  # same as step settings
-  name_of_step_2:
+  other_step_invocation_id:
     settings: { }
   ...
 ```
 
-ZenML provides a convenient method that takes a pipeline instance and generates a config template based on its settings automatically:
-
+You can also use the following method to generate a config template for your pipelines:
```
fa9r, line 399: Could be a bit clearer that the resulting template is custom-tailored to this pipeline

fa9r: Also, does the template already include the currently configured options as defaults somehow? (not sure if it should, but if it does then maybe mention that)

fa9r (approved):

docs/book/user-guide/advanced-guide/configure-steps-pipelines.md
schustmi: Good point, I'll mention that this is for the specific pipeline/stack combination!

Regarding the second point: No, but the existing configuration in code (which would also exist the next time the user uses that pipeline) will be merged with the config file.

PR #1686, opened 2023-07-12
Misc slack fixes

schustmi:
## Describe changes

- Change order in resolving sources to account for virtual environments created inside the ZenML source root
- Improve error message when passing wrong parameter key

## Pre-requisites

## Types of changes


AlexejPenner (approved):
:seal:

src/zenml/utils/source_utils.py
```diff
@@ -308,13 +308,9 @@ def is_distribution_package_file(file_path: str, module_name: str) -> bool:
         if Path(path).resolve() in absolute_file_path.parents:
             return True
 
-    if _get_package_for_module(module_name=module_name):
-        return True
-
-    # TODO: Both of the previous checks don't detect editable installs because
+    # TODO: The previous checks does not detect editable installs because
```
fa9r, line 311: "check does" or "check do" 😛 

```suggestion
    # TODO: The previous check does not detect editable installs because
```

src/zenml/utils/source_utils.py
```diff
@@ -308,13 +308,9 @@ def is_distribution_package_file(file_path: str, module_name: str) -> bool:
         if Path(path).resolve() in absolute_file_path.parents:
             return True
 
-    if _get_package_for_module(module_name=module_name):
-        return True
```
fa9r, line 312: What does this change do exactly?

schustmi: I'm not exactly sure why this was even here in the first place to be honest:
It checks for third-party packages that are installed outside of the site packages directory, which would be the case for editable installs. However as the comment below states, they aren't detected by this anyway so probably a leftover when trying to get it to work.

In some cases (e.g. our failing tests), this lead to detecting our test files (in modules `tests.unit...`) as third-party modules, because one of the packages accidently installed their tests and registered the `tests` prefix as an installed package.

fa9r (approved):
SGFMA 🚀

PR #1689, opened 2023-07-13
Make the workspace statistics endpoint more performant

AlexejPenner:
## Describe changes
I implemented a count method for the sql_zen_store for stacks, components, pipelines and runs to make the call to `.../workspaces/<iD>/statistics` more performant.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/rest_zen_store.py
+++ b/src/zenml/zen_stores/rest_zen_store.py
@@ -1309,6 +1309,7 @@ class RestZenStore(BaseZenStore):
         Returns:
             The number of pipelines in the workspace.
         """
+        raise NotImplementedError
 
     @track(AnalyticsEvent.UPDATE_PIPELINE)
     def update_pipeline(
--- a/src/zenml/zen_stores/rest_zen_store.py
+++ b/src/zenml/zen_stores/rest_zen_store.py
@@ -495,6 +495,19 @@ class RestZenStore(BaseZenStore):
             filter_model=stack_filter_model,
         )
 
+    def count_stacks(
+        self, workspace_id: Optional[UUID]
+    ) -> int:
+        """List all stacks matching the given filter criteria.
+
+        Args:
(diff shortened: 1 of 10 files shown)


src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -1272,6 +1298,18 @@ def list_pipelines(
...
+        self, workspace_id: Optional[UUID]
+    ) -> int:
+        """Count all pipelines, optionally within a workspace scope.
+
+        Args:
+            workspace_id: The workspace to use for counting pipelines
+
+        Returns:
+            The number of pipelines in the workspace.
+        """
```
strickvl, line 1311:
```suggestion
        """
        raise NotImplementedError
```

maybe?

fa9r (commented):
Implementation LGTM but I have a few general questions/suggestions:

1. Why do we need these count methods in the first place? The `Page` returned by `list_X()` already contains a `total` field which is obtained using basically the same query. If these count methods are not intended as public API, couldn't we just do `list_X(size=0).total` to get the same result with (more or less) similar runtime for the dashboard?

2. All count methods on the `RestZenStore` currently raise `NotImplementedError`. I guess this is because these methods are not intended as public API so we don't want to add API endpoints for them. In that case, I think these methods should not even be part of the ZenStore interface. The underlying issue here is that the `zen_store` global in the server interface is a `BaseZenStore` type even though it cannot even be initialized as non-sql. If we change the typing there to `def zen_store() -> SqlZenStore` then we don't need to implement the count methods in the interface / rest stores and can get rid of that dead boilerplate code.

3. The implementation of the count methods is basically always the same, just with a different schema, so there's a lot of duplicate code right now that we could get rid of by moving the implementation to some util method `def _count_entity(self, schema, workspace_id)`.

fa9r (approved):
LGTM, what a beautiful implementation 👏 🥸

schustmi (approved):

PR #1694, opened 2023-07-17
fix: :card_file_box: Extend pipeline spec storage length

francoisserra:
Increase pipeline spec DB field to support length up to  2**24 bytes (MEDIUMTEXT in MySQL)

## Describe changes
I fixed the `Pipeline` schema to extend the `spec` field storage length.
It turns out that when having deep complex pipeline parameters (e.g. `dict` that contains `list[str]`), the JSON serialization of these parameters can quickly reach the actual DB storage limit (`TEXT` datatype in MySQL allows 2^16=65536 chars).
(In our case we hit this limit by specifying the list of input features that are numeric/categorical as pipeline parameters, and we have ~300 long features names) 

You can see this PR as a sister of #1503 which extended the `Step_run` `parameters` and `configuration` fields.

## Pre-requisites

## Types of changes


schustmi (approved):
Thanks!

fa9r (approved):

PR #1652, opened 2023-06-26
Remove Python 3.7 Support

fa9r:
## Describe changes
Python 3.7 reached its end of life on June 27th, 2023.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [ ] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Other (add details above)

PR #1695, opened 2023-07-17
Fix Python 3.10 PyYAML Installation Issues

fa9r:
## Background
- PyYAML 5 no longer works with the new Cython 3, resulting in broken ZenML installations under Python 3.10 https://github.com/yaml/pyyaml/issues/601
- PyYAML is apparently not going to release a fix for this: https://github.com/yaml/pyyaml/pull/726

## Describe changes
- Bumped `pyyaml` to `>=6.0.1`
- Bumped `evidently` version to `[0.2.7-0.2.8]` which are the only versions that support both the old Evidently API which our integration code is using as well as PyYAML 6.
- Pinned `kfp` to `1.8.22` which is the only `1.x` version that supports PyYAML 6.
- Pinned `kfk-tekton` to `1.7.1` which is the only version compatible with `kfp==1.8.22`
- Removed `docker-compose` as dev dependency since all available versions require PyYAML 5.  As a workaround, I adjusted the corresponding CI action to lazy-install the package instead.
- Removed Python 3.7 CI checks since the new Tekton version is no longer compatible with Python 3.7 -> let's also merge #1652 before the next release

## Pre-requisites

## Types of changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ pyparsing = "<3,>=2.4.0"
 python = ">=3.7.1,<3.11"
 python-dateutil = "^2.8.1"
 python-terraform = { version = "^0.10.1" }
-pyyaml = ">5.4.1"
+pyyaml = ">=6.0.1"
 rich = {extras = ["jupyter"], version = "^12.0.0"}
 sqlalchemy_utils = "0.38.3"
 sqlmodel = "~0.0.8"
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -160,8 +160,6 @@ types-setuptools = { version = "^57.4.2", optional = true }
 types-six = { version = "^1.16.2", optional = true }
 types-termcolor = { version = "^1.1.2", optional = true }
 types-psutil = { version = "^5.8.13", optional = true }
-# test framework dependencies
```
stefannica, line 163: Removing this will break the CI

fa9r: I adjusted the integration test GH action to lazily install `docker-compose` in non-default environments now. I think that should work since the CI only runs non-default environment tests under Python 3.8 where PyYAML shouldn't be broken in the first place.

schustmi (approved):

stefannica (approved):
Amazing work figuring out all those PyYAML version dependencies !

PR #1703, opened 2023-07-20
catch exceptions in `list_model_versions`

avishniakov:
## Describe changes
I fixed an issue with MlFlow Model Registry operating with bad underlying data to achieve better stability in the long run. Corrupted versions in MlFlow are not usable anyway, so it is safe to filter them out.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/mlflow/model_registries/mlflow_model_registry.py
+++ b/src/zenml/integrations/mlflow/model_registries/mlflow_model_registry.py
@@ -616,11 +616,13 @@ class MLFlowModelRegistry(BaseModelRegistry):
                         mlflow_model_version=mlflow_model_version,
                     )
                 )
-            except (AttributeError, OSError):
-                # Sometimes, the Model Registry in MlFlow can become unusable
+            except (AttributeError, OSError) as e:
+                # Sometimes, the Model Registry in MLflow can become unusable
                 # due to failed version registration or misuse. In such rare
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/mlflow/model_registries/mlflow_model_registry.py
```diff
@@ -608,12 +608,20 @@ def list_model_versions(
...
+                model_versions.append(
+                    self._cast_mlflow_version_to_model_version(
+                        mlflow_model_version=mlflow_model_version,
+                    )
+                )
+            except (AttributeError, OSError):
+                # Sometimes, the Model Registry in MlFlow can become unusable
+                # due to failed version registration or misuse. In such rare
+                # cases, it's best to suppress those versions that are not usable.
+                pass
```
fa9r, line 623: silent error handling is a bit dangerous, probably we should log some warning here at least

```suggestion
            except (AttributeError, OSError) as e:
                # Sometimes, the Model Registry in MLflow can become unusable
                # due to failed version registration or misuse. In such rare
                # cases, it's best to suppress those versions that are not usable.
                logger.warning(
                    f"Error encountered while loading MLflow model version: {e}"
                )
```

fa9r (approved):
Great catch, would suggest raising a warning in those cases, otherwise LGTM!

schustmi (approved):

stefannica (approved):

PR #1549, opened 2023-05-22
Decouple MLflow Components

fa9r:
## Describe changes
This PR decouples the MLflow deployer and model registry from the experiment tracker so they can be used without any models having been logged beforehand.

### Example
```python
@step
def digits_data_loader():
    ...

@step  # No experiment tracker used here
def svc_trainer(X_train: np.ndarray, y_train: np.ndarray) -> ClassifierMixin:
    model = SVC(gamma=0.001)
    model.fit(X_train, y_train)
    return model

from zenml.integrations.mlflow.steps.mlflow_deployer import mlflow_model_deployer_step

@pipeline
def first_pipeline():
    X_train, X_test, y_train, y_test = digits_data_loader()
    model = svc_trainer(X_train, y_train)
    mlflow_model_deployer_step(model)
```

### Implementation Details
- Both `mlflow_model_deployer_step` and `mlflow_register_model_step` now log the model to MLflow in a new run if the model was not found; both steps were also heavily refactored
- Instead of getting the MLflow connection from the experiment tracker in the stack, the model registry and deployer can now be configured separately
- To achieve this, `MLFlowStackComponentMixin` and `MLFlowConfigMixin` classes were added which all three MLflow components now inherit from

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [x] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #1708, opened 2023-07-23
bump `MLflow` to 2.5.0

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


fa9r (approved):

src/zenml/integrations/mlflow/__init__.py
```diff
@@ -35,9 +35,9 @@ class MlflowIntegration(Integration):
     # does not pin it. They fixed this in a later version, so we can probably
     # remove this once we update the mlflow version.
     REQUIREMENTS = [
-        "mlflow>=1.24.0,<=2.2.2",
-        "mlserver>=0.5.3",
-        "mlserver-mlflow>=0.5.3",
+        "mlflow>=2.1.1,<=2.5.0",
```
avishniakov, line 38: Why do we go for `2.1.1` here? What makes `2.1.1` better than `2.0.1`?
If limiting the lower boundary above the earliest `2x` version, I'd go for `2.2.1` as it was delivering important fixes.

safoinme: The reason for `2.1.1` is a bug fixed in that version related to the model registry so the previous version would still have that bug.

avishniakov: For my understanding - which [one of those](https://github.com/mlflow/mlflow/releases/tag/v2.1.1) is critical for our case?

safoinme: I think it's more of a feature within mlflow rather than a bug but for the zenml model registry, it was detected as a bug. 
At the development of the model registry stack component, it was designed so that it uses the same backend store as the experiment tracker, in a local case our mlflow experiment tracker uses the file store [this one](https://github.com/mlflow/mlflow/blob/branch-2.1/mlflow/store/tracking/file_store.py) However before version `2.1.x` of MLflow there was no implementation of a backend file store for the registry there were only [rest_store](https://github.com/mlflow/mlflow/blob/branch-2.1/mlflow/store/model_registry/rest_store.py) and [sqlalchemy_store](https://github.com/mlflow/mlflow/blob/branch-2.1/mlflow/store/model_registry/rest_store.py). The absence of a file_store was causing an error because within zenml we were trying to configure both experiment tracker and model registry to use the same store and save all artifact in the same path, however starting from the version 2.1 the file_store was introduced to the model registry which you can find [here](https://github.com/mlflow/mlflow/blob/branch-2.1/mlflow/store/model_registry/file_store.py) which was also mentioned in the release notes from [here](https://github.com/mlflow/mlflow/releases/tag/v2.1.0). With how coupled experiment tracking and registry components are (it's not set to change, you can see an open PR [here](https://github.com/zenml-io/zenml/pull/1549)) it was a challenging task to make the registry support an older version that doesn't have a file_store so for that we just limited the version to 2.1 only for when the model registry is used which you can find from [here](https://github.com/zenml-io/zenml/blob/main/src/zenml/integrations/mlflow/model_registries/mlflow_model_registry.py#L110)

safoinme: Technically we can still have the validation we do for verifying if a local model registry is used and the version installed is earlier than `2.1.x` and have a range of supported mlflow versions starting from`2.0` @avishniakov do you think that would be better? or we just move the `2.2.1` which has even more important fixes as you mentioned?

avishniakov: I see, thanks for the clarifications!
My suggestion would be to go with `2.1.1` as a minimum then, but raise a warning in case `<2.2.1` is used, so users are better aware of security concerns it might bring. Sounds reasonable?

avishniakov (approved):
Lgtm, see comments

stefannica (approved):

PR #1707, opened 2023-07-21
Rename `transition_model_stage` to `transition_model_version_stage`

avishniakov:
## Describe changes
I fixed the typo in the MlFlow client method name.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [ ] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #1710, opened 2023-07-24
E2E flow example for templates

avishniakov:
## Describe changes
This PR contains an e2e flow example to be used as the basis for the template project later on. It is 2 pipelines + steps. It has a training and predictions pipeline, which exchanges the states (we rely on certain Artifacts from trains in predict).

[README](https://github.com/zenml-io/zenml/blob/feature/FRW-2134-add-e2e-example/examples/e2e/README.md)

For better readability, this is the current flow outlook:
Train pipeline:
image
Batch predictions pipeline:
image

It basically has an ETL phase, HP tunning, Train phase, promotion, inference, a bit of DQ, and Slack alerting on top of everything. It can be a representation of a weekly/daily batch prediction with retrain flow.

It still needs a polish and shouldn't be considered as a final state of thinking.

This PR depends on the following:
- https://github.com/zenml-io/zenml/pull/1707
- https://github.com/zenml-io/zenml/pull/1711
- https://github.com/zenml-io/zenml/pull/1715
- https://github.com/zenml-io/zenml/pull/1703

## Pre-requisites

## Types of changes


examples/e2e/config.py
```diff
@@ -0,0 +1,81 @@
+#  Copyright (c) ZenML GmbH 2023. All Rights Reserved.
```
fa9r, line 1: This is very interesting, I think some time ago we were already discussing whether/how to make configuration management part of ZenML itself. Might also solve/help with the hyperparam tuning issue. Perhaps we could revisit this now that you're on the team!

examples/e2e/pipelines/batch_inference.py
```diff
@@ -0,0 +1,90 @@
...
+from zenml.logger import get_logger
+from zenml.steps.external_artifact import ExternalArtifact
+
+logger = get_logger(__name__)
+
+
+@pipeline(
+    settings={"docker": PipelinesConfig.docker_settings},
+    on_success=notify_on_success,
+    on_failure=notify_on_failure,
```
fa9r, line 39: If I remember right, you mentioned this is not working? I believe it should, but if not, you could just do `notify_on_failure` for each step and have `notify_on_success` as a separate pipeline step that runs at the very end of the pipeline

examples/e2e/steps/training/model_evaluator.py
```diff
@@ -0,0 +1,113 @@
...
+    if fail_on_accuracy_quality_gates and messages:
+        raise RuntimeError(
+            "Model performance did not meet the minimum criteria:\n"
+            + "\n".join(messages)
+        )
+    else:
+        for message in messages:
+            logger.warning(message)
+
+    mlflow.log_metric("testing_accuracy_score", tst_acc)
```
fa9r, line 111: You might want to move this up, otherwise the metric is not logged if it's too low

examples/e2e/steps/training/model_hp_tuning.py
```diff
@@ -0,0 +1,103 @@
...
+from sklearn.model_selection import RandomizedSearchCV
+
+from zenml import step
+from zenml.logger import get_logger
+
+logger = get_logger(__name__)
+
+
+@step
+def model_hp_tuning(
```
fa9r, line 30: Yeah doing the entire tuning in a single step is the only way to do that right now I think. Maybe enabling step calls from within other steps could alleviate this issue since you could then just call your `trainer` step below with different configs from within here directly.

examples/e2e/utils/artifacts.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    pipeline = Client().get_pipeline(pipeline_name)
+    latest_run = pipeline.runs[0]
+    artifacts = [a for a in latest_run.artifacts if a.name == artifact_name]
+    if artifacts:
+        return artifacts[0].id
+    else:
+        raise ValueError(
+            f"`{artifact_name}` not found in last run of `{pipeline_name}`"
+        )
```
fa9r, line 49: Maybe this is slightly more efficient/readable

```suggestion
    for artifact in latest_run.artifacts:
        if artifact.name == artifact_name:
            return artifact.id

    raise ValueError(
        f"`{artifact_name}` not found in last run of pipeline `{pipeline_name}`."
    )
```

examples/e2e/steps/etl/data_loader.py
```diff
@@ -0,0 +1,59 @@
...
+from zenml.client import Client
+from zenml.logger import get_logger
+
+logger = get_logger(__name__)
+
+artifact_store = Client().active_stack.artifact_store
+
+
+@step
+def data_loader(
```
fa9r, line 31: For the final example, I think using an actual ML dataset might be more interesting than the random dataset you're using here. E.g., one of the maybe: https://scikit-learn.org/stable/datasets/toy_dataset.html

examples/e2e/steps/alerts/slack_notifications.py
```diff
@@ -0,0 +1,53 @@
...
+        str: Prepared message.
+    """
+    step_context = get_step_context()
+    run_url = get_run_url(step_context.pipeline_run)
+
+    return (
+        f"Pipeline `{step_context.pipeline_name}` [{str(step_context.pipeline.id)}] {status}!\n"
+        f"Run `{step_context.pipeline_run.name}` [{str(step_context.pipeline_run.id)}]\n"
+        f"URL: {run_url}"
+    )
```
fa9r, line 41: I think it might make sense to add this into the Slack alerter by default. We already have a `include_format_blocks` option, which, when set, prettifies the message you send via markdown and that already includes the pipeline and step names. Adding run and run URL might make sense there IMO

examples/e2e/pipelines/batch_inference.py
```diff
@@ -0,0 +1,90 @@
...
+        n_samples=10_000,
+        drop_target=True,
+    )
+    preprocess_pipeline_id = find_artifact_id(
+        pipeline_name=MetaConfig.pipeline_name_training,
+        artifact_name="preprocess_pipeline",
+    )
+    dataset_inf = inference_data_preprocessor(
+        dataset_inf=dataset_inf,
+        preprocess_pipeline=ExternalArtifact(id=preprocess_pipeline_id),
```
fa9r, line 62: Another interesting use I haven't seen before. Very cool if this works since it doesn't redundantly save the artifact again. If not, you could try wrapping `find_artifact_id` in a `step` instead which does `artifact.load()` to load and returns the artifact again (but, as I said, that would be less efficient since the artifact would be redundantly saved again then).

examples/e2e/steps/inference/inference_predict.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        dataset_inf: The inference dataset.
+        model_version: Model Version in Model Registry.
+
+    Returns:
+        The processed dataframe: dataset_inf.
+    """
+    ### ADD YOUR OWN CODE HERE - THIS IS JUST AN EXAMPLE ###
+    deployment: MLFlowDeploymentService = mlflow_model_registry_deployer_step(
```
fa9r, line 54: Wait, does this work?

fa9r: Anyways, maybe nicer to split this up by moving `mlflow_model_registry_deployer_step` invocation into the pipeline itself and passing the resulting service into this step as an input

examples/e2e/steps/promotion/promote_model.py
```diff
@@ -0,0 +1,145 @@
...
+from zenml.logger import get_logger
+from zenml.model_registries.base_model_registry import ModelVersionStage
+
+logger = get_logger(__name__)
+
+model_registry = Client().active_stack.model_registry
+
+
+@step
+def promote_model(
```
fa9r, line 37: Also very interesting, I think we've never considered how to do that properly in the past. Maybe it would make sense to add a specific prebuilt step for this to `zenml.integrations.mlflow.steps` since I assume this is a common use case

fa9r: Also, have you tried whether this works actually? In particular, the two `mlflow_model_registry_deployer_step` calls? It looks a bit questionable to me but TBH I also don't know how else you could do this.

fa9r (commented):
Wow, amazing how many new ideas you're bringing to the team! 🙌 🙌 🙌

In particular, I love how you mix steps and non-step function in your pipeline and how you pass artifacts between them. I've never seen that before actually. Also lot's of interesting steps like promotion, hyper param tuning, ...

For now I only left review comments on a conceptual level but what I saw so far looks very very good. Let me know once you think the example is ready to merge then I'll do another more detailed review including docstrings, coding style, ...

examples/e2e/pipelines/batch_inference.py
avishniakov: Yeah, that's actually better - wrapped `notify_on_success` into step.

examples/e2e/steps/training/model_evaluator.py
avishniakov: Thanks, moved up

examples/e2e/utils/artifacts.py
avishniakov: Thanks - reworked

examples/e2e/pipelines/batch_inference.py
avishniakov: Yes, it works. I also didn't want to go with load and passing cause of the double saving issue.
I had to use the standard function to evaluate `id` on pipeline compose, cause otherwise it would fail.
I was also thinking to put this helper function somewhere in the standard, cause it seems to be a rather generic case for 2+  pipelines per product.

examples/e2e/steps/inference/inference_predict.py
avishniakov: Yes, it also works. Moving this out would be fine here, but might be a bit complex for the promotion stage. I'll think about it a bit more.

examples/e2e/steps/promotion/promote_model.py
avishniakov: All works - no issues, but here it will be complex to split steps. I might think of refactoring `mlflow_model_registry_deployer_step`  to call a function `mlflow_model_registry_deployer_function` for consistency and use the same function calls in promotion or other steps internally. What do you think?
Logic inside the deployer is too complex to reimplement, imho.

examples/e2e/steps/inference/inference_predict.py
avishniakov: Carved out

examples/e2e/pipelines/batch_inference.py
avishniakov: Ok, I went high stakes and modified the `ExternalArtifact` class to fit the purpose.
Can you have a close look [here](https://github.com/zenml-io/zenml/pull/1710/commits/fe0111892e2a53ce4c27dcd78ef0a68bcfe61dd7) please @schustmi @fa9r @bcdurak

avishniakov:
@schustmi I added you here for `ExternalArtifact` check - no need to review whole thing, if this is not interesting or out of the capacity 🙂

examples/e2e/steps/etl/data_loader.py
avishniakov: Absolutely, have it on my TODOs. Was just not sure, if the toy dataset is anyhow better than random.
While bringing a real-life case with external DWH overloads this code with the hassle of configuring such a source on all possible clouds (S3, GCP, local, Azure).

If the toy dataset is better from your PoV - I'll rework it, with no arguments.

src/zenml/steps/external_artifact.py
```diff
@@ -138,6 +153,17 @@ def upload_if_necessary(self) -> UUID:
                     "issues loading the artifact. Please make sure to only "
                     "reference artifacts stored in your active artifact store."
                 )
+        elif self._pipeline_name and self._artifact_name:
```
schustmi, line 156: I think this would benefit from the same check as the one with `self._id` above, making sure that the artifact is located in the same artifact store as the one being used by the current pipeline run.

schustmi (changes requested):

src/zenml/steps/external_artifact.py
schustmi: Other than that, love the idea of passing pipeline/artifact name! We could even add a small reference to that in the docs here: https://docs.zenml.io/user-guide/advanced-guide/pipelining-features/configure-steps-pipelines#pass-any-kind-of-data-to-your-steps

fa9r:
> @schustmi I added you here for `ExternalArtifact` check - no need to review whole thing, if this is not interesting or out of the capacity 🙂

Nothing to add from my side except for what Michael already wrote, LGTM 👍

src/zenml/steps/external_artifact.py
avishniakov: Thanks @schustmi  - I reworked the logic to incorporate that, can you have a second look pls?

I'm also unsure how to update docs - will reach out by other channels for guidance 🙂

avishniakov:
Few ongoing things from side discussions:
- HP tunning should be done in parallel in multiple steps
- Config should be revisited for highly mutable params (e.g. target column and so on). For theoretically immutable (like pipeline names) it is ok. If a class is used for mutable parameters, it might bring predicted side effects in pipeline behavior due to the cache not being invalidated.
- Custom Materializers might be needed to support config work, due to sklearn classes in.

schustmi (approved):

examples/e2e/config.py
avishniakov: I put some thinking into the configuration topic and didn't find a better way to do it rather than using class as now.
It can be relatively simple to handle step-level configs with `extra`, but things are getting messed up once we need access to config during pipeline composition.
For the use-case of HP tuning, I generate a set of steps to run the search in parallel based on the config and also rely on it inside the step, because serializing sklearn, scipy, numpy, etc objects of all kinds is super tricky. The search grid for HP tuning might even have some lambdas or whatever else.

So, I prefer to keep this as-is here for now and try to shape better how we deal with configurations.
Thoughts - @fa9r @bcdurak @htahir1 ?

examples/e2e/steps/alerts/slack_notifications.py
avishniakov: This will not work on just `Alerter` isn't it?  I'm relying on stack info to get alerter and fire plain `post` to it.

Btw, a side note: why do we allow only one alerter in the stack? People often want email and Slack (or other means), maybe also with some flexibility on recipient lists.

examples/e2e/config.py
htahir1: @avishniakov Can you make a framework PRD in the framework board and add basic details and context to it? Then put it in a rough timeline for next quarter . We’ll take it up when the time is right

examples/e2e/steps/etl/data_loader.py
avishniakov: done

avishniakov:
@fa9r you asked me to ping you once this PR will be close to its final state - it is now, imho 🙂

avishniakov:
Hey @fa9r @htahir1 @bcdurak , are you good to merge this today, or still need some more time to read through it?
I'll stop the PR check run now, cause [the last one](https://github.com/zenml-io/zenml/actions/runs/5682584364) was green and since then I only modified README.md

htahir1 (approved):
Beautiful! Disclaimer: I have only seen the README, but on that end its an approval

fa9r:
@avishniakov yeah no need to run CI again if code didn't change. I think it's fine to merge but I'll give it another quick review now

src/zenml/steps/external_artifact.py
```diff
@@ -173,6 +178,32 @@ def upload_if_necessary(self) -> UUID:
...
+                        response = artifact
+                        break
+
+            if response is None:
+                raise RuntimeError(
+                    f"Artifact with name `{self._artifact_name}` was not found "
+                    f"in last successful run of pipeline `{self._pipeline_name}`. "
+                    "Please check your inputs and try again."
+                )
+            elif response.artifact_store_id != artifact_store_id:
```
fa9r, line 198: I'd suggest to refactor this a bit since the `if response is None` does not apply to the ID case

```suggestion
            if self._id:
                response = Client().get_artifact(artifact_id=self._id)
            else:  # self._pipeline_name and self._artifact_name
                pipeline = Client().get_pipeline(self._pipeline_name)
                for artifact in pipeline.last_successful_run.artifacts:
                    if artifact.name == self._artifact_name:
                        response = artifact
                        break
                else:
                    raise RuntimeError(
                        f"Artifact with name `{self._artifact_name}` was not found "
                        f"in last successful run of pipeline `{self._pipeline_name}`. "
                        "Please check your inputs and try again."
                    )

            if response.artifact_store_id != artifact_store_id:
```

fa9r (approved):
The example looks great, really cool use case that you built, in particular I love all the promotion logic. Very nice showcase of ZenML 👏  👏  👏 

If you have some more time before next week, I'd have two minor code improvement suggestions for the `src/` changes. Perhaps you could also add some unit tests for the three different cases how `ExternalArtifact` can now be used (by ID, by value, by pipeline + artifact names). 

But I'd also be fine with just merging this PR as it is and tackling the testing in a separate PR later (in that case, at least add a ticket to the backlog for this so we don't forget)

src/zenml/steps/external_artifact.py
avishniakov: mypy was going crazy with such an implementation, so I had to rewrite it like this.
https://github.com/zenml-io/zenml/actions/runs/5676919051/job/15384458918#step:5:33

src/zenml/steps/external_artifact.py
```diff
@@ -128,17 +143,7 @@ def upload_if_necessary(self) -> UUID:
...
-                raise RuntimeError(
-                    f"The artifact {response.name} (ID: {response.id}) "
-                    "referenced by an external artifact is not stored in the "
-                    "artifact store of the active stack. This will lead to "
-                    "issues loading the artifact. Please make sure to only "
-                    "reference artifacts stored in your active artifact store."
-                )
-        else:
+        if self._value:
             assert self._value is not None
```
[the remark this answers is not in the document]
avishniakov: Good catch!

PR #1711, opened 2023-07-24
Flexible use of `ignore_cols` in `evidently_report_step`

avishniakov:
## Describe changes
I implemented `suppress_missing_ignored_cols_error` param in `evidently_report_step` to achieve flexibility while using this step on unequal datasets and avoid the use of wrapper steps. This param is defaulted to `False`, so the behavior is unchanged until the user explicitly raises the flag.
I checked the docs and this change is not affecting it, since `ignore_cols` is not looked at in depth there. Should it be extended then?

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/evidently/steps/evidently_report.py
+++ b/src/zenml/integrations/evidently/steps/evidently_report.py
@@ -55,10 +55,6 @@ def evidently_report_step(
         download_nltk_data: whether to download the NLTK data for the report
             step. Defaults to False.
 
-    Raises:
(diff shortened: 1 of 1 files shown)


fa9r (approved):
Cool, love it 👍

stefannica (approved):
Neat improvement ! You could have been even bolder and always clean up these "extra" ignored columns without the need for a new parameter. Simply logging a warning should be sufficient.

PR #1718, opened 2023-07-26
Add external artifacts and direct links to run DAG

fa9r:
## Describe changes
I refactored the pipeline run DAG creation logic to include:
- input artifacts of type `ExternalArtifact`
- direct links between nodes that have `after=[...]` dependencies

All changes should be fully backwards compatible and should not break any existing dashboard functionality.

### ExternalArtifact

`ExternalArtifacts` are now correctly shown as input artifacts. The correspoding nodes have `node.data.status="external"` so we can distinguish them in the frontend to potentially visualize them differently in the future.

Screenshot 2023-07-26 at 16 44 28

### `.after`
`.after` dependencies between steps are now correctly reflected in the DAG via corresponding direct edges:

Screenshot 2023-07-26 at 16 44 14

## Pre-requisites

## Types of changes


src/zenml/lineage_graph/lineage_graph.py
```diff
@@ -39,17 +45,124 @@ class LineageGraph(BaseModel):
...
+        for step in run.steps.values():
+            for artifact_name, artifact in step.inputs.items():
+                artifact_id = ARTIFACT_PREFIX + str(artifact.id)
+                if artifact_id not in nodes_ids:
+                    self.add_artifact_node(
+                        artifact=artifact,
+                        id=artifact_id,
+                        name=artifact_name,
+                        step_id=str(artifact.producer_step_run_id),
+                        status="External",
```
AlexejPenner, line 116: Feels a bit like we are misusing the status field here in order to make artifacts fit the mold. Especially with the loosening of the BaseNodeDetails.status typing. Ideally I would keept this as an enum at least so we have a source of truth of allowed statuses.

AlexejPenner (approved):
General approval. Got a small nit to pick, but feel free to ignore it if you feel that it doesn't matter.

src/zenml/lineage_graph/lineage_graph.py
fa9r: Long-term the entire DAG creation should probably move to the frontend, IMO it doesn't make sense that the backend is doing this behind a separate endpoint since all the required information is already contained in the `PipelineRunResponseModel`.

avishniakov (approved):
LGTM. Thanks for quick fix 👍🏼

schustmi (approved):

PR #1740, opened 2023-08-15
Publish server helm chart as part of CI

wjayesh:
## Describe changes
I implemented a gha to publish the zenml server helm chart to an ECR repo.

## Pre-requisites

## Types of changes


.github/workflows/publish_helm_chart.yaml
```diff
@@ -0,0 +1,44 @@
+# package and push the zenml helm chart to the ECR repository
+name: Publish Helm Chart
+
+on: workflow_call
```
AlexejPenner, line 4: Might make sense to also allow this to be called from workflow dispatch in case it fails on release

.github/workflows/publish_helm_chart.yaml
```diff
@@ -0,0 +1,44 @@
...
+      PYTHONIOENCODING: 'utf-8'
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      # add tag version to Chart.yaml file for variable version
+      - name: Add tag version to Chart.yaml
+        run: |
+          sed -i "s/version: .*/version: \"${{ github.ref_name }}\"/" src/zenml/zen_server/deploy/helm/Chart.yaml
```
AlexejPenner, line 22: When it comes to commands like this, I like it when they are commented a bit verbosly:

# Finds like that contains "version: " and replaces the complete line with "version: ___"

src/zenml/zen_server/deploy/helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
-name: zenml
-version: "1.0"
+name: zenml-server
+version: "0.42.1"
```
AlexejPenner, line 3: SHouldn't this be 0.43.0 already? How do we do it moving forward. DOes this get manually updated as well? If Not, could it make sense to leave the version number empty?

AlexejPenner (commented):
Got a few suggestions/ thoughts

.github/workflows/publish_helm_chart.yaml
AlexejPenner: Didin't mean to scream it :D

.github/workflows/publish_helm_chart.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: arn:aws:iam::715803424590:role/gh-action-role-zenml-helm
```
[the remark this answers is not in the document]
wjayesh: yes, exactly

src/zenml/zen_server/deploy/helm/Chart.yaml
wjayesh: yes, it does get updated autonomously as part of the CI. A step fetches the latest verison and updates this file before it is packaged. You're right, it makes sense to leave it empty; I'll make the change.

.github/workflows/publish_helm_chart.yaml
wjayesh: ahaha, point taken

src/zenml/zen_server/deploy/helm/templates/hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2
```
stefannica, line 2: What implications does this have on the supported Kuberentes cluster versions ? Should this be done with an if/else depending on that kube API version ?

src/zenml/zen_server/deploy/helm/templates/hpa.yaml
```diff
@@ -17,12 +17,16 @@ spec:
     - type: Resource
       resource:
         name: cpu
-        targetAverageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+        target:
+          type:  Utilization
+          averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
```
stefannica, line 22: Same question here: is this change required because some Kubernetes cluster APIs don't support the previous syntax ? If so, shouldn't it be in an if/else condition to maintain backwards compatibility with previous versions ?

stefannica (changes requested):

src/zenml/zen_server/deploy/helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
-name: zenml
-version: "1.0"
+name: zenml-server
```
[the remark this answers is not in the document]
wjayesh: No strong reason. I felt that this reflected what the helm chart deploys, more appropriately, since it is now in the public light more than ever. Let me know if I should consider going back

src/zenml/zen_server/deploy/helm/templates/hpa.yaml
wjayesh: With this change, we would stop supporting clusters with versions starting 1.23 and older. I think it's reasonable to no longer support that since 1.23 has reached end of support almost 6 months back and most cloud providers use 1.27 as the default cluster version.

src/zenml/zen_server/deploy/helm/templates/hpa.yaml
wjayesh: Yes, the new syntax is need for v2 of the HPA resource. The older versions would no longer be supported. What I can do is add a different chart version that has support for v1beta1 for the same zenml version but that might need some changes across the board. Still the safest bet and allows us to support everyone.

src/zenml/zen_server/deploy/helm/Chart.yaml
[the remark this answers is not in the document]
stefannica: No need to go back, just considering the implications in the name change. No idea what this does for current deployments.

src/zenml/zen_server/deploy/helm/templates/hpa.yaml
stefannica: You don't need a different chart version, just an if condition here. See this thread for example: https://stackoverflow.com/questions/70346448/deploy-conditional-charts-based-on-kubeversion-helm

stefannica (approved):
Fine to merge it as is as long as we're not breaking existing deployments.

PR #1750, opened 2023-08-18
Bump `rich` and `uvicorn`

jlopezpena:
## Describe changes
I changed the dependency on rich to allow for version 13. This should not break anything (famous last words!) but is needed in order to test more recent versions of Evidently

## Pre-requisites

## Types of changes


strickvl:
Seems like the dependency change broke something. `psutil` used to be there but this loosening of either `rich` or `uvicorn` has removed it. Will need more investigation + probably a more fine-grained specification/pinning of both those dependency changes.

strickvl:
Agreed @jlopezpena. I'll just rerun that failing test now while the runners are underutilized and hopefully it'll go through fine.

fa9r:
@strickvl @jlopezpena `Jinja2` was missing as a server dependency now that `uvicorn` no longer installs it, shoutout to @stefannica for helping debug this ❤️

PR #1763, opened 2023-08-25
Lock `MYSQL` Database during DB migrations 

safoinme:
## Describe changes
We are locking the `MYSQL` DB during migration to prevent the failing that happens when deploying zenml in replicas mode in k8s

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/alembic.py
```diff
@@ -34,6 +34,7 @@
 from zenml.zen_stores import schemas
 
 ZENML_ALEMBIC_START_REVISION = "alembic_start"
+DATABASE_LOCK_TIMEOUT = 15
```
strickvl, line 37: Is this long enough? Wonder whether it should be increased to 30 just in case?

strickvl (commented):
Looks ok to me. Assume you tested it in the specific scenario you mentioned?

src/zenml/zen_stores/migrations/alembic.py
safoinme: I think it is, it was even working with 5s

strickvl (approved):

bcdurak (approved):

PR #1767, opened 2023-08-28
Add `mlstacks` compatibility check to CI

strickvl:
I added a check for the `zenml` release workflow which checks that the current package (editable install on the release branch) can be installed in the same environment as `mlstacks`.

## Pre-requisites

## Types of changes


safoinme (approved):

strickvl:
> How critically does `zenml` rely on `mlstacks`? Do we want this CI check just to inform us or actually it's a hard requirement before releasing

That's a good question. If you can't install the two in the same environment then you can't use `zenml stack deploy...` etc, so I'd say if `mlstacks` is a core part of the ZenML flow / functionality then it should be a hard requirement.

But open to other opinions...

avishniakov (commented):
Hm, I have a feeling that we better catch it earlier rather than on release... Here it is too late to make fixes, so why we not include this as part of regular CI as well?

strickvl:
@avishniakov yeah can do. It doesn't add much in terms of time. I'll update.

fa9r (approved):
Cool, great idea 🙌

PR #1775, opened 2023-08-31
Add `bandit` to CI for security linting

strickvl:
Added bandit to the CI to detect medium and high severity warnings about code we write.

(Made some fixes to handle the errors it found -- some were ignored (like hashing algorithms where security was not the intended purpose) and other small fixes were just made.)

For the `timeout` changes, I read that it's best to choose something that is 1 larger than a multiple of 3, and then beyond that I just chose values that I thought made sense for that particular piece of code.

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/lint-unit-test.yml
+++ b/.github/workflows/lint-unit-test.yml
@@ -99,6 +99,9 @@ jobs:
         # Only run the docstring check on ubuntu-latest and python 3.8
         if: ${{ inputs.os == 'ubuntu-latest' && inputs.python-version == '3.8' }}
 
+      - name: Security check
+        run: bash scripts/check-security.sh
+
       - name: Markdown link check
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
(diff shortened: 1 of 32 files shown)


strickvl:
@avishniakov good question. The problem with changing the hashing algorithm relates to backwards compatibility. i.e. anyone upgrading to a newer version would have the hashes stored in the database using the old algo but this update would mean we essentially wipe / invalidate their cache (in a way that perhaps does unexpected things). To do it in a graceful way would be fairly involved, I felt, where we'd have to maybe store the hashing algorithm in the database along with the hash itself. 

Since none of the places where we're currently using old hashing algorithms involve security, I felt like it was an ok tradeoff. It will prevent us from adding any *new* places where we're using broken hashing algorithms and we can potentially consider switching out the places where it's ignored going forward as part of a bigger ticket.

strickvl:
You're right though that the fix is basically replacing md5 with sha256, but that would break things in unpredictable ways, I thought.

avishniakov (approved):

schustmi (commented):
What was the reason for the various different timeout adjustments (5/7/31/#nosec) for the different requests?

strickvl:
> What was the reason for the various different timeout adjustments (5/7/31/#nosec) for the different requests?

@schustmi see the PR description. I didn't want to make it generically long, but that said I can also see that it might be confusing. Not sure what the right way to handle this is. Raw `request.get()` without a timeout passed in was flagged as a medium risk.

schustmi (approved):

PR #1783, opened 2023-09-04
Improve `ConnectionError` error message

fa9r:
## Describe changes
I improved the error message shown when being connected to a ZenML server that is no longer running or reachable.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/base_zen_store.py
+++ b/src/zenml/zen_stores/base_zen_store.py
@@ -25,9 +25,11 @@ from typing import (
     Type,
     Union,
 )
+from urllib.parse import urlparse
 from uuid import UUID
 
 from pydantic import BaseModel
+from requests import ConnectionError
 
 import zenml
 from zenml.analytics.utils import analytics_disabler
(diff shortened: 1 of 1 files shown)


src/zenml/zen_stores/base_zen_store.py
```diff
@@ -133,6 +135,25 @@ def __init__(
...
+                recommendation = (
+                    "Please run `zenml disconnect` and `zenml up` to restart "
+                    "the server."
+                )
+            else:
+                recommendation = (
+                    "Please run `zenml disconnect` and `zenml connect --url "
+                    f"{self.url}` to reconnect to the server."
+                )
+            raise RuntimeError(f"{error_message} {recommendation}") from e
```
strickvl, line 155:
```suggestion
            raise RuntimeError(f"{error_message}\n{recommendation}") from e
```

Maybe instead?

src/zenml/cli/server.py
```diff
@@ -259,7 +259,8 @@ def down() -> None:
     deployer = ServerDeployer()
     deployer.remove_server(server.config.name)
 
-    GlobalConfiguration()
+    gc = GlobalConfiguration()
+    gc.set_default_store()
```
strickvl, line 263: Although obvious somehow, this feels like magic. Might it be worth adding another message to say that the current stack has been set to the default stack? I've certainly been burned by running a pipeline with the wrong stack in the past.

fa9r: > that the current stack has been set to the default stack

It doesn't do that though, I just tested manually. The logic of this function is quite complex but AFAIK this only tells ZenML to revert back to the default SQLZenStore, which is used in this function to disconnect from the local server in case it was already dead and couldn't be shut down.

strickvl (approved):
LGTM pending passing tests ;)

PR #1792, opened 2023-09-06
Prevent simultaneous connection to local and remote servers

fa9r:
## Describe changes
Being connected to both a local and remote server at the same time can lead to unexpected and confusing behavior, such as the wrong dashboard opening after `zenml up` or `zenml show`. Therefore, we now raise an error when trying to `zenml up` while being connected to a remote server or when trying `zenml connect` while having a local server running.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/server.py
+++ b/src/zenml/cli/server.py
@@ -120,7 +120,22 @@ def up(
     """
     from zenml.zen_server.deploy.deployer import ServerDeployer
 
+    if connect:
+        logger.warning(
+            "The `--connect` flag is deprecated, has no effect, and will be "
+            "removed in a future release."
+        )
+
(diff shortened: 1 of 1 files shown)


src/zenml/cli/server.py
```diff
@@ -120,8 +120,23 @@ def up(
     """
     from zenml.zen_server.deploy.deployer import ServerDeployer
 
+    if connect:
+        logger.warning(
```
strickvl, line 124: maybe add it here? https://www.notion.so/zenml/Deprecations-903bcaacb5c04954b9f6133b3340a312

bcdurak (approved):
🦭

PR #1793, opened 2023-09-06
Stop old MLflow services when deploying new ones

fa9r:
## Describe changes
We now stop old MLflow deployment service when starting new ones.

This is the main change (L154):
```python
for existing_service in existing_services:
    existing_service.stop(timeout)
```

While implementing this, I also refactored the `mlflow_model_deployer_step` to make it more maintainable. One important functional change is that the step now returns `None` when no model was deployed and no existing service reused instead of returning some mock service.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/mlflow/steps/mlflow_deployer.py
+++ b/src/zenml/integrations/mlflow/steps/mlflow_deployer.py
@@ -56,7 +56,7 @@ def mlflow_model_deployer_step(
     workers: int = 1,
     mlserver: bool = False,
     timeout: int = DEFAULT_SERVICE_START_STOP_TIMEOUT,
-) -> MLFlowDeploymentService:
+) -> Optional[MLFlowDeploymentService]:
     """Model deployer pipeline step for MLflow.
 
     This step deploys a model logged in the MLflow artifact store to a
@@ -79,7 +79,7 @@ def mlflow_model_deployer_step(
         timeout: the number of seconds to wait for the service to start/stop.
 
     Returns:
-        MLflow deployment service
+        MLflow deployment service or None if no service was deployed or found
 
     Raises:
         ValueError: if the MLflow experiment tracker is not found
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/mlflow/steps/mlflow_deployer.py
```diff
@@ -111,98 +111,83 @@ def mlflow_model_deployer_step(
...
-                f"includes a step with a MLflow experiment configured that "
-                "trains a model and logs it to MLflow. This could also happen "
-                "if the current pipeline run did not log an MLflow model  "
-                f"because the training step was cached."
-            )
-            # return an inactive service just because we have to return
-            # something
-            return service
+    # Check whether to deploy a new service
+    if model_uri and deploy_decision:
```
avishniakov, line 129: What is the reason for this `deploy_decision`? Why one would call a deployer without a deploy decision being made?

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
```diff
@@ -111,98 +111,83 @@ def mlflow_model_deployer_step(
...
-        # return the existing service
-        return service
-
-    # even when the deploy decision is negative, if an existing model server
-    # is not running for this pipeline/step, we still have to serve the
-    # current model, to ensure that a model server is available at all times
-    if not deploy_decision and existing_services:
+        logger.info("Stopping existing services...")
+        for existing_service in existing_services:
+            existing_service.stop(timeout)
```
avishniakov, line 155: Major question: why not reuse? I would search existing by Model Name and Model Version and if same - just reuse.

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
```diff
@@ -111,98 +111,83 @@ def mlflow_model_deployer_step(
...
-        f"MLflow deployment service started and reachable at:\n"
-        f"    {new_service.prediction_url}\n"
-    )
-
-    return new_service
+    # Log a warning if no service was deployed or found
+    elif deploy_decision and not model_uri:
+        logger.warning(
+            f"An MLflow model with name `{model_name}` was not "
+            f"logged in the current pipeline run and no running MLflow "
```
avishniakov, line 183: Not sure that I get it: `model_uri` depends on `mlflow_run_id`, which we got from the MlFlow client based on the experiment or pipeline name.
First: I doubt that naming Experiment as Pipelines automatically is not well protected from collisions, but this is not related to this PR.
Second: if the user fed in the experiment name it will be confusing how Pipeline Run and the model not found in my experiment relate to one another.
So I suggest rephrasing this or making more flexible depending on user inputs.

avishniakov (commented):
I left a few comments here and there, they are mostly conceptual and might not be related to this PO, so ignored if you will.
I have left also a question to the main part of this PR 🙂

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
fa9r: Main reason is that we don't have dynamic pipelines yet, so if you want to decide on the fly whether to deploy a model or not (i.e. by checking whether it outperforms previous models or similar) then you can at least pass the result of your check as an artifact into this step to achieve the behavior you want

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
```diff
@@ -111,98 +111,83 @@ def mlflow_model_deployer_step(
...
             run_id=mlflow_run_id, artifact_path=model_name
         )
 
-    # fetch existing services with same pipeline name, step name and model name
+    # Fetch existing services with same pipeline name, step name and model name
     existing_services = model_deployer.find_model_server(
         pipeline_name=pipeline_name,
         pipeline_step_name=step_name,
         model_name=model_name,
     )
```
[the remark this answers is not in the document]
fa9r: No clue, maybe @safoinme knows?

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
fa9r: Yeah currently the MLflow deployer always assumes that you had an MLflow experiment tracker in a previous stack that logged the model as an MLflow artifact with name "model". We already know this UX is not ideal, refactoring this is actually the main point of the "Complete the model story" PRD is about AFAIK.

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
avishniakov: But why do we deploy, if we find the existing model with the same version? It is the same model - nothing changed.
It is related directly to the previous question.

Now we identify the Server by pipeline, step, and model name and this is probably a reason, cause we have no clue if the model actually changed or not, but if we switch to name+version, there would be no good reason to redeploy the same version, imho.

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
avishniakov: Not really, I checked the code there and how I read this:
- User provides experiment name in params - search inside of it
- User skipped experiment name - search in experiment named pipeline name

To me, this logging is fine for use-case 2, but doesn't make sense for use-case 1.

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
fa9r: Actually model version only exists for model registries right? In this specific step there might not be a model registry so models might not be versioned at all.

fa9r: But thanks for mentioning this, I almost forgot to adjust the model registry deployer step as well

src/zenml/integrations/mlflow/steps/mlflow_deployer.py
fa9r: I didn't fully get that, but anyways I think we should fundamental discussions for the big refactoring coming soon; the goal of this PR was just to stop existing servers when deploying new models

avishniakov (approved):
I messed up with these 2 methods initially, so just drop my comments 👍🏼 
LGTM

safoinme (approved):

PR #1794, opened 2023-09-06
Prevent adding private components into shared stacks

fa9r:
## Describe changes
`zenml stack update ...` previously didn't check whether added components are shared before adding them into shared stacks. I now added such a check to prevent unusable shared stacks containing private components.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -1670,9 +1670,9 @@ class Client(metaclass=ClientMetaClass):
             stack_spec_path=stack_spec_file,
         )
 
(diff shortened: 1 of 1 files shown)


strickvl (approved):
This looks good. Wondering, though, if there's an argument to be made for some kind of force or recursive update flag to just handle this? This way where we just `Error` out is probably safer, but it's also more annoying as a user experience.

fa9r:
It's an edge case anyways so IMO error is fine

PR #1797, opened 2023-09-07
Fix AWS container registry image pushing with service connectors

fa9r:
## Describe changes
Using the AWS container registry via service connectors previously raised a boto authentication error when trying to check for existing repositories if local AWS credentials were not present. As a temporary fix, we simply catch this error now, raise a warning instead, and try to push images regardless of whether repositories could be found or not.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/aws/container_registries/aws_container_registry.py
+++ b/src/zenml/integrations/aws/container_registries/aws_container_registry.py
@@ -17,7 +17,7 @@ import re
 from typing import List, Optional, cast
 
 import boto3
-from botocore.exceptions import ClientError
+from botocore.exceptions import ClientError, NoCredentialsError
 
 from zenml.container_registries.base_container_registry import (
     BaseContainerRegistry,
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/aws/container_registries/aws_container_registry.py
```diff
@@ -70,9 +70,27 @@ def prepare_image_push(self, image_name: str) -> None:
...
+        except NoCredentialsError:
+            logger.warning(
+                "Amazon ECR requires you to create a repository before you can "
+                f"push an image to it. ZenML is trying to push the image "
+                f"{image_name} but could not find any repositories because "
+                "your local AWS credentials are not set. We will try to push "
+                "anyway, but in case it fails you need to create a repository "
+                f"named `{repo_name}`."
+            )
+            return
```
strickvl, line 92: Do we need this `return` if it's going to continue onwards to run the push anyway?

strickvl: or even do we want it?

fa9r: This function doesn't actually do the push, it just checks what kind repositories exist and raises a warning if it does not find the repository it expects. Thus, the `return` simply prevents the other checks (since we have already raised a warning and don't want to raise another back to back)

strickvl (approved):

schustmi (approved):

PR #1804, opened 2023-09-11
Use the container registry credentials to build images with the local image builder

stefannica:
## Describe changes
When a custom base image is used to build container images for containerized orchestrators, the base image can also be a private container image coming from the same container registry where the resulting image is pushed. In this case, linking the container registry to a service connector doesn’t authenticate the Docker client in the image building phase, so authentication errors will occur.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/image_builders/local_image_builder.py
+++ b/src/zenml/image_builders/local_image_builder.py
@@ -101,7 +101,12 @@ class LocalImageBuilder(BaseImageBuilder):
         """
         self._check_prerequisites()
 
-        docker_client = DockerClient.from_env()
+        if container_registry:
+            # Use the container registry's docker client, which may be
+            # authenticated to access additional registries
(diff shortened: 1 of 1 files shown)


schustmi (approved):

schustmi:
Maybe long-term the image builder could instead be linked to a different docker service connector to pull the image from?

stefannica:
> Maybe long-term the image builder could instead be linked to a different docker service connector to pull the image from?

@schustmi I also thought of that, but should that be symmetric then and also use the image builder service connector to push images ? Doesn't that completely replace the container registry ?

fa9r (approved):

schustmi:
I was actually thinking that we should maybe have it to allow the asymmetry of pulling from/pushing to different container registries. However that might actually be solvable also by having different container registries in the stack and not tying the image builder to a container registry.

PR #1814, opened 2023-09-15
External authenticator support

stefannica:
## Describe changes

Auth changes required to enable external authenticator support and other improvements:
* consolidated ZenML Server configuration in its own class
* moved JWT token authentication related logic to its own class
* add support for more JWT token claims (issuer, audience) for increased security
* add support for JWKS (public key criptography) secret keys
* add support in client (REST Store) to automatically re-authenticate when a temporary JWT token expires
* add support to store JWT access tokens in HTTP-only cookies with configurable domain and cookie name for increased security (TBD UI needs to be updated to not store tokens)

## Pre-requisites

## Types of changes


src/zenml/zen_server/jwt.py
```diff
@@ -0,0 +1,156 @@
...
+        user_id: The id of the authenticated User.
+        api_key_id: The id of the API key used to authenticate the user.
+        permissions: The permissions scope of the authenticated user.
+        claims: The original token claims.
+    """
+
+    user_id: UUID
+    api_key_id: Optional[UUID] = None
+    permissions: List[str]
+    claims: Dict[str, Any] = {}
```
AlexejPenner, line 49: Have you discussed this with @schustmi  ? will we continue using jwt permissions and scopes for rbac going forward?

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -114,12 +131,316 @@ def token(
...
+        return generate_access_token(
+            user_id=auth_context.user.id,
+            api_key_id=auth_context.api_key.id
+            if auth_context.api_key
+            else None,
+            response=response,
+        )
+
+
+if server_config().auth_scheme == AuthScheme.EXTERNAL:
```
AlexejPenner, line 208:
```suggestion
elif server_config().auth_scheme == AuthScheme.EXTERNAL:
```

src/zenml/zen_stores/migrations/versions/b0e68bb87611_add_api_key_table.py
```diff
@@ -0,0 +1,64 @@
...
+revision = "b0e68bb87611"
+down_revision = "47f55c2589f6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
AlexejPenner, line 22: I'm so used to see batch_op operations in all of our migration scripts. I assume you tested this one and it works fine on sqlite?

src/zenml/zen_stores/migrations/versions/47f55c2589f6_add_external_id_to_user.py
```diff
@@ -0,0 +1,38 @@
+"""add external id to user [47f55c2589f6].
+
+Revision ID: 47f55c2589f6
```
AlexejPenner, line 3: Why not combine both alembic migrations into one file?

src/zenml/zen_server/routers/users_endpoints.py
```diff
@@ -159,185 +168,214 @@ def get_user(
...
-        user_name_or_id: Name or ID of the user.
-        user_update: the user to use for the update.
-
-    Returns:
-        The updated user.
-    """
-    user = zen_store().get_user(user_name_or_id)
+# When the auth scheme is set to EXTERNAL, users cannot be updated via the
+# API.
+if server_config().auth_scheme != AuthScheme.EXTERNAL:
```
AlexejPenner, line 173: Any reason you query this multiple times instead of wrapping all the endpoints in one if condition?

AlexejPenner (commented):
Left some comments looking through this. One question that came to mind for me, have we thought about the impact that this has on analytics? We probably need to use the external_id for analytics calls now so theanalytics calls from cloud_api and server are attached tot he same distinct mixpanel API? 

@bcdurak I summon you

src/zenml/cli/api_key.py
```diff
@@ -0,0 +1,222 @@
...
+@click.argument("name_or_id", type=str, required=True)
+@click.option("--name", type=str, required=False, help="The new name.")
+@click.option(
+    "--description", type=str, required=False, help="The new description."
+)
+@click.option(
+    "--active",
+    type=bool,
+    required=False,
+    help="Whether the API key is active.",
```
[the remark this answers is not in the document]
stefannica: It's to set the API key to active/inactive. I'll update the help string.

src/zenml/client.py
```diff
@@ -4958,6 +4966,174 @@ def get_service_connector_type(
...
+        description: Optional[str] = None,
+        active: Optional[bool] = None,
+    ) -> APIKeyResponseModel:
+        """Update an API key.
+
+        Args:
+            name_id_or_prefix: Name, ID or prefix of the API key to update.
+            name: New name of the API key.
+            description: New description of the API key.
+            active: Whether the API key is active or not.
```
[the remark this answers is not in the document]
stefannica: It's to set the API key to active/inactive. I updated the help string.

src/zenml/config/server_config.py
```diff
@@ -0,0 +1,199 @@
...
+        if cors_allow_origins := values.get("cors_allow_origins"):
+            origins = cors_allow_origins.split(",")
+            values["cors_allow_origins"] = origins
+        else:
+            values["cors_allow_origins"] = ["*"]
+
+        return values
+
+    @classmethod
+    def get_server_config(cls) -> "ServerConfiguration":
```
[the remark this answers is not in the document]
stefannica: Yes, the prefix and the fact that we don't use dotenv for the global configuration or store configurations either. I wouldn't introduce dotenv here just for the server options.

src/zenml/zen_server/jwt.py
stefannica: I kept these here for backwards compatibility only, but they make no sense when RBAC is introduced, because the authorization is decoupled from the JWT tokens and managed internally.

src/zenml/zen_stores/migrations/versions/b0e68bb87611_add_api_key_table.py
stefannica: `batch_op` is only used to update existing tables, not to create new ones.

src/zenml/zen_server/routers/users_endpoints.py
stefannica: There is an order to these endpoints that I don't want to mess up. I'm only disabling the PUTs and POSTs that are no longer valid.

fa9r (changes requested):
Looks good but couldn't see any tests for this - @schustmi perhaps you could add a few before merging?

PR #1819, opened 2023-09-18
[Model WatchTower] add ModelConfig to step and pipeline decorators

avishniakov:
## Describe changes
I added support to pass `ModelConfig` via `model_config` to `@step` and `@pipeline`. It is also accessible via step context, where `@step` setting dominates over `@pipeline` setting.

Attachment of runs and artifact logging will follow in other PRs.

How to use:
```python
from zenml import step,pipeline, get_step_context
from zenml.model import ModelConfig

@step(model_config=ModelConfig(name="foo",save_models_to_registry=True,recovery=True,create_new_model_version=True,))
def dummy():
    ctx = get_step_context()
    print(ctx.parameters)
    print(ctx.pipeline_run.config.extra)
    print(ctx.step_run.config.extra)
    print(ctx.step_run.config.model_config)

@pipeline(model_config=ModelConfig(name="bar",save_models_to_registry=True,recovery=True,create_new_model_version=True,))
def p():
    dummy()

if __name__=="__main__":
    p()
```

P.S. I carved `ModelBaseModel` out of `models` to `model` to break the hell of cyclic imports I was locked in for a while.

## Pre-requisites

## Types of changes


src/zenml/model/base_model.py
```diff
@@ -0,0 +1,61 @@
...
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Model base model to support Model WatchTower feature."""
+
+from typing import List, Optional
+
+from pydantic import BaseModel, Field
+
+STR_FIELD_MAX_LENGTH = 255
+TEXT_FIELD_MAX_LENGTH = 65535
```
strickvl, line 21: We already have both of these constants in `/Users/strickvl/coding/zenml/repos/zenml/src/zenml/models/constants.py` FYI

src/zenml/models/model_models.py
strickvl, line 1: `src/zenml/models/model_models.py` feels a bit of a confusing abstraction to me ;) though I guess it's probably not too user-facing?

tests/integration/functional/steps/test_model_config.py
```diff
@@ -0,0 +1,63 @@
+from zenml import get_step_context, pipeline, step
+from zenml.model.model_config import ModelConfig
+
+
+@step
+def _assert_that_model_config_set(name="foo", version="bar"):
```
strickvl, line 6: add docstring?

strickvl (approved):
A few tiny comments but otherwise this LGTM.

src/zenml/model/base_model.py
avishniakov: Yeah, but I had to move `ModelBaseModel` out of `models` to break circular loops, thus redefined these here. Not the best one, but serves the purpose.

src/zenml/models/model_models.py
avishniakov: Some Response models are user-facing in the sense that they will come in response to certain actions, but generally setting a `model_config`in `@step`, `@pipeline` or `ArtifactConfig` will cover most of the uses automatically.

strickvl: ok then I think it's fine. just typing `model` three times one after another felt a bit off.

src/zenml/model/base_model.py
fa9r: Strong opinion that this should be kept in `models/` and I'm sure we can find a cleaner way to break the circular imports (send me the stack trace if you need help on this)

fa9r (changes requested):
Partial review - currently many line changes are related to moving `ModelBaseModel` which I think should be undone first

src/zenml/model/base_model.py
avishniakov: This is my way to approach this: https://github.com/zenml-io/zenml/pull/1819/commits/0fa0d5a78bf77376b28e945e55a234ccf0bccaf5
In short: avoid exposing it out of the modules and use by full link from the file aside. Looks rather good to me.

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -79,7 +79,7 @@
 )
 from zenml.io import fileio
 from zenml.logger import get_logger
-from zenml.model.model_stages import ModelStages
+from zenml.model import ModelStages
```
fa9r, line 82: Was this changed intentionally?

```suggestion
from zenml.model. model_stages import ModelStages
```

fa9r (approved):
`pipelines/base_pipeline.py`, `pipelines/pipeline_decorator.py`, `steps/base_step.py` and `steps/step_decorator.py` are all deprecated so I wouldn't adjust them (we'll delete them anyways soon - no point to risk braking them now)

Aside from that, looks good. Can't really tell if all the necessary places have been adjusted but I think we should be good to go if none of the new tests are failing 👍

PR #1825, opened 2023-09-20
[Model WatchTower] Delete running versions on failed pipelines with `delete_new_version_on_failure ` option

avishniakov:
## Describe changes
I implemented a clean-up step `delete_running_versions_without_recovery` in `pipleine.run_` - it runs on the failure of the pipeline and cleans up all hanging ModelVersions in `running` created during the pipeline run, but not having a `delete_new_version_on_failure` option.

Here we have one conceptual renaming: `recovery` of `ModelConfig` becomes `delete_new_version_on_failure` to avoid confusion on model version retrieval logic going forward.

Also inside `Pipeline` new method `get_new_version_requests` is added to fetch all model configurations across the pipeline requesting new versions to be created. This aims for several goals:
- Output of it is passed to `register_running_versions` on success to make `running` versions normal numbered versions or to `delete_running_versions_without_recovery` on failure to clean-up versions not requesting to keep them on failure.
- It warns users if all steps inside the pipeline are defining their own model configuration and overrides one defined in the pipeline if defined.
- It warns users that several model configurations (in pipeline and step(s) or in several steps) request new model version creation for the same model - this is not impacting logic, still, one version will be created, but this is a poor design pattern and the new version should be requested in one place only.

I also touched a bit `ModelVersionResponseModel` to better support versioned artifacts:
- `artifact_object_ids`, `deployment_ids` and `model_object_ids` becomes `Dict[str, Dict[str, UUID]]`: `{"artifact_name":{"1":[...],"2":[...]}}`, where `1` and `2` are artifact link versions for versioned artifact
- `get_model_object` and others also accept version now

The issue with this implementation to be fixed in https://github.com/zenml-io/zenml/pull/1839:
- Using artifact_name as a key in `artifact_object_ids` and others of `ModelVersionResponseModel` causes collisions with not uniquely named artifacts

## Pre-requisites

## Types of changes


src/zenml/models/model_models.py
```diff
@@ -70,15 +70,16 @@ class ModelVersionResponseModel(
     model: "ModelResponseModel" = Field(
         title="The model containing version",
     )
-    model_object_ids: Dict[str, UUID] = Field(
+    # TODO: rework to Dict[str,Dict[UUID]] and so on
+    model_object_ids: Dict[str, Dict[str, UUID]] = Field(
```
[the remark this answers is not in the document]
fa9r: Ah, name -> version -> ID?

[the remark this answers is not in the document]
fa9r: Feels a bit clunky but not sure how else to implement it so feel free to ignore

src/zenml/models/model_models.py
```diff
@@ -143,44 +153,55 @@ def pipeline_runs(self) -> Dict[str, PipelineRunResponseModel]:
             for name, pr in self.pipeline_run_ids.items()
         }
 
-    def get_model_object(self, name: str) -> ArtifactResponseModel:
-        """Get model object linked to this version.
+    def get_model_object(self, name: str) -> Dict[str, ArtifactResponseModel]:
```
fa9r, line 156: IMO think would be a more intuitive interface - same for the other two methods below

```suggestion
    def get_model_object(self, name: str, version:str) -> ArtifactResponseModel:
```

src/zenml/new/pipelines/pipeline.py
```diff
@@ -642,6 +645,14 @@ def _run(
...
             try:
                 stack.deploy_pipeline(deployment=deployment_model)
+            except Exception as e:
+                runs = Client().list_pipeline_runs(
+                    deployment_id=deployment_model.id,
+                    sort_by="desc:start_time",
+                    size=1,
+                )
+                self.delete_running_versions_without_recovery(runs.items[0])
+                raise e
```
fa9r, line 655: Multiple runs can be linked to the same deployment, so this might not be safe. E.g., if I have a pipeline that I run on schedule every 1h, but each run takes >1h, then it could happen that this call queries the wrong pipeline run

src/zenml/new/pipelines/pipeline.py
```diff
@@ -785,6 +796,51 @@ def register_running_versions(
...
+                    model_name_or_id=model_name,
+                    model_version_name_or_id=RUNNING_MODEL_VERSION,
+                )
+            except KeyError as e:
+                if model_name == pipeline_model_name:
+                    logger.warning(
+                        f"Failed to register stable model version of `{model_name}` model. "
+                        f"No `{RUNNING_MODEL_VERSION}` found. "
+                        "Most probable root cause: you set ModelConfig on pipeline level and "
+                        "override it in all steps inside that pipeline."
```
fa9r, line 839: We have this warning in several places now - couldn't we check and silently handle this case at compile time somehow?

src/zenml/new/pipelines/pipeline.py
fa9r: Ah, if you only need the configs, then here's an easy fix: The deployment already has both the pipeline config and all step configs in it, so you can just pass that into `delete_running_versions_without_recovery()` directly

fa9r (changes requested):
Few minor comments but overall this looks good to go 👍

src/zenml/new/pipelines/pipeline.py
avishniakov: Good catch, I will include this into https://zenml.atlassian.net/browse/OSS-2463. I don't have runs linked yet.

src/zenml/model/model_config.py
```diff
@@ -204,18 +204,19 @@ def _get_or_create_model_version(
...
+        try:
+            # trying to get existing `running` version
+            # if recovery is disable it can be only a version from current run
+            # if it exists - it was created in previous steps
+            mv = zenml_client.zen_store.get_model_version(
+                model_name_or_id=self.name,
+                model_version_name_or_id=self.version,
+            )
+        except KeyError:
+            if self.recovery:
```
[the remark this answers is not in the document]
avishniakov: As the description of `recovery` states - it is not controlling that but rather controls if the `running` version will get deleted on failing or not. So this is by design.
`recovery: Whether to keep failed runs with new versions for later recovery from it.`

But I will rename it for better readability.

src/zenml/new/pipelines/pipeline.py
avishniakov: I can completely suppress it, just wanted to give users visibility that they are doing weird stuff.

avishniakov: I reworked that part heavily :)

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5830,6 +5830,10 @@ def create_model_version_artifact_link(
             if existing_model_version_artifact_link is not None:
                 version = existing_model_version_artifact_link.version + 1
 
+            version = 1
+            if existing_model_version_artifact_link is not None:
+                version = existing_model_version_artifact_link.version + 1
+
```
avishniakov, line 5836: This is duplicated code, I already have a commit to remove it on my local, but will wait for CI to complete first.

src/zenml/model/model_config.py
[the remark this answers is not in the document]
fa9r: My main point is: `create_new_model_version=True` will not always create new model versions. Which goes against what the docstring says:

> New version will be created if `create_new_model_version`, otherwise will try to fetch based on `version`.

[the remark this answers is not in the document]
avishniakov: Make sense, I will reflect in docs better. Thanks!

src/zenml/new/pipelines/pipeline.py
fa9r: Would suggest to change it in this PR directly so we don't forget - after #1835 gets merged the runs and steps themselves won't even have configs anymore so your current will be broken in general

src/zenml/model/model_config.py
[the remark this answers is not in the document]
avishniakov: It will look like this after I commit:
```
        A new model is created implicitly if missing, otherwise existing model is fetched.

        A new model version is only created if `create_new_model_version` is True, but if there is an
        existing model version left over from the previous failed run with `delete_new_version_on_failure`
        is set to False, it will be fetched without creating a new one or deleting the old one.

        If `create_new_model_version` is set to False, a new model version will be fetched based on `version`.
```
Waiting for CI again :)

fa9r (changes requested):
See unresolved comments from last review

src/zenml/new/pipelines/pipeline.py
avishniakov: Ah, this one was only about runs, got it now. I read it a bit differently: what if I have a scheduled pipeline and it creates a new version of the model? Runtime is 2h and the schedule is every 1h - there will be a clash and mess in versions, but for this, I need to link pipeline runs first.

I prepared a fix for passing deployment object into `get_new_version_requests` 👍🏼

src/zenml/model/model_config.py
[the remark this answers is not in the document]
fa9r: - If there is an existing model version left over from the previous failed run with `delete_new_version_on_failure` set to False, it will be reused.
- Otherwise, if `create_new_model_version` is True, a new model version is created.
- If `create_new_model_version` is False and no version from a previous failed run exists, a model version will be fetched based on `version`.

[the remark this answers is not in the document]
fa9r: Maybe like this clearer?

fa9r (approved):
Nice, thanks for adjusting, looks good to me now 👍

src/zenml/model/model_config.py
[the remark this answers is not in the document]
avishniakov: This is not accurate, cause the leftover version will only be used when a new version is requested, otherwise, if `version` is set - it will dominate and a specific version will be fetched.

My yet another attempt to make it clear and reflect reality:
```
        A new model is created implicitly if missing, otherwise existing model is fetched. Model
        name is controlled by the `name` parameter.

        Model Version returned by this method is resolved based on model configuration:
        - If there is an existing model version leftover from the previous failed run with
        `delete_new_version_on_failure` is set to False and `create_new_model_version` is True,
        leftover model version will be reused.
        - Otherwise if `create_new_model_version` is True, a new model version is created.
        - If `create_new_model_version` is False a model version will be fetched based on the version:
            - If `version` is not set, the latest model version will be fetched.
            - If `version` is set to a string, the model version with the matching version will be fetched.
            - If `version` is set to a `ModelStage`, the model version with the matching stage will be fetched.
```

avishniakov:
Test failure caused by not stable behavior of MySQL due to the way or storing datetime. Fixed that, but will not wait for full CI once again.

PR #1826, opened 2023-09-20
Improve `zenml stack describe` to show `mlstacks` outputs

strickvl:
Added a way for people to display `mlstacks` outputs using the standard `zenml stack describe` command. Output will look something like this:

ScreenShot 2023-09-20 at 10 50 07

This restores the old functionality (and as such is a sort of bug fix) we had with stack recipes where users can get the Terraform outputs using the zenml CLI, albeit from within the `stack describe` subcommand which I thought preferable to any other new `... outputs` subcommand.

## Pre-requisites

## Types of changes


src/zenml/cli/utils.py
```diff
@@ -430,6 +465,15 @@ def print_stack_configuration(
...
 
+def print_mlstacks_stack_outputs(stack: "StackResponseModel") -> None:
+    """Prints outputs for stacks deployed with mlstacks.
+
+    Args:
+        stack: Instance of a stack model.
+    """
+    pass
+
+
```
fa9r, line 476: This can be deleted I guess?

```suggestion
```

src/zenml/cli/utils.py
```diff
@@ -391,6 +391,41 @@ def format_integration_list(
...
+    rich_table = table.Table(
+        box=box.HEAVY_EDGE,
+        title="MLStacks Outputs",
+        caption=stack_caption,
+        show_lines=True,
+    )
+    rich_table.add_column("OUTPUT_KEY", overflow="fold")
+    rich_table.add_column("OUTPUT_VALUE", overflow="fold")
+
+    from mlstacks.utils.terraform_utils import get_stack_outputs
```
fa9r, line 413: `mlstacks` is a core dependency now so this will never fail right?

fa9r (approved):

src/zenml/cli/utils.py
strickvl: Oh you're totally right I should add a check for that. Thank you! (`mlstacks` is not a core dependency just as soon with #1611 `python-terraform` won't be either.)

strickvl: Added the check for mlstacks installation now.

safoinme (approved):
LGTM

PR #1835, opened 2023-09-23
Optimizing the performance through database changes

bcdurak:
## Describe changes

This PR aims to improve the performance of ZenML by optimizing the queries made to the ZenStore, changing the Schemas & Models (of pipeline deployments, pipeline runs, and step runs), and modifying the hydration logic.

# Changes in the Models

- **Pipeline Deployment Models:**
    - Client- and server versions have been added to the `PipelineDeploymentBaseModel`.
- **Pipeline Run Models:**
    - The following fields have been removed from the `PipelineRunBaseModel`: `schedule_id`, `enable_cache`, `config`, `num_steps`, `client_version`, and `server_version`
    - A `ScheduleResponseModel` and the `config` have been added to the `PipelineRunResponseModel`
    - The `PipelineRunRequestModel` now gets a `PipelineDeploymentResponseModel` as an input instead of stack, pipeline, build, and deployment IDs.
- **Step Run Models:**
    - The `config` and `spec` objects are moved from the `StepRunBaseModel` to the `StepRunResponseModel`.
    - A `deployment_id` is added to the `StepRunResponseModel`.
    - The `StepRunRequestModel` now gets a `PipelineDeploymentResponseModel` as an input.

# Changes in the Schemas

Previously, the hydration (`to_model` method) of the several schemas (Artifacts, Step Run, and Pipeline Runs) was dependent on various additional input that was not accessed through the defined schema. For instance, every time an artifact got hydrated to a model, you needed to provide a `producer_step_id`. However, this ID was not accessed through the schema.

As a result, every time you wanted to hydrate one of these schemas, you needed to use a helper method defined in the `SQLZenStore` instead of calling the `to_model()` method of the given schema. Not only this goes against the concept of schema hydration but this helper method also had to conduct several additional queries to be able to figure out the required input for this operation. This behavior has now changed. Each schema relies solely on its own attributes and relationships to conduct the `to_model()` hydration.

- **Artifact Schema**
    - The `output_to_step_runs` relationship has been modified to be used in the hydration logic.
    - The `UserResponseModel` in the `ArtifactResponseModel` is not fully hydrated anymore.
- **Pipeline Deployment Schema**
    - Client- and server versions have been added as new fields.
    - The `UserResponseModel` in the `PipelineDeploymentResponseModel` is not fully hydrated anymore.
- **Pipeline Run Schema**
    - `stack_id`, `pipeline_id`, `build_id`, `schedule_id` have been removed as foreign keys.
    - `enable_cache`, `enable_artifact_metadata`, `pipeline_configuration`, `num_steps`, `client_version`, `server_version`, `client_environment`, `git_sha` have been removed as fields.
    - The values that are required for the hydration process are accessed by using the `deployment_id`.
- **Run Metadata Schema**
    - The `UserResponseModel` in the `ArtifactResponseModel` is not fully hydrated anymore.
- **Step Run Schema**
    - `enable_cache`, `enable_artifact_metadata`, `entrypoint_name`, `parameters`, `step_configuration`, `caching_parameters`, `num_outputs` have been removed as fields.
    - The values that are required for the hydration process are accessed by using the `deployment_id`.
    - The `UserResponseModel` in the `StepRunResponseModel` is not fully hydrated anymore.

Additionally, several unused relationships have been removed.

## Pre-requisites

## Types of changes


src/zenml/models/pipeline_run_models.py
```diff
@@ -286,12 +237,13 @@ def generate_filter(
...
+            PipelineRunSchema,
+        )
+
         if self.code_repository_id:
-            from zenml.zen_stores.schemas import (
-                CodeReferenceSchema,
-                PipelineDeploymentSchema,
-                PipelineRunSchema,
-            )
+            from zenml.zen_stores.schemas import CodeReferenceSchema
```
fa9r, line 246: Maybe cleaner to move the other import up aswell

src/zenml/models/pipeline_run_models.py
```diff
@@ -317,10 +298,7 @@ class PipelineRunRequestModel(
     """Pipeline run model with user, workspace, pipeline, and stack as UUIDs."""
 
     id: UUID
-    stack: Optional[UUID]  # Might become None if the stack is deleted.
-    pipeline: Optional[UUID]  # Unlisted runs have this as None.
-    build: Optional[UUID]
-    deployment: Optional[UUID]
+    deployment: "PipelineDeploymentResponseModel"
```
fa9r, line 301: Again not a fan of requiring a response model as input for the request model - shouldn't the deployment ID be enough?

src/zenml/zen_stores/schemas/run_metadata_schemas.py
```diff
@@ -129,7 +129,9 @@ def to_model(self) -> "RunMetadataResponseModel":
             value=json.loads(self.value),
             type=self.type,
             workspace=self.workspace.to_model(),
-            user=self.user.to_model() if self.user else None,
+            user=self.user.to_model(_block_recursion=True)
+            if self.user
+            else None,
```
fa9r, line 134: feels very arbitrary - why only the user, why only for this specific model?

fa9r: Would suggest we keep those as is and tackle hydration separately

src/zenml/zen_stores/schemas/pipeline_schemas.py
```diff
@@ -79,13 +78,15 @@ class PipelineSchema(NamedSchema, table=True):
...
     user: Optional["UserSchema"] = Relationship(back_populates="pipelines")
+    workspace: "WorkspaceSchema" = Relationship(back_populates="pipelines")
 
     schedules: List["ScheduleSchema"] = Relationship(
         back_populates="pipeline",
     )
     runs: List["PipelineRunSchema"] = Relationship(
-        back_populates="pipeline", sa_relationship_kwargs={"cascade": "delete"}
+        sa_relationship_kwargs={"cascade": "delete"}
     )
```
fa9r, line 90: Shouldn't this have a relationship to deployments now instead and cascade-delete all pipeline deployments which in turn cascade-delete all of their runs?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3986,7 +3851,6 @@ def list_artifacts(
                 query=query,
                 table=ArtifactSchema,
                 filter_model=artifact_filter_model,
-                custom_schema_to_model_conversion=self._artifact_schema_to_model,
```
fa9r, line 3989: Now that we no longer need these for runs / steps / artifacts, I think we could get rid of the `custom_schema_to_model_conversion` arg in `filter_and_paginate()` entirely.

src/zenml/zen_stores/schemas/pipeline_deployment_schemas.py
```diff
@@ -121,21 +130,18 @@ class PipelineDeploymentSchema(BaseSchema, table=True):
...
-        sa_column=Column(
-            String(length=MEDIUMTEXT_MAX_LENGTH).with_variant(
-                MEDIUMTEXT, "mysql"
-            ),
-            nullable=False,
-        )
-    )
-    client_environment: str = Field(sa_column=Column(TEXT, nullable=False))
+    pipeline_runs: List["PipelineRunSchema"] = Relationship()
+    step_runs: List["StepRunSchema"] = Relationship()
```
fa9r, line 144: These should cascade delete I think

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -147,11 +134,20 @@ def to_model(
...
+                break
+
         return ArtifactResponseModel(
             id=self.id,
             name=self.name,
             artifact_store_id=self.artifact_store_id,
-            user=self.user.to_model() if self.user else None,
+            user=self.user.to_model(_block_recursion=True)
+            if self.user
+            else None,
```
fa9r, line 150: Same here, let's tackle hydration separately
```suggestion
            user=self.user.to_model() if self.user else None,
```

src/zenml/zen_stores/migrations/versions/3e86a04263a5_optimizing_database.py
```diff
@@ -0,0 +1,235 @@
...
+                mysql.MEDIUMTEXT(), "mysql"
+            ),
+            existing_nullable=False,
+        )
+        batch_op.alter_column(
+            "client_environment",
+            existing_type=sa.TEXT(),
+            type_=sqlmodel.sql.sqltypes.AutoString(),
+            existing_nullable=False,
+        )
```
fa9r, line 51: Might cause issues to change this type - probably we changed it to TEXT before because someone reached the limit of regular strings before

src/zenml/zen_stores/migrations/versions/3e86a04263a5_optimizing_database.py
```diff
@@ -0,0 +1,235 @@
...
+
+    with op.batch_alter_table("pipeline_run", schema=None) as batch_op:
+        batch_op.alter_column(
+            "status", existing_type=sa.VARCHAR(), nullable=True
+        )
```
fa9r, line 56: Why is status nullable now?

fa9r (changes requested):

src/zenml/orchestrators/publish_utils.py
```diff
@@ -162,15 +162,14 @@ def update_pipeline_run_status(pipeline_run: PipelineRunResponseModel) -> None:
...
-    assert pipeline_run.num_steps is not None
     client = Client()
     steps_in_current_run = depaginate(
         partial(client.list_run_steps, pipeline_run_id=pipeline_run.id)
     )
 
     new_status = get_pipeline_run_status(
         step_statuses=[step_run.status for step_run in steps_in_current_run],
-        num_steps=pipeline_run.num_steps,
+        num_steps=len(pipeline_run.steps),
```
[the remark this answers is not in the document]
bcdurak: Fixed it through the pipeline spec as discussed.

src/zenml/models/pipeline_run_models.py
bcdurak: Since the models are also referenced in the respective schema.py files, the imports are here to avoid a circular dependency and thus can not be moved to the top of the file. Following your suggestion, I still moved them outside of their respective if blocks.

src/zenml/zen_stores/schemas/pipeline_deployment_schemas.py
bcdurak: For both `pipeline_runs` and `step_runs`, the `cascade delete` behavior is already defined when we define the foreign keys:

```python
    deployment_id: UUID = build_foreign_key_field(
        source=__tablename__,
        target=PipelineDeploymentSchema.__tablename__,
        source_column="deployment_id",
        target_column="id",
        ondelete="CASCADE",
        nullable=True,
    )
```

Defining these args in the `Relationship` object only matters if you do the deletion (or any other operation for that matter) on the schema object directly which we never do. I think defining such relationships creates a lot of confusion in the code that's why I was trying to remove the redundant parts as much as possible.

src/zenml/zen_stores/schemas/run_metadata_schemas.py
bcdurak: Even though this is a minimal change compared to the changes in the configuration, this has an effect on the final representation of a pipeline run. The run metadata models are repeated quite a few times already and this change just excludes the roles in the user model (which are never used). If it is alright, I would leave this change in.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: I would rather do this in a separate PR.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -124,32 +79,22 @@ class PipelineRunSchema(NamedSchema, table=True):
...
-    orchestrator_environment: Optional[str] = Field(
-        sa_column=Column(TEXT, nullable=True)
+    pipeline_id: Optional[UUID] = build_foreign_key_field(
+        source=__tablename__,
+        target=PipelineSchema.__tablename__,
+        source_column="pipeline_id",
+        target_column="id",
+        ondelete="SET NULL",
+        nullable=True,
     )
```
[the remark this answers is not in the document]
bcdurak: We need the link for the status as we discussed.

src/zenml/models/step_run_models.py
```diff
@@ -248,6 +251,7 @@ class StepRunRequestModel(StepRunBaseModel, WorkspaceScopedRequestModel):
         title="Logs associated with this step run.",
         default=None,
     )
+    deployment: "PipelineDeploymentResponseModel"
```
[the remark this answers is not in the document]
bcdurak: Changed it according to our discussion.

src/zenml/models/pipeline_run_models.py
bcdurak: Changed it according to our discussion.

src/zenml/zen_stores/migrations/versions/3e86a04263a5_optimizing_database.py
bcdurak: Must be fixed now.

src/zenml/zen_stores/migrations/versions/3e86a04263a5_optimizing_database.py
bcdurak: Must be fixed now.

src/zenml/zen_stores/migrations/versions/3e86a04263a5_optimizing_database.py
```diff
@@ -0,0 +1,235 @@
...
+        batch_op.drop_column("client_environment")
+        batch_op.drop_column("enable_artifact_metadata")
+        batch_op.drop_column("schedule_id")
+
+    with op.batch_alter_table("step_run", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "deployment_id", sqlmodel.sql.sqltypes.GUID(), nullable=True
+            )
+        )
```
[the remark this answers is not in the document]
bcdurak: Must be fixed now.

src/zenml/zen_stores/schemas/pipeline_schemas.py
bcdurak: You are right, fixed the relationship.

src/zenml/models/step_run_models.py
```diff
@@ -251,7 +250,10 @@ class StepRunRequestModel(StepRunBaseModel, WorkspaceScopedRequestModel):
         title="Logs associated with this step run.",
         default=None,
     )
-    deployment: "PipelineDeploymentResponseModel"
+    deployment: UUID = Field(
```
fa9r, line 253:
```suggestion
    deployment_id: UUID = Field(
```

fa9r (changes requested):
This will not work - just kidding, it should work fine and looks good overall now, just have two minor changes I'd still like to see before merging :)

src/zenml/models/pipeline_run_models.py
```diff
@@ -298,7 +293,9 @@ class PipelineRunRequestModel(
     """Pipeline run model with user, workspace, pipeline, and stack as UUIDs."""
 
     id: UUID
-    deployment: "PipelineDeploymentResponseModel"
+    deployment: UUID = Field(
```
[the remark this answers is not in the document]
bcdurak: Same as above

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -136,10 +144,9 @@ def to_model(self) -> ArtifactResponseModel:
 
         producer_step_run_id = None
 
-        for step_run in self.output_of_step_runs:
-            if step_run.step_run.status == ExecutionStatus.COMPLETED:
-                producer_step_run_id = step_run.step_run.id
-                break
+        if self.output_of_step_runs:
+            step_run = self.output_of_step_runs[0].step_run
+            producer_step_run_id = step_run.original_step_run_id
```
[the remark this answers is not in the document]
bcdurak: Good catch. Fixed it.

src/zenml/models/step_run_models.py
fa9r: Oh, yes, you're right. I remember all of them were `X_id` at some point so I thought that was our convention. But apparently this was changed sometime.

fa9r (approved):
This will work 👍

src/zenml/config/compiler.py
```diff
@@ -50,6 +51,20 @@
 logger = get_logger(__file__)
 
 
+def get_zenml_versions() -> Tuple[str, str]:
+    """Returns the version of the ZenML on the client- and server side.
+
+    Returns:
+        the ZenML versions on the client- and server side respectively.
```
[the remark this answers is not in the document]
bcdurak: Good catch!

strickvl (approved):

PR #1840, opened 2023-09-26
ZenML API Keys and Service Accounts

stefannica:
### ZenML Service Accounts

Service accounts are similar to users, with the difference that they are used to authenticate automated workloads like CI/CD jobs where the user-based web login isn't possible.

Internally, service accounts and users are represented using the same legacy user models and database table. This is required for backwards compatibility reasons. At the same time, new models and REST API routes are implemented for use exclusively with service accounts.  

### ZenML API Keys

Add support to generate and rotate long-lived API keys associated with a service account that can be used to authenticate to the ZenML Server.

* API keys can be created only through the CLI at the moment (dashboard support will follow)
* the API keys are associated with a service account
* API keys can be used to authenticate the CLI instead of a username/password.
* when an API key is rotated, the user has the option to configure a grace period during which both keys continue to be valid
* the implementation is done in a way that only exposes the API key value when generated or rotated, but otherwise the (hashed) API key value is kept hidden in the DB layer and never exposed through the REST API.

## TODO

* [x] documentation
* [x] unit tests

## Pre-requisites

## Types of changes


src/zenml/cli/service_accounts.py
```diff
@@ -0,0 +1,515 @@
...
+    is_flag=True,
+)
+@click.option(
+    "--role",
+    "-r",
+    "initial_role",
+    help="Give the service account an initial role.",
+    required=False,
+    type=str,
+    default="admin",
```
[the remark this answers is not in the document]
AlexejPenner: And probably should use an enum?

src/zenml/cli/service_accounts.py
```diff
@@ -0,0 +1,515 @@
...
+)
+def create_service_account(
+    service_account_name: str,
+    skip_api_key: bool = False,
```
AlexejPenner, line 116: create_api_key instead of skip, seems more natural to me imo. Less mental gymnastics required to understand whats going on ( don’t skip being a sort of double negative )

src/zenml/cli/service_accounts.py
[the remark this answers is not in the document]
AlexejPenner: Maybe sync with michael to get one of the new tenant default roles (maybe the one that can run pipelines but not create stacks)

AlexejPenner (commented):
Some initial thoughts, will dive deeper tomorrow :)

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2029,16 +2063,495 @@ def delete_user(self, user_name_or_id: Union[str, UUID]) -> None:
...
+        with Session(self.engine) as session:
+            # Check if a user or service account with the given name already
+            # exists
+            existing_user = session.exec(
+                select(UserSchema).where(
+                    UserSchema.name == service_account.name
+                )
+            ).first()
+            if existing_user is not None:
+                raise EntityExistsError(
```
schustmi, line 2148: Is it necessary to enforce unique names for users + service accounts? While they share a single table in the database, it's not really something that users need to be aware of right?

We could still enforce unique user names and unique service account names but include the `is_service_account` flag in the unique constraint?

schustmi (changes requested):
Amazing job @stefannica, this looks very nice already and I only have a few minor comments. Can't wait to integrate this in my RBAC branch 🎉

src/zenml/models/api_key_models.py
```diff
@@ -0,0 +1,296 @@
...
+            raise ValueError("Invalid API key.")
+
+    def encode(self) -> str:
+        """Encodes the API key in a base64 string that includes the key ID and prefix.
+
+        Returns:
+            The encoded API key.
+        """
+        encoded_key = b64_encode(self.json())
+        return f"{ZENML_API_KEY_PREFIX}{encoded_key}"
```
AlexejPenner, line 78: This is also something stripe does. Separating the prefix from the rest of the key makes it easier to spot and read

```suggestion
        return f"{ZENML_API_KEY_PREFIX}_{encoded_key}"
```

src/zenml/models/service_account_models.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class ServiceAccountBaseModel(BaseModel):
+    """Base model for service accounts."""
+
+    name: str = Field(
+        title="The unique username for the account.",
+        max_length=STR_FIELD_MAX_LENGTH,
+    )
+
```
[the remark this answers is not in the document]
AlexejPenner: +1

src/zenml/zen_stores/migrations/versions/7cea2e1c2007_add_service_accounts.py
```diff
@@ -0,0 +1,65 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "7cea2e1c2007"
+down_revision = "0.45.6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
```
AlexejPenner, line 19: Is this complete? I can't find the APIKey table

src/zenml/zen_stores/sql_zen_store.py
AlexejPenner: I think you should not be able t create a service account with the same name as a user within a tenant simply for clarity, if a service account with name michael runs a pipeline, that would be confusing (even if we find soem way in the frontend to visually separate users from service accounts)

src/zenml/zen_stores/schemas/user_schemas.py
```diff
@@ -200,3 +257,33 @@ def to_model(
                 updated=self.updated,
                 roles=[ra.role.to_model() for ra in self.assigned_roles],
             )
+
+    def to_service_account_model(
```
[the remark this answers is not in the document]
stefannica: There are guards in the SQL zen store to prevent against such a thing (and integration tests for it). Failing here would be redundant.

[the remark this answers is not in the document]
stefannica: The reverse is not true: `to_model()` is still called in lots of places to hydrate the owner of various resources in a `UserResponseModel` format. I have to allow that, otherwise massive code changes would be required in both core and dashboard to allow an `owner` to be either `UserResponseModel` or `ServiceAccountResponseModel`.

src/zenml/models/service_account_models.py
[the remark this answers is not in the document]
stefannica: Love that idea !

src/zenml/zen_stores/sql_zen_store.py
stefannica: We use the "user name" in various places (CLI, dashboard) as a visual hint for the owner of various resources. I agree with @AlexejPenner , it would be confusing to have overlapping names in this space, at least until the dashboard catches up with the service account concept. But you did make me aware of a case where this restriction is problematic: creating a service account with the name `stefan@zenml.io` would block me from having a user account with that same name. Especially in the ZenML Cloud context where usernames are externally managed and ZenML user accounts are created automatically based on SSO data, this would indeed be a problem.

I don't know how to solve this elegantly. Forcing the service account name to contain only alphanumeric characters would take care of our needs at least. Any ideas @schustmi @AlexejPenner ?

stefannica: I might just go ahead and do what @schustmi suggested and make `name`+`is_service_account` the unique constraint to avoid this new situation.

src/zenml/constants.py
```diff
@@ -172,6 +172,7 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 DEFAULT_ZENML_SERVER_DEVICE_AUTH_TIMEOUT = 60 * 5  # 5 minutes
 DEFAULT_ZENML_SERVER_DEVICE_AUTH_POLLING = 5  # seconds
 DEFAULT_HTTP_TIMEOUT = 30
+ZENML_API_KEY_PREFIX = "ZENKEY"
```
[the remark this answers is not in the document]
stefannica: It was suggested to me by @wjayesh that branding our API keys this way will make it easier to detect leaked keys e.g. with github actions. I think it's a good idea. `SK` is too generic.

src/zenml/models/api_key_models.py
stefannica: done (in the `ZENML_API_KEY_PREFIX` definition)

src/zenml/zen_stores/migrations/versions/7cea2e1c2007_add_service_accounts.py
stefannica: I think you meant to ask about the `ServiceAccount` table ? Service accounts and user accounts share the same `User` table.

src/zenml/constants.py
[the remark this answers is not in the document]
AlexejPenner: great point by @wjayesh <3

src/zenml/zen_stores/migrations/versions/7cea2e1c2007_add_service_accounts.py
AlexejPenner: No this question was about the Table for API Keys, I couldn't find any mention  of it in the migrations. Does that table already exist?

schustmi: Isn't that the only thing that this migration script is doing though, it creates an `api_key` table :P

src/zenml/cli/service_accounts.py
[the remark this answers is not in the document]
stefannica: This will go away after Michael's RBAC changes.

src/zenml/zen_stores/sql_zen_store.py
stefannica: This is the solution I went for, for future reference:
* user accounts and service accounts can have overlapping names
* calling `get_user` with an account name does not return a service account, even if one has the same name
* calling `get_service_account` never returns a user account
* however, calling `get_user` with a service account UUID will return that service account (for backwards-compatibility purposes).

src/zenml/zen_stores/migrations/versions/7cea2e1c2007_add_service_accounts.py
AlexejPenner: feeling so stupid rn, what the hell was I thinking

AlexejPenner (approved):
:seal: from my side

tests/unit/conftest.py
```diff
@@ -413,6 +413,7 @@ def sample_user_model() -> UserResponseModel:
         name="axl",
         created=datetime.now(),
         updated=datetime.now(),
+        is_service_account=True,
```
schustmi, line 416: Shouldn't this be false?

schustmi (changes requested):

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1882,6 +1897,118 @@ def delete_flavor(self, flavor_id: UUID) -> None:
...
+                    continue
+                assert fk.parent is not None
+                foreign_key_attr = fk.parent.name
+                break
+
+            assert foreign_key_attr is not None
+
+            foreign_keys.append((target_schema, foreign_key_attr))
+
+        return foreign_keys
```
AlexejPenner, line 1980: :cold_sweat:

stefannica: I have added plenty of unit tests for this, no worries

Issue #1833, opened 2023-09-22
[BUG]: Issue when supplying user in docker settings


mlops-rf:
### Contact Details [Optional]

mlops@richfox.com

### System Information

ZENML_LOCAL_VERSION: 0.44.2
ZENML_SERVER_VERSION: 0.44.2

### What happened?

Docker image build fails when user is specified in docker settings. Issue is due to https://github.com/zenml-io/zenml/blob/main/src/zenml/utils/pipeline_docker_image_builder.py.
Lines are:
```
 if docker_settings.user:
            lines.append(f"USER {docker_settings.user}") # specifying user without creating it beforehand
            lines.append(f"RUN chown -R {docker_settings.user} .") # setting chown on directory with user without rights to execute 
chown [only root can execute chown]
```
If the docker settings are specifed as following:
```
docker_settings = DockerSettings(required_integrations=[MLFLOW, SKLEARN],
                                 user="test")
```
Then, image will be built automatically and here is the issue
```
Step 9/12 : COPY . .
Step 10/12 : RUN chmod -R a+rw .
Step 11/12 : USER test # here user is specified while in code is not added. One can add user before with RUN adduser --disabled-password --gecos '' runner or some similar command
Step 12/12 : RUN chown -R test . # this will pop issue due to user changing permissions. On unix only root can change permissions
```

Fix should be quite easy for this.

### Reproduction steps

1. Specify user in docker settings and pass to pipeline.
```
docker_settings = DockerSettings(required_integrations=[MLFLOW, SKLEARN, GITLAB],
                                 user="test-user")

settings = {
    "docker": docker_settings,
}
@pipeline(enable_cache=False, settings=settings)
def test_pipeline(n_estimators) -> None:
  pass

```

### Relevant log output

```shell
RuntimeError: Docker error: The command '/bin/sh -c chown -R test-user .' returned a 
non-zero code: 1.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

strickvl:
@mlops-rf Opened a PR to fix this bug / issue + expect it'll make it into the next release. I'll let you know when that happens!

strickvl:
This was fixed in #1857. Thanks for reporting it!

PR #1857, opened 2023-10-04
Allow for `user` param to be specified (successfully) in `DockerSettings`

strickvl:
Addresses #1833, where a community member found that we were incorrectly setting the USER in our pipeline `Dockerfile` generator function.

I updated it so that it:

- Identified a permission issue occurring when trying to change file ownership using the `chown` command as a non-root user.
  - Kept the `RUN chown -R {docker_settings.user} .` instruction to change the file ownership to the specified user.
  - Added another `USER {docker_settings.user}` instruction to switch back to the specified user for any subsequent instructions in the Dockerfile.
- This modification ensures that the `chown` command is executed as the root user, resolving the permission issue while still allowing for a specified user to be set for subsequent instructions.

## Pre-requisites

## Types of changes


src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -638,8 +638,12 @@ def _generate_zenml_pipeline_dockerfile(
         lines.append("RUN chmod -R a+rw .")
 
         if docker_settings.user:
-            lines.append(f"USER {docker_settings.user}")
+            # Switch to root user to change file ownership
+            lines.append("USER root")
```
fa9r, line 642: Isn't this redundant since `root` is the default? And even if you were not root at this point, wouldn't `RUN chmod -R a+rw .` already have failed before even reaching this?

schustmi (approved):

fa9r (approved):

tests/unit/utils/test_pipeline_docker_image_builder.py
```diff
@@ -42,6 +42,7 @@ def test_check_user_is_set():
             download_files=False,
         )
     )
+    assert "USER root" in generated_dockerfile
```
fa9r, line 45: :)
```suggestion
```

PR #1858, opened 2023-10-04
Connect to Service-connector at component registration

safoinme:
## Describe changes
Adding the ability to set a service connector to connect to when registering the component for the first time to avoid running 2 commands. The implementation uses the function used to connect and calls it after the component is registered to ensure all validation we have in the connect command is called without actually re-writing it.

## Pre-requisites
- [ ] I have updated the documentation if my change requires a change to docs.
- [ ] I have based my new branch on `develop` and the open PR targets `develop`. If your branch wasn't based on develop read [Contribution guide on rebasing branch to develop](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md#-pull-requests-rebase-your-branch-on-develop).

## Types of changes


src/zenml/cli/stack_components.py
```diff
@@ -288,6 +305,14 @@ def register_stack_component_command(
...
             print_model_url(get_component_url(component))
 
+        connect_stack_component_with_service_connector(
+            component_type=component_type,
+            name_id_or_prefix=name,
+            connector=connector,
+            interactive=False,
+            no_verify=False,
+        )
+
```
stefannica, line 315: You also need a resource ID argument for some resource types, such as cloud Kubernetes clusters. You're also missing out on the best part of linking a component to a connector: the interactive mode.

src/zenml/cli/stack_components.py
```diff
@@ -288,6 +305,14 @@ def register_stack_component_command(
             )
             print_model_url(get_component_url(component))
 
+        connect_stack_component_with_service_connector(
```
stefannica, line 308: you probably need to run this only `if connector`

stefannica (changes requested):

fa9r (approved):
`connect_stack_component_with_service_connector` needs to be in the `if` condition, otherwise LGTM 👍

PR #1859, opened 2023-10-05
[Model Control Plane] parallel running versions support

avishniakov:
## Describe changes
The ultimate goal of this PR is to enable multiple parallel runs of pipelines creating new ModelVersions. This is achieved by the following:
- Enable users to pass in version display name to act as a distinguisher between different running versions. If no name is explicitly passed a default `running` name will be used.
- On failed run, we decide to keep or to drop the current running version based on delete on failure flag:
  - If deletion is requested - the version is removed
  - If recovery is requested - the version is kept with an original name for future runs.
- On the successful run we keep the version original name for user-provided names, for the default name we substituted it with the model version number (a continuous auto-increment number)
- If a new run is started - a new version can only be created but will fail if it is already in place. Default names are substituted on success, so these pipelines can run as long they don't collapse with one another - 2 or more running default versions are not possible!

To support those there were a few extra changes:
- ModelVersion schema was shaken a bit: `version` is now `name` and the new field `number` is added
- `model_version_number` added to `ExternalArtifact` and `model_version` renamed to `model_version_name`
- All get methods for ModelVersion extended to get in version number as well
- Renamed `ArtifactConfig.model_version_name` to `ArtifactConfig.model_version` and unified it to support version number input also
- `ModelConfig` extended to receive `version_name`, `version_number` and `version_description`. The root validator moved to the parent class.
- Groomed new model version requests logic in pipelines and steps.
- Pipeline run is linked to Model Version at the beginning of the step from context and at the end from artifacts. This was needed to ensure parallel runs on the same running version protection.
- Merged alembic migrations

Bug fixes:
- `ExternalArtifact.model_version` and `ExternalArtifact. model_artifact_name `  typing was off - changed that.

## Pre-requisites

## Types of changes


src/zenml/artifacts/external_artifact.py
```diff
@@ -99,7 +100,8 @@ def _validate_all(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         pipeline_name = values.get("pipeline_name", None)
         artifact_name = values.get("artifact_name", None)
         model_name = values.get("model_name", None)
-        model_version = values.get("model_version", None)
+        model_version_name = values.get("model_version_name", None)
+        model_version_number = values.get("model_version_number", None)
```
fa9r, line 104: I don't fully understand why this is needed. Feels like `model_version_name` is the model version and `model_version_number` is the version of the model version. If that's correct then I think this needs to be redesigned since have multi-layered versioning makes it to confusing in my opinion

fa9r (changes requested):
Need to clarify below before more detailed review

src/zenml/artifacts/external_artifact.py
avishniakov: This supports what @stefannica  asked for.
Model Version is a container entity, which is represented by a display name (`model_version_name`) and continuous number (`model_version_number`). A display name is what the user can control, while a number is an auto-incremental integer assigned to it during creation. This change gives the user the flexibility to fetch the model version either by display name or by number.
These two are split into separate args, cause I thought it might be too confusing to understand why a "1" would be fetched by name, while 1 would be fetched by number while keeping it mixed in all internal methods.

src/zenml/artifacts/external_artifact_config.py
```diff
@@ -42,12 +42,33 @@ class ExternalArtifactConfiguration(BaseModel):
...
+        """
+        if values.get("model_version_number", None) and values.get(
+            "model_version_name", None
+        ):
+            logger.warning(
+                "`model_version_number` has higher priority then `model_version_name`."
+                "Setting `model_version_name` to `None`."
+            )
+            values["model_version_name"] = None
+        return values
```
fa9r, line 70: IMO this feels a bit clunky - why does the user need to specify whether it's a number or name in the first place?

src/zenml/models/model_base_model.py
```diff
@@ -64,18 +70,91 @@ class ModelBaseModel(BaseModel):
...
+                "`version_number` cannot be used with `create_new_model_version`."
+            )
+
+        if version_number is not None and version_name is not None:
+            logger.warning(
+                "`version_number` has higher priority then `version_name`."
+                "Setting `version_name` to `None`."
+            )
+            version_name = None
+            values["version_name"] = None
```
fa9r, line 142: Same here - feels a bit clunky from a UX perspective

tests/integration/functional/steps/test_model_config.py
```diff
@@ -807,3 +841,110 @@ def _inner_pipeline():
...
+)
+def test_that_two_pipelines_cannot_run_at_the_same_time_requesting_new_version_and_with_recovery(
+    version,
+):
+    """Test that if second pipeline for same new version is started in parallel - it will fail."""
+    with model_killer():
+        long_runtime = 120
+        long_run_name = f"long_{uuid4()}"
+        # start a long running pipeline with a request for a new version
+        long_process = subprocess.Popen(
```
fa9r, line 863: Very cool you decided to test this via actual parallelism 👍 👏 

Only thing I was wondering, is there a specific reason why you use `subprocess` over `multiprocessing`?

In the latter case you wouldn't need to leave Python, e.g. this should work:

```python
from multiprocessing import Process

p1 = Process(target=lambda: pipeline_1())
p1.start()
p2 = Process(target=lambda: pipeline_2())
p2.start()
p1.join()
p2.join()
```

fa9r (commented):
The main change (pipeline.py L840) looks good and I also love all the new tests you added 🙌

However, I still don't fully get why we need both `model_version_name` and `model_version_number`, from scanning the code I didn't see any concrete advantage of adding this and all the new warnings/errors make it quite clunky from a UX perspective IMO

tests/integration/functional/steps/test_model_config.py
avishniakov: I started with this, but it was not working as expected in the CI of Github, not sure why, so I had to switch to `subprocess` which is a more OS-natural approach.

avishniakov: But I tried with `Pool` actually, not sure if `Process` would be any better, but since I'll do some rework anyway today, I'll give a shot to this proposal as well.

avishniakov: I restored what happened back then - in CI it will start throwing `Unable to determine source root because the main module does not have an associated file. This could be because you're running in an interactive Python environment.`, thus I had to switch to a subprocess calling script.

src/zenml/artifacts/external_artifact_config.py
avishniakov: I simplified it 👍🏼

src/zenml/models/model_base_model.py
avishniakov: I simplified it 👍🏼

PR #1864, opened 2023-10-06
Allow disconnecting service-connector from stack component

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -2523,7 +2523,7 @@ def put(
         Returns:
             The response body.
         """
-        logger.debug(f"Sending PUT request to {path}...")
+        logger.debug(f"Sending PUT request to {path} and ...")
```
fa9r, line 2526: and ...?

```suggestion
        logger.debug(f"Sending PUT request to {path ...")
```

src/zenml/cli/stack_components.py
```diff
@@ -1703,6 +1703,55 @@ def connect_stack_component_command(
...
+        "name_id_or_prefix",
+        type=str,
+        required=True,
+    )
+    def disconnect_stack_component_command(name_id_or_prefix: str) -> None:
+        """Rename a stack component.
+
+        Args:
+            name_id_or_prefix: The name of the stack component to rename.
+        """
```
fa9r, line 1729:
```suggestion
        """Disconnect a stack component from a service connector.

        Args:
            name_id_or_prefix: The name of the stack component to disconnect.
        """
```

src/zenml/cli/stack_components.py
```diff
@@ -1833,6 +1882,15 @@ def command_group() -> None:
         help=f"Connect {singular_display_name} to a service connector.",
     )(connect_command)
 
+    # zenml stack-component connect
+    disconnect_command = generate_stack_component_disconnect_command(
+        component_type
+    )
+    command_group.command(
+        "disconnect",
+        help=f"Disconnect {singular_display_name} to a service connector.",
```
fa9r, line 1891:
```suggestion
        help=f"Disconnect {singular_display_name} from a service connector.",
```

src/zenml/client.py
```diff
@@ -2263,10 +2266,23 @@ def update_stack_component(
             }
             update_model.labels = existing_labels
 
-        if connector_id is not None:
+        if disconnect:
+            logger.info(update_model)
```
fa9r, line 2270: I guess this was a debug print? Or what does this do?

```suggestion
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1525,8 +1525,15 @@ def update_stack_component(
                         "Service connector with ID "
                         f"{component_update.connector} not found."
                     )
+            else:
+                existing_component.connector = None
+                existing_component.connector_resource_id = None
```
fa9r, line 1530: Currently, if any value in an update model is `None`, it means that this value was not provided and should not be updated. So I think the current design is breaking this convention since `connector=None` now disconnects the existing connector and thereby leads to updates.

Instead, I would suggest to add `disconnect` as a field of `StackComponentUpdateModel` explicitly and then use this here instead:

```suggestion
            elif component_update.disconnect:
                existing_component.connector = None
                existing_component.connector_resource_id = None
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1525,8 +1525,15 @@ def update_stack_component(
...
 
             if service_connector:
+                logger.debug(
+                    "Updating component %s with connector %s",
+                    existing_component,
+                )
                 existing_component.connector = service_connector
```
fa9r, line 1537: I'd suggest to move this entire block up into the `if component_update.connector:` block after L1521

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1525,8 +1525,15 @@ def update_stack_component(
...
                     )
+            else:
+                existing_component.connector = None
+                existing_component.connector_resource_id = None
 
             if service_connector:
+                logger.debug(
+                    "Updating component %s with connector %s",
+                    existing_component,
+                )
```
fa9r, line 1536: String has two `%s` but you only give one formatting input (connector is not passed)

fa9r (changes requested):
Great functionality, looking forward to using that 👍

Overall the implementation looks good, however I would prefer if `disconnect` was a separate field in the `StackComponentUpdateModel` since `connector=None` usually just means that this value was not provided and should not be updated. See comment below.

Other than that only minor issues, looks like we can merge it today 🚀

stefannica (approved):
LGTM aside from the changes that Felix already suggested.

src/zenml/zen_stores/sql_zen_store.py
stefannica: I think it's fine as it is, because this immediately fixes the UI as well. There is no disconnect field in the component update model and we shouldn't add one. The alternative also requires dashboard changes.

src/zenml/zen_stores/sql_zen_store.py
stefannica: I think it's fine as it is, because this immediately fixes the UI as well. The alternative also requires dashboard changes. If it helps, this is the same approach used for service connector updates: some update fields are "special", in the sense that if omitted from the update request, they are interpreted as a value removal rather than a value omission.

PR #1867, opened 2023-10-09
[Artifacts Tab] Globally Unique Default Artifact Names

fa9r:
## Describe changes
Artifacts now have globally unique default names `{pipeline_name}::{step_name}::output`.

Note that this only affects the name of the artifact itself (`artifact.name`), not the respective output name (`step.outputs[...]`). This means, e.g., that `step.outputs["output"]` will still work as before, while `step.outputs["output"].name` now returns `{pipeline_name}::{step_name}::output` instead of `output`.

I also added a `has_custom_name` field into the artifact DB table, which we can use in the frontend to decide which artifacts to show by default in the artifact list (the idea is that only named ones are shown by default to prevent clutter)

## Pre-requisites

## Types of changes


src/zenml/orchestrators/step_runner.py
```diff
@@ -567,14 +574,25 @@ def _store_output_artifacts(
 
             uri = output_artifact_uris[output_name]
             materializer = materializer_class(uri)
+            has_custom_name = output_has_custom_name[output_name]
+
+            # Override the artifact name if it is not a custom name.
+            if has_custom_name:
+                artifact_name = output_name
```
avishniakov, line 581: For the reasons I outlined in PRD, I would not go with this but rather have a `f"{pipeline_name}::{step_name}::{output_name}"` pattern all over the place.

src/zenml/models/artifact_models.py
```diff
@@ -46,9 +46,13 @@ class ArtifactBaseModel(BaseModel):
     """Base model for artifacts."""
 
     name: str = Field(
-        title="Name of the output in the parent step.",
+        title="Name of the artifact.",
         max_length=STR_FIELD_MAX_LENGTH,
     )
```
avishniakov, line 51: I would suggest storing pipeline_name and step_name separately, rather than combining them here, cause users might want to use `::` in their names and it might somehow affect the logic. Lookups by separate fields would also be a bit more readable, IMO.

avishniakov (commented):
Left a few conceptual comments, not critical

src/zenml/models/artifact_models.py
fa9r: > cause users might want to use :: in their names and it might somehow affect the logic
Now that we also store `is_custom_name` it should be no problem to discern user-created names with `::` in it from the auto-generated ones, so I don't think this is an issue. 

Technically we actually have all this info in the DB already - the `ArtifactResponseModel` has a `producer_step_run_id` field which is the ID of the step that produced it (which in turn has a link to the run which has a link to the pipeline). So technically we wouldn't need to provide `name` in the request model at all. But we already had the field, so might as well use it 😄

TL;DR: Overall I prefer the current approach since it's the easiest and doesn't require additional DB columns

src/zenml/orchestrators/step_runner.py
fa9r: @schustmi and I discussed this before and decided to not do that. 

The main reason is that you might want to create new versions of the same artifact from different pipelines or steps, or you might want to rename your pipeline or steps at some point which would break the lineage if custom names also need to follow this prefix.

Also, if users want to follow this naming convention, they can easily do so themselves, so why force them

avishniakov (approved):

PR #1869, opened 2023-10-09
[Artifacts Tab] Artifact Versioning

fa9r:
## Describe changes
I added auto-increment artifact versioning and adjusted the CLI and Client methods accordingly.

Docs will be included in a separate PR.

## Pre-requisites

## Types of changes


src/zenml/cli/artifact.py
```diff
@@ -63,7 +62,13 @@ def list_artifacts(**kwargs: Any) -> None:
 
 
 @artifact.command("delete", help="Delete an artifact.")
-@click.argument("artifact_id")
+@click.argument("artifact_name_or_id")
+@click.option(
+    "--version",
+    "-v",
+    default=None,
```
strickvl, line 69:
```suggestion
    default=None,
    type=str,
```

src/zenml/cli/artifact.py
```diff
@@ -63,7 +62,13 @@ def list_artifacts(**kwargs: Any) -> None:
 
 
 @artifact.command("delete", help="Delete an artifact.")
-@click.argument("artifact_id")
+@click.argument("artifact_name_or_id")
```
strickvl, line 65: When you delete the artifact without specifying a name or id, does it delete all versions by default?

src/zenml/cli/artifact.py
strickvl, line 1: Update to CLI docs needed?

src/zenml/client.py
```diff
@@ -3279,6 +3234,7 @@ def list_artifacts(
             created: Use to filter by time of creation
             updated: Use the last updated date for filtering
             name: The name of the run to filter by.
+            version: The version of the artifact to filter by.
```
strickvl, line 3237:
```suggestion
            version: The version of the artifact by which to filter.
```

src/zenml/client.py
```diff
@@ -5017,6 +4990,66 @@ def _get_entity_by_id_or_name_or_prefix(
             f"only one of the {entity_label}s."
         )
 
+    @staticmethod
+    def _get_versioned_entity_by_id_or_name_or_prefix(
```
strickvl, line 4994:
```suggestion
    def _get_entity_version_by_id_or_name_or_prefix(
```

Maybe?

src/zenml/utils/artifact_utils.py
```diff
@@ -369,6 +372,8 @@ def upload_artifact(
             be stored.
         extract_metadata: If artifact metadata should be extracted and returned.
         include_visualizations: If artifact visualizations should be generated.
+        version: The version of the artifact. If not provided, a new
+            auto-increment version will be used.
```
strickvl, line 376:
```suggestion
            auto-incremented version will be used.
```

Maybe?

src/zenml/utils/artifact_utils.py
```diff
@@ -422,6 +428,30 @@ def upload_artifact(
     return response.id
 
 
+def _get_new_artifact_version(artifact_name: str) -> int:
+    """Get the next auto-increment version for an artifact name.
```
strickvl, line 432:
```suggestion
    """Get the next auto-incremented version for an artifact name.
```

src/zenml/utils/artifact_utils.py
```diff
@@ -422,6 +428,30 @@ def upload_artifact(
...
+
+    Args:
+        artifact_name: The name of the artifact.
+
+    Returns:
+        The next auto-increment version.
```
strickvl, line 438:
```suggestion
        The next auto-incremented version.
```

src/zenml/cli/artifact.py
fa9r: Do you think that should be the expected behavior?

fa9r: Right now it only deletes the latest version, but we could easily change it to delete all versions I suppose

src/zenml/client.py
fa9r: Technically better but we use that wording in 130 other places and I'll not bother adjusting all of them 😅

src/zenml/cli/artifact.py
strickvl: I think I'd expect it to delete all versions... but not sure. Was more a clarifying question because if it doesn't do that by default, then there should also be a way to make sure it does that I reckon.

src/zenml/utils/artifact_utils.py
```diff
@@ -427,6 +433,30 @@ def upload_artifact(
...
 
 
+def _get_new_artifact_version(artifact_name: str) -> int:
+    """Get the next auto-incremented version for an artifact name.
+
+    Args:
+        artifact_name: The name of the artifact.
+
+    Returns:
+        The next auto-incremented version.
```
strickvl, line 443: Is it too obvious to mention that auto-incremented means +1?

strickvl: Because people can have non-int versions, too, no?

strickvl:
Also, will there be a Docs PR for this whole set of changes?

src/zenml/cli/artifact.py
avishniakov: I would expect full deletion, not only the latest, UX-wise.

src/zenml/artifacts/external_artifact_config.py
```diff
@@ -167,7 +167,7 @@ def get_artifact_id(
         client = Client()
 
         if self.id:
-            response = client.get_artifact(artifact_id=self.id)
+            response = client.get_artifact(self.id)
```
avishniakov, line 170: Why remove the attribute name?

src/zenml/cli/artifact.py
```diff
@@ -83,37 +89,40 @@ def list_artifacts(**kwargs: Any) -> None:
...
+        artifact_name_or_id: Name or ID of the artifact to delete.
+        version: Version of the artifact to delete.
         only_artifact: If set, only delete the artifact but not its metadata.
         only_metadata: If set, only delete metadata and not the actual artifact.
         yes: If set, don't ask for confirmation.
     """
     if not yes:
         confirmation = cli_utils.confirmation(
-            f"Are you sure you want to delete artifact '{artifact_id}'?"
+            f"Are you sure you want to delete artifact '{artifact_name_or_id}'?"
```
avishniakov, line 109:
```suggestion
            f"Are you sure you want to delete version '{version or 'latest'}' from artifact '{artifact_name_or_id}'?"
```

src/zenml/cli/artifact.py
```diff
@@ -83,37 +89,40 @@ def list_artifacts(**kwargs: Any) -> None:
...
+            name_id_or_prefix=artifact_name_or_id,
+            version=version,
             delete_metadata=not only_artifact,
             delete_from_artifact_store=not only_metadata,
         )
     except (KeyError, ValueError) as e:
         cli_utils.error(str(e))
     else:
-        cli_utils.declare(f"Artifact '{artifact_id}' deleted.")
+        cli_utils.declare(f"Artifact '{artifact_name_or_id}' deleted.")
```
avishniakov, line 125:
```suggestion
        cli_utils.declare(f"Artifact version '{version or 'latest'}' from '{artifact_name_or_id}' deleted.")
```

src/zenml/cli/artifact.py
```diff
@@ -166,7 +175,7 @@ def prune_artifacts(
     for unused_artifact in unused_artifacts:
         try:
             Client().delete_artifact(
-                artifact_id=unused_artifact.id,
+                name_id_or_prefix=unused_artifact.id,
```
avishniakov, line 178: It will only wipe our latest versions of them or?

src/zenml/orchestrators/input_utils.py
```diff
@@ -79,7 +79,7 @@ def resolve_step_inputs(
         artifact_id = external_artifact.get_artifact_id(
             model_config=model_config
         )
-        input_artifacts[name] = Client().get_artifact(artifact_id=artifact_id)
+        input_artifacts[name] = Client().get_artifact(artifact_id)
```
avishniakov, line 82: Why remove the attribute name?

src/zenml/utils/artifact_utils.py
avishniakov: +1 to this question

avishniakov: Generally, how these auto-versions will cross with custom versions? Isn't it too confusing for users? Maybe going with Version Name / Version Number approach as we have in Model Versions?

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -160,9 +162,16 @@ def to_model(self) -> ArtifactResponseModel:
             else:
                 producer_step_run_id = step_run.original_step_run_id
 
+        version: Union[str, int]
+        try:
+            version = int(self.version)
+        except ValueError:
+            version = self.version
```
avishniakov, line 169: Does it change anything if you will not cast numerics to int here?

avishniakov (changes requested):
Looks good, left a few comments. On a general note: I would love to see some tests for not first version and also for custom string names. If it will follow in other PRs - ignore.

src/zenml/artifacts/external_artifact_config.py
fa9r: Since it changed :P

src/zenml/cli/artifact.py
fa9r: Reworked that part heavily

src/zenml/cli/artifact.py
fa9r: Reworked that part heavily

src/zenml/cli/artifact.py
fa9r: Reworked that part heavily

src/zenml/orchestrators/input_utils.py
fa9r: Same as above, since it changed

src/zenml/zen_stores/schemas/artifact_schemas.py
fa9r: No actually not. I think I'll just keep them as string to make the typing easier so we don't always need `Union` everywhere

src/zenml/utils/artifact_utils.py
fa9r: > Is it too obvious to mention that auto-incremented means +1?

In my opinion, yes, since it's a super short function that's not even user-facing.

In the docs I'd explain that in more detail of course

fa9r:
@strickvl For smaller changes, I'll try to incorporate docs in the respective PR, but things that will require entire new pages (like versioning) will have their docs coming in a separate PR in the end.

fa9r:
Tests I always want to add with the respective changes though, so these are still missing in this PR

src/zenml/cli/artifact.py
fa9r: Okay, changed it to delete all versions if only the name is provided 👍

src/zenml/cli/artifact.py
```diff
@@ -83,37 +90,76 @@ def list_artifacts(**kwargs: Any) -> None:
...
+    artifact_name_or_id: str,
+    version: Optional[str] = None,
     only_artifact: bool = False,
     only_metadata: bool = False,
     yes: bool = False,
 ) -> None:
-    """Delete an artifact.
+    """Delete an artifact by ID or name.
+
+    If an artifact name without a version is provided, all artifacts with that
```
strickvl, line 101:
```suggestion
    If an artifact name without a version is provided, all artifact versions with that
```

Perhaps?

src/zenml/cli/artifact.py
```diff
@@ -83,37 +90,76 @@ def list_artifacts(**kwargs: Any) -> None:
...
+                f"'{artifact_name_or_id}'?"
+            )
+        elif version:
+            confirmation = cli_utils.confirmation(
+                f"Are you sure you want to delete version '{version}' of "
+                f"artifact '{artifact_name_or_id}'?"
+            )
+        else:
+            confirmation = cli_utils.confirmation(
+                f"Are you sure you want to delete all artifacts with name "
```
strickvl, line 126:
```suggestion
                f"Are you sure you want to delete all artifact versions with name "
```

And wherever else this repeats?

strickvl (commented):
More or less there now, I'd say.

Q: so when people compare versions, they're comparing string to string now? We should warn about that somewhere in docs maybe? (Given `"10" < "2"` evaluates to `True` etc)

fa9r:
> Q: so when people compare versions, they're comparing string to string now? We should warn about that somewhere in docs maybe? (Given "10" < "2" evaluates to True etc)

Ugh yeah, that's a problem. This also breaks the auto-increment feature since the DB column itself is string. Perhaps we do need to add version name / version number representation on a DB level at least. And revert to my original changes were the version in the model was typed as `Union` I suppose.

src/zenml/artifacts/external_artifact_config.py
avishniakov: Maybe better to use new name then?

src/zenml/utils/artifact_utils.py
avishniakov: > > Generally, how these auto-versions will cross with custom versions? Isn't it too confusing for users? Maybe going with Version Name / Version Number approach as we have in Model Versions?
> 
> No, I still think that version name + number is a fundamentally flawed design from a UX perspective

I think you came to this already, but I was not saying that you need to ask users for numbers and names, but have it in DB, same as we do now in Models - user can input 1, "1", "Some name", ModelStages.STAGING, "staging", etc all in the same field. The same logic can be achieved here.
Initially, I was also against extracting numbers, but it is simplifying backend life a bit, IMO. The only downside is that you have to ban numerical version names from being inputted by the user, if you rely on the uniqueness of a name and auto-populate the name with a number, if no user input is given.

src/zenml/cli/artifact.py
avishniakov: I'm not getting what is the difference between this and uuid case. Aren't they the same?

src/zenml/cli/artifact.py
```diff
@@ -83,37 +90,76 @@ def list_artifacts(**kwargs: Any) -> None:
...
+            cli_utils.declare(f"Artifact '{artifact_name_or_id}' deleted.")
+        elif version:
+            cli_utils.declare(
+                f"Version '{version}' of artifact '{artifact_name_or_id}' "
+                f"deleted."
+            )
+        else:
+            cli_utils.declare(
+                f"All artifacts with name '{artifact_name_or_id}' deleted."
+            )
```
avishniakov, line 162: Same question here - why is this different from uuid case?

src/zenml/utils/artifact_utils.py
```diff
@@ -427,6 +433,30 @@ def upload_artifact(
...
+    """
+    models = depaginate(
+        partial(
+            Client().list_artifacts,
+            name=artifact_name,
+            sort_by="desc:version",
+        )
+    )
```
avishniakov, line 451: This can be heavy, maybe adding some extra helper store function to retrieve the maximum numerical name would be a better way. This should be achievable with SQLModels interface.
If you go with Number in DB - ignore this.

tests/integration/functional/models/test_artifact.py
```diff
@@ -31,6 +31,28 @@
     from zenml.client import Client
 
 
+def test_auto_incremented_artifact_versioning(
```
avishniakov, line 34: Good 👍🏼

src/zenml/cli/artifact.py
fa9r: each artifact version has a different ID, so if you specify an ID you only delete that version, whereas for a name there are multiple versions associated to it

src/zenml/utils/artifact_utils.py
fa9r: yeah I think I will go with number in DB 👍

src/zenml/artifacts/external_artifact_config.py
fa9r: don't really think we need arg names. here since there's only a single arg provided and it's obvious what it is used for

src/zenml/cli/artifact.py
fa9r: See comment below - UUID is unique for a version but name applies to multiple versions

src/zenml/cli/artifact.py
avishniakov: Ah, gotca

src/zenml/utils/artifact_utils.py
fa9r: Yeah that's pretty much what I did now:
- in the models and all user-facing methods, `version` is a single field that is `Union[str, int]`, which is intuitive and also allows comparisons of int versions as one would expect.
- in the DB I added version_number as an optional field which is only set for numbered versions to make it more efficient to retrieve numbered versions only (e.g. for auto-increment)

fa9r: At the moment I don't need to ban numeric custom versions and I don't see yet why I would need to - anyway, I will implement that in the follow-up PR that adds manual string versions if we need it

src/zenml/utils/artifact_utils.py
```diff
@@ -375,6 +375,9 @@ def upload_artifact(
         include_visualizations: If artifact visualizations should be generated.
         version: The version of the artifact. If not provided, a new
             auto-incremented version will be used.
+        has_custom_name: If the artifact name is custom and should be listed in
+            the dashboard "Artifacts" tab.
```
strickvl, line 379: Do we need to state 'cloud dashboard' here?

strickvl: Since it won't be on the OSS dashboard

strickvl (approved):

avishniakov (commented):
All good, one conceptual question and one nit.

fa9r:
@avishniakov no, you will get what you ask for, i.e.: 
> I create an artifact version named "2' and it gets assigned number 1, right?
No, if you request "2" you will get 2. 

> I create an unnamed artifact version and what name it gets? Also, it gets number 2 attached, I bet. Right?
No, the current highest number upon creation is 2, so the second artifact you create will have version 3.

> I fetch the artifact version using "2" as input, what do I get in return?
The first artifact you created that you named "2".

avishniakov:
> @avishniakov no, you will get what you ask for, i.e.:
> 
> > I create an artifact version named "2' and it gets assigned number 1, right?
> 
> No, if you request "2", the artifact will have number 2.
> 
> > I create an unnamed artifact version and what name it gets? Also, it gets number 2 attached, I bet. Right?
> 
> No, the current highest number upon creation is 2, so the second artifact you create will have version 3.
> 
> > I fetch the artifact version using "2" as input, what do I get in return?
> 
> If you fetch "2", you get "2", i.e. the first artifact you created

Not sure how to approach this, cause now we have inconsistent behavior between Artifacts and Models. We need to align it somehow, either here or in Model CP, cause it might confuse users.

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -118,9 +120,15 @@ def from_request(
         Returns:
             The converted schema.
         """
+        try:
+            version_number = int(artifact_request.version)
+        except ValueError:
+            version_number = None
```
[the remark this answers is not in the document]
fa9r: might fail if `artifact_request.version` is already of type `int`

fa9r:
@avishniakov Agreed, we probably need @stefannica in that discussion though. Anyway, artifacts tab will not go live this sprint so we can still consolidate this later I suppose. I'm creating a ticket for it 👍

avishniakov (approved):

avishniakov:
> @avishniakov Agreed, we probably need @stefannica in that discussion though. Anyway, artifacts tab will not go live this sprint so we can still consolidate this later I suppose. I'm creating a ticket for it (2521) 👍

Yeah, let's follow up with him. He was advocating the immutability of version number, but let's spin this thread again 🙂

PR #1876, opened 2023-10-10
Evaluate YAML based config early + OSS-2511

avishniakov:
## Describe changes
I implemented an early evaluation of the `config_path` YAML config passed to a pipeline to make it available to `get_pipeline_context` properly and passing of `ModelConfig` as part of YAML config.
User will get warned about overriding of the config in cases like:
```python
@pipeline
def my_pipeline():
  ...
  
p = my_pipeline.with_options(config_path="config.yaml")
p.configuration.extra["a"]  # 1

p.configure(extra={"a": 2}) # warning about overriding existing config and implications of it
p.configuration.extra["a"]  # 2

p() # here extra is again {"a":1}
```

## Pre-requisites

## Types of changes


schustmi (changes requested):
I'm wondering whether this could lead to some weird behaviour due to the config being applied twice, once when the `with_options` method is called and once during compilation which is most likely triggered when the pipeline is executed.

For example, we now apply the config from the yaml file and overwrite its values by options given in code (`**kwargs`). These will stay configured on the pipeline until it gets compiled, at which point the yaml config will be re-applied.

Maybe we can modify it so only the `steps` section of the yaml config gets applied during compilation?

Additionally, I would suggest using `dict_utils.recursive_update(original_dict, update_dict)` instead of `from_config.update(...)` to get recursive merging of dict keys as it happens in other places when people re-configure pipelines/steps

avishniakov:
> I'm wondering whether this could lead to some weird behaviour due to the config being applied twice, once when the `with_options` method is called and once during compilation which is most likely triggered when the pipeline is executed.
> 
> For example, we now apply the config from the yaml file and overwrite its values by options given in code (`**kwargs`). These will stay configured on the pipeline until it gets compiled, at which point the yaml config will be re-applied.
> 
> Maybe we can modify it so only the `steps` section of the yaml config gets applied during compilation?
> 
> Additionally, I would suggest using `dict_utils.recursive_update(original_dict, update_dict)` instead of `from_config.update(...)` to get recursive merging of dict keys as it happens in other places when people re-configure pipelines/steps

Thanks, addressed in recent commits

avishniakov:
I'll suppress CI for now, to give capacity to release

src/zenml/new/pipelines/pipeline.py
```diff
@@ -1108,7 +1108,17 @@ def _compile(
         integration_registry.activate_integrations()
 
         if config_path:
-            run_config = PipelineRunConfiguration.from_yaml(config_path)
+            with open(config_path, "r") as f:
+                from_config: Dict[str, Any] = yaml.load(
+                    f, Loader=yaml.SafeLoader
+                )
+            # pull out parameters relevant for PipelineRunConfiguration
+            from_config = {
```
schustmi, line 1116: Does it make sense to only use the `steps` key here as the rest was already configured earlier?

avishniakov: I just made it extensible, if in future we update the structure of the underlying models for any reason.

schustmi: But if I understand correctly, this now still applies the config twice right? Imagine the following:

```yaml
# config.yaml
extra:
  a: 1
```

```python
@pipeline
def my_pipeline():
  ...
  
p = my_pipeline.with_options(config_path="config.yaml")
p.configuration.extra["a"]  # 1

p.configure(extra={"a": 2})
p.configuration.extra["a"]  # 2

p()
# When running the pipeline, the config file gets reapplied and leads to (IMO) unexpected result of extra["a"] == 1
```

schustmi: I guess this also isn't ideal currently, as the config file only gets applied when running
```
...

p = my_pipeline.with_options(config_path="config.yaml")
p.configuration.extra["a"]  # KeyError
```

avishniakov: I see, let me think about it 👍🏼

avishniakov: Besides this, what about this approach to avoid double applying?
```python
        if config_path:
            with open(config_path, "r") as f:
                from_config: Dict[str, Any] = yaml.load(
                    f, Loader=yaml.SafeLoader
                )
            # pull out parameters relevant for PipelineRunConfiguration and not used in configure method
            configure_args = inspect.getfullargspec(self.configure)[0]
            from_config = {
                k: v
                for k, v in from_config.items()
                if k in PipelineRunConfiguration.__fields__
                and k not in configure_args
            }
            run_config = PipelineRunConfiguration(**from_config)
```

avishniakov: > I guess this also isn't ideal currently, as the config file only gets applied when running
> 
> ```
> ...
> 
> p = my_pipeline.with_options(config_path="config.yaml")
> p.configuration.extra["a"]  # KeyError
> ```

What's inside your `config.yaml`?
This works for me:
```python
In [8]: p_ = p.with_options(config_path="tmp.yaml")

In [9]: p_.configuration
Out[9]: PipelineConfiguration(enable_cache=None, enable_artifact_metadata=None, enable_artifact_visualization=None, enable_step_logs=None, settings={}, extra={'a': 1}, failure_hook_source=None, success_hook_source=None, model_config_model=None, name='p')

In [10]: p.configuration
Out[10]: PipelineConfiguration(enable_cache=None, enable_artifact_metadata=None, enable_artifact_visualization=None, enable_step_logs=None, settings={}, extra={}, failure_hook_source=None, success_hook_source=None, model_config_model=None, name='p')
```
with
```yaml
extra:
  a: 1
```

avishniakov: [8e96223](https://github.com/zenml-io/zenml/pull/1876/commits/8e96223721649f1789dce84b82c6aed7dd7651ed) should solve the double application issue while keeping it flexible for future extensions.

schustmi: Yep I like it. As for the example with the `KeyError`, I didn't actually run anything but assumed that's what was happening. However what I meant by "currently" was actually the code on `develop` and not the current state of your PR, so that was probably the misunderstanding

schustmi (approved):

src/zenml/new/pipelines/pipeline.py
```diff
@@ -354,6 +373,36 @@ def configure(
...
+                    and value_ != self._from_config_file[param_]
+                ):
+                    to_be_reapplied.append(
+                        (param_, self._from_config_file[param_], value_)
+                    )
+            if to_be_reapplied:
+                msg = ""
+                reapply_during_run_warning = (
+                    "The value of parameter '{name}' has changed from "
+                    "'{file_value}' to '{new_value}' set in you configuration file.\n"
```
schustmi:
```suggestion
                    "'{file_value}' to '{new_value}' set in your configuration file.\n"
```

schustmi (approved):
LGTM

PR #1877, opened 2023-10-10
Fixing the `upgrade` migration script after the database changes

bcdurak:
## Describe changes
I have fixed the `upgrade` section of the migration script. 

Context:

- With the new optimization changes, we have added the client version and server version to the pipeline deployment table as non-nullable fields. However, for previous deployments, this has become an issue, as they did not have a version. Now these columns get created as nullable columns first, we populate them with the values from previous runs and then alter the columns to be non-nullable again.

- Similar to the previous point, we added a non-nullable foreign key to the step runs, called `deployment_id`. With the same approach, we make sure that this field is filled with the right value after the migration.

## Pre-requisites

## Types of changes


safoinme (approved):

AlexejPenner (approved):
:seal:

PR #1937, opened 2023-10-11
[Artifacts Tab] Artifact Tags, Renaming & Manual Versioning

fa9r:
## Describe changes

Users can now specify name, version, and tags of an artifact as part of the `ArtifactConfig`:

### Before

```python
from zenml import step
from zenml.model.artifact_config import ArtifactConfig

@step
def my_step() -> Annotated[int, "aria", ArtifactConfig(...)]`:
    ...
```

### After

```python
from zenml import ArtifactConfig, step

@step
def my_step() -> (
    Annotated[
        int,
        ArtifactConfig(
            name="aria", 
            version="super_cat", 
            tags=["cat", "grumpy"],
           ...
        ),
    ]
):
    ...
```

### Docs Page
https://app.gitbook.com/o/-MCl1Hlw9oU4xibZ-ymf/s/vlf6HVLGPRknfyMwMtR2/user-guide/advanced-guide/artifact-management/artifact-versioning

### Detailed changes
- [Breaking Change] Moved `zenml.model.artifact_config.py` to `zenml.artifacts.artifact_config.py` and adjusted the following attributes of `ArtifactConfig`:
    - (new) `version`, and `tags` 
    - (renamed) `artifact_name` -> `name`
    - (renamed) `overwrite` -> `overwrite_model_link`
    - (refactored) turned class variables `IS_MODEL_ARTIFACT` and `IS_DEPLOYMENT_ARTIFACT` into regular fields `is_model_artifact` and `is_deployment_artifact` and deleted `ModelArtifactConfig` and `DeploymentArtifactConfig` classes.
- [Breaking Change] Removed unused `zenml.model.link_output_to_model.py`
- [Breaking Change] Removed package level imports in `zenml.model`, i.e., `from zenml.model import ...` no longer works
- [Breaking Change] Removed the following unnecessary attributes of `ModelVersionArtifact`:
    - name
    - link_version
    - pipeline_name
    - step_name
- [Breaking Change] Removed the following unnecessary attributes of `ModelVersionPipelineRun`:
    - name
- Merged and moved `StepRunner._link_artifacts_to_model` + `ArtifactConfig.link_to_model()` -> `zenml.model.utils.link_step_artifacts_to_model`
- Added `tags` as a field to the artifact models / schemas.
- Added an `update_artifact` endpoint / ... that can be used to update tags or rename artifacts after they were created.

### TODO
- adjust / fix integration tests

## Pre-requisites
Please ensure you have done the following:
[...]

PR #1943, opened 2023-10-13
Artifacts Tab

fa9r:
## Describe changes
Backend changes required to add an "Artifacts" tab to the ZenML dashboard.

### High-Level Changes
- Artifacts now have globally unique default names `{pipeline_name}::{step_name}::output` (#1867)
- All artifacts are now versioned:
    - Users can specify a specific version for an artifact as part of the `ArtifactConfig` (#1937)
    - Otherwise, if no version is specified by the user, an auto-increment integer is assigned (#1869)
- Artifacts can now have tags (#1937)
- `ArtifactConfig` has been heavily reworked and is now decoupled from models (#1937)
    - Users can now specify name, version, and tags of an artifact as part of the ArtifactConfig
- `ExternalArtifacts` have been heavily reworked and are now decoupled from models (#1947)
- `ModelVersion.load_artifact()` is how artifacts of a model can now be loaded instead
- `ModelVersionArtifact` has been heavily reworked (#1937)
- Reworked model CLI and Client methods
- Added util functions for saving and loading ZenML artifacts and attaching metadata to them (#2038)

### New Docs Pages
- https://app.gitbook.com/o/-MCl1Hlw9oU4xibZ-ymf/s/vlf6HVLGPRknfyMwMtR2/user-guide/advanced-guide/artifact-management/artifact-versioning
- https://app.gitbook.com/o/-MCl1Hlw9oU4xibZ-ymf/s/vlf6HVLGPRknfyMwMtR2/user-guide/advanced-guide/artifact-management/artifact-saving-loading

### Comprehensive Code Example

```python
from typing import Optional, Tuple

import numpy as np
from sklearn.base import ClassifierMixin
from sklearn.datasets import load_digits
from sklearn.svm import SVC
import zenml
from zenml import ArtifactConfig, ExternalArtifact, pipeline, step


@step
def versioned_data_loader_step() -> (
    Annotated[
        Tuple[np.ndarray, np.ndarray],
        ArtifactConfig(
            name="my_dataset",
            tags=["digits", "computer vision", "classification"],
        ),
    ]
):
    """Loads the digits dataset as a tuple of flattened numpy arrays."""
    digits = load_digits()
    return (digits.images.reshape((len(digits.images), -1)), digits.target)


@step
def model_finetuner_step(
    model: ClassifierMixin, dataset: Tuple[np.ndarray, np.ndarray]
) -> Annotated[
    ClassifierMixin, ArtifactConfig(name="my_model", tags=["SVC", "trained"])
]:
    """Finetunes a given model on a given dataset."""
    model.fit(dataset[0], dataset[1])
    return model


@pipeline
def model_finetuning_pipeline(
    dataset_version: Optional[str] = None,
    model_version: Optional[str] = None,
):
    # Either load a previous version of "my_dataset" or create a new one
    if dataset_version:
        dataset = ExternalArtifact(name="my_dataset", version=dataset_version)
    else:
        dataset = versioned_data_loader_step()

    # Load the model to finetune
    # If no version is specified, the latest version of "my_model" is used
    model = ExternalArtifact(name="my_model", version=model_version)

    # Finetune the model
    # This automatically creates a new version of "my_model"
    model_finetuner_step(model=model, dataset=dataset)


def main():
    # Save an untrained model as first version of "my_model"
    untrained_model = SVC(gamma=0.001)
    zenml.save_artifact(
        untrained_model, name="my_model", version="1", tags=["SVC", "untrained"]
    )

    # Create a first version of "my_dataset" and train the model on it
    model_finetuning_pipeline()

    # Finetune the latest model on an older version of the dataset
    model_finetuning_pipeline(dataset_version="1")

    # Run inference with the latest model on an older version of the dataset
    latest_trained_model = zenml.load_artifact("my_model")
    old_dataset = zenml.load_artifact("my_dataset", version="1")
    latest_trained_model.predict(old_dataset[0])

if __name__ == "__main__":
    main()

```

This would create the following pipeline run DAGs:

#### Run 1:

Screenshot 2023-11-17 at 11 12 54

#### Run 2:

Screenshot 2023-11-17 at 11 12 23

### Breaking Changes
- `ExternalArtifact` no longer supports `artifact_name`, `model_name`, `model_version`, `model_artifact_name`, and `model_artifact_version` arguments
- Moved `zenml.model.artifact_config.py` to `zenml.artifacts.artifact_config.py` 
- `ArtifactConfig` no longer supports `artifact_name` and `overwrite` args
- Deleted `ModelArtifactConfig` and `DeploymentArtifactConfig`
- Deleted `zenml.model.link_output_to_model.py`
-`log_artifact_metadata()` now expects a completely different set of arguments
- Removed package level imports in `zenml.model`, i.e., `from zenml.model import ...` no longer works
- `ModelVersionArtifact` no longer supports `name`, `link_version`, `pipeline_name`, and `step_name` arguments
- `ModelVersionPipelineRun` no longer supports `name` argument
- Reworked API endpoints and client / zen store methods for listing model version artifact / pipeline run links
- `model_version.get_X_artifact()` functions no longer support `pipeline_name` and `step_name` args

## Pre-requisites

## Types of changes


fa9r:
Previous PR with open discussions: https://github.com/zenml-io/zenml/pull/1937

bcdurak (commented):
I still have some files to review, but I am making a checkpoint here. Will continue in the morning.

docs/book/user-guide/advanced-guide/artifact-management/artifact-versioning.md
```diff
@@ -0,0 +1,143 @@
...
+
+This will assign tags "my_tag_1" and "my_tag_2" to all artifacts created by
+this step, which can later be used to filter for artifacts in the dashboard.
+
+## Link artifacts to models
+
+`ArtifactConfig` contains several other properties that you can set to modify
+how artifacts are linked to models. 
+
+TODO: link to model control plane section
```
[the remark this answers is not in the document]
fa9r: Yeah thanks, we'll probably not have this docs section any time soon so I wrote some short placeholder text for the meanwhile 👍

docs/book/user-guide/advanced-guide/artifact-management/artifact-saving-loading.md
```diff
@@ -0,0 +1,128 @@
...
+import zenml
+
+x = np.ndarray([1, 2, 3])
+zenml.save_artifact(x, name="my_numpy_array")
+x = zenml.load_artifact("my_numpy_array")
+```
+
+{% hint style="info" %}
+It is also possible to use these functions inside your ZenML steps. However, it is usually cleaner to return the artifacts as outputs of your step to save them, or to use [External Artifacts](#external-artifacts) to load them instead.
+{% endhint %}
```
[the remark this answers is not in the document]
fa9r: Great catch!

src/zenml/__init__.py
```diff
@@ -32,19 +32,34 @@
 
 __path__ = extend_path(__path__, __name__)
 
+# Need to import zenml.models before zenml.config to avoid circular imports
```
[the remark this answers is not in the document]
fa9r: Couldn't come up with any better solution, but I'm open to suggestions

> importing the entire models module can lead to a chain of imports and significantly reduce the performance of our CLI

Models are also imported by the Client, which in turn is imported in nearly every CLI file, so my change shouldn't change anything there. Also, aren't models supposed to have light-weight top-level imports only?

src/zenml/models/model_base_model.py
```diff
@@ -33,28 +33,35 @@ class ModelBaseModel(BaseModel):
     license: Optional[str] = Field(
         title="The license model created under",
         max_length=TEXT_FIELD_MAX_LENGTH,
+        default=None,
```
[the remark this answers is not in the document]
fa9r: Functionally it's the same, however my IDE shows errors without the `default=None`, so I added those default to quite the errors

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -167,6 +188,8 @@ def to_model(self, hydrate: bool = False) -> ArtifactResponse:
 
         # Create the body and metadata of the model
         body = ArtifactResponseBody(
+            version=self.version_number or self.version,
+            has_custom_name=self.has_custom_name,
```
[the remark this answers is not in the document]
fa9r: Yeah good catch - how is it possible this didn't cause mypy or any tests to fail? 🤔

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -192,3 +217,17 @@ def to_model(self, hydrate: bool = False) -> ArtifactResponse:
             body=body,
             metadata=metadata,
         )
+
+    def update(self, artifact_update: ArtifactUpdate) -> "ArtifactSchema":
```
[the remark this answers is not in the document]
fa9r: I agree, but I think that should be addressed separately since tags are also used for other entities where the logic is the same

src/zenml/zen_stores/schemas/model_schemas.py
```diff
@@ -386,53 +381,44 @@ class ModelVersionArtifactSchema(NamedSchema, table=True):
...
     )
-    artifact_id: Optional[UUID] = build_foreign_key_field(
+    artifact_id: UUID = build_foreign_key_field(
         source=__tablename__,
         target=ArtifactSchema.__tablename__,
         source_column="artifact_id",
         target_column="id",
         ondelete="CASCADE",
-        nullable=True,
+        nullable=False,
```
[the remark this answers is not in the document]
fa9r: @avishniakov is there any scenario where this could have been `null` before?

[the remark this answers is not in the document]
fa9r: > this change is not fully covered in the migration script

It is, on line 88

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -54,10 +57,21 @@ class ArtifactSchema(NamedSchema, table=True):
     __tablename__ = "artifact"
 
     # Fields
+    has_custom_name: bool
+    version: str
```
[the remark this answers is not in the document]
fa9r: TLDR: No, we need both

- `version_number` is needed for efficient retrieval of the highest numbered version, which is needed for auto-increment versioning
- `version` as a str field is necessary for custom string versioning

src/zenml/model/__init__.py
```diff
@@ -12,23 +12,4 @@
...
-
-
-__all__ = [
-    "DataArtifactConfig",
-    "EndpointArtifactConfig",
-    "ModelArtifactConfig",
-    "ModelVersion",
-    "link_output_to_model",
-]
+"""Concepts related to the Model Control Plane feature."""
```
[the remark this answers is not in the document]
fa9r: Two reasons:
- Main reason: it leads to all sorts of cyclic imports that I didn't know how to resolve otherwise
- We expose `zenml.model.ModelVersion` now as a top-level ZenML import `from zenml import ModelVersion` and if `zenml.models.__init__` was not empty, this would cause unnecessary chain-imports

src/zenml/models/v2/core/artifact.py
```diff
@@ -286,6 +346,14 @@ class ArtifactFilter(WorkspaceScopedFilter):
         default=None,
         description="Name of the artifact",
     )
+    version: Optional[str] = Field(
```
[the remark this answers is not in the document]
fa9r: Agreed, but I think filtering by tags should be done in a separate PR since there are also other entities that have tags which could benefit from this

src/zenml/zen_stores/schemas/model_schemas.py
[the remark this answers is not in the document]
avishniakov: Create method relies on `ModelVersionArtifactBaseModel` with `artifact: UUID`, so technically (and also by common sense), I'd say: no, it is not possible. If it is not true - this record has to be removed as corrupted anyway.

src/zenml/zen_stores/schemas/model_schemas.py
```diff
@@ -515,15 +497,15 @@ class ModelVersionPipelineRunSchema(NamedSchema, table=True):
...
     )
-    pipeline_run_id: Optional[UUID] = build_foreign_key_field(
+    pipeline_run_id: UUID = build_foreign_key_field(
         source=__tablename__,
         target=PipelineRunSchema.__tablename__,
         source_column="run_id",
         target_column="id",
         ondelete="CASCADE",
-        nullable=True,
+        nullable=False,
```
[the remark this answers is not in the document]
avishniakov: Same here - request model is not nullable.

src/zenml/artifacts/external_artifact.py
```diff
@@ -40,29 +39,32 @@ class ExternalArtifact(ExternalArtifactConfiguration):
...
-        model_artifact_version: Version of a model artifact to search for.
+        name: Name of an artifact to search. If none of
+            `version`, `pipeline_run_name`, or `pipeline_name` are set, the
+            latest version of the artifact will be used.
+        version: Version of the artifact to search. Only used when `name` is
+            provided.
+        pipeline_run_name: Name of a pipeline run to search artifacts in. Only
+            used when `name` is provided.
+        pipeline_name: Name of a pipeline in which to search for the artifact.
+            Only used when `name` is provided.
```
avishniakov, line 67: Conceptually, I'm not getting why we are giving privilege to use pipeline objects as selectors, but not model versions? Pipeline run name to me is somewhat of the same level of granularity as a model version.

src/zenml/artifacts/external_artifact_config.py
```diff
@@ -39,14 +33,10 @@ class ExternalArtifactConfiguration(BaseModel):
     """
 
     id: Optional[UUID] = None
+    name: Optional[str] = None
+    version: Optional[str] = None
+    pipeline_run_name: Optional[str] = None
     pipeline_name: Optional[str] = None
```
avishniakov, line 39: Same question as above, `pipeline_run_name` seems similar level selector as `model_version`.

src/zenml/cli/model.py
```diff
@@ -559,57 +538,43 @@ def _print_artifacts_links_generic(
...
+        columns=["artifact", "created"],
     )
 
 
 @version.command(
     "data_artifacts",
     help="List data artifacts linked to a model version.",
 )
-@click.argument("model_name_or_id")
-@click.argument("model_version_name_or_number_or_id", default="0")
```
avishniakov, line 595: Don't like this change in the way it is done now. Can we move this up to model group? Version group assume that version is an argument in all methods.

src/zenml/cli/model.py
```diff
@@ -619,25 +584,25 @@ def list_model_version_data_artifacts(
     "model_artifacts",
     help="List model artifacts linked to a model version.",
 )
-@click.argument("model_name_or_id")
-@click.argument("model_version_name_or_number_or_id", default="0")
```
avishniakov, line 623: Don't like this change in the way it is done now. Can we move this up to model group? Version group assume that version is an argument in all methods.

src/zenml/cli/model.py
```diff
@@ -647,25 +612,25 @@ def list_model_version_model_artifacts(
     "endpoint_artifacts",
     help="List endpoint artifacts linked to a model version.",
 )
-@click.argument("model_name_or_id")
-@click.argument("model_version_name_or_number_or_id", default="0")
```
avishniakov, line 651: Don't like this change in the way it is done now. Can we move this up to model group? Version group assume that version is an argument in all methods.

src/zenml/cli/model.py
```diff
@@ -675,30 +640,25 @@ def list_model_version_endpoint_artifacts(
     "runs",
     help="List pipeline runs of a model version.",
 )
-@click.argument("model_name_or_id")
-@click.argument("model_version_name_or_number_or_id", default="0")
```
avishniakov, line 679: Don't like this change in the way it is done now. Can we move this up to model group? Version group assume that version is an argument in all methods.

src/zenml/constants.py
```diff
@@ -222,6 +222,8 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 SERVICE_CONNECTOR_CLIENT = "/client"
 MODELS = "/models"
 MODEL_VERSIONS = "/model_versions"
+MODEL_VERSION_ARTIFACTS = "/model_version_artifacts"
+MODEL_VERSION_PIPELINE_RUNS = "/model_version_pipeline_runs"
```
avishniakov, line 226: @Cahllagerfeld @AlexejPenner are you aware of this and ok with it?

src/zenml/zen_server/routers/artifacts_endpoints.py
```diff
@@ -116,6 +117,29 @@ def get_artifact(
     return zen_store().get_artifact(artifact_id, hydrate=hydrate)
 
 
+@router.put(
```
avishniakov, line 120: @Cahllagerfeld @AlexejPenner please review changes below

src/zenml/zen_server/routers/model_versions_endpoints.py
```diff
@@ -155,15 +157,20 @@ def delete_model_version(
 # Model Version Artifacts
 ##########################
 
+model_version_artifacts_router = APIRouter(
```
avishniakov, line 160: @Cahllagerfeld @AlexejPenner please review changes below

tests/integration/functional/steps/test_external_artifact.py
```diff
@@ -38,171 +38,115 @@ def consumer(external_artifact: int, expected_value: int):
...
-@pytest.mark.parametrize(
-    "model_version_name,expected",
-    [[1, 42], ["1", 42], ["foo", 23]],
-    ids=[
-        "By model version number",
-        "By model version number as string",
-        "By model version name",
-    ],
-)
-def test_direct_consumption(model_version_name, expected):
```
avishniakov, line 194: Deletion of this also correlate with my question about removal of model version

avishniakov (changes requested):
I'm good with most of it. My major concern: why not allowing people to use ExternalArtifact in pair with model version instead of pipeline run? I might be misunderstanding something though.

Please, also document somehow API changes and make @Cahllagerfeld aware of it before merging.

Lastly, I'm getting right that it is still possible to do `Annotated[int,"my_name"]` without full-blown config object?

src/zenml/artifacts/external_artifact.py
fa9r: Fair enough - I just added it because we had it in the previous version but I'd be happy to take them out as well

src/zenml/cli/model.py
fa9r: It was already an optional argument before so conceptually nothing really changed

fa9r: but sure we can move it to model

avishniakov: I'm just upvoting consistency here, if I call `zenml model version` any command, I expect to be able to pass `MODEL_NAME MODEL_VERSION` as unnamed params. This is the only concern. Rest is all good.

src/zenml/artifacts/external_artifact.py
avishniakov: Let's kick them out, if not needed. But, if we want to merge this - don't change it now, just make a follow-up PR. It would be better.

src/zenml/cli/model.py
avishniakov: We can also do it in subsequent PR, tbh.

avishniakov (approved):
Please, take all my change requests as requests for a follow-up PR, since it is not that critical to get into another loop of aligning this PR.

src/zenml/artifacts/external_artifact.py
fa9r: Ok done 👍

fa9r:
@avishniakov I adjusted `ExternalArtifact` now to no longer have `pipeline_run` and `pipeline` args either, which feels indeed much cleaner.

@Cahllagerfeld and I also just aligned on the API changes and how we will make the latest backend changes available to him in general.

As for your last question, yes, `Annotated[int, "my_name"]` is still possible and is functionally equivalent to `Annotated[int, ArtifactConfig("my_name")]`. I think it still makes sense to keep the former as a short-cut syntax since the name is by far the most common thing users will need to configure.

src/zenml/cli/model.py
fa9r: Already done 👍

avishniakov:
> @avishniakov I adjusted `ExternalArtifact` now to no longer have `pipeline_run` and `pipeline` args either, which feels indeed much cleaner.
> 
> @Cahllagerfeld and I also just aligned on the API changes and how we will make the latest backend changes available to him in general.
> 
> As for your last question, yes, `Annotated[int, "my_name"]` is still possible and is functionally equivalent to `Annotated[int, ArtifactConfig("my_name")]`. I think it still makes sense to keep the former as a short-cut syntax since the name is by far the most common thing users will need to configure.

We discussed API for artifact links with @Cahllagerfeld and the conclusion is. that we need the following fields to be in the response model (non-hydrated one, for listing even): `name (from artifact), type (from artifact), user (from artifact), updated (from link table)`

src/zenml/models/model_models.py
```diff
@@ -484,12 +483,16 @@ class ModelVersionArtifactRequestModel(
...
 
+    artifact: UUID
+
 
 class ModelVersionArtifactResponseModel(
     ModelVersionArtifactBaseModel, WorkspaceScopedResponseModel
 ):
     """Model version link with artifact response model."""
 
+    artifact: ArtifactResponse
```
avishniakov, line 494: Seems that this model still lacks the update from the link, also the full artifact response might be a big thing, not worth pulling here. I would seek advice from @bcdurak here, but I would, maybe, limit to only needed fields instead (name, type, user details).
@Cahllagerfeld , what fits better or full response is fine here?

P.S. same applies to the `ModelVersionPipelineRunResponseModel`

fa9r: Full artifact only has `name`, `version`, `type`, and `uri`. I.e., only `uri` is redundant, and IMO this is much cleaner than adding `artifact_name`, `artifact_type` as separate fields into the link.

fa9r: User of the artifact will not be in the response, but the link itself already has a user (which will be the same in 99% of cases) so we can simply display that instead.

fa9r: I don't think it makes sense for @Cahllagerfeld to work on anything related to models before #2066

We already discussed that we'll build him a Docker image including all changes once this PR and #1999 are merged into #2066

avishniakov (approved):

src/zenml/models/v2/core/artifact.py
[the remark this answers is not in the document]
bcdurak: Before we resolve this it might be nice to create a ticket for it then.

bcdurak (approved):

PR #1945, opened 2023-10-13
Fixing the migration issues

bcdurak:
## Describe changes
I fixed the problems caused by the alembic migration.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/37835ce041d2_optimizing_database.py
```diff
@@ -202,13 +176,6 @@ def downgrade() -> None:
                 "parameters", sa.VARCHAR(length=16777215), nullable=False
             )
         )
-        batch_op.add_column(
```
fa9r, line 205: downgrade is not fully reflective of the upgrade changes

fa9r (changes requested):
Changes look good to me, however before we merge I'd suggest we:
- combine the two new migrations into one
- either implement the downgrade correctly or just raise a `NotImplementedError`

bcdurak:
@fa9r I implemented the suggested changes, merged two scripts into one, and put a `NonImplementedError` for the downgrade.

fa9r (commented):
죽을래?

fa9r (approved):

src/zenml/zen_stores/migrations/versions/6f707b385dc1_fix_model_artifacts.py
```diff
@@ -10,7 +10,7 @@
 
 # revision identifiers, used by Alembic.
 revision = "6f707b385dc1"
-down_revision = "0.45.0"
+down_revision = "7f603e583dd7"
 branch_labels = None
```
AlexejPenner, line 14: @bcdurak changing this in the upgrade tree seems very wrong to me, I think this is rewriting history in a way that might break things for people who are running on 0.45.0 now. Is this file somethign that was merged into develop after the 0.45.0 release?

AlexejPenner: ah went to investigate, all good with this.

PR #1954, opened 2023-10-17
Increase reuse of `ModelConfig`

avishniakov:
## Describe changes
This PR removes `ModelConfigModel`; all codes now use `ModelConfig` instead. `ModelConfig` now has a `model` and `model_version` to avoid multiple calls to DB while working inside the same context.
`ModelConfig` is still instantiated on every step due to how reading of configs from DB works currently - it gets a dict and converts to_model where applicable, but no extra calls to DB are done with the current architecture.

## Pre-requisites
- [x] I have based my new branch on `develop` and the open PR targets `develop`. If your branch wasn't based on develop read [Contribution guide on rebasing branch to develop](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md#-pull-requests-rebase-your-branch-on-develop).

## Types of changes


src/zenml/model/model_config.py
```diff
@@ -30,24 +38,182 @@
...
-    delete_new_version_on_failure: Whether to delete failed runs with new versions for later recovery from it.
-    """
+    name: str
+    license: Optional[str]
+    description: Optional[str]
+    audience: Optional[str]
+    use_cases: Optional[str]
+    limitations: Optional[str]
+    trade_offs: Optional[str]
+    ethic: Optional[str]
```
strickvl, line 51:
```suggestion
    ethics: Optional[str]
```

Probably should be this (and below in relevant places)

strickvl (changes requested):

src/zenml/model/model_config.py
avishniakov: Done

src/zenml/new/pipelines/pipeline.py
```diff
@@ -825,46 +816,52 @@ def get_new_version_requests(
...
+        elif deployment.pipeline_configuration.model_config is not None:
             logger.warning(
                 f"ModelConfig of pipeline `{self.name}` is overridden in all steps. "
             )
 
         self._validate_new_version_requests(new_versions_requested)
 
+        for other_model_config in other_model_configs:
+            other_model_config._validate_config_in_runtime()
+
```
fa9r, line 864: Nitpick: wouldn't it be cleaner to this on top where the configs are checked directly? I.e., is there a reason why do we need to save them in `other_model_configs` and do the validation at the end?

src/zenml/model/model_config.py
```diff
@@ -123,40 +296,42 @@ def _create_model_version(
...
 
         Returns:
             The model version based on configuration.
         """
-        if self._model_version is not None:
-            return self._model_version
+        if self.model_version is not None:
+            from zenml.models.model_models import ModelVersionResponseModel
+
+            return ModelVersionResponseModel(**dict(self.model_version))
```
fa9r, line 317: Isn't `self.model_version` already a `ModelVersionResponseModel`?

src/zenml/model/model_config.py
```diff
@@ -103,8 +274,10 @@ def _create_model_version(
...
```
fa9r, line 280: Same here, isn't `self.model_version` already a `ModelVersionResponseModel`?

src/zenml/model/model_config.py
```diff
@@ -57,15 +228,17 @@ def get_or_create_model(self) -> "ModelResponseModel":
         Returns:
             The model based on configuration.
         """
-        if self._model is not None:
-            return self._model
+        from zenml.models.model_models import ModelResponseModel
+
+        if self.model is not None:
+            return ModelResponseModel(**dict(self.model))
```
fa9r, line 234: Same here, isn't `self.model` already a `ModelResponseModel`?

src/zenml/model/model_config.py
```diff
@@ -30,24 +38,187 @@
...
+    create_new_model_version: bool = False
+    save_models_to_registry: bool = True
+    delete_new_version_on_failure: bool = True
 
-    _model: Optional["ModelResponseModel"] = PrivateAttr(default=None)
-    _model_version: Optional["ModelVersionResponseModel"] = PrivateAttr(
-        default=None
-    )
+    model: Optional[Any] = None
+    model_version: Optional[Any] = None
```
fa9r, line 60: Why are these `Any` types? (Probably related to my questions below)

src/zenml/model/model_config.py
```diff
@@ -30,24 +38,187 @@
...
+            ethics=ethics,
+            tags=tags,
+            version=version,
+            version_description=version_description,
+            create_new_model_version=create_new_model_version,
+            save_models_to_registry=save_models_to_registry,
+            delete_new_version_on_failure=delete_new_version_on_failure,
+            model=kwargs.get("model", None),
+            model_version=kwargs.get("model_version", None),
+            user_not_yet_warned=kwargs.get("user_not_yet_warned", True),
```
fa9r, line 120: Usually the cleaner way to do this in Pydantic is to have validators for each of these arguments instead of overriding the init

fa9r: Just saw you already have a root validator below, why don't you do this in the root validator instead? Then you don't need these 60LoC of init

src/zenml/model/model_config.py
```diff
@@ -30,24 +38,187 @@
...
+        delete_new_version_on_failure = values.get(
+            "delete_new_version_on_failure", True
+        )
+        user_not_yet_warned = values.get("user_not_yet_warned", True)
+        if not delete_new_version_on_failure and not create_new_model_version:
+            if user_not_yet_warned:
+                logger.warning(
+                    "Using `delete_new_version_on_failure=False` and `create_new_model_version=False` has no effect."
+                    "Setting `delete_new_version_on_failure` to `True`."
+                )
```
fa9r, line 153: I don't even think this needs a warning, it's not like this is a dangerous configuration, one arg is simply not used, so why would users need to care?

src/zenml/cli/model.py
```diff
@@ -468,24 +468,30 @@ def _print_artifacts_links_generic(
...
     model_version = Client().get_model_version(
         model_name_or_id=model_name_or_id,
-        model_version_name_or_number_or_id=model_version_name_or_number_or_id,
+        model_version_name_or_number_or_id=None
+        if model_version_name_or_number_or_id == "0"
+        else model_version_name_or_number_or_id,
+    )
+    type_ = (
+        "artifacts"
+        if only_artifacts
```
fa9r, line 477: Nitpick: "only_artifacts" doesn't really make sense semantically since the deployments and models are also artifacts

src/zenml/cli/model.py
```diff
@@ -515,7 +521,7 @@ def _print_artifacts_links_generic(
     help="List artifacts linked to a model version.",
 )
 @click.argument("model_name_or_id")
-@click.argument("model_version_name_or_number_or_id")
+@click.argument("model_version_name_or_number_or_id", default="0")
```
fa9r, line 524: This feels a bit weird, why can't we use `None` as default / latest value instead?

tests/unit/model/test_model_config_init.py
```diff
@@ -0,0 +1,66 @@
+from unittest.mock import patch
+
+import pytest
+
+from zenml.enums import ModelStages
+from zenml.model import ModelConfig
+
+
+@pytest.mark.parametrize(
```
fa9r, line 9: See comment above - why is it important that this configuration gives a warning?

fa9r (changes requested):
Nice, this feels much cleaner 🙌

Just have a few more minor suggestions, otherwise this looks good to merge 👍

src/zenml/cli/model.py
avishniakov: Cause it is strings - cli arguments and `-1` which I planned to use is treated as a parameter name, so 0 seemed a solid choice. Moreover - user will not specify it directly, it will be set by just skipping this arg.
`zenml model version artifacts my_model` -> will return artifacts for latest in `my_model`

src/zenml/model/model_config.py
avishniakov: Init is here for a reason: by this, I control which fields are user-facing. `model, model_version, user_not_yet_warned` are not user-facing and not expected to be used by users.
Why not `PrivateAttr`? They don't go into `dict()` while we store it in DB and the setup from DB is reused in many places (e.g. each step do it), so I want to keep track, if this object was instantiated from code (has `user_not_yet_warned=True` by skipping it from user) or from DB (has `user_not_yet_warned=False` coming from DB)

avishniakov: I'll redesign this

src/zenml/model/model_config.py
avishniakov: I'll redesign this

src/zenml/model/model_config.py
avishniakov: I'll redesign this

src/zenml/model/model_config.py
avishniakov: I'll redesign this

src/zenml/new/pipelines/pipeline.py
avishniakov: Yes, if I do it in the end, all model versions are already created.
Example:
```python
@step(model_config=ModelConfig(name="foo",create_new_version=True))
def _step():
    return 1

@pipeline(model_config=ModelConfig(name="foo"))
def _pipeline():
    _step()

_pipeline()
```
This snippet will fail, if I do validations of reading model configs, before creators.

src/zenml/model/model_config.py
avishniakov: I decided to get rid of this caching, as it produces unexpected results and always returns the version cached at the very beginning of the pipeline run.

src/zenml/model/model_config.py
avishniakov: I decided to get rid of this caching, as it produces unexpected results and always returns the version cached at the very beginning of the pipeline run.

src/zenml/model/model_config.py
avishniakov: I decided to get rid of this caching, as it produces unexpected results and always returns the version cached at the very beginning of the pipeline run.

src/zenml/model/model_config.py
avishniakov: Init is out along with overall redesign of this part.

tests/unit/model/test_model_config_init.py
avishniakov: removed

src/zenml/cli/model.py
fa9r: I don't get it, why can't it be `Optional[str]` instead? I.e., why would this not work?

```python
@click.argument("model_name_or_id")
@click.argument("model_version_name_or_number_or_id")
def list_model_version_artifacts(
    model_name_or_id: str,
    model_version_name_or_number_or_id: Optional[str] = None,
) -> None:
    ...

def _print_artifacts_links_generic(
    model_name_or_id: str,
    model_version_name_or_number_or_id: Optional[str] = None,
    ...
):
    ...
```

avishniakov: It is a click limitation, as I see this. If I set `default=None` it is treated as no default, making arg mandatory. Changes on the level of functions args are not effective at all - `@click.argument` is the king here.

fa9r: Ah, yes, `click.argument` means mandatory, for optional inputs you need to use `click.option`

strickvl (approved):
Looks great + lovely to have it well tested, too.

PR #1965, opened 2023-10-19
Fix DB initialization when using external authentication

schustmi:
## Describe changes

This PR fixes two issues regarding database initialization in the Helm chart when using external authentication:
- The init container previously wasn't configured to use external authentication and therefore wrongly created a default user and stack/componens.
- The Helm chart used `zenml status` to initialize the database. The `zenml status` command however also prints the active user (which falls back to the default user) which fails when using external authentication. This PR introduces a separate CLI command which migrates the DB with as little side-effects as possible.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
+++ b/src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
@@ -48,7 +48,7 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.zenml.image.repository }}:{{ .Values.zenml.image.tag | default .Chart.Version }}"
           imagePullPolicy: {{ .Values.zenml.image.pullPolicy }}
-          args: ["status"]
+          args: ["migrate-database"]
           command: ['zenml']
           env:
             {{- if .Values.zenml.debug }}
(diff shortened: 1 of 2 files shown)


strickvl (approved):

src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
```diff
@@ -61,12 +61,24 @@ spec:
               value: {{ .Values.zenml.defaultProject | quote }}
             - name: ZENML_DEFAULT_USER_NAME
               value: {{ .Values.zenml.defaultUsername | quote }}
-            {{- if .Values.zenml.database.url }}
```
stefannica, line 64: This should stay in to conditionally configure `ZENML_STORE_TYPE` to `sql`

stefannica (changes requested):

src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
schustmi: If I'm not mistaken, the entire initContainer is wrapped in a `{{- if .Values.zenml.database.url }}` which is why I removed this duplicate

stefannica: Oh, right :)

stefannica (approved):

PR #1969, opened 2023-10-20
fix: `BasePyTorchMaterliazer` -> `Materializer`

cameronraysmith:
`BasePyTorchMaterliazer` contains a typo, for example, at https://github.com/zenml-io/zenml/blob/f333d27099783da7da2b784f855c6b98a0bc8bcc/src/zenml/integrations/pytorch/materializers/base_pytorch_materializer.py#L28 

See this [search](https://github.com/search?q=repo%3Azenml-io%2Fzenml+BasePyTorchMaterliazer&type=code) for the incorrect form, which should become empty after merge

- [x] rename `BasePyTorchMaterliazer` to `BasePyTorchMaterializer`


strickvl (approved):
Thank you! Good catch. Looks good to me.

fa9r (approved):
Oh nice, good catch, thanks 👍

PR #1988, opened 2023-10-25
Server-Side Component Config Validation

fa9r:
## Describe changes
When creating or updating stack components through the dashboard, we now also validate that the component config matches the corresponding configuration class and that the component is configured correctly.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/stack/utils.py
+++ b/src/zenml/stack/utils.py
@@ -1,0 +1,148 @@
+#  Copyright (c) ZenML GmbH 2023. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at:
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Util functions for handling stacks, components, and flavors."""
+
+from typing import Any, Dict, Optional
+
+from zenml.client import Client
+from zenml.enums import StackComponentType, StoreType
+from zenml.logger import get_logger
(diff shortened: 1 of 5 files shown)


src/zenml/stack/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _warn_if_config_server_mismatch(
+    configuration: StackComponentConfig, zen_store: BaseZenStore
+) -> None:
+    """Warn if the configuration is mismatched with the ZenML server."""
+    if configuration.is_remote and zen_store.is_local_store():
+        if zen_store.type != StoreType.REST:
+            logger.warning(
+                "You are configuring a stack component that is running "
```
schustmi, line 75: When this is running on the server now, won't this always run into this log as the ZenStore running on the server will be local as it's SQLite/MySQL?

fa9r: Yeah good point 👍

strickvl (changes requested):
Can we add a test(s) for the updated client functionality? Otherwise looks good to me.

fa9r:
@strickvl I didn't add any Client functionality. Did you mean the util functions?

fa9r:
> You added the validation

That was already part of the client before, all I did is refactor it out into a util function that is now called by the endpoints as well

strickvl (approved):
LG from my side, pending a green test suite.

src/zenml/stack/utils.py
```diff
@@ -0,0 +1,152 @@
...
+        component_type=component_type,
+        zen_store=zen_store,
+    )
+    configuration = flavor_class.config_class(**configuration_dict)
+    if not configuration.is_valid:
+        raise ValueError(
+            f"Invalid stack component configuration. Please verify "
+            f"the configurations set for {component_type}."
+        )
+    if not zen_store:  # Doesn't make sense to warn inside a server.
```
schustmi, line 62: Instead of having this unclear `if` condition here, I suggest we don't call this warning here, but instead simply call the `_warn_if_config_server_mismatch` function from the client after the configuration was validated

fa9r: Yeah feels cleaner, agreed 👍

schustmi (approved):

safoinme (approved):

PR #1992, opened 2023-10-26
Fix Build Options of `GCPImageBuilder`

fa9r:
## Describe changes
Fixed build options format for `GCPImageBuilder` and also added stack validation before image building to ensure the image builder is in a compatible stack before trying to build.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
+++ b/src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
@@ -130,6 +130,9 @@ class GCPImageBuilder(BaseImageBuilder, GoogleCredentialsMixin):
             build_context=build_context,
             parent_path_directory_name="cloud-build-contexts",
         )
+        docker_build_options = {
+            "--" + key: value for key, value in docker_build_options.items()
+        }
         build = self._configure_cloud_build(
             image_name=image_name,
             cloud_build_context=cloud_build_context,
--- a/src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
+++ b/src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
@@ -181,10 +181,10 @@ class GCPImageBuilder(BaseImageBuilder, GoogleCredentialsMixin):
             if isinstance(value, list):
                 for val in value:
                     docker_build_args.extend([option, val])
-            elif value is None or isinstance(value, bool):
-                docker_build_args.append(option)
-            else:
-                docker_build_args.extend([key, value])
+            elif value is not None and not isinstance(value, bool):
+                docker_build_args.extend([option, value])
+            elif value is not False:
+                docker_build_args.extend([option])
 
         return cloudbuild_v1.Build(
             source=cloudbuild_v1.Source(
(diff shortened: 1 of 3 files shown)


src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
```diff
@@ -130,6 +130,9 @@ def build(
             build_context=build_context,
             parent_path_directory_name="cloud-build-contexts",
         )
+        docker_build_options = {
+            "--" + key: value for key, value in docker_build_options.items()
```
schustmi, line 134: Does this actually work? When I try running a local `docker build --rm True ...` this fails for me 🤔

While checking the code, I also saw that in `_configure_cloud_build` we create a list `docker_build_args` which converts the build options dict into a list, but that seems to be completely unused?

fa9r: I don't know why but yes https://console.cloud.google.com/cloud-build/builds;region=global/7f5ab3a5-2dd7-467e-aba4-200070db7639?project=zenml-core 😅

fa9r: > While checking the code, I also saw that in _configure_cloud_build we create a list docker_build_args which converts the build options dict into a list, but that seems to be completely unused?

Hahaha yes it is never used, instead it uses `*build_options`, which only passes in the keys, that's why my "fix" above made it just receive `["--pull", "--rm"]` without the `True` values 😄

src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
```diff
@@ -177,9 +177,12 @@ def _configure_cloud_build(
         # Convert the docker_build_options dictionary to a list of strings
         docker_build_args = []
         for key, value in build_options.items():
+            option = f"--{key}"
             if isinstance(value, list):
                 for val in value:
-                    docker_build_args.extend([key, val])
+                    docker_build_args.extend([option, val])
+            elif value is None or isinstance(value, bool):
+                docker_build_args.append(option)
```
schustmi, line 185: Now passing `rm=False` as an option would be translated to `--rm` right?

(I don't have any idea how to solve this in a nice way btw, seems impossible)

src/zenml/integrations/gcp/image_builders/gcp_image_builder.py
```diff
@@ -177,11 +177,14 @@ def _configure_cloud_build(
         # Convert the docker_build_options dictionary to a list of strings
         docker_build_args = []
         for key, value in build_options.items():
+            option = f"--{key}"
```
schustmi, line 180: So the current implementation now works for the two args that we currently provide which is good, I'm pretty sure it's not generic and will break in some other cases.

I guess we just go with it for now? I don't have any idea which works generically other than not passing those arguments here and instead have them configure it as something separate in the `DockerSettings`.

Maybe something like `DockerSettings.image_builder_build_args` or something, but maybe not necessary for now.

src/zenml/new/pipelines/pipeline.py
```diff
@@ -635,6 +635,7 @@ def _run(
                 )
 
             stack = Client().active_stack
+            stack.validate()
```
schustmi, line 638: This validation in already happening in `stack.prepare_pipeline_deployment` later in this method. I guess you want it here so we make sure the image builder doesn't break right? Can we somehow verify that at stack registration/update? I guess not as it won't be called when registering from the dashboard at least?

schustmi (changes requested):

src/zenml/new/pipelines/pipeline.py
fa9r: Yeah this is technically redundant but I added the call here to ensure we'll always validate before trying to run any pipelines on a stack

PR #1995, opened 2023-10-27
Adjust model endpoints and methods to conform to the rest of code

AlexejPenner:
## Describe changes

As a rule of thumb (besides workspaces) we do not have endpoints with 2 layers of unique resources. 

`/model_version/{id}` already uniquely identifies the model_version so
`/model/{id}/model_version{id}` is redundant.

Also

get `.../model/{id}/model_versions` as a list call did not work as the `{id}` was completely ignored, instead the `model_id` field of the filter model was a required field. This behaved differently than filters and list calls for all other resources. So I moved /model_versions into their own group

Here is the new API docs:
image

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/routers/models_endpoints.py
+++ b/src/zenml/zen_server/routers/models_endpoints.py
@@ -34,13 +34,14 @@ from zenml.models import (
     ModelResponseModel,
     ModelUpdateModel,
     ModelVersionArtifactFilterModel,
+    ModelVersionArtifactRequestModel,
     ModelVersionArtifactResponseModel,
     ModelVersionFilterModel,
     ModelVersionPipelineRunFilterModel,
+    ModelVersionPipelineRunRequestModel,
     ModelVersionPipelineRunResponseModel,
     ModelVersionResponseModel,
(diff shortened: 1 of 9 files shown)


src/zenml/zen_server/routers/models_endpoints.py
```diff
@@ -210,57 +212,51 @@ def get_model_version(
...
-@router.put(
-    "/{model_id}" + MODEL_VERSIONS + "/{model_version_id}",
+@model_versions_router.put(
+    "/{model_version_id}",
     response_model=ModelVersionResponseModel,
     responses={401: error_response, 404: error_response, 422: error_response},
 )
 @handle_exceptions
 def update_model_version(
-    model_version_id: UUID,
```
fa9r, line 227: Isn't this still needed?

AlexejPenner:
image
I have removed these endpoints, however, I think we also need to discuss if their ORM and transport Models should be workspace scoped. The Model Version already uniquely locks them into a workspace scope

avishniakov:
Closing in favor of #2003

PR #1999, opened 2023-10-30
Cloud RBAC

schustmi:
## Describe changes

This PR introduces changes to enable RBAC for all server endpoints.

### RBAC
To check whether a user is allowed to perform an action on a resource, we verify that either
- the user is the owner of the resource (e.g. created the `stack` that he wants to read/update/delete)
- the user has an assigned role that gives them permission to perform the action. This is done using any concrete implementation of the `zenml.zen_server.rbac.rbac_interface.RBACInterface` interface.

#### Dehydration
Some of our response models contain other response models in some of their attributes. In some cases, users might have access to read the parent model but not the child models contained within it. To address this, we "dehydrate" any response model and redact information from child models which the user is not allowed to access.

#### Service accounts
Service accounts are currently stored in the server database which makes them incompatible with the Cloud RBAC backend (which requires an entry in the cloud user table). For this reason, service accounts currently are excluded from the RBAC and have full permissions.

### Some additional changes that needed to be implemented
- Sharing of `stacks`, `stack components` and `service connectors` was removed. In case of overlapping entity names which are required to be unique, the owner id is appended as the suffix.
- Roles and teams were removed. This includes any kind of authorization on server endpoints when RBAC is disabled.
- The default stack and stack components are now created server-side. Additionally, there is just a single default stack per server and not one per user. This was necessary as some users might not have permissions to even create a stack, but the client code requires a working active stack in lots of places.
- To make caching less error-prone when using a shared local (e.g. `default`) stack, the `client ID` is not included in the cache key for local artifact stores.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/base_zen_store.py
```diff
@@ -309,37 +299,19 @@ def get_default_store_config(path: str) -> StoreConfiguration:
...
             except KeyError:
-                self._create_default_stack(
-                    workspace_name_or_id=default_workspace.id,
-                    user_name_or_id=default_user.id,
-                )
+                self._create_default_user()
+
+        self._get_or_create_default_stack(
+            workspace=default_workspace,
+        )
```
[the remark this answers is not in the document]
stefannica: I guess this is true for the default stack belonging to the default workspace, but this doesn't apply to other workspaces. Yes, we don't officially allow/support multiple workspaces ATM, but I can see why it can't hurt to create those default stacks in the migration script.

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -309,37 +299,19 @@ def get_default_store_config(path: str) -> StoreConfiguration:
...
-                )
+                _ = self._default_user
             except KeyError:
-                self._create_default_stack(
-                    workspace_name_or_id=default_workspace.id,
-                    user_name_or_id=default_user.id,
-                )
+                self._create_default_user()
+
+        self._get_or_create_default_stack(
```
stefannica, line 312: Given that the default workspace and stack are created during the store DB initialization which happens at server startup (or when the local DB is initialized on local deployment), I think we should only call `_get_or_create_default_stack` and `_get_or_create_default_workspace` here in this initialization block, but everywhere else we should call `_get_default_stack` and `_get_default_user` and just fail if they're not present. The reason is that I'm afraid to have these created ad-hoc past the point where the store is initialized now. This is especially relevant if the client calls these instead of the server.

I'm talking here mostly about the `validate_active_config` method which should no longer create default entities if they're not already there.

stefannica: I would even go as far as moving the creation of the default stack/user/workspace to the SQL zen store, where they actually belong.

stefannica: This is all just optional BTW, it works fine as it is.

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -826,15 +702,13 @@ def _default_workspace(self) -> WorkspaceResponse:
...
+    @abstractmethod
     def _create_default_workspace(self) -> WorkspaceResponse:
         """Creates a default workspace.
 
         Returns:
             The default workspace.
         """
-        workspace_name = self._default_workspace_name
-        logger.info(f"Creating default workspace '{workspace_name}' ...")
-        return self.create_workspace(WorkspaceRequest(name=workspace_name))
```
stefannica, line 837: This is what I meant earlier: it now also makes sense that the default stack (and maybe even the default user) are also only created in the SQL zen store.

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -470,6 +452,23 @@ def _initialize_database(self) -> None:
...
+            f"Unable to create default stack in workspace "
+            f"{workspace.name}."
+        )
+
+    def _create_default_workspace(self) -> WorkspaceResponse:
+        raise RuntimeError(
+            f"Unable to create default workspace "
+            f"{self._default_workspace_name}."
+        )
+
```
stefannica, line 471: Instead of using an abstract approach, maybe you could only call these in the SQL zen store DB initialization to begin with.

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -499,9 +482,7 @@ def api_token(
     pipeline_id: Optional[UUID] = None,
     schedule_id: Optional[UUID] = None,
     expires_minutes: Optional[int] = None,
-    auth_context: AuthContext = Security(
-        authorize, scopes=[PermissionType.WRITE]
-    ),
+    auth_context: AuthContext = Security(authorize),
```
stefannica, line 485: This endpoint can be used by any user or service account regardless of their permissions to generate an API token. I guess it's harmless, because the user can already authenticate to the server to be even allowed to call the endpoint, but just wanted to point it out.

src/zenml/zen_server/routers/flavors_endpoints.py
```diff
@@ -102,60 +113,50 @@ def get_flavor(
...
     """
-    if flavor.user != auth_context.user.id:
-        raise IllegalOperationError(
-            "Creating flavors for a user other than yourself "
-            "is not supported."
-        )
-
-    created_flavor = zen_store().create_flavor(
-        flavor=flavor,
+    return verify_permissions_and_create_entity(
```
stefannica, line 126: `FlavorRequest` doesn't inherit from `UserScopedRequest`. This means that anyone can theoretically create a flavor belonging to another user. Theoretically, it could inherit from `WorkspaceScopedRequest`, but it can't because the user (owner) can be Null in case of the built-in flavors. This looks a lot like the default stack/components use-case, where you used `server_owned_request_model` and an internal request model, maybe the same can be applied here (i.e. `register_builtin_flavors` could use an internal flavor request model).

src/zenml/zen_server/routers/flavors_endpoints.py
```diff
@@ -102,60 +113,50 @@ def get_flavor(
...
-        flavor_update: Team update.
+        flavor_id: ID of the flavor to update.
+        flavor_update: Flavor update.
 
     Returns:
         The updated flavor.
     """
-    return zen_store().update_flavor(
-        flavor_id=flavor_id, flavor_update=flavor_update
+    return verify_permissions_and_update_entity(
```
stefannica, line 155: Same thing here: the flavor model doesn't inherit from `UserScopedResponseModel` so the owner won't be recognized by these utility functions.

stefannica (changes requested):
Second round done, still have about a third to go through.

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -245,8 +228,8 @@ def generate_access_token(
     access_token = JWTToken(
         user_id=user_id,
         device_id=device.id if device else None,
+        permissions=[],
```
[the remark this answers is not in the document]
schustmi: Done 👍

src/zenml/zen_server/deploy/helm/values.yaml
```diff
@@ -347,6 +347,11 @@ zenml:
   # mounted as environment variables in the ZenML server container.
   secretEnvironment: {}
 
+  # Source pointing to a class implementing the RBAC interface defined by
+  # `zenml.zen_server.rbac_interface.RBACInterface`. If not specified,
+  # RBAC will not be enabled for this server.
+  rbacImplementationSource:
+
```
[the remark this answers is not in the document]
schustmi: Sure, done 👍

src/zenml/client.py
```diff
@@ -4898,7 +4248,6 @@ def list_service_connector_resources(
             connectors have access to.
         """
         return self.zen_store.list_service_connector_resources(
-            user_name_or_id=self.active_user.id,
```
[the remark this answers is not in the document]
schustmi: Yes exactly!

src/zenml/zen_stores/base_zen_store.py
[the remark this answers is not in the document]
schustmi: Yep exactly, if someone for some reason has created a different workspace they also get a default stack during the migration. Creating a new workspace however will not create a default stack, which is why I'm also not sure we should keep the logic in the migration here.

src/zenml/zen_stores/base_zen_store.py
schustmi: I've moved all the database initialization into the SQLZenStore as you suggested, much better :)

src/zenml/zen_stores/base_zen_store.py
schustmi: I've moved all the database initialization into the SQLZenStore as you suggested, much better :)

src/zenml/zen_stores/rest_zen_store.py
schustmi: I've moved all the database initialization into the SQLZenStore as you suggested, much better :)

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -309,37 +299,19 @@ def get_default_store_config(path: str) -> StoreConfiguration:
...
             try:
-                default_user = self._default_user
-            except KeyError:
-                default_user = self._create_default_user()
-            try:
-                self._get_default_stack(
-                    workspace_name_or_id=default_workspace.id,
-                    user_name_or_id=default_user.id,
-                )
+                _ = self._default_user
```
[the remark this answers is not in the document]
schustmi: Added that method for the `SQLZenStore`!

src/zenml/zen_server/auth.py
```diff
@@ -763,31 +724,15 @@ def oauth2_authentication(
     Raises:
         HTTPException: If the JWT token could not be authorized.
     """
-    if security_scopes.scopes:
```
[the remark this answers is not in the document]
schustmi: Done

src/zenml/zen_server/routers/flavors_endpoints.py
schustmi: I think it can't inherit from `WorkspaceScopedRequest`/`WorkspaceScopedResponse` as our builtin/integration flavors are registered without a workspace ID so we don't duplicate them in the DB in case someone creates multiple workspaces.

I like your idea of going with the `server_owned_request_model`, I'll try to inherit from the `UserScoped..` classes and see if that works!

src/zenml/zen_server/routers/flavors_endpoints.py
schustmi: Flavors are now user-scoped as you suggested, nice catch!

src/zenml/zen_server/rbac/utils.py
```diff
@@ -0,0 +1,569 @@
...
+        assert auth_context
+
+        resources = get_subresources_for_model(model)
+        permissions = rbac().check_permissions(
+            user=auth_context.user, resources=resources, action=Action.READ
+        )
+
+    dehydrated_fields = {}
+
+    for field_name in model.__fields__.keys():
```
stefannica, line 115: Same here: should you exclude private fields, just in case a model already uses or decides to use that ?

src/zenml/zen_server/rbac/utils.py
```diff
@@ -0,0 +1,569 @@
...
+        model: The original model.
+        keep_id: If `True`, the model ID will not be replaced.
+        keep_name: If `True`, the model name will not be replaced.
+
+    Returns:
+        The model with attribute values replaced by default values.
+    """
+    values = {}
+
+    for field_name, field in model.__fields__.items():
```
stefannica, line 241: Same here: should you exclude private fields, just in case a model already uses or decides to use that ?

src/zenml/zen_server/rbac/utils.py
```diff
@@ -0,0 +1,569 @@
...
+
+    Args:
+        model: The model for which to get all the resources.
+
+    Returns:
+        All resources of a model which need permission verification.
+    """
+    resources = set()
+
+    for field_name in model.__fields__.keys():
```
stefannica, line 532: Should you exclude private fields ?

src/zenml/models/v2/base/base.py
```diff
@@ -255,15 +256,35 @@ def get_body(self) -> AnyBody:
 
         Returns:
             The body field of the response.
+
+        Raises:
+            IllegalOperationError: If the user lacks permission to access the
+                entity represented by this response.
         """
+        if not self.body:
```
stefannica, line 264: This same error message is now raised if by some mistake, the body was not populated. Does it make sense to check if `permission_denied` is True and only then raise the permission exception if then, but raise some other exception if `permission_denied` is False ?

src/zenml/models/v2/base/base.py
```diff
@@ -255,15 +256,35 @@ def get_body(self) -> AnyBody:
...
+
+        Raises:
+            IllegalOperationError: If the user lacks permission to access this
+                entity represented by this response.
         """
+        if not self.body:
+            raise IllegalOperationError(
+                f"Missing permissions to access {type(self).__name__} with "
+                f"ID {self.id}."
+            )
```
stefannica, line 286: Same here: raise a different exception if `permission_denied` is set to False ?

tests/integration/functional/zen_stores/test_zen_store.py
```diff
@@ -3662,118 +3236,6 @@ def test_connector_name_update_fails_if_exists():
...
-                resource_types=["cat"],
-                is_shared=False,
-            ) as other_connector:
-                with pytest.raises(EntityExistsError):
-                    other_store.update_service_connector(
-                        other_connector.id,
-                        update=ServiceConnectorUpdate(is_shared=True),
-                    )
-
-
```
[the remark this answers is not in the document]
schustmi: Enabled both tests on REST stores

src/zenml/models/v2/base/filter.py
```diff
@@ -380,6 +385,67 @@ def sorting_params(self) -> Tuple[str, SorterOps]:
...
+        from sqlmodel import or_
+
+        if not self._rbac_configuration:
+            return query
+
+        expressions = []
+
+        for column_name, allowed_ids in self._rbac_configuration[1].items():
+            if allowed_ids is not None:
+                expression = getattr(table, column_name).in_(allowed_ids)
```
stefannica, line 428: I probably mentioned this before: this query can explode if a lot of UUIDs are passed in. I think you should put a limit on how many IDs can be passed in a single column `IN` query like this, either here or wherever this is called (in the RBAC utilities, or the RBAC implementation ?) and raise an error. If the limit is enforced at the source (e.g. whenever someone adds an ID to an existing resource type for a role), then we can prevent performance degradation on this side of things.

There are some ways to implement large `IN` queries without degrading performance, like creating temporary entries in table, but perhaps we can avoid that altogether.

stefannica (changes requested):
One more round...

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4537,53 +4166,42 @@ def get_service_connector_client(
 
     def list_service_connector_resources(
         self,
-        user_name_or_id: Union[str, UUID],
         workspace_name_or_id: Union[str, UUID],
         connector_type: Optional[str] = None,
         resource_type: Optional[str] = None,
         resource_id: Optional[str] = None,
+        filter_model: Optional[ServiceConnectorFilter] = None,
```
stefannica, line 4173: This will break the Liskov Substitution Principle because the interface signature is not the same. Maybe you should just add the filter to all methods, including the REST store and Client. I also think you can get rid of the other arguments and only rely on what's set in the filter. The exception is the `resource_id`, which is applied separately and not as part of the filter when you call `list_service_connectors`.

src/zenml/models/base_models.py
```diff
@@ -107,6 +118,36 @@ def get_analytics_metadata(self) -> Dict[str, Any]:
         metadata["entity_id"] = self.id
         return metadata
 
+    @property
+    def partial(self) -> bool:
```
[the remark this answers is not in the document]
schustmi: Yep this is an old relic, didn't remember that was still in. I've removed it

src/zenml/models/base_models.py
```diff
@@ -107,6 +118,36 @@ def get_analytics_metadata(self) -> Dict[str, Any]:
...
+            if isinstance(value, BaseResponseModel):
+                return value.partial
+            elif isinstance(value, Dict):
+                return any(_helper(v) for v in value.values())
+            elif isinstance(value, (List, Set, tuple)):
+                return any(_helper(v) for v in value)
+            else:
+                return False
+
+        for field_name in self.__fields__.keys():
```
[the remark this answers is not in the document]
schustmi: Attributes starting with `_` are not included in this dictionary:

```python

class A(BaseModel):
  some_field: str
  _some_private_field: int
 
print(A.__fields__)
{'some_field': ModelField(name='some_field', type=str, required=True)}
```

src/zenml/zen_server/rbac/utils.py
schustmi: See above, not included

src/zenml/zen_server/rbac/utils.py
schustmi: See above, not included

src/zenml/zen_server/rbac/utils.py
schustmi: See above, not included

src/zenml/models/v2/base/base.py
schustmi: Done

src/zenml/models/v2/base/base.py
schustmi: Done

src/zenml/models/v2/core/flavor.py
```diff
@@ -117,12 +121,9 @@ class FlavorUpdate(FlavorRequest):
 # ------------------ Response Model ------------------
 
 
-class FlavorResponseBody(BaseResponseBody):
+class FlavorResponseBody(UserScopedResponseBody):
```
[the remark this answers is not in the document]
schustmi: Yes, our builtin/integration flavors are not workspace scoped as that would just duplicate them in the DB.

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -1064,7 +1028,7 @@ def create_code_repository(
 @handle_exceptions
 def get_workspace_statistics(
     workspace_name_or_id: Union[str, UUID],
-    _: AuthContext = Security(authorize, scopes=[PermissionType.READ]),
+    _: AuthContext = Security(authorize),
```
stefannica, line 1031: More of a consistency issue: the stats should reflect the number of items shown in the various lists (stacks, components, pipelines, runs) for the current user. There's no filtering done here however, so the number of items will be >= than those visible to the active user.

stefannica (changes requested):
Last batch of comments. This was a pleasure to review ! Awesome job abstracting everything and avoiding code duplication !

src/zenml/zen_stores/sql_zen_store.py
schustmi: Are you sure it breaks it? It seems to me that adding an additional argument **with default value** to a method of a subclass would not break it, as the subclass can still be used in all places where the superclass can be used. `mypy` seems to agree and only complain if the argument does not have a default value.

I'm a little afraid of adding the filter model also for the rest case, as we would now send that as part of the body instead of query parameters which means potential changes for the frontend would be required?

src/zenml/zen_stores/schemas/pipeline_schemas.py
```diff
@@ -166,14 +166,5 @@ def update(self, pipeline_update: "PipelineUpdate") -> "PipelineSchema":
...
         """
-        if pipeline_update.name:
-            self.name = pipeline_update.name
-
-        if pipeline_update.docstring:
-            self.docstring = pipeline_update.docstring
-
-        if pipeline_update.spec:
-            self.spec = pipeline_update.spec.json(sort_keys=True)
-
```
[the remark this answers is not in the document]
schustmi: A `pipeline` is essentially a `pipeline version` which in my opinion should be an immutable object. Especially the `spec` and `docstring` are something that we parse from the user code when running the pipeline and not something that should be manually adjusted later (Especially modifying the `spec` might mess with future pipeline runs).

I guess we could argue whether the name should be mutable, in which case we'd have to add the line back in and also implement checks for `name` + `version` uniqueness in the `update_pipeline` method of the `SQLZenStore`.

src/zenml/models/v2/base/filter.py
schustmi: What would you say is a reasonable limit for this? I guess we could also limit the amount of ids per role, but assigning multiple roles for a user would again allow them to exceed this limit?

stefannica: at a maximum, we could limit 100 IDs per role and 100 roles per user, this way we don't exceed 1000 IDs per query

src/zenml/zen_stores/schemas/pipeline_schemas.py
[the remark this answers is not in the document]
stefannica: Okay, I'll leave this up to you.

src/zenml/zen_stores/sql_zen_store.py
stefannica: Fine by me if mypy isn't complaining.

src/zenml/zen_stores/base_zen_store.py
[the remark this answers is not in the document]
schustmi: I actually had to change this, creating a new workspace now automatically creates a default stack for that workspace. We still need it in the migration, as default stack creation only happens when the workspace is created and not for existing ones.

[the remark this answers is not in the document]
stefannica: I saw that, good idea !

src/zenml/zen_server/routers/secrets_endpoints.py
```diff
@@ -138,8 +156,11 @@ def update_secret(
             if key not in secret_update.values:
                 secret_update.values[key] = None
 
-    return zen_store().update_secret(
-        secret_id=secret_id, secret_update=secret_update
+    return verify_permissions_and_update_entity(
+        id=secret_id,
+        update_model=secret_update,
+        get_method=zen_store().get_secret,
+        update_method=zen_store().update_secret,
```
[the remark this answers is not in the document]
schustmi: Left this out for now as I wasn't sure how to handle permissions for schedules. They are currently essentially tied to a pipeline deployment which we could use to verify permissions for them. However, thinking ahead to the triggers shaping schedules might actually become a standalone powerful resource that could be used as a trigger/assigned to pipelines/...

For that reason, it probably makes sense to already introduce a separate `schedule` resource type, do you agree?

src/zenml/zen_server/routers/schedule_endpoints.py
```diff
@@ -50,7 +49,7 @@ def list_schedules(
         make_dependable(ScheduleFilter)
     ),
     hydrate: bool = False,
-    _: AuthContext = Security(authorize, scopes=[PermissionType.READ]),
+    _: AuthContext = Security(authorize),
```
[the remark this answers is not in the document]
schustmi: I left that out for now as I wasn't sure how to handle permissions for schedules. Currently they are pretty useless and in a 1-1 relationship with pipeline deployments which we could use for permission verification.

However, thinking ahead to the triggers shaping schedules might become a more useful/standalone entity that can be used as a trigger/attached to pipelines/...
For that reason it probably makes sense to already introduce them as a separate resource type, do you agree?

src/zenml/zen_server/routers/secrets_endpoints.py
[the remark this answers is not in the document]
schustmi: No there currently is no hierarchy of permissions implemented.

src/zenml/zen_server/routers/workspaces_endpoints.py
schustmi: Fixed, we can now pass filter models to the `count_...()` methods of the `SQLZenStore` which allows us to apply RBAC (and also any other) filters.

src/zenml/zen_server/routers/auth_endpoints.py
schustmi: The generated token will have the same permissions as the user that generated it, correct? In other words, when someone authenticates with this API token the user in the auth context is the same as the one that generated the token?

src/zenml/zen_server/rbac/utils.py
```diff
@@ -0,0 +1,569 @@
...
+        FlavorResponse: ResourceType.FLAVOR,
+        ServiceConnectorResponse: ResourceType.SERVICE_CONNECTOR,
+        ComponentResponse: ResourceType.STACK_COMPONENT,
+        StackResponse: ResourceType.STACK,
+        PipelineResponse: ResourceType.PIPELINE,
+        CodeRepositoryResponse: ResourceType.CODE_REPOSITORY,
+        SecretResponseModel: ResourceType.SECRET,
+        ModelResponseModel: ResourceType.MODEL,
+        ArtifactResponse: ResourceType.ARTIFACT,
+    }
```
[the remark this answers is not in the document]
schustmi: I completely agree, even stuff like the "surrogate model" could be handled on the model classes themselves instead of having util functions that potentially contain too much knowledge.

The reason I did not implement it like this because I would have to duplicate everything on both the old and new base model classes until we convert the remaining models. If it's okay with you, I'd create a ticket to refactor this once we got rid of the old models.

src/zenml/models/v2/base/filter.py
schustmi: There's at least one `0` missing or too much for this to make sense 🤔

schustmi: I like the idea though, nicer than limiting it here when applying it as that would just lead to wrong results or error which could have been avoided at role creation/assignment time.

stefannica: Right, a zero missing :smiley: 50 IDs per role and 20 roles per user should be enough, I think

schustmi: Alright, I'll add such a limitation in the cloud backend

stefannica: Doesn't need to be done in this PR though. Can be a follow-up.

src/zenml/zen_server/routers/secrets_endpoints.py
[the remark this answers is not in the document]
stefannica: Nice ! Yeah, that makes sense.

src/zenml/zen_server/rbac/utils.py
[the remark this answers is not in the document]
stefannica: Fine by me, of course.

src/zenml/zen_server/routers/schedule_endpoints.py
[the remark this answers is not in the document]
stefannica: No, let's leave that for later, you're right. Maybe there won't be a schedule resource, but a trigger resource and the schedule is just one of the forms a trigger can take.

src/zenml/zen_server/routers/workspaces_endpoints.py
stefannica: Love it ! This could even become one of the CRUD functions that we implement for every entity, so we'll have `count` in addition to create/read/list/update/delete. CRLCUD !!

src/zenml/zen_server/routers/auth_endpoints.py
stefannica: Yes, correct.

stefannica: The endpoint is currently only used to generate API tokens to run pipelines though. I guess we could condition it on the pipeline run CREATE action, for the lack of a better thing.

stefannica (approved):
Ready to land !

src/zenml/zen_server/routers/workspaces_endpoints.py
schustmi: Haha CRLCUD 😅

src/zenml/models/__init__.py
```diff
@@ -76,11 +76,6 @@
     WorkspaceScopedResponse,
     WorkspaceScopedResponseBody,
     WorkspaceScopedResponseMetadata,
-    ShareableRequest,
```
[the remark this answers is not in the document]
schustmi: All gone!

src/zenml/zen_server/rbac/endpoint_utils.py
```diff
@@ -0,0 +1,159 @@
...
+
+AnyRequestModel = TypeVar(
+    "AnyRequestModel", bound=Union[BaseRequestModel, BaseRequest]
+)
+AnyResponseModel = TypeVar(
+    "AnyResponseModel", bound=Union[BaseResponseModel, BaseResponse]  # type: ignore[type-arg]
+)
+AnyFilterModel = TypeVar("AnyFilterModel", bound=BaseFilter)
+AnyUpdateModel = TypeVar("AnyUpdateModel", bound=BaseModel)
+UUIDOrStr = TypeVar("UUIDOrStr", UUID, Union[UUID, str])
```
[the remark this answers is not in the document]
schustmi: Making `mypy` happy is the only reason 😄 Some of the methods are `UUID` and some of the are `Union[UUID, str]`, and just having this type var have `Union[UUID, str]` means that it thinks we pass a string to a method which only accepts UUIDs.

src/zenml/zen_server/rbac/models.py
```diff
@@ -0,0 +1,83 @@
...
+    STACK_COMPONENT = "stack_component"
+    PIPELINE = "pipeline"
+    CODE_REPOSITORY = "code_repository"
+    MODEL = "model"
+    MODEL_VERSION = "model_version"
+    SERVICE_CONNECTOR = "service_connector"
+    ARTIFACT = "artifact"
+    SECRET = "secret"
+    TAG = "tag"
+    SERVICE_ACCOUNT = "service_account"
```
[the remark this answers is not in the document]
schustmi: Yep exactly, those are only the resource types which require explicit permissions. Other models such as `StepRuns` are implicitly verified using an associated model, the `PipelineRun` in this example.

src/zenml/zen_server/rbac/endpoint_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    Args:
+        id: The ID of the entity to fetch.
+        get_method: The method to fetch the entity.
+        get_method_kwargs: Keyword arguments to pass to the get method.
+
+    Returns:
+        A model of the fetched entity.
+    """
+    model = get_method(id, **get_method_kwargs)
+    verify_permission_for_model(model, action=Action.READ)
```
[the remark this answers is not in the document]
schustmi: Dehydration only looks at sub-models of a model and potentially removes their body.
We want this in two steps instead for the following reason:
- If a user does not have permission to `READ` the model that they're trying to fetch, we return them a `403` response: This is covered by the first `verify_permission_for_model(...)` call
- After we've verified that we want to return something from the endpoint, we now remove all the info contained in the response model that the user is not allowed to see.

src/zenml/zen_server/rbac/endpoint_utils.py
```diff
@@ -0,0 +1,159 @@
...
+        get_method: The method to fetch the entity.
+        update_method: The method to update the entity.
+
+    Returns:
+        A model of the updated entity.
+    """
+    model = get_method(id)
+    verify_permission_for_model(model, action=Action.UPDATE)
+    updated_model = update_method(model.id, update_model)
+    return dehydrate_response_model(updated_model)
```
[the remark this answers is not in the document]
schustmi: See above

src/zenml/zen_server/routers/stacks_endpoints.py
```diff
@@ -111,9 +118,22 @@ def update_stack(
...
+    if stack_update.components:
+        updated_components = [
+            zen_store().get_stack_component(id)
+            for ids in stack_update.components.values()
+            for id in ids
+        ]
+
+        batch_verify_permissions_for_models(
+            updated_components, action=Action.READ
+        )
```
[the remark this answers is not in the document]
stefannica: I don't know, I think this needs to be evaluated on a case-by-case basis. Not all relationships are the same. For example, the relationship between the service connector and its secret shouldn't even be visible to the client (I have a ticket to fix this). The relationship between a pipeline build and its pipeline is different: you need to have read access to the pipeline to see its details to be able to incorporate it in a build (at least that's how it feels to me).

I think we can be lax about these things for now and leave them as they are, but create a ticket to evaluate them at a later time with feedback from users.

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -73,13 +80,23 @@ def list_run_steps(
...
+    )
+    step_run_filter_model.configure_rbac(
+        authenticated_user_id=auth_context.user.id,
+        pipeline_run_id=allowed_pipeline_run_ids,
+    )
+
+    page = zen_store().list_run_steps(
         step_run_filter_model=step_run_filter_model, hydrate=hydrate
     )
+    return dehydrate_page(page)
```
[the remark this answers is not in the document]
schustmi: I'll add this to the follow-up ticket that we refactor some of the things once we get converted everything to the new models, is that okay?

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -706,13 +666,12 @@ def create_deployment(
...
             f"not supported."
         )
-    if deployment.user != auth_context.user.id:
-        raise IllegalOperationError(
-            "Creating deployments for a user other than yourself "
-            "is not supported."
-        )
 
-    return zen_store().create_deployment(deployment=deployment)
+    return verify_permissions_and_create_entity(
```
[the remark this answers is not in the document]
schustmi: See the discussion above: I'd say we keep this pretty relaxed and then increase the permissions required for certain endpoints based on user feedback/our own experiences. Does that sound good to you?

src/zenml/models/v2/base/filter.py
schustmi: The `teams` we have in the cloud poses some additional questions: Do we include the roles assigned via teams in this limit?
- If yes, we'd have to check the role limit for each user of a team whenever we assign a role to the team. This also means single users of a team might block assigning a role to a team.
- If no, people could avoid the role limit by creating an arbitrary amount of teams and using those to assign the roles.

src/zenml/zen_server/rbac/models.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ResourceType(StrEnum):
+    """Resource types of the server API."""
+
+    STACK = "stack"
+    FLAVOR = "flavor"
+    STACK_COMPONENT = "stack_component"
+    PIPELINE = "pipeline"
+    CODE_REPOSITORY = "code_repository"
+    MODEL = "model"
```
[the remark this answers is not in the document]
schustmi: We thought that this model itself does not contain any sensitive information:
- Users can also get IDs and names for "permission denied models" as part of a dehydrated model. For example, when fetching a `PipelineRun` which ran on a stack that you don't have permission to read, the `PipelineRunResponse` will still contain a `StackResponse` without body/metadata but including the stack ID and name.
- Other models contained in this reponse (e.g. the `ArtifactResponse`) are protected and will be removed from the response if the user does not have permission to view them.

Overall, I think adding separate permissions for these "link models" is not worth the effort, especially as i think we both agree that they ideally shouldn't even have a separate endpoint.

src/zenml/zen_server/routers/steps_endpoints.py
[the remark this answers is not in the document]
schustmi: https://zenml.atlassian.net/jira/software/projects/CLOUD/boards/10/backlog?selectedIssue=CLOUD-473

bcdurak (approved):

PR #2011, opened 2023-11-02
Add warning when artifact store cannot be loaded

strickvl:
Added:

- a more verbose warning when trying to import an artifact store flavor without all dependencies installed
- a boolean flag to specify that you want artifacts loading using only the current stack's artifact store

## Pre-requisites

## Types of changes


src/zenml/utils/artifact_utils.py
```diff
@@ -114,11 +114,17 @@ def load_model_from_metadata(model_uri: str) -> Any:
...
+def load_artifact(
+    artifact: "ArtifactResponseModel",
+    use_current_stack_only: bool = False,
+) -> Any:
     """Load the given artifact into memory.
 
     Args:
         artifact: The artifact to load.
+        use_current_stack_only: If True, only the current stack will be used to
+            load the artifact. If False, the artifact store will be loaded from
```
schustmi, line 126:
```suggestion
            load the artifact. If False, the artifact will be loaded from
```

schustmi (changes requested):
I think I misremembered this function when talking about it on discord. I thought this was some kind of utility function which takes as input a `run_name`, `step_name` and `output_name` and then fetches/loads the artifact.

If this is really just loading a specific artifact then it seems like a user error to try loading it when the integrations aren't available. Maybe in that case my suggestion of having this boolean flag does not make much sense, and we should rather just go with the improved error message that you already implemented as part of this PR as well?

strickvl:
@schustmi the improved error message doesn't help with the fact that I might not want it to assume to use an artifact store other than the one in my current stack. (Agreed though that not having the right dependencies installed is just user error which we can make a bit nicer with a nicer error message).

strickvl:
@schustmi ok you're probably right. The confusion (when using it) came when getting an artifact via the `last_successful_run` (where potentially this might have been from a different artifact store and which (when getting the artifact) was calling `load_artifact` under the hood. This flag wouldn't have helped in that case. I'll remove it.

schustmi:
Yes, the flag should maybe be on the `last_succesful_run` method instead, where people can use it to fetch just the runs on the active stack/artifact store?
But then it's probably too far removed from the artifact loading for anyone to realize its use..

schustmi (approved):

stefannica (approved):

PR #2020, opened 2023-11-03
Add simple database migration testing

strickvl:
PLACEHOLDER WHILE TESTING SCRIPT ON GH RUNNERS

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,9 @@ jobs:
           python-version: "3.8"
 
       - name: Install current package as editable
-        run: pip install -e .
+        run: |
+          pip install -U pip
+          pip install -e .
 
       - name: Install mlstacks package
         run: pip install mlstacks
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,12 +47,10 @@ jobs:
     - name: Set up Python 3.9
       uses: actions/setup-python@v2
       with:
-        python-version: 3.9
+        python-version: 3.10
 
     - name: Test migrations across versions
-      run: |
-        chmod +x scripts/test_migrations.sh
-        scripts/test_migrations.sh
+      run: scripts/test_migrations.sh
 
   ubuntu-setup-and-test-python:
     if: github.event.pull_request.draft == false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,23 @@ jobs:
       - name: Check for broken dependencies
         run: pip check
 
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+
+    - name: Test migrations across versions
+      run: |
+        chmod +x scripts/test_migrations.sh
+        scripts/test_migrations.sh
+
   ubuntu-setup-and-test-python:
     if: github.event.pull_request.draft == false
     strategy:
--- a/scripts/test-migrations.sh
+++ b/scripts/test-migrations.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+function run_tests_for_version() {
+    local VERSION=$1
+
+    # Initialize zenml with the appropriate template
+    if zenml init --template supported > zenml init --template starter --template-with-defaults; then
+        echo "Used the supported template"
+    else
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -29,14 +29,38 @@ jobs:
...
+          pip install -U pip
+          pip install -e .
 
       - name: Install mlstacks package
         run: pip install mlstacks
 
       - name: Check for broken dependencies
         run: pip check
 
+  db-migration-testing:
```
schustmi, line 42: I'm guessing this should not run on every PR, but maybe only in the release flow?

scripts/test-migrations.sh
```diff
@@ -0,0 +1,42 @@
...
+
+    # Confirm DB works and is accessible
+    zenml pipeline runs list >> dump.txt && echo "Success listing pipeline runs on $VERSION"
+
+    cd ..
+    rm -rf test_starter template-starter
+}
+
+# List of versions to test
+VERSIONS=("0.39.1" "0.40.0" "0.40.3" "0.41.0" "0.43.0" "0.44.1" "0.44.3" "0.45.2" "0.45.3" "0.45.4" "0.45.5" "0.45.6")
```
schustmi, line 25: While this tests that migration works for existing releases, I think the more interesting case would be to consider if the current local files on the PR/branch also work on top of that.

schustmi (changes requested):

scripts/test-migrations.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+function run_tests_for_version() {
+    set -e  # Exit immediately if a command exits with a non-zero status
+    local VERSION=$1
+
+    # Initialize zenml with the appropriate template
+    git clone -b "release/0.43.0" https://github.com/zenml-io/template-starter
```
fa9r, line 8: shouldn't this use `VERSION` instead of 0.43.0?

fa9r (commented):
Good starting point, but as @schustmi noted I think the most important test is the migration from the latest release to the latest/dev/PR/... alembic version, which is currently missing in this PR

scripts/test-migrations.sh
strickvl: No, since we don't make a new release of the `template-starter` every time we do a ZenML release. The 0.43.0 release branch / tag is simply the latest working version of the `template-starter` template...

fa9r (approved):

schustmi (approved):
LGTM!

I have a few questions/follow-up ideas:
- Have you checked that this actually fails when adding a migration that fails on the current branch?
- This currently starts at version `0.39.1` which does not cover the database state of long-term users. I think we should extend that soon, as those are IMO the most likely to fail and hardest to test manually. (The recent DB migration issues were due to pipeline runs which were probably from `0.2x.x` releases.)
- Does the `0.43` template starter also work for the old `0.39.1` releases? Long-term we need a better way to handle this, probably manual scripts for each release?
- This only tests running with SQLite at the moment. I think it would be crucial to run the same script with a MySQL DB as soon as possible. This actually wouldn't require many modifications, e.g.
```bash
docker run --name mysql -d -p 3306:3306 -e MYSQL_ROOT_PASSWORD=password mysql:8.0
zenml connect --url mysql://127.0.0.1/zenml --username root --password password
```

avishniakov (approved):

strickvl:
@schustmi all your points are taken + I've created tickets for those extensions of the limited test that this PR represents. Fully agreed that we'll want to have separate scripts between specific versions. This limit test already revealed some issues with the older versions + unpinned dependencies (like `sqlmodel`).

scripts/test-migrations.sh
```diff
@@ -0,0 +1,78 @@
...
+# Test the version of the current branch
+set -e
+python3 -m venv ".venv-current-branch"
+source ".venv-current-branch/bin/activate"
+
+pip3 install -U pip setuptools wheel
+pip install -e ".[templates,server]"
+pip3 install importlib_metadata
+
+run_tests_for_version current_branch
```
schustmi, line 76: Maybe I'm missing something but this doesn't seem to be defined?

schustmi (changes requested):

fa9r (approved):

scripts/test-migrations.sh
strickvl: It doesn't actually change the functionality. It just gets passed to the logged / `echo` statements. Installations are separately handled, so it's just so that we can track which version is being tested in the logged output.

avishniakov (approved):

PR #2021, opened 2023-11-03
Move db-migration to `job` instead of `init-container` to allow replicas

safoinme:
## Describe changes
This PR implements a pivotal change in our Helm chart deployment process by replacing the existing init container for database migration with a dedicated Helm Job. To ensure timely execution of database migration, Helm hooks with `pre-install` and `pre-upgrade` annotations are applied. This enhancement resolves migration issues arising when multiple replicas are deployed, guaranteeing database consistency and aligning the migration Job's execution precisely before the deployment. Extensive testing confirms the smooth operation of this new approach, both with single and multiple replicas in a Kubernetes cluster. The PR also adheres to Helm best practices by setting `helm.sh/hook-delete-policy` to `hook-succeeded`. Reviewers, please verify the code changes and execution order during the Helm installation and upgrade.

## Pre-requisites

## Types of changes


src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
```diff
@@ -0,0 +1,141 @@
...
+kind: Job
+metadata:
+  name: {{ include "zenml.fullname" . }}-db-migration
+  labels:
+    {{- include "zenml.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-1"
+    "helm.sh/hook-delete-policy": hook-succeeded
+{{- if .Values.zenml.database.url }}
```
stefannica, line 11: I think this condition should guard the entire job definition instead of just its `spec`.

stefannica (changes requested):
This is an amazing solution ! So good, in fact, that I think we might even be able to use something similar to implement automated DB downgrades (e.g. for rollbacks or even zenml server downgrades), which is something that would not have been previously possible with the init container.

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
```diff
@@ -0,0 +1,141 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "zenml.fullname" . }}-db-migration
+  labels:
+    {{- include "zenml.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-1"
+    "helm.sh/hook-delete-policy": hook-succeeded
```
stefannica, line 10:
```suggestion
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
This way, previous invocations of jobs that have failed are also cleaned up when the helm chart is re-applied.

stefannica:
> This is an amazing solution ! So good, in fact, that I think we might even be able to use something similar to implement automated DB downgrades (e.g. for rollbacks or even zenml server downgrades), which is something that would not have been previously possible with the init container.

I went a bit overboard here and forgot the real reason why we can't implement DB downgrades is that we need the new version image during rollback, which isn't possible ATM.

stefannica:
> > This is an amazing solution ! So good, in fact, that I think we might even be able to use something similar to implement automated DB downgrades (e.g. for rollbacks or even zenml server downgrades), which is something that would not have been previously possible with the init container.
> 
> I went a bit overboard here and forgot the real reason why we can't implement DB downgrades is that we need the new version image during rollback, which isn't possible ATM.

I just got an idea of how we might be able to implement this after all:

* always use `latest` as the container image used for the DB migration job
* pass the target zenml version as a parameter (command line argument) to the job
* the job figures out what the current DB version is, compares it to the one received as a parameter and performs an upgrade/downgrade as the case may be

stefannica (approved):

strickvl (approved):

PR #2023, opened 2023-11-06
Make error message nicer for when two artifacts that share a prefix are found

strickvl:
Addresses this situation:
CleanShot 2023-11-06 at 11 39 28@2x

with nicer error messages.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -5391,11 +5391,15 @@ class Client(metaclass=ClientMetaClass):
 
         # If more than one entity with the same name is found, raise an error.
         entity_label = get_method.__name__.replace("get_", "") + "s"
+        formatted_entity_items = [
+            f"- {item.name}: (id: {item.id}, flavor: {item.flavor})\n"
+            for item in entity.items
+        ]
         raise ZenKeyError(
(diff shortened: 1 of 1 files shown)


src/zenml/client.py
```diff
@@ -5391,11 +5391,15 @@ def _get_entity_by_id_or_name_or_prefix(
 
         # If more than one entity with the same name is found, raise an error.
         entity_label = get_method.__name__.replace("get_", "") + "s"
+        formatted_entity_items = [
+            f"- {item.name}: (id: {item.id}, flavor: {item.flavor})\n"
```
fa9r, line 5395: This won't work since only stack components have a `.flavor` field, but this util function is used for many other entities as well

fa9r (changes requested):

src/zenml/client.py
strickvl: Updated to address that. Do you think we should conditionally output that if they are stack components? Or anything else you'd think we might want to include in the output here that's present for all entities?

fa9r (approved):
Good enough ;)

avishniakov (approved):

PR #2030, opened 2023-11-07
Remove `create_new_model_version` arg of `ModelConfig`

avishniakov:
## Describe changes
I simplified the `ModelConfig` interface to avoid explicitly passing around `create_new_model_version`.
New behavior expects that the user keeps `version==None` if the new unnamed version is needed, otherwise model version will be retrieved/created based on the `version` argument.
Also new `ModelStages.LATEST` was introduced in order to give the ability to fetch the LATEST model version.

**[BREAKING CHANGE]** Old configurations using `create_new_model_version` might lead to unpredicted results, but the change needed is simple:
```python
ModelConfig(name="foo",create_new_model_version=True) -> ModelConfig(name="foo")
ModelConfig(name="foo", version="bar", create_new_model_version=True) -> ModelConfig(name="foo", version="bar")
ModelConfig(name="foo", version="bar", create_new_model_version=False) -> ModelConfig(name="foo", version="bar")
ModelConfig(name="foo", version=None, create_new_model_version=False) -> ModelConfig(name="foo", version=ModelStages.LATEST)
```

**[BEHAVIOR CHANGE]** If a signature like `ModelConfig(name="foo", version="bar", create_new_model_version=True)` was used, it prevented 2 parallel pipelines from running pointing to same model version, this is not the case anymore for named versions.

Tests updates mostly relate to signature updates, but also some tests were removed as no longer valid ones.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/model/model_config.py
+++ b/src/zenml/model/model_config.py
@@ -128,12 +128,11 @@ class ModelConfig(BaseModel):
                             "This run also operates with unnamed model version - "
                             "new run will be stopped to prevent unexpected behavior."
                         )
-
                 if self.delete_new_version_on_failure:
                     raise RuntimeError(
                         f"Cannot create version `{self.version}` "
                         f"for model `{self.name}` since it already exists. "
-                        "This could happen for unforseen reasons (e.g. unexpected "
+                        "This could happen for unforeseen reasons (e.g. unexpected "
                         "intteruption of previous pipeline run flow).\n"
                         "If you would like to remove the staling version use "
                         "following CLI command:\n"
@@ -241,8 +240,8 @@ class ModelConfig(BaseModel):
             model_version = zenml_client.create_model_version(
                 model_version=mv_request
             )
-            logger.info(f"New model version `{self.version}` was created.")
             self.was_created_in_this_run = True
+            logger.info(f"New model version `{self.version}` was created.")
         return model_version
 
     def _merge(self, model_config: "ModelConfig") -> None:
--- a/src/zenml/model/model_config.py
+++ b/src/zenml/model/model_config.py
@@ -132,7 +132,12 @@ class ModelConfig(BaseModel):
                 if self.delete_new_version_on_failure:
                     raise RuntimeError(
                         f"Cannot create version `{self.version}` "
-                        f"for model `{self.name}` since it already exists"
+                        f"for model `{self.name}` since it already exists. "
+                        "This could happen for unforseen reasons (e.g. unexpected "
+                        "intteruption of previous pipeline run flow).\n"
+                        "If you would like to remove the staling version use "
+                        "following CLI command:\n"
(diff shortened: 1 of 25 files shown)


src/zenml/model/model_config.py
```diff
@@ -151,18 +118,25 @@ def _validate_config_in_runtime(self) -> None:
...
-            if self.create_new_model_version:
+            if self.version is None or self.version == RUNNING_MODEL_VERSION:
+                self.version = RUNNING_MODEL_VERSION
                 for run_name, run in model_version.pipeline_runs.items():
                     if run.status == ExecutionStatus.RUNNING:
                         raise RuntimeError(
-                            f"New model version was requested, but pipeline run `{run_name}` "
-                            f"is still running with version `{model_version.name}`."
+                            f"New unnamed model version was requested, "
+                            f"but pipeline run `{run_name}` have not finished yet. "
```
fa9r, line 127:
```suggestion
                            f"but pipeline run `{run_name}` has not finished yet. "
```

src/zenml/model/model_config.py
```diff
@@ -151,18 +118,25 @@ def _validate_config_in_runtime(self) -> None:
...
+                            "This run also operates with unnamed model version - "
+                            "new run will be stopped to prevent unexpected behavior."
```
fa9r, line 129: I don't fully get the error message - would be better if you explain what the user is supposed to do now? Rename his version? Or wait for this other run to complete?

src/zenml/model/model_config.py
```diff
@@ -307,3 +260,22 @@ def _merge(self, model_config: "ModelConfig") -> None:
         self.delete_new_version_on_failure &= (
             model_config.delete_new_version_on_failure
         )
+
+    def __hash__(self) -> int:
```
fa9r, line 264: what's this used for?

fa9r (changes requested):
The main change looks good but I'm not a fan of the changes to the client, models, and zenstore methods. For these, fetching the latest version by default made sense and was more convenient to use with the version arg being optional. I'd suggest reverting these changes as indicated in the comments below, otherwise LGTM.

src/zenml/model/model_config.py
avishniakov: [Here](https://github.com/zenml-io/zenml/blob/07b93568230a0eb1ca6ba6a75d67448b729bc869/src/zenml/new/pipelines/pipeline.py#L851)

src/zenml/model/model_config.py
avishniakov: reworked

avishniakov:
@fa9r , tests passed, if you don't see any more changes needed here, I would love to merge this to proceed with renaming ticket.

fa9r (approved):
LGTM

PR #2031, opened 2023-11-07
Nightly builds

strickvl:
These changes set up the following:
- a scheduled run at midnight every night
- this first runs the unit tests on Ubuntu to prevent pushing something truly broken
- then it pushes the package as `zenml-nightly` with the version e.g. 0.46.0-2023-11-07
- it then pushes all the docker images (server and Python 3.10) with the `nightly` tag

## Pre-requisites

## Types of changes


.github/workflows/publish_to_pypi_nightly.yml
```diff
@@ -0,0 +1,62 @@
...
+
+              # Update the version to the nightly version
+              poetry version $NIGHTLY_VERSION
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Install Poetry
```
schustmi, line 48: How can the previous steps even work if poetry is installed afterwards? 🤔

.github/workflows/publish_docker_image_nightly.yml
```diff
@@ -0,0 +1,36 @@
...
+          service_account_email: ${{ secrets.GCP_CLOUDBUILD_EMAIL }}
+          service_account_key: ${{ secrets.GCP_CLOUDBUILD_KEY }}
+          project_id: ${{ secrets.GCP_CLOUDBUILD_PROJECT }}
+
+      # Cloudbuild
+      - name: Build docker images
+        run: |-
+          gcloud builds submit \
+            --quiet \
+            --config=release-cloudbuild-nightly.yaml \
```
schustmi, line 35: Can we reuse the other workflow and make the config configurable? Or are there more changes in this file which complicate things?

avishniakov (approved):
Looks good, I'll trust it's working 👍🏼

release-cloudbuild-nightly.yaml
```diff
@@ -0,0 +1,138 @@
+steps:
```
schustmi, line 1: I'm wondering in general whether we need to release/build Docker images for all python versions for these nightly builds?

Is the idea behind it to make sure our release workflows run without issues, or are we expecting people to actively use these releases in environments in which they're restricted to a specific python version?

schustmi (changes requested):

release-cloudbuild-nightly.yaml
strickvl: The main use case is internal, tbh, for demos. But it is also part of our reworking of the test process in general. I'd be fine only pushing one tagged image for now @ Python 3.10 or something. We can always add more if people ask for them.

.github/workflows/publish_docker_image_nightly.yml
strickvl: You're right. I've refactored it away to take an argument that specifies the config file. That was the only thing that distinguishes the two workflows.

.github/workflows/nightly_build.yml
```diff
@@ -0,0 +1,36 @@
...
+        runs-on: ubuntu-latest
+        needs: publish-python-package
+        steps:
+          - name: Sleep for 4 minutes
+            run: sleep 240
+            shell: bash
+
+    publish-docker-image:
+        needs: wait-for-package-release
+        uses: ./.github/workflows/publish_docker_image_nightly.yml
```
schustmi, line 33: This should not have the nightly suffix I think?

schustmi: I think we also need to pass the normal release config in the existing release workflow?

schustmi (changes requested):

.github/workflows/nightly_build.yml
strickvl: No need for the normal release config to be passed in for the existing release workflow as it's specified explicitly as the default in `publish_docker_image.yml`. But the nightly suffix can be deleted, yes.

stefannica (approved):
LGTM. We might need to include other python versions when/if people start asking for them.

schustmi (approved):

PR #2036, opened 2023-11-09
Create tags table

avishniakov:
## Describe changes
I implemented Tags and TagsResources tables as follows:
image

The connection between Models (as a first entity) and linking table TagResource is soft and controlled by SQLModel, shown only for demo purposes - it is not a real FK on the DB level.

Tags are equipped with the full scope of CLI commands:
```bash
zenml tag --help
Usage: zenml tag [OPTIONS] COMMAND [ARGS]...

  Interact with tags.

Options:
  --help  Show this message and exit.

Commands:
  delete    Delete an existing tag.
  list      List tags with filter.
  register  Register a new tag.
  update    Update an existing tag.
  ```

**How does the connection between entities and tags work?**
TagResources
```python
    resource_id: UUID
    resource_type: int
    model: List["ModelSchema"] = Relationship(
        back_populates="tags",
        sa_relationship_kwargs=dict(
            primaryjoin=f"and_(TagResourceSchema.resource_type=={TaggableResourceTypes.MODEL.value}, foreign(TagResourceSchema.resource_id)==ModelSchema.id)",
        ),
    )
```
Models (same for any other entity)
```python
tags: List["TagResourceSchema"] = Relationship(
        back_populates="model",
        sa_relationship_kwargs=dict(
            primaryjoin=f"and_(TagResourceSchema.resource_type=={TaggableResourceTypes.MODEL.value}, foreign(TagResourceSchema.resource_id)==ModelSchema.id)",
            cascade="delete",
        ),
    )
```
**What are the needed changes for adding other Taggable entities?**
1. Extend `zenml.enums.TaggableResourceTypes`
2. Add relationship between  tables as shown above
3. On update handle tags separately inside update in SQL store:
```python
            if model_update.add_tags:
                self._attach_tags_to_resource(
                    tag_names=model_update.add_tags,
                    resource_id=existing_model.id,
                    resource_type=TaggableResourceTypes.MODEL,
                )
            model_update.add_tags = None
            if model_update.remove_tags:
                self._detach_tags_from_resource(
                    tag_names=model_update.remove_tags,
                    resource_id=existing_model.id,
                    resource_type=TaggableResourceTypes.MODEL,
                )
            model_update.remove_tags = None

            existing_model.update(model_update=model_update)
```
5. Update object creation codes:
```python
if model.tags:
    attach_tags_to_resource(
        tag_names=model.tags,
        resource_id=model_schema.id,
        resource_type=TaggableResourceTypes.MODEL,
        sql_store=self,
    )
session.commit()
```

@Cahllagerfeld , this PR changes the Update model for Models by removing `tags` and introducing `add_tags` and `remove_tags`

## Pre-requisites

## Types of changes


src/zenml/enums.py
```diff
@@ -291,3 +291,25 @@ class ModelStages(StrEnum):
...
+    TEAL = 8
+    TURQUOISE = 9
+    MAGENTA = 10
+    BLUE = 11
+
+
+class TaggableResourceTypes(Enum):
+    """All possible resource types for tagging."""
+
+    MODEL = 1
```
AlexejPenner, line 315: any reason we aren't using StrEnum here? it would make it more readable in the DB

src/zenml/enums.py
```diff
@@ -291,3 +291,25 @@ class ModelStages(StrEnum):
     STAGING = "staging"
     PRODUCTION = "production"
     ARCHIVED = "archived"
+
+
+class ColorVariants(Enum):
```
AlexejPenner, line 296: Also same question, any reason we aren't using String enums here?

src/zenml/utils/tag_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                    tag_id=tag.id,
+                    resource_id=resource_id,
+                    resource_type=resource_type,
+                )
+            )
+        except EntityExistsError:
+            pass
+
+
+def delete_links(
```
AlexejPenner, line 61: No idea where to write this - but do we have any way of implementing cascading delete? Like if I delete resource 1 - will all entries for tags on resource 1 be deleted

src/zenml/cli/model.py
```diff
@@ -267,7 +267,7 @@ def update_model(
             trade_offs=tradeoffs,
             ethics=ethical,
             limitations=limitations,
-            tags=tag,
+            add_tags=tag,
```
AlexejPenner, line 270: I would expect this to take a list of strings not just one string

AlexejPenner (commented):
Just a general question, how implicit should this whole tagging be. The way I alway imagined it was when I create a resource with a list of tags (strings) the whole db entity creation happens completely abstracted away from me as a user. Even at the API layer I would imagine that we use updates on the resource to set/remove tags. What do you think about this in general.

src/zenml/enums.py
avishniakov: Save space - the only reason. 4 bytes vs 5+ (for non-unicode db, otherwise 10+)

src/zenml/enums.py
avishniakov: Save space - the only reason. 4 bytes vs 5+ (for non-unicode db, otherwise 10+)

src/zenml/utils/tag_utils.py
avishniakov: It will be cascaded - this is need for removal of tags via update, I'll rename it for clarity

avishniakov: This folk is cascading (set in Models and other taggable entities going forward)
```python
tags: List["TagResourceSchema"] = Relationship(
        back_populates="model",
        sa_relationship_kwargs=dict(
            primaryjoin=f"and_(TagResourceSchema.resource_type=={TaggableResourceTypes.MODEL.value}, foreign(TagResourceSchema.resource_id)==ModelSchema.id)",
            cascade="delete",
        ),
    )
```

src/zenml/cli/model.py
avishniakov: It is a list, just the limitation of click - the name of arg must match the name of arg in the function and the user pass it as `-t a -t b ...`

avishniakov:
> Just a general question, how implicit should this whole tagging be. The way I alway imagined it was when I create a resource with a list of tags (strings) the whole db entity creation happens completely abstracted away from me as a user. Even at the API layer I would imagine that we use updates on the resource to set/remove tags. What do you think about this in general.

It is exactly like you described - the user passes in a list of strings, and creation/retrieval happens under the hood. Next with Update models one can add or remove tags

src/zenml/enums.py
avishniakov: Do you think it is not worth saving?

fa9r: +1 to using StrEnum for more readable DB entries

fa9r: Few bytes per link is practically irrelevant compared to other stuff we dump into DB atm 😅

avishniakov: This makes me feel pain as a former DB developer back in the day, but ok, let's go like this!

src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -2039,3 +2045,119 @@ def delete_authorized_device(self, device_id: UUID) -> None:
...
+        tag_resource_id: UUID,
+    ) -> None:
+        """Deletes a tag resource relationship.
+
+        Args:
+            tag_resource_id: id of the tag<>resource to delete.
+
+        Raises:
+            KeyError: specified ID not found.
+        """
```
fa9r, line 2163: Same here, this should be handled in the `update_{resource}` functions instead

src/zenml/zen_stores/schemas/model_schemas.py
```diff
@@ -73,7 +72,13 @@ class ModelSchema(NamedSchema, table=True):
...
     trade_offs: str = Field(sa_column=Column(TEXT, nullable=True))
     ethics: str = Field(sa_column=Column(TEXT, nullable=True))
-    tags: str = Field(sa_column=Column(TEXT, nullable=True))
+    tags: List["TagResourceSchema"] = Relationship(
+        back_populates="model",
+        sa_relationship_kwargs=dict(
+            primaryjoin=f"and_(TagResourceSchema.resource_type=={TaggableResourceTypes.MODEL.value}, foreign(TagResourceSchema.resource_id)==ModelSchema.id)",
+            cascade="delete",
+        ),
+    )
```
fa9r, line 81: 🙌

src/zenml/zen_stores/schemas/model_schemas.py
```diff
@@ -151,9 +155,13 @@ def update(
...
         """
+        from zenml.utils.tag_utils import create_links, remove_links
+
         for field, value in model_update.dict(exclude_unset=True).items():
-            if field == "tags":
-                setattr(self, field, json.dumps(value))
+            if field == "add_tags":
+                create_links(value, self.id, TaggableResourceTypes.MODEL)
+            elif field == "remove_tags":
+                remove_links(value, self.id)
```
fa9r, line 164: This is problematic since `create_links` and `remove_links` use the `Client()` under the hood, but the client should never be initialized like this inside the server since that leads to all sorts of unexpected behavior.

Instead, simply write those links to the DB directly here (or in the corresponding ZenStore method if it's not possible to do it inside the schema's `update()` function)

fa9r (changes requested):
Awesome, love the overall design, very clean how tags and resources are connected now, well done!!

Overall, the entire PR looks great and very polished already. High-level the only change I'd request is to handle the link creation inside the `update_{resource}` ZenStore methods and get rid of the TagResource endpoints.

Also, I noticed the RestZenStore methods are still missing, but I'm sure integration tests would reveal that sooner or later anyways.

fa9r:
> Just a general question, how implicit should this whole tagging be. The way I alway imagined it was when I create a resource with a list of tags (strings) the whole db entity creation happens completely abstracted away from me as a user. Even at the API layer I would imagine that we use updates on the resource to set/remove tags. What do you think about this in general.

Yes, that's exactly what I meant as well with links being created in update methods instead 👍

src/zenml/enums.py
avishniakov: Done

src/zenml/enums.py
avishniakov: Done

src/zenml/models/model_models.py
```diff
@@ -633,9 +636,14 @@ class ModelResponseModel(
...
     """Model response model.
 
+    tagged: list of tag names attached to the model
     latest_version: name of latest version, if any
     """
 
+    tags: List[TagResponseModel] = Field(
+        title="Tags associated with the model",
+    )
+    tagged: List[str]
```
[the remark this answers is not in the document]
avishniakov: This is for CLI only, yes it is a bit redundant

src/zenml/models/tag_models.py
```diff
@@ -0,0 +1,143 @@
...
+            values["color"] = random.choice(list(ColorVariants)).name.lower()
+        else:
+            _validate_color(values["color"])
+        return values
+
+
+class TagResponseModel(TagBaseModel, BaseResponseModel):
+    """Response model for tags."""
+
+    tagged_count: int = Field(
```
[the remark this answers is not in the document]
avishniakov: It is used in CLI + Zuri has it in designs

src/zenml/zen_server/routers/tags_endpoints.py
```diff
@@ -0,0 +1,214 @@
...
+    tag_resource_id: UUID,
+    _: AuthContext = Security(authorize, scopes=[PermissionType.WRITE]),
+) -> None:
+    """Deletes a tag resource relationship.
+
+    Args:
+        tag_resource_id: id of the tag<>resource to delete.
+
+    """
+    zen_store().delete_tag_resource(tag_resource_id)
```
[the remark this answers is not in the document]
avishniakov: Removed 👍🏼

src/zenml/zen_stores/zen_store_interface.py
avishniakov: Removed 👍🏼

src/zenml/zen_stores/schemas/model_schemas.py
avishniakov: Reworked to use SQL Store instance instead

src/zenml/constants.py
```diff
@@ -224,6 +224,8 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 DEVICE_AUTHORIZATION = "/device_authorization"
 DEVICE_VERIFY = "/verify"
 API_TOKEN = "/api_token"
+TAGS = "/tags"
+TAG_RESOURCES = "/tag_resources"
```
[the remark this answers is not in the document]
avishniakov: Removed

src/zenml/zen_stores/migrations/versions/6119cd9b93c2_tags_table.py
```diff
@@ -0,0 +1,160 @@
...
+        "tag",
+        sa.Column("color", sa.SMALLINT(), nullable=False),
+        sa.Column("id", sqlmodel.sql.sqltypes.GUID(), nullable=False),
+        sa.Column("created", sa.DateTime(), nullable=False),
+        sa.Column("updated", sa.DateTime(), nullable=False),
+        sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    # fetch current tags in Model table
```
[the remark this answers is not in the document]
avishniakov: Did my best in testing this part, hope it will work out 🥲

src/zenml/models/model_models.py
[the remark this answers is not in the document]
fa9r: In that case I'd suggest to delete that arg and simply call `tag.name for tag in tags` wherever you need it

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6560,14 +6563,17 @@ def create_tag_resource(
     def delete_tag_resource(
```
fa9r, line 6563: Is this method and the other still used now that the interface no longer exists?

fa9r (approved):
Beautiful 🙌

src/zenml/zen_stores/sql_zen_store.py
avishniakov: Yeah, called inside helpers.

src/zenml/models/model_models.py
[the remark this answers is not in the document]
avishniakov: Agreed

src/zenml/utils/tag_utils.py
```diff
@@ -0,0 +1,85 @@
+#  Copyright (c) ZenML GmbH 2022. All Rights Reserved.
```
stefannica, line 1: The functionality in this file feels more like it belongs under the `zenml.zen_stores` module than here, maybe even in the `sql_zen_store` module itself. The functions that take `SqlZenStore` as an argument in particular are a good fit for methods under that class.

stefannica (approved):
Beautifully crafted PR ! I only have a couple of suggestions, but nothing that would keep this from being merged.

src/zenml/utils/tag_utils.py
avishniakov: Yeah, agree, makes sense - moved it around.

src/zenml/zen_stores/schemas/tag_schemas.py
```diff
@@ -0,0 +1,153 @@
...
+        Returns:
+            The created `TagResponseModel`.
+        """
+        return TagResponseModel(
+            id=self.id,
+            name=self.name,
+            color=ColorVariants(self.color),
+            created=self.created,
+            updated=self.updated,
+            tagged_count=len(self.links),
```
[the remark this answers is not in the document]
avishniakov: This is just a look-up of the pretty narrow table with links, I assume not a big impact vs complications it will bring to integrate that in all standard functions of sql_store, like listing for instance.

PR #2048, opened 2023-11-14
Use placeholder runs to show pipeline runs in the dashboard without delay

schustmi:
## Describe changes

This PR allows us to create a `PipelineRun` in the database before actually executing the pipeline using the orchestrator.
With this in place, we can now return a reference to this pipeline run when someone is running a pipeline, and can also show the pipeline run in the dashboard immediately.

### Notes regarding the migration
This PR adds a unique constraint for the combination of the `deployment_id` and `orchestrator_run_id` of the `pipeline_run` table. These columns were introduced (and have been set to non-null values since)
- release `0.21.0` for the `orchestrator_run_id`
- release `0.34.0` for the `deployment_id`

For this unique constraint to work, we have to consider these scenarios:
- pipeline runs that happened before release `0.21.0`: For these both columns are `NULL`. We solve this by writing some unique dummy value in the `orchestrator_run_id` column.
- pipeline runs that happened between releases `0.21.0` and `0.34.0`: In this case only the `orchestrator_run_id` is set. This is only a problem if we assume people run with custom orchestrators that do not generate a globally unique `orchestrator_run_id`.
- pipeline runs that happened after release `0.34.0`: For these both `deployment_id` and `orchestrator_run_id` are set and the combination of the two is unique, otherwise it would have failed earlier when trying to run those pipelines.

What the migration currently does not account for:
- If users manually modified their database to set/delete values of the `orchestrator_run_id` column.
- If users deleted `deployments` in their database, which sets the `deployment_id` column to `None`. Similar to above, this is only a problem if we assume people run with custom orchestrators that do not generate a globally unique `orchestrator_run_id`.

### TODO
- Add a new icon for the `initializing` state of a pipeline run in the dashboard.
  - This currently shows the same icon that is displayed when the pipeline is running. IMO, we should still update it but it's not blocking any release. (-> https://zenml.atlassian.net/browse/PROD-160)
- ~~Make sure the dashboard can handle an empty `orchestrator_environment` in a `PipelineRun` response. It is optional on the model but might not be handled correctly in the dashboard.~~
  - This works and just displays nothing.

## Pre-requisites

## Types of changes


src/zenml/models/pipeline_run_models.py
```diff
@@ -150,6 +150,25 @@ def produced_artifacts(self) -> List["ArtifactResponseModel"]:
...
+            if current_run_state.status != ExecutionStatus.RUNNING:
+                for key in current_run_state.__fields_set__:
+                    setattr(self, key, getattr(current_run_state, key))
+                return
+
+            duration = time.time() - start_time
+            if timeout > 0 and duration > timeout:
+                return
+
+            time.sleep(5)
```
htahir1, line 170: might make sense to use a backoff strategy here rather than a fixed time

schustmi: Absolutely, I'm not even sure we want this method like that on this model class. As a user I would expect that to be immutable, but what I do in this method is to update the attributes after fetching an updated copy of the model - not ideal IMO.

Maybe we want an intermediate object as the return value of running a pipeline instead:
```python
something = my_pipeline()
pipeline_run = something.wait()

# The `something.wait()` happens in the pipeline call
pipeline_run = my_pipeline(sync=True)
```

fa9r: Yeah not a big fan of the model fetching the info continuously - but is this even necessary? IMO it's fine to just write a pipeline run with something like `ExecutionStatus.PENDING` or similar to indicate that this is not actually running yet and let the user fetch the run again whenever he wants

src/zenml/new/pipelines/pipeline.py
```diff
@@ -691,6 +693,33 @@ def _run(
 
             self.log_pipeline_deployment_metadata(deployment_model)
 
+            # TODO: We should probably only do this if there is no schedule?
```
fa9r, line 696: 100%, when running a pipeline with schedule we already create the schedule object from which runs will then be created at the corresponding scheduling times - don't think it makes sense to create a placeholder for the first run of a schedule from a UX perspective

src/zenml/new/pipelines/pipeline.py
```diff
@@ -691,6 +693,33 @@ def _run(
...
+                    run_name_template=deployment_model.run_name_template
+                ),
+                orchestrator_run_id=None,
+                user=Client().active_user.id,
+                workspace=Client().active_workspace.id,
+                deployment=deployment_model.id,
+                pipeline=deployment_model.pipeline.id
+                if deployment_model.pipeline
+                else None,
+                # TODO: Maybe we want a state called `preparing` or `launching`
```
fa9r, line 710: Yeah agreed, see comment above 👍

src/zenml/new/pipelines/pipeline.py
```diff
@@ -691,6 +693,33 @@ def _run(
...
+                # to differentiate between waiting for the first step to get
+                # started
+                status=ExecutionStatus.RUNNING,
+                # Is this the actual start time now, or do we want to set this
+                # once the first step started? Especially for scheduled runs
+                # this would be strange
+                start_time=datetime.utcnow(),
```
fa9r, line 717: Depends how we want to define the execution time of a run - does it make sense for this to include the image building / pushing times or should it only be the cummulative execution times of all steps? Maybe @htahir1 has an opinion here?

In any case, we already store `created` as a separate field in the DB so we will also always know when this object was created even if we set the `start_time` later.

Also, scheduling can be ignored IMO since we shouldn't write such placeholder runs for scheduled runs anyway (see comment above).

src/zenml/new/pipelines/pipeline.py
```diff
@@ -691,6 +693,33 @@ def _run(
...
             self.log_pipeline_deployment_metadata(deployment_model)
 
+            # TODO: We should probably only do this if there is no schedule?
+            # It will still work, but it might be hours/days until the run gets
+            # populated
+            pipeline_run_request = PipelineRunRequestModel(
+                name=get_run_name(
+                    run_name_template=deployment_model.run_name_template
+                ),
+                orchestrator_run_id=None,
```
fa9r, line 703: Do we even still need the `orchestrator_run_id` as a field?

src/zenml/new/pipelines/pipeline.py
```diff
@@ -704,32 +733,19 @@ def _run(
...
-                run_url = dashboard_utils.get_run_url(runs[0])
-                if run_url:
-                    logger.info(f"Dashboard URL: {run_url}")
-                else:
-                    logger.info(
-                        "You can visualize your pipeline runs in the `ZenML "
-                        "Dashboard`. In order to try it locally, please run "
-                        "`zenml up`."
-                    )
+            self.register_running_versions(new_version_requests)
```
fa9r, line 736: on another note, this should probably be renamed to `register_running_model_versions` @avishniakov

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -63,7 +71,7 @@ class PipelineRunSchema(NamedSchema, table=True):
     )
 
     # Foreign keys
-    deployment_id: UUID = build_foreign_key_field(
+    deployment_id: Optional[UUID] = build_foreign_key_field(
```
fa9r, line 74: Why is this optional now? The mock run creation still provides a deployment and your alembic migration doesn't make this optional either

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -283,3 +290,43 @@ def update(
...
+
+        Raises:
+            RuntimeError: If the DB entry does not represent a placeholder run.
+            ValueError: If the run request does not match the deployment or
+                pipeline ID of the placeholder run.
+
+        Returns:
+            The updated `PipelineRunSchema`.
+        """
+        if self.orchestrator_run_id:
```
fa9r, line 311: Might also make sense to check for `self.status` here instead and set `self.status = ExecutionStatus.RUNNING` if the placeholder was updated successfully

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4027,22 +4094,45 @@ def get_or_create_run(
         Args:
             pipeline_run: The pipeline run to get or create.
 
+        Raises:
+            ValueError: If the request
```
fa9r, line 4098: ...

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4027,22 +4094,45 @@ def get_or_create_run(
...
+            # No placeholder run available for this run -> We now try to create
+            # a new run in the database, which will fail if a run with the same
+            # orchestrator_run_id and deployment_id already exists. In that case
+            # we fetch and return the existing run.
+            pass
+
         # We want to have the 'create' statement in the try block since running
         # it first will reduce concurrency issues.
         try:
             return self.create_run(pipeline_run), True
```
fa9r, line 4124: If I understand it right, creation from non-placeholder run will only be reached for scheduled runs, right? In that case, you might want to make the docstrings and comments a bit clearer to explain in which scenarios which case is reached, this is currently not easy to follow

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -52,6 +53,13 @@ class PipelineRunSchema(NamedSchema, table=True):
     """SQL Model for pipeline runs."""
 
     __tablename__ = "pipeline_run"
+    __table_args__ = (
+        UniqueConstraint(
+            "orchestrator_run_id",
+            "deployment_id",
+            name="unique_orchestrator_run_id_for_deployment",
+        ),
+    )
```
fa9r, line 62: Is a new deployment created every time you do `python run.py`? Or is there any scenario other than scheduling where deployments are shared among runs?

I'm asking because placeholder runs are now created with empty `orchestrator_run_id` which would fail if there was any other placeholder run with the same deployment ID.

fa9r (commented):
Fundamentally I don't see a reason why this shouldn't work, but let me summarize to check whether I understand it right:
- If we run a pipeline with a schedule, we'll do the same as before: create a schedule in DB, hand over to orchestrator, create run when the orchestrator starts the first step and link it to the schedule, link other steps to same run by searching for runs by orchestrator ID
- If we run a pipeline without schedule, we directly write a run to the DB with empty orchestrator ID, hand over to orchestrator, set orchestrator ID when the orchestrator starts the first step, link other steps to same run by searching for similar orchestrator ID

Logically this makes sense to me and I think it should work 👍 

The only possible failure case I can see is if we try to create a placeholder run for a deployment that already has a placeholder run. But I'm not sure whether that scenario can be reached or not, @schustmi you should know this best 😁

src/zenml/models/pipeline_run_models.py
fa9r: TL;DR I'd suggest we delete this method entirely and let the user figure it out ;)

src/zenml/new/pipelines/pipeline.py
schustmi: I would argue that we should not use the time here as the start time to keep it consistent with our other runs. Scheduled runs (and reruns) will have their start time set to when the first step starts executing, so I think that should also be the case for placeholder runs.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
schustmi: Actually it always was, just with a wrong type annotation. If you take a look at the `build_foreign_key_field` function, we pass `nullable=True` there. This is even necessary, as pipeline runs without deployments exist from a time before we introduced these deployments.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
schustmi: Yes definitely, I added a new `initializing` execution status now and set it to the status of the replacement request here.

src/zenml/zen_stores/sql_zen_store.py
schustmi: Not only scheduled runs, but also rerunning a run on e.g. Kubeflow reuses the deployment and therefore creates new runs. I'll try to add some more comments to explain it better.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
schustmi: Yes, that's correct. The only place where we create deployments and placeholder runs is in the `Pipeline._run` method which means there is a 1-1 relationship between those.

src/zenml/new/pipelines/pipeline.py
schustmi: Currently, we create this schedule object even if the orchestrator does not support scheduling. Do you think we should do the same in this case, only create the placeholder run if the user does not pass a schedule.

That means if they pass a schedule but the orchestrator does not support scheduling, it'll still work but they won't be able to see it in the dashboard immediately.

The alternative would be to add a property `supports_scheduling` on the orchestrator I guess, and do it only if a schedule is passed and the orchestrator supports it?

src/zenml/new/pipelines/pipeline.py
schustmi: While I would still prefer to set the start date only once the first step starts, I think it's nicer to have this timestamp match the one which is used in the run name.

For this reason we decided to go with the simplest approach, which is to simply set both `run_name` and `start_date` already when creating the placeholder run, and not updating it down the line.

src/zenml/orchestrators/utils.py
```diff
@@ -178,8 +160,24 @@ def get_config_environment_vars(
...
+    Returns:
+        The run name derived from the template.
+    """
     date = datetime.utcnow().strftime("%Y_%m_%d")
     time = datetime.utcnow().strftime("%H_%M_%S_%f")
 
-    return run_name_template.format(date=date, time=time)
+    run_name = run_name_template.format(date=date, time=time)
+
+    if run_name == "":
```
fa9r, line 180: When would this ever be reached? Only if the user specifies an empty string as run name manually? In that case, I think that check should happen even earlier, no?

src/zenml/zen_stores/migrations/versions/6917bce75069_add_pipeline_run_unique_constraint.py
```diff
@@ -0,0 +1,65 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "6917bce75069"
+down_revision = "0.47.0"
+branch_labels = None
+depends_on = None
+
+
+def add_orchestrator_run_id_for_old_runs() -> None:
```
fa9r, line 18: No `baris_number` column? :(

fa9r (approved):
Cool, looks good from my side, I think this should work 🙌

src/zenml/orchestrators/utils.py
schustmi: Never :P
As you said, we already catch this earlier during compilation. I just added this check here in the function in case we ever decide to use it somewhere else.

src/zenml/zen_stores/migrations/versions/6917bce75069_add_pipeline_run_unique_constraint.py
schustmi: Decided to add ugly values to existing columns instead of creating a new column just to add ugly values 😄

schustmi: Btw I added some notes on this migration in the PR description, do you think this is an acceptable risk or do we have to find a way around that potential issue?

fa9r: So the main risk is that there might be users with custom orchestrators where the `orchestrator_run_id` is not unique? I would say that's an acceptable risk as long as it does not happen with any of our official orchestrators.

That said, I suppose we could handle this case in the migration if we wanted by first searching for rows that violate the constraints and abandoning the migration with a corresponding error message if there are duplicates. Whether it's worth implementing this is another question - I'll leave that up to you 😉

schustmi: I like that idea, better to be safe with migrations. Shouldn't be too difficult to implement either, I'll give it a try.

schustmi: Done

fa9r (approved):
Let's go to main 🚀

P.S.: Nice tests!

src/zenml/new/pipelines/pipeline.py
avishniakov: This method is removed from develop already - no need.

avishniakov (approved):

fa9r (approved):
Looks still good to me

htahir1 (approved):

PR #2055, opened 2023-11-16
Separate helm hook resources from regular resources

stefannica:
## Describe changes
Hook resources are deletes after the hook phase. This PR duplicates the secret and service account to prevent that the helm hooks delete resources that are required by the workloads.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
+++ b/src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
@@ -25,14 +25,18 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.serviceAccount.create }}
+      serviceAccountName: {{ include "zenml.serviceAccountName" . }}-db-migration
+      {{- else }}
       serviceAccountName: {{ include "zenml.serviceAccountName" . }}
+      {{- end }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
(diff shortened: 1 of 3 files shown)


src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
```diff
@@ -25,14 +25,18 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.serviceAccount.create }}
+      serviceAccountName: {{ include "zenml.serviceAccountName" . }}-db-migration
+      {{- else }}
       serviceAccountName: {{ include "zenml.serviceAccountName" . }}
```
wjayesh, line 31: this can just be "default", correct? `zenml.serviceAccountName` would also check the same condition (- if .Values.serviceAccount.create) and set this to default.
or maybe we can define a zenml.jobServiceAccountName in helpers and just assign it here directly, to keep things consistent.

wjayesh (approved):
feel free to keep it this way too. LGTM!

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
wjayesh: I'm talking about the else condition here.

stefannica: @wjayesh if the service account is created by the user, then we need to use that same account in the db migration job instead of creating a second service account for the job. I don't know what you mean by "default", there's a "default" service account and there's a `default` helm chart template statement.

PR #1638, opened 2023-06-20
Fix run configuration parameter merging

schustmi:
## Describe changes
I implemented/fixed _ to achieve _.

## Pre-requisites
Please ensure you have done the following:
- [ ] I have read the **CONTRIBUTING.md** document.
- [ ] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/component-gallery/integrations) table and the [corresponding website section](https://zenml.io/integrations).
- [ ] I have added tests to cover my changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #2058, opened 2023-11-17
Passing pipeline parameters as yaml config

avishniakov:
## Describe changes
Add ability to pass in pipeline params as yaml config:
```yaml
parameters:
  environment: production
```
Also added a `RuntimeError` for the cases when runtime config conflicts with YAML config (both steps and pipeline params) - @schustmi please confirm that this is acceptable or shall I revert to config-dominant silent mode.
Docs: https://app.gitbook.com/o/-MCl1Hlw9oU4xibZ-ymf/s/9oLgDi1mkUbwQiTmrEit/user-guide/advanced-guide/pipelining-features/configure-steps-pipelines#parameters-for-your-steps

**BREAKING CHANGES**
- What passed silently before now throws an error (step/pipeline params configured in file and code differently)
```yaml
# config.yaml
parameters:
    some_param: 24

steps:
  my_step:
    parameters:
      input_2: 42
```
```python
# run.py
@step
def my_step(input_1: int, input_2: int) -> None:
    pass

@pipeline
def my_pipeline(some_param: int):
    # here an error will be raised since `input_2` is
    # `42` in config, but `43` was provided in the code
    my_step(input_1=42, input_2=43)

if __name__=="__main__":
    # here an error will be raised since `some_param` is
    # `24` in config, but `23` was provided in the code
    my_pipeline(23)
```

## Pre-requisites

## Types of changes


avishniakov:
@schustmi , these changes clash a bit with what you created in https://github.com/zenml-io/zenml/pull/1638.
Can I have your perspective on this? I made an error to be raised if a user provides confusing configurations (step input is set in code and config), but before config, it just dominated it. WDYT, shall I revert to the old behavior or update the test case instead?

avishniakov:
@bcdurak @schustmi can you have a look?

src/zenml/new/pipelines/pipeline.py
```diff
@@ -428,11 +431,50 @@ def prepare(self, *args: Any, **kwargs: Any) -> None:
...
+        param_name: value1
+            
+            
+# pipeline.py
+@pipeline
+def pipeline_(param_name: str):
+    step_name()
+
+if __name__=="__main__":
+    pipeline_(param_name="value2")
```
schustmi, line 469: This would also need to call `with_options(config_file=...)` to showcase the actual error right?

src/zenml/new/pipelines/pipeline.py
```diff
@@ -428,11 +431,50 @@ def prepare(self, *args: Any, **kwargs: Any) -> None:
...
         self._parameters = {}
         self._invocations = {}
 
+        conflicting_parameters = {}
+        parameters_ = (self.configuration.parameters or {}).copy()
+        if from_file_ := self._from_config_file.get("parameters", None):
+            parameters_ = dict_utils.recursive_update(parameters_, from_file_)
+        if parameters_:
+            for k, v_runtime in kwargs.items():
+                if v_config := parameters_.get(k, None):
```
schustmi, line 449: `None` or `False` might also be valid parameter values which will not get detected by this logic I think?

src/zenml/new/pipelines/pipeline.py
```diff
@@ -428,11 +431,50 @@ def prepare(self, *args: Any, **kwargs: Any) -> None:
...
+                    conflicting_parameters[k] = (v_config, v_runtime)
```
schustmi, line 450: In the step you added an additional check if the config and runtime parameter are different, I think that also makes sense here?

src/zenml/steps/base_step.py
```diff
@@ -871,33 +873,47 @@ def _apply_configuration(
...
                 parameters were configured.
+            RuntimeError: If the step has parameters configured differently in
+                configuration file and code.
         """
         if not parameters:
             return
 
+        conflicting_parameters = {}
         for key, value in parameters.items():
+            if runtime_value := runtime_parameters.get(key, None):
```
schustmi, line 914: Same as for the pipeline, `None`/`False`/`0`/... as parameter values will probably not be included in these checks.

schustmi (changes requested):

src/zenml/new/pipelines/pipeline.py
avishniakov: Fixed

src/zenml/steps/base_step.py
avishniakov: Fixed

PR #2072, opened 2023-11-22
Alembic migration fix for databases with scheduled pipelines

bcdurak:
## Describe changes

This is the initial version of the PR. I will test this in different scenarios and share the details about the scenarios and the outcomes here.

## Pre-requisites
- [X] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/stacks-and-components/component-guide) table and the [corresponding website section](https://zenml.io/integrations).

## Types of changes


avishniakov (approved):

src/zenml/zen_stores/migrations/versions/37835ce041d2_optimizing_database.py
```diff
@@ -52,7 +73,7 @@ def upgrade() -> None:
         """
         UPDATE pipeline_deployment
         SET client_version = (
-            SELECT pipeline_run.client_version
+            SELECT max(pipeline_run.client_version)
```
schustmi, line 76: The fact that we have to merge/take the maximum of these values indicates to me that the right place for these columns is actually the `PipelineRun` instead of the `PipelineDeployment`. This is because multiple runs of the same deployment might get started from different environments (especially once running a pipeline through the API is implemented) and also run with different server versions if it was upgraded in the meantime.

I wouldn't say that we implement this right now, but I would keep it in mind/create a ticket for the future.

schustmi: A question more relevant for this PR: Have you checked how the SQL `max()` function behaves for strings? Does it use the maximum version in this case? And is that even something we care about which version gets picked in case there are different ones?

bcdurak: I agree with you about the placement. We can look into how to solve this in a future PR. As for the `max()` function, I tested it with both SQLite and MySQL, it seems to be working as intended.

schustmi (approved):

PR #2073, opened 2023-11-23
Improve alembic migration safety

fa9r:
## Describe changes
As part of the data versioning migration we make the foreign keys `artifact_id` and `pipeline_run_id` of the model links non-nullable. Under MySQL, this requires us to drop the FK constraints before the DB change. Additionally, I added an additional safety measure here which all empty links before the migration.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/389046140cad_data_versioning.py
```diff
@@ -83,20 +83,48 @@ def upgrade() -> None:
         )
 
     # Model Version Artifacts and Runs Tables
+    # Drop all entries with null IDs
+    op.execute(
```
schustmi, line 87: In the cloud case where we discovered this issue, there wasn't actually any entry where the `artifact_id` or `pipeline_run_id` is `NULL`, but the issue was related to the lower fix here which deletes/recreates the foreign key. Do we still want this to happen, as it potentially deletes user data that might be relevant to them?

In my pipeline run migration, I encountered a similar case and instead
- checked whether there were any entries that would fail the constraint
- raise an exception which explains the reason why the migration failed

If we do this as the initial step of the migration, nothing else will happen and users can just re-run it once they fixed the erroneous entries.

schustmi (changes requested):

src/zenml/zen_stores/migrations/versions/389046140cad_data_versioning.py
fa9r: It's just links of a model version to an artifact and any links without such IDs would be completely useless and even break the dashboard, so I think we should drop these regardless

schustmi (approved):

src/zenml/zen_stores/migrations/versions/389046140cad_data_versioning.py
avishniakov: How it happen that `artifact_id` is null in this table? It is not possible according to the Request models. This might be a sign of a bigger issue, rather than just something we need to fix with a plug like that. Can I have a look at some instances in such a state?

avishniakov: It should cascade then
```python
    model_versions_artifacts_links: List[
        "ModelVersionArtifactSchema"
    ] = Relationship(
        back_populates="artifact",
        sa_relationship_kwargs={"cascade": "delete"},
    )
```

avishniakov (approved):
I would like some justification of the need to do `DELETE FROM model_versions_artifacts WHERE artifact_id IS NULL` and the same for pipeline links. Rest is good.

On the other hand, if this was the case before, we will see some errors here and there after this migration, since NULL is not accepted anymore.

PR #2102, opened 2023-12-04
Model Versions are taggable

avishniakov:
## Describe changes
I implemented tags support for Model Versions in Client, CLI, and endpoints.
New Model Version will also get tags from `ModelVersion` on every run, if Model Version or Model exists - tag has to be updated by CLI/ Client/Frontend instead.

## Pre-requisites

## Types of changes


avishniakov:
> @avishniakov I think its a good idea to ask Zuri, but it might also be a permission scope thing right? Like maybe you dont have the ability to edit the model so then your pipeline run will fail?
> 
> What is the current behavior? Does the pipeline fail or just ignore/give a warning? If its the latter, then im happy to keep that behavior

Now it just ignoring tags - they are used only if new ModelVersion/Model is created

htahir1:
@avishniakov ok then lets leave that behavior until someone else objects

strickvl (commented):
LGTM aside from two things:

1) you've update the schemas, so doesn't this require a DB migration?
2) could you update the `src/zenml/cli/__init__.py` to include instructions on adding / using these tags in the appropriate section?

avishniakov:
> LGTM aside from two things:
> 
> 1. you've update the schemas, so doesn't this require a DB migration?
> 2. could you update the `src/zenml/cli/__init__.py` to include instructions on adding / using these tags in the appropriate section?

1. It is all "virtual" references - alembic produces nothing on this change
2. I added a section to CLI docs since it was never there.

src/zenml/cli/__init__.py
```diff
@@ -882,6 +882,75 @@
...
+zenml model update <MODEL_NAME_OR_ID> --tag <TAG> --tag <TAG> .. 
+   --remove-tag <TAG> --remove-tag <TAG> ..
+```
+
+To delete a model, use:
+```bash
+zenml model delete <MODEL_NAME_OR_ID>
+```
+
+Model CLI interface also helps to navigate through artifacts linked to a specific model versions.
```
strickvl, line 917:
```suggestion
The CLI interface for models also helps to navigate through artifacts linked to a specific model versions.
```

src/zenml/cli/__init__.py
```diff
@@ -882,6 +882,75 @@
...
+```bash
+zenml model data_artifacts <MODEL_NAME_OR_ID> [-v <VERSION>]
+zenml model endpoint_artifacts <MODEL_NAME_OR_ID> [-v <VERSION>]
+zenml model model_artifacts <MODEL_NAME_OR_ID> [-v <VERSION>]
+```
+
+Model CLI interface also helps to navigate through pipeline runs linked to a specific model versions.
```
strickvl, line 924:
```suggestion
You can also navigate the pipeline runs linked to a specific model versions:
```

src/zenml/cli/__init__.py
```diff
@@ -882,6 +882,75 @@
...
+```bash
+zenml model runs <MODEL_NAME_OR_ID> [-v <VERSION>]
+```
+
+To list model versions of a specific model, use:
```
strickvl, line 929:
```suggestion
To list the model versions of a specific model, use:
```

src/zenml/cli/__init__.py
```diff
@@ -882,6 +882,75 @@
...
+To delete a model version, use:
+```bash
+zenml model version delete <MODEL_NAME_OR_ID> <VERSION>
+```
+
+To update a model version, use:
+```bash
+zenml model version update <MODEL_NAME_OR_ID> <VERSION> [--MODEL_VERSION_OPTIONS]
+```
+Important options for update of model version are:
```
strickvl, line 943:
```suggestion
These are some of the more common uses of model version updates:
```

src/zenml/cli/__init__.py
```diff
@@ -882,6 +882,75 @@
...
+- stage
```
strickvl, line 944:
```suggestion
- stage (i.e. promotion)
```

strickvl (approved):
Right got it. It's just creating the new fields from things it already has rather than setting up new places to store things. Nice one!

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6496,6 +6503,22 @@ def update_model_version(
                             f"Model version {existing_model_version_in_target_stage.name} has been set to {ModelStages.ARCHIVED.value}."
                         )
 
+            if model_version_update_model.add_tags:
+                self._attach_tags_to_resource(
+                    tag_names=model_version_update_model.add_tags,
+                    resource_id=existing_model_version.id,
+                    resource_type=TaggableResourceTypes.MODEL_VERSION,
+                )
+            model_version_update_model.add_tags = None
```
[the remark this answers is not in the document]
avishniakov: No need, removed

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6496,6 +6503,22 @@ def update_model_version(
...
+            if model_version_update_model.remove_tags:
+                self._detach_tags_from_resource(
+                    tag_names=model_version_update_model.remove_tags,
+                    resource_id=existing_model_version.id,
+                    resource_type=TaggableResourceTypes.MODEL_VERSION,
+                )
+
+            model_version_update_model.remove_tags = None
+
```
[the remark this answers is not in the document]
avishniakov: Not sure that there is a real explanation behind that. Can you share some Schemas having such "extended" updates? For instance in `run_metadata` we use an extra method to create those, so it is not really closely coupled with the object itself (e.g. `create_run_metadata` is called as one of the steps in `save_artifact` utility).
It might turn out that having those very related entities coupled in one update method is a bit better choice, IMO. But it is definitely not a battle I want to fight in, so if you share examples and reject my explanations - I will rework 👍🏼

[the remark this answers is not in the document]
bcdurak: I think in their first iterations, most schemas had an `update `function which was relatively more extended. However, as I went through the current implementation, I see that this is not the case anymore. I don't think we need to rework anything in the scope of this PR though, ideally, this would be a separate cleaning-up ticket in the future.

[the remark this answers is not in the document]
bcdurak: If we need examples in any case:

- [Code repository schemas](https://github.com/zenml-io/zenml/blob/main/src/zenml/zen_stores/schemas/code_repository_schemas.py): The entire update on the schema happens within the update method not in the SQLZenStore.
- [Stack component schemas](https://github.com/zenml-io/zenml/blob/main/src/zenml/zen_stores/schemas/component_schemas.py): The update on the configuration and labels are happening inside the update function whereas the update on the connectors is happening on the SQLZenStore.,

This is the inconsistency that I find suboptimal. IMO, the update should either happen fully on the `SQLZenStore` or fully on the `update` method of a schema. Otherwise, the code becomes very hard to track and read.

PR #2131, opened 2023-12-11
Fixing the colab issues with stack cache

bcdurak:
## Describe changes
Allow me to quickly summarize what went wrong in the Colab with the order of how things are executed.

- We install ZenML and register a stack **without a model registry.**
- Then, we execute some ZenML-related code, which instantiates a ZenML `Client`.
- This `client`, which is a singleton, has now the `active_stack` set **(again without a model registry).**
- Afterwards, we update the stack through a CLI command.
- Even though the stack gets updated correctly, since this is a CLI command, this does not affect either the `_STACK_CACHE` or the `client` singleton.
- This is where things take an interesting turn. Consecutively, we try to launch a pipeline.
- The pipeline uses the `step_launcher` which is initiated with a `deployment` object.
- This `deployment` object is fetched from the DB by the `load_deployment` method of the step entrypoint.
- Since the fetch happens after the stack is updated, this `deployment` object has the correct and updated version of the stack.
- The step launcher calls the `Stack().from_model` method which, in return, puts the updated second version of the stack in the cache. 
- Even though now the second version of the stack is in the cache, the client singleton has no notion that the stack was actually updated and does not update the `active_stack` property.
- The step eventually tries to use this property to fetch the model registry but can not find it and it fails.

With the proposed solution, if we ever write an updated version of the active stack to the cache, we automatically update the client accordingly.

## Pre-requisites
- [X] If I have added an integration, I have updated the [integrations](https://docs.zenml.io/stacks-and-components/component-guide) table and the [corresponding website section](https://zenml.io/integrations).

## Types of changes

### DIFF
--- a/src/zenml/models/v2/core/stack.py
+++ b/src/zenml/models/v2/core/stack.py
@@ -14,6 +14,7 @@
 """Models representing stacks."""
 
 import json
+from datetime import datetime
 from typing import Any, ClassVar, Dict, List, Optional, Union
 from uuid import UUID
 
@@ -91,6 +92,13 @@ class StackUpdate(StackRequest):
 
 class StackResponseBody(WorkspaceScopedResponseBody):
     """Response body for stacks."""
+
+    created: datetime = Field(
+        title="The timestamp when this component was created."
+    )
+    updated: datetime = Field(
+        title="The timestamp when this component was last updated.",
+    )
 
 
 class StackResponseMetadata(WorkspaceScopedResponseMetadata):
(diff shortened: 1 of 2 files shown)


avishniakov (approved):
Go for it 🦭

avishniakov:
Before we merge, I'm curious why this call was not updating properly?
https://github.com/zenml-io/zenml/blame/00f3039e509cd3253077a859ab861146395e3ecf/src/zenml/client.py#L1179-L1183

bcdurak:
AFAIK, since the update is done through a CLI command, it has no effect on the Python runtime of the first client.

PR #2144, opened 2023-12-13
Warn if Model(Version) config fluctuates from DB state

avishniakov:
## Describe changes
I implemented a warning logic to notify users about differences in the current Model(Version) configuration vs what is registered in the DB, since it will not be updated for free. Plus added guidance about how to make it sync again using CLI.

Side change: I enabled storing of `save_models_to_registry` config in DB schema, which was missed before.

## Pre-requisites

## Types of changes


examples/nlp-case/run.py
```diff
@@ -67,7 +67,7 @@
 @click.option(
     "--no-cache",
     is_flag=True,
-    default=False,
+    default=True,
```
strickvl, line 70: shouldn't this get updated through the CI?

src/zenml/model/model_version.py
```diff
@@ -470,6 +496,25 @@ def _get_model_version(self) -> "ModelVersionResponse":
...
+        configured_tags = set(self.tags or [])
+        db_tags = {t.name for t in mv.tags}
+        if db_tags != configured_tags:
+            difference["tags added"] = list(configured_tags - db_tags)
+            difference["tags removed"] = list(db_tags - configured_tags)
+        if difference:
+            logger.warning(
+                "Provided model version configuration does not match existing model "
+                f"version `{self.name}::{self.version}` with the following "
+                f"changes: {difference}. If you want to update the model version "
```
strickvl, line 514: Will this display nicely? I guess it's just logged but wondering if we can make it a bit more readable?

examples/nlp-case/run.py
avishniakov: It was: https://github.com/zenml-io/zenml/pull/2144/commits/d8ce54bb4c6e224e2081c61e0bd18832ca11e470

src/zenml/model/model_version.py
strickvl: Was thinking to leave a linebreak at least, but if you think it looks ok then that's good enough for me. It's logging output anyway, not our CLI output etc.

PR #2154, opened 2023-12-14
Feature: use service connectors to authenticate secrets stores.

stefannica:
## Describe changes
Reuse the AWS/GCP/Azure Service Connector functionality to authenticate the AWS/GCP/Azure Secrets Store instead of relying on the limited configuration currently available in the Secrets Store implementations.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
+++ b/src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
@@ -26,12 +26,13 @@ from typing import (
     List,
     Optional,
     Type,
+    cast,
 )
-from uuid import UUID
+from uuid import UUID, uuid4
 
 import boto3
 from botocore.exceptions import ClientError
-from pydantic import SecretStr
+from pydantic import BaseModel, Field, root_validator
 
 from zenml.analytics.enums import AnalyticsEvent
 from zenml.analytics.utils import track_decorator
@@ -43,6 +44,12 @@ from zenml.enums import (
     SecretsStoreType,
 )
 from zenml.exceptions import EntityExistsError
+from zenml.integrations.aws.service_connectors.aws_service_connector import (
+    AWS_CONNECTOR_TYPE,
+    AWS_RESOURCE_TYPE,
+    AWSAuthenticationMethods,
+    AWSServiceConnector,
+)
 from zenml.logger import get_logger
 from zenml.models import (
     Page,
@@ -50,6 +57,10 @@ from zenml.models import (
     SecretRequestModel,
     SecretResponseModel,
     SecretUpdateModel,
+    ServiceConnectorRequest,
+)
+from zenml.service_connectors.service_connector_registry import (
+    service_connector_registry,
 )
 from zenml.zen_stores.secrets_stores.base_secrets_store import (
     BaseSecretsStore,
@@ -59,6 +70,10 @@ logger = get_logger(__name__)
 
 
 AWS_ZENML_SECRET_NAME_PREFIX = "zenml"
+
+
+class AWSSecretsStoreConnector(BaseModel):
+    """AWS secrets store connector configuration."""
 
 
 class AWSSecretsStoreConfiguration(SecretsStoreConfiguration):
@@ -86,18 +101,40 @@ class AWSSecretsStoreConfiguration(SecretsStoreConfiguration):
     """
 
     type: SecretsStoreType = SecretsStoreType.AWS
-    region_name: str
-    aws_access_key_id: Optional[SecretStr] = None
-    aws_secret_access_key: Optional[SecretStr] = None
-    aws_session_token: Optional[SecretStr] = None
+
+    auth_method: AWSAuthenticationMethods = AWSAuthenticationMethods.SECRET_KEY
+    auth_config: Dict[str, Any] = Field(default_factory=dict)
+
     list_page_size: int = 100
     secret_list_refresh_timeout: int = 0
+
+    @root_validator(pre=True)
+    def populate_config(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        """Populate the connector configuration from legacy attributes.
+
+        Args:
+            values: Dict representing user-specified runtime settings.
+
+        Returns:
+            Validated settings.
+
+        Raises:
+            ValueError: If the connector attribute is not set.
+        """
+        if "auth_method" not in values or "auth_config" not in values:
+            values["auth_config"] = dict(
+                aws_access_key_id=values.get("aws_access_key_id"),
+                aws_secret_access_key=values.get("aws_secret_access_key"),
+                region=values.get("region_name"),
+            )
+
+        return values
 
     class Config:
         """Pydantic configuration class."""
 
         # Forbid extra attributes set in the class.
-        extra = "forbid"
+        extra = "allow"
 
 
 class AWSSecretsStore(BaseSecretsStore):
@@ -159,6 +196,7 @@ class AWSSecretsStore(BaseSecretsStore):
     ] = AWSSecretsStoreConfiguration
 
     _client: Optional[Any] = None
+    _connector: Optional[AWSServiceConnector] = None
 
     @property
     def client(self) -> Any:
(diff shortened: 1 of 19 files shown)


src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
```diff
@@ -193,13 +211,21 @@ def client(self) -> Any:
...
 
-        # Initialize the AWS client early, just to catch any configuration or
-        # authentication errors early, before the Secrets Store is used.
-        _ = self.client
+        Args:
+            client: The authenticated client object returned by the service
+                connector.
+
+        Returns:
+            The GCP Secrets Manager client.
```
wjayesh, line 222:
```suggestion
            The AWS Secrets Manager client.
```

src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
```diff
@@ -144,46 +193,15 @@ class AWSSecretsStore(BaseSecretsStore):
...
-                aws_secret_access_key=self.config.aws_secret_access_key.get_secret_value()
-                if self.config.aws_secret_access_key
-                else None,
-                aws_session_token=self.config.aws_session_token.get_secret_value()
-                if self.config.aws_session_token
-                else None,
-            )
-        return self._client
+    SERVICE_CONNECTOR_TYPE: ClassVar[str] = AWS_CONNECTOR_TYPE
+    SERVICE_CONNECTOR_RESOURCE_TYPE: ClassVar[str] = AWS_RESOURCE_TYPE
```
wjayesh, line 204: Does `aws-generic` include checks for accessing the secrets store API?

wjayesh: Same question for GCP and Azure.

src/zenml/zen_stores/secrets_stores/azure_secrets_store.py
```diff
@@ -154,39 +180,7 @@ def client(self) -> SecretClient:
...
-                    client_id=self.config.azure_client_id.get_secret_value(),
-                    client_secret=self.config.azure_client_secret.get_secret_value(),
-                )
-            else:
-                credential = DefaultAzureCredential()
-            self._client = SecretClient(
-                vault_url=vault_url, credential=credential
-            )
-        return self._client
+        return cast(SecretClient, super().client)
```
wjayesh, line 183: - Why is there a need to cast this object into `SecretClient`? Don't we already return that type from super().client?

- Why do we not do the same for, say, AWS?

src/zenml/zen_stores/schemas/code_repository_schemas.py
```diff
@@ -109,13 +109,13 @@ def to_model(self, hydrate: bool = False) -> "CodeRepositoryResponse":
             user=self.user.to_model() if self.user else None,
             source=json.loads(self.source),
             logo_url=self.logo_url,
+            created=self.created,
```
wjayesh, line 112: Why was this done?

wjayesh (commented):
Happy to see connectors being used for secrets; less duplication and cleaner code. Thanks!

src/zenml/zen_stores/secrets_stores/service_connector_secrets_store.py
```diff
@@ -0,0 +1,179 @@
...
+    @property
+    def client(self) -> Any:
+        """Get the secrets store API client.
+
+        Returns:
+            The secrets store API client instance.
+        """
+        # Multiple API calls can be made to this secrets store in the context
+        # of different threads. We want to make sure that we only initialize
+        # the client once, and then reuse it. We have to use a lock to treat
```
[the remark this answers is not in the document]
stefannica: Yes, correct, this is one of the reasons. The way the REST API works is by servicing every API request in its own thread. However, there is a single Secrets Store instance that is used in common by all these requests.

With Service Connectors now in the picture, it can happen that the temporary credential issued by the service connector expires. When that happens, we have to issue a new credential (same as we do with orchestrators like Vertex AI and other stack components that can be linked to a service connector).

Because we refresh credentials on demand, this needs to be an atomic operation, otherwise two or more threads doing this at the same time may create race conditions that can result in errors.

We could issue a separate credential for every API request, but this is expensive and would make the secret API calls very slow.

src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
stefannica: No, `aws-generic` doesn't care about the particular cloud service that the credentials are used to access. We could add another resource type and call it `secrets-manager` for example, and include those checks, but that would take more effort and expose "secrets manager" as a public resource type that the service connectors are capable of managing, like `docker-registry` and `kubernetes-cluster`. However, there's no clear advantage for making this an established service connector feature, because there are no stack components that can be "linked" to a secrets manager (we had the Secrets Manager stack component, but we decided to remove it).

src/zenml/zen_stores/schemas/code_repository_schemas.py
stefannica: You caught me. This was a minor, unrelated bug that I fixed here for @Cahllagerfeld . Too small to open a separate PR for it :)

src/zenml/zen_stores/secrets_stores/azure_secrets_store.py
stefannica: This is mainly for mypy and to facilitate auto-complete in the IDE. The `super().client` property returns `Any`. We just cast it here so the Azure specific code knows it is working with an Azure specific client object:

```
    @property
    def client(self) -> Any:
        """Get the secrets store API client.

        Returns:
            The secrets store API client instance.
        """
        # Multiple API calls can be made to this secrets store in the context
        # of different threads. We want to make sure that we only initialize
        # the client once, and then reuse it. We have to use a lock to treat
        # this method as a critical section.
        with self.lock:
            return self._get_client()
```

stefannica: We do the same for GCP:

```
    @property
    def client(self) -> SecretManagerServiceClient:
        """Initialize and return the GCP Secrets Manager client.

        Returns:
            The GCP Secrets Manager client instance.
        """
        return cast(SecretManagerServiceClient, super().client)
```

but not for AWS, (I suppose) because the boto3 library isn't typed and is not covered by mypy (i.e. it uses `Any` for the client type).

stefannica: Yes, if you look in the `pyproject.toml` file, `boto3` is mentioned in the list of python modules that have mypy disabled:

```
[[tool.mypy.overrides]]
module = [
    "tensorflow.*",
    "apache_beam.*",
    "pandas.*",
    ...
    "boto3.*",
    "botocore.*",
    ...
]
```

src/zenml/zen_stores/secrets_stores/service_connector_secrets_store.py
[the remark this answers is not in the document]
wjayesh: I see, thanks for the explanation. I get it now.

src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
wjayesh: Makes sense to not "expose" the secrets manager as a resource, I agree. However, does this mean that we can't verify if the credentials that are provided can access the secrets API? Shouldn't be a big deal in any case, we would get the error from the API anyways.

src/zenml/zen_stores/secrets_stores/azure_secrets_store.py
wjayesh: Ah, okay. Got it, thanks!

src/zenml/zen_stores/schemas/code_repository_schemas.py
wjayesh: Haha, coolcool!

src/zenml/zen_stores/secrets_stores/aws_secrets_store.py
stefannica: There is some basic verification happening still, but it's mainly to verify that the credentials are valid and can be used to authenticate to the target AWS/GCP/Azure cloud:

* for AWS, we create a boto3 session and call `get_caller_identity`:
```        sts_client = session.client("sts")
        try:
            sts_client.get_caller_identity()
        except (ClientError, BotoCoreError) as err:
            msg = f"failed to verify AWS account access: {err}"
            logger.debug(msg)
            raise AuthorizationException(msg) from err
```

* for GCP, we use the GCP auth library to exchange the credentials for an OAuth2 token:
```
            try:
                with requests.Session() as session:
                    req = Request(session)
                    credentials.refresh(req)
            except google.auth.exceptions.GoogleAuthError as e:
                raise AuthorizationException(
                    f"Could not fetch GCP OAuth2 token: {e}"
                )
```

* for Azure, we don't really verify anything, because the Azure authentication API is the least advanced of the three 

So generally speaking, we do check that the credentials are valid (well formed, can be used to authenticate to the target cloud). However, there is no verification done concerning permissions to access the Secrets Manager API though, not until the user actually tries to create/get/list/delete secrets.

wjayesh: Thanks for the detailed response!

wjayesh (approved):
LGTM 🛷!

PR #2158, opened 2023-12-15
Add DB backup and recovery during DB schema migrations

wjayesh:
## Describe changes
To help with failed DB migrations, this PR implements database backup and restoration features directly in the SQL ZenML store and uses them to automatically recover from DB migration failures.

Four different DB backup strategies are implemented, depending on where and how the backup is stored:

1. DB backup is disabled
2. in-memory: the database schema and its entire contents are stored in-memory.
3. dump file: the database schema and its entire contents are stored in a JSON file on disk (e.g. in a persistent volume, for helm chart deployments)
4. database: the database is cloned in the same database server using a different database

For SQLite databases, the backup/restore operations are performed differently: the SQLite database file itself is copied/restored.

The backup/restore are performed automatically for all types of deployments when DB migrations are executed. They can also be invoked manually via two newly implemented CLI commands.

### Implementation Details

- The DB connection values are derived from the Zen store configuration
- The location where the file dump is stored:
  - A persistent volume, if a `zenml.database.storageClass` value was provided to the Helm chart. Different Kubernetes providers have different storage classes and we can't set a default for the chart.
  OR
  - An empty directory volume, otherwise.
- The database backup/recovery is performed during DB migrations and only if a schema migration will actually happen (i.e.  the alembic head and current revisions don't match).
- A proprietary JSON file format is used for the DB dump instead of a traditional SQL dump for the following reasons:
  - avoids the pains of having to install the `mysqldump` binary version aligned with the target DB server type and version
  - not vulnerable to SQL injection attacks
  - somewhat easier to read, understand and fix manually in case of problems
- The data being transferred between the source database and the target backup is processed in a manner that minimizes memory consumption, which means this scales well with larger databases.

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
	- Implemented new commands for database backup and restoration.
	- Introduced database backup strategies and related configurations.

- **Refactor**
	- Enhanced database migration utilities and methods for improved reliability.

- **Style**
	- Updated script for better compatibility across environments.


src/zenml/cli/base.py
```diff
@@ -629,12 +629,152 @@ def migrate_database(skip_default_registrations: bool = False) -> None:
...
+        password=db_password,
+    )
+
+    # check if the database exists
+    # if it does not exist, return
+    with connection.cursor() as cursor:
+        result = cursor.execute(f'SHOW DATABASES LIKE "{db_name}"')
+        if result == 0:
+            logger.debug("Database does not exist. No backup created.")
+            return
```
stefannica, line 707: same thing here: you can reuse a lot of the existing code in the `SqlZenStore` to achieve this instead of reinventing the wheel and duplicating the code here.

src/zenml/cli/base.py
```diff
@@ -629,12 +629,152 @@ def migrate_database(skip_default_registrations: bool = False) -> None:
...
+        BaseZenStore.create_store(
+            store_config, skip_default_registrations=skip_default_registrations
+        )
+        logger.debug("Database migration successful.")
+    except Exception as e:
+        mysql_restore_from_s3(store_config.url)
+        cli_utils.error(f"Database migration failed with error: {e}")
+
+
+def mysql_dump_to_s3(url: str) -> None:
```
stefannica, line 680: to s3 ?

src/zenml/cli/base.py
```diff
@@ -629,12 +629,152 @@ def migrate_database(skip_default_registrations: bool = False) -> None:
...
+    # then create the database
+    with connection.cursor() as cursor:
+        cursor.execute(f"CREATE DATABASE {db_name}")
+
+    # restore the database from the dump file
+    with connection.cursor() as cursor:
+        cursor.execute(f"USE {db_name}")
+        cursor.execute(f"SOURCE {backups_directory}/{db_name}.sql")
+
+    logger.debug(f"Database restored from {backups_directory}/{db_name}.sql")
```
stefannica, line 777: nice !

src/zenml/zen_server/deploy/helm/values.yaml
```diff
@@ -194,6 +194,8 @@ zenml:
     # sslCert: /path/to/client-cert.pem
     # sslKey: /path/to/client-key.pem
     # sslVerifyServerCert: True
+    backupStorageSize: 1Gi
+    # storageClass: standard
```
stefannica, line 198: `backupStorageClass` is a better name for this, otherwise it can be confused with the volume used for the sqlite DB (which we don't support yet, but maybe we will).

src/zenml/zen_server/deploy/helm/templates/server-db-pvc.yaml
```diff
@@ -0,0 +1,18 @@
+# if a storage class is set, create the pvc that will hold a backup of the database
+{{- if .Values.zenml.database.storageClass }}
+apiVersion: v1
+kind: PersistentVolumeClaim
```
stefannica, line 4: I think you'll find that there are problems if you mix helm-hook resources with regular resources. I'm not sure about volumes, but I think the same will happen here: the volume won't be created until _after_ the DB job runs successfully (which can't happen).

stefannica (changes requested):
Looks good so far.

src/zenml/zen_server/deploy/helm/templates/server-db-pvc.yaml
wjayesh: Makes sense, good catch!

src/zenml/zen_server/deploy/helm/values.yaml
wjayesh: Good point, I agree.

src/zenml/cli/base.py
wjayesh: Makes the code cleaner, I like it. Thanks!

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -917,11 +923,74 @@ def _initialize(self) -> None:
...
+            # TODO add a try catch block around this fn call
+            # perform DB restore if exception is raised.
+            try:
+                self.migrate_database()
+            except Exception as e:
+                logger.error(
+                    "Failed to migrate the database. Attempting to restore "
+                    "the database from a backup."
+                )
+                self.restore_database()
```
stefannica, line 944: better to add a try/finally around this and log the exception with `logger.exception` in case the restoration doesn't work either

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -917,11 +923,74 @@ def _initialize(self) -> None:
...
+
+        # create the backup directory if it does not exist
+        if not os.path.exists(self.config.backup_directory):
+            os.makedirs(self.config.backup_directory)
+
+        # dump the database to a file
+        os.system(
+            f"mysqldump --set-gtid-purged=OFF -h {self.engine.url.host} -u {self.config.username} "
+            f"-p{self.config.password} {db_name} > {self.db_backup_file_path}"
+        )
```
stefannica, line 991: If you check the SQL store configuration, you'll notice there are more configuration attributes that control how to connect to the database in addition to host, username, password and database name. I'm referring to the SSL configuration: the CA certificate as well as the client certificate and key. These should all be reflected here.

stefannica (changes requested):

src/zenml/zen_stores/sql_zen_store.py
wjayesh: Good catch!

src/zenml/zen_stores/sql_zen_store.py
wjayesh: Makes sense, i've added it.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -961,11 +968,41 @@ def _initialize(self) -> None:
...
+                    "the database from a backup."
+                )
+                try:
+                    self.restore_database()
+                except Exception:
+                    logger.exception(
+                        "Failed to restore the database from a backup. "
+                        "Please check the logs for more details."
+                    )
+                raise e
```
stefannica, line 1005:
```suggestion
            except Exception:
                logger.error(
                    "Failed to migrate the database. Attempting to restore "
                    "the database from a backup."
                )
                try:
                    self.restore_database()
                except Exception:
                    logger.exception(
                        "Failed to restore the database from a backup. "
                        "Please check the logs for more details."
                    )
                raise
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1005,6 +1042,111 @@ def _initialize(self) -> None:
...
+
+        # create the backup directory if it does not exist
+        if not os.path.exists(self.config.backup_directory):
+            os.makedirs(self.config.backup_directory)
+
+        if self.config.driver == SQLDatabaseDriver.SQLITE:
+            # dump the database to a file
+            os.system(
+                f"sqlite3 {self.config.database} .dump > {self.db_backup_file_path}"
+            )
```
stefannica, line 1095:
```suggestion
            )
            return
```
otherwise you fall over to the mysql code.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1005,6 +1042,111 @@ def _initialize(self) -> None:
...
+        # if a backup file does not exist, return
+        assert self.db_backup_file_path is not None
+        db_name = self.db_backup_file_path.split("/")[-1].split(".")[0]
+        if not os.path.exists(self.db_backup_file_path):
+            logger.debug("No backup found. No restore performed.")
+            return
+
+        with self.engine.begin() as connection:
+            # drop the database if it exists
+            connection.execute(text(f"DROP DATABASE IF EXISTS {db_name}"))
```
stefannica, line 1056: Pretty sure this doesn't work on SQLite. But then again, that shouldn't need running SQL statements, it should be enough to copy the SQLite DB backup file back onto the current DB location.

stefannica: This also reminds me that in some scenarios, the database user might not be allowed to drop the database or even to create it for that matter, because it only is allowed access to this database to begin with. Not a big deal for now, just that we need to account for this in our own deployments, if we want to implement granular, per-database user access. Automated DB restoration might not work in that case.

stefannica (changes requested):
This is really close to a final shape. Really looking forward to seeing this in action !

src/zenml/zen_stores/sql_zen_store.py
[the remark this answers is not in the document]
stefannica: done

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1064,15 +1206,32 @@ def migrate_database(self) -> None:
...
+                    logger.info("Backing up the database")
+                    self.backup_database()
+                except Exception as e:
+                    logger.exception(
+                        "Failed to backup the database. Please check the logs "
+                        "for more details."
+                    )
+                    # TODO catch exception when the directory doesn't exist
+                    # write permissions etc.
+                    raise e
```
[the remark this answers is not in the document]
stefannica: done

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1058,24 +1250,98 @@ def migrate_database(self) -> None:
 
         alembic_logger.addHandler(get_console_handler())
 
-        # We need to account for 3 distinct cases here:
+        # We need to account for 2 distinct cases here:
```
schustmi, line 1253:
```suggestion
        # We need to account for 3 distinct cases here:
```

schustmi (commented):
Haven't looked at anything in-depth so don't count this as a review, just wanted to say great job @wjayesh @stefannica , I wish we would have had this a long time ago, would have saved us many headaches from failed migrations 🎉

strickvl:
@coderabbitai review

src/zenml/cli/base.py
```diff
@@ -659,3 +659,113 @@ def migrate_database(skip_default_registrations: bool = False) -> None:
...
+        assert isinstance(store, SqlZenStore)
+        store.restore_database(
+            strategy=DatabaseBackupStrategy(strategy) if strategy else None,
+            location=location,
+        )
+        cli_utils.declare("Database restore finished.")
+    else:
+        cli_utils.warning(
+            "Cannot restore database while connected to a ZenML server."
+        )
```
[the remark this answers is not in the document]
stefannica: nope. This is merely for mypy checks. It is otherwise guaranteed to happen.

src/zenml/zen_stores/migrations/utils.py
```diff
@@ -0,0 +1,584 @@
...
+        if self._master_engine is None:
+            self._master_engine = self.create_engine()
+        return self._master_engine
+
+    def create_drop_mysql_database(
+        self,
+        database: Optional[str] = None,
+        create: bool = True,
+        drop: bool = False,
+    ) -> None:
```
AlexejPenner, line 97: I'm a bit confused why create and drop are sharing the same method. The two methods only really share 2 lines of code.

src/zenml/zen_stores/migrations/utils.py
```diff
@@ -0,0 +1,584 @@
...
+                    dict(table=table.name, create_stmt=create_table_stmt)
+                )
+
+                # 2. extract the table data in batches
+
+                # If the table has a `created` column, we use it to sort
+                # the rows in the table starting with the oldest rows.
+                # This is to ensure that the rows are inserted in the
+                # correct order, since some tables have inner foreign key
+                # constraints.
```
AlexejPenner, line 173: Is this strategy guaranteed to avoid key constraint failures, or does it just make failure less likely?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -416,6 +423,41 @@ def _remove_grpc_attributes(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                "The backup database name must be set if the backup strategy "
+                "is set to use a backup database."
+            )
+
+        if backup_strategy == DatabaseBackupStrategy.DUMP_FILE and (
+            not values.get("backup_directory")
+        ):
+            values["backup_directory"] = os.path.join(
+                GlobalConfiguration().config_directory,
+                SQL_STORE_BACKUP_DIRECTORY_NAME,
```
AlexejPenner, line 456: should this also maybe validate that there are sufficient permissions to write to this directory?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1060,22 +1252,96 @@ def migrate_database(self) -> None:
...
+                    logger.info("Backing up the database before migration.")
+                    (
+                        backup_location_msg,
+                        backup_location,
+                    ) = self.backup_database()
+                except Exception:
+                    logger.exception(
+                        "Failed to backup the database. Please check the logs "
+                        "for more details. The upgrade will continue as usual "
+                        "but no recovery is possible if something goes wrong."
```
AlexejPenner, line 1299: Hmm, so even if I configure a backup strategy, I might have misconfigured something, and the migration will still continue and possibly fail, without any recourse? Can't we make it fail here, and give users a overrride strategy?

AlexejPenner (commented):
Did not expect this level of complexity with 3 different backup strategies. I left some questions/concerns. @stefannica

src/zenml/zen_stores/migrations/utils.py
AlexejPenner: I saw further down, that this method can do both in once go. I still feel like this should be two separate methods. Feel free to ignore my comment though.

src/zenml/zen_stores/migrations/utils.py
stefannica: generally speaking, not all tables have a `created` column, so I'd say it makes it less likely. But for tables that do have a `created` column, this is very near to guaranteeing it.

src/zenml/zen_stores/sql_zen_store.py
stefannica: we validate it by writing to it

src/zenml/zen_stores/sql_zen_store.py
stefannica: done

src/zenml/zen_stores/migrations/utils.py
stefannica: done

AlexejPenner (approved):
:seal:

stefannica (approved):
Looks good to me, for what that's worth

PR #2174, opened 2023-12-20
Remove space saver action + upgrade other actions

strickvl:
Overdue to upgrade / update some of our really ancient or deprecated actions. 

Also since we're on custom runners for ubuntu I removed the special action that was clearing space on the runner. (Should save a bit of time).

## Summary by CodeRabbit

- **Chores**
  - Updated GitHub Actions to newer versions across various workflow files for improved performance and security.


strickvl:
@coderabbitai review

.github/actions/setup_environment/action.yml
```diff
@@ -29,7 +29,7 @@ runs:
   using: "composite"
   steps:
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5.0.0
```
avishniakov, line 32: Here and in other places, why you are so specific as `5.0.0`? Why not just rely on majors like `5`?

avishniakov (approved):
One nitpick, rest good to go!

.github/actions/setup_environment/action.yml
strickvl: AFAIK you specify the literal tag name here and if there isn't a `v5` tag name it fails.

PR #2190, opened 2023-12-22
Fix API docs environment setup

strickvl:
Temporary fix for API building process using @AlexejPenner's pip lockfile and list of requirements.

At some point (soon?) the requirements file will get out of date so we need a better way of installing all our requirements in a single environment.

https://github.com/zenml-io/zenml/actions/runs/7298987028 shows it tested + working (with the 'publish / push the built docs' part commented out obv)


avishniakov (approved):
The file might be worth renaming to something more abstract, such as `requirements-docs-frozen.txt` 🙂

PR #2192, opened 2023-12-22
Prune of unused artifacts links via client

avishniakov:
## Describe changes
I implemented pruning of unused artifacts and their versions via Client to bring it in sync with CLI.

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
  - Introduced an artifact cleanup tool to remove unused artifacts and their versions, helping maintain a clutter-free environment.

- **Tests**
  - Added new tests to ensure the artifact pruning functionality works as expected.


avishniakov:
@coderabbitai review

tests/integration/functional/test_client.py
```diff
@@ -1128,6 +1129,17 @@ def test_basic_crud_for_entity(
             pass
 
 
+class TestArtifact:
+    def test_prune(self, clean_client: "Client"):
```
strickvl, line 1133: can you test that the two args for the function (deleting only artifacts and only_metadata) are also doing what they're meant to do. this test just tests it works generically, but we should test the edge cases too.

strickvl (changes requested):

src/zenml/client.py
```diff
@@ -2677,6 +2677,43 @@ def delete_artifact(
...
+                self.delete_artifact_version(
+                    name_id_or_prefix=unused_artifact_version.id,
+                    delete_metadata=not only_artifact,
+                    delete_from_artifact_store=not only_metadata,
+                )
+                unused_artifact = unused_artifact_version.artifact
+                if not unused_artifact.versions and not only_artifact:
+                    self.delete_artifact(unused_artifact.id)
+
+            except Exception as e:
```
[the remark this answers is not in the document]
avishniakov: Not really, it is for unexpected errors only, delete_artifact raises nothing. I can get rid of it and let exceptions flow uninterrupted also.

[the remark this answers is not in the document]
strickvl: Ah ok I thought maybe `delete_artifact` raised something defined. Then leave it as it is.

tests/integration/functional/test_client.py
avishniakov: add tests for that

tests/integration/functional/test_client.py
```diff
@@ -1138,6 +1138,28 @@ def test_prune(self, clean_client: "Client"):
...
+
+    def test_prune_only_artifact(self, clean_client: "Client"):
+        """Test that artifact pruning works with only artifacts flag."""
+        artifact_id = ExternalArtifact(value="foo").upload_by_value()
+        artifact = clean_client.get_artifact_version(artifact_id)
+        assert artifact is not None
+        clean_client.prune_artifacts(only_artifact=True)
+        assert clean_client.get_artifact_version(artifact_id).id == artifact.id
+        with pytest.raises(FileNotFoundError):
+            os.listdir(artifact.uri)
```
strickvl, line 1162:
```suggestion
        assert not os.path.exists(artifact.uri)
```

Maybe nicer?

tests/integration/functional/test_client.py
```diff
@@ -1138,6 +1138,28 @@ def test_prune(self, clean_client: "Client"):
...
+
+    def test_prune_only_metadata(self, clean_client: "Client"):
+        """Test that artifact pruning works with only metadata flag."""
+        artifact_id = ExternalArtifact(value="foo").upload_by_value()
+        artifact = clean_client.get_artifact_version(artifact_id)
+        assert artifact is not None
+        clean_client.prune_artifacts(only_metadata=True)
+        with pytest.raises(KeyError):
+            clean_client.get_artifact_version(artifact_id)
+        assert os.listdir(artifact.uri)
```
strickvl, line 1152:
```suggestion
        assert os.path.exists(artifact.uri)
```

strickvl (commented):
LGTM, but two tiny nits to make it a bit more readable

strickvl (approved):

src/zenml/client.py
```diff
@@ -2786,6 +2786,43 @@ def delete_artifact(
...
+        for unused_artifact_version in unused_artifact_versions:
+            try:
+                self.delete_artifact_version(
+                    name_id_or_prefix=unused_artifact_version.id,
+                    delete_metadata=not only_artifact,
+                    delete_from_artifact_store=not only_metadata,
+                )
+                unused_artifact = unused_artifact_version.artifact
+                if not unused_artifact.versions and not only_artifact:
+                    self.delete_artifact(unused_artifact.id)
```
stefannica, line 2818: Similar to the PR where you implemented the "delete all model version artifacts", there is a performance hit with this type of loop that ends up calling the REST API multiple times. You should consider implementing this as a REST API endpoint and looping in the SQL zen store instead (perhaps even deleting entities en-masse through a single SQL query, if possible). This opens the possibility of using this functionality through the dashboard.

stefannica (approved):
Just my usual performance related suggestions, but good to merge otherwise.

src/zenml/client.py
avishniakov: Hey, I'm half way through implementing this change, but stuck at the `verify_permissions_and_delete_entity`. It was rather clear what to check on single entry deletion, but here I prune Artifact, Artifact Version and Data at scale. Any thoughts? Shall I check for Admin permissions to do so?

stefannica: I only have a half-baked idea about how to get around this: maybe we need a new `prune` permission that you check globally at server level, as opposed to checking against every resource. But that requires more work, not only here, but probably also at Cloud API level.

@schustmi what do you think ?

PR #2193, opened 2023-12-22
Refactor secrets stores to store all secret metadata in the DB

stefannica:
## Describe changes
Refactor all secrets store to ONLY store secret values in the external back-end, as opposed to storing all the meta-information (UUID, name, user, workspace, create & update timestamps) in the external back-end.

* the secret CRUD methods are moved to the SQL/REST zen store
* the SQL zen store always keeps track of secrets in the ZenML database, but not their values
* the secrets store is now an implementation detail belonging to the SQL zen store
* the `list_secrets` method in particular _doesn't need to interact with the external secrets manager anymore_, which means that ZenML no longer needs "list all secrets" permissions from the Cloud Secrets Manager APIs.
* the AWS/GCP/Azure/HashiCorp Secrets Manager are only called to store, update and delete the secret key-values for the ZenML secrets.
* the SQL secrets store is a mere extension of the SQL Zen store now and will store/update/delete secret key-values in the ZenML database
* the REST secrets store has been completely removed

This change is fully backwards-compatible. All existing secrets stored in external secrets stores will be added to the DB as part of the DB migration, while their values and their external secret IDs will continue to be stored externally. Existing users will not be aware that this change even happened.

### TODO

* [x] DB migration script to populate the SQL DB for secrets currently stored in external back-ends

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **Documentation**
  - Updated the IAM user permissions guide for AWS S3 buckets to recommend specific rather than broad policies.

- **New Features**
  - Introduced a new secret management interface with enhanced validation and configuration options.
  - Added support for handling secret values updates and deletions more effectively.

- **Refactor**
  - Streamlined secret store configurations, focusing on environment-specific setups.
  - Removed deprecated methods and refactored existing ones for better secret handling across various platforms.

- **Bug Fixes**
  - Fixed issues related to secret value data types to ensure proper handling of optional and required fields.

- **Deprecations**
  - Deprecated certain methods in favor of new implementations for secret management.

- **Chores**
  - Cleaned up unnecessary imports and outdated references across multiple files related to secret stores.


stefannica:
@coderabbitai review

src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
```diff
@@ -86,6 +86,28 @@ class GCPUserAccountCredentials(AuthenticationConfig):
...
+            values: The configuration values.
+
+        Returns:
+            The validated configuration values.
+
+        Raises:
+            ValueError: If the user account credentials JSON is not a valid
+                JSON object.
+        """
+        if isinstance(values.get("user_account_json"), dict):
```
stefannica, line 105: This is just something left over from the "use service connectors to authenticate secrets stores" PR and revealed during testing this PR.

src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
```diff
@@ -146,6 +168,24 @@ class GCPServiceAccountCredentials(AuthenticationConfig):
...
+    ) -> Dict[str, Any]:
+        """Convert the service account credentials to JSON if given in dict format.
+
+        Args:
+            values: The configuration values.
+
+        Returns:
+            The validated configuration values.
+        """
+        if isinstance(values.get("service_account_json"), dict):
```
stefannica, line 183: This is just something left over from the "use service connectors to authenticate secrets stores" PR and revealed during testing this PR.

src/zenml/zen_stores/migrations/versions/7b651bf6822e_track_secrets_in_db.py
```diff
@@ -0,0 +1,94 @@
...
+
+revision = "7b651bf6822e"
+down_revision = "1ac1b9c04da1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    from zenml.config.global_config import GlobalConfiguration
```
schustmi, line 28: I commented something similar on a cloudapi PR before where we decided importing these classes in migrations is okay as it will just run once.

I think here it is a little more complicated: This migration will also run in `zenml 0.75.x` in a few months/years from now, which means we need to keep those classes and all their (deprecated) methods around. Do you think this will lead to any trouble in the future?

schustmi (changes requested):

src/zenml/zen_stores/migrations/versions/7b651bf6822e_track_secrets_in_db.py
stefannica: You're absolutely correct, we have to keep a lot of deprecated code around, even though it's only used in a DB migration script. Specifically, I'm talking about the deprecated "list_secrets" method in each secrets store implementation. We have two inter-related situations that I need to address here:

1. getting rid of deprecated code: this is a bit laborious, because I'd have to retest the migration again for all possible store types, but easy. I just have to move all those functions and their dependents inside the migration script.
2. getting rid of all imports in the migration script: technically impossible. These imports are strictly required to allow the migration script to connect to the external secrets store back-end. This needs to happen today, tomorrow, a few months from now or even a few years from now. As long as we have to support direct upgrades from 0.50.0 or earlier versions. On the bright side, we're only importing major concepts here: the global configuration and the SQL store. As long as we don't call attributes or methods prone to removal or deprecation on these objects (see the previous point), this should be safe.

schustmi: So the point is, we won't be able to remove those classes anyway as long as we want to support upgrading from pre-0.50 right? In that case I guess that's the best case we have, even though ideally this would not even be an automated migration in my opinion but rather something users do manually as there might be multiple reasons for this to fail.

schustmi (approved):

PR #2195, opened 2023-12-22
Force forward slashes when saving artifacts

christianversloot:
## Describe changes
When saving visualizations into S3 artifact stores, Windows-based pipeline runs generate backward slash based URIs because that's how `os.path.join` works on Windows. An example would be `s3://bucket\\visualize_principal_components\\output\\1a6d995d-1417-4c8c-89d0-99d0e5897d8e\\image_file.PNG`. Following that, loading the artifact from the artifact store fails because of a malformed key.

This PR attempts fixing the issue by forcing all paths to have forward slashes when saving artifacts.

* All URIs are forced to have forward slashes: those that are passed to `save_artifact` and when none are passed. 
* Windows and Linux based file systems seem to disallow creating files that have backslashes in their name, meaning that the risk of any failure modes being introduced there is _very low_ to _nonexistent_ (checked the approach with @strickvl).

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility of artifact saving across different operating systems by ensuring URIs use forward slashes.


strickvl:
@coderabbitai review

christianversloot:
Unfortunately that proved to be too simple: replacing the URI at that location would only result in replacing the base path, not the eventual join between the base path and the file name.

This, unfortunately, required adding replaces at materializer level.

strickvl:
Nice! This is potentially a breaking change, I think, specifically for anyone who already has an artifact saved that includes a backslash?

christianversloot:
Two more changes:

* Added the backslash conversion to `materializer_object.uri` as well, to ensure that the UI displays the path correctly:

image

* Added an assertion to the general unit tests for materializers writing visualizations, to assert that `\\` is not present in any of the URIs in other words requiring any materializer that writes visualizations to add backslash conversion.

christianversloot:
> Nice! This is potentially a breaking change, I think, specifically for anyone who already has an artifact saved that includes a backslash?

Not sure if breaking. Old pipeline runs with failing paths keep failing after applying this PR, only new ones are saved correctly. For example, this path was saved with the original PR contents (before adding backslash conversion to the materializers themselves); even though it is now working, the original path keeps failing, so strictly speaking no backwards incompatibility there:

```
File 's3://BUCKET/visualize_raw_data/output/9c1e5696-f324-41db-a3f0-23eb6cb6e590\image_file.PNG' does not exist in artifact store 'store'.
```

If you'd like to correct these as well (which I'm not sure of) we'll probably need to add a database migration which applies backslash conversion to the paths of these artifacts / visualizations as well.

strickvl:
@christianversloot no your right. I think no need to retrospectively migrate / correct things that are still working. They're just saved in the DB using the 'incorrect' naming but they'd still be accessible etc as you say.

strickvl:
@coderabbitai review

strickvl (approved):
Looks good to me and the test failures are known.

strickvl:
@christianversloot thanks for handling this!

PR #2212, opened 2024-01-03
A faster pandas materializer

htahir1:
## Describe changes

I create a faster pandas materializer that uses `snappy` and chunks to write to disk. This works better in cases where the dataframe is >10000 rows of data. I have ensured that it is backwards compatible and not a breaking change, however I have not tested the case of backwards compatibility myself (Some help here is appreciated).

This materializer has been tested manually on a S3 artifact store.

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
  - Introduced environment variables specific to the Pandas materializer for enhanced configuration.
  - Added support for chunked parquet file handling in the Pandas materializer, improving performance for large datasets.

- **Documentation**
  - Updated the advanced guide with instructions for setting environment variables in DockerSettings.
  - Corrected documentation links for better user navigation.

- **Bug Fixes**
  - Adjusted default compression settings for parquet files to optimize storage and access efficiency.

### DIFF
--- a/src/zenml/materializers/pandas_materializer.py
+++ b/src/zenml/materializers/pandas_materializer.py
@@ -20,10 +20,10 @@ import pandas as pd
 
 from zenml.enums import ArtifactType, VisualizationType
 from zenml.io import fileio
-from zenml.utils import io_utils
 from zenml.logger import get_logger
 from zenml.materializers.base_materializer import BaseMaterializer
 from zenml.metadata.metadata_types import DType, MetadataType
+from zenml.utils import io_utils
 
 logger = get_logger(__name__)
 
@@ -36,7 +36,7 @@ CHUNK_SIZE = os.getenv("ZENML_PANDAS_CHUNK_SIZE", 100000)
 CSV_FILENAME = "df.csv"
 
 
-class FastPandasMaterializer(BaseMaterializer):
+class PandasMaterializer(BaseMaterializer):
     """Materializer to read data to and from pandas fast."""
 
     ASSOCIATED_TYPES: ClassVar[Tuple[Type[Any], ...]] = (
@@ -89,10 +89,14 @@ class FastPandasMaterializer(BaseMaterializer):
                 dfs = []
 
                 # Read the length of the dataframe from the metadata file
-                len_df = int(io_utils.read_file_contents_as_string(self.metadata_path))
-                
+                len_df = int(
+                    io_utils.read_file_contents_as_string(self.metadata_path)
+                )
+
                 for i in range(0, len_df, CHUNK_SIZE):
-                    with fileio.open(f"{self.parquet_path}_{i//CHUNK_SIZE}", mode="rb") as f:
+                    with fileio.open(
+                        f"{self.parquet_path}_{i//CHUNK_SIZE}", mode="rb"
+                    ) as f:
                         dfs.append(pd.read_parquet(f))
 
                 # concatenate all the dataframes to one
@@ -158,13 +162,19 @@ class FastPandasMaterializer(BaseMaterializer):
 
         if self.pyarrow_exists:
             # Write the length of the dataframe to a file for later user
-            io_utils.write_file_contents_as_string(self.metadata_path, len(df))
-            
+            io_utils.write_file_contents_as_string(
+                self.metadata_path, str(len(df))
+            )
+
             # Write the dataframe to a parquet file in chunks
             for i in range(0, len(df), CHUNK_SIZE):
-                chunk = df.iloc[i:i+CHUNK_SIZE]
-                with fileio.open(f"{self.parquet_path}_{i//CHUNK_SIZE}", mode="wb") as f:
-                    chunk.to_parquet(f, compression=COMPRESSION_TYPE, engine="pyarrow")
+                chunk = df.iloc[i : i + CHUNK_SIZE]
+                with fileio.open(
+                    f"{self.parquet_path}_{i//CHUNK_SIZE}", mode="wb"
+                ) as f:
+                    chunk.to_parquet(
+                        f, compression=COMPRESSION_TYPE, engine="pyarrow"
+                    )
         else:
             with fileio.open(self.csv_path, mode="wb") as f:
                 df.to_csv(f, index=True)
--- a/src/zenml/materializers/pandas_materializer.py
+++ b/src/zenml/materializers/pandas_materializer.py
@@ -1,4 +1,4 @@
-#  Copyright (c) ZenML GmbH 2021. All Rights Reserved.
+#  Copyright (c) ZenML GmbH 2024. All Rights Reserved.
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


htahir1:
@coderabbitai review

schustmi (approved):

src/zenml/materializers/pandas_materializer.py
```diff
@@ -77,7 +82,36 @@ def load(self, data_type: Type[Any]) -> Union[pd.DataFrame, pd.Series]:
...
+                )
+
+                for i in range(0, len_df, CHUNK_SIZE):
+                    with fileio.open(
+                        f"{self.parquet_path}_{i//CHUNK_SIZE}", mode="rb"
+                    ) as f:
+                        dfs.append(pd.read_parquet(f))
+
+                # concatenate all the dataframes to one
+                df = pd.concat(dfs)
```
safoinme, line 103: does `df.concat(dfs)` uses memory? will it be a problem in the case of a very large dataframe?

safoinme (approved):

src/zenml/materializers/pandas_materializer.py
htahir1: Hmm wouldn’t be more of a problem than what was currently status quo :)

src/zenml/materializers/pandas_materializer.py
```diff
@@ -23,17 +23,21 @@
...
+COMPRESSION_TYPE = os.getenv("ZENML_PANDAS_COMPRESSION_TYPE", "snappy")
+# Get the chunk size from the environment variable
+CHUNK_SIZE = int(os.getenv("ZENML_PANDAS_CHUNK_SIZE", "100000"))
 
 CSV_FILENAME = "df.csv"
 
 
 class PandasMaterializer(BaseMaterializer):
-    """Materializer to read data to and from pandas."""
+    """Materializer to read data to and from pandas fast."""
```
bcdurak, line 40: 🤣

src/zenml/materializers/pandas_materializer.py
strickvl: @coderabbitai what do you think? is there a better way of doing this?

src/zenml/materializers/pandas_materializer.py
```diff
@@ -23,17 +23,21 @@
 from zenml.logger import get_logger
 from zenml.materializers.base_materializer import BaseMaterializer
 from zenml.metadata.metadata_types import DType, MetadataType
+from zenml.utils import io_utils
 
 logger = get_logger(__name__)
 
-PARQUET_FILENAME = "df.parquet.gzip"
-COMPRESSION_TYPE = "gzip"
+PARQUET_FILENAME = "df.parquet"
```
bcdurak, line 30: Doesn't this break the functionality where there is an artifact saved as a parquet file with the old materializer, and people upgrade their ZenML version?

In more detail, in this case, the `load` function would check:
- if the metadata file exists and would skip,
- then it would check whether the parquet file exists and since the filename has changed would skip again,
- finally, would try to load a csv file and fail.

bcdurak (changes requested):
Left a comment regarding the backward compatibility.

src/zenml/materializers/pandas_materializer.py
htahir1: No guys we dont need to touch this, of course we have to hold the dataframe in memory... thats what materialization means right?

src/zenml/materializers/pandas_materializer.py
htahir1: > then it would check whether the parquet file exists and since the filename has changed would skip again,

It would check if the parquet files exists and would not skip as it would exist

bcdurak: Since it was generated by the old materializer, the filepath of the old artifact would be `self.uri/df.parquet.gzip`. But the new materializer will look for it on `self.uri/df.parquet`. or am I missing something?

strickvl:
@htahir1 My bigger question is around the specific use case that triggered this PR. In some tests I ran, materializing (saving and loading), was twice as slow on this branch when compared to what we already have. (Details in the table, but the key comparison is 28-million-row dataframe materialized to a GCP artifact store). (It works a *little* bit faster in the local artifact store case, which is what we'd want / expect).

CleanShot 2024-01-04 at 13 56 11@2x

Probably the repeated saving (at the 100K chunk size that we specify here) that causes the slowdown. In any case, presumably we want this to work faster in the remote artifact store case?

The code used for these tests:

```python
import pandas as pd
import numpy as np
from datetime import datetime, timedelta


@step(enable_cache=False)
def save_df(num_rows: int = 10000) -> pd.DataFrame:
    """Reads the data from the source."""
    data = {
        "ID": range(1, num_rows + 1),
        "Value": np.random.rand(num_rows),  # Random float values
        # Create dates cycling through a year to avoid 'date value out of range' error
        "Date": [datetime.now() - timedelta(days=x % 365) for x in range(num_rows)],
    }
    return pd.DataFrame(data)


@step(enable_cache=False)
def read_df(df: pd.DataFrame) -> pd.DataFrame:
    """Reads the data from the source."""
    return df


@pipeline(enable_cache=False)
def my_pipeline():
    # df = save_df()  # 10k rows
    # df = save_df(num_rows=1000000)  # 1M rows
    df = save_df(num_rows=28000000)  # 28M rows
    reloaded_df = read_df(df=df)


my_pipeline()
```

strickvl:
Parallelising the chunk saving (i.e. fanning out the saving of the artifacts to multiple threads) brings a decent improvement to the process, though then we run the risk of overwhelming the users' machines. WDYT?

CleanShot 2024-01-04 at 14 37 54@2x

```python
def save_chunk(self, df, start, chunk_size, path, compression):
        """Save a chunk of the dataframe to a parquet file."""
        chunk = df.iloc[start : start + chunk_size]
        with fileio.open(path, mode="wb") as f:
            chunk.to_parquet(f, compression=compression)

    def save(self, df: Union[pd.DataFrame, pd.Series]) -> None:
        """Writes a pandas dataframe or series to the specified filename in parallel chunks."""
        if isinstance(df, pd.Series):
            df = df.to_frame(name="series")

        if self.pyarrow_exists:
            # Write the length of the dataframe to a file for later use
            io_utils.write_file_contents_as_string(
                self.metadata_path, str(len(df))
            )

            # Define the maximum number of threads
            # Fallback to 10 if cpu_count is None
            max_workers = os.cpu_count() or 10

            # Prepare the thread pool for parallel writes
            with concurrent.futures.ThreadPoolExecutor(
                max_workers=max_workers
            ) as executor:
                futures = []
                for i in range(0, len(df), CHUNK_SIZE):
                    chunk_path = f"{self.parquet_path}_{i//CHUNK_SIZE}"
                    # Submit the chunk-saving task to the executor
                    futures.append(
                        executor.submit(
                            self.save_chunk,
                            df,
                            i,
                            CHUNK_SIZE,
                            chunk_path,
                            COMPRESSION_TYPE,
                        )
                    )

                # Wait for all futures to complete
                for future in concurrent.futures.as_completed(futures):
                    # Check for exceptions and handle them if necessary
                    future.result()
        else:
            with fileio.open(self.csv_path, mode="wb") as f:
                df.to_csv(f, index=True)
```

strickvl:
Reducing the number of workers to be no more than the # of CPU cores on the machine (i.e. 16 on my machine) made it even faster. So almost at twice as fast as the original implementation on this PR, but it's still quite a bit slower than the original `gzip` implementation.

CleanShot 2024-01-04 at 14 59 30@2x

htahir1:
Closing due to @strickvl 's results, lets table it

PR #2236, opened 2024-01-08
Adjust migration settings based on database engine

strickvl:
This pull request adjusts the migration settings based on the requirements of the database engine. It fixes the primary key requirement for MySQL and MariaDB engines.

We want to run DB migration testing for MariaDB following #2155, but before we can do that, we need to fix this one migration which is incompatible with MariaDB. I added a check for mariadb and execute a similar setting of the variable, but note that it is [only able to be set globally](https://jira.mariadb.org/browse/MDEV-28412).

## Summary by CodeRabbit

- **Refactor**
  - Improved database schema to enhance data management.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
   update-templates-to-examples:
     # this doesn't work on forked repositories (i.e. outside contributors)
     # so we disable template updates for those PRs / branches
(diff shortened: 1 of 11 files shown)


strickvl:
@coderabbitai review

schustmi (approved):

safoinme (approved):

PR #2274, opened 2024-01-12
Fast/slow CI core scaffold

strickvl:
Restructures / updates the CI, splitting it into two parts:

- fast CI (running by default / as usual) which uses `ubuntu-dind-runners` and Python 3.10. Removes some smaller checks.
- slow CI (currently disabled pending https://zenml.atlassian.net/browse/OSSK-344 implementation) runs the full matrix of tests; smaller checks split up so that they run faster.

Fast CI [runs in 36 mins](https://github.com/zenml-io/zenml/actions/runs/7502791455?pr=2274), which will speed up with sharding.
Slow CI [takes a good deal longer](https://github.com/zenml-io/zenml/actions/runs/7502791462?pr=2274) but that's as slow as the availability of MacOS and Windows runners, so...


.github/workflows/ci-fast.yml
```diff
@@ -3,72 +3,42 @@ name: Fast CI
...
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-dind-runners
     steps:
       - name: Checkout code
         uses: actions/checkout@v4.1.1
       - name: Set up Python
         uses: actions/setup-python@v5.0.0
         with:
-          python-version: '3.8'
+          python-version: '3.10'
```
avishniakov, line 26: Why not `3.11`? It is reported to be faster than `3.10` in general

avishniakov (approved):
Good one, just a question about `3.10` instead of `3.11`

.github/workflows/ci-fast.yml
avishniakov: Sure, go ahead with 3.10

safoinme (approved):

PR #2280, opened 2024-01-15
Fix `yamlfix` script to use `--no-yamlfix` flag

strickvl:
Script wasn't handling the case when extra `SRC` values were passed in. Updated and works / tested locally.

Also updated some related comments in the `CONTRIBUTING.md` document for OSS contributors.


avishniakov (approved):
LGTM, assuming that linting pass on Windows runners (please wait for it to happen)

strickvl:
@avishniakov it passed all unit tests, so working on all OSes.

PR #2271, opened 2024-01-12
Allow `sklearn` versions > 1.3

Vishal-Padia:
## Describe changes
I fixed #2216 to achieve that sklearn version > 1.3 is required to be installed.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [x] If my change requires a change to docs, I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] I have based my new branch on `develop` and the open PR is targeting `develop`. If your branch wasn't based on develop read [Contribution guide on rebasing branch to develop](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md#-pull-requests-rebase-your-branch-on-develop).
- [ ] If my changes require changes to the dashboard, these changes are communicated/requested.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

The changes in `tests/conftest.py` on lines 133 and 139 are just formatting changes that happen when I save the file in VSCode, I haven't changed anything inside it.
The change in `tests/conftest.py` is that I removed the deprecated `execfile` which was replaced with `File` in the new version.

## Summary by CodeRabbit

- **Documentation**
	- Updated the user guide to reflect the new command for installing the latest version of `sklearn`.

- **New Features**
	- Adjusted the `sklearn` integration to support versions above 1.3.

- **Tests**
	- Enhanced the testing configuration to align with updated dependencies and environment management.

PR #2281, opened 2024-01-15
Free `sklearn` dependency to allow all versions

strickvl:
I merged #2271 without checking / noting that a key change request wasn't actually made. This makes the fix for the dependency as well as the docs change which also wasn't added.

Regarding CI: already tested and passed on #2271.


avishniakov (approved):
🦭

PR #2317, opened 2024-01-18
Add conditional checks to prevent scheduled actions running inside forked repositories

strickvl:
Some of our scheduled actions are currently running on forked repositories. They are failing (mainly due to missing passwords etc) but still we should confine these to just running from within ZenML. Added some other conditionals in places where it seemed to make sense that we'd want to prevent these from running outside as well.

## Summary by CodeRabbit

- **Chores**
  - Improved workflow conditions across various GitHub Actions to ensure they run exclusively for the primary repository (`zenml-io/zenml`).


strickvl:
@coderabbitai review

avishniakov (approved):

PR #2333, opened 2024-01-22
Add warning for GCP integration when using Python >=3.11

strickvl:
Added warnings to docs where GCP integration is mentioned in a significant way.


htahir1 (approved):
Small nit - can we also do this in code?

strickvl:
> Small nit - can we also do this in code?

How would that work @htahir1? Check if GCP is one of the integrations included when someone runs `zenml integration install` and then output some message then?

strickvl:
@htahir1 added something along those lines

htahir1 (approved):

PR #2334, opened 2024-01-22
Add repeating tests and instafail error logging to testing in CI

strickvl:
Added two dev dependencies and two new behaviours to our test suite:

- `pytest-instafail` which outputs the logs / errors / stacktrace etc immediately when a test fails (instead of current behaviour which waits till the end of the entire test suite run to log what went wrong.
- `pytest-rerunfailures` which reruns failed / failing tests 3 times when they fail. We can limit this to specific tests if this ends up leading to unwanted reruns (for obvious failures that don't need to be rerun) but currently it's set to do this whenever any test fails.


safoinme (approved):
Great One

PR #2343, opened 2024-01-23
Use GitHub as trusted publisher for PyPI publication

strickvl:
Poetry publication using a static username and password is no longer possible. This PR makes the necessary fixes for generating a one-time token to enable publication using Poetry. See https://github.com/marketplace/actions/pypi-token-mint for more.

At some point in the future we might want to do it fully in [the way that PyPi recommend](https://docs.pypi.org/trusted-publishers/using-a-publisher/) but for now this should get us back in business quickly.


strickvl:
@coderabbitai

strickvl:
@coderabbitai review

safoinme (approved):
LGTM

PR #2360, opened 2024-01-25
Mocked API docs building

avishniakov:
## Describe changes
**Added**:
- The core script is `docs.sys_modules_mock` - in check mode it tries to import all ZenML code to verify that all mockers are in place in clean env. On import, it substitutes all imports mentioned in `mockers` with a custom class suitable for mocked docs building. It is imported in `mkdocs.yml` for docs building.
- Step in Fast CI to check if all necessary objects are added to the mocked list. I can also automate this, but for now, I would prefer to see it semi-automated. You will get instructions on what to add to the mockers list to overcome failures. If it fails, you will see something like this: https://github.com/zenml-io/zenml/actions/runs/7724622397/job/21057142169?pr=2360#step:4:1366. Just go ahead and paste this suggested list in `docs/mocked_libs.json`, if you find that this is expected.

**Changed**:
- I sorted items on the main page for consistency.
- No integrations are needed to build docs as of now - only `pip install -e ".[server,dev]"` is executed
- I updated `.github/workflows/publish_api_docs.yml` to use Python 3.11 and skip environment provisioning since generation script is now self-inclusive

**Observed deviations in docs vs current version**:
- Obviously, ordering in `index.html` and `404.html` due to introduced sorting
- Some objects got a bit more documentation (like `StrEnum` has `__format__` now documented) - I don't see this to be an issue.
- Class names used in typing for mocked imports might vary. I tried to overcome this, but it is impossible since originally it was using internal information of imported integrations, which is not accessible anymore. Generally, it is more full compared to the previous state, so I don't see this as a showstopper as well. Few examples:
```diff
"integration_code_docs/integrations-lightgbm/index.html"
-      <td><code>Booster</code></td>
+      <td><code>lightgbm.Booster</code></td>
-        <td><code>Booster</code></td>
+        <td><code>lightgbm.Booster</code></td>
-        <td><code>Dataset</code></td>
+        <td><code>lightgbm.Dataset</code></td>
-      <td><code>Dataset</code></td>
+      <td><code>lightgbm.Dataset</code></td>
-        <td><code>Dataset</code></td>
+        <td><code>lightgbm.Dataset</code></td>

"integration_code_docs/integrations-gcp/index.html"
-      <td><code>GCSFileSystem</code></td>
+      <td><code>gcsfs.GCSFileSystem</code></td>
-      <td><code>FunctionServiceClient</code></td>
+      <td><code>google.cloud.functions_v2.FunctionServiceClient</code></td>
```

P.S. Changes in `src/zenml/materializers/base_materializer.py` looks weird in GitHub, but it is just top `if` added.

## Pre-requisites

## Types of changes


strickvl:
@coderabbitai review

docs/sys_modules_mock.py
```diff
@@ -0,0 +1,307 @@
...
+            root = root[len("src/") :]
+
+            no_meaningful_files = len(subdirs) == 0
+            for file in files:
+                for mf in meaningful_files:
+                    if file.endswith(mf):
+                        no_meaningful_files = False
+                        break
+                if file.endswith(".py") and not file == "__init__.py":
+                    module_name = root.replace("/", ".") + "." + file[:-3]
```
strickvl, line 281:
```suggestion
                    # removes '.py'
                    module_name = root.replace("/", ".") + "." + file[:-3]
```

Maybe?

docs/sys_modules_mock.py
```diff
@@ -0,0 +1,307 @@
...
+                    try:
+                        importlib.import_module(module_name)
+                    except ModuleNotFoundError as e:
+                        msg: str = e.args[0]
+                        module = msg[msg.find("'") + 1 :]
```
strickvl, line 286: What's the +1 doing here? @coderabbitai  any ideas?

Maybe add a comment just to make it a bit clearer?

scripts/generate-docs.sh
```diff
@@ -63,49 +65,38 @@ rm docs/mkdocs/index.md || true
...
-  # zenml integration install -y --ignore-integration feast --ignore-integration label_studio --ignore-integration kserve --ignore-integration airflow --ignore-integration bentoml
-  # pip install jinja2==3.0.3 protobuf==3.20.0 numpy~=1.21.5
-  # pip install typing-extensions --upgrade
-  # pip install feast --upgrade  # The integration feast version has unsupported googleapis-common-protos >=1.52.* requirement
-  
-  # TEMPORARY FIX
-  pip install -e ".[server,dev,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs]"
-  pip install -r docs/requirements-docs-frozen.txt
+  pip3 install -e ".[server,dev]"
+  pip3 install "Jinja2==3.0.3"
```
strickvl, line 69: is it so precarious that we have to pin it so specifically?

docs/sys_modules_mock.py
avishniakov: Find function return a pos of a symbol, e.g. in `"str!payload".find("!")==4` and I need `payload`, so have to do `[4+1:]`

scripts/generate-docs.sh
avishniakov: I can test without it, but failed on local clean env always

docs/sys_modules_mock.py
avishniakov: I will add a comment

scripts/generate-docs.sh
avishniakov: Yeah, collapses without pinned version: `error: cannot import name 'contextfunction' from 'jinja2' (/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/jinja2/__init__.py)` same as on my local

strickvl (approved):

.github/workflows/publish_api_docs.yml
```diff
@@ -26,14 +26,10 @@ jobs:
...
-        uses: ./.github/actions/setup_environment
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5.0.0
         with:
-          cache_version: ${{ secrets.GH_ACTIONS_CACHE_KEY }}-api-docs
-          python-version: '3.8'
-          os: ubuntu-latest
-          # We need to install integrations in a different way to avoid conflicts
-          install_integrations: 'no'
+          python-version: '3.11'
```
AlexejPenner, line 32: is there a reason that we use python 3.11?

docs/sys_modules_mock.py
```diff
@@ -0,0 +1,114 @@
+import sys
+from unittest.mock import MagicMock
+import json
+from pathlib import Path
+
+mockers = json.load(open(Path(__file__).parent.resolve() / "mocked_libs.json"))
+
+
+class DocsMocker(MagicMock):
```
AlexejPenner, line 9: Would love to have some docstrings in this file to make it a bit easier to understand

docs/sys_modules_mock.py
```diff
@@ -0,0 +1,114 @@
...
+            "Consider mocking : \n["
+            + ", ".join([f'"{mm}"' for mm in missing_modules])
+            + "]\n  To do so, add this list entries to `mocked_libs.json` "
+            "in `docs` folder and retry."
+        )
+        logger.error(msg)
+
+    if empty_dirs:
+        logger.warning(
+            f"Directories without any 'meaningful_files' {meaningful_files} files detected. This might affect docs building process:",
```
AlexejPenner, line 107: I think its not fully clear here what a "meaningful_file" is

AlexejPenner (commented):
Very excited for this

.github/workflows/publish_api_docs.yml
avishniakov: It is the fastest, any objections? No deps installed, so I pick fastest.

strickvl: Given how often we return to fix these things, I'm happy to keep it a future-facing version as it'll be a while (hopefully) before we return to update this functionality / code.

src/zenml/materializers/base_materializer.py
```diff
@@ -51,52 +51,52 @@ def __new__(
...
-            )
-
-        # Validate associated artifact type.
-        if cls.ASSOCIATED_ARTIFACT_TYPE:
-            try:
-                cls.ASSOCIATED_ARTIFACT_TYPE = ArtifactType(
-                    cls.ASSOCIATED_ARTIFACT_TYPE
-                )
-            except ValueError:
+        if not cls._DOCS_BUILDING_MODE:
```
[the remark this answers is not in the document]
avishniakov: Need to think about it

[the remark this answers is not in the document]
avishniakov: So far only came up with quite dirty hack, IMO:
```python
base_materializer_meta_hack = '''
class BaseMaterializerMeta(type):
    def __new__(mcs, name, bases, dct):
        return cast(
            Type["BaseMaterializer"], super().__new__(mcs, name, bases, dct)
        )
'''

with open("src/zenml/materializers/base_materializer.py", "r") as f:
    base_materializer = f.readlines()
    start_ln,end_ln = -1,-1
    for ln,l in enumerate(base_materializer):
        if "class BaseMaterializerMeta(type)" in l:
            start_ln = ln
        if start_ln>=0 and "return cls" in l:
            end_ln = ln
            break
    base_materializer_modified = base_materializer[:start_ln] + [l+"\n" for l in base_materializer_meta_hack.split("\n")] + base_materializer[
        end_ln+1:
    ]
with open("src/zenml/materializers/base_materializer.py", "w") as f:
    f.writelines(base_materializer_modified)
```

It works but creates debt for future maintainers. 
@AlexejPenner , if this is better, I will run a full-blown validation of generated docs once again tomorrow.

.github/workflows/publish_api_docs.yml
AlexejPenner: no objection, was just surprised as I forgot that we already support 3.11

src/zenml/materializers/base_materializer.py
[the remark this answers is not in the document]
AlexejPenner: I think you're right, your first implementation of this is much easier to maintain. Lets keep it as is for the time being, we can always come back at a later point to remove it from our core code.

docs/mkdocs.yml
```diff
@@ -19,6 +19,8 @@ plugins:
       default_handler: python
       handlers:
         python:
+          setup_commands:
+            - import docs.sys_modules_mock
```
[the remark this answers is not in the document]
AlexejPenner: Ah, ok, now I understand, thanks

AlexejPenner (approved):
Once the docstrings have made it in, i think this is ready to go :)

PR #2363, opened 2024-01-25
Fix CI issues

strickvl:
- Slow CI now runs when someone makes a PR 'ready for review'. (this requires making it a draft and then setting it as 'ready'
- removed all the code relating to comment-driven triggering of CI (freeing up unit tests etc to work freely again)


avishniakov (approved):
Oh, all checks are collected in one place - beautiful!

Issue #2211, opened 2024-01-03
Fix step logging when using GCS Artifact Store


strickvl:
## Open Source Contributors Welcomed!
Please comment below if you would like to work on this issue!

### Contact Details [Optional]

support@zenml.io

### What happened?
There seems to be an issue with StepLogging when using GCS (Google Cloud Storage) as the artifact store. Specifically, only the last parts of the logs appear in the file, which suggests a problem with the log writing or saving mechanism.

### Steps to Reproduce
Here's a snippet to reproduce the issue:

```python
import gcsfs
from zenml.client import Client
from zenml.logging.step_logging import StepLogsStorage

client = Client()
_ = client.active_stack

TEST_FILE="gs://<<your_bucket>>/test_log.log"

log_storage = StepLogsStorage(logs_uri=TEST_FILE, max_messages=5)
for i in range(0,11):
    log_storage.write(f"I'm log line #{i}")
log_storage.save_to_file()

fs = gcsfs.GCSFileSystem()
with fs.open(TEST_FILE, 'r') as f:
    all_of_it = f.read()

print(all_of_it)
```

### Expected Behavior
All log lines should be saved and visible in the GCS file, not just the last few.

### Potential Solution
Consider using the logging.StreamHandler facility to temporarily write logs to the remote file (GCS, S3, etc.). Here's an example:

```python
import logging
import fsspec

f = fsspec.open("gs://<<my_gcs_bucket>>/test_log.log", "w")
with f as of:
    log_handler = logging.StreamHandler(of)
    logger = logging.getLogger()  # Root logger
    logger.addHandler(log_handler)
    for i in range(0, 5000):
        logger.warning(f"I'm log line #{i}")
    logger.removeHandler(log_handler)
```

This approach could fit nicely in the `StepLogsStorageContext` class.

### Additional Context
Proper log handling is crucial for debugging and monitoring pipeline performance, especially when dealing with large-scale data processing in cloud environments.

### Code of Conduct
- [ ] I agree to follow this project's Code of Conduct

adtygan:
Hello @strickvl, I'm trying to reproduce this issue but can't. I made a GCS bucket and tried to run the first snippet and got the following error. Please let me know if you need the traceback.

`ValueError: No file systems were found for the scheme: gs://. Please make sure that you are using the right path and the all the necessary integrations are properly installed.` 

The error was raised for the following line,

`log_storage = StepLogsStorage(logs_uri=TEST_FILE, max_messages=5)`

[...]

PR #2366, opened 2024-01-26
Fix GCP step logging

adtygan:
## Describe changes
The issue arises because GCS artifacts are immutable [Stack Overflow thread](https://stackoverflow.com/questions/52715217/how-to-open-a-file-in-a-gcs-bucket-in-append-mode-using-python). To fix the issue, I rewrite the existing file with its old contents and buffer's content appended together.

Code to test the change: (Credits: @strickvl)

>Please use GCS stack

```
import gcsfs
from zenml.logging.step_logging import StepLogsStorage

client = Client()
_ = client.active_stack

TEST_FILE="gs://zenml-2211/test.txt"

log_storage = StepLogsStorage(logs_uri=TEST_FILE, max_messages=5)
for i in range(0,11):
    log_storage.write(f"I'm log line #{i}")
log_storage.save_to_file()

fs = gcsfs.GCSFileSystem()
with fs.open(TEST_FILE, 'r') as f:
    all_of_it = f.read()

print(all_of_it)
```

## Pre-requisites

## Types of changes


adtygan:
I accidentally included my GCP keys during the PR. I later made another commit to delete it. But GitGuardian still shows error. I have not worked on contributing to opensource before and so request some assistance on this.

htahir1 (commented):
Thanks for the PR!

Two questions

- Will this effect other artifact stores like S3?
- Is this still performant? Somehow I feel like we are doing too many IO operations... Maybe we should slow it down a bit?

adtygan:
Thanks for reviewing @htahir1 !

- **Will this effect other artifact stores like S3?**
I do not think so. This is because the existing solution appended the logs to the file. The proposed solution overwrites existing file with updated content and hence should work.
- **Is this still performant? Somehow I feel like we are doing too many IO operations... Maybe we should slow it down a bit?**
I did try to think of better solutions. The main limitation of GCS files being immutable lead me this option. Could you suggest any alternative approaches that I could try?

Thanks

htahir1:
@adtygan As the keys remain git hit history, please invalidate them from GCP ASAP to make it safe for your cloud.

For the performance, maybe a good way to do it would be to benchmark some running pipelines with varying logs. I have noticed if you use a [rich ](https://rich.readthedocs.io/en/stable/progress.html) or [TQDM progress bar](https://www.analyticsvidhya.com/blog/2021/05/how-to-use-progress-bars-in-python/) it slows down  a LOT, and id love some benchmarks on the local store vs the GCS store for varying scripts :-)

adtygan:
Thanks for the suggestion @htahir1 , I have invalidated my key. With regard to benchmarks, please give me some time. I will get back on this and update you.

adtygan:
Hello @htahir1 , I want to confirm with you if I understand what you said correctly. I'm planning to measure the running time for logging 100, 1,000 and 10,000 lines. For each of these, I'm going to measure the running times for local, GCP, local with TQDM, GCP with TQDM. In total this should give 12 run time values. I need to measure these 12 values for the current version of the code and my PR version. 

Is this correct? Thanks.

htahir1:
Yes this is correct!

Hamza Tahir
Co-Creator & CTO

[image: ZenML] <https://zenml.io/>

 Github  <https://github.com/zenml-io/zenml> Twitter
<https://twitter.com/zenml_io> Linkedin <https://linkedin.com/company/zenml>
ZenML Inc./GmbH, Schellingstr. 36, 80799 Munich
HRB Munich 268487, MD/GF: Adam Probst, Hamza Tahir

adtygan:
I did some parts of the bechmarking and noticed a bunch of issues.

Here is the details of the run for writing 100 lines of logs (averaged over 10 runs):

- My version on GCP stack: 27.807 seconds
- Develop branch version on GCP stack: (did not run this because it does not properly write logs)
- My version on local stack: (this was creating file size of >1 GB which I don't yet know why it is happening, on GCP stack it works fine)
- Develop branch version on local stack: 0.003 seconds 

I'm noticing 2 big issues with my fix
1. It is not fast enough 
2. It creates a huge file on local stack and I can't understand why it is doing.

I need some more time to look into this issue.
Thanks.

strickvl:
@adtygan note that you're getting some linting failures on the CI. if you could fix those as well that'd be great!

adtygan:
Hello @strickvl , I don't think my current code can be optimized to improve performance. Instead, I checked the Potential Solution you had mentioned in the initial post of the issue (https://github.com/zenml-io/zenml/issues/2211#issue-2063754569). This option looks like the best choice. However, I want to clarify how to go about incorporating it.

If I understand correctly, you are suggesting to open the log file in write mode and then proceed with logging. This would write all the contents. I have tested on GCP stack and it works. But the only issue I realize is it is going to overwrite past logs.

Can I work on a solution where we create a temporary file to store the logs, and then using the __exit__() method append this file's contents to the main log file? 

Thanks

htahir1:
@adtygan this sounds like a reasonable plan to try out! Would love to see how this new approach would benchmark against the old one

adtygan:
Sorry @htahir1 , I took a break from work for a few weeks and did not keep you updated. I will get back to working on the issue.

adtygan:
Hello @htahir1  and @strickvl , I have opened a PR (https://github.com/zenml-io/zenml/pull/2533). Please review it. To the best of my knowledge, I think this does not have any errors.

PR #2371, opened 2024-01-29
Reflect env variables on global configuration

safoinme:
## Describe changes
Fix for the following issue:
image

### Details

This PR implements an environment variable overlay that is applied on top of the current store configuration, regardless of whether it is loaded from the global configuration file or explicitly configured using `set_store`. The environment variables are used to either replace the active store configuration or merged on top of them, depending on how complete they are.

Downside: if any environment variables are used to configure the zen store, they will also be reflected in the configuration file, thus replacing what was there before. There is no easy way to prevent this, because we rely on the actual running store configuration to be dumped back into the configuration file for persistence purposes (e.g. to persist the API token).

More changes included to clean up the already overloaded global configuration:
* got rid of `copy_configuration` methods that are no longer used
* got rid of any attributes and methods in the global configuration class that are no longer needed
* implemented a `store_configuration` property that is the source of truth regarding the actual store configuration, with env variables applied and used it to replace any references to the `.store` property that is not always as accurate and does not include environment variable overrides.
* moved the method that converts the global config into env variables (i.e. the opposite operation) into the global config class  

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **Refactor**
	- Improved the logic for accessing the Zen store, ensuring more reliable and direct checking for store availability.


safoinme:
@coderabbitai review

src/zenml/config/global_config.py
```diff
@@ -669,11 +669,8 @@ def zen_store(self) -> "BaseZenStore":
         Returns:
             The current zen store.
         """
-        if not self.store:
+        if self._zen_store is None:
```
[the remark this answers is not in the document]
safoinme: does adding a check on `self.store` after env variables solve the issue? since it will only go to return the `get_default_store_config` if no env variables where set

src/zenml/config/global_config.py
```diff
@@ -566,6 +566,8 @@ def get_default_store(self) -> StoreConfiguration:
             config = StoreConfiguration(
                 **env_store_config,
             )
+        elif self.store:
+            config = self.store
```
avishniakov, line 570: To add to Michael's comment: it seems to me that in this case either ENV go or File, but not a merger of them, isn't it?

safoinme: That's a good point! maybe I understood the ticket differently. In that case, I imagine we can do something like the following, where we see if self.config is set and then update config based on existing env variables. @schustmi @avishniakov 

```
# Update self.config based on environment variables
if self.config:
    # Update the store configuration from environment variables
    if env_store_config:
        if "type" not in env_store_config and "url" in env_store_config:
            env_store_config["type"] = BaseZenStore.get_store_type(env_store_config["url"])

        # Merge environment variables into self.config for store
        for key, value in env_store_config.items():
            self.config[key] = value

    # Update the secrets store configuration
    for key, value in env_secrets_store_config.items():
        self.config.secrets_store[key] = value

    # Update the backup secrets store configuration
    for key, value in env_backup_secrets_store_config.items():
        self.config.backup_secrets_store[key] = value

    logger.debug("Updated the configuration with environment variables.")
else:
    if len(env_store_config):
        if "type" not in env_store_config and "url" in env_store_config:
            env_store_config["type"] = BaseZenStore.get_store_type(env_store_config["url"])

        logger.debug("Using environment variables to configure the default store")

        self.config = StoreConfiguration(**env_store_config)
    else:
        self.config = BaseZenStore.get_default_store_config(
            path=os.path.join(self.local_stores_path, DEFAULT_STORE_DIRECTORY_NAME)
        )
```

avishniakov: Yeah, something like this might do. Can you write some tests to cover this use-case?

safoinme: @avishniakov Looking back on this, we can't merge. With the env variables, we need to initialize a new zen store which in this case will be the `RestZenStore` if we use the already existing one which is SQLZenStore config and we just updated (merge) to initialize the rest one it will fail because [rest zen store](https://github.com/zenml-io/zenml/blob/develop/src/zenml/zen_stores/rest_zen_store.py#L255) doesn't the same attributes as [sql one](https://github.com/zenml-io/zenml/blob/develop/src/zenml/zen_stores/sql_zen_store.py#L352) and each of these stores has different attributes besides type, URL, and SSL.

stefannica: I implemented a custom merge logic that in some cases replaces the config file store completely, in other cases it only overrides some of its attributes.

src/zenml/config/global_config.py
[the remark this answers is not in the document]
stefannica: Updated now to do the right thing: use the merged configuration to initialize the store.

avishniakov (approved):
🦭

stefannica (approved):
FWIW this looks good to me

PR #2372, opened 2024-01-29
HyperAI integration: orchestrator and service connector

christianversloot:
## Describe changes
HyperAI (hyperai.ai) has recently become one of our suppliers of GPU instances. Unfortunately, unlike the major public cloud providers, HyperAI does not yet have an SDK like `sagemaker`. Still, it is critical for us to keep using ZenML, as many of our pipelines are built that way. This PR implements a HyperAI integration by means of an orchestrator and service connector.

**Service connector:**

* The service connector is effectively a SSH based service connector. Using `paramiko`, it provides an authenticated `SSHClient` given the configured `ip_address` (with an optional `instance_name` serving as a nickname to distinguish multiple IP addresses), `username`, `base64_ssh_key` and optionally `ssh_passphrase`. 
* Unfortunately, the ZenML CLI does not support multiline entry via cli. That is why the key can be provided in Base64 encoded format; the service connector ensures that it is decoded.
* It has support for multiple key types: RSA, DSA (DSS), ECDSA and ED255519.

**The orchestrator:**

* Effectively uses Docker Compose and its `service_completed_successfully` `depends_on` condition to compose a Docker Compose file which guarantees the order of execution, including more complex pipelines by using `step.spec.upstream_steps`.
* Uses the provided SSH client to upload the file to the instance and then executes it (and hence assumes a very lightweight setup at the user side with Docker and Docker Compose being installed).
* Has support for scheduled pipelines (then also assumes Cron daemon is running).
* Has a built-in cleanup mechanism for non-scheduled pipeline runs, which are automatically deleted after 7 days upon starting a new pipeline run. Following the paradigm where users are responsible for cleaning up their schedule pipeline runs (per the ZenML docs), it does nothing with those.
* Optionally (via configuration, but set to False by default recognizing possible security implications) is able to authenticate the instance to the stack's configured container registry by logging in. Once again, this is entirely optional: if, from a security perspective, the user does not want this, they are free to leave it to False; then, they must ensure that the instance is logged in themselves.
* Allows for mounts to be made between folders on the instance and the container, hence allowing data stored on these GPU instances to be readily available to the pipeline run. Mounts that are made can be configured by the user in component configuration and can be changed at any time.

**Logo:**
The code assumes a HyperAI logo to be present in a seemingly public bucket on your end. I can ask the HyperAI team to provide a proper logo that can be put in this bucket so that it's visible within the ZenML dashboard.

This way:

* Users have full flexibility as to whether they want to create one service connector/orchestrator combination per instance; reuse service connectors (and thus keys) with multiple instances; even provide single users with different data access patterns by giving them different orchestrators with different mounts.
* Users have full freedom to deploy any Docker based pipeline using this orchestrator: as with SageMaker, the orchestrator fully respects the Docker configuration set by the user (in fact, I used the local Docker orchestrator to find inspiration).

Testing:

* I have added limited tests because I could not find many nor do I have an understanding about any testing infrastructure you may have.
* I did test thoroughly via a local ZenML / ZenML dashboard setup and it worked very well.

image

image

image

As discussed with @htahir1 , we're actively awaiting HyperAI usage (and so is the HyperAI team, as they've been in touch as far as I know) so we'd prefer to start using this integration as soon as possible. Do note that this does not mean that we should merge it recklessly, we're just very excited :) 

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
	- Introduced the HyperAI Service Connector for enhanced authentication options, supporting various key types.
	- Added comprehensive documentation for configuring HyperAI Connectors and orchestrating pipelines on HyperAI instances, including support for Docker Compose and scheduled runs.
	- Implemented the ZenML HyperAI orchestrator for deploying machine learning pipelines on HyperAI instances, with support for GPU-backed hardware via CUDA.
- **Documentation**
	- New guides and documentation for setting up and using the HyperAI Service Connector and orchestrator.
- **Tests**
	- Added tests for the HyperAI orchestrator to ensure correct attribute settings.


htahir1:
@coderabbitai review

htahir1:
Fantastic PR! Will get reviews out ASAP.

Immediate problem I see is the conflict with another integration: 

```
    The user requested paramiko>=3.4.0
    azureml-core 1.48.0 depends on paramiko<3.0.0 and >=2.0.8
```

Maybe lets pun paramiko to those?

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,399 @@
...
+                    file_path, password=ssh_passphrase
+                )
+
+            # Trim whitespace from the IP address
+            ip_address = str(self.config.ip_address).strip()
+
+            paramiko_client = paramiko.client.SSHClient()
+            paramiko_client.set_missing_host_key_policy(
+                paramiko.AutoAddPolicy()
+            )
```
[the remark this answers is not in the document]
christianversloot: This was added deliberately to ensure that ZenML can connect to instances seamlessly. I assume that we can consider the connection to be safe given that ZenML server should be safe and the user configured the instance themselves from within an access controlled environment.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,419 @@
...
+    """HyperAI orchestrator settings.
+
+    Attributes:
+        mounts_from_to: A dictionary mapping from paths on the HyperAI instance
+            to paths within the Docker container. This allows users to mount
+            directories from the HyperAI instance into the Docker container that runs
+            on it.
+    """
+
+    mounts_from_to: Dict[str, str] = {}
```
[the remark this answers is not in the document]
christianversloot: Added mount path validation in https://github.com/zenml-io/zenml/pull/2372/commits/850f7db95257c5989e04560f6fe5ce735f35b9bd.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,419 @@
...
+            # Get cron expression for scheduled pipeline
+            cron_expression = deployment.schedule.cron_expression
+            logger.info("Scheduling ZenML pipeline on HyperAI instance.")
+            logger.info(f"Cron expression: {cron_expression}")
+
+            # Create cron job for scheduled pipeline on HyperAI instance
+            stdin, stdout, stderr = paramiko_client.exec_command(
+                f"(crontab -l ; echo '{cron_expression} cd {directory_name} && docker compose up -d') | crontab -"
+            )
+
```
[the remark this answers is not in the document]
christianversloot: 1. They are, I believe. Little to no user input is required and all user input required needs to be configured via an access controlled interface.
2. Let's investigate.
3. In my tests, it is performant enough, so I'd propose no further optimization.
4. Let's investigate.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,399 @@
...
+                supports multiple instances.
+            kwargs: Additional implementation specific keyword arguments to use.
+
+        Raises:
+            NotImplementedError: If the connector auto-configuration fails or
+                is not supported.
+        """
+        raise NotImplementedError(
+            "Auto-configuration is not supported by the HyperAI connector."
+        )
```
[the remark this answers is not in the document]
christianversloot: Done.

christianversloot:
> Fantastic PR! Will get reviews out ASAP.
> 
> Immediate problem I see is the conflict with another integration:
> 
> ```
>     The user requested paramiko>=3.4.0
>     azureml-core 1.48.0 depends on paramiko<3.0.0 and >=2.0.8
> ```
> 
> Maybe lets pun paramiko to those?

Thanks Hamza!

Seemingly `azureml-core==1.54.0.post1` has support for `paramiko>3.0.0`:

```
pip show azureml-core
Name: azureml-core
Version: 1.54.0.post1
Summary: Azure Machine Learning core packages, modules, and classes
Home-page: https://docs.microsoft.com/python/api/overview/azure/ml/?view=azure-ml-py
Author: Microsoft Corp
Author-email:
License: https://aka.ms/azureml-sdk-license
Location: /home/chris/miniconda3/envs/paramikotest/lib/python3.11/site-packages
Requires: adal, argcomplete, azure-common, azure-core, azure-graphrbac, azure-mgmt-authorization, azure-mgmt-containerregistry, azure-mgmt-keyvault, azure-mgmt-network, azure-mgmt-resource, azure-mgmt-storage, backports.tempfile, contextlib2, docker, humanfriendly, jmespath, jsonpickle, knack, msal, msal-extensions, msrest, msrestazure, ndg-httpsclient, packaging, paramiko, pathspec, pkginfo, PyJWT, pyopenssl, python-dateutil, pytz, requests, SecretStorage, urllib3

Name: paramiko
Version: 3.4.0
Summary: SSH2 protocol library
Home-page: https://paramiko.org
Author: Jeff Forcier
Author-email: jeff@bitprophet.org
License: LGPL
Location: /home/chris/miniconda3/envs/paramikotest/lib/python3.11/site-packages
Requires: bcrypt, cryptography, pynacl
Required-by: azureml-core
```

I'm not sure if downgrading the `paramiko` requirement would be a great idea because in my view it would be best to keep up to date with existing packages. Hence, I would suggest to adapt the `azure` integration instead, requiring `azureml-core>=1.54.0.post1` or something.

WDYT?

htahir1:
@christianversloot Yes thats the ideal solution but then we'll have to test the upgrade on Azure: https://github.com/zenml-io/zenml/blob/main/src/zenml/integrations/azure/__init__.py#L41 

If you want to go down that route, I can provide an Azure subscription for you to test

christianversloot:
Is that a "press a button and go" thing or does it require many manual steps? If so, what steps?

htahir1:
Well... Not really "press a button".  The following components have to be tested"

https://docs.zenml.io/stacks-and-components/component-guide/artifact-stores/azure
https://docs.zenml.io/stacks-and-components/component-guide/container-registries/azure
https://docs.zenml.io/stacks-and-components/component-guide/step-operators/azureml

i'll let the others on the internal ZenML team decide how to proceed here... for now you can try upgrading it in this PR and see what happens in the tests

christianversloot:
Done, let's see :)

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+                        self._generate_valid_path_format(mount_from),
+                        self._generate_valid_path_format(mount_to),
+                    )
+                    for mount_from, mount_to in step_settings.mounts_from_to.items()
+                ],
+            }
+
+            # Add dependency on upstream steps if applicable
+            upstream_steps = step.spec.upstream_steps
+            if isinstance(upstream_steps, list) and len(upstream_steps) > 0:
```
schustmi, line 235: I think you can probably get rid of this check entirely, as
- pydantic makes sure this is a `list`
- The `for` loop below just won't do anything if the list has length 0

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+        # Get container registry autologin setting
+        container_registry_autologin = self.config.container_registry_autologin
+        if container_registry_autologin:
+            logger.info(
+                "Attempting to automatically log in to container registry used by stack."
+            )
+
+            # Select stack container registry
+            container_registry = None
+            for component in stack.components.values():
```
schustmi, line 288: I think you can simply do `stack.container_registry` here as a shortcut

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,405 @@
...
+            base64_key_value = self.config.base64_ssh_key.get_secret_value()
+            ssh_key = base64.b64decode(base64_key_value).decode("utf-8")
+            paramiko_key = None
+
+            # Create temporary file and write Docker Compose file to it
+            with tempfile.NamedTemporaryFile(mode="w", delete=True) as f:
+                # Get file path
+                file_path = f.name
+
+                # Write Docker Compose file to temporary file
```
schustmi, line 241:
```suggestion
                # Write SSH key file to temporary file
```

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,405 @@
...
+                with f.file as f_:
+                    f_.write(ssh_key)
+
+                paramiko_key = self._paramiko_key_type_given_auth_method().from_private_key_file(
+                    file_path, password=ssh_passphrase
+                )
+
+            # Trim whitespace from the IP address
+            ip_address = str(self.config.ip_address).strip()
```
schustmi, line 250: This is already a string so I think we don't need the conversion?

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,405 @@
...
+                "SSH error: %s. A common cause for this error is selection of the wrong key type in your service connector.",
+                e,
+            )
+        except Exception as e:
+            logger.error(
+                "Unknown error while connecting to HyperAI instance: %s. Please check your network connection, IP address, and authentication details.",
+                e,
+            )
+
+    def _connect_to_resource(
```
schustmi, line 279: I think it would be nice if we can extract the shared code of this method and `_authorize_client(...)` above into a helper method to avoid duplicate code.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+
+        # Get pipeline run id and create directory for it
+        orchestrator_run_id = self.get_orchestrator_run_id()
+        directory_name = f"{directory_name}/{orchestrator_run_id}"
+        stdin, stdout, stderr = paramiko_client.exec_command(
+            f"mkdir -p {directory_name}"
+        )
+
+        # Remove all folders from nonscheduled pipelines if they are 7 days old or older
+        stdin, stdout, stderr = paramiko_client.exec_command(
```
schustmi, line 334: Is this maybe something that we should make configurable?

stefannica:
@christianversloot amazing contribution, I can't wait to review / try this. For now, I just wanted to mention that the multi-line limitation can eventually be worked around by using the `@file-name` CLI syntax to read attributes from a file rather than passing the value inline, e.g.:

```
zenml service-connector register my-hyperai-sc --type hyperai --ssh_key=@ssh-key-file.pub --ip_address=1.2.3.4 --username=admin
```

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+            RuntimeError: If a step fails.
+        """
+        # Basic Docker Compose definition
+        compose_definition = {"version": "3", "services": {}}
+
+        # Get deployment id
+        deployment_id = deployment.id
+
+        # Set environment
+        os.environ[ENV_ZENML_HYPERAI_RUN_ID] = str(deployment_id)
```
schustmi, line 198: I think using the deployment ID here will lead to issues with scheduled pipelines, as there will be multiple runs for the same deployment. (This ID needs to be something that is unique for each run). Maybe docker compose has something like an internal run ID that they use, which we can expose in the environment with some placeholder? 🤔

christianversloot:
> @christianversloot amazing contribution, I can't wait to review / try this. For now, I just wanted to mention that the multi-line limitation can eventually be worked around by using the `@file-name` CLI syntax to read attributes from a file rather than passing the value inline, e.g.:
> 
> ```
> zenml service-connector register my-hyperai-sc --type hyperai --ssh_key=@ssh-key-file.pub --ip_address=1.2.3.4 --username=admin
> ```

Thanks! That is really interesting!

One question before I maybe switch to that approach: is it UI compatible? In other words, does a "File upload" form appear there where the SSH key can be uploaded? If so, how is it stored and processed?

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
schustmi: Also, I think you don't actually need to set this in the local environment, only in the containers where the steps get executed.

schustmi (changes requested):
This is awesome, great job @christianversloot!

I especially liked the docker compose file generation, I think we can even extract this into a separate class at some point in the future to reuse it across other integrations or make it available locally as well.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: I will test again to make sure I don't, but I added it there because I _did_ get an error related to setting this run ID. If I don't anymore, I will remove it. Thanks!

I will read into possible internal run IDs.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: We can, yes. But what do you think about setting it to `True` by default? It is all too common to forget cleanup, and although these are small files, they can pile up quickly.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+        """
+        try:
+            return os.environ[ENV_ZENML_HYPERAI_RUN_ID]
+        except KeyError:
+            raise RuntimeError(
+                "Unable to read run id from environment variable "
+                f"{ENV_ZENML_HYPERAI_RUN_ID}."
+            )
+
+    def _generate_valid_path_format(self, path) -> str:
```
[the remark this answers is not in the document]
christianversloot: Thanks, I changed this in https://github.com/zenml-io/zenml/pull/2372/commits/2c19d2c530e206745b47f8497cabc419b57589c1. Let me know if you have a different suggestion.

stefannica:
> > @christianversloot amazing contribution, I can't wait to review / try this. For now, I just wanted to mention that the multi-line limitation can eventually be worked around by using the `@file-name` CLI syntax to read attributes from a file rather than passing the value inline, e.g.:
> > ```
> > zenml service-connector register my-hyperai-sc --type hyperai --ssh_key=@ssh-key-file.pub --ip_address=1.2.3.4 --username=admin
> > ```
> 
> Thanks! That is really interesting!
> 
> One question before I maybe switch to that approach: is it UI compatible? In other words, does a "File upload" form appear there where the SSH key can be uploaded? If so, how is it stored and processed?

No, unfortunately the UI expects an inline string, although multi-line strings are supported in the UI. It's up to you if you want to switch to that approach, both make sense in my opinion.

christianversloot:
> > > @christianversloot amazing contribution, I can't wait to review / try this. For now, I just wanted to mention that the multi-line limitation can eventually be worked around by using the `@file-name` CLI syntax to read attributes from a file rather than passing the value inline, e.g.:
> > > ```
> > > zenml service-connector register my-hyperai-sc --type hyperai --ssh_key=@ssh-key-file.pub --ip_address=1.2.3.4 --username=admin
> > > ```
> > 
> > 
> > Thanks! That is really interesting!
> > One question before I maybe switch to that approach: is it UI compatible? In other words, does a "File upload" form appear there where the SSH key can be uploaded? If so, how is it stored and processed?
> 
> No, unfortunately the UI expects an inline string, although multi-line strings are supported in the UI. It's up to you if you want to switch to that approach, both make sense in my opinion.

In that case I would like to suggest to keep the current approach because it both supports UI and CLI; then switch to file-based configuration once the UI supports it, too.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+
+if TYPE_CHECKING:
+    from zenml.models import PipelineDeploymentResponse
+
+logger = get_logger(__name__)
+
+ENV_ZENML_HYPERAI_RUN_ID = "ZENML_HYPERAI_ORCHESTRATOR_RUN_ID"
+
+
+class HyperAIOrchestratorSettings(BaseSettings):
```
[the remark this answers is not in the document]
christianversloot: Done

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+    BaseOrchestratorConfig, HyperAIOrchestratorSettings
+):
+    """Configuration for the HyperAI orchestrator.
+
+    Attributes:
+        container_registry_autologin: If True, the orchestrator will attempt to
+            automatically log in to the container registry specified in the stack
+            configuration on the HyperAI instance. This is useful if the container
+            registry requires authentication and the HyperAI instance has not been
+            manually logged in to the container registry. Defaults to `True`.
```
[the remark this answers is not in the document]
christianversloot: Done

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
christianversloot: Done

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,445 @@
...
+            )
+
+    def _generate_valid_path_format(self, path) -> str:
+        """Validates if a given string is in a valid path format.
+
+        Args:
+            path (str): The path to be validated.
+
+        Returns:
+            bool: True if the path has a valid format, False otherwise.
```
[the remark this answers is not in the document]
christianversloot: Done

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: Done

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: Done

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: Upon removal of that line I'm getting this error:

```
│ /mnt/c/users/cvers/repos/zenml/src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator │
│ .py:93 in get_orchestrator_run_id                                                                │
│    90 │   │   try:                                                                               │
│    91 │   │   │   return os.environ[ENV_ZENML_HYPERAI_RUN_ID]                                    │
│    92 │   │   except KeyError:                                                                   │
│ ❱  93 │   │   │   raise RuntimeError(                                                            │
│    94 │   │   │   │   "Unable to read run id from environment variable "                         │
│    95 │   │   │   │   f"{ENV_ZENML_HYPERAI_RUN_ID}."                                             │
│    96 │   │   │   )                                                                              │
RuntimeError: Unable to read run id from environment variable ZENML_HYPERAI_ORCHESTRATOR_RUN_ID.
```

I'm using it there because both the Airflow and Local Docker orchestrators have similar `get_orchestrator_run_id` defs.

Any suggestions how to proceed?

christianversloot: Also, is `get_orchestrator_run_id` called within the container during a pipeline run? If so, we could perhaps set a unique run ID there via e.g. a timestamp.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+        except paramiko.ssh_exception.SSHException as e:
+            logger.error(
+                "SSH error: %s. A common cause for this error is selection of the wrong key type in your service connector.",
+                e,
+            )
+        except Exception as e:
+            logger.error(
+                "Unknown error while connecting to HyperAI instance: %s. Please check your network connection, IP address, and authentication details.",
+                e,
+            )
```
stefannica, line 273: You should also raise all these error messages as `AuthorizationException` exceptions. The base service connector class (and in extension all its consumers) expect methods like `_connect_to_resource` and `_verify` to raise `AuthorizationException` when the credentials are incorrectly configured or invalid.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+            AuthenticationException: If the authentication credentials are
+                invalid.
+            SSHException: If there is an SSH related error.
+        """
+        logger.info("Verifying connection to HyperAI instance...")
+
+        try:
+            paramiko_client = self._create_paramiko_client()
+            if paramiko_client is None:
+                raise RuntimeError(
```
stefannica, line 291: Raising `RuntimeError` in the context of the ZenML server results in an internal server error, whereas this is better reported to the UI or the CLI calling this function as an `AuthorizationException`, which is more specific and even expected in this situation.

But this code will probably become obsolete if you raise `AuthorizationException` in `_create_paramiko_client`.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+        Returns:
+            An authenticated Paramiko SSH client.
+        """
+        logger.info("Connecting to HyperAI instance...")
+        assert self.resource_id is not None
+
+        try:
+            paramiko_client = self._create_paramiko_client()
+            if paramiko_client is None:
+                raise RuntimeError(
```
stefannica, line 321: same thing here:  `AuthorizationException` is more nuanced.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+                    "Could not connect to instance due to missing paramiko client."
+                )
+            else:
+                return paramiko_client
+        except Exception as e:
+            logger.error(
```
stefannica, line 327: same here: this should be raised as an `AuthorizationException`.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+                "Unknown error while connecting to HyperAI instance: %s. Please check your network connection, IP address, and authentication details.",
+                e,
+            )
+
+    def _configure_local_client(
+        self,
+        **kwargs: Any,
+    ) -> None:
+        """There is no local client for the HyperAI connector, so it does nothing."""
+        pass
```
stefannica, line 337: This sort of thing might be useful when someone needs to access the HyperAI instance to debug something. You could use the stored SSH key / passphrase to configure the local SSH client.

If you choose not to implement this method though, could you please raise `NotImplementedError` instead ? This is better handled in the CLI.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+
+        Args:
+            resource_type: The type of resource to verify. Must be set to the
+                Docker resource type.
+            resource_id: The HyperAI instance to verify.
+
+        Returns:
+            The resource ID if the connection can be established.
+        """
+        self._authorize_client()
```
stefannica, line 385: same here: please raise `AuthorizationException` if any errors occur related to misconfigured or invalid credentials. Or just let `AuthorizationException` fall through from `self._authorize_client()`

docs/book/stacks-and-components/auth-management/hyperai-service-connector.md
```diff
@@ -0,0 +1,50 @@
...
+The HyperAI Service Connector supports HyperAI instances.
+
+## Authentication Methods
+ZenML creates an SSH connection to the HyperAI instance in the background when using this Service Connector. It then provides these connections to stack components requiring them, such as the HyperAI Orchestrator. Multiple authentication methods are supported:
+
+1. RSA key based authentication.
+2. DSA (DSS) key based authentication.
+3. ECDSA key based authentication.
+4. ED25519 key based authentication.
+
```
stefannica, line 36: It might be worth adding a "warning" hint here making users aware that the SSH private keys configured in the connector will be distributed to all clients that use them to run pipelines with the HyperAI orchestrator.

This is an obvious security weakness, as the SSH keys are long-lived credentials that give unrestricted access to the HyperAI instances.

stefannica (changes requested):
I only looked over the service connector implementation and left a few suggestions for improvement. It goes without saying that you have my thanks and appreciation for tackling such a complex subject. It's not every day that someone implements a brand new service connector.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+    ssh_passphrase: Optional[SecretStr] = Field(
+        default=None,
+        title="SSH key passphrase",
+    )
+
+
+class HyperAIConfiguration(HyperAICredentials):
+    """HyperAI client configuration."""
+
+    ip_address: str = Field(
```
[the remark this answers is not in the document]
christianversloot: I could change it into `hostname`. That way, the field is equal to how `paramiko` calls it, and it would be agnostic to DNS entries or IP addresses. WDYT?

src/zenml/integrations/hyperai/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+"""Initialization of the HyperAI integration."""
+from typing import List, Type
+
+from zenml.integrations.constants import HYPERAI
+from zenml.integrations.integration import Integration
+from zenml.stack import Flavor
+
+# Service connector constants
+HYPERAI_CONNECTOR_TYPE = "hyperai"
+HYPERAI_RESOURCE_TYPE = "hyperai-generic"
```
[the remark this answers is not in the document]
christianversloot: Will be changed in future commit 2d36b6e12.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
christianversloot: Changed in 57f2c9219.

I am raising the exception generically (i.e. not one exception per `except` case, just one at the end, as everything would have led there anyway). If client creation succeeds the `paramiko_client` will be returned, so it will either be:

* A connected client
* An `AuthorizationException`

```
            return paramiko_client

        except paramiko.ssh_exception.BadHostKeyException as e:
            logger.error("Bad host key: %s", e)
        except paramiko.ssh_exception.AuthenticationException as e:
            logger.error("Authentication failed: %s", e)
        except paramiko.ssh_exception.SSHException as e:
            logger.error(
                "SSH error: %s. A common cause for this error is selection of the wrong key type in your service connector.",
                e,
            )
        except Exception as e:
            logger.error(
                "Unknown error while connecting to HyperAI instance: %s. Please check your network connection, IP address, and authentication details.",
                e,
            )

        raise AuthorizationException("Could not create SSH client for HyperAI instance.")
```

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
christianversloot: I have in fact removed this `if` statement in ce40f594a because the changes applied in 57f2c9219 (i.e. raise `AuthorizationException` if no client can be created) disallow ever having `paramiko_client == None` here (it is simply not returned anymore). This removes the need for the if statement.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
christianversloot: I have in fact removed this `if` statement in ce40f594a because the changes applied in 57f2c9219 (i.e. raise `AuthorizationException` if no client can be created) disallow ever having `paramiko_client == None` here (it is simply not returned anymore). This removes the need for the if statement.

PS: ah I see that you had already written that in your last phrase :)

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
christianversloot: I have removed the whole `try/except` block in ae7b069d4 because it is no longer necessary: `_create_paramiko_client` will raise the `AuthorizationException` if no client can be created.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
christianversloot: If I would want to implement it, I could imagine that I would return the `paramiko.client.SSHClient` by calling `self._create_paramiko_client()` - or does this def have a different purpose?

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: image

Added and works!

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: I think we're close then, because we will then probably run into the issue that two scheduled pipelines are not supported (as they share the same environment variable).

How about changing

```
f"(crontab -l ; echo '{cron_expression} cd {directory_name} && docker compose up -d') | crontab -"
```

into

```
deployment_id = deployment.id
deployment_id_uppercase = deployment_id.upper()
f"(crontab -l ; echo '{cron_expression} cd {directory_name} && export ZENML_HYPERAI_RUN_ID_{deployment_id_uppercase}=\"{deployment_id}_$(date +%s)\" && docker compose up -d') | crontab -"

environment_run_id = f"{ENV_ZENML_HYPERAI_RUN_ID_{deployment_id_uppercase}"
environment[environment_run_id] = str(deployment_id)
```

I believe this means that:

* Each run will have its own unique run ID.
* Multiple scheduled pipelines will never have conflicting runs environment variable wise.

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
```diff
@@ -0,0 +1,386 @@
...
+        ResourceTypeModel(
+            name="HyperAI instance",
+            resource_type=HYPERAI_RESOURCE_TYPE,
+            description="""
+Allows users to access a HyperAI instance as a resource. When used by
+connector consumers, they are provided a pre-authenticated SSH client
+instance.
+""",
+            auth_methods=HyperAIAuthenticationMethods.values(),
+            supports_instances=False,
```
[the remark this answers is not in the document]
stefannica: I think you covered most of it:
* you have to mark your resource type as `supports_instances=True`. This opens up different code paths in the base connector implementation.
* you also have to update `_verify` to return all the configured hostnames in that list
* `_get_default_resource_id` you can throw away, it's only called for single-instance connectors

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
schustmi: That would not work with the implementation of `get_orchestrator_run_id(...)` which checks for a fixed env variable, right?

schustmi: But I don't think we need the `_{deployment_id_uppercase}` suffix in the environment variable, we can just keep that as `ZENML_HYPERAI_RUN_ID` and set it to different values right?

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
stefannica: This method serves a different purpose. It's basically meant to be the exact opposite of discovering and lifting up credentials from your local machine. In this case, it would most likely create a file in the `~/.ssh/` folder containing the  key stored in the connector configuration and instruct the user on how to use e.g. by showing some example `ssh -i <private-key-file> <hostname>` commands.

christianversloot: Hm, I'd say we leave that out for now. I'll make sure to raise `NotImplementedError`.

christianversloot: https://github.com/zenml-io/zenml/pull/2372/commits/6ca8b97a0e257e87aa427dd717ffd2419870d4dd

src/zenml/integrations/hyperai/service_connectors/hyperai_service_connector.py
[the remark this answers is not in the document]
christianversloot: https://github.com/zenml-io/zenml/pull/2372/commits/4ad6d79d4fba7355878ab5ff1e858805f5202a09

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
schustmi: I guess if you ssh into the instance, you get a separate shell in which you set the environment variable, which won't interfere with other sessions? At least that was my assumption, not sure it holds true though. As for long-running pipelines: I'm not sure if the environment variables get resolved by `docker-compose up` or once the container starts running (the latter could potentially be an issue). Is there maybe a way to run `docker-compose up -e ZENML_HYPERAI_RUN_ID=...` or something comparable 🤔)

christianversloot: You're right! I tested as follows:

```
# m h  dom mon dow   command
* * * * * export VARIABLE_CONTENT="first_job" && echo $VARIABLE_CONTENT > /home/chris/first_job.txt
* * * * * export VARIABLE_CONTENT="second_job" && echo $VARIABLE_CONTENT > /home/chris/second_job.txt
```

The result:

```
(zenmldevelop) chris@MSI:~$ cat first_job.txt
first_job
(zenmldevelop) chris@MSI:~$ cat second_job.txt
second_job
```

So unless you have a different approach that works better, I'll make sure that the crontab entry is created as follows tomorrow - and then test it 😄 

```
f"(crontab -l ; echo '{cron_expression} cd {directory_name} && export ZENML_HYPERAI_RUN_ID=\"{deployment_id}_$(date +%s)\" && docker compose up -d') | crontab -"
```

christianversloot: (As for longer-running pipelines: good question... looks like something is available. I'll look into it tomorrow: https://docs.docker.com/compose/environment-variables/set-environment-variables/#set-environment-variables-with-docker-compose-run---env).

stefannica (approved):
Service connector implementation looks good and works like a charm.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,324 @@
...
+            # Get settings
+            step_settings = cast(
+                HyperAIOrchestratorSettings, self.get_settings(step)
+            )
+
+            # Define container name as combination between deployment id and step name
+            container_name = f"{deployment_id}-{step_name}"
+
+            # Make Compose service definition for step
+            compose_definition["services"][container_name] = {
```
stefannica, line 172: I personally haven't ran Docker containers with GPU support, but don't we need to set some special options to allow the containers to access the GPUs ? Something like this maybe: https://stackoverflow.com/questions/25185405/using-gpu-from-a-docker-container ?

stefannica: @michael-zenml or is this something that ZenML's docker settings can be used to configure on a per-container level ?

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: Current idea:

* Write env var to a pipeline-specific env file in the pipeline folder created by the orchestrator (deterministic .env file location)
* Include it in the Docker Compose file for the services.

That way, each time it's executed, the new env var is set and then passed to the Docker daemon. We can then also test what happens with longer-running pipelines.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
htahir1: @schustmi and me discussed this and we definitely need cuda installed but not sure we need anything else

christianversloot: As far as I know:

* The NVIDIA driver for Linux: https://www.nvidia.com/en-us/drivers/unix/
* The NVIDIA Container Toolkit (https://github.com/NVIDIA/nvidia-container-toolkit) must be installed and configured (https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html)

That's it! No CUDA toolkit needs to be installed on the host, it only needs to be available within the container for which many packages are available:

image

Perhaps we can add a section about running GPU containers and these necessities into the orchestrator docs? I think it's difficult to install these from within the orchestrator or while installing the integration.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: ENV INTERACTION FOR DIFFERENT PIPELINE RUNS:
To see if two different pipelines can interfere each other when scheduled, I've tested with the following crontab:

12 * * * * cd /mnt/c/users/cvers/repos/composetest/serv1 && echo "SOME_ENV_VAR=1" > .env1 && docker compose up -d
11 * * * * cd /mnt/c/users/cvers/repos/composetest/serv2 && echo "SOME_ENV_VAR=2" > .env2 && docker compose up -d

Both entries start up a sequential series of 3 containers, each of which echos $SOME_ENV_VAR and then sleeps for 120 seconds. Because one `compose up` starts one minute after the other, there are 3 windows of overlap where envvar interaction might take place. If it doesn't, we'd expect 3x1 and 3x2 when inspecting the logs after the whole sequence finishes.

Results align with expectations - when logging the container logs I see 3x1 and 3x2. We can thus be safe to assume that 

root@MSI:/mnt/c/users/cvers/repos/composetest/serv2# docker logs 3c74
Contents: 2

ENV INTERACTION FOR SAME PIPELINE RUN:
This also means that if a pipeline is scheduled twice to run at different intervals, it will work nicely. That is because ZenML will place the pipeline twice in a different folder based on the following run ID, which I just tested is different every single time you run a pipeline (for scheduling):

```
orchestrator_run_id = self.get_orchestrator_run_id()
```

...so there will be no interference when scheduling the same pipeline twice...

```
zenml@ubuntu-4gb-fsn1-2:~/scheduled-pipeline-runs$ ls
1a5501d9-af43-4f98-b0f8-2c3336f689df  4d2514c6-1a4e-4359-9ac2-7907b54b5f88
```

```
zenml@ubuntu-4gb-fsn1-2:~/scheduled-pipeline-runs$ crontab -l
*/5 * * * * cd /home/zenml/scheduled-pipeline-runs/1a5501d9-af43-4f98-b0f8-2c3336f689df && docker compose up -d
*/5 * * * * cd /home/zenml/scheduled-pipeline-runs/4d2514c6-1a4e-4359-9ac2-7907b54b5f88 && docker compose up -d
```

Hence, I'll work on the proposed implementation now :)

christianversloot: Implemented in https://github.com/zenml-io/zenml/pull/2372/commits/7bb3405f4b831b591b7c39ecbb984079e0cbc1d6 and tested with both scheduled and nonscheduled pipeline runs. a92dd2195 adds more information to the docs.

With schedule (`schedule = Schedule(cron_expression="*/1 * * * *")`) the pipeline run nicely executes every minute.

Without the schedule the pipeline starts up immediately and finishes nicely.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: @schustmi @htahir1 WDYT?

schustmi: I think our `DockerSettings` can be used to configure an image that is GPU enabled, I'm not 100% sure if there's anything else needed in the docker-compose file to make it work though (something like the `--gpus=all` flag in @stefannica link above). If not, I don't think there is any need to mention this in the orchestrator docs as it's a general solution that applies to all orchestrators.

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
schustmi: Nice!

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: Ah, yes. We will probably need to add this to the Docker Compose file generation:

```
    resources:
        reservations:
          devices:
            - driver: nvidia
              capabilities: [gpu]
```

That way, all GPUs should become accessible to the container (because we omit setting `count` and `device_ids`, see link below).

Perhaps, we can make this configurable (i.e. allow orchestrators to use parts of the GPU) later.

https://docs.docker.com/compose/gpu-support/

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -0,0 +1,336 @@
...
+            stdin, stdout, stderr = paramiko_client.exec_command(
+                f"docker login -u {container_registry_username} --password-stdin {container_registry_url} <<< {container_registry_password}"
+            )
+
+            # Log stdout
+            for line in stdout.readlines():
+                logger.info(line)
+
+        # Set up pipeline-runs directory if it doesn't exist
+        nonscheduled_directory_name = "/home/zenml/pipeline-runs"
```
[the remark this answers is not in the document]
christianversloot: Very fair question. I'd like to keep things a bit standardized here especially because the cleanup mechanism has assumptions about where things are stored (i.e. if you configure the scheduled and nonscheduled directories in the same way, you risk losing pipeline files that are necessary per the schedule). Still, the dir may not be writable. However, we know the `username` with which the service connector logs in. I think it is rare that a user cannot write to their own home directory. Hence, I would like to propose to use `f"/home/{username}/pipeline-runs"` (or its scheduled equivalent) as it is likely writable all the time and then in line with service connector configuration.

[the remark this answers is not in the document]
schustmi: Sounds good to me!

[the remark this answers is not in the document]
christianversloot: 31d0a2849

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
christianversloot: I've just tested on a GPU equipped instance with a local Docker image (`tensorflow:latest-gpu` with ZenML installed locally). It then does run (and not only create) the machine, compared to a CPU-only instance (so the configurable disabling of GPU reservation would be a solution). Within the pipeline, I ran

```
    import tensorflow as tf

    # check physical devices
    logger.info('Physical devices:')
    logger.info(tf.config.list_physical_devices())
```

and got

```
Physical devices:
[PhysicalDevice(name='/physical_device:CPU:0', device_type='CPU'), PhysicalDevice(name='/physical_device:GPU:0', device_type='GPU')]
```

so the way GPU configuration is built into the orchestrator is working, under the condition that the NVIDIA Driver and NVIDIA Container Toolkit are installed on the instance!

schustmi (approved):
LGTM 🦭

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -271,22 +275,27 @@ def prepare_or_run_pipeline(
...
-                container_registry_username,
-                container_registry_password,
-            ) = container_registry.credentials
-
-            # Log in to container registry using --password-stdin
-            stdin, stdout, stderr = paramiko_client.exec_command(
-                f"docker login -u {container_registry_username} --password-stdin {container_registry_url} <<< {container_registry_password}"
-            )
+            credentials = container_registry.credentials
+            if credentials is not None:
```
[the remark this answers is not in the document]
stefannica: done

src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
```diff
@@ -226,7 +230,7 @@ def prepare_or_run_pipeline(
 
         # Convert into yaml
         logger.info("Finalizing Docker Compose definition.")
-        compose_definition = yaml.dump(compose_definition)
+        compose_definition_yaml = yaml.dump(compose_definition)
```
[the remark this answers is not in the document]
stefannica: done

PR #2398, opened 2024-02-06
Add Vertex Step Operator network parameter

francoisserra:
## Describe changes
I implemented extra config parameters of the Vertex Step Operator:

- Support of the network and reserved_ip_ranges parameters (allow the CustomJob to run in private/shared VPC)
- Support specifying the GCP service account for the CustomJob execution

## Pre-requisites
- [?] If my changes require changes to the dashboard, these changes are communicated/requested.

## Types of changes


docs/book/stacks-and-components/component-guide/step-operators/vertex.md
```diff
@@ -122,9 +122,38 @@ more about how ZenML builds these images and how you can customize them.
 
 #### Additional configuration
 
+You can specify the service account, network and reserved IP ranges to use for the VertexAI CustomJob by passing the `service_account`, `network` and `reserved_ip_ranges` parameters to the `step-operator register` command:
```
strickvl, line 125:
```suggestion
You can specify the service account, network and reserved IP ranges to use for the VertexAI `CustomJob` by passing the `service_account`, `network` and `reserved_ip_ranges` parameters to the `step-operator register` command:
```

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
```diff
@@ -250,7 +250,14 @@ def launch(
...
                         },
                     }
-                ]
+                ],
+                "service_account": self.config.service_account,
+                "network": self.config.network,
+                "reserved_ip_ranges": (
+                    self.config.reserved_ip_ranges
+                    if self.config.reserved_ip_ranges
+                    else []
```
strickvl, line 259: for my understanding, why do we take in a string for this in the flavor, but it's a list here when defining?

strickvl (commented):
Just one small question, otherwise looks good to me.

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
francoisserra: hmmm yep I agree there's a smelly thing overthere..  Let me check (need to test deeper)

francoisserra: (I think GH Copilot misled me...)

francoisserra: @strickvl  am I right that the [parse_name_and_extra_arguments](https://github.com/zenml-io/zenml/blob/cdd14528d674e0cafe00173755e556ea00c6343c/src/zenml/cli/utils.py#L787) utility function that handles the specific args of a (step operator) flavor doesn't handle list type parameters (i.e. same behavior as the `multiple=True` option in `@click.option()` decorator)?

What is the standard way of handling such _list parameters_ in ZenML CLI ?

strickvl: I think till now we've managed to avoid this situation :) AFAICS, `parse_name_and_extra_arguments` would need to be updated to handle the case of a list of inputs, but let me check with a colleague and get back to you.

strickvl: So as far as I understand it, I think the `parse_name_and_extra_arguments` currently doesn't handle that so would need to be updated to take care of a string that represented a list of arguments.

src/zenml/integrations/gcp/flavors/vertex_step_operator_flavor.py
```diff
@@ -79,6 +86,12 @@ class VertexStepOperatorConfig(  # type: ignore[misc] # https://github.com/pydan
     # will be applied to all Vertex AI resources if set
     encryption_spec_key_name: Optional[str] = None
 
+    network: Optional[str] = None
+
+    reserved_ip_ranges: Optional[str] = None
```
safoinme, line 91: should this be a list of str?

safoinme (changes requested):

strickvl (approved):
LGTM now. Thanks for the update!

safoinme (approved):

PR #2405, opened 2024-02-06
Add new property for filtering service account events

strickvl:
Added a way to filter for events made by real users vs those made via service accounts.

Tested to ensure that the new property shows up in mixpanel.


bcdurak (approved):

stefannica (approved):
Looks fine to me. Just a quick question for @bcdurak maybe: does this also need to go in the identify request ? I don't think this is covered here.

bcdurak (approved):

bcdurak:
@stefannica You are right. As we discussed, it would be great if we could add this as a trait for our identify requests. We will merge this in for now and handle it in a separate PR.

PR #2435, opened 2024-02-09
Correctly expose the flavors. List endpoint with familiar Pagination …

AlexejPenner:
## Describe changes
This PR contains quite a few changes to the Response Models.

1) BaseResponse no longer assumes that the Entity is stored in the DB with a id, created and updated fields
2)  All Models that formerly inherited from the BaseModel now inherit from the BaseIdentifiedModel which assumes the entity has an `id`, `created` and `updated` field.
3) BaseSchemas  now implement the to_model() method with metadata and resources 
4) All flavor plugins can now be generically fetched at the `/plugin-flavors` route
5) This PR also enables pagination of in-memory lists
6) ResponseBody.resources now returns related resources of the main response entity in accordance with the new API guidelines

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **Refactor**
	- Updated internal response model references from `BaseResponse` to `BaseDBResponse` across various files for consistency and clarity.
- **New Features**
	- Introduced a new error class `EventSourceExistsError` for handling duplicate event source registrations.
	- Renamed `PLUGINS` constant to `PLUGIN_FLAVORS` to better reflect its purpose.
	- Enhanced event source models with additional metadata for source and filter configurations.
	- Improved plugin flavor response models with new classes for handling response body and metadata.
- **Chores**
	- Refactored `PluginFlavorRegistry` class for better clarity and consistency in method naming and data structure.
	- Updated pagination logic and endpoint refactoring for plugin flavors listing.
- **Bug Fixes**
	- Adjusted error messaging for event source registration conflicts.


strickvl:
@coderabbitai review

src/zenml/models/v2/plugin/event_flavor.py
```diff
@@ -14,11 +14,27 @@
...
+
+
+class EventFlavorResponseMetadata(BasePluginResponseMetadata):
+    """Response metadata for event flavors."""
 
     source_config_schema: Dict[str, Any]
     filter_config_schema: Dict[str, Any]
+
+
+class EventFlavorResponse(
```
stefannica, line 35: Heh, I failed to mention this until now and now it's spreading: this should be called `EventSourceFlavorResponse`. Same thing applies to the new body/metadata classes. Even the filename should be `event_source_flavor.py`

src/zenml/models/v2/plugin/plugin_flavor.py
```diff
@@ -12,14 +12,43 @@
...
-    plugin_type: PluginType
-    plugin_subtype: PluginSubType
+    flavor_name: str = Field(title="Name of the flavor.")
+    plugin_type: PluginType = Field(title="Type of the flavor.")
+    plugin_subtype: PluginSubType = Field(title="Subtype of the flavor.")
+
+    class Config:
+        """Configuration for base plugin flavor response."""
+
+        extra = Extra.allow
```
stefannica, line 54: "Allow" means that the plugins can add whatever they want to the core models defined by ZenML and they will be serialized and show up in the responses sent through the REST API. Are you sure you don't mean "ignore" here? "Ignore" will also not raise errors if extra attributes are present, but will silently discard them.

src/zenml/models/v2/plugin/event_flavor.py
```diff
@@ -14,11 +14,27 @@
...
+    BasePluginFlavorResponse[
+        EventFlavorResponseBody, EventFlavorResponseMetadata
+    ]
+):
```
stefannica, line 39: Shouldn't you do the same for the `ActionFlavorResponse` ?

src/zenml/zen_server/routers/plugin_endpoints.py
```diff
@@ -37,82 +44,61 @@
...
-    responses={401: error_response, 404: error_response, 422: error_response},
-)
-@handle_exceptions
-def list_plugin_types(
-    flavor_name: str,
+def list_flavors(
+    plugin_type: PluginType,
+    plugin_subtype: PluginSubType,
+    page: int = PAGINATION_STARTING_PAGE,
+    size: int = PAGE_SIZE_DEFAULT,
```
stefannica, line 55: I love what you did here with the pagination and code contraction!

src/zenml/zen_server/routers/plugin_endpoints.py
```diff
@@ -37,82 +44,61 @@
...
+    page_items = [
+        flavor.get_plugin_flavor_response_model() for flavor in flavors
+    ][start:end]
+
+    return_page = Page(
+        index=page,
+        max_size=size,
+        total_pages=total_pages,
+        total=total,
+        items=page_items,
```
stefannica, line 83: With most (maybe all ?) of the other "list" endpoints, we usually just call the sql zen store to handle everything related to filtering and pagination. Do you think we could also to that here and move all this logic into the plugin registry ?

src/zenml/plugins/plugin_flavor_registry.py
```diff
@@ -44,54 +44,89 @@ class PluginFlavorRegistry:
...
-        """Get a list of all subtypes for a specific flavor and type."""
         if (
-            flavor in self.plugin_flavors
-            and _type in self.plugin_flavors[flavor]
+            _type in self.plugin_flavors
+            and sub_type in self.plugin_flavors[_type]
         ):
-            return list(self.plugin_flavors[flavor][_type].keys())
-        return []
+            return self.plugin_flavors[_type][sub_type]
```
stefannica, line 85: else return ?

src/zenml/plugins/plugin_flavor_registry.py
```diff
@@ -44,54 +44,89 @@ class PluginFlavorRegistry:
...
+            page: For optional pagination
+            size: For optional pagination
+        """
+        flavor_names = list(
+            self._flavor_entries(_type=_type, sub_type=sub_type).keys()
+        )
+        if page > 0 and size > 0:
+            start = page * size
+            end = start + size
+            return flavor_names[start:end]
```
stefannica, line 108: else return ?

src/zenml/zen_server/routers/plugin_endpoints.py
```diff
@@ -37,82 +44,61 @@
...
-
-@plugin_router.get(
-    "/{flavor_name}",
-    response_model=List[str],
-    responses={401: error_response, 404: error_response, 422: error_response},
-)
-@handle_exceptions
-def list_plugin_types(
-    flavor_name: str,
+def list_flavors(
```
stefannica, line 51: You're missing the `hydrate` argument.

src/zenml/zen_server/routers/plugin_endpoints.py
```diff
@@ -37,82 +44,61 @@
...
+    "/{flavor_name}",
     responses={401: error_response, 404: error_response, 422: error_response},
 )
 @handle_exceptions
 def get_flavor(
     flavor_name: str,
-    plugin_type: PluginType,
-    plugin_subtype: PluginSubType,
+    plugin_type: PluginType = Query(..., alias="type"),
+    plugin_subtype: PluginSubType = Query(..., alias="subtype"),
```
stefannica, line 99: You're missing a `hydrate` argument here.

src/zenml/event_sources/base_event_source.py
```diff
@@ -557,9 +561,12 @@ def get_event_source_config_schema(cls) -> Dict[str, Any]:
     def get_plugin_flavor_response_model(cls) -> EventFlavorResponse:
```
stefannica, line 561: I think you're missing the `hydrate` flag here here that tells you whether you populate the metadata or not.

src/zenml/event_sources/base_event_source.py
```diff
@@ -557,9 +561,12 @@ def get_event_source_config_schema(cls) -> Dict[str, Any]:
...
         """Convert the Flavor into a Response Model."""
         return EventFlavorResponse(
+            body=EventFlavorResponseBody(),
```
stefannica, line 564: Shouldn't you the same for `BaseActionFlavor` ?

stefannica (changes requested):
I have to say I really like the idea of an "in-memory" base filter. I haven't looked closely at the plugin registry logic, because it's just too difficult to make sense of at the moment. 

You should just go all the way and actually implement proper hydration: add flags everywhere, including in the client, and populate the metadata field or not based on this. You're 90% there already.

src/zenml/models/v2/base/base.py
```diff
@@ -99,17 +99,21 @@ class BaseResponseMetadata(BaseZenModel):
...
+    """Base domain model for all responses."""
 
     # Body and metadata pair
     body: Optional["AnyBody"] = Field(title="The body of the resource.")
     metadata: Optional["AnyMetadata"] = Field(
         title="The metadata related to this resource."
     )
 
+
+class BaseDBResponse(BaseResponse, Generic[AnyBody, AnyMetadata]):
```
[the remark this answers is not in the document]
AlexejPenner: IdentifiedEntityResponse <- this could work, as the thing that this adds is the id field (and the date fields)

src/zenml/models/v2/plugin/plugin_flavor.py
AlexejPenner: problem is the endpoint is decorated with BasePluginFlavorResponse - however the plugin registry returns a girhubwebhookeventsourceflavorresponse, which has extra fields. but i guess this might be soled by moving those into the body as well

src/zenml/zen_server/routers/plugin_endpoints.py
AlexejPenner: Exactly what I wanted to do, was holding myself back though to optimize for speed. Your comment makes me itch to do it anyways now :D

src/zenml/models/v2/base/base.py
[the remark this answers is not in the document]
stefannica: What did you think of my suggestion though ? You should avoid renaming the base class if the base class isn't the thing being changed anyway.

src/zenml/zen_server/routers/plugin_endpoints.py
AlexejPenner: I actually wanted to even have a generic way to paginate in_memory list on the InMemoryFilterModel, but I think thats for the future

src/zenml/models/v2/base/base.py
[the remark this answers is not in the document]
AlexejPenner: I think conceptually the old base class just sin't a base class. Its a specific type of response model with id, created and updated fields. the real base class would be the new one. BaseResponse vs BasicResponse also sounds too confusing i feel. Sure I can still undo this though, but I still had to touch every single of those files for other reasons

src/zenml/zen_server/routers/plugin_endpoints.py
```diff
@@ -37,93 +44,61 @@
...
+        subtype: Subtype of Plugin
 
     Returns:
         The requested flavor response.
     """
-    # TODO: Figure out if typing of the response can be reintroduced
-    return plugin_flavor_registry.get_flavor_class(
-        flavor=flavor_name, _type=plugin_type, subtype=plugin_subtype
-    ).get_plugin_flavor_response_model()
+    plugin_flavor = plugin_flavor_registry.get_flavor_class(
```
schustmi, line 101: These endpoints do not have any RBAC checks at the moment, but I guess there isn't really anything worth hiding in the flavors anyway?

schustmi (changes requested):
Even after reading the new API guidelines, I'm still not really sure what the `resources` part of the new `BaseResponse` refers to. It also seems to me like it is empty for every subclass that you created, so I assume this is a change that we need going forward?

src/zenml/models/v2/base/base_plugin_flavor.py
```diff
@@ -0,0 +1,77 @@
...
+from zenml.enums import PluginSubType, PluginType
+from zenml.models.v2.base.base import (
+    BaseResponse,
+    BaseResponseMetadata,
+    BaseResponseResources,
+    BaseZenModel,
+)
+
+
+class BasePluginResponseBody(BaseZenModel):
```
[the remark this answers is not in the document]
AlexejPenner: correct

src/zenml/zen_server/routers/plugin_endpoints.py
AlexejPenner: Yeah I think the flavors themselves need no RBAC - you're allows to know a plugin exists and what is needed to configure one

src/zenml/models/v2/base/base_plugin_flavor.py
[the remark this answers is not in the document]
AlexejPenner: Well, the BaseResponseBody actually has a creation and update date, but this is a irregularity - I fixed it now

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1560,7 +1561,7 @@ def create_api_key(
             session.add(new_api_key)
             session.commit()
 
-            api_key_model = new_api_key.to_model(hydrate=True)
+            api_key_model = new_api_key.to_model(include_resources=True)
```
stefannica, line 1564:
```suggestion
            api_key_model = new_api_key.to_model(include_metadata=True)
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1588,7 +1589,7 @@ def get_api_key(
                 api_key_name_or_id=api_key_name_or_id,
                 session=session,
             )
-            return api_key.to_model(hydrate=hydrate)
+            return api_key.to_model(include_resources=hydrate)
```
stefannica, line 1592:
```suggestion
            return api_key.to_model(include_metadata=hydrate)
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1702,7 +1703,7 @@ def update_api_key(
 
             # Refresh the Model that was just created
             session.refresh(api_key)
-            return api_key.to_model(hydrate=True)
+            return api_key.to_model(include_resources=True)
```
stefannica, line 1706:
```suggestion
            return api_key.to_model(include_metadata=True)
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1733,7 +1734,7 @@ def update_internal_api_key(
...
```
stefannica, line 1737:
```suggestion
            return api_key.to_model(include_metadata=True)
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1823,7 +1824,7 @@ def create_artifact(self, artifact: ArtifactRequest) -> ArtifactResponse:
             artifact_schema = ArtifactSchema.from_request(artifact)
             session.add(artifact_schema)
             session.commit()
-            return artifact_schema.to_model(hydrate=True)
+            return artifact_schema.to_model(include_resources=True)
```
stefannica, line 1827:
```suggestion
            return artifact_schema.to_model(include_metadata=True)
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1850,7 +1851,7 @@ def get_artifact(
                     f"Unable to get artifact with ID {artifact_id}: No "
                     "artifact with this ID found."
                 )
-            return artifact.to_model(hydrate=hydrate)
+            return artifact.to_model(include_resources=hydrate)
```
stefannica, line 1854:
```suggestion
            return artifact.to_model(include_metadata=hydrate)
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1917,7 +1918,7 @@ def update_artifact(
             session.add(existing_artifact)
             session.commit()
             session.refresh(existing_artifact)
-            return existing_artifact.to_model(hydrate=True)
+            return existing_artifact.to_model(include_resources=True)
```
stefannica, line 1921:
```suggestion
            return existing_artifact.to_model(include_metadata=True)
```

src/zenml/zen_stores/schemas/base_schemas.py
```diff
@@ -26,6 +29,20 @@ class BaseSchema(SQLModel):
     created: datetime = Field(default_factory=datetime.utcnow)
     updated: datetime = Field(default_factory=datetime.utcnow)
 
+    def to_model(
```
stefannica, line 32: I went through the SQL zen store code as well as a few schemas and it looks like 1. the `include_resources` arg is never set to True in `to_model` calls and 2. the `include_resources` arg is never actually used in schema implementations. What did you have in mind as a follow-up step for this flag ?

src/zenml/zen_stores/schemas/api_key_schemas.py
```diff
@@ -113,18 +113,20 @@ def from_request(
...
+    def to_model(
+        self, include_metadata: bool = False, include_resources: bool = False
+    ) -> APIKeyResponse:
         """Convert a `APIKeySchema` to an `APIKeyResponse`.
 
         Args:
-            hydrate: bool to decide whether to return a hydrated version of the
-                model.
+            include_metadata: Whether the metadata will be filled.
+            include_resources: Whether the metadata will be filled.
```
stefannica, line 123:
```suggestion
            include_resources: Whether the resources will be filled.
```

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -99,12 +99,14 @@ def from_request(
...
+    def to_model(
+        self, include_metadata: bool = False, include_resources: bool = False
+    ) -> ArtifactResponse:
         """Convert an `ArtifactSchema` to an `ArtifactResponse`.
 
         Args:
-            hydrate: bool to decide whether to return a hydrated version of the
-                model.
+            include_metadata: Whether the metadata will be filled.
+            include_resources: Whether the metadata will be filled.
```
stefannica, line 109:
```suggestion
            include_resources: Whether the resources will be filled.
```

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -266,12 +268,14 @@ def from_request(
...
+    def to_model(
+        self, include_metadata: bool = False, include_resources: bool = False
+    ) -> ArtifactVersionResponse:
         """Convert an `ArtifactVersionSchema` to an `ArtifactVersionResponse`.
 
         Args:
-            hydrate: bool to decide whether to return a hydrated version of the
-                model.
+            include_metadata: Whether the metadata will be filled.
+            include_resources: Whether the metadata will be filled.
```
stefannica, line 278:
```suggestion
            include_resources: Whether the resources will be filled.
```

I won't suggest this for other schemas, but they all have the same issue

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -32,45 +32,13 @@
     TriggerResponseMetadata,
     TriggerUpdate,
 )
+from zenml.models.v2.core.trigger import TriggerResponseResources
```
stefannica, line 35:
```suggestion
from zenml.models import TriggerResponseResources
```

src/zenml/plugins/base_plugin_flavor.py
```diff
@@ -19,7 +19,7 @@
 
 from zenml.config.global_config import GlobalConfiguration
 from zenml.enums import PluginSubType, PluginType
-from zenml.models import BasePluginFlavorResponse
+from zenml.models.v2.base.base_plugin_flavor import BasePluginFlavorResponse
```
stefannica, line 22:
```suggestion
from zenml.models import BasePluginFlavorResponse
```

src/zenml/models/v2/base/base.py
```diff
@@ -126,6 +124,102 @@ def get_hydrated_version(self) -> "BaseResponse[AnyBody, AnyMetadata]":
...
+            RuntimeError: If the body was not included in the response.
+        """
+        if not self.body:
+            raise RuntimeError(
+                f"Missing response body for {type(self).__name__}."
+            )
+
+        return self.body
+
+    def get_metadata(self) -> "AnyMetadata":
```
stefannica, line 143: The `get_body` and `get_metadata` method in the BaseIdentifiedResponse class are mostly duplicates of those in the base class. Could you please just call these method from those methods instead ?

Second idea: what about keeping `permission_denied` in the base class, in which case you won't need to override these methods at all in the `BaseIdentifiedResponse` class ? Or don't you think we'll need permissions for in-memory models too ?

src/zenml/models/v2/base/base.py
```diff
@@ -94,28 +80,40 @@ class BaseResponseMetadata(BaseZenModel):
...
 AnyBody = TypeVar("AnyBody", bound=BaseResponseBody)
 AnyMetadata = TypeVar("AnyMetadata", bound=BaseResponseMetadata)
+AnyResources = TypeVar("AnyResources", bound=BaseResponseResources)
 
 
-class BaseResponse(GenericModel, Generic[AnyBody, AnyMetadata], BaseZenModel):
-    """Base domain model."""
-
-    id: UUID = Field(title="The unique resource id.")
-    permission_denied: bool = False
```
stefannica, line 105: why not keep `permission_denied` in the base class ? we'll probably need to run permission checks for in-memory models too.

stefannica (changes requested):
Only half way through...

src/zenml/models/v2/base/base.py
AlexejPenner: Had a discussion about this with Michael - for the time being no, no permission checks there. The rbac implementation also runs with the assumption that resources have ids. If we want to enable permission checks for in-memory items in the future, we can move it then

src/zenml/models/v2/base/base.py
AlexejPenner: sure, only differences are the error messages.

src/zenml/zen_stores/schemas/artifact_schemas.py
AlexejPenner: will fix everywhere

src/zenml/zen_stores/schemas/base_schemas.py
AlexejPenner: as we slowly migrate to the new paradigm, the include resources flag will be used in the places where they should be included. 
I fixed all the places where include_resources was falsely used

stefannica (changes requested):

src/zenml/models/v2/base/base.py
```diff
@@ -126,6 +124,100 @@ def get_hydrated_version(self) -> "BaseResponse[AnyBody, AnyMetadata]":
...
+    Used as a base class for all body models associated with responses.
+    Features a creation and update timestamp.
+    """
+
+    created: datetime = Field(
+        title="The timestamp when this resource was created."
+    )
+    updated: datetime = Field(
+        title="The timestamp when this resource was last updated."
+    )
```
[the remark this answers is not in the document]
stefannica: Or not. In fact, this is an improvement.

src/zenml/models/v2/base/base.py
```diff
@@ -150,7 +242,8 @@ def __eq__(self, other: Any) -> bool:
             return False
 
     def _validate_hydrated_version(
-        self, hydrated_model: "BaseResponse[AnyBody, AnyMetadata]"
+        self,
+        hydrated_model: "BaseIdentifiedResponse[AnyDatedBody, AnyMetadata, AnyResources]",
```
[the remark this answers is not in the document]
stefannica: Done

src/zenml/models/v2/base/base_plugin_flavor.py
```diff
@@ -0,0 +1,77 @@
...
+    subtype: PluginSubType = Field(title="Subtype of the plugin.")
+
+    class Config:
+        """Configuration for base plugin flavor response."""
+
+        extra = Extra.ignore
+
+    def get_hydrated_version(
+        self,
+    ) -> "BasePluginFlavorResponse[Any, Any, Any]":
```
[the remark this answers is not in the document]
stefannica: done

src/zenml/actions/base_action.py
```diff
@@ -90,5 +92,7 @@ def to_model(self) -> ActionFlavorResponse:
         """Convert the Flavor into a Flavor Response Model."""
         return ActionFlavorResponse(
             name=self.FLAVOR,
+            type=self.TYPE,
+            subtype=self.SUBTYPE,
             config_schema=self.get_action_config_schema(),
         )
```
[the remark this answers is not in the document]
stefannica: done

src/zenml/models/v2/core/event_source.py
```diff
@@ -125,13 +129,13 @@ class EventSourceResponseBody(WorkspaceScopedResponseBody):
...
     )
+
     created: datetime = Field(
-        title="The timestamp when this event filter was created."
+        title="The timestamp when this component was created."
     )
     updated: datetime = Field(
-        title="The timestamp when this event filter was last updated.",
+        title="The timestamp when this component was last updated.",
     )
```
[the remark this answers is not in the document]
stefannica: @AlexejPenner I made some changes related to this:
* made created and updated non-optional in the base model, just like you did before
* removed created and updated fields from all inheriting bodies mentioned here

src/zenml/models/v2/base/base.py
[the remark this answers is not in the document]
AlexejPenner: They were optional in fact, Apparently we do have some models in the database with no created field :cry:

src/zenml/models/v2/base/base.py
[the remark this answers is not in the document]
AlexejPenner: THanks

src/zenml/models/v2/core/event_source.py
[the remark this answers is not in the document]
AlexejPenner: nice, I really wanted this, just didn't get around to handling this for all implementations that have no created/updated fields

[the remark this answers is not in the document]
stefannica: which implementations/schemas exactly are you talking about ? Now that created and updated are mandatory in all Identified models, I expected mypy to complain if we happen to convert schemas that don't have created/updated fields to Identified models, but everything is green. I hope I'm not missing something that would explode at runtime (but then again, that's why we have unit tests).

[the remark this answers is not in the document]
stefannica: anyway, I went through all schemas and all those that had `to_model` methods also had created/updated fields, so I don't think there are exceptions

src/zenml/models/v2/base/base.py
stefannica: done (the first part)

stefannica (approved):

PR #2442, opened 2024-02-15
Use `uv` to speed up pip installs & the CI in general

strickvl:
Uses [`uv`](https://github.com/astral-sh/uv) to speed up Python package installation in the CI.

I also removed the cache completely, so @safoinme not sure if we need the infra for that any more? (`uv` is so fast we don't need it).


strickvl:
Pausing work on this until https://github.com/astral-sh/uv/issues/1526 is added / released.

tests/unit/utils/test_source_utils.py
strickvl, line 1: @schustmi not sure if this counts as 'fixing the test' or not. `uv` seems to install Python differently on Windows (see full discussion [here](https://github.com/astral-sh/uv/issues/2056)) which results in different behaviour when on windows for these two source utils for `defaultdict`. Any objections, or should I dive deeper into what's different?

safoinme:
Let's start with no caching, but we will need to have it later on as part of the Azure cost optimization for these runners, it's just a question of how does uv deal with cache

avishniakov (approved):

tests/unit/utils/test_source_utils.py
strickvl: How about like this @schustmi? Unfortunately `int` doesn't have an associated file, so we can't check for that dunder attribute.

PR #2452, opened 2024-02-20
Upgrading SQLModel to the latest version

bcdurak:
## Describe changes
I upgraded the version of the [SQLModel](https://sqlmodel.tiangolo.com) package that we are using from `0.0.8` to `0.0.16`.

The most critical factor in this upgrade stems from [this PR right here](https://github.com/tiangolo/sqlmodel/pull/669). With this change, they have changed the way that they handle `Enum` values.

For instance, if you are familiar with our [component schema](https://github.com/zenml-io/zenml/blob/main/src/zenml/zen_stores/schemas/component_schemas.py) (which we defined through SQLModel), we added a field called `type` which was a `StackComponentType`:

```python
# the schema of the stack component
class StackComponentSchema(...):
    ...
    type: StackComponentType
    ...
    
# and the stack component type looked like this:
class StackComponentType(StrEnum):
    ...
    ARTIFACT_STORE = "artifact_store"
    ...
```

With this setup, when we registered, for instance, a new artifact store, we created an entry in the components table of our DB where the column `type` had the string `artifact_store` stored in it as a value. However, with the new changes, SQLModel now gives higher priority to `Enum` fields and saves the value `ARTIFACT_STORE` instead. While this is alright if you are starting from scratch, if you have any entry in a table with an `Enum` field zenml will fail after the upgrade. Instead of taking the migration route, we decided to adjust our schemas to use str fields instead and updated the corresponding `to_model`, `update` and `from_request` methods.

@avishniakov every enum field is using a similar approach to the one that you took with the `TagSchema`s and their color field.

**Additionally**, I have removed `fastapi_utils` as a dependency as it was completely unused and tweaked the code a bit to work with the new SQLModel version.

## Pre-requisites

## Types of changes


strickvl:
@coderabbitai review

avishniakov (changes requested):
Agree with most things @coderabbitai highlighted + test are failing now

src/zenml/integrations/evidently/__init__.py
```diff
@@ -54,7 +54,7 @@ class EvidentlyIntegration(Integration):
     """[Evidently](https://github.com/evidentlyai/evidently) integration for ZenML."""
 
     NAME = EVIDENTLY
-    REQUIREMENTS = ["evidently>0.2.6,<0.4.5"]  # supports pyyaml 6
+    REQUIREMENTS = ["evidently>0.2.6,<0.4.5", "sqlmodel<0.0.10"]  # supports pyyaml 6
```
strickvl, line 57: should add a comment on why we've added sqlmodel here

src/zenml/integrations/airflow/__init__.py
```diff
@@ -30,7 +30,7 @@ class AirflowIntegration(Integration):
     """Definition of Airflow Integration for ZenML."""
 
     NAME = AIRFLOW
-    REQUIREMENTS = ["apache-airflow~=2.4.0"]
+    REQUIREMENTS = ["apache-airflow~=2.4.0", "sqlmodel<0.0.10"]
```
strickvl, line 33: should add a comment on why we've added sqlmodel here

src/zenml/integrations/evidently/__init__.py
strickvl: and maybe a note in the docs that this is what's happening?

src/zenml/integrations/airflow/__init__.py
strickvl: and maybe a note in the docs that this is what's happening?

src/zenml/integrations/great_expectations/__init__.py
```diff
@@ -35,6 +35,7 @@ class GreatExpectationsIntegration(Integration):
         "great-expectations>=0.15.0,<=0.15.47",
         # typing_extensions 4.6.0 and above doesn't work with GE
         "typing_extensions<4.6.0",
+        "sqlmodel<0.0.10",
```
strickvl, line 38: should add a comment on why we've added sqlmodel here. and maybe a note in the docs that this is what's happening?

strickvl (changes requested):
LGTM aside from the comments I already made.

strickvl (approved):

src/zenml/integrations/airflow/__init__.py
```diff
@@ -32,7 +32,14 @@ class AirflowIntegration(Integration):
...
     # remove pendulum version requirement once Airflow supports
     # pendulum>-3.0.0
-    REQUIREMENTS = ["apache-airflow~=2.4.0", "pendulum<3.0.0"]
+    REQUIREMENTS = [
+        "apache-airflow~=2.4.0",
+        "pendulum<3.0.0",
+        # We need to add this as an extra dependency to manually downgrade
+        # SQLModel. Otherwise, the initial installation of ZenML installs
+        # a higher version SQLModel and a version mismatch is created.
+        "sqlmodel>=0.0.9,<=0.0.16",
```
avishniakov, line 41: Is this still needed? You have exactly the same in `pyproject.toml` now...

src/zenml/integrations/evidently/__init__.py
```diff
@@ -54,7 +54,13 @@ class EvidentlyIntegration(Integration):
     """[Evidently](https://github.com/evidentlyai/evidently) integration for ZenML."""
 
     NAME = EVIDENTLY
-    REQUIREMENTS = ["evidently>0.2.6,<0.4.5"]  # supports pyyaml 6
+    REQUIREMENTS = [
+        "evidently>0.2.6,<0.4.5",  # supports pyyaml 6
+        # We need to add this as an extra dependency to manually downgrade
+        # SQLModel. Otherwise, the initial installation of ZenML installs
+        # a higher version SQLModel and a version mismatch is created.
+        "sqlmodel>=0.0.9,<=0.0.16"
```
avishniakov, line 62: same

src/zenml/integrations/airflow/__init__.py
bcdurak: Yes, if we don't put this here, when someone installs ZenML, it installs sqlmodel 0.0.16. When they install the airflow integration on top, it downgrades sqlalchemy but not the sqlmodel, which creates a drift.

david101-hunter:
Hi, after I upgrade sqlmodel version, I encounter issue
`zenml 0.56.3 requires sqlmodel==0.0.8, but you have sqlmodel 0.0.16 which is incompatible.`

I want to intergate feast (0.36.0), however:

`zenml 0.56.3 requires pydantic<1.11,>=1.9.0`
`feast 0.36.0 requires pydantic>=2.0.0, but you have pydantic 1.10.15 which is incompatible`

How I can fix that.

PR #2453, opened 2024-02-20
Add service account and auth window to trigger

stefannica:
## Describe changes
Associates a mandatory service account with every trigger. A service account is required for two reasons:
* to generate API tokens needed by remote workloads like pipeline runs to authenticate and communicate with the server.  
* to associate an identity with every operation that the action will perform on ZenML resources

Other related changes:
* moved plugin registry singleton and initialization to the REST API to avoid circular deps
* re-generated DB migration scripts
* removed unnecessary scheduler files

Important points:

1. there is no implicit global RBAC check performed on operations that the action execution implementation executes while running in the context of the server. For example, fetching a stack or a pipeline is completely left as a responsibility of the action implementation. Currently, the pipeline run action implementation doesn't perform any checks to verify that the service account is allowed to access the pipeline, deployment, stack, components etc.

2. the action workloads (e.g. remote pipeline runs) are issued a temporary API token associated with the service account linked to the trigger instead of the API key for the following reasons:
  * the API key is a long-lived credential that is not safe to distribute to external workloads.
  * the API key is only visible during creation and rotation and is otherwise stored encrypted in the DB
The API token has an expiration configurable through the auth_window trigger attribute. If not explicitly configured for the trigger, the action implementation decides what default value to use (48 hours in the case of the pipeline run action).

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/479103df60b6_add_triggers.py
```diff
@@ -0,0 +1,161 @@
+"""add triggers [479103df60b6].
+
```
AlexejPenner, line 2: You've really thought of everything -  I was getting ready to do this one once your branch was merged.

AlexejPenner (approved):
:seal: Lets get this to our mega branch

PR #2457, opened 2024-02-21
Disable methods for local case

AlexejPenner:
## Describe changes
Some methods should not be allowed for the fully local setup where the client is directly connected to the SqlZenStore. This is because there is some custom logic, that is only run in the layer between endpoint and database:

Client -> RestZenStore -> Server-Endpoint -> Custom handler -> SqlZenStore

This custom layer would be bypassed in the local setup:

Client -> SqlZenStore

For this reason the relevant methods are disallowed.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -188,6 +188,7 @@ if TYPE_CHECKING:
 logger = get_logger(__name__)
 
 AnyResponse = TypeVar("AnyResponse", bound=BaseIdentifiedResponse)  # type: ignore[type-arg]
+T = TypeVar("T")
 
 
 class ClientConfiguration(FileSyncModel):
@@ -365,7 +366,7 @@ class Client(metaclass=ClientMetaClass):
         cls._global_client = client
 
     @staticmethod
-    def _fail_for_sql_zen_store(method):
+    def _fail_for_sql_zen_store(method: Callable[..., T]) -> Callable[..., T]:
         """Decorator for all methods, that are disallowed when the client is not connected through REST API.
 
         Args:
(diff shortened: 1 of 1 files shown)


src/zenml/client.py
```diff
@@ -364,6 +364,30 @@ def _reset_instance(cls, client: Optional["Client"] = None) -> None:
...
+            method: The method
+
+        Returns:
+            The decorated method.
+        """
+
+        @functools.wraps(method)
+        def wrapper(self, *args, **kwargs):
+            # No isinstance check to avoid importing ZenStore implementations
+            breakpoint()
```
avishniakov, line 381: I think this shouldn't be here :)

src/zenml/client.py
```diff
@@ -364,6 +364,30 @@ def _reset_instance(cls, client: Optional["Client"] = None) -> None:
...
+            if self.zen_store.__class__.__name__ == "SqlZenStore":
+                raise TypeError(
+                    "This method is not allowed when not connected "
+                    "to a ZenML Server through the API interface.."
+                )
+            return method(*args, **kwargs)
```
avishniakov, line 387:
```suggestion
        @functools.wraps(method)
        def wrapper(self, *args, **kwargs):
            # No isinstance check to avoid importing ZenStore implementations
            if self.zen_store.__class__.__name__ == "SqlZenStore":
                raise TypeError(
                    "This method is not allowed when not connected "
                    "to a ZenML Server through the API interface.."
                )
            return method(self, *args, **kwargs)
```

src/zenml/client.py
AlexejPenner: This is definitely a necessary part :stuck_out_tongue:

AlexejPenner:
> With the suggested changes E2E seems to run smoothly on my end. Consider adding test cases for this wrapper.

Tests will be written on the mega branch that we are merging to. applied your review

stefannica (approved):

stefannica (changes requested):
Alexej, can you please look at the linter errors ? I think you have to add some type annotations to the decorator.

AlexejPenner:
> Alexej, can you please look at the linter errors ? I think you have to add some type annotations to the decorator.

done

stefannica (approved):

PR #2459, opened 2024-02-21
Fix local Airflow install + docs instructions

strickvl:
Our docs were out of date and it was no longer working for local airflow installation.


bcdurak (approved):
:seal:

avishniakov (approved):

PR #2464, opened 2024-02-23
Add option to add base URL for zenml server (with support for cloud)

wjayesh:
## Describe changes
I implemented changes to allow

- people to set base URLs for their ZenML server
- for ZenML Cloud tenants to set the same along with an option to add organization ID and tenant ID to it to construct a full  cloud URL.

This is exposed through the `/info` endpoint of the server. The use case is to allow ZenML to use the URL to construct addresses for stuff like pipeline runs and other info on the server.

## Pre-requisites

## Types of changes


AlexejPenner:
I think for the future we will also need the org id separately, I would not make it part of the base url.

Base url should be: https://cloud.zenml.io/
then the tenant should also have its organization_id available to it to construct a full url

I believe the tenant_id is already there

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1434,6 +1434,19 @@ def get_store_info(self) -> ServerModel:
...
+        # if ZENML_SERVER_BASE_URL is set as an environment variable, use it
+        # as the base URL for the server
+        base_url = os.environ.get("ZENML_SERVER_BASE_URL")
+
+        # if cloud org ID and tenant ID are set as environment variables, construct
+        # the URL by appending the org ID and tenant ID to the base URL
+        if base_url:
+            model.server_url = base_url
+            cloud_org_id = os.environ.get("ZENML_CLOUD_ORGANIZATION_ID")
+            cloud_tenant_id = os.environ.get("ZENML_CLOUD_TENANT_ID")
```
stefannica: This works, but makes the ZenML OSS code too interlinked with the Cloud API and that makes it difficult to roll out changes independently.

Imagine that in the future, we might change the tenant URL to be constructed differently. All tenants currently deployed at that time will be instantly invalidated and they would all need to be upgraded for the change to take effect.

May I suggest a better solution that keeps things nicely decoupled:

* construct and pass the full `{model.url}/organizations/{cloud_org_id}/tenants/{cloud_tenant_id}"` URL in the `ZENML_SERVER_BASE_URL` env variable instead
* don't use any mention of cloud organization and tenant here

stefannica (changes requested):

src/zenml/zen_stores/sql_zen_store.py
stefannica: @AlexejPenner what do you think about this?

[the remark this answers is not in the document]
stefannica: @wjayesh one more thing: as @AlexejPenner pointed out, the `external_server_id` attribute in the `ServerConfiguration` is already set to the tenant UUID, you won't need another attribute.

AlexejPenner: I'm not sure I understand what you are suggesting. Are you saying we shouldn't have ORG_ID as a separate property in the sql zen store or as as separate env_var in general?

I believe we will need the org id for event aggregation - I wouldn't want to have to splice this out of some URL from the environment. I also think its fine here to acknoledge on the Tenant side, that something like a ZenML Cloud Org entity exists. However, I don't see why we would need the org_id in the sql zen store

schustmi: I agree with @stefannica that this should probably be passed as the full tenant URL.

For the analytics we still need the organization ID, it would be nice if you could include that as part of this PR in the `ServerConfiguration`

wjayesh: I added a `cloud_organization_id` as a variable in the ServerConfiguration. I was leaning towards adding the `metadata` field instead and having the org ID as a key in it but how far would it go in decoupling OSS and cloud?
When we want to access the value for analytics, we would still need to know what that key is called and this would inevitably lead to coupling. Am I missing something? @stefannica

stefannica: @wjayesh if you're still exploring this: you could add a `metadata: Dict[str, str]` field in the `ServerConfiguration` and pass that as an environment variable containing additional cloud specific values e.g. `ZENML_SERVER_METADATA='{"cloud_organization_id: "<org-UUID>"}'. You wouldn't need to know what the key is called on the OSS side if we pass the whole `server_config().metadata` dictionary to the telemetry without caring what's inside it.

stefannica (approved):
This looks good now, but if you want to go the extra mile and model the `cloud_organization_id` setting as a generic `metadata` dictionary instead, it would be even better.

src/zenml/zen_stores/sql_zen_store.py
stefannica: With a generic `metadata` field in the server settings, we can add whatever values we want in there on the Cloud API side without the OSS code even being aware of it and have those values published in the telemetry, shown in the server information and probably used in other places in the future (e.g. event logging).

src/zenml/models/v2/misc/server_models.py
```diff
@@ -73,6 +73,14 @@ class ServerModel(BaseModel):
...
         title="The authentication scheme that the server is using.",
     )
+    base_url: str = Field(
+        "",
+        title="The Base URL of the server.",
+    )
+    metadata: dict = Field(
+        {},
+        title="The metadata associated with the server.",
+    )
```
stefannica, line 83:
```suggestion
    metadata: Dict[str, str] = Field(
        default_factory=dict,
        title="The metadata associated with the server.",
    )
```

and you probably need to import Dict from typing too

stefannica (changes requested):

src/zenml/config/server_config.py
```diff
@@ -144,6 +149,8 @@ class ServerConfiguration(BaseModel):
     external_cookie_name: Optional[str] = None
     external_server_id: Optional[UUID] = None
 
+    metadata: Optional[Dict[str, Any]] = None
```
stefannica, line 152: You need to add some code to the root validator below to convert the metadata from string/JSON format into dictionary. Something like this:

```
        if isinstance(values.get("metadata"), str):
            try:
                values["metadata"] = json.loads(values["metadata"])
            except json.JSONDecodeError as e:
                raise ValueError(
                    f"The server metadata is not a valid JSON object: {e}"
                )
```

For an example of this, please see here: https://github.com/zenml-io/zenml/blob/develop/src/zenml/zen_stores/secrets_stores/service_connector_secrets_store.py#L72-L80

stefannica (changes requested):

src/zenml/config/server_config.py
wjayesh: ah yes, good catch! I had the impression that I had to load it into json somewhere but then this thought got lost with all the code re-org between the repos

stefannica (approved):

src/zenml/zen_server/deploy/helm/templates/_environment.tpl
```diff
@@ -82,6 +82,9 @@ external_server_id: {{ .ZenML.auth.externalServerID | quote }}
 {{- if .ZenML.rootUrlPath }}
 root_url_path: {{ .ZenML.rootUrlPath | quote }}
 {{- end }}
+{{- if .ZenML.baseURL }}
```
schustmi, line 85: Should the metadata also be a value here?

src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
```diff
@@ -91,6 +91,10 @@ spec:
               value: {{ $value | quote }}
             {{- end }}
             {{- end }}
+            {{- if .Values.zenml.baseURL }}
```
schustmi, line 94: Same here for the metadata? Or is this something that's supposed to be a little more hidden that we only use for cloud tenants?

wjayesh: Yes, that was my intention. I have kept it out of the helm chart for now but we can add it in if there's a need.

PR #2468, opened 2024-02-24
Telemetry guarding for CI & editable installs

strickvl:
This pull request adds extra telemetry guarding to the codebase. It also includes environment variables `ZENML_ANALYTICS_OPT_IN` and `ZENML_DEBUG` with their respective values set to `false` and `true`. This applies to the CI systems in action.

Also uses editable installs for DB testing (was installing from PyPI previously).


htahir1 (approved):
I hope this just guardrails ...

avishniakov (approved):

PR #2470, opened 2024-02-26
Update `.coderabbit.yaml`

strickvl:
Tiny nit. Just noticed the config file wasn't quite right.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -2,14 +2,14 @@ language: "en"
 early_access: false
 reviews:
   high_level_summary: true
-  poem: false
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -2,14 +2,14 @@ language: "en"
 early_access: false
 reviews:
   high_level_summary: true
-  poem: false
+  poem: true
```
avishniakov, line 5: What does it mean?

strickvl: :) it writes a poem about your commit.

PR #2473, opened 2024-02-27
Fix broken docs links

strickvl:
As per Slack message:

> "FYI, several documentation links are referencing non-existing README files from the repo, for example https://docs.zenml.io/user-guide/production-guide/cloud-orchestration has a link to https://github.com/zenml-io/zenml/blob/release/0.55.3/docs/book/stacks-and-components/component-guide/orchestrators/README.md which does not exist (I've seen more examples like this)"


bcdurak (approved):
🦭

PR #2474, opened 2024-02-28
Use `uv pip compile` for environment setup in CI

strickvl:
We now use `uv pip compile` to resolve all our installations on runners for testing in setup.

This allows our CI to install everything, but has the downside that we're starting to depart a bit from the tooling that users would / do use to install zenml (i.e. `pip`). On the other side of that is the fact that previously as per the original setup behaviour, we actually install zenml in a way that no user would -- i.e. almost all integrations in the same environment, and the same with all extras -- but probably for the moment the tradeoff of this update / change.

We're also now installing zenml as well as langchain during the integration installation function, so I removed the duplicate installation steps for both.


strickvl:
Merging now as CI previously passed with this code -- https://github.com/zenml-io/zenml/pull/2474/checks?sha=859d280b37eae7e27ca1304143ef87d2879a1502.

PR #2484, opened 2024-02-29
Rate limiting for login API

avishniakov:
## Describe changes
I fixed the missing rate limit for login API to prevent misuse and too frequent login attempts.

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
	- Introduced rate limiting on authentication requests to enhance security and manage traffic effectively. Users can now make up to 5 requests per minute and 1000 requests per day.


avishniakov:
@coderabbitai review

AlexejPenner (approved):
:seal: - Looks good as is, feel free to unify with cloud rate limiting

pyproject.toml
```diff
@@ -73,6 +73,7 @@ fastapi-utils = { version = "~0.2.1", optional = true }
 orjson = { version = "~3.8.3", optional = true }
 Jinja2 = { version = "*", optional = true }
 ipinfo = { version = ">=4.4.3", optional = true }
+slowapi = { version = ">=0.1.9", optional = true }
```
strickvl, line 76: Don't we have a clash between our pytest and the one specified in slowapi? 

`pytest = "~=6.2.5"`

https://github.com/laurentS/slowapi/blob/6547679463b6a6fa87d9e254a8558bbc6075e263/pyproject.toml#L31C1-L31C19

AlexejPenner: @strickvl this might not be the case anymore if we switch to fastapi_limiter which Andrei
 told me he would look into

stefannica: fastapi_limiter uses redis to perform rate limiting. Redis needs to be installed separately. This is not something that we want to make mandatory.

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -256,6 +259,7 @@ def generate_access_token(
     response_model=Union[OAuthTokenResponse, OAuthRedirectResponse],
 )
 @handle_exceptions
+@limiter.limit("5/minute;1000/day")
```
stefannica, line 262: Adding this will throw an exception that is not handled by the `@handle_exceptions` decorator and ends up being incorrectly reported as a 500 internal server error. It should be properly caught and converted into a user-friendly authorization error.

This is what the client sees:

```
RuntimeError: Error initializing rest store with URL 'http://localhost:8237': <slowapi.wrappers.Limit object at 0x7fc80523edc0>
```

This is what the server throws:

```
ratelimit 5 per 1 minute (127.0.0.1) exceeded at endpoint: /api/v1/login
API error
  File "/home/stefan/aspyre/src/zenml/src/zenml/zen_server/utils.py", line 304, in decorated
    return func(*args, **kwargs)
  File "/home/stefan/aspyre/src/zenml/.venv/lib/python3.8/site-packages/slowapi/extension.py", line 764, in sync_wrapper
    self._check_request_limit(request, func, False)
  File "/home/stefan/aspyre/src/zenml/.venv/lib/python3.8/site-packages/slowapi/extension.py", line 630, in _check_request_limit
    self.__evaluate_limits(request, _endpoint_key, all_limits)
  File "/home/stefan/aspyre/src/zenml/.venv/lib/python3.8/site-packages/slowapi/extension.py", line 528, in __evaluate_limits
    raise RateLimitExceeded(failed_limit)
slowapi.errors.RateLimitExceeded: <slowapi.wrappers.Limit object at 0x7fc80523edc0>
INFO:     127.0.0.1:47208 - "POST /api/v1/login HTTP/1.1" 500 Internal Server Error
```

stefannica (changes requested):
I love how simple `slowapi` is. I don't recommend we use fastapi_limiter for this, it's overkill.

I already closed the original vulnerability as informational on account of us not advertising the username/password authentication method as a full-fledged production-grade security scheme. If `slowapi` doesn't work out, we could also give up and not fix this at all.

src/zenml/zen_server/routers/auth_endpoints.py
avishniakov: Yeah, good catch. I just moved it around and now on UI user will see `5 per 1 minute` and on the server:
```
ratelimit 5 per 1 minute (127.0.0.1) exceeded at endpoint: /api/v1/login
INFO:     127.0.0.1:63151 - "POST /api/v1/login HTTP/1.1" 429 Too Many Requests
```
This seems good enough for me.

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -255,6 +258,7 @@ def generate_access_token(
     LOGIN,
     response_model=Union[OAuthTokenResponse, OAuthRedirectResponse],
 )
+@limiter.limit("5/minute;1000/day")
```
stefannica, line 261: @avishniakov could you also make these values configurable? I recommend you use the ServerSettings class for this and just add two more configuration options for the 5 and 1000 values.

avishniakov: Ah, yeah, I wanted to do this but forgot - thx for the reminder. We talked this through with Alexej yesterday

AlexejPenner:
> I love how simple `slowapi` is. I don't recommend we use fastapi_limiter for this, it's overkill.
> 
> I already closed the original vulnerability as informational on account of us not advertising the username/password authentication method as a full-fledged production-grade security scheme. If `slowapi` doesn't work out, we could also give up and not fix this at all.

good catch - forgot about redis

stefannica (approved):

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -255,6 +258,10 @@ def generate_access_token(
     LOGIN,
     response_model=Union[OAuthTokenResponse, OAuthRedirectResponse],
 )
+@limiter.limit(
+    f"{server_config().login_rate_limit_minute}/minute;"
+    f"{server_config().login_rate_limit_day}/day"
+)
```
stefannica, line 264: I'm a bit worried about the fact that this rate limiting is counting all calls to this endpoint, even those that are legit. There might be cases in which logging with valid credentials will run into rate limiting errors. One example I can think of is many pipeline run steps running from the same location and logging in with service account API keys for example.

stefannica (changes requested):

src/zenml/zen_server/routers/auth_endpoints.py
stefannica: I think there's a way to customize the conditions for rate-limiting. This GitHub issue might be useful: https://github.com/laurentS/slowapi/issues/31

avishniakov: There are ways to customize that, cause I just set a key function to differentiate requests. The problem is that we have only the request at hand, not the response at the time of the decision.
I will merge this and do more research.

avishniakov: @stefannica WDYT: https://github.com/zenml-io/zenml/pull/2484/commits/f6aaf7270b7c01bf8ad12d3e18e2832568099a7c

stefannica:
I just realized there's another reason why we need to be very very careful with this: rate-limiters rely on the ability to extract the *real* source IP address from an incoming request, but this can be difficult to achieve, especially with cloud or Kubernetes deployments, because a load balancer or proxy usually sits in front of the ZenML server and rewrites the source IP address.

If mis-configured, this can have disastrous consequences, because all requests are "seen" as coming from the same IP address. For example, someone trying to brute-force their way into the server can become a DoS attack, because everyone will be locked out of the server.

More info here:
* https://stackoverflow.com/questions/60098005/fastapi-starlette-get-client-real-ip
* https://www.uvicorn.org/settings/#http

I think the safe thing to do is to have rate-limiting turned off by default and have proper documentation around how to properly configure it and the services around it to avoid problems.

avishniakov:
> I just realized there's another reason why we need to be very very careful with this: rate-limiters rely on the ability to extract the _real_ source IP address from an incoming request, but this can be difficult to achieve, especially with cloud or Kubernetes deployments, because a load balancer or proxy usually sits in front of the ZenML server and rewrites the source IP address.
> 
> If mis-configured, this can have disastrous consequences, because all requests are "seen" as coming from the same IP address. For example, someone trying to brute-force their way into the server can become a DoS attack, because everyone will be locked out of the server.
> 
> More info here:
> 
> * https://stackoverflow.com/questions/60098005/fastapi-starlette-get-client-real-ip
> * https://www.uvicorn.org/settings/#http
> 
> I think the safe thing to do is to have rate-limiting turned off by default and have proper documentation around how to properly configure it and the services around it to avoid problems.
> 
> See this PR for a bug related to the same problem: #2499

Yeah, this makes sense to me, but let's build on top of that:
If k8s or other infra is bombarding REST with incorrect credentials, why should it pass? Successful requests do not affect limits. This might be an issue indeed for users behind VPN, for instance (they all might land on the same IP).

Due to all those complications, we can drop this feature completely for OSS. Based on my research - everything inside HTTP request can be spoofed, so we cannot rely really on any header or other info to make a weighted decision if this is the same requestor or not. I don't know what folks from CDN solution use to prevent DDoS and other things, but we are not CDN provider 🙂

stefannica:
> Yeah, this makes sense to me, but let's build on top of that: If k8s or other infra is bombarding REST with incorrect credentials, why should it pass? Successful requests do not affect limits. This might be an issue indeed for users behind VPN, for instance (they all might land on the same IP).
> 
> Due to all those complications, we can drop this feature completely for OSS. Based on my research - everything inside HTTP request can be spoofed, so we cannot rely really on any header or other info to make a weighted decision if this is the same requestor or not. I don't know what folks from CDN solution use to prevent DDoS and other things, but we are not CDN provider 🙂

There's probably been a misunderstanding, it's not really k8s or other infra that's bombarding REST with incorrect credentials. This is what can happen if the ZenML server or the kubernetes cluster or load balancer, or proxy is incorrectly configured:
* let's say you have 100 legit users regularly logging in
* let's also say you have an attacker trying to brute-force his way into the server
* when incorrectly configured, the server "sees" all requests coming from a single IP: the IP of the proxy or load-balancer
* the "bad" requests will pile up until the limiter starts blocking traffic from that IP address, effectively stopping everyone from logging in, even the legit users

We should still offer the limiter as a feature, because it's a great thing to have for security, but have some way of configuring it and keeping it disabled by default. If the admins enable it, it's their responsibility to correctly configure the surrounding infrastructure (proxy/load-balancer/whatever).

avishniakov:
@stefannica I added some docs describing how to configure it and cancel the current CI - the last one was all green. Please have a look.

docs/book/deploying-zenml/zenml-self-hosted/deploy-with-docker.md
```diff
@@ -44,6 +44,9 @@ The following environment variables can be passed to the container:
...
+* **ZENML\_SERVER\_RATE\_LIMIT\_ENABLED**: This variable controls the rate limiting for ZenML API (currently only LOGIN endpoint). It is disabled by default, so tet it to `1` only if you need to enable rate limiting. To determine unique users a `X_FORWARDED_FOR` header or `request.client.host` is used, so before enabling this make sure that your network configuration is associating proper information to your clients in order to avoid disruptions for legitimate requests.
+* **ZENML\_SERVER\_LOGIN\_RATE\_LIMIT\_MINUTE**: If rate limiting is enabled, this variables controls how many requests will be allowed to login endpoint in one minute interval. Set it to a desired integer value, defaults to `5`.
+* **ZENML\_SERVER\_LOGIN\_RATE\_LIMIT\_DAY**: If rate limiting is enabled, this variables controls how many requests will be allowed to login endpoint in one day interval. Set it to a desired integer value, defaults to `1000`.
```
strickvl, line 49:
```suggestion
* **ZENML\_SERVER\_RATE\_LIMIT\_ENABLED**: This variable controls the rate limiting for ZenML API (currently only for the `LOGIN` endpoint). It is disabled by default, so set it to `1` only if you need to enable rate limiting. To determine unique users a `X_FORWARDED_FOR` header or `request.client.host` is used, so before enabling this make sure that your network configuration is associating proper information with your clients in order to avoid disruptions for legitimate requests.
* **ZENML\_SERVER\_LOGIN\_RATE\_LIMIT\_MINUTE**: If rate limiting is enabled, this variable controls how many requests will be allowed to query the login endpoint in a one minute interval. Set it to a desired integer value; defaults to `5`.
* **ZENML\_SERVER\_LOGIN\_RATE\_LIMIT\_DAY**: If rate limiting is enabled, this variable controls how many requests will be allowed to query the login endpoint in an interval of day interval. Set it to a desired integer value; defaults to `1000`.
```

strickvl (changes requested):

src/zenml/zen_server/utils.py
```diff
@@ -318,6 +335,70 @@ def decorated(*args: Any, **kwargs: Any) -> Any:
...
+                    identifier = limiter._key_func(request)
+                    if "memory" not in (limiter._storage.STORAGE_SCHEME or []):
+                        raise NotImplementedError(
+                            "Limiter not running in memory is not yet supported."
+                        )
+                    for k in limiter._storage.storage.keys():  # type: ignore[union-attr]
+                        if api_route in k:
+                            _, identifier_ = k.split("//")[0].split("/")
+                            if identifier_ == identifier:
+                                limiter._storage.storage[k] -= 1  # type: ignore[union-attr]
```
stefannica, line 394: You're digging deep into the slowapi implementation, essentially hacking its internal state, to achieve a behavior that the tool clearly wasn't designed for. This makes it really difficult to maintain this, for the following reasons:

* it's really difficult to understand what this hack does without knowing a lot more about the slowapi internals
* the slowapi library itself is still in alpha stage, which means it's still subject to change a lot over the next releases and that means it's very probable that this code will be invalidated when we decide to update to a newer version

Instead, we should try to use the public (documented) APIs to achieve our goal, or work to implement these new features in the original project. As a last resort, this could be achieved by temporarily sub-classing the `Limiter` class and adding the missing functionality there, if this is even possible.

stefannica: I gave this more thought and spend a lot of time trying to find a solution for this that doesn't require us to mess with the Limiter's internals and I realized we don't even have a good understanding of what it is we want to achieve, so I'll try to summarize it here, maybe get some alignment.

IMO there are two end goals:

1. make it really difficult for someone to use brute-force to hack into an account (e.g. guess a username, password, an API token or invitation token) by allowing only a limited number of failed login attempts. We should also make sure that even if a successful login attempt is made after a series of failed ones, nothing in the server's response can be used to infer this. 
2. don't impact "good" users that use valid credentials to log in on the first try. Even if they fail to login for a couple of times, we should make it easy for them to recover.

I think the following mechanism achieves both goals, but I'm open to suggestions for how we can make it more robust:

1. the limiter should be applied for a source IP (i.e. start counting requests and raising limiter errors) only when the first failed login happens, but not before that.
2. the limiter should stop being applied for a source IP until the following happens, at which time, the limiter should be reset to its original state (i.e. its counters should be reset):
  * a successful login request is received
  * there is no time restriction in place (i.e. the request would be allowed by the limiter)

If we can't achieve this with the official slowapi features, I'd be more inclined to implement our own, slim, in-memory limiter, targeted to our reduced set of features, rather than hacking slowapi.

strickvl (approved):

src/zenml/zen_server/utils.py
```diff
@@ -318,6 +337,150 @@ def decorated(*args: Any, **kwargs: Any) -> Any:
...
+        Decorated function.
+    """
+    limiter = RequestLimiter(day_limit=day_limit, minute_limit=minute_limit)
+
+    def decorator(func: F) -> F:
+        @wraps(func)
+        def decorated(
+            request: Request,
+            response: Response,
+            auth_form_data: "OAuthLoginRequestForm",
```
stefannica, line 464: modelling this explicitly as an argument makes the decorator less reusable. Why not have `*args` and `**kwargs` instead?

src/zenml/zen_server/utils.py
```diff
@@ -318,6 +337,150 @@ def decorated(*args: Any, **kwargs: Any) -> Any:
...
+                ret = func(
+                    request=request,
+                    response=response,
+                    auth_form_data=auth_form_data,
+                )
+            except Exception as e:
+                raise e
+            else:
+                limiter.reset_limiter(request)
+            return ret
```
stefannica, line 477:
```suggestion
            limiter.hit_limiter(request)
            ret = func(
                request=request,
                response=response,
               auth_form_data=auth_form_data,
            )

            limiter.reset_limiter(request)
            return ret
```

This is equivalent.

src/zenml/zen_server/utils.py
```diff
@@ -318,6 +337,150 @@ def decorated(*args: Any, **kwargs: Any) -> Any:
...
+        older_index = None
+        for i, limiter_hit in enumerate(self.limiter[requester]):
+            if limiter_hit < day_ago:
+                older_index = i
+            else:
+                break
+        if older_index is not None:
+            self.limiter[requester] = self.limiter[requester][
+                older_index + 1 :
+            ]
```
stefannica, line 395:
```suggestion
        from bisect import bisect_left
        # remove failures older than a day
        older_index = bisect_left(self.limiter[requester], day_ago)
        self.limiter[requester] = self.limiter[requester][older_index:]
```

With large sorted lists, you can leverage a binary search algorithm to reduce the time to find the delimitation.

stefannica (approved):
Everything looks great. I love the minimalistic limiter implementation, definitely a lot better than having to hack our way into slowapi. I left a couple of comments with improvement suggestions. I leave it up to you if you want to address them or not.

PR #2487, opened 2024-03-01
Allow latest `ruff` versions

strickvl:
Templates are using any version of ruff, while zenml had pinned it to a lower version. This restores the balance so that all are using the latest version.


avishniakov:
I tried the latest, and it reformats a lot of files at once, you might need to do that on your local before pushing, otherwise lint check should fail.

avishniakov (approved):

PR #2498, opened 2024-03-06
Relax azure, gcfs and s3 dependencies

strickvl:
This pull request updates the version requirements for the azure, gcfs, and s3 dependencies to allow for more flexibility in the versions used.

In particular, this would unblock https://github.com/zenml-io/zenml/pull/2376.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ jinja2-time = { version = "^0.2.0", optional = true }
 python-terraform = { version = "^0.10.1", optional = true }
 
 # Optional dependencies for the AWS secrets store
-boto3 = { version = ">=1.16.0,<=1.24.59", optional = true }
+boto3 = { version = ">=1.16.0", optional = true }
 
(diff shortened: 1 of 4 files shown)


strickvl:
@stefannica just drawing your attention to this since I think a few of these restrictions were originally set by you. I have little sense of the consequence of relaxing the dependencies, but we need to find a way to allow for the huggingface_hub installation somehow and this was the path I found that seems to work...

safoinme (approved):

avishniakov (approved):
🦭

stefannica:
> @stefannica just drawing your attention to this since I think a few of these restrictions were originally set by you. I have little sense of the consequence of relaxing the dependencies, but we need to find a way to allow for the huggingface_hub installation somehow and this was the path I found that seems to work...

IIRC, one of the reasons why I pinned down boto3 was to speed up the pip dependency resolution. But that's no longer an issue, so this should be good to go.

PR #2506, opened 2024-03-08
Log URL for pipelines and model versions when running a pipeline

wjayesh:
## Describe changes
- I implemented a `get_model_version_url` to construct the URL for a model version inside dashboard utils.
- Updated the `get_base_url` function to use the URL from server configuration if it's set.
- Log pipeline URL before it is run.

## Pre-requisites

## Types of changes


src/zenml/model/model.py
```diff
@@ -722,7 +723,18 @@ def _get_or_create_model_version(
...
+            )
             logger.info(f"New model version `{self.version}` was created.")
+            if model_version_url:
+                logger.info(
+                    f"Model version dashboard URL: {model_version_url}"
+                )
+            else:
+                logger.info(
+                    "You can now view your Models in the ZenML dashboard using ZenML Cloud."
+                )
```
avishniakov, line 737:
```suggestion
            
            logger.info(f"New model version `{self.version}` was created.")
        model_version_url = dashboard_utils.get_model_version_url(
            model_version.id
        )
        if model_version_url:
            logger.info(
                f"Model version dashboard URL: {model_version_url}"
            )
        else:
            logger.info(
                "You can now view your Models in the ZenML dashboard using ZenML Cloud."
            )
```
This will make it visible for both cases: the new and existing model versions.

src/zenml/utils/dashboard_utils.py
```diff
@@ -34,6 +35,13 @@ def get_base_url() -> Optional[str]:
     client = Client()
 
     if client.zen_store.type == StoreType.REST:
+        # if the server config has a base URL use that
+        server_config = ServerConfiguration.get_server_config()
+        if server_config.base_url:
+            return (
+                server_config.base_url
```
schustmi, line 42: If the base URL is a cloud one, there is no `/workspaces/...` right? Maybe we need to differentiate here whether this is a cloud tenant (I guess if the base url contains `cloud.zenml.io`?) until we open-source the cloud dashboard and can unify these?

schustmi (changes requested):

src/zenml/utils/dashboard_utils.py
wjayesh: good catch, i've added a check

avishniakov (changes requested):
Smells like a circular dep, based on the CI output

wjayesh:
Hmm, I can't seem to figure out why that's happening. Let me take a deeper look

src/zenml/model/model.py
```diff
@@ -505,7 +505,17 @@ def _root_validator(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+
+        model_version = self._get_or_create_model_version()
+
+        model_version_url = get_model_version_url(model_version.id)
+        if model_version_url:
+            logger.info(f"Model version dashboard URL: {model_version_url}")
+        else:
+            logger.info(
+                "You can now view your Models in the ZenML dashboard using ZenML Cloud."
+            )
```
avishniakov, line 518: Is it indeed the best place to output this? It will be called for every model version used in the pipeline. So if I set something in the pipeline and set something in 10 out of 20 steps this function will be called 11 times for each one of them (assuming they are all different, for sure).

src/zenml/new/pipelines/pipeline.py
```diff
@@ -758,6 +754,10 @@ def _run(
                         "`zenml up`."
                     )
 
+            deploy_pipeline(
+                deployment=deployment_model, stack=stack, placeholder_run=run
+            )
+
```
avishniakov, line 760: Why this reorder is needed?

src/zenml/model/model.py
```diff
@@ -505,7 +505,17 @@ def _root_validator(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
     def _validate_config_in_runtime(self) -> None:
         """Validate that config doesn't conflict with runtime environment."""
-        self._get_or_create_model_version()
+        from zenml.utils.dashboard_utils import get_model_version_url
+
+        model_version = self._get_or_create_model_version()
+
+        model_version_url = get_model_version_url(model_version.id)
+        if model_version_url:
+            logger.info(f"Model version dashboard URL: {model_version_url}")
```
avishniakov, line 514: It is unclear which model are you talking about (see my other comment with the multi-model version setting)

src/zenml/model/model.py
```diff
@@ -505,7 +505,17 @@ def _root_validator(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+        else:
+            logger.info(
+                "You can now view your Models in the ZenML dashboard using ZenML Cloud."
```
avishniakov, line 517: Again for multi-model version use-case, I bet this shall be displayed only once.

avishniakov (changes requested):

src/zenml/new/pipelines/pipeline.py
wjayesh: this is needed so that the URL for the pipeline is displayed before the pipeline run logs start.

src/zenml/model/model.py
wjayesh: good catch, I'll add some ID to the logs

src/zenml/new/pipelines/run_utils.py
```diff
@@ -241,4 +242,11 @@ def _validate_new_version_requests(
                 "that `Model` requesting new version is configured only in one "
                 "place of the pipeline."
             )
-        data.model._validate_config_in_runtime()
+        is_cloud_model = (
```
avishniakov, line 245: I know, that it is always True or False, but just for consistency, if this will get changed sometime later,  I'd use `is_cloud_model &= (`

avishniakov (approved):
All good, one small suggestion.

src/zenml/new/pipelines/run_utils.py
wjayesh: thanks, I agree. will push the change

src/zenml/model/model.py
```diff
@@ -503,9 +503,25 @@ def _root_validator(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+
+        model_version = self._get_or_create_model_version()
+
+        model_version_url = get_model_version_url(model_version.id)
+        if model_version_url:
+            logger.info(
+                f"Dashboard URL for Model Version with name {model_version.name} "
+                f": {model_version_url}"
+            )
+            return True
```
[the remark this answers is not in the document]
wjayesh: I put the logic in there to avoid having to call `_get_or_create_model_version()` twice, as the validate function already makes that call. Maybe I can return the output of that fn (a model version) from the validation function and then call a utiils function to make the cloud server check?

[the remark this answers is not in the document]
schustmi: Yep that sounds good!

schustmi (approved):
LGTM, thanks for the adjustments!

PR #2510, opened 2024-03-11
Allow installing packages using UV

schustmi:
## Describe changes
Allow installing packages using UV when building Docker images.

## Pre-requisites

## Types of changes

### DIFF
--- a/examples/quickstart/README.md
+++ b/examples/quickstart/README.md
@@ -24,7 +24,7 @@ Along the way we will also show you how to:
 
 You can use Google Colab to see ZenML in action, no signup / installation required!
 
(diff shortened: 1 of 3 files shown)


strickvl (approved):
Looks good, but a bigger question would be to ask, what is the reason to not turn this on by default? We're thinking of using `uv` for the `zenml integration install` installer too, and I have the same question there...

schustmi:
Mhm I'm actually not too familiar with `uv` yet, would you say it is as reliable as `pip`?

strickvl:
> Mhm I'm actually not too familiar with `uv` yet, would you say it is as reliable as `pip`?

Actually that's probably a good point. I think not yet is probably the right call, but maybe in 3-5 months yes.

PR #2519, opened 2024-03-13
Add support for LLM template

schustmi:
## Describe changes
Adds support for the new LLM Finetuning template.

## Pre-requisites

## Types of changes


schustmi:
@avishniakov The CI ran (successfully) but didn't actually commit the files for the first time. Is that intended and I need to commit the initial version myself?

avishniakov:
> @avishniakov The CI ran (successfully) but didn't actually commit the files for the first time. Is that intended and I need to commit the initial version myself?

No actually, it should always commit changes anyway

.typos.toml
```diff
@@ -7,6 +7,7 @@ extend-exclude = [
     "tests/unit/materializers/test_built_in_materializer.py",
     "tests/integration/functional/cli/test_pipeline.py",
     "src/zenml/zen_server/dashboard/",
+    "examples/llm_finetuning/lit_gpt/"
```
avishniakov, line 10: Why do you need to exclude the new template from the typos check? Can the typos be fixed on the template side instead?

pyproject.toml
```diff
@@ -303,6 +303,12 @@ exclude = [
     "venv",
     '__init__.py',
     'src/zenml/cli/version.py',
+    # LitGPT files from the LLM Finetuning example
+    'examples/llm_finetuning/evaluate',
+    'examples/llm_finetuning/finetune',
+    'examples/llm_finetuning/generate',
+    'examples/llm_finetuning/lit_gpt',
+    'examples/llm_finetuning/scripts',
```
avishniakov, line 311: Also not very confident here - why is this needed? You have linting in your `copier.yaml`, right? Why this is not aligned with what we have here?

avishniakov (commented):
Left a few questions. Can you also add the image optimizer to the template repo, so optimization is not happening here and will not be causing infinite updates?
https://github.com/zenml-io/template-starter/blob/main/.github/workflows/image-optimizer.yml

.typos.toml
schustmi: The template contains third-party code which I tried not to modify, as we'll have to swap that from time to time to upgrade it.

pyproject.toml
schustmi: Same as above.

schustmi:
@avishniakov Regarding the image optimizer: That's already in the repo, but probably didn't run for this PR yet. What would be the best solution to resolve this? Push the optimized image to the other repo and re-tag?

avishniakov:
> @avishniakov Regarding the image optimizer: That's already in the repo, but probably didn't run for this PR yet. What would be the best solution to resolve this? Push the optimized image to the other repo and re-tag?

It would the easiest way, yeah. Let's do it like this. In this case - this PR can go directly!

schustmi:
It seems like it's actually just the same image 🤔
I'll re-run the job here to see if it still detects changes

PR #2520, opened 2024-03-13
Update `pip check` command to use `uv`

strickvl:
Use uv pip check instead of normal pip check.

(Awaiting next `ruff` release for this command to make its way in).


avishniakov (approved):
LGTM, there is a failure in CI checks - need to be resolved before merging

strickvl:
Yep @avishniakov relates to this issue I created: https://github.com/astral-sh/uv/issues/2520. I'll have to dig into exactly why there are two different pips installed, but seems simple enough to address.

PR #2521, opened 2024-03-13
Add test to install dashboard script

strickvl:
This pull request adds a test to the install dashboard script. The test ensures that the dashboard directories are not ignored by Git.

### DIFF
--- a/scripts/install-dashboard.sh
+++ b/scripts/install-dashboard.sh
@@ -25,6 +25,17 @@ verifySupported() {
   fi
 }
 
+# checkGitIgnore checks if the dashboard directories are ignored by Git
+checkGitIgnore() {
+  if [ -f ".gitignore" ]; then
+    if grep -q -E "(^|\/)dashboard($|\/)" ".gitignore" || grep -q -E "(^|\/)src\/zenml\/zen_server\/dashboard($|\/)" ".gitignore"; then
+      echo "Error: The '/dashboard' or 'src/zenml/zen_server/dashboard' directory is ignored by Git."
+      echo "Please remove the corresponding entries from the .gitignore file to proceed with the installation."
+      exit 1
+    fi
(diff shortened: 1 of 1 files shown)


avishniakov (commented):
Will it fail only on release now?

strickvl:
> Will it fail only on release now?

Yeah I guess that's the only time we run this script. I can add an equivalent in our CI somewhere else?

avishniakov:
> > Will it fail only on release now?
> 
> Yeah I guess that's the only time we run this script. I can add an equivalent in our CI somewhere else?

Yep, it is just too late on the release, since we will need to recreate release branch and all that hassle

.github/workflows/ci-slow.yml
```diff
@@ -112,6 +112,15 @@ jobs:
...
           bash scripts/check-alembic-branches.sh
+      - name: Check .gitignore
+        run: |
+          if [ -f ".gitignore" ]; then
+            if grep -q -E "(^|\/)dashboard($|\/)" ".gitignore" || grep -q -E "(^|\/)src\/zenml\/zen_server\/dashboard($|\/)" ".gitignore"; then
+              echo "Error: The '/dashboard' or 'src/zenml/zen_server/dashboard' directory is ignored by Git."
+              echo "Please remove the corresponding entries from the .gitignore file."
+              exit 1
+            fi
+          fi
```
stefannica, line 123: Or just install the dashboard?

```suggestion
      - name: Install latest dashboard
        run: bash scripts/install-dashboard.sh
```

stefannica (approved):

.github/workflows/ci-slow.yml
avishniakov: +1

avishniakov (approved):

PR #2522, opened 2024-03-13
Fix the pagination in the database backup

stefannica:
## Describe changes
The pagination done during the database backup can inadvertently include the same entry in multiple consecutive pages, because it only uses "created" as a sorting field and this field is not unique. Including the same field multiple times in different pages is a big problem because it will cause primary key constraint integrity failures at restoration time.

To break this tie, the "id" field is also included in the sort criteria.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/migrations/utils.py
+++ b/src/zenml/zen_stores/migrations/utils.py
@@ -236,9 +236,17 @@ class MigrationUtils(BaseModel):
                 # correct order, since some tables have inner foreign key
                 # constraints.
                 if "created" in table.columns:
-                    order_by = table.columns["created"]
+                    order_by = [table.columns["created"]]
                 else:
-                    order_by = None
+                    order_by = []
+                if "id" in table.columns:
+                    # If the table has an `id` column, we also use it to sort
+                    # the rows in the table, even if we already use "created"
+                    # to sort the rows. We need a unique field to sort the rows,
+                    # to break the tie between rows with the same "created"
+                    # date, otherwise the same entry might end up multiple times
+                    # in subsequent pages.
+                    order_by.append(table.columns["id"])
 
                 # Fetch the number of rows in the table
                 row_count = conn.scalar(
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/migrations/utils.py
```diff
@@ -236,9 +236,17 @@ def backup_database_to_storage(
                 # correct order, since some tables have inner foreign key
                 # constraints.
                 if "created" in table.columns:
-                    order_by = table.columns["created"]
+                    order_by = [table.columns["created"]]
                 else:
-                    order_by = None
+                    order_by = []
+                if "id" in table.columns:
```
AlexejPenner, line 242: do we currently have any entitiy without id? In that case what would we use as tiebreak?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -870,9 +870,14 @@ def filter_and_paginate(
...
-            query = query.order_by(desc(getattr(table, column)))
+            sort_clause = desc(getattr(table, column))
         else:
-            query = query.order_by(asc(getattr(table, column)))
+            sort_clause = asc(getattr(table, column))
+
+        # We always add the `id` column as a tiebreaker to ensure a stable,
+        # repeatable order of items, otherwise subsequent pages might contain
+        # the same items.
+        query = query.order_by(sort_clause, asc(table.id))
```
AlexejPenner, line 880: are we sure here that every entity in the db has an id?

src/zenml/zen_stores/migrations/utils.py
stefannica: All entities that have a `create` field also have an `id` field. This is guaranteed by the fact that they both come from the `BaseSchema` class.

As for entities that don't have either, we don't need a tiebreak. The implicit ordering supported by the storage engine (e.g. insert order or primary key order) is used. This implicit ordering does not create duplicates.

src/zenml/zen_stores/sql_zen_store.py
stefannica: Not every entity in the DB has an ID, no. However, all entities supported by the generic filter and pagination feature inherit from `BaseSchema` and they all have an ID column.

avishniakov (approved):
🦭

wjayesh (approved):
Thanks for figuring this out! LGTM 🐬

PR #2523, opened 2024-03-13
Fix bug in HyperAI orchestrator depends_on parallelism

christianversloot:
## Describe changes
There is a newly discovered issue with the HyperAI orchestrator which leads to unexpected behavior when a step depends on multiple previous steps. Upon investigation a bug was identified as currently the `depends_on` dict that eventually leads to the Docker Compose file is overwritten for every step a downstream one is dependent on. This PR fixes the issue by first initializing the dict to an empty one if it discovers that a step is dependent on others; then adds them one by one by updating it. I have just tested locally with the following Compose file, and multi-container dependency works nicely, so the fix also works Docker side.

```
version: '3.8'
services:
  hello-world:
    image: busybox
    command: /bin/sh -c "sleep 30"
  hello-world2:
    image: hello-world
  hello-world3:
    image: hello-world
    depends_on:
      hello-world:
        condition: service_completed_successfully
      hello-world2:
        condition: service_completed_successfully
```

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
+++ b/src/zenml/integrations/hyperai/orchestrators/hyperai_orchestrator.py
@@ -230,17 +230,23 @@ class HyperAIOrchestrator(ContainerizedOrchestrator):
 
             # Add dependency on upstream steps if applicable
             upstream_steps = step.spec.upstream_steps
-            for upstream_step_name in upstream_steps:
-                upstream_container_name = (
-                    f"{deployment_id}-{upstream_step_name}"
-                )
+
+            if len(upstream_steps) > 0:
                 compose_definition["services"][container_name][
                     "depends_on"
-                ] = {
-                    upstream_container_name: {
-                        "condition": "service_completed_successfully"
-                    }
-                }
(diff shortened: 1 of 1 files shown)


avishniakov (approved):

schustmi (approved):

strickvl:
@christianversloot could you run the formatting script? Otherwise this looks good to go.

christianversloot:
@strickvl absolutely, had already done so, but realized it used an older `ruff` version. Noticed in the pipeline that `ruff==0.3.2` is used, so I've now formatted with that version. If you could launch the pipeline, that would be great.

strickvl:
Yeah we now just use whatever latest and greatest `ruff` is, same with `uv`, coupled with more specific opt-in rule sets so we don't get surprised by whatever newness they add support for.

christianversloot:
@strickvl in the pipelines, when using `uv`, what are the performance gains you're seeing?

christianversloot:
@strickvl Interesting! If I don't forget I will try it as soon as the new version is published :)

PR #2529, opened 2024-03-14
Fix CI (`uv` bug)

strickvl:
Addresses this https://github.com/astral-sh/uv/issues/2450. Will revert + allow higher versions once the fix is released in 0.1.21.


avishniakov (approved):
🦭

PR #2533, opened 2024-03-16
Fix GCP step logging

adtygan:
## Describe changes
The issue arises because GCS artifacts are immutable [Stack Overflow thread](https://stackoverflow.com/questions/52715217/how-to-open-a-file-in-a-gcs-bucket-in-append-mode-using-python). To fix the issue, I suggest a fix based on @strickvl 's idea mentioned below

### [Alex Strick van Linschoten](https://github.com/strickvl)'s Idea

Consider using the logging.StreamHandler facility to temporarily write logs to the remote file (GCS, S3, etc.). Here's an example:

```
import logging
import fsspec

f = fsspec.open("gs://<<my_gcs_bucket>>/test_log.log", "w")
with f as of:
    log_handler = logging.StreamHandler(of)
    logger = logging.getLogger()  # Root logger
    logger.addHandler(log_handler)
    for i in range(0, 5000):
        logger.warning(f"I'm log line #{i}")
    logger.removeHandler(log_handler)
```

This approach could fit nicely in the StepLogsStorageContext class.

### My Solution

Set buffer size to `sys.maxsize` and copy the existing contents of the log file to the buffer. This prevents the triggering of the part of code that appends buffer contents to log file once buffer is full and flushes buffer. ([reference](https://github.com/adtygan/zenml/blob/34d179ca47be2927e1e0af77f5dd25d4971e186a/src/zenml/logging/step_logging.py#L126))

### Code to test the change:

>Please use GCS stack

```
import gcsfs
from zenml.logging.step_logging import StepLogsStorage, StepLogsStorageContext

client = Client()
_ = client.active_stack

TEST_FILE="gs://<<my_gcs_bucket>>/test.txt"

with StepLogsStorageContext(logs_uri=TEST_FILE) as log_storage:
    for i in range(0,10):
        print(f"I'm log line #{i}")

fs = gcsfs.GCSFileSystem()

with fs.open(TEST_FILE, 'r') as f:
    all_of_it = f.read()

print('Final Contents:')
print(all_of_it)
```

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the initialization process of `StepLogs` to support a maximum message limit and improved handling of log file reading errors.


strickvl:
@coderabbitai review

strickvl:
@adtygan there's a linting error that needs fixing

adtygan:
@strickvl , I ran the below commands before I made the PR

```
zenml integration install gcp 
pip install click~=8.0.3
mypy --install-types
```

After that I ran 

```
bash scripts/format.sh
bash scripts/run-ci-checks.sh >> .local/run-ci-checks.log
```

For `format.sh`, I got the following output

Screenshot 2024-03-17 at 12 21 15 AM

For `run-ci-checks.sh`, I got the following output in the terminal

Screenshot 2024-03-17 at 12 23 46 AM

I am attaching the log file for `run-ci-checks.sh`. I see that the lint script ran without errors. I am not sure what I am doing wrong. Kindly request your assistance. 

[run-ci-checks.log](https://github.com/zenml-io/zenml/files/14624513/run-ci-checks.log)

strickvl:
https://github.com/zenml-io/zenml/actions/runs/8307099335/job/22735937951?pr=2533#step:5:27 this is the specific thing you need to fix first of all.

adtygan:
In the commit I had made I do see the import at Line 25
SHA is: `34d179ca47be2927e1e0af77f5dd25d4971e186a`
Screenshot 2024-03-17 at 12 44 43 AM


But after this, there was another commit made by you, and in this version I cannot see the import anymore
SHA is: `64a1e745e4b38efc5ff4ce9c6738e95199f4dd50`

Screenshot 2024-03-17 at 12 47 42 AM


I am not sure why this is happening.

src/zenml/logging/step_logging.py
```diff
@@ -170,7 +170,14 @@ def __init__(self, logs_uri: str) -> None:
         Args:
             logs_uri: the URI of the logs file.
         """
-        self.storage = StepLogsStorage(logs_uri=logs_uri)
+        self.storage = StepLogsStorage(
+            logs_uri=logs_uri, max_messages=sys.maxsize
+        )
+        try:
+            with fileio.open(logs_uri, "r") as file:
```
strickvl, line 177: @adtygan this is the line the failing linting is talking about...

adtygan: As far as I understand, L177 is using `fileio` which has not been imported, and this is raising the error. Please correct me if I'm wrong.

strickvl:
Yes that's exactly it.

src/zenml/logging/step_logging.py
adtygan: I have added the import now.

adtygan: I ran the command `bash scripts/lint.sh --fix` locally. In `step_logging.py` it swapped the order of imports, and seems to be fixing the Lint Check at https://github.com/zenml-io/zenml/actions/runs/8347170346/job/22846195340. But running this command has also changed a lot of other files. Should I still push the change to this branch?

strickvl: The script to run is `bash scripts/format.sh` to format the files. Not sure where the `--fix` came from, but you should undo any changes made by `bash scripts/lint.sh --fix` and run `bash scripts/format.sh` instead before repushing/committing.

adtygan: I ran the `format.sh` instead and it still made a lot of changes in other files. I pushed the commit and it has added a lot of commits here. I'm not sure if this is intended. It's my first time working on open-source and I apologise for any inconvenience.

strickvl:
It seems like maybe you're using an old version of the `ruff` formatter @adtygan? Can you `pip install "zenml[server,dev]"` again in your environment to make sure you're using the latest version, then rerun the script and push the changes? We def shouldn't be seeing changes to all these files in the PR.

adtygan:
I figured out that I was working on a fork of the repo so my `develop` branch was out of date. I pulled the updates from upstream and then pushed my changes. Please review.

strickvl:
There are quite a few updated files in here that shouldn't need to be updated. Not sure what happened on your side or with the fork, but since the changes are small, it might be worth resetting this branch somehow so that only the changes you made are reflected in the changed files?

adtygan:
I only made a change to the `logging/step_logging.py` file. But when I run the `format.sh` script it modifies a lot of other files. Can I run the script on only the login file and push a commit? I think that should fix the issue.

strickvl:
You could try that, but the point is that the other previously-modified
files will still have been previously committed. have you made sure to
installed / upgraded the latest version of the dev tools? Probably if you
do `pip install -U ruff` that would be enough, since that's the package
that's modifying / formatting files...

adtygan:
I upgraded `zenml` but not `ruff`. I will do this also and then try once more like you mentioned.

strickvl:
Yeah the updates to the important formatting packages will only come when you update `zenml[dev]` packages, not just core zenml @adtygan

adtygan:
I had already run `pip install "zenml[server,dev]"` the last time. But this was using `ruff 0.1.7`. I ran `pip install -U ruff` after this like you suggested and it upgraded to version `0.3.4`. When I ran `format.sh` now, it doesn't raise any errors. So I have now pushed my changes.

adtygan:
May I please know if there is any action required from my end to solve this issue?

strickvl:
@bcdurak is looking into it / at it

bcdurak (changes requested):
Hey @adtygan, sorry about the late response regarding this issue. 

I have checked the original problem and the implementation in the PR again and there are a few points that I wanted to mention here:

- Currently, the logic around how ZenML stores the log messages works as follows:
	- Whenever we start the execution of a step, we create the proper context manager (`StepLogsStorageContext`) based on the properties of this step and a unique identifier. This process is conducted only once per step execution.
	- During the execution, ZenML queues up the incoming log messages in a buffer. If the buffer reaches a certain size or a certain amount of time has passed, we empty it by saving the messages in the designated location using the active artifact store.

The problem arises due to the immutability of the designated file in GCS as you mentioned above. While this PR fixes the problem to some degree, unfortunately, this does not propose a full solution. 

- If I am not missing anything, in the current version of the implementation, it is redundant to read any existing log messages (and add them to the buffer) as it is impossible to for ZenML create a context manager which points to an already existing log file. As I have mentioned, this process is done once per step execution by using a unique identifier.

- The current solution works because it removes the first threshold of the buffer by setting the `max_messages` to a high enough value. However, you would run into the same issue if the buffer is cleared not due to the size but due to the time limit. For instance, if you add a `time.sleep(20)` in between two log messages, you will run into the same issue.

- Finally, the current changes impact the functionality of the entire implementation and deem the `max_messages` parameter useless unless it is explicitly defined. I would perhaps propose a solution where a patch is applied only if the current artifact store is a GCP artifact store.

Thank you once again for working on this issue. The logs storage is an important topic and we are actively looking for different ways to improve our implementation, so your contribution is much appreciated :)

adtygan:
Thank you for your review @bcdurak . I agree with the points you have mentioned above. If I am allowed to propose a GCP specific solution, I have 2 approaches in mind.
1. Patch the time limit as well so that my current solution can work.
2. GCP, if I'm not wrong, allows some other way to create mutable files. But as far as I know, it won't be a GCS object. I will look into using these tools to solve the problem.

Which of the 2 routes do you suggest?

Thanks

strickvl:
linked to #2366 (closed in favor of this one)

bcdurak:
Hey @adtygan,

I think the second approach would be better here if possible. The first one is still viable but I would only apply it if the active artifact store is a GCS artifact store. 

As I mentioned before, logging has been an integral part of ZenML and we will be working on improving the user experience when it comes to handling these log messages. @safoinme is currently taking a look at our implementation. He can also share any insights and recommendations he comes across.

adtygan:
Thank you for your inputs @bcdurak . I will see how I can implement option 2. @safoinme could you please share your insights?

avishniakov:
GCP issue shall be closed with https://github.com/zenml-io/zenml/pull/2755

htahir1:
Closed with https://github.com/zenml-io/zenml/pull/2755

PR #2535, opened 2024-03-18
Bug fix HyperAI orchestrator: Offload scheduled pipeline execution to bash script

christianversloot:
## Describe changes
I've discovered a new issue with the HyperAI orchestrator in case of longer-running scheduled pipelines. Recall that pipeline scheduling with the HyperAI orchestrator means that a Docker Compose file is created in combination with a crontab entry, which is responsible for kicking off the Docker Compose `up` command and ensuring that the pipeline completes.

Except that it doesn't. It does create all containers; the dependent ones nicely wait for the others to be completed, but the pipeline never finishes. After the completion of some steps (i.e. typically after 5-10 minutes have passed) the pipeline just hangs and the other containers remain in `Created` state. I've verified that this is not an issue with Docker Compose itself, because a nonscheduled variant of the pipeline completes without issues.

Then, I've found that when offloading the actual command to a `bash` script, the pipeline completes successfully even when scheduled (I've had 10 pipelines complete with this test setups over the weekend, 100% success compared to 0% before). In other words, it seems to be the case that for some reason the `up` command breaks when the cron is executed if it's executed directly. So, this PR ensures that my test setup is moved to the orchestrator:

* The pipeline execution command for scheduled pipelines is moved to a separate bash script that will be scp'd into the same folder as the Docker Compose file.
* The crontab entry kicks off the bash script instead of the actual command.

Tested locally -- and looked good (equal to how I manually configured the crontab).

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
	- Added functionality to securely transfer files to a HyperAI instance, enhancing pipeline preparation and execution.


strickvl:
@coderabbitai review

strickvl (approved):

PR #2545, opened 2024-03-20
Allow loading of artifacts without needing to activate the artifact store (again)

avishniakov:
## Describe changes
I fixed the logic of reading artifacts from ZenML to achieve the following snippet to work as usual, without the need to activate the stack before that:
```python

artifact = Client().get_artifact_version('some-uuid')
loaded_artifact = artifact.load()
```

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/materializers/service_materializer.py
+++ b/src/zenml/materializers/service_materializer.py
@@ -17,6 +17,7 @@ import os
 import uuid
 from typing import TYPE_CHECKING, Any, ClassVar, Dict, Tuple, Type
 
+from zenml.client import Client
 from zenml.enums import ArtifactType
 from zenml.materializers.base_materializer import BaseMaterializer
 from zenml.services.service import BaseDeploymentService, BaseService
@@ -49,8 +50,7 @@ class ServiceMaterializer(BaseMaterializer):
         with self.artifact_store.open(filepath, "r") as f:
             service_id = f.read().strip()
 
-        client = Client()
-        service = client.get_service(name_id_or_prefix=uuid.UUID(service_id))
+        service = Client().get_service(name_id_or_prefix=uuid.UUID(service_id))
         return BaseDeploymentService.from_model(service)
 
     def save(self, service: BaseService) -> None:


strickvl (approved):

schustmi (commented):
Haven't looked at it in detail yet, just a more general question: What's the technical reason why this is not possible with `fileio` anymore and we have to use the artifact store methods in the materializers?

avishniakov:
> Haven't looked at it in detail yet, just a more general question: What's the technical reason why this is not possible with `fileio` anymore and we have to use the artifact store methods in the materializers?

There is one main reason behind it: `fileio` works smoothly (meaning in a protected fashion, according to previous PRs) for all FS types, except local, while local is used in many other places outside of materializers' story, so to keep writing of artifacts and logs in scope-aware mode without the need to care about the underlying FS - I moved them to go via Artifact Stores. This PR is not changing that (it was already done and released) it just adds the possibility to set an artifact store explicitly in the materializer class, so we can pull smoothly without the need to switch stacks.

schustmi:
I see, with that in mind the changes in this PR look good to me.
However I still don't really get why we made those changes to use the artifact stores in the first place, doesn't `fileio` contains some useful methods built on top of the artifact store interface (e.g. copying from one store to another) that are very useful inside the materializers?

avishniakov:
> I see, with that in mind the changes in this PR look good to me. However I still don't really get why we made those changes to use the artifact stores in the first place, doesn't `fileio` contains some useful methods built on top of the artifact store interface (e.g. copying from one store to another) that are very useful inside the materializers?

This might be indeed the case, that usage of some `fileio` methods might come in handy on top of the artifact store methods, but the PR merged previously was aiming to close the gap with the usage of unsafe paths (e.g. outside of artifact store bounds) since it was reported as a bug.

I cannot think of a scenario, where a materializer has to work with some data from outside of the configured artifact store. My reasoning: materializer is used only in steps (using them directly is possible, but is not what is officially supported) and inside a step, we always have a stack set, so we never face multi-artifact store use-case, as of now. If going forward, we will support different artifact stores for different steps inside one pipeline, my assumption still stands, since in one step we only deal with one AS. Does it make any sense or I missed something important? @schustmi

safoinme (approved):

schustmi (approved):

avishniakov:
> The scenario I was thinking of was actually copying local directories <-> artifact store.

I see, so `fileio.copy` still covers this case, but the previous PR I mentioned is protecting a bit of our standard materialized from malicious use. I understand that cannot fully get rid of the cases when we go beyond artifact store bounds and this can be kept as is in other materialized with more complex logic.

I will merge this one now and we can create a follow-up for that if you find this needed.

What can be actually done for the use case you described is also achievable with the artifact store interface, but with a bit more sugar around:
```python
with self.artifact_store.open("s3://some_path/05b56688-931a-44f8-8777-035b046e191f.log","rb") as src:
     with open("local-file.log","wb") as tgt:
         tgt.write(src.read())
```

PR #2568, opened 2024-03-28
Add VS Code extension to README / docs

strickvl:
Update docs with information about the VS Code extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@
     🎉 Version 0.56.2 is out. Check out the release notes
     <a href="https://github.com/zenml-io/zenml/releases">here</a>.
     <br />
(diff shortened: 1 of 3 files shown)


README.md
htahir1, line 1: Might make sense to add a small section even with a GIF hidden behind a accordion arrow?

docs/book/user-guide/production-guide/understand-stacks.md
htahir1, line 1: I like where this is but might make sense to add it the pipelines page as well at the start of the starter guide?

htahir1 (changes requested):

htahir1 (approved):

PR #2571, opened 2024-03-28
Catch missing jupyter installation for `zenml go`

strickvl:
I added a check to catch when someone runs `zenml go` and they don't have Jupyter notebooks installed.

### DIFF
--- a/src/zenml/cli/base.py
+++ b/src/zenml/cli/base.py
@@ -28,7 +28,13 @@ from zenml.analytics.enums import AnalyticsEvent
 from zenml.analytics.utils import email_opt_int, track_handler
 from zenml.cli import utils as cli_utils
 from zenml.cli.cli import cli
-from zenml.cli.utils import confirmation, declare, error, warning
+from zenml.cli.utils import (
+    confirmation,
+    declare,
+    error,
+    is_jupyter_installed,
+    warning,
+)
 from zenml.client import Client
 from zenml.config.global_config import GlobalConfiguration
 from zenml.console import console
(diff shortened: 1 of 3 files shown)


strickvl:
@coderabbitai review

src/zenml/cli/utils.py
```diff
@@ -2649,3 +2667,17 @@ def verify_mlstacks_prerequisites_installation() -> None:
...
+
+
+def is_jupyter_installed() -> bool:
+    """Checks if Jupyter notebook is installed.
+
+    Returns:
+        bool: True if Jupyter notebook is installed, False otherwise.
+    """
+    try:
+        import notebook
```
bcdurak, line 2679: Just a small nitpick here.

While this is much simpler, it might be more performant and safer to use the same functionality that we used for [our integrations to check](https://github.com/zenml-io/zenml/blob/cb57213189ae8a30591fdcfbd294d77201dbd856/src/zenml/integrations/integration.py#L64) whether the `notebook` package is installed, instead of importing the whole package.

Again, for this use case, we can still leave this as is.

bcdurak (approved):
Looks good.

A few small notes (feel free to ignore):

- Do you think that the version of the `notebook` package could change anything here?
- I like how you added a comment to the README, it might be nice to do the same to the docs if we have it anywhere.
- Just out of curiosity, did you do the formatting manually?

strickvl:
> * Do you think that the version of the `notebook` package could change anything here?

No I think jupyter takes pretty good care with backwards compatibility. I don't think we need to specify a specific version.

> * I like how you added a comment to the README, it might be nice to do the same to the docs if we have it anywhere.

I'll take a look to see where is appropriate.
> * Just out of curiosity, did you do the formatting manually?

Not sure what you mean here.

bcdurak (approved):

src/zenml/cli/base.py
```diff
@@ -475,7 +482,24 @@ def go() -> None:
...
+        try:
+            subprocess.check_call(
+                ["jupyter", "notebook"], cwd=zenml_tutorial_path
+            )
+        except subprocess.CalledProcessError as e:
+            cli_utils.error(
+                "An error occurred while launching Jupyter Notebook. "
+                "Please make sure Jupyter is properly installed and try again."
+            )
+            raise e
```
avishniakov, line 496: I think this code is never reached, since `error` does this inside: `raise click.ClickException(message=click.style(text, fg="red", bold=True))`

avishniakov (approved):
Tiny nit

src/zenml/cli/base.py
strickvl: Correct, but if I leave it out then mypy complains. and same with the weird `e` in the docstring. If I leave that out, then `darglint` complains.

PR #2574, opened 2024-03-30
`enable_cache` option shouldn't be set to `False` for one of the steps

ruvilonix:
## Describe changes
I changed `training_pipeline.with_options(enable_cache=False)()` to `training_pipeline()` under where it says `logger.info("\n\nFirst step cached, second not due to settings")`.

## Pre-requisites
- [X ] I have read the **CONTRIBUTING.md** document.

## Types of changes

- [X ] Other (add details above)
Documentation error


strickvl:
Thanks! + good catch!

strickvl (approved):

docs/book/user-guide/starter-guide/cache-previous-executions.md
```diff
@@ -166,7 +166,7 @@ if __name__ == "__main__":
     # This will disable cache for all steps
```
bcdurak, line 166: Since you spotted the mistake here, I took another look at the entire example and I already see a few more lines we can fix here:

- The comment on this line is incorrect. The following code only disables the cache for the `svc_trainer`.
- Additionally, there is another comment a few lines above this one which is also misleading. **There is no `enable_cache=False` in the `@step` decorator.** The statement `ZenML will detect a different value for the
gamma input of the second step and disable caching for that step.` is correct though.

bcdurak (changes requested):
Good catch! I left a comment with a few more changes.

bcdurak (approved):

bcdurak:
Thanks @ruvilonix for spotting the mistake and fixing it.

PR #2572, opened 2024-03-29
Add `start_time` support to HyperAI orchestrator scheduled pipelines

christianversloot:
## Describe changes
Currently, the HyperAI orchestrator supports scheduled pipelines, but only via crontab entries - meaning that pipeline runs are started in intervals.

This PR adds support for `Schedule.start_time` to the HyperAI orchestrator, as we've found use cases for running pipelines just once (such as scheduling them to start during the night, when other peoples' GPU workloads have finished). It uses the `at` Linux command line utility that can be installed in HyperAI instances (and reflects this in the docs). 

I still need to test this locally but already made the PR so that the review process and tests can start.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [x] If my change requires a change to docs, I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] I have based my new branch on `develop` and the open PR is targeting `develop`. If your branch wasn't based on develop read [Contribution guide on rebasing branch to develop](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md#-pull-requests-rebase-your-branch-on-develop).
- [x] If my changes require changes to the dashboard, these changes are communicated/requested.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

## Summary by CodeRabbit

- **New Features**
	- Enhanced the HyperAI orchestrator to support more flexible scheduling of pipeline runs. Users can now schedule pipelines using either cron expressions or specific start times, providing greater control over their pipeline executions.

PR #2577, opened 2024-04-02
Upgrade ruff and yamlfix to latest versions before running formatting

christianversloot:
## Describe changes
When working on https://github.com/zenml-io/zenml/pull/2572 I got a large diff when running `scripts/format.sh` because my version of `ruff` installed in the environment was quite old.

This PR implements a suggested approach given how ZenML tests are performed these days (at least that's what I learned from @strickvl) - which is to automatically attempt `pip install --upgrade` for both `ruff` and `yamlfix` in `scripts/format.sh`, unless the user explicitly chooses not to.

Feel free to decline this PR if you don't think this should be in there, but at least it makes testing a bit more convenient when developing.

## Pre-requisites

## Types of changes


christianversloot:
`uv` sounds like a good idea to me. I can also see your point about doing it the other way around; however, I chose to do it this way because forgetting about the upgrades can then never happen. I'm looking forward to the others' opinions and I'll make the necessary changes.

avishniakov:
Thanks for contributing @christianversloot !

My take on this:
- I run `format.sh` as on save function in my VSCode, so running `pip install` on every file save isn't very nice to me
- On the other hand - it is still fast enough, maybe a couple of seconds to run
- What about just switching it to `uv` here without dev deps? Dev deps can go as some separate flag, but I don't see big benefits of that right away, tbh. @strickvl up to you to advise, cause my experiments with running `uv pip install --upgrade ruff yamlfix` right away ended up in acquiring pydantic v2 immediately 🙁

christianversloot:
@avishniakov does your VSCode setup allow for flags? If so, you could set things up e.g. as `bash format.sh --no-pip-install` and it should work as it currently does. 

Also, do we want to assume that `uv` is installed by the user? Or do we want to install it within the script (taking a few extra seconds)? My intuition says the first, and maybe reflect this as a comment in the script e.g. `# Assumes uv is installed, if not run pip install uv`.

christianversloot:
Made some preliminary changes based on the feedback :)

avishniakov (approved):
🦭

scripts/format.sh
```diff
@@ -26,6 +29,12 @@ do
     fi
 done
 
+# Automatically upgrade packages to newest versions
+# Assumes uv is installed, if not run pip install uv
+if [ "$SKIP_UPGRADE" = false ]; then
+    uv pip install --no-deps --upgrade ruff yamlfix
```
avishniakov, line 35:
```suggestion
    pip install uv
    uv pip install --no-deps --upgrade ruff yamlfix
```

avishniakov (changes requested):

scripts/format.sh
christianversloot: @avishniakov in that case, we can drop line 33, can't we?

avishniakov: Yep, sure thing!
The reason why I think it would be better to add this install line is the fact that `uv` might be not so popular around yet, so people can get confused about why it is needed and what it is actually. Having the disabler flag at hand and not so drastic a runtime impact, I'd prefer to make it easier to use, even if it is a bit slower. Normally people are not running `format.sh` on every save, like I do, I guess 🙂

christianversloot: Good point. Let me commit your suggestion then drop the line with an additional commit :)

christianversloot:
@avishniakov @strickvl both commits have been pushed, if you'd like the tests can be run again.

avishniakov (approved):
Thanks for your work 🙂

christianversloot:
@bcdurak I like the suggestion. Some questions in return before I'll adapt

* Is it guaranteed that `zenml[dev]` contains the dependencies that we need here, being `yamlfix` and `ruff`?
* What do you think about adding an `--upgrade` flag to the `uv pip install` line? Reason why is that if `zenml[dev]` is already installed, one may still end up with an outdated ruff version this way.
* To be 100% sure, I think the `if` statement should check that `installed_version` is not empty, not `install_version`, shouldn't it? Or maybe I'm missing something :)

bcdurak:
@christianversloot Those are some good questions. If I go through them one by one:

- Yes. Both `ruff` and `yamlfix` are included [in our dev dependencies](https://github.com/zenml-io/zenml/blob/a96754a35770b89143c6238588b37cb77d583654/pyproject.toml#L132). This is also how we utilize them in our internal development.
- This is the one question I am not so sure about. 
    - On the one hand, I like the idea because it keeps everything fresh and updated without leaving the boundaries of the ZenML installation. 
    - However, on the other hand, it might upgrade some packages that were originally required by ZenML, which might eventually change the environment people are developing in. Also, in such cases, it would take a few more seconds to run. 
    - Adding it might actually be a blessing in disguise because a change in the environment would only occur if a new version of a dependency is released. In such cases, when you are developing, it is a nice advantage to be in the most up-to-date state as possible.
    - Right now, I am more in favor of adding it but maybe we can ask @avishniakov's opinion on this as well.
- Lastly, you are right, that's a typo from my side when I was sketching it. Sorry about the confusion :)

christianversloot:
Thanks for your response @bcdurak. I'll await @avishniakov's as well before making changes that turn out fruitless.

christianversloot:
Thanks @avishniakov! We could choose to keep things as they are atm and then make the more complex changes should they be necessary in the future? WDYT @strickvl @bcdurak?

strickvl:
Agreed

bcdurak:
@avishniakov @strickvl The more I think about this, I favor my previous suggestion a bit more.

- For your first point, I would argue that it is more beneficial for anyone using a "development" version of ZenML that they are as close to the most up-to-date environment as possible. Otherwise, you would be developing in a different environment than people who just did "pip install zenml". This forces the developer's hand to stay up to date with the requirements of the current ZenML package. Once it is set up, it basically takes the same amount of time. There is also an option to opt out of it for any developer who wants to disable the feature.

- Secondly, you can not keep it as it is and remove the `--no-deps`, right?. This is simply the same as doing:

	```bash
	pip install "zenml[dev]"
	pip install uv 
	uv pip install --upgrade ruff yamlfix
	```

	which ends up installing pydantic>2.0.

- As for the last option, I initially thought about this as well. However, due to our other dependencies, a developer might end up with a `ruff` and `yamlfix` version which is not the max version when they do `pip install "zenml[dev]"`. This might end up breaking some dependencies.

christianversloot:
Since what looked like a simple change has eaten more time than anticipated, I unfortunately need to withdraw my contribution to this PR, as for me, this is a nice to have. However, do feel free to proceed on your own and if not feel free to close the PR.

scripts/format.sh
```diff
@@ -26,6 +29,12 @@ do
     fi
 done
 
+# Automatically upgrade packages to newest versions
+if [ "$SKIP_UPGRADE" = false ]; then
+    pip install uv
+    uv pip install --no-deps --upgrade ruff yamlfix
+fi
```
avishniakov, line 36:
```suggestion
# Check for ruff and yamlfix versions
if [ "$SKIP_UPGRADE" = false ]; then
    RED='\033[0;31m'
    NC='\033[0m'
    
    pip install uv -q
    uv_output=$(uv pip install ".[dev]" --dry-run --upgrade 2>&1 | grep "+")
    ruff_version_change=$(echo "$uv_output" | grep "+ ruff")
    yamlfix_version_change=$(echo "$uv_output" | grep "+ yamlfix")
    
    if [ -n "$ruff_version_change" ]; then
        echo "${RED}ruff version is outdated and might lead to CI failures. Consider upgrading to ${ruff_version_change:3}.${NC}"
    fi
    if [ -n "$yamlfix_version_change" ]; then
        echo "${RED}yamlfix version is outdated and might lead to CI failures. Consider upgrading to ${yamlfix_version_change:3}.${NC}"
    fi
fi
```

bcdurak:
Hey @christianversloot,

@avishniakov and I just had a discussion about this and ended up implementing a new solution. 

In short, it will give you a warning if your current `ruff` and `yamlfix` installations do not match the requirements of the ZenML dev package. 

Thank you for your contribution on this, it will be very helpful going forward.

christianversloot:
Thanks @bcdurak @avishniakov !

bcdurak (approved):
🦭

bcdurak:
Hey @jlopezpena, that's a valid concern. 

However, this PR only changes the formatting script and thus, does not affect the users but the developers IMO. I would argue that it is more important for the developers to stay up-to-date with the latest usable versions of `ruff` and `yamlfix` in order to conduct a proper merge. It is also possible to disable this behavior by setting the `SKIP_UPGRADE` to True.

PR #2608, opened 2024-04-11
Add Python 3.12 support

strickvl:
This pull request adds support for Python 3.12 by updating the matrix of Python versions in the CI workflows.


strickvl:
@jlopezpena yeah the list of new dependencies that become available as listed above are going to be the thing that will cause the issues.

PR #2609, opened 2024-04-11
Add support for `uv` package installer for `zenml integration ...` commands

strickvl:
This pull request adds support for the uv package installer as an alternative to pip. It includes the necessary changes to the `install_packages` and `uninstall_package` functions to handle package installation and uninstallation with uv. Additionally, a new function `is_uv_installed` is added to check if uv is installed in the current environment.

Docs updates are included as well as updates with `--uv` flags for the CI commands.


bcdurak (approved):
Changes look good! 

❤️ uv

src/zenml/cli/utils.py
```diff
@@ -1021,20 +1025,27 @@ def install_packages(
...
-        ] + packages
+        command = (
+            [
+                sys.executable,
+                "-m",
+            ]
+            + pip_command
+            + [
+                "install",
+                "--upgrade",
```
avishniakov, line 1038: Feels like we need `--system` for `uv`

avishniakov (changes requested):

src/zenml/cli/utils.py
avishniakov: Just in case, I get right, that if venv is set, but `--system` is passed we still install into venv?

avishniakov: No, it might be good as is, but then I have another request: can you add a catcher for the cases when Venv is not there and print properly?

strickvl: Ok yes that's a good point. I'm pretty sure that `uv` will spit out its own error, but I can catch it and make it a nice error.

src/zenml/cli/utils.py
```diff
@@ -1051,23 +1062,56 @@ def install_packages(
...
+            [
+                sys.executable,
+                "-m",
+            ]
+            + pip_command
+            + [
+                "uninstall",
+                quiet_flag,
+                "-y",
+                package,
```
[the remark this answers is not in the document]
strickvl: Doesn't apply here.

src/zenml/cli/utils.py
strickvl: Done

avishniakov (approved):

PR #2610, opened 2024-04-12
Mixpanel grouping improvements

schustmi:
## Describe changes
This PR contains two improvements for mixpanel groups:
- Use `server_id` instead of `group_id` in the segment group call
- Enable customizing the `server_id` when deploying a server so we can set it to the tenant ID when deploying cloud tenants

## Pre-requisites

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1377,11 +1378,10 @@ def migrate_database(self) -> None:
                     SQLModel.metadata.create_all  # type: ignore[arg-type]
                 )
             with Session(self.engine) as session:
-                session.add(
-                    IdentitySchema(
-                        id=str(GlobalConfiguration().user_id).replace("-", "")
-                    )
+                deployment_id = os.getenv(
```
schustmi, line 1381: @stefannica I'm unsure whether this is the right place to use this environment variable. When implemented like this, the environment variable will only be respected when deploying the server for the first time. We could use it to have precedence over the DB entry by checking for the env variable existence in `SQLZenStore.get_deployment_id()`. What do you think?

src/zenml/analytics/context.py
```diff
@@ -211,15 +211,10 @@ def group(
         """
         success = False
         if self.analytics_opt_in and self.user_id is not None:
-            if traits is None:
-                traits = {}
-
-            traits.update({"group_id": group_id})
```
bcdurak, line 217: Just out of curiosity, why did we remove this here?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1377,11 +1378,10 @@ def migrate_database(self) -> None:
...
+                    ENV_ZENML_DEPLOYMENT_ID, GlobalConfiguration().user_id
```
bcdurak, line 1382: When are we setting this ID?

schustmi: When creating cloud tenants to have their ID align with the tenant ID

src/zenml/analytics/context.py
schustmi: This was (I guess) intended for mixpanel at some point. However it never worked, as that `group_id` keyword was never included in the `track` events. This is why I switched this to `server_id` (added as a trait when calling this function) as we also send the `server_id` with all our track events.

src/zenml/zen_stores/sql_zen_store.py
stefannica: I suggest not adding another yet environment variable for this, it just complicates this even more. We already have `ZENML_USER_ID` that can be used to override the `GlobalConfiguration().user_id`.

If I understand correctly, you're just trying to override the server identity as reported by the server (`server_info.id`) with the tenant ID when the server is deployed as a cloud tenant:

* use the `ZENML_SERVER_EXTERNAL_SERVER_ID` environment variable (or call `server_config().external_server_id` if set and if running on server), instead of this new `ENV_ZENML_DEPLOYMENT_ID` env var. This takes care of newly deployed tenants.
* for existing tenants, you can either manually update the IDs in the database to retrospectively match their tenant IDs, or you can also patch `SQLZenStore.get_deployment_id()` to use the same override. I prefer the former, because it keeps things nice and consistent and persists all information in the DB. Also easier to apply without upgrades.

Please note that this change is disruptive in many ways, because the deployment ID is used in quite a few places:

* all existing API tokens are invalidated (users will have to re-login, running pipelines will fail unless they use service accounts)
* obviously, it's like a complete change of identity, so new telemetry events will be disconnected from the old ones, unless some other means are used to group them

stefannica (changes requested):

src/zenml/zen_stores/sql_zen_store.py
schustmi: I agree we should probably reuse the `ZENML_SERVER_EXTERNAL_SERVER_ID` env variable. But even if doing that, my question remains the same: Should I just use it once when the DB is first initialized and from that point on use the value stored in the `Identity` table of the database. Or should this env variable be checked anytime the `server_info.id` is checked?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1377,11 +1377,11 @@ def migrate_database(self) -> None:
...
                 )
             with Session(self.engine) as session:
-                session.add(
-                    IdentitySchema(
-                        id=str(GlobalConfiguration().user_id).replace("-", "")
-                    )
+                id_ = (
+                    ServerConfiguration.get_server_config().external_server_id
+                    or GlobalConfiguration().user_id
                 )
```
stefannica, line 1383: You missed the `replace("-", "")` part here. Is that intentional ? In the database, UUIDs are stored without their hyphens.

stefannica (approved):
Aside from the comment regarding extra hyphens, this looks good

src/zenml/zen_stores/sql_zen_store.py
schustmi: Yeah that was intentional. The UUID is stored as a string without their hyphens, but sqlmodel/pydantic takes care of that for us (We use UUID in lots of other places). What really happened here I think is the following:
- We convert the UUID to a string and replace the hyphens
- We create the pydantic/SQLModel object, which converts the string to a UUID.
- When storing in the DB, SQLModel again converts to a string and replaces the hyphens 😄

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Got it! Thanks.

bcdurak (approved):

PR #2613, opened 2024-04-12
Allow DB backup failures if the database version is 0.56.3 or earlier

stefannica:
## Describe changes
The database backup and restore feature wasn't fully functional until now due to problems in the DB schema:

* table dependency loop impacting all backup strategies
* improper pagination when the database backup strategy is used

Moreover, even if the DB upgrade is successful, the database backup step will fail if the database backup strategy is used, and basically interrupt the entire upgrade process.

We want to avoid any unnecessary administrative operations to be incurred by ZenML users and thus we allow database backup failures to occur with earlier database versions. In these cases, the upgrade process will continue uninterrupted without a backup. Things will go back to normal (i.e. DB backup errors will be reported as failures and interrupt the upgrades) after ZenML is upgraded to the next release.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -1321,9 +1321,19 @@ class SqlZenStore(BaseZenStore):
                     # database schema. If the database is at version 0.56.3
                     # or earlier and if the backup fails, we only log the
                     # exception and leave the upgrade process to proceed.
-                    if version.parse(current_revisions[0]) <= version.parse(
-                        "0.56.3"
-                    ):
+                    allow_backup_failures = False
+                    try:
+                        if version.parse(
+                            current_revisions[0]
+                        ) <= version.parse("0.56.3"):
+                            allow_backup_failures = True
+                    except version.InvalidVersion:
+                        # This can happen if the database is not currently
+                        # stamped with an official ZenML version (e.g. in
+                        # development environments).
+                        pass
+
+                    if allow_backup_failures:
                         logger.exception(
                             "Failed to backup the database. The database "
                             "upgrade will proceed without a backup."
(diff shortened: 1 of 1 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1315,13 +1316,26 @@ def migrate_database(self) -> None:
...
-                        "functionality, set the `backup_strategy` attribute "
-                        "of the store configuration to `disabled`."
-                    ) from e
+                    # The database backup feature was not entirely functional
+                    # in ZenML 0.56.3 and earlier, due to inconsistencies in the
+                    # database schema. If the database is at version 0.56.3
+                    # or earlier and if the backup fails, we only log the
+                    # exception and leave the upgrade process to proceed.
+                    if version.parse(current_revisions[0]) <= version.parse(
+                        "0.56.3"
```
strickvl, line 1325: This does full semantic comparision?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1315,13 +1316,26 @@ def migrate_database(self) -> None:
...
+                    ):
+                        logger.exception(
+                            "Failed to backup the database. The database "
+                            "upgrade will proceed without a backup."
+                        )
+                    else:
+                        raise RuntimeError(
+                            f"Failed to backup the database: {str(e)}. "
+                            "Please check the logs for more details."
```
strickvl, line 1334:
```suggestion
                            "Please check the logs for more details. "
```

strickvl (changes requested):

src/zenml/zen_stores/sql_zen_store.py
stefannica: Yes, but you're right about one thing: this will break if not running on an actual official release. Let me fix that.

stefannica: Done

strickvl (approved):

PR #2616, opened 2024-04-17
Fix GCP commands in docs for `project_id`

strickvl:
Docs were incorrectly referring to `project-id` when it should have been `project_id`. (This came up in a slack conversation).


stefannica (approved):
Thanks so much for fixing this!

PR #2617, opened 2024-04-17
Slimmer and more secure Docker container images

stefannica:
## Describe changes
Brings a number of improvements to the container image build process and results:

* use the slimmest recommended base image that has python and pip installed (python:3.x-slim-bookworm instead of python:3.x). NOTE: alpine is not recommended because it can lead to [longer build times, obscure bugs, and performance issues](https://pythonspeed.com/articles/alpine-docker-python/).
* using separate stages for building and runtime
* disabling cache and cleaning up unneeded files
* use a virtual environment to install Python packages instead of system wide
* switch to Python 3.11 for the server image
* ignore vulnerabilities for which a patch isn't published yet, of which there are about 100 at the moment. The reason is we can't possibly fix those vulnerabilities until a patched package version is available.
* enable debug logging in the server development image by default
* installing some useful utilities in the development images (network and database utilities, curl, git, etc.)
* enable proxy headers for uvicorn so that the server logs display the real source IP address
* disable `PYTHONUNBUFFERED` in client images because this causes containerized pipeline steps to be very slow when using an external artifact store like S3 or GCS. Buffering logging messages means having to call the external object storage less frequent and incurring fewer delays. Ideally, this should be an asynchronous process, but this fix will have to do for now.

## Side Changes

This PR also:

* fixes an infinite loop in the step logging feature, visible only when debug logging is turned on
* fixes the test harness to use the same python version for building the development base image as the system's python version

## Breaking Changes

Users who utilize the ZenML base (client) or server image as a base image to build their own custom images should be aware of the following changes:

* zenml and its dependencies are now installed in a virtual environment under `/opt/venv` instead of being installed system-wide
* the base image is python:3.x-slim-bookworm instead of python:3.x, which means some system packages might now be missing

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **Security Enhancements**
  - Updated vulnerability scanning tools to ignore unfixed vulnerabilities, enhancing security without hindering development.
  - Improved Docker configurations to exclude unnecessary files, reducing potential exposure to security risks.

- **Infrastructure Updates**
  - Upgraded base Docker images and split the build process to enhance efficiency and security.
  - Adjusted nightly and regular build processes to use updated Python versions, ensuring compatibility with newer software.

- **New Features**
  - Exposed port 8080 for the ZenML server, facilitating easier access and integration.


strickvl (approved):

strickvl:
@coderabbitai review

strickvl:
I'd be careful about using `uv` by default. There are enough random issues to make me wonder whether the `uv`-ification of our Docker builds shouldn't wait a month or three... I know @schustmi had similar concerns. And @avishniakov had some issues with `uv` as well this week, though something unrelated.

avishniakov:
Agree with @strickvl , UV, I'm pretty sure, even broke the LLM template, not talk about the other issues we faced. It is still quite unstable at the moment.

stefannica:
> I'd be careful about using `uv` by default. There are enough random issues to make me wonder whether the `uv`-ification of our Docker builds shouldn't wait a month or three... I know @schustmi had similar concerns. And @avishniakov had some issues with `uv` as well this week, though something unrelated.

Alright, I removed uv from the dockerfiles. Something left for when uv is more stable.

strickvl (approved):
Looks good from my side, but I'd like to merge this in *after* today's release so that we have a bit of time to kick the wheels before the next release.

PR #2628, opened 2024-04-22
Label studio settings and optional port

htahir1:
I added an api key settings to the label studio annotator so that we can share the stack component.
Also port is now optional when using the annotator.

## Summary by CodeRabbit

- **New Features**
	- Introduced new settings for Label Studio integration to enhance API key management.


strickvl:
@coderabbitai review

strickvl:
Should probably merge https://github.com/zenml-io/zenml/pull/2626 here too

htahir1:
@strickvl I merged your branch, you can close yours and just review this

src/zenml/integrations/label_studio/annotators/label_studio_annotator.py
```diff
@@ -49,13 +53,26 @@ def config(self) -> LabelStudioAnnotatorConfig:
         """
         return cast(LabelStudioAnnotatorConfig, self._config)
 
+    @property
+    def settings_class(self) -> Type[LabelStudioAnnotatorSettings]:
+        """Settings class for the Kubeflow orchestrator.
```
strickvl, line 58:
```suggestion
        """Settings class for the Label Studio annotator.
```

strickvl (approved):

src/zenml/integrations/label_studio/flavors/label_studio_annotator_flavor.py
```diff
@@ -30,18 +32,31 @@
...
     Attributes:
         instance_url: URL of the Label Studio instance.
         port: The port to use for the annotation interface.
+        api_key: The api_key for label studio.
     """
 
     instance_url: str = DEFAULT_LOCAL_INSTANCE_URL
-    port: int = DEFAULT_LOCAL_LABEL_STUDIO_PORT
+    port: Optional[int] = DEFAULT_LOCAL_LABEL_STUDIO_PORT
+    api_key: Optional[str] = SecretField()
```
bcdurak, line 46: In most integrations (that use an API key or a port as a part of a flavor implementation), we placed these fields as parts of the `Config` class instead of the `Settings` class. Is there a reason why we opted out of this behavior? If not, I would propose that the `port` and `api_key` actually belong to the `Config` object and have to be defined upon the registration of the component.

bcdurak (commented):
Left a small comment. Otherwise LGTM.

src/zenml/integrations/label_studio/flavors/label_studio_annotator_flavor.py
strickvl: We did this specifically so as to allow quick switching for this component when 3 people were using the same annotator, but we didn't want to each separately to have to configure a separate component. We wanted to use the same component in our shared server, but with each passing in different values for the `api_key` via a config file.

bcdurak (approved):

PR #2630, opened 2024-04-23
ZenML Server activation and user on-boarding

stefannica:
## Describe changes

This PR changes the way that ZenML server are being provisioned immediately after deployment to exclude the automated creation of the so called `default` user. A new server activation flow is introduced that works as follows:

* newly deployed ZenML servers will not have any user accounts configured by default anymore. The "default user" concept is removed. Users cannot log into the server until they've gone through the server activation flow (see below)
* after deployment, the ZenML server is in an `inactive` state, marked by a False value for the new `active` field in the server info object. When the dashboard is opened in this state, the user will be walked through the server activation flow, which consists of the following:
  * supplying a username for the initial admin account
  * supplying a password for the initial admin account
  * setting the server name
  * after all this information is collected, the dashboard calls the new `/api/v1/activate` endpoint that 1. activates the server (toggles `active` to True) 2. creates an initial admin user account with the given name/password 3. automatically logs in the user using the new account

The server activation feature only works with the new OSS dashboard. If users want to explicitly configure new ZenML server deployments to use the old dashboard, the legacy behavior of automatically activating the server and automatically creating the default admin user account will be automatically enabled (FYI @avishniakov ).

The legacy behavior can also be enabled by setting the `ZENML_SERVER_AUTO_CREATE_DEFAULT_USER` environment variable to `True` for those who need a ready-to-go ZenML server that is automatically pre-activated on deployment.

The following server deployments are unaffected by this and will continue to have a `default` user account created automatically on deployment:

* local client deployments connected directly to a database
* local ZenML servers deployed with `zenml up`
* test harness deployments
* deployments that are explicitly configured to use the old dashboard

## Side changes

This includes some changes required to support server settings and user on-boarding:

* the `identity` table has been renamed to `server_settings` and now contains new server-wide settings attributes in addition to the previous `id` deployment identifier:
  * the new `active` attribute
  * analytics opt in
  * whether to display announcements in the dashboard
  * whether to display information about zenml updates in the dashboard
  * onboarding status - server global on-boarding metadata that can be set by the dashboard
* the user now has a new `user_metadata` attribute that can be used by the dashboard to store additional information

## Backwards compatibility

Existing server deployments are already active and have at least one admin account already configured.

## Pre-requisites

## Types of changes


src/zenml/analytics/context.py
```diff
@@ -77,7 +77,12 @@ def __enter__(self) -> "AnalyticsContext":
         from zenml.config.global_config import GlobalConfiguration
 
         gc = GlobalConfiguration()
-        self.analytics_opt_in = gc.analytics_opt_in
+        store_info = gc.zen_store.get_store_info()
+
+        if self.in_server:
```
avishniakov, line 82: Cannot this in init, what's this?

src/zenml/analytics/context.py
```diff
@@ -77,7 +77,12 @@ def __enter__(self) -> "AnalyticsContext":
         from zenml.config.global_config import GlobalConfiguration
 
         gc = GlobalConfiguration()
-        self.analytics_opt_in = gc.analytics_opt_in
+        store_info = gc.zen_store.get_store_info()
```
avishniakov, line 80: Maybe push this below the `if self.in_server`? It is not needed anywhere else.

src/zenml/config/server_config.py
```diff
@@ -206,6 +206,21 @@ class ServerConfiguration(BaseModel):
...
+        display_updates: Whether to display notifications about ZenML updates in
+            the dashboard. Used only during initial deployment. Can be changed
+            later as a part of the server settings.
+        auto_create_default_user: Whether to automatically create a default
+            admin user account with an empty password. Used only during initial
+            deployment.
+        auto_activate: Whether to automatically activate the ZenML server during
+            deployment. Used only during initial deployment. If
+            `auto_create_default_user` is set to `True`, this setting is ignored
+            and the server is activated automatically.
```
avishniakov, line 223: Can you explain why they are different? When I want to use `auto_activate` instead of `auto_create_default_user`?

docs/book/deploying-zenml/zenml-self-hosted/deploy-with-docker.md
```diff
@@ -266,6 +263,13 @@ The following secure headers environment variables are supported:
 * **ZENML\_SERVER\_SECURE\_HEADERS\_CACHE**: The `Cache-Control` HTTP header value. The default value is `no-store, no-cache, must-revalidate`.
 * **ZENML\_SERVER\_SECURE\_HEADERS\_PERMISSIONS**: The `Permissions-Policy` HTTP header value. The default value is `accelerometer=(), camera=(), geolocation=(), gyroscope=(), magnetometer=(), microphone=(), payment=(), usb=()`.
 
+If you prefer to activate the server automatically during the initial deployment and also automate the creation of the initial admin user account, this legacy behavior can be brought back by setting the following environment variables:
+
+* **ZENML\_SERVER\_AUTO\_CREATE\_DEFAULT\_USER**: Set this to `1` to automatically create the initial admin user account when the server is first deployed. Defaults to `0`.
+* **ZENML\_DEFAULT\_USER\_NAME**: The name of the initial admin user account created by the server on the first deployment, during database initialization. Defaults to `default`.
+* **ZENML\_DEFAULT\_USER\_PASSWORD**: The password to use for the initial admin user account. Defaults to an empty password value, if not set.
```
avishniakov, line 270: Maybe I misread the whole thing, but it seems that Name and Password were whipped from everywhere, but left here. Is that right?

avishniakov (changes requested):
Cool, that Server Settings are in the DB now 👍🏼 
Left a few minor changes/questions.

docs/book/deploying-zenml/zenml-self-hosted/deploy-with-docker.md
stefannica: Yes, we'll need to point people to this section in the docs if they also want to deploy the old dashboard instead of the new one. The old dashboard doesn't support the server activation flow.

avishniakov: Ah, clear

src/zenml/analytics/context.py
schustmi: It is needed below.

avishniakov: Yes, found it now, thanks

src/zenml/analytics/context.py
avishniakov: My bad again, I was not looking good enough to notice that this is a property of `AnalyticsContext` - I only checked the `__init___` method

src/zenml/config/server_config.py
stefannica: you're probably right, it is overkill to have two configuration options. I'll remove the `auto_activate` one.

PR #2633, opened 2024-04-24
Send workflow dispatch event to Cloud Plugins repo on release

wjayesh:
## Describe changes
I implemented an addition to the Docker image building workflow to send an event to the `zenml-io/zenml-cloud-plugins` repo when zenml server images are built.

This allows us to build new `zenml-cloud-server` images based on top of the latest zenml server images.

## Testing

I have tested this to send an event to the plugins repo.
- action on this repo: https://github.com/zenml-io/zenml/actions/runs/8833520859
- action on the plugins repo: https://github.com/zenml-io/zenml-cloud-plugins/actions/runs/8833523496/job/24253038897

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/publish_docker_image.yml
+++ b/.github/workflows/publish_docker_image.yml
@@ -51,7 +51,7 @@ jobs:
               owner: 'zenml-io',
               repo: 'zenml-cloud-plugins',
               workflow_id: 'main_image_builder.yml',
-              ref: 'feature/automate-image-building-on-zenml-tag-push',
+              ref: 'cloud/send-workflow-dispatch-on-release',
               inputs: {
                 tag: '0.56.3',
               }
--- a/.github/workflows/publish_docker_image.yml
+++ b/.github/workflows/publish_docker_image.yml
@@ -8,7 +8,6 @@ on:
         required: false
         type: string
         default: release-cloudbuild.yaml
-
 jobs:
   publish_to_docker:
     name: Publish Docker 🐋 image 🖼️ to Dockerhub
@@ -37,7 +36,6 @@ jobs:
             --quiet \
             --config=${{inputs.config_file}} \
             --substitutions=TAG_NAME=${{github.ref_name}} .
-
     wait-for-package-release:
       runs-on: arc-runner-set
       needs: publish_to_docker
@@ -50,12 +48,12 @@ jobs:
     dispatch_to_cloud_plugins_repo:
       runs-on: ubuntu-latest
       needs: wait-for-package-release
-      steps:  
+      steps:
         - name: Trigger ZenML Cloud server image creation
           uses: actions/github-script@v7
           with:
             github-token: ${{ secrets.CLOUD_PLUGINS_REPO_PAT }}
-            script: |
+            script: |-
               await github.rest.actions.createWorkflowDispatch({
                 owner: 'zenml-io',
                 repo: 'zenml-cloud-plugins',
--- a/.github/workflows/publish_docker_image.yml
+++ b/.github/workflows/publish_docker_image.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Trigger ZenML Cloud server image creation
         uses: actions/github-script@v7
         with:
-          github-token: ${{ secrets.CLOUD_BE_REPO_PAT }}
+          github-token: ${{ secrets.CLOUD_PLUGINS_REPO_PAT }}
           script: |
             await github.rest.actions.createWorkflowDispatch({
               owner: 'zenml-io',
--- a/.github/workflows/publish_docker_image.yml
+++ b/.github/workflows/publish_docker_image.yml
@@ -8,56 +8,54 @@ on:
         required: false
         type: string
         default: release-cloudbuild.yaml
-  push:
-    branches:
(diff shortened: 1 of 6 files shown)


.github/workflows/publish_docker_image.yml
```diff
@@ -36,3 +37,19 @@ jobs:
             --quiet \
             --config=${{inputs.config_file}} \
             --substitutions=TAG_NAME=${{github.ref_name}} .
+
```
schustmi, line 40: The only thing I'm wondering is whether we need some delay here so the images are actually available and ready to pull from GCP? I know that we also sleep for a few minutes after the PyPI release before we start the docker builds because we ran into issues before.

strickvl: Probably can't hurt.

strickvl (approved):

.github/workflows/publish_docker_image.yml
wjayesh: Agreed. I added the sleep job before sending the dispatch event. Thanks!

avishniakov (approved):

.github/workflows/publish_docker_image.yml
```diff
@@ -36,3 +36,30 @@ jobs:
...
+    # Create a workflow dispatch to the ZenML Cloud API repository
+    dispatch_to_cloud_plugins_repo:
+      runs-on: ubuntu-latest
+      needs: wait-for-package-release
+      steps:
+        - name: Trigger ZenML Cloud server image creation
+          uses: actions/github-script@v7
+          with:
+            github-token: ${{ secrets.CLOUD_PLUGINS_REPO_PAT }}
+            script: |-
```
schustmi, line 56: I have to say I'm still not happy with how this is handled on the receiving end of this workflow dispatch.
IMO, this action should maybe create a tag in the cloud plugins repo, which then triggers the release? Right now this is all a mess because of different ways of doing a release
- workflow dispatch
- manual tag of cloud plugins repo in which case we query the latest OSS release as a base (even this is not a given, we might want to patch older releases)

I think we have to figure out a way to make this consistent (e.g. have tags for all releases on cloud-plugins), otherwise this will be very hard to keep track of.

schustmi (changes requested):

.github/workflows/publish_docker_image.yml
wjayesh: I don't want this workflow to create tags on the plugins repo because it doesn't make sense to have a new tag if nothing changes on the plugins repo.

## How I see the flow:
### From the zenml-io/zenml repo side:
- You make a ZenML release.
- This workflow runs and sends a dispatch to the plugins repo with the release zenml tag.
- The plugin repo takes the latest plugins code (that is released) and uses this zenml tag as base and creates a new cloud server image.

### From the zenml-io/zenml-cloud-plugins repo:
- You make some changes to the plugins repo, while the zenml repo stays the same.
- You make a release there by pushing a tag. 
- By default, the plugins repo chooses the latest released version of zenml and builds a new cloud server image using that as base and adds your latest plugins release code on top of it.

## Customization
What you can also do with the PR now is the following:

### Release plugins repo but use a different zenml OSS version than the latest
- You can set what zenml version you want to use inside the workflow yaml file and when you push a tag on the plugins repo, that zenml version gets picked up.

### Release zenml-io/zenml but build an image using a different plugins repo release
- rare event but option present nonetheless

### Create dev images
- choose whatever plugins repo branch you want.
- choose the zenml version you want as base.
- choose what region to push to.
- call it from the github UI.

schustmi: Why do you think tagging in `zenml-cloud-plugins` is not a good idea when triggered by the zenml release workflow?

I see it as follows:
- The `zenml-cloud-plugins` builds a Docker image as its release artifact.
- I would like to see which code version in `zenml-cloud-plugins` was used for a specific release of that Docker image. I would like to see that in my code repository, not by inspecting some hidden Docker image tags.
- I don't see the issue of having multiple tags on the same commit in the `zenml-cloud-plugins` repository. Having two tags `0.60.0` and `0.61.0` would actually give me just the information I want when looking at tags in that repository: There were two releases made, both with the same `zenml-cloud-plugins` code version but based on different OSS releases.

As mentioned above, the inconsistency in the different release processes just makes things unnecessarily complex in my opinion:
- There is no simple way to look at all releases and their associated plugins code version
- There is not simple way to look at all releases and their associated OSS code version

wjayesh:
@schustmi is this one good to go?

schustmi:
@wjayesh I see that you modified the workflow that gets dispatched to only create a tag. Wouldn't it be easier to just create a tag in that repo directly, or is this workflow dispatch event still the way to go?

wjayesh:
@schustmi I wanted to keep the workflow dispatch in place just in case we want to change the implementation in the future. This setup just gives us an easy way to communicate with the other repo.

schustmi:
I think it's still weird to do it to be honest, if we ever needed this again we could just copy-paste this from stackoverflow, no need to keep that code around. But if we keep it around, we should at least give the workflow a name that resembles what it does..

PR #2641, opened 2024-04-25
Pigeon annotator

strickvl:
Implemented a new annotator:

pigeon

It only works within Jupyter notebooks and supports a limited feature set but is great for experimentation and demos.

Labels get saved to a user-specified directory as JSON files and also returned back to the user as a list of tuples.

(@schustmi: told you I'd do this one day ;) )

# Breaking Change

The interface for the base class of the annotator stack component has been updated to account for the fact that not all annotator will launch with a specific URL. So there is no longer an `url` argument passed in.


src/zenml/integrations/pigeon/annotators/pigeon_annotator.py
```diff
@@ -0,0 +1,322 @@
...
+        Args:
+            dataset_name: Name of the dataset (annotation file).
+
+        Returns:
+            A tuple containing (num_labeled_examples, num_unlabeled_examples).
+        """
+        dataset_path = os.path.join(self.config.output_dir, dataset_name)
+        num_labeled_examples = sum(1 for _ in open(dataset_path))
+        num_unlabeled_examples = 0  # Assuming all examples are labeled
+        return num_labeled_examples, num_unlabeled_examples
```
safoinme, line 105: shall we catch errors in case can't be open:
```suggestion
    dataset_path = os.path.join(self.config.output_dir, dataset_name)
    try:
        with open(dataset_path, 'r') as file:
            num_labeled_examples = sum(1 for _ in file)
        num_unlabeled_examples = 0  # Update this logic if you start handling unlabeled data
        return num_labeled_examples, num_unlabeled_examples
    except FileNotFoundError:
        logger.error(f"File not found: {dataset_path}")
        return (0, 0)
```

src/zenml/integrations/pigeon/annotators/pigeon_annotator.py
```diff
@@ -0,0 +1,322 @@
...
+        annotations = []
+        current_index = 0
+        out = widgets.Output()
+
+        def show_next() -> None:
+            nonlocal current_index
+            if current_index >= len(examples):
+                with out:
+                    clear_output(wait=True)
+                    print("Annotation done.")
```
safoinme, line 133: logger instead of print?

src/zenml/integrations/pigeon/annotators/pigeon_annotator.py
```diff
@@ -0,0 +1,322 @@
...
+        def submit_annotations(btn: widgets.Button) -> None:
+            """Submit all annotations and save them to a file.
+
+            Args:
+                btn: The button that triggered the event.
+            """
+            self._save_annotations(annotations)
+            with out:
+                clear_output(wait=True)
+                print("Annotations saved.")
```
safoinme, line 163: same here

safoinme (changes requested):

avishniakov (approved):
Looks good 👍🏼

safoinme (approved):

strickvl:
Had passed the CI previously and no meaningful changes made since.

PR #2665, opened 2024-05-03
Prevent special whitespaces in the names of entities

avishniakov:
## Describe changes
I added masking using the banned set of symbols (only special characters `\t\n\r\v\f` are considered banned with this PR) for tags, models, artifacts, stacks, and stack components.
This impacts only the creation of new entities - old ones with invalid names will remain to work, as before, but new ones cannot be created using banned symbols.

Original motivation: symbols like `\n` would be allowed in API calls, but it breaks the Dashboard subsequently.

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
  - Enhanced name validation across various components and requests to ensure names contain only allowed characters.
  
- **Bug Fixes**
  - Implemented additional name validation logic in artifact, component, model, model version, stack, and tag requests to prevent invalid name entries.

- **Tests**
  - Added new tests to verify that invalid names are correctly handled and trigger exceptions in various system components.


strickvl:
@coderabbitai review

tests/integration/functional/zen_stores/test_zen_store.py
```diff
@@ -2634,6 +2651,31 @@ def test_register_stack_fails_when_stack_exists():
                     )
 
 
+def test_register_stack_fails_with_invalid_name():
+    """Tests registering stack fails with invalid name."""
+    client = Client()
+    store = client.zen_store
```
strickvl, line 2657:
```suggestion
    store = Client().zen_store
```

tests/integration/functional/zen_stores/test_zen_store.py
```diff
@@ -2497,6 +2497,23 @@ def test_count_stack_components():
         assert store.count_stack_components(filter_model) == count_before + 1
 
 
+def test_stack_component_create_fails_with_invalid_name():
+    """Tests that creating a stack component with an invalid name fails."""
+    client = Client()
+    store = client.zen_store
```
strickvl, line 2503:
```suggestion
    store = Client().zen_store
```

strickvl (commented):
Just nits. Otherwise looks good to me. I look to @bcdurak for the Pydantic perspective, however.

tests/integration/functional/zen_stores/test_zen_store.py
avishniakov: USed below: `workspace=client.active_workspace.id`

tests/integration/functional/zen_stores/test_zen_store.py
avishniakov: Used below: `client.active_workspace.id`

tests/integration/functional/zen_stores/test_zen_store.py
```diff
@@ -4051,6 +4100,19 @@ def test_list_by_tag(self, clean_client: "Client"):
             )
             assert len(ms) == 0
 
+    def test_create_fails_with_invalid_name(self):
+        """Test that creation fails with invalid name."""
+        c = Client()
+        zs = c.zen_store
```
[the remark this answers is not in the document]
avishniakov: Used below: `c.active_user.id`

avishniakov:
> Just nits. Otherwise looks good to me. I look to @bcdurak for the Pydantic perspective, however.

I made it in a bit more functional way, so no impact on Pydantic.

avishniakov:
@coderabbitai review

strickvl (approved):

bcdurak (approved):
I like the changes. Thank you for making it less pydantic heavy as well, makes my job a lot easier. 😄 One question though, when it comes to named entities, I thought this approach would also go well with pipelines and pipeline runs too (not sure if I am forgetting about anything else). Is there a reason why we left those out?

Feel free to merge it. Going forward, I think we can implement a simple validator for names like this Pydantic v2 to generalize this even further.

avishniakov:
> I like the changes. Thank you for making it less pydantic heavy as well, makes my job a lot easier. 😄 One question though, when it comes to named entities, I thought this approach would also go well with pipelines and pipeline runs too (not sure if I am forgetting about anything else). Is there a reason why we left those out?
> 
> Feel free to merge it. Going forward, I think we can implement a simple validator for names like this Pydantic v2 to generalize this even further.

Make sense, we have way more named entities indeed. I'll create a ticket with low prior to revisit this later.
The problem with making it as validator in the Pydantic is the fact that old "poorly" named entities should still work, so it might be tricky to restrict this on the model level.

PR #2671, opened 2024-05-06
[k8s orchestrator] Fix credentials refresh and don't use service connector for incluster auth

wjayesh:
## Describe changes
This PR fixes a bug which prevented long running pipelines from running on the Kubernetes orchestrator. The source of the problem was the credentials not being refreshed inside the Kubernetes orchestrator pod.

### Core Changes:
- the `get_kube_client` function of the k8s orchestrator is now called directly inside the `wait_pod` method.
- this ensures that the function always checks/fetches the latest credentials every second (the frequency of the wait pod check).

### Additional Changes:
- we don't use the credentials from the service connector for processes running inside the cluster and use the k8s service account credentials instead.
- We add a buffer to the token validity check for the boto3 session to avoid any edge cases where the token expires at the very next moment after a check.

## Questions
- Is it okay to pass in the orchestrator to the `wait_pod` function?
- Should I make `wait_pod` a function of the `KubernetesOrchestrator` class?

## Rationale behind not using service connector credentials

This section details the reasoning behind why using the k8s service account credentials over the service connector creds makes sense.

## Background
- Every pod in kubernetes has a service account attached to it which is used to perform actions on its behalf.
- The service account has some creds (a token) associated with it, that identifies this SA uniquely.
- This token is automatically generated by the kubelet and mounted to the pod at some path in its filesystem.
- When you load cluster config inside any process running in the pod, using any kubernets client library (in our case, the python library) with `incluster` set to `True`, this file is read, and the token is extracted.
- This token corresponds to the service account so only the permissions attached to this service account can be used by the process wielding the token.

## Why it makes logical sense

- When running inside the pod, you would want that the process respects the permissions that you allow to the pod.
- We create a service account already for the orchestrator pod and add permissions that it will use.
- By using the service connector credentials instead, we "assume" the role of which ever account was used while configuring the service connector and the permissions that come with the mapping of that role with k8s roles.
- This means that the pod can potentially have more power to perform actions than what we are "aware" of.

## Robust to edge cases

This mechanism is also robust to edge cases where the token may expire between the time you check its validity and when its used. Our service connector option also has the same robustness. However, I'm putting this block here to alleviate concerns that this route might not be robust to that scenario.

- The token that is attached to the service account and mounted to the pod is automatically refreshed and doesn't depend on any function call to get client to check if the validity has run out. As opposed to the service connector route, which sits with its token for as long as until the validity is explicitly checked. So in short, more frequent refreshes reduce the probability that the token you have might expire soon.
- The kubelet refreshes the token before it expires so the probability that the token you received from the mounted path at the pod further reduces or is non-existent.

## Pre-requisites

## Types of changes


src/zenml/integrations/aws/service_connectors/aws_service_connector.py
```diff
@@ -713,7 +713,10 @@ def get_boto3_session(
             # Refresh expired sessions
             now = datetime.datetime.now(datetime.timezone.utc)
             expires_at = expires_at.replace(tzinfo=datetime.timezone.utc)
-            if expires_at > now:
+            # check if the token expires in the next 10 minutes
```
schustmi, line 716:
```suggestion
            # check if the token expires in the next 5 minutes
```

This or change the timedelta :)

I can't see any issues with this change but I'm also not super familiar with service connectors. Worst case this would lead to refreshing the session more often, correct? Is this something we would also need in other service connector implementations?

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
```diff
@@ -169,26 +174,29 @@ def get_pod(
 
 
 def wait_pod(
-    core_api_fn: Callable[[], k8s_client.CoreV1Api],
+    orchestrator: "KubernetesOrchestrator",
```
schustmi, line 177: I don't think we want a dependency in our general kubernetes utils on the orchestrator, these might be used in other places and should not even be aware of what an orchestrator is. Instead I would suggest passing in a function as before, but maybe one that returns a kubernetes client? We could then pass in `orchestrator.get_kube_client` and achieve the same result, correct?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -88,6 +88,14 @@ def get_kube_client(
         if incluster is None:
             incluster = self.config.incluster
 
+        if incluster:
```
schustmi, line 91: If we use the incluster option, why is that better than using the service connector? Is there no expiration on the token that the incluster client uses?

schustmi (changes requested):

src/zenml/integrations/aws/service_connectors/aws_service_connector.py
wjayesh: - I'll make the change to 5. thanks for the catch :)
- a session is valid for 1 hour and so it shouldn't lead to frequent refreshing of session creds.
- we already have a buffer when refreshing tokens for the EKS cluster (which we get using the session) so there is precedent. I remember we did this because someone ran into this edge case on slack. This change here is just precautionary and there have been no reports yet.
- I'll take a look at other service connector implementations to check if this is needed elsewhere.

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
wjayesh: Makes sense! I can pass `get_kube_client` instead of the orchestrator and call it inside the `wait_pod` function. Let me implement and test if that works.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
wjayesh: It's not "better" but it is the right thing to do.
Processes running in a pod should use the credentials that are made available to the pod through its service account. This ensures that only the permissions that the service account is allowed, area accessible to any functions that run inside the pod.

Currently, we use the credentials of an AWS IAM role that was used for the purpose of authenticating to the cluster. And that entity is mapped to the master role in our kubernetes cluster, which means it has all permissions in every namespace.

src/zenml/integrations/aws/service_connectors/aws_service_connector.py
schustmi: Alright sounds good 👌

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
schustmi: I agree it's the right thing to do. Can it be confusing for users though because of the following scenario: They give ZenML certain permissions (using the service connector) but the service account used for the pod is missing some of those permissions which are needed to run the pipeline?

In any case, for the reason mentioned above this could be a breaking change for some users/scenarios right?

wjayesh: The pod in question here is the orchestrator pod whose only purpose is to query the kubernetes api server to fetch information about the pod status of the different step pods. Thus, the permissions that it needs are kubernetes permissions to read pod resources. 

We create a custom k8s service account for this pod to use and give it the necessary permissions. Now, the code is implemented in a way that this service account is not being used at all. Had this been the intended behavior, we would not create this service account explicitly in the first place.

wjayesh: The step code where the user might expect to use any permissions runs in different pods. These pods don't even have a custom k8s service account and just use the default. They are not intended to be used to access external resources (like AWS resources) directly. 

When a user wants to allow its pipelines to use external services, they should use service connectors, 
- either explicit methods, or 
- the implicit method where they leverage the IAM role we create for every tenant and can be used for their service connectors without providing any credentials.
- this IAM role currently has only permissions that we give but we can allow users to edit it.

wjayesh: In other words, the EKS connector is only supposed to help users connect to the EKS resource and not all other AWS resources. This connector's only purpose is to allow the user to authenticate and submit k8s jobs to the cluster.

Any other IAM permissions are out of scope for this connector and it is not aware of them.

wjayesh: This will also not lead to any breaking changes because the orchestrator pod's service account that we will now start using already has all the Kubernetes permissions that the orchestrator pod needs. And it also reflects on my tests so far.

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
```diff
@@ -180,15 +181,17 @@ def wait_pod(
...
+            periodically and is used to get a `CoreV1Api` client for 
+            the Kubernetes API. It should cache the client to avoid
             unnecessary overhead but should also instantiate a new client if
             the previous one is using credentials that are about to expire.
         pod_name: The name of the pod.
         namespace: The namespace of the pod.
         exit_condition_lambda: A lambda
             which will be called periodically to wait for a pod to exit. The
             function returns True to exit.
+        incluster: Whether to load the in-cluster config. Defaults to False.
```
schustmi, line 194: Remove this

schustmi (approved):

PR #2677, opened 2024-05-07
Accelerate scaler

avishniakov:
## Describe changes
This PR introduces the concept of Scalers, which can be passed into the steps' definitions to allow step run on some parallelization engine. Scalers included here:
- AccelerateScaler (a major goal of this PR): allows to run a step function via `accelerate run...` without any effort for rewriting the step (except the fact, that it should properly handle accelerate by itself, like save in the main process, etc.)
- AggregateScaler - this is more of a demo of what we can do for other use cases. This scaler just parallelizes the step logic locally and later aggregates back the results using the chosen aggregate function.

Now I would like to gather some early feedback from you.

## Examples

```python
from zenml import step
from zenml.integrations.accelerate import AccelerateScaler

@step(scaler=AccelerateScaler(num_processes=42))
def training_step(some_param: int, ...):
    # your training code is below
    ...
```

```python
from zenml.scalers import AggregateScaler

@step(scaler=AggregateScaler(parameters={"a":[1,2,3],"b":[4,5,6]}, agg_function="sum"))
def training_step_with_sum_aggregation(a:int = None, b:int = None, c:int = 2)->int:
    # your code is below
    return a+b+c

@pipeline
def pipeline_with_aggregate_scaler():
    training_step_with_sum_aggregation(c=3)
# actual step output would be (1+4+3)+(2+5+3)+(3+6+3) = 30,
# where last "+3" comes from constant `c` parameter
```

## To be done still in this PR:
- Tests
- Docs
- Make accelerate a proper integration

How this works in the wild: https://github.com/zenml-io/zenml-projects/pull/102

## Pre-requisites

## Types of changes


avishniakov:
@htahir1 not putting you to reviewers, but you might have what to add to this story.

src/zenml/integrations/accelerate/__init__.py
strickvl, line 1: is this an accelerate integration or is this just part of our huggingface integration? I'm pretty sure you'll get accelerate already with the packages we have defined there?

src/zenml/utils/function_utils.py
```diff
@@ -0,0 +1,221 @@
...
+
+def _is_valid_collection_arg(arg_type: Any) -> bool:
+    if getattr(arg_type, "__origin__", None) in _ALLOWED_COLLECTIONS:
+        if arg_type.__args__[0] not in _ALLOWED_TYPES:
+            return False
+        return True
+    return False
+
+
+def _is_valid_optional_arg(arg_type: Any) -> bool:
```
strickvl, line 79:
```suggestion
def _is_valid_optional_arg(arg_type: Any) -> bool:
    """Check if the given argument type is a valid Optional type.

    A valid Optional type is defined as a Union with two arguments, where:
    - The first argument is either an allowed type or a valid collection type.
    - The second argument is the NoneType.

    Args:
        arg_type: The type to check.

    Returns:
        True if the argument type is a valid Optional type, False otherwise.
    """
```

htahir1:
Wow what a great idea. I havnt look to deep but what is the difference between this and step operators?

avishniakov:
> Wow what a great idea. I havnt look to deep but what is the difference between this and step operators?

Thx, IMO, going forward one can define a step which, for example, takes some slicing parameters and inside the step, some intense data processing is happening and instead of running it with step operator k8s or other, I say scale it using step operator k8s or other. 
This is not yet implemented for sure or shaped well, but this is how I see this: a mixture of specific libs and the reuse of some step operators to give a performance boost in parallelisable operations.
Does that answer?

htahir1:
The only reservation I have is that there are simply too many concepts in ZenML and this is a new one that I haven't really understood yet

avishniakov:
> The only reservation I have is that there are simply too many concepts in ZenML and this is a new one that I haven't really understood yet

That's very valid, based on the feedback we saw. Do you feel that saying `AccelerateStepOperator` is a better option here? On the other hand, it shall be still backed by `VertexStepOperator` or another GPU compute, which makes things really complicated to implement.

avishniakov:
Closed in favor of #2746

PR #2655, opened 2024-05-01
Prodigy annotator

strickvl:
This pull request adds the [Prodigy](https://prodi.gy/) Annotator integration to ZenML. Prodigy is a powerful annotation tool that allows for efficient data labeling. With this integration, users can now connect ZenML with Prodigy and leverage its annotation capabilities in their ML pipelines.

prodigy-annotator

Note, there are no tests included in this PR since `prodigy` the package is only installable via a custom wheel they ship when you buy a license (and thus isn't possible to install and use in our CI). I also made one small change to the base annotator interface to  accommodate this new annotator.

I've developed and tested this on my machine (since I have a license).

## Summary by CodeRabbit

- **New Features**
  - Introduced Prodigy integration for ZenML, enhancing data annotation capabilities.
  - Added new documentation pages for ProdigyAnnotator and updated Table of Contents.
  - Enhanced `BaseAnnotator` to accept additional keyword arguments, improving flexibility.

- **Documentation**
  - Updated documentation to include guides and references for Prodigy integration.

- **Chores**
  - Adjusted installation scripts to include Prodigy in the list of ignored integrations during development setup.

- **Refactor**
  - Modified CLI functions to handle different annotator flavors and improve error handling during dataset operations.

- **Bug Fixes**
  - Fixed package installation handling in utility scripts to manage Prodigy package exceptions.

Issue #2579, opened 2024-04-03
Add Argilla annotator stack component


strickvl:
ZenML currently supports only [Label Studio](https://docs.zenml.io/stacks-and-components/component-guide/annotators/label-studio) as an [annotator stack component](https://docs.zenml.io/stacks-and-components/component-guide/annotators). The work for this issue would add the option to use [Argilla](https://github.com/argilla-io/argilla) as a second option for annotation.

The work would involve (high-level):

- implementing the core interface for annotators as is suitable for Argilla
- updates to documentation where appropriate (at a minimum, a new annotator component page)
- testing, where appropriate and where features have been added that can be tested (esp standalone util functions and so on)

You'll probably want to have a basic example to show that it's able to work together with a simple annotation workflow. A more involved example is [available here](https://github.com/zenml-io/zenml-projects/tree/main/label_studio_annotation) but you can probably think of something smaller to showcase the use of Argilla within a ZenML pipeline.

Depending on when this issue is completed, we might have [an integration with mlstacks](https://github.com/zenml-io/mlstacks/issues/149), in which case that helps solve some of the issues around deployment since it is easily deployed on the Hugging Face Hub.

PR #2687, opened 2024-05-09
Argilla annotator

strickvl:
Added the third and final (for me, for now 😄) annotator integration (alongside https://github.com/zenml-io/zenml/pull/2655 and https://github.com/zenml-io/zenml/pull/2641), this time for Argilla.

argilla_annotator

This works both locally-deployed and when the annotation instance lives in the cloud (i.e. in the Hugging Face Spaces deployment which they recommend).

I tried including some of the changes that have been made in the other two PRs, but depending on the merge order there might need to be some final tweaks. (Also, there probably can be some final updates to the `BaseAnnotator` interface at a later date now that I have a far stronger sense of what's needed and what's not (which we didn't have when we first implemented just Label Studio)).

The Argilla team have mentioned that their SDK / interface will be updated to a 2.x version where lots of things will be fixed, at which point we can update this to fit it.

Relates to https://github.com/argilla-io/argilla/issues/4277 and https://github.com/zenml-io/zenml/issues/2579

# To do
- [x] make a note about breaking change to interface
- [x] ~~Adding datasets should use ‘FeedbackDataset’ as the interface or offer different types~~


src/zenml/integrations/argilla/annotators/argilla_annotator.py
```diff
@@ -0,0 +1,299 @@
...
+
+        Raises:
+            ValueError: when unable to access the Argilla API key.
+        """
+        config = self.config
+        if config.api_key and config.authentication_secret:
+            raise ValueError(
+                "Both API key and authentication secret are provided. "
+                "Please provide only one."
+            )
```
safoinme, line 83: why not prioritize only one of them even if they are both passed, also we don't check if non of them are passed

src/zenml/integrations/argilla/annotators/argilla_annotator.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        dataset_name = kwargs.get("dataset_name")
+        if not dataset_name:
+            raise ValueError("`dataset_name` keyword argument is required.")
+
+        try:
+            if rg.FeedbackDataset.from_argilla(name=dataset_name) is not None:
+                return rg.FeedbackDataset.from_argilla(name=dataset_name)
+            else:
+                return self._get_client().get_dataset(name=dataset_name)
```
safoinme, line 241:
```suggestion
            return rg.FeedbackDataset.from_argilla(name=dataset_name) or self._get_client().get_dataset(name=dataset_name)
```

src/zenml/integrations/argilla/annotators/argilla_annotator.py
```diff
@@ -0,0 +1,299 @@
...
+        except (NotFoundApiError, ValueError) as e:
```
safoinme, line 242: why are we expecting `ValueError` here?

safoinme (changes requested):

src/zenml/integrations/argilla/annotators/argilla_annotator.py
strickvl: We get a value error if `self._get_client().get_dataset(name=dataset_name)` fails, and the other error if the `rg.FeedbackDataset` fails.

src/zenml/integrations/argilla/annotators/argilla_annotator.py
strickvl: Good point. I updated to address this.

src/zenml/integrations/argilla/annotators/argilla_annotator.py
```diff
@@ -0,0 +1,298 @@
...
+            **kwargs: Additional keyword arguments to pass to the Argilla
+                client.
+
+        Returns:
+            An Argilla dataset object.
+
+        Raises:
+            ValueError: if 'dataset_name' and 'dataset' aren't provided.
+        """
+        dataset_name = kwargs.get("dataset_name")
```
bcdurak, line 175: While I am loving the fact that the overall implementation/abstraction for our annotators is coming together, I think the approach that we are using here with the keyword arguments is not very scalable. Especially with a high enough number of "required" parameters, the implementation can become very cluttered and we might end up having to write `kwargs.get("parameter_name")` multiple times like here. I briefly looked into possible solutions but unfortunately could not find a proper one myself right away. However, still wanted to write the comment here as a reminder.

docs/book/stacks-and-components/component-guide/annotators/argilla.md
```diff
@@ -0,0 +1,145 @@
...
+
+(Visit the Argilla documentation and interface to obtain your API key.)
+
+Then register your annotator with ZenML:
+
+```shell
+zenml annotator register argilla --flavor argilla --authentication_secret=argilla_secrets
+```
+
+When using a deployed instance of Argilla, the instance URL must be specified
```
bcdurak, line 65: Can we perhaps handle this with a pydantic validator?

src/zenml/integrations/argilla/annotators/argilla_annotator.py
```diff
@@ -0,0 +1,298 @@
...
+        return f"{self.get_url()}/dataset/{dataset_id}/annotation-mode"
+
+    def get_datasets(self) -> List[Any]:
+        """Gets the datasets currently available for annotation.
+
+        Returns:
+            A list of datasets.
+        """
+        old_datasets = self._get_client().list_datasets()
+        new_datasets = rg.FeedbackDataset.list()
```
bcdurak, line 129: Just for my understanding, can you explain the difference between the `old_datasets` and `new_datasets` here?

bcdurak (commented):
The implementation looks good overall. I left a few small comments. 

I see that you have a few more PRs with different annotator flavors. How does the timeline look for all these flavors?

src/zenml/integrations/argilla/annotators/argilla_annotator.py
strickvl: Yeah so my thought here was that I am also feeling the limits of this strategy. I think when I originally did `label_studio` we wanted to leave it open since I wasn't sure what the API would look like. Now that I've made 4 annotators, I think I have a bit more a solid feel for what the interface should look like, but I was thinking we can first merge these in and then I'd make a subsequent PR that would standardise / update the annotators API for all of them. I think that it can be improved, but I'd do that as a subsequent step. (And maybe I could do this so we merge them in just after a release, and then I'd make the fixes before the next release so that users aren't exposed to the API changes.

src/zenml/integrations/argilla/annotators/argilla_annotator.py
strickvl: :) Ok this is some Argilla weirdness. So they currently have two different datasets APIs within their package at the same time. They have some old dataset types which use one interface, and they have a new type (the `FeedbackDataset` which has a whole other implementation / interface). So depending on when you started using Argilla, you'll be using one or the other API. (Or both!) They say that they're phasing out the old interface, but probably it's going to be weeks / months before this stabilises, so for now we just have to support both somehow. It's unfortunate.

safoinme (approved):

docs/book/stacks-and-components/component-guide/annotators/argilla.md
strickvl: Done. Not sure if this is the canonical way to use a validator, but I figure it's easier just that we handle it rather than forcing the user to go to the trouble of updating their stack component etc...

src/zenml/integrations/argilla/annotators/argilla_annotator.py
bcdurak: That's a good idea. Can resolve this now.

bcdurak (approved):
🦭

PR #2700, opened 2024-05-14
Separate actions from triggers

schustmi:
## Describe changes
This PR separates the action configuration for triggers from the trigger configuration.

Additionally, initial steps towards allowing schedules for triggers have been implemented on the model layer.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1709,6 +1715,205 @@ def _auto_activate_server(self) -> None:
...
+                because it's used by triggers.
+        """
+        with Session(self.engine) as session:
+            action = self._get_action(action_id=action_id, session=session)
+
+            # Prevent deletion of action if it is used by a trigger
+            if action.triggers:
+                raise IllegalOperationError(
+                    f"Unable to delete action with ID `{action_id}` "
+                    f"as it is used by {len(action.triggers)} triggers."
```
AlexejPenner, line 1911: Should we maybe actually allow this and propagate?

AlexejPenner (commented):
Some initial thoughts

src/zenml/models/v2/core/action.py
```diff
@@ -0,0 +1,290 @@
...
+        title="The subtype of the action.",
+        max_length=STR_FIELD_MAX_LENGTH,
+    )
+    configuration: Dict[str, Any] = Field(
+        title="The configuration for the action.",
+    )
+    service_account_id: UUID = Field(
+        title="The service account that is used to execute the action.",
+    )
+    auth_window: Optional[int] = Field(
```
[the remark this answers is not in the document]
stefannica: We talked about this and decided that the service account should be as close to the action as possible, because the action implementation is the component that actually uses it. The trigger should be lightweight because it keeps the concerns of the event hub limited to relaying and filtering events.

src/zenml/zen_server/routers/triggers_endpoints.py
```diff
@@ -81,54 +80,10 @@ def list_triggers(
     Returns:
         All triggers.
     """
-
```
stefannica, line 84: This was a way to interpose the action handler between trigger REST API calls and the database as a sort of admission controller. Now that the actions are detached, there's no immediate need for this, but in the future, when we have a proper event hub implementation, the event hub will basically have to play that same role: intervene in every trigger REST API call and act as an admission controller. In other words the final event hub class should implement `get_trigger`, `create_trigger`, `update_trigger` and `delete_trigger` and basically talk to whatever back-end implements the message hub to keep it updated with these events.

stefannica (approved):
Looks great, thank you for taking this off my hands!

src/zenml/models/v2/core/action.py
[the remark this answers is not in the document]
AlexejPenner: Alright, I will defer this to you then

src/zenml/zen_stores/sql_zen_store.py
schustmi: No strong opinion here, maybe @stefannica had something concrete in mind when he implemented it that way?

PR #2703, opened 2024-05-14
Add `model_version_id` as part of the Model config

avishniakov:
## Describe changes
I added `model_version_id` as a public field of the Model. It can be used to configure a Model Version using ID or will be used to represent a more stable way of working with pipelines under Model Version, for example, if the pipeline was started with the version set to some stage - the same model ID will be used in all steps using this configuration, regardless of the fact, that some other model version could have been moved to that stage during the pipeline run.
I added a specific test case for the following case:
    - Pipeline starts with Model `foo` stage `staging`
    - Do some steps
    - Some step moves another model version to `staging`
    - Validate that subsequent steps can still work with the initial model version

@Cahllagerfeld @atzin-escandia , FYI, now the Step or Model config will have a field `model` as before, but now it is enriched with `model_version_id`. The logic remains the same: check StepRun config for the model, if not null pick `id_` and leave, if null check PipelineRun config for the model, if not null pick `model_version_id` and leave.
 
## Pre-requisites

## Types of changes


src/zenml/model/model.py
```diff
@@ -73,12 +73,12 @@ class Model(BaseModel):
     tags: Optional[List[str]] = None
     version: Optional[Union[ModelStages, int, str]] = None
     save_models_to_registry: bool = True
+    id_: Optional[UUID] = None
```
bcdurak, line 76: The changes look good! However, when I first started the review, the naming tricked me a bit. Since this is the `Model` class, I expected the id to be associated with a `ModelResponse` rather than a `ModelVersionResponse`. Since we already have the `model_id`, if it is possible, I would recommend changing this to a `model_version_id`.

bcdurak (approved):
Left a small comment regarding the naming, otherwise, the changes look good.

src/zenml/model/model.py
avishniakov: Since it is still meant to be local only, I don't care much. This class specifically was renamed into `Model` in the course of concepts flattening some time back. I'll rename it.

avishniakov:
Sorry for retagging you @bcdurak , but while fixing a tiny failure in one of the tests I discovered biggerer problems, which worth another pair of eyes: https://github.com/zenml-io/zenml/pull/2703/commits/d04f7b702c7abd06cd0a56f7ceaed75e7b42595b

bcdurak (approved):

PR #2705, opened 2024-05-14
Add how-to section & restructure/update documentation

AlexejPenner:
WIP

https://zenml-io.gitbook.io/felixs-fecilitations/uiD80kmXqxCUBvbvW7BD/ for Gitbook preview


docs/book/how-to/overview/README.md
```diff
@@ -0,0 +1,35 @@
...
+    labels = [0, 1, 0]
+    return {'features': training_data, 'labels': labels}
+
+@step
+def train_model(data: dict) -> None:
+    total_features = sum(map(sum, data['features']))
+    total_labels = sum(data['labels'])
+    
+    # Train some model here
+
```
htahir1, line 22:
```suggestion
```

docs/book/how-to/overview/use-pipeline-step-parameters.md
```diff
@@ -0,0 +1,110 @@
...
+```
+{% endhint %}
+
+**Parameters and caching**
+
+When an input is passed as a parameter, the step will only be cached if all parameter values are exactly the same as for previous executions of the step.
+
+**Artifacts and caching**
+
+When an artifact is used as a step function input, the step will only be cached if all the artifacts are exactly the same as for previous executions of the step. This means that if any of the upstream steps that produce the input artifacts for a step were not cached, the step itself will always be executed.
```
htahir1, line 110: We can link to caching here?

docs/book/how-to/overview/use-pipeline-step-parameters.md
```diff
@@ -0,0 +1,110 @@
...
+    my_step(input_1=42, input_2=43)
+
+if __name__=="__main__":
+    # here an error will be raised since `some_param` is
+    # `24` in config, but `23` was provided in the code
+    my_pipeline(23)
+```
+{% endhint %}
+
+**Parameters and caching**
```
htahir1, line 104: Maybe we can nest these two in a ## Pipeline and caching section?

docs/book/how-to/overview/control-caching-behavior.md
htahir1, line 1: I think whats missing here is a small paragraph on the perils of using an external API within a step like loading data and expecting the cache to work magically. We should tell users that if you have a data loading step e.g. they should explicitly set it to false

Also, I think changing configuration like the stack or Model also invalidates the cache, so we should talk about that. We can also put a screenshot of the DAG visualizer to talk about what a cached step looks like

docs/book/how-to/overview/step-output-typing-and-annotation.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+# Not allowed: Variable length tuple annotation when using
+# multiple output artifacts
+@step
+def my_step() -> Tuple[int, ...]:
+    return 0, 1
+```
+{% endhint %}
+
```
htahir1, line 81: maybe an image of the DAG vsiualizer here showing artifact names and types might make sense?

docs/book/how-to/overview/trigger-a-pipeline-from-another.md
htahir1, line 1: First, this isnt really triggering a pipeline from another, this is a "composing" pipelines. But I like the title because its more relatable. But you can clarify this in the first paragraph

"In ZenML, you can call a pipeline from another pipeline, which ZenML interprets to compose the steps of the two pipelines together in one big pipeline. Note that the resulting pipeline would be counted as one pipeline and not two.

Second, We need to write more here. Like the limitations and use-cases where this makes sense

docs/book/how-to/overview/control-execution-order-of-steps.md
htahir1, line 1: This section needs sub-headings to make the content easily readable

docs/book/how-to/overview/using-a-custom-step-invocation-id.md
htahir1, line 1: I think we're missing the fact that the invocation ID is most of the time the same as the step name, but in some cases can be different. I'd give an example here of a for loop, and how the invocation ID gets into play here

docs/book/how-to/overview/name-your-pipeline-and-runs.md
htahir1, line 1: This is also missing some sort of sub-section (h2 headers) to make this easier to parse and read

docs/book/how-to/overview/hyper-parameter-tuning.md
htahir1, line 1: Lets take out the reference to our roadmap here because its an unfulfilled promise and it isn't true anymore. Lets just say this is not the best way to do it and maybe using Optuna within a step might make more sense for some use-cases

docs/book/how-to/overview/fetch-metadata-within-steps.md
htahir1, line 1: My suggestion: Let's move the metadata subject into its own section and call it "Tracking metrics and metadata" and talk there about the metadata topic because its also a new value proposition

htahir1: We can leave some in the indiivudla sections though like this chapter is fine

docs/book/how-to/overview/enable-or-disable-logs-storing.md
htahir1, line 1: Similar to metadata, I feel like logs deserve their own section, as there are so many topics about it . Of the top of my head:

* Enabling/disabling logs
* Setting up the server in a way you can see logs in the dashboard
* Which logs are tracked and how?
* Viewing local logs (there will be a new `zenml pipeline logs..` command that might fit here

docs/book/deploying-zenml/zenml-cloud/README.md
htahir1, line 1: this needs to change to the new Pro naming (here and elsewhere in the docs)

htahir1 (commented):
This is a tremendous effort! I'm going to stop here mid-way to make this feedback more digestible. happy to pair and split up the work because my suggestions are quite heavy sometimes.

# Things I noticed are missing

* How do I debug errors when a pipeline fails, especially remotely?  What logs are available to me?
* How do I run a pipeline "async"?
* How to use my YAML file to pass same parameters to all steps withour repeating?
* Visualziations and how to use them properly?
* Questions regarding configuring stack components like : do all the stack components need to be connected to the same service connector? and "How do I add some settings like GPUs or CPUs for Sagemaker Orchestrator? How do I configure settings for Kubernetes Orchestrator? For the VertexOrchestrator?"

# Specific 

* I feel like "Fetching pipelines" can now go into the "Build a pipeline" How-to section instead. I always though it did not really fit in the Starter guide anyway!
* In so many sections in the How-to, almost every section could use a visual screen of the DAG or some CLI commands. I do miss that... should we do a sweep and try to find places to put? Ill comment where I think we can add screenshots or CLI commands

# General 

I have an idea for the docs which I've recorded on Loom:

https://www.loom.com/share/1cfd061b610845358bcaae6981fc481b

strickvl:
Some general comments from me. I really like the way this is going. 

I think maybe the how-to sections could be improved by having a final h3 'see also' section which is basically a bullet point list of 2 or 3 other pages (could be to a concept or a page or even a subsection of a page (pref not, since the link will break :) ) AND (important to have this 'and' here) a brief explanation (1 sentence / phrase) of what this link is and why someone browsing for 'how to use service connectors' e.g. might also want to know about x and y. I think this helps the how-to guides work as an ecosystem and not just 100 islands.

I like @htahir1's sort of idea in the loom about breaking open the stacks and components section. Personally I'd like to see the component guides with their own section at the top level. They're so buried and they are so valuable + well used by users...

Finally, I think neither you nor @htahir1 saw much value in having an overview page, but I just want to take a final moment to advocate for a top level page in the how-to section which allows people to navigate all the pages in a single view (without having to click sections to unfold them etc), and where they can do a quick text search and also just have it open as the reference point etc. Not sure how to enable that with Gitbook, but I think it's important.

bcdurak (changes requested):
please no

bcdurak (approved):

PR #2707, opened 2024-05-15
Don't use pod resources/affinity for kubernetes orchestrator pod

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
@@ -395,10 +395,15 @@ class KubernetesOrchestrator(ContainerizedOrchestrator):
             # Remove all settings that specify on which pod to run for the
             # orchestrator pod. These settings should only be used
             # for the pods executing the actual steps.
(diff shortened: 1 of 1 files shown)


AlexejPenner (commented):
Can any of these values be set from the kubernetes orchestrator configuration? IF so are those removed through this change?

schustmi:
The settings that get fetched (and that I override) already include the configuration of the stack component as well as any runtime settings that were provided as an override.

AlexejPenner (approved):
:seal:

dragosmc:
Thanks for this feature as it makes sense not to use a huge instance that a step requires just to run the orchestrator. With this change, though, what's the way to specify pod_settings for the orchestrator? We still need to force it to run on specific nodes.

AlexejPenner:
> Thanks for this feature as it makes sense not to use a huge instance that a step requires just to run the orchestrator. With this change, though, what's the way to specify pod_settings for the orchestrator? We still need to force it to run on specific nodes.

Thank you for raisin g your concern here, we've come up with a new way to do just that [here](https://github.com/zenml-io/zenml/pull/2732)

PR #2708, opened 2024-05-15
Extract project ID from GCP service connector SA credentials

stefannica:
## Describe changes
GCP service accounts are already scoped to a GCP project and the project ID is part of the JSON credentials. We therefore don't need to provide the project ID as a mandatory part of the service connector configuration.

## Pre-requisites

## Types of changes


htahir1 (changes requested):
Thanks for this but but two points:

* Are we sure there is not occasions where a service account for one project is given access to other projects (I am not aware of any but just want to make sure)
* If the project is always in cfg.service_account_json and it always has to be the same as what the user says, then why do we ask the user to input it in the first place? Why cant we just take it from there?
* As mentioned internally, please do add more logs around establishing a connection to debug what happens

stefannica:
> Thanks for this but but two points:
> 
> * Are we sure there is not occasions where a service account for one project is given access to other projects (I am not aware of any but just want to make sure)

Yes, 100% sure. This is a misconfiguration and should be avoided.

> * If the project is always in cfg.service_account_json and it always has to be the same as what the user says, then why do we ask the user to input it in the first place? Why cant we just take it from there?

This is a fair question. I'll look into it.

> * As mentioned internally, please do add more logs around establishing a connection to debug what happens

Unfortunately, the connection establishment is located deep inside the GCP Cloud Python library where I have no granular control, but I'll try to enable more verbose logs for those libraries overall.

htahir1:
@stefannica feel free to merge this and make a follow up ticket for the rest of the work here ahead of the release tomorrow

htahir1 (approved):

stefannica:
@htahir1 I modified the PR do automatically extract the GCP project ID from the service account credentials. This way, the user doesn't have to explicitly provide the project ID in addition to the service account JSON.

I also tried to enable debug logs for the GCP libraries, but nothing I tried worked. Either they don't log any debug information or they use some other way of enabling them which I couldn't find documented anywhere.

htahir1 (approved):

PR #2725, opened 2024-05-24
Handle error when using `zenml info` with missing dependencies

strickvl:
When using `zenml info -s`, a `ModelNotFoundError` gets raised if your environment doesn't have all the necessary dependencies installed for the components in your stack. 

This is raised from the `Client().active_stack` property which in turn calls `Stack.from_model(self.active_stack_model)` which is where the error is raised.

This PR handles the error gracefully to log an error message instead of completely failing.

### DIFF
--- a/src/zenml/cli/base.py
+++ b/src/zenml/cli/base.py
@@ -653,7 +653,14 @@ def info(
         cli_utils.print_user_info(user_info)
 
     if stack:
-        cli_utils.print_debug_stack()
+        try:
+            cli_utils.print_debug_stack()
+        except ModuleNotFoundError:
+            cli_utils.warning(
+                "Could not print debug stack information. Please make sure "
+                "you have the necessary dependencies and integrations "
+                "installed for all your stack components."
+            )
 
 
 @cli.command(
--- a/src/zenml/cli/base.py
+++ b/src/zenml/cli/base.py
@@ -655,12 +655,13 @@ def info(
     if stack:
         try:
             cli_utils.print_debug_stack()
-        except ModuleNotFoundError:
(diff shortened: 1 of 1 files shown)


src/zenml/cli/base.py
```diff
@@ -653,7 +653,14 @@ def info(
...
     if stack:
-        cli_utils.print_debug_stack()
+        try:
+            cli_utils.print_debug_stack()
+        except ModuleNotFoundError:
+            cli_utils.warning(
+                "Could not print debug stack information. Please make sure "
+                "you have the necessary dependencies and integrations "
+                "installed for all your stack components."
+            )
```
avishniakov, line 663: Maybe partially print error info here, e.g. which module is missing to give some hint? Or, if possible, list the flavors set of this stack.

strickvl: Yeah ok that's a good idea.

src/zenml/cli/base.py
```diff
@@ -653,7 +653,14 @@ def info(
         cli_utils.print_user_info(user_info)
 
     if stack:
-        cli_utils.print_debug_stack()
+        try:
+            cli_utils.print_debug_stack()
```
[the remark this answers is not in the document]
strickvl: Perhaps, but I recall this being fairly fiddly when I originally implemented it
and don't necessarily want to get to stuck into this. I'll take a look, though.

src/zenml/cli/base.py
strickvl: Added this now.

schustmi (approved):

avishniakov (approved):

PR #2731, opened 2024-05-28
Step logs pagination

schustmi:
## Describe changes
This PR adds `offset` and `length` parameters to the step logs endpoint to avoid loading the entire file in memory.

## Pre-requisites

## Types of changes


strickvl:
another FYI @avishniakov since you're working on logging things at the moment

src/zenml/artifacts/utils.py
```diff
@@ -763,7 +767,19 @@ def _load_file_from_artifact_store(
...
+                try:
+                    text_file.seek(offset, os.SEEK_END)
+                except OSError:
+                    # The negative offset was too large for the file, we seek
+                    # to the start of the file
+                    text_file.seek(0, os.SEEK_SET)
+            elif offset > 0:
+                text_file.seek(offset, os.SEEK_SET)
+
+            return text_file.read(length)
```
stefannica, line 782: This method is also used to load visualization files in the dashboard. You should test that still works properly with this (i.e. not sure what happens if length is None here)

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -265,6 +269,10 @@ def get_step_logs(
...
     return str(
         _load_file_from_artifact_store(
-            logs.uri, artifact_store=artifact_store, mode="r"
-        )
+            logs.uri,
+            artifact_store=artifact_store,
+            mode="rb",
+            offset=offset,
+            length=length,
+        ).decode()
```
stefannica, line 277: why did you add the extra `decode` here ?

stefannica (approved):
This looks like a good quick solution for the unbound logs loading problems we've been having. It also requires some UI support to ensure proper pagination but also some buffering and pre-caching (e.g. you wouldn't want to call this endpoint for every line of log being scrolled).

src/zenml/zen_server/routers/steps_endpoints.py
schustmi: I had to change the `mode` to binary reads, as the seeking is only available for the cloud filesystems when using this mode.

src/zenml/artifacts/utils.py
schustmi: `read(None)` should be equivalent to just `read()` which was the previous behaviour

avishniakov (approved):
These changes shall land quite smoothly on the merged logs file, while the performance of multiple files logging might be not improved.

PR #2752, opened 2024-06-05
Server scalability improvements

stefannica:
## Describe changes

This PR brings a number of improvements meant to make the ZenML server more scalable and resilient under load:

* make the health check endpoint async. This makes the health check responsive even when all the worker threads are otherwise occupied with other requests. It prevents Kubernetes from killing and restarting ZenML server pods when they are under sustained load.
* configurable thread pool size (default is 40, same as FastAPI's). In addition to Kubernetes resources, auto-scaling and replica count options, this can be used to tweak the ZenML server configuration to achieve an optimal resource use.
* sync the cloud API requests connection pool to the server's thread pool size. The previous configuration used the default value of 10 connections which is misaligned with the number of worker threads in the ZenML Pro tenants and therefore leading to warnings such as:

```
Connection pool is full, discarding connection: cloudapi.zenml.io. Connection pool size: 10
```

* make all dashboard endpoints async as well. Given that these endpoints do not make any blocking I/O calls, they can also be implemented as asyncio coroutines which means then will no longer be competing over the limited amount of worker threads. This further improves the ZenML server responsiveness under load.

## Pre-requisites

## Types of changes


strickvl:
Just adding @avishniakov as a reviewer on the PR as it feels like OSS team should have a sense of these changes

src/zenml/zen_server/cloud_utils.py
```diff
@@ -154,12 +155,29 @@ def session(self) -> requests.Session:
...
             retries = Retry(total=5, backoff_factor=0.1)
-            self._session.mount("https://", HTTPAdapter(max_retries=retries))
+            self._session.mount(
+                "https://",
+                HTTPAdapter(
+                    max_retries=retries,
+                    # We only use one connection pool to be cached because we
+                    # only communicate with one remote server (the control
+                    # plane)
+                    pool_connections=1,
```
wjayesh, line 177: For my understanding, is it costly to have more cached hosts? I understand we need only one but want to know what impact this would have.

src/zenml/zen_server/deploy/helm/templates/server-secret.yaml
```diff
@@ -21,6 +21,12 @@ data:
   {{- if .Values.zenml.database.sslKey }}
   ZENML_STORE_SSL_KEY: {{ .Files.Get .Values.zenml.database.sslKey | b64enc }}
   {{- end }}
+  {{- if .Values.zenml.database.poolSize }}
+  ZENML_STORE_POOL_SIZE: {{ .Values.zenml.database.poolSize | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.zenml.database.maxOverflow }}
+  ZENML_STORE_POOL_SIZE: {{ .Values.zenml.database.maxOverflow | b64enc | quote }}
```
wjayesh, line 28: Duplicate key?

wjayesh (commented):
Looks good to me save for a few comments.

src/zenml/zen_server/zen_server_api.py
```diff
@@ -11,18 +11,28 @@
...
+from time import sleep
 from typing import Any, List
 
+from anyio import to_thread
 from fastapi import FastAPI, HTTPException, Request
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import ORJSONResponse
 from fastapi.staticfiles import StaticFiles
 from fastapi.templating import Jinja2Templates
+from py import log
```
schustmi, line 35: This and the `sleep` import seem unused?

schustmi (approved):

src/zenml/zen_server/cloud_utils.py
stefannica: There is no impact in this case. We can set this to 10 or 100, there will only be one connection pool because we only need to connect to one host. This is better covered here, if you want the gruesome details: https://stackoverflow.com/questions/34837026/whats-the-meaning-of-pool-connections-in-requests-adapters-httpadapter

wjayesh: Really well explained answer; had this in bookmarks for a while ;) thanks!

src/zenml/zen_server/deploy/helm/templates/server-secret.yaml
stefannica: good catch ! thank you !

src/zenml/zen_server/zen_server_api.py
stefannica: Correct, I removed them. The linter was also flagging them.

PR #2769, opened 2024-06-11
Fix nightly Docker build

strickvl:
Addresses this failure: https://github.com/zenml-io/zenml/actions/runs/9457690125/job/26061375888

Fixed the logging of which version was being used.

Updated the branch used to determine which version we're currently on.

It's possible I still didn't get to the root cause here. Might need some more investigation.


safoinme (approved):

avishniakov (approved):
Can we test it somehow before merging?

avishniakov (approved):

PR #2772, opened 2024-06-12
Adjust docs url for flavors

AlexejPenner:
## Describe changes
Stack component flavors have a "docs_url" property which is used in the frontend and CLI to point users in the right direction to get more context. With the recent changes to legacy docs, the package needs to return a different URL for the latest zenml version vs. all older versions.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/package_utils.py
+++ b/src/zenml/utils/package_utils.py
@@ -12,6 +12,7 @@
 #  or implied. See the License for the specific language governing
 #  permissions and limitations under the License.
 """Utility functions for the package."""
+
 import requests
 from packaging import version
 
@@ -41,7 +42,9 @@ def is_latest_zenml_version() -> bool:
         )
 
     # Compare versions
-    if version.parse(latest_published_version) > version.parse(current_local_version):
+    if version.parse(latest_published_version) > version.parse(
+        current_local_version
+    ):
         return False
     else:
         return True
--- a/src/zenml/utils/package_utils.py
+++ b/src/zenml/utils/package_utils.py
@@ -28,20 +28,20 @@ def is_latest_zenml_version() -> bool:
     from zenml import __version__
 
     # Get the current version of the package
-    current_version = __version__
+    current_local_version = __version__
 
     # Get the latest version from PyPI
     try:
         response = requests.get("https://pypi.org/pypi/zenml/json")
         response.raise_for_status()
-        latest_version = response.json()["info"]["version"]
+        latest_published_version = response.json()["info"]["version"]
     except Exception as e:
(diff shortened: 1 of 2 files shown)


strickvl:
This will be released in a subsequent version of ZenML, so when would there be a case for it to point to an older version of the docs? Just trying to understand why there are options for the docs URL now? @AlexejPenner

AlexejPenner:
> This will be released in a subsequent version of ZenML, so when would there be a case for it to point to an older version of the docs? Just trying to understand why there are options for the docs URL now? @AlexejPenner

Ok lets say we are at 0.58.2 and upgrade to 0.58.3. Part of the release flow is now to deprecate the 0.58.2 version of the docs and move it to the legacy docs domain. When somebody now runs their old 0.58.2 code and the docs_url property is loaded the local cove version 0.58.2 is no longer the latest (0.58.3) and so the legacy docs url is returned.

strickvl:
Wouldn't the person on 0.58.3 just get the new docs url which we set in code? Why is it getting an old docs url version? I might just be misunderstanding something really simple here, so feel free to ignore me.

AlexejPenner:
> Wouldn't the person on 0.58.3 just get the new docs url which we set in code? Why is it getting an old docs url version? I might just be misunderstanding something really simple here, so feel free to ignore me.

This is for people still running on an old version of zenml. Another example think 5 releases into the future. The code from this PR will be in as from 0.58.3 - if somebody does not upgrade to 0.65.9, their stack components might not have the latest and greatest attributes that were introduced up to this potential new latest version. So when they look at the docs for their stack components, we point to the docs at the right version for them

src/zenml/utils/package_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    Returns:
+        True in case the current running zenml code is the latest available version on PYPI, otherwise False.
+
+    Raises:
+        RuntimeError: In case something goe wrong
+    """
+    from zenml import __version__
+
+    # Get the current version of the package
+    current_version = __version__
```
strickvl, line 31: this name feels like it might be updated? current_local_version or just local_version?

src/zenml/utils/package_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Get the latest version from PyPI
+    try:
+        response = requests.get("https://pypi.org/pypi/zenml/json")
+        response.raise_for_status()
+        latest_version = response.json()["info"]["version"]
```
strickvl, line 37: Also here, latest_published_version or something? or latest_pypi_version

PR #2774, opened 2024-06-12
Add installation instructions for Macs running on Apple Silicon

strickvl:
Adds information on installation for Mac users running on M1 / M2 systems (i.e. Apple Silicon). (Also adds a note about Windows support to FAQ).

### DIFF
--- a/scripts/install-zenml-dev.sh
+++ b/scripts/install-zenml-dev.sh
@@ -31,7 +31,7 @@ install_zenml() {
     uv pip install $PIP_ARGS -e ".[server,templates,terraform,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,dev]"
 
     # TODO: Remove. Temporary installation from the upgrade branch.
-    uv pip install $PIP_ARGS git+https://github.com/zenml-io/mlstacks.git@feature/upgrade-to-pydantic-v2
+    uv pip install $PIP_ARGS git+https://github.com/zenml-io/mlstacks.git@develop
 }
 
 install_integrations() {


docs/book/getting-started/installation.md
```diff
@@ -26,6 +26,18 @@ pip install "zenml[server]"
 We highly encourage you to install ZenML in a virtual environment. At ZenML, We like to use [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/) or [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) to manage our Python virtual environments.
 {% endhint %}
 
+## Installing onto MacOS with Apple Silicon (M1, M2)
```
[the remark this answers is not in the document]
schustmi: Yes, although that seems very unlikely? For the most common case (a MacOS client) this is not needed though.

docs/book/getting-started/installation.md
```diff
@@ -26,6 +26,18 @@ pip install "zenml[server]"
...
+
+A change in how forking works on Macs running on Apple Silicon means that you
+should set the following environment variable which will ensure that your
+connections to the server remain unbroken:
+
+```bash
+export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES
+```
+
+You can read more about this [here](http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html).
```
strickvl, line 39:
```suggestion
You can read more about this [here](http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html). This environment variable is needed if you are working with a local server on your Mac, but if you're just using ZenML as a client / CLI and connecting to a deployed server then you don't need to set it.
```

schustmi (approved):
LGTM. I personally would probably not read the docs before installing/trying, so I have an additional suggestion:

When someone is running `zenml up`, we could detect that they're on macos and then check whether this environment variable is set, and if not warn them?

Or maybe point them towards a page to read up on the environment variable? At least when I read it initially, I remember reading that setting this can have some negative effect as well, might be worth also linking in the docs. (Quite old info though, haven't done any recent research). If it's unproblematic, we could even set it ourselves when spinning up a server

AlexejPenner:
> LGTM. I personally would probably not read the docs before installing/trying, so I have an additional suggestion:
> 
> When someone is running `zenml up`, we could detect that they're on macos and then check whether this environment variable is set, and if not warn them?
> 
> Or maybe point them towards a page to read up on the environment variable? At least when I read it initially, I remember reading that setting this can have some negative effect as well, might be worth also linking in the docs. (Quite old info though, haven't done any recent research). If it's unproblematic, we could even set it ourselves when spinning up a server

Also think its great to have this here, but it should be a warning in the code as well in case we don't have that yet.

AlexejPenner (approved):
:seal:

strickvl:
Not sure if it was too much to just error out of the process, but it felt better than just to emit a warning message...

src/zenml/cli/utils.py
```diff
@@ -2748,3 +2748,19 @@ def is_jupyter_installed() -> bool:
...
+    This is for the case where a user is on a MacOS system, trying to run a
+    local server but is missing the `OBJC_DISABLE_INITIALIZE_FORK_SAFETY`
+    environment variable.
+
+    Returns:
+        bool: True if a warning needs to be shown, False otherwise.
+    """
+    return (
+        not os.getenv("OBJC_DISABLE_INITIALIZE_FORK_SAFETY")
+        and sys.platform == "darwin"
```
schustmi, line 2765: I think this env variable is only necessary on M1 macs, can we detect that somehow?

schustmi: Actually this is wrong, but it depends on the system version which we could detect I assume?

strickvl: What do you mean system version? like the operating system version? Sonoma vs snow leopard etc? We can detect that with:

```python
import platform

# Get the macOS version
macos_version = platform.mac_ver()[0]
```

I didn't realise that this was only a thing above a certain version. I'll look into when that was introduced.

schustmi: Yes, the link you mention in the docs explains that this is relevant for macos 10.13+ only

AlexejPenner (approved):
:seal:

schustmi (approved):

PR #2780, opened 2024-06-13
Add `--ignore-errors` flag for `zenml artifact prune`

strickvl:
We had [a report on Slack](https://zenml.slack.com/archives/C01FWQ5D0TT/p1718210372292429?thread_ts=1718208717.747519&cid=C01FWQ5D0TT) of a user who was unable to use the `zenml artifact prune` subcommand because of the errors that were being thrown. I added a flag to allow the user to bypass these errors and just to delete the artifacts that are able to be deleted. Regardless of flag or not, the errors are still output to the terminal.


src/zenml/cli/artifact.py
```diff
@@ -229,8 +229,17 @@ def update_artifact_version(
     is_flag=True,
     help="Don't ask for confirmation.",
 )
+@click.opion(
```
avishniakov, line 232:
```suggestion
@click.option(
```

avishniakov (changes requested):

avishniakov (approved):

schustmi (approved):

PR #2791, opened 2024-06-18
TF tests + various integration (un)install improvements

avishniakov:
## Describe changes
Not an actual fix for 3.8 and TF, but:
- I allowed all runtimes for `tests/integration/examples/test_tensorflow.py` except Ubuntu/3.8 only (before was banned for 3.8,3.10 and 3.11 on all OS)
- Allowed `tests/integration/examples/test_huggingface.py::test_sequence_classification` for 3.8, since it works fine
- Allowed `tests/integration/examples/test_huggingface.py:: test_token_classification ` for all versions/OS, since it works fine, but there was a bug and this test was completely skipped
- Throw a warning for users on 3.8 using Tensorflow integration

Basically, the fix is simple - go above 3.8.

**Completely Side-Effect** produced by review comments:
- I added a new `REQUIREMENTS_IGNORED_ON_UNINSTALL` integrations class setting, so we can skip some requirements on uninstall, like `pydantic` in `mlflow`, `typing-extensions` in `tensorflow` and so on. To ensure that we don't break other integrations by uninstalling others. Unfortunately, it will mean that some will be not uninstallable, say `Kubernetes` is used in many places, but none is uninstalling it in new behavior to protect others. *Sahll we make it better and lookup for other installed integrations for intersections and skip ignore, in case no harm to others, or this is good enough?* I personally tend to the latest.

## Pre-requisites

## Types of changes


.github/workflows/integration-test-slow.yml
strickvl, line 1: haha what even was this?

src/zenml/integrations/tensorflow/__init__.py
```diff
@@ -31,14 +34,20 @@ def activate(cls) -> None:
...
             import tensorflow_io  # type: ignore
 
         from zenml.integrations.tensorflow import materializers  # noqa
 
+        if sys.version_info.minor == 8:
+            logger.warning(
+                "Python 3.8 with TensorFlow is not fully "
+                "compatible with Pydantic 2 requirements. "
+                "Consider upgrading to a higher Python "
+                "version, if you would like to use "
```
strickvl, line 47:
```suggestion
                "version if you would like to use the "
```

strickvl: I was wondering if we have this error message when someone does `zenml integration install tensorflow` (and/or).

strickvl (changes requested):

.github/workflows/integration-test-slow.yml
avishniakov: Left over 🙂

src/zenml/integrations/tensorflow/__init__.py
avishniakov: Not really, if I plug it into requirements it is too annoying since requirements are being read in many places. This one will pop-up on the run of a pipeline. Do you think I need it in the CLI command as well?

strickvl (approved):

src/zenml/cli/integration.py
```diff
@@ -279,6 +272,22 @@ def install(
                     f"{list(integration_registry.integrations.keys())}"
                 )
 
+    # TODO: remove once python 3.11 gcp integration issue is resolved
+    if sys.version_info >= (3, 11) and "gcp" in integrations:
```
bcdurak, line 276: Since you moved this warning, it caught my attention. I have tried it again after the pydantic upgrade and I believe this issue is resolved now and we can remove this warning. What do you think @strickvl?

src/zenml/cli/integration.py
```diff
@@ -359,11 +368,27 @@ def uninstall(
     for integration_name in integrations:
         try:
             if integration_registry.is_installed(integration_name):
-                requirements += (
-                    integration_registry.select_integration_requirements(
-                        integration_name
+                # TODO: keep only `else` branch once 3.8 is deprecated
+                if (
```
bcdurak, line 372: This is a quite interesting catch. As far as I understand, we have to keep the `typing-extensions` even if we uninstall the `tensorflow` integration over the CLI otherwise it will lead to issues.

I can think of another example that might need to adjust similarly to the `uninstall` command. `mlflow` now has `pydantic` V2 as a requirement, otherwise when installed through the CLI, it downgrades it to V1. So, in its current state, if I uninstalled the `mlflow` integration through the CLI, it will remove pydantic altogether 🥲 Would be amazing to fix that here as well.

bcdurak (commented):
Great changes. Even after the pydantic upgrade, these tests were still on my mind. Thank you for working on them.

I left a few comments, but feel free to merge.

src/zenml/cli/integration.py
strickvl: Yes I think we can remove it. And from the docs page for GCP (can't remember exactly which one, but it's in there in at least one page) as well, I think.

avishniakov: I generalized the concept and added more ignores by going through the integration list. I will re-request reviews, since it is a bold move 🙂

avishniakov: *GCP warning is now removed from all the palces.

src/zenml/cli/integration.py
avishniakov: I generalized the concept and added more ignores by going through the integration list. I will re-request reviews, since it is a bold move 🙂

strickvl (approved):

bcdurak (approved):
The new changes look great!

PR #2795, opened 2024-06-20
Merged log files in Step Ops steps might be not available on main process, due to merge in the step op

avishniakov:
## Describe changes
I fixed the max recursion issue with step-op controlled steps on the main process.
Additionally, I simplified the merge function to avoid such issues in general.
There is one **BIG ASSUMPTION** here: time on client and step op should be quite in sync, otherwise the order of the log might go crazy. Generally, this is not a super critical thing, since in the worst case it can plug all client logs to the end or to the beginning of the log file, so the major logs from step op will flow in chronological order.

@bcdurak this one also should solve the retry missing logs thingy.

## Pre-requisites

## Types of changes


schustmi (commented):
This changes the logic of the merging right: Previously we only merged the last X log files, which means we would not end up reading/writing the full logs every time we merge. Unless I'm mistaken, this is now different and we always
- load all previous log files
- write them to a single merged file

For long-running steps, won't this again be an issue that it reads and writes potentially many megabytes of logs every few seconds?

avishniakov:
> This changes the logic of the merging right: Previously we only merged the last X log files, which means we would not end up reading/writing the full logs every time we merge. Unless I'm mistaken, this is now different and we always
> 
> * load all previous log files
> * write them to a single merged file
> 
> For long-running steps, won't this again be an issue that it reads and writes potentially many megabytes of logs every few seconds?

Valid concerns, thanks! I reworked this to be less heavy for the execution:
- During the step execution the roll-up would be happening from time to time for long-running jobs, but only rolling up the most recent files, keeping ones with the `_merged` suffix behind
- On step exit, we will try to roll up all files into one big one for faster lookup in the future, but will suppress the errors during final roll-up, since it is non-critical, but might be failing on very big files indeed.

schustmi (approved):

PR #2802, opened 2024-06-24
Fix S3 artifact store memory leak and other improvements

stefannica:
## Describe changes
This PR fixes a memory leak was observed while using the ZenML dashboard to view the logs of a pipeline or the visualization of an artifact that logged through an S3 Artifact Store linked to an AWS Service Connector.

### How to reproduce

* configure an AWS Service Connector that uses the IAM role authentication method.
* configure an S3 artifact store and link it to an S3 bucket through the service connector
* configure a stack with the S3 artifact store and the default orchestrator
* run a pipeline, preferably one that has rich visualizations, like the evidently integration test example (you also need the evidently data validator in the stack for that)
* visualize the artifacts or the logs of the pipeline in the dashboard. The memory usage will keep going up.

NOTE: even with this patch applied, the memory usage may still go up until it reaches a stable value. This depends largely on the number of threads that the server is allowed to create (default: 40). It is recommended to configure the number of threads to something low like 2 or 3 (by using the `ZENML_SERVER_THREAD_POOL_SIZE` environment variable when running the server) to reach that memory usage plateau faster.

### Details

The investigation lead to the following findings:

1. the s3fs library used in the S3 Artifact Store caches all its `S3FileSystem` client instances permanently, based on the constructor input arguments. In a more dynamic and long-lived environment like the ZenML server, this means that a new `S3FileSystem` is created (and never de-allocated) for every new set of credentials. Given that the AWS Service Connector generates _temporary credentials_, the number of cached `S3FileSystem` clients slowly grows with every invocation of the pipeline logs or artifact visualization endpoint until the server runs out of memory.

Solution: `S3FileSystem` instance caching is a class-level setting. This PR disables instance caching by sub-classing `S3FileSystem` and using the sub-class in the ZenML S3 Artifact Store implementation:

```
class ZenMLS3Filesystem(s3fs.S3FileSystem):

    cachable = False
```

2. with the caching out of the picture, a second bug was revealed: `S3FileSystem` client instances were incorrectly de-allocated by the garbage collector, leading to additional memleaks. The exact root cause of this was not clearly identified even though it was fixed, but a likely explanation is included here.
 
`s3fs` is using asyncio at its core, but also provides support for synchronous calls by maintaining its own asyncio event loop thread and converting sync calls in asyncio tasks. `s3fs` also uses `aiobotocore` instead of `botocore` for that same reason and the aiobotocore clients are implemented using asynchronous context managers (more details on this below).

Normally, an async botocore client is used like this, where the client is created and deallocated in the same code block:

```
    session = aiobotocore.session.AioSession(**self.kwargs)
    async with self.session.create_client("s3", config=self.conf) as client:
        # do work with client
```

However, in the case of a class like `S3FileSystem`, where the client needs to be reused across different calls happening at different times,  the aiobotocore client context manager needs to be **entered** during initialization and **exited** during cleanup:

```
async def initialize_client(self):
    self.session = aiobotocore.session.AioSession(**self.kwargs)
    client = self.session.create_client(
                "s3", config=self.conf
    )
    self._s3 = await s3creator.__aenter__()

async def cleanup_client(self):
    await self._s3.__aexit__(None, None, None)
````

The problem comes from _when_ and _how_ the cleanup is called: it needs to run as an asyncio task and therefore it needs to be invoked at a time when an asyncio event loop still exists i.e. before the garbage collector deallocates the event loop resources. The cleanup might also need to be run using the same event loop as the initialization - which is the assumed source for the second memleak: he way `s3fs` was doing de-initialization was buggy in that it was not using the same event loop as the one used for initialization.

### Side-changes

Some more changes are included that help make the ZenML server leaner and more efficient concerning memory usage:

* add a `cleanup` method to the stack component base class to be called in the ZenML server to proactively deallocate the client when done working with it instead of waiting for the garbage collector
* stop registering the Artifact Store instances as I/O filesystems in the ZenML server. We don't use this feature in the ZenML server (we access artifacts through the artifact store directly) and it just keeps an artifact store instance constantly loaded in memory
* stop verifying service connectors again when they are used to generate credentials for their clients (stack components). This speeds up the artifact and logs visualization a bit. 

## Pre-requisites

## Types of changes

### CODE REVIEW

File: src/zenml/integrations/s3/artifact_stores/s3_artifact_store.py [hunk 1]
```diff
@@ -38,10 +39,77 @@
...
+    correct cleanup implementation by overriding the `close_session` method
+    the S3 aiobotocore client. The original one provided by s3fs was causing
+    memory leaks by creating a new event loop in the destructor instead of
+    using the existing one.
+
+    A `close` method is also provided to allow for synchronous on-demand cleanup
+    of the S3 client.
+    """
+
+    cachable = False
```
avishniakov, line 65: I would assume that it is a typo in s3fs, it should be `cacheable` normally.

schustmi (approved):

avishniakov (approved):
Nice catch!

htahir1 (changes requested):
Could you please put something in the docs about the `ZENML_SERVER_THREAD_POOL_SIZE ` :-) Ty!

PR #2808, opened 2024-06-26
Stack wizard CLI + Endpoints

avishniakov:
## Describe changes
With this PR we introduce:
- `zenml stack register --provider aws` - it will spin up the wizard to create a new stack with a cloud connector pointing to existing cloud infrastructure pieces. Users will be asked to select artifact store, orchestrator, and container registry. The only supported cloud provider in this PR is AWS.
- New endpoint to support the "full stack" creation, which takes `FullStackRequest` as input and creates all needed objects (service connectors, components, etc.)

## Pre-requisites

## Types of changes


src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -2166,6 +2167,24 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+
+        Returns:
+            The created stack.
+
+        Raises:
+            EntityExistsError: If a service connector with the same name
+                already exists.
+            StackComponentExistsError: If a stack component with the same name
+                already exists.
+            StackExistsError: If a stack with the same name already exists.
```
stefannica, line 2185:
```suggestion
```

No longer accurate.

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -349,6 +351,72 @@ def create_stack(
...
+            verify_permission_for_model(model=component, action=Action.READ)
+        else:
+            is_component_create_needed = True
+    if is_component_create_needed:
+        verify_permission(
+            resource_type=ResourceType.STACK_COMPONENT,
+            action=Action.CREATE,
+        )
+
+    verify_permission(resource_type=ResourceType.STACK, action=Action.CREATE)
```
stefannica, line 412: Beautiful !

stefannica (changes requested):
A couple of initial suggestions, still have to look at this in depth.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6919,6 +6920,230 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+                    self.delete_service_connector(
+                        service_connector_id=service_connector_id
+                    )
+                for component_id in components_created_ids:
+                    self.delete_stack_component(component_id=component_id)
+
+                raise RuntimeError(
+                    f"Full Stack creation has failed {e}. Cleaning up the "
+                    f"created entities."
+                )
```
stefannica, line 7144: The real error is completely hidden from the logs. You should use `raise from` or `logger.exception()` here at least if debug logs are enabled. For example, this is what I got:

```
2024-06-28 18:37:23,876 - zenml.zen_server.utils - ERROR - API error (utils.py:457)
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_stores/sql_zen_store.py", line 7015, in create_full_stack
    service_connector = service_connectors[
                        ^^^^^^^^^^^^^^^^^^^
IndexError: list index out of range

During handling of the above exception, another exception occurred:

  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_server/utils.py", line 447, in decorated
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_server/routers/workspaces_endpoints.py", line 417, in create_full_stack
    return zen_store().create_full_stack(full_stack)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_stores/sql_zen_store.py", line 7141, in create_full_stack
    raise RuntimeError(
RuntimeError: Full Stack creation has failed list index out of range. Cleaning up the created entities.
INFO:     3.78.248.73:0 - "POST /api/v1/workspaces/default/full-stack HTTP/1.1" 500 Internal Server Error
```

No idea where in the code this happened and which index is out of range.

stefannica (changes requested):

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6919,6 +6920,230 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+                                components_created_ids.append(component.id)
+                                break
+                            except EntityExistsError:
+                                component_name = f"{full_stack.name}-{random_str(4)}".lower()
+                                continue
+
+                        if component_info.service_connector_index is not None:
+                            service_connector = service_connectors[
+                                component_info.service_connector_index
+                            ]
```
stefannica, line 7017: The index might be out of bounds here. You should raise a user friendly error for that

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6919,6 +6920,230 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+                            if resources.resource_ids:
+                                if len(resources.resource_ids) > 1:
+                                    raise RuntimeError(
+                                        f"Multiple {resource_type} resources are "
+                                        "available for the selected connector. "
+                                        "Please use the a specific resource-id to "
+                                        f"configure a {resource_type} resource."
+                                    )
+                                else:
+                                    resource_id = resources.resource_ids[0]
```
stefannica, line 7096: Verification isn't necessary and you should avoid doing any type of verification on the service connector at this point. This should be exclusively about registration. Verification is time-consuming and will yield non-deterministic results.

The verification itself should be done before this endpoint is called.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6919,6 +6920,230 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+
+                handler.metadata.update(
+                    stack_response.get_analytics_metadata()
+                )
+            except Exception as e:
+                for service_connector_id in service_connectors_created_ids:
+                    self.delete_service_connector(
+                        service_connector_id=service_connector_id
+                    )
+                for component_id in components_created_ids:
```
stefannica, line 7138: `components_created_ids` is declared inside the try-except block:

```
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_stores/sql_zen_store.py", line 6965, in create_full_stack
    service_connector_response = self.create_service_connector(
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/zenml/analytics/utils.py", line 178, in inner_func
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_stores/sql_zen_store.py", line 6102, in create_service_connector
    service_connector.validate_and_configure_resources(
  File "/opt/venv/lib/python3.11/site-packages/zenml/models/v2/core/service_connector.py", line 196, in validate_and_configure_resources
    _validate_and_configure_resources(
  File "/opt/venv/lib/python3.11/site-packages/zenml/models/v2/core/service_connector.py", line 1013, in _validate_and_configure_resources
    raise ValueError(
ValueError: connector configuration is not valid: missing required attribute 'role_arn'

During handling of the above exception, another exception occurred:

  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_server/utils.py", line 447, in decorated
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_server/routers/workspaces_endpoints.py", line 417, in create_full_stack
    return zen_store().create_full_stack(full_stack)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/zenml/zen_stores/sql_zen_store.py", line 7138, in create_full_stack
    for component_id in components_created_ids:
                        ^^^^^^^^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'components_created_ids' where it is not associated with a value

```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6919,6 +6920,230 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+                        continue
+
+                handler.metadata.update(
+                    stack_response.get_analytics_metadata()
+                )
+            except Exception as e:
+                for service_connector_id in service_connectors_created_ids:
+                    self.delete_service_connector(
+                        service_connector_id=service_connector_id
+                    )
```
stefannica, line 7137: I think you need to delete stack components first, otherwise you run into a validation error because the service connector is still referenced

stefannica (changes requested):

stefannica:
FYI regular stack registration is also currently broken:

```
 zenml stack register aws-s3-kube -o k8s -a aws-s3 -c aws --set
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/stefan/aspyre/src/zenml/.venv/lib/python3.10/site-packages/pydantic/main.py:176 in         │
│ __init__                                                                                         │
│    173 │   │   """                                                                               │
│    174 │   │   # `__tracebackhide__` tells pytest and some other tools to omit this function fr  │
│    175 │   │   __tracebackhide__ = True                                                          │
│ ❱  176 │   │   self.__pydantic_validator__.validate_python(data, self_instance=self)             │
│    177 │                                                                                         │
│    178 │   # The following line sets a flag that we use to determine when `__init__` gets overr  │
│    179 │   __init__.__pydantic_base_init__ = True  # pyright: ignore[reportFunctionMemberAccess  │
ValidationError: 2 validation errors for FullStackRequest
service_connectors.0.uuid
  UUID input should be a string, bytes or UUID object [type=uuid_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.7/v/uuid_type
service_connectors.0.ServiceConnectorInfo
  Input should be a valid dictionary or instance of ServiceConnectorInfo [type=model_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.7/v/model_type
```

src/zenml/models/v2/misc/full_stack.py
```diff
@@ -0,0 +1,75 @@
...
+
+from zenml.constants import STR_FIELD_MAX_LENGTH
+from zenml.enums import StackComponentType
+from zenml.models.v2.base.base import BaseRequest
+
+
+class ServiceConnectorInfo(BaseModel):
+    """Information about the service connector when creating a full stack."""
+
+    connector_type: str
```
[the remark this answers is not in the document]
bcdurak: Sure, renaming it now.

src/zenml/models/v2/misc/full_stack.py
```diff
@@ -0,0 +1,75 @@
...
+    auth_type: str
```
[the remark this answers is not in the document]
bcdurak: That's a good idea, changing this as well.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Good catch, the same issue applied to the connectors as well. Fixing them both now.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Oh, that makes sense. I am switching them.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: You are right. However, I am not sure how to properly handle this. We can add the error but this case should never come to pass IMO. 

The index field here refers to a list that exists on a different model and the `FullStackRequest` brings these two models together. It feels like we should have a validator for the request model. What do you think?

avishniakov:
> zenml stack register aws-s3-kube -o k8s -a aws-s3 -c aws

This is fixed

src/zenml/zen_stores/sql_zen_store.py
avishniakov: Addressed in eeb8bea961946d1594fa1edcf20637a9c36423df

src/zenml/models/v2/misc/full_stack.py
```diff
@@ -0,0 +1,93 @@
...
+        "`service_connectors` from `FullStackRequest`.",
+    )
+    service_connector_resource_id: Optional[str] = None
+    configuration: Dict[str, Any] = {}
+
+
+class FullStackRequest(BaseRequest):
+    """Request model for a full-stack."""
+
+    user: Optional[UUID] = None
```
schustmi, line 52: Isn't there a `UserScoped/WorkspaceScoped Request` class for these?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6919,6 +6920,209 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
+                while True:
+                    try:
+                        stack_request = StackRequest(
+                            user=full_stack.user,
+                            workspace=full_stack.workspace,
+                            name=stack_name,
+                            description=full_stack.description,
+                            components=components_mapping,
+                        )
+                        stack_response = self.create_stack(stack_request)
```
schustmi, line 7102: This now means the `REGISTERED_STACK` gets sent twice, right?

schustmi (changes requested):

src/zenml/models/v2/misc/full_stack.py
avishniakov: @bcdurak , this is also on the note of the lining issue

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Yes. Creating the full stack can potentially mean that you have to create service connectors, components, and a stack. To avoid code duplication, I used the methods that were already provided by the `SqlZenStore`.

However, as you wrote, this leads to a place where we have two `REGISTERED_STACK` calls happening within the boundaries of the same function. Potentially, we can solve this in different ways:

1. We can change the event to a different one, which is specific to this endpoint. Perhaps something like `REGISTERED_FULL_STACK`?
2. Or, we can copy over the code of the `create_stack` into the `create_full_stack` method so that we don't call the same analytics event twice.

schustmi: Or maybe you don't call the `REGISTERED_STACK` in your full stack method if it doesn't create a stack. This is a helper method, and the actual stack is being registered in `create_stack(...)`, which correctly calls the analytics event.

schustmi: Why not just add a `metadata` field to the stack request and DB schema, where the `generated_by_wizard` can be stored? This would be the right approach IMO, that solves both this problem and also others like detecting which stacks were created by the wizard later on.

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -349,6 +351,68 @@ def create_stack(
     )
 
 
+@router.post(
+    WORKSPACES + "/{workspace_name_or_id}" + FULL_STACK,
+    response_model=StackResponse,
+    responses={401: error_response, 409: error_response, 422: error_response},
+)
+@handle_exceptions
+def create_full_stack(
```
schustmi, line 360: Why is this endpoint needed actually? Is this just to implement this quicker in the frontend, or some other reason? It seems like one could achieve the same thing with the existing endpoints

src/zenml/zen_stores/sql_zen_store.py
bcdurak: @schustmi As we discussed before, we have solved this over the `labels` field (previously present in components and connectors, now available for stacks). We also introduced a new notion of labels which start with `zenml:` which directly gets added to the analytics metadata. Would be great if you could take a quick look once again.

schustmi: LGTM

src/zenml/zen_server/routers/workspaces_endpoints.py
stefannica: Yes, quicker for the frontend, but also easier to register the whole shebang with one API call in the AWS cloud formation stack, where we only have access to very basic python libraries (urllib) and the python script needs to be supplied inline. Both reasons why having an endpoint behind which we can hide most of the logic that would otherwise be duplicated in the CLI, front-end and cloud stack deployment templates.

bcdurak (approved):
Really excited for this stuff!! 🚀

schustmi (approved):

PR #2818, opened 2024-07-01
Fail pipeline run if error happens during deployment

schustmi:
## Describe changes
Update the run status to `Failed` if an error happens before the pipeline starts when running from the server.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/pipeline_deployment/utils.py
+++ b/src/zenml/zen_server/pipeline_deployment/utils.py
@@ -17,7 +17,7 @@ from zenml.constants import (
     ENV_ZENML_ACTIVE_STACK_ID,
     ENV_ZENML_ACTIVE_WORKSPACE_ID,
 )
-from zenml.enums import StackComponentType, StoreType
+from zenml.enums import ExecutionStatus, StackComponentType, StoreType
 from zenml.integrations.utils import get_integration_for_module
 from zenml.models import (
     CodeReferenceRequest,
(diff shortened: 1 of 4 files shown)


src/zenml/zen_server/pipeline_deployment/utils.py
```diff
@@ -125,52 +124,62 @@ def run_pipeline(
         deployment_id=new_deployment.id
     )
 
-    def _task() -> None:
-        pypi_requirements, apt_packages = get_requirements_for_stack(
-            stack=stack
-        )
+    placeholder_run = create_placeholder_run(deployment=new_deployment)
+    assert placeholder_run
```
AlexejPenner, line 128: Not sure if it's just me, but assertions in productive code rubs me the wrong way. Are we that certain that this will exist? Can't we instead do an if check and throw an error instead? or do a try except around the create_placeholder_run?

AlexejPenner: even if you just end up reraising the same error, which I guess you would wanna do here

schustmi: Yes we're sure, the deployment does not contain a schedule which is the only reason no placeholder run would be created.

I'm not sure what you mean by reraising, there is no error being raised in these lines of code?

AlexejPenner: sorry, I get what you mean, I made the assumption that we need the assertion there because we assume create_placeholder_run() would fail and raise an error in all other cases. I get now that you are making this assertion because create_placeholder_run returns Optional[ResponseModel]. 

I also noticed that this was preexisting code. Won't hold this up anymore.

AlexejPenner (approved):
:seal:

PR #2820, opened 2024-07-03
Update triggers docs with information on previously-run pipelines

strickvl:
Updated the docs as discussed.

### DIFF
--- a/docs/book/how-to/trigger-pipelines/trigger-a-pipeline-from-another.md
+++ b/docs/book/how-to/trigger-pipelines/trigger-a-pipeline-from-another.md
@@ -45,6 +45,10 @@ def loads_data_and_triggers_training():
     trigger_pipeline(df)  # Will trigger the other pipeline
 ```
 
+{% hint style="info" %}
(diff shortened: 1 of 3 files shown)


docs/book/how-to/trigger-pipelines/trigger-a-pipeline-from-another.md
```diff
@@ -45,11 +45,15 @@ def loads_data_and_triggers_training():
     trigger_pipeline(df)  # Will trigger the other pipeline
 ```
 
+{% hint style="info" %}
+The pipeline that you're triggering (i.e. `training_pipeline` in the above example) has to have been run previously. In other words, the functionality to trigger a pipeline from another only works when a Docker image has previously been built for that pipeline. In most cases this will be because you ran the pipeline already, but in some cases you might have built the image separately.
```
schustmi, line 49: Should we include "on a remote stack" in here somewhere, or do you think the Docker image part covers that?

Also, can you copy this part also to the other "trigger a pipeline" pages as its relevant for those as well

strickvl: Done.

schustmi (approved):

PR #2821, opened 2024-07-03
Small improvements and bug fixes

schustmi:
## Describe changes
This PR contains a number of small fixes and improvements:
- The `pagination_utils.depaginate(...)` function now allows passing keyword arguments that will be used when calling the list function. This means we don't have to use `functools.partial(...)` to apply those arguments to the function before passing it to the depagination.
- The `model` property of a pipeline had a wrong error message that was adjusted.
- When running a pipeline from the server, the deprecated `name` property of the `StepConfigurationUpdate` led to an issue in some cases. That property is now explicitly removed from the update dict.
- The `exclude_none` argument of the `pydantic_utils.update_model(...)` function was not working when the update was a pydantic model, only if it was a dict. There is not a single place in our code where this function is being called with a dictionary however. I fixed the bug and updated the default value for `exclude_none` to `False` to keep the existing behaviour.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/artifact.py
+++ b/src/zenml/cli/artifact.py
@@ -13,7 +13,6 @@
 #  permissions and limitations under the License.
 """CLI functionality to interact with artifacts."""
 
-from functools import partial
 from typing import Any, Dict, List, Optional
 
 import click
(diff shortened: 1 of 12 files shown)


avishniakov (approved):

strickvl (approved):
Was mainly looking at this:

> "When running a pipeline from the server, the deprecated name property of the StepConfigurationUpdate led to an issue in some cases. That property is now explicitly removed from the update dict."

but LGTM

PR #2827, opened 2024-07-04
Fixed edge-case where step run is stored incompletely

AlexejPenner:
## Describe changes

I try to catch the error that we ran across today when a pipeline run fails and an incomplete step run is stored in the database which fails on retrieval. 

Whats missing for this PR is the ability to actually delete this step run/the pipeline.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/schemas/step_run_schemas.py
+++ b/src/zenml/zen_stores/schemas/step_run_schemas.py
@@ -213,8 +213,6 @@ class StepRunSchema(NamedSchema, table=True):
             for artifact in self.output_artifacts
         }
 
-        # the step configuration moved into the deployment - the else case is to
-        # guarantee backwards compatibility
         full_step_config = None
         if self.deployment is not None:
             step_configuration = json.loads(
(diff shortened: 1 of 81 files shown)


src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -212,11 +213,27 @@ def to_model(
...
-                json.loads(self.deployment.step_configurations)[self.name]
+            step_configuration = json.loads(
+                self.deployment.step_configurations
             )
-        elif self.step_configuration is not None:
+            if self.name in step_configuration:
+                full_step_config = Step.model_validate(
+                    step_configuration[self.name]
+                )
+            elif self.step_configuration:
```
schustmi, line 226: Isn't this case already handled below?

Maybe 
```python
  ...
elif not self.step_configuration:
  raise ValueError(...)
```

AlexejPenner: but it won't ever end up in the elif below once it goes into the else statement

schustmi: My suggestion is to get rid of the `else` end use an `elif` instead, see the code above

schustmi:
```python
if self.name in step_configuration:
    full_step_config = Step.model_validate(
        step_configuration[self.name]
    )
elif not self.step_configuration:
    raise ValueError(...)
```

schustmi (approved):

PR #2830, opened 2024-07-08
Remove pipeline versioning and add run templates

schustmi:
## Describe changes

This PR includes two big features:
1. Pipelines are now just a namespace and are not tied to code (previously using the `spec`) or a version anymore. As part of this, the temporary `PipelineNamespace` models and endpoints were removed.
2. The introduction of run templates which can be used to run a pipeline from the dashboard/server.

## Pre-requisites

## Types of changes


bcdurak (commented):
I know this is still in the draft stage, but as I was going through the PR, a few things caught my attention and I wanted to share it here. Hope it helps.

src/zenml/models/v2/core/run_template.py
```diff
@@ -0,0 +1,385 @@
...
+    )
+
+
+# ------------------ Response Model ------------------
+
+
+class RunTemplateResponseBody(WorkspaceScopedResponseBody):
+    """Response body for run templates."""
+
+    runnable: bool = Field(
```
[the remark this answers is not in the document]
schustmi: It'll be used in the endpoint to run a template (not finished yet) and also by the frontend to deactivate the run button on a template.

src/zenml/models/v2/core/run_template.py
```diff
@@ -0,0 +1,385 @@
...
+        default=None, title="The pipeline associated with the template."
+    )
+    build: Optional[PipelineBuildResponse] = Field(
+        default=None,
+        title="The pipeline build associated with the template.",
+    )
+    code_reference: Optional[CodeReferenceResponse] = Field(
+        default=None,
+        title="The code reference associated with the template.",
+    )
```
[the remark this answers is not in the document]
schustmi: You mean in the resources of the deployment?

I added them here explicitly to make work in the frontend easier by avoiding extra requests.

src/zenml/zen_stores/schemas/run_template_schemas.py
```diff
@@ -0,0 +1,284 @@
...
+        Args:
+            update: The update model.
+
+        Returns:
+            The updated schema.
+        """
+        for field, value in update.model_dump(
+            exclude_unset=True, exclude_none=True
+        ).items():
+            setattr(self, field, value)
```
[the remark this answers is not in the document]
schustmi: Yeah not a fan of this either, but I think for consistency I'd leave it this way:
- All our schemas have an `update(...)` method
- The schemas can only update themselves as they don't have access to a DB session (which makes sense I think), which is why extra relation updates happen in the SQLZenStore directly.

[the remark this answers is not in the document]
bcdurak: Exactly, perhaps we can look to improve our approach here later on. Resolving this...

src/zenml/models/v2/core/run_template.py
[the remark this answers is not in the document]
bcdurak: I see. For the sake of consistency, I would still add the property then.

src/zenml/models/v2/core/run_template.py
```diff
@@ -0,0 +1,385 @@
...
+        default=None,
+        title="The ID of the latest run of the run template.",
+    )
+    latest_run_status: Optional[ExecutionStatus] = Field(
+        default=None,
+        title="The status of the latest run of the run template.",
+    )
+    tags: List[TagResponse] = Field(
+        title="Tags associated with the run template.",
+    )
```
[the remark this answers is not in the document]
bcdurak: Resources are the way here. It would work with the tags and since the status of pipeline runs are also in their body, we can move the latest run as well. 💪

src/zenml/models/v2/core/run_template.py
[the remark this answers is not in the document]
bcdurak: Yes, I was talking about the resources of the deployment.

Even though I think it's good practice to change it whenever we have the chance, since the resources of a response won't have their own resources, it wouldn't help in this case. We can correct it later on.

src/zenml/models/v2/core/run_template.py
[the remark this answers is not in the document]
schustmi: You mean having `latest_run` in the resources? I would agree with you, but right now a run has lots of stuff in it's body that we don't want to have in this response I think: `stack`, `pipeline`, `build`, `schedule`, `code_reference`. All stuff that should be in the run resources instead, but I don't want to change that as part of this PR and for backwards compatability.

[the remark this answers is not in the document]
schustmi: I moved the tags though, for templates, pipelines and runs (all the entities where I introduced them).

src/zenml/models/v2/core/run_template.py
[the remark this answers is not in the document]
schustmi: Yep added that 👍

src/zenml/new/pipelines/build_utils.py
```diff
@@ -585,3 +532,32 @@ def verify_custom_build(
...
+    Returns:
+        The checksum.
+    """
+    hash_ = hashlib.md5()  # nosec
+
+    # This checksum is used to see if the stack has been updated since a build
+    # was created for it. We create this checksum not with specific requirements
+    # as these might change with new ZenML releases, but they don't actually
+    # invalidate those Docker images.
+    required_integrations = sorted(
```
bcdurak, line 552: Can any `custom` flavors lead to weird cases here?

src/zenml/zen_stores/migrations/versions/7d1919bb1ef0_add_run_templates.py
```diff
@@ -0,0 +1,102 @@
...
+from sqlalchemy.dialects import mysql
+
+# revision identifiers, used by Alembic.
+revision = "7d1919bb1ef0"
+down_revision = "b59aa68fdb1f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
```
bcdurak, line 21: Would it make sense here to merge the two migration scripts?

src/zenml/zen_server/routers/run_templates_endpoints.py
```diff
@@ -0,0 +1,211 @@
...
+    @router.post(
+        "/{template_id}/runs",
+        responses={
+            401: error_response,
+            404: error_response,
+            422: error_response,
+        },
+    )
+    @handle_exceptions
+    def create_template_run(
```
bcdurak, line 171:
```suggestion
    @router.post(
        "/{template_id}/run",
        responses={
            401: error_response,
            404: error_response,
            422: error_response,
        },
    )
    @handle_exceptions
    def run_run_template(
```

I think the naming of the endpoint and the function is a bit confusing here.

- For the endpoint, I would propose something like: `/run_templates/{template_id}/run`. Because it would execute and action instead of handling an entity.
- As for the function, I would rename it to either `run_template` or `run_run_template` which is similar to how it is defined in the `zen_store_interface`.

src/zenml/utils/template_utils.py
```diff
@@ -0,0 +1,181 @@
...
+    config_template = {
+        "run_name": deployment.run_name_template,
+        "steps": steps_configs,
+        **pipeline_config,
+    }
+    return config_template
+
+
+def generate_config_schema(
+    deployment: PipelineDeploymentResponse, stack: StackResponse
```
bcdurak, line 76: Doesn't the deployment response already include the stack response?

src/zenml/new/pipelines/build_utils.py
bcdurak: Also, any change in the configuration of the component can lead to problems when executing runs even though the required integrations remain the same. 

Perhaps we can use the `updated` field of the stack and compare it to the `created` field of the run template and throw a warning if the stack is updated after the creation of the run template.

bcdurak (commented):
Added some questions.

src/zenml/config/source.py
```diff
@@ -245,5 +246,7 @@ def convert_source(source: Any) -> Any:
 
 
 SourceWithValidator = Annotated[
-    SerializeAsAny[Source], BeforeValidator(convert_source)
+    SerializeAsAny[Source],
+    BeforeValidator(convert_source),
+    WithJsonSchema({"type": "string"}),
```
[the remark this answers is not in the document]
schustmi: We use a schema to validate/autocomplete the YAML config editor in the frontend. In this editor, I wanted people to not have to write a source object, but instead just write a string like this:
```yaml
failure_hook_source: my_utils_file.failure_hook
```

The `WithJsonSchema` function makes sure that anytime the `SourceWithValidator` is used as an attribute on a pydantic model, it shows up in the schema as a string. <- This is also the reason why I had to remove this again: This also affected the OpenAPI schemas that the frontend uses, which now had the wrong type in the schema for e.g. artifact versions. I solved this in a different way now.

src/zenml/new/pipelines/build_utils.py
schustmi: Custom flavors have `integration=None` which gets filtered out. This is just for the checksum computation however, running templates from the server does extra checks that prevent custom flavors completely, so we don't have to worry about it.

schustmi: As for the configuration update: You're right that it might lead to problems, however I think in most cases it actually doesn't. Changing the experiment name of an experiment tracker, adding labels to a container registry, ... . So while my implementation here is not perfect, it definitely covers cases that will always fail (new integrations in the stack that are not built into the Docker image). IMO we should go with this, and solve the corner cases by introducing real stack versioning. WDYT?

src/zenml/zen_stores/migrations/versions/7d1919bb1ef0_add_run_templates.py
schustmi: Is there any real benefit? I think these are two separate features and much clearer if separated

src/zenml/new/pipelines/build_utils.py
bcdurak: Sure, the implementation is quite nice right now. My proposal was not to make it fail with the misalignment between the `updated` and `created` fields but rather to throw just a warning. 

Also, you are right. The ultimate solution here is the stack versioning. Until that is implemented, I think this is a good interim solution. 👍

src/zenml/utils/template_utils.py
schustmi: Yes, but not hydrated. This is actually a bigger thing I noticed. I also thought about this, and I think the best way would actually be to update these methods to work with a `DeploymentSchema` instead to avoid multiple DB calls.

There are actually many improvements here:
- Even with the `StackResponse` being passed in here, I still need to call the ZenStore again for each flavor of the stack components, which is lots of unnecessary DB calls.
- All our models have `Client().zen_store.get_something` in their `get_hydrated_version()` implementation. This means when we (accidently) hydrate a model on the server, it uses the `Client` which is something we should avoid. Apparently this is not failing at the moment, but I think we should look at that and fix it before it causes any issues.

src/zenml/zen_server/routers/run_templates_endpoints.py
schustmi: I think @AlexejPenner would disagree :P

Your suggestion is not a restful endpoint that deals with resources. While yes this does execute an action, it also creates a `run` resource for a specific template. No strong opinion from my side though.

If the endpoint stays as is, I think the current name is better. Otherwise I agree with your suggestion 👍

bcdurak: I wrote this thinking @AlexejPenner would be proud of me. Let's see what he thinks 😆

AlexejPenner: I'm all for /runs - i don't like action endpoints outside of /action/...

bcdurak (approved):

PR #2843, opened 2024-07-10
Fix requests vulnerability

stefannica:
## Describe changes
Unpin `requests` and upgrade `docker` to remediate: https://github.com/zenml-io/zenml/security/dependabot/279

Related: https://github.com/zenml-io/zenml/pull/2716

The `docker-compose` package has been deprecated and no longer works with newer versions of the `docker` Python package. For that reason, we have to explicitly install an older version of `docker` (and also an older version of `requests`) whenever we need to use `docker-compose` in our CI. 

## Pre-requisites

## Types of changes


avishniakov (changes requested):
Deps not installable, please check

avishniakov (dismissed):
Now all Dockerized tests throw `TypeError: kwargs_from_env() got an unexpected keyword argument 'ssl_version'` and some templates died with `ModuleNotFoundError: No module named 'packaging'`

stefannica:
> Now all Dockerized tests throw `TypeError: kwargs_from_env() got an unexpected keyword argument 'ssl_version'` and some templates died with `ModuleNotFoundError: No module named 'packaging'`

I had to make more changes to remove our test environment dependency on the old and deprecated `docker-compose` python package.

stefannica:
> > Now all Dockerized tests throw `TypeError: kwargs_from_env() got an unexpected keyword argument 'ssl_version'` and some templates died with `ModuleNotFoundError: No module named 'packaging'`
> 
> I had to make more changes to remove our test environment dependency on the old and deprecated `docker-compose` python package.

That didn't work either due to some other docker-compose v2 dependencies. I chose to simply install an older version of docker whenever docker-compose is also required.

avishniakov:
> > > Now all Dockerized tests throw `TypeError: kwargs_from_env() got an unexpected keyword argument 'ssl_version'` and some templates died with `ModuleNotFoundError: No module named 'packaging'`
> > 
> > 
> > I had to make more changes to remove our test environment dependency on the old and deprecated `docker-compose` python package.
> 
> That didn't work either due to some other docker-compose v2 dependencies. I chose to simply install an older version of docker whenever docker-compose is also required.

How should the users proceed in this case? One needs docker-compose to run on cloud stacks, but the package requests a higher docker version. Wouldn't it give them problems?

.github/workflows/integration-test-slow.yml
```diff
@@ -155,7 +155,7 @@ jobs:
         if: inputs.test_environment != 'default'
         run: |
           pip install uv
-          uv pip install --system "pyyaml==5.3.1" docker-compose
+          uv pip install --system "pyyaml==5.3.1" "docker==6.1.3" docker-compose
```
strickvl: Feels like at the very least we should include a #TODO here to address this at some point? Or will this just work forever?

strickvl (commented):
Also you probably should ask someone other than @avishniakov for a review since he's away on holiday...

strickvl:
Also there's a conflict to resolve

.github/workflows/integration-test-slow.yml
stefannica: The only way to fix this is to get rid of docker-compose (which is just a thin Python layer on top of the `docker-compose` shell commands anyway) and replace it with something else. We may also need to do that sooner rather than later if it turns out that's the reason why our Docker CI jobs are still failing.

strickvl (approved):

PR #2847, opened 2024-07-12
add image pull secrets to k8s pod settings

wjayesh:
## Describe changes
I added image pull secrets to k8s pod settings to allow pods to pull images when they are running on private clusters and other environments where implicit access is not easy to provide.

An alternative to this would be to create a custom service account, attach pull secrets to it and then use it for your orchestrator and step pods but the more options, the merrier. This would also be easier to use, imo.

## Pre-requisites

## Types of changes


schustmi:
In general I think ZenML should even be able to set those image pull secrets for users if they're using service connectors, correct?

wjayesh:
@schustmi hundred percent. but this change should also include other orchestrators in general, eventually. Right now, service connectors are only useful client side. I can start with this one and we can deal with other orchestrators after this, what say?

stefannica:
> In general I think ZenML should even be able to set those image pull secrets for users if they're using service connectors, correct?

This would be just a bit more difficult to set up, because you would have to give the ZenML pods permissions to create those secrets, but 100% that it should be doable.

PR #2861, opened 2024-07-16
Misc cleanup after release

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/utils/source_utils.py
```diff
@@ -57,7 +58,9 @@
     type=SourceType.BUILTIN,
 )
 
-_CUSTOM_SOURCE_ROOT: Optional[str] = None
+_CUSTOM_SOURCE_ROOT: Optional[str] = os.getenv(
+    ENV_ZENML_CUSTOM_SOURCE_ROOT, None
+)
```
safoinme, line 63: I thought setting custom root from env variable won't work because it would check for repo, that's why we had to add it in `source_utils.set_custom_source_root`

safoinme (changes requested):

src/zenml/utils/source_utils.py
schustmi: Setting the ZenML repository (what `zenml init` creates) using the `ZENML_REPOSITORY_PATH` env variable does not work. This however is the new one you introduced, which is fine. I just set it at a different place, but because it's set globally it will be run before `source_utils.get_source_root` is called for the first time

safoinme (approved):

PR #2866, opened 2024-07-17
Adding a `logo_url` and the of the `integration` to component responses

bcdurak:
## Describe changes

I added `logo_url` and `integration` as new attributes to the components' response bodies. Unfortunately, the component schema was not directly related to a flavor schema so I had to create a new relationship where we match the flavor by name and type.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/schemas/component_schemas.py
+++ b/src/zenml/zen_stores/schemas/component_schemas.py
@@ -38,6 +38,7 @@ from zenml.zen_stores.schemas.user_schemas import UserSchema
 from zenml.zen_stores.schemas.workspace_schemas import WorkspaceSchema
 
 if TYPE_CHECKING:
+    from zenml.zen_stores.schemas.flavor_schemas import FlavorSchema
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/schemas/component_schemas.py
```diff
@@ -85,6 +86,13 @@ class StackComponentSchema(NamedSchema, table=True):
     run_metadata: List["RunMetadataSchema"] = Relationship(
         back_populates="stack_component",
     )
+    flavor_schema: Optional["FlavorSchema"] = Relationship(
```
schustmi, line 89: Can we not remove the `flavor: str` field now that we have the explicit reference here, and then call this attribute `flavor` to be consistent? Or is that out of scope for this ticket?

bcdurak: Out of scope for this one.

schustmi (approved):

PR #2868, opened 2024-07-18
Unpin `requests` dependency

strickvl:
We pinned requests on account of this: https://github.com/docker/docker-py/issues/3256

New versions have been released which fix this, so I just remove the restriction on requests since it's such a fundamental library and it was blocking other dependency upgrades / usage.

### DIFF
--- a/examples/e2e_nlp/config.yaml
+++ b/examples/e2e_nlp/config.yaml
@@ -17,6 +17,8 @@
 
 settings:
   docker:
+    python_package_installer: uv
+    install_stack_requirements: False
     required_integrations:
       - aws
       - skypilot_aws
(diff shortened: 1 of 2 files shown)


avishniakov (approved):

stefannica:
@strickvl I think you can close it in favor of this PR I already had open: https://github.com/zenml-io/zenml/pull/2843

stefannica (approved):

stefannica:
I don't think removing the requests pin is enough to address the reason it was added there in the first place (https://github.com/docker/docker-py/issues/3256). Namely, this still fails for me:

```
pip install  "docker==6.1.3"
python -c "from docker.client import DockerClient; DockerClient.from_env()"
```

but this works:

```
pip install  "docker==6.1.3" "requests<2.32.0"
python -c "from docker.client import DockerClient; DockerClient.from_env()"
```

stefannica:
The only `docker` version that works with `requests` higher than 2.32.0 is `docker==7.1.0`, but that has the unfortunate side-effect that it's no longer compatible with docker-compose, which we rely on to run our CI.

strickvl:
No longer needed since we have #2843

PR #2870, opened 2024-07-19
Disable notebook error for Kubernetes orchestrator

strickvl:
Disables an error from getting raised when running pipelines in the Kubernetes orchestrator from within a notebook.

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
@@ -338,15 +338,15 @@ class KubernetesOrchestrator(ContainerizedOrchestrator):
             RuntimeError: If trying to run from a Jupyter notebook.
         """
         # First check whether the code is running in a notebook.
-        if Environment.in_notebook():
-            raise RuntimeError(
-                "The Kubernetes orchestrator cannot run pipelines in a notebook "
-                "environment. The reason is that it is non-trivial to create "
-                "a Docker image of a notebook. Please consider refactoring "
-                "your notebook cells into separate scripts in a Python module "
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -337,16 +337,17 @@ def prepare_or_run_pipeline(
...
-        if Environment.in_notebook():
-            raise RuntimeError(
-                "The Kubernetes orchestrator cannot run pipelines in a notebook "
-                "environment. The reason is that it is non-trivial to create "
-                "a Docker image of a notebook. Please consider refactoring "
-                "your notebook cells into separate scripts in a Python module "
-                "and run the code outside of a notebook when using this "
-                "orchestrator."
-            )
+        # if Environment.in_notebook():
```
schustmi, line 342: We can just remove this right, no need to keep it commented out?

schustmi (changes requested):

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
AlexejPenner: yeah delete this and this is good to go imo

AlexejPenner (approved):
:seal:

wjayesh (changes requested):
Let's go! Although this might not work if the pipelines and steps are defined inside the blocks. Have you tried running a pipeline solely from within the notebook? 

In any case, we can at least keep this as a warning and not raise an error.

schustmi (approved):

wjayesh (approved):

PR #2899, opened 2024-08-01
Run steps defined in notebooks with remote orchestrators

schustmi:
## Describe changes
This PR allows users to define steps in self-contained notebook cells and then run these steps remotely.

## Pre-requisites

## Types of changes


src/zenml/materializers/base_materializer.py
```diff
@@ -98,6 +98,10 @@ def __new__(
                     associated_type, cls
                 )
 
+            from zenml.utils import notebook_utils
+
+            notebook_utils.save_notebook_cell_id(cls)
```
avishniakov, line 103: Just in case, maybe `try_save_notebook_cell_id` is a bit more readable?

src/zenml/utils/notebook_utils.py
```diff
@@ -0,0 +1,249 @@
...
+                _ACTIVE_NOTEBOOK_PATH = ip.user_ns["__vsc_ipynb_file__"]
+            else:
+                try:
+                    _ACTIVE_NOTEBOOK_PATH = str(ipynbname.path())
+                except FileNotFoundError:
+                    logger.warning(
+                        "Unable to detect active notebook. You can use the %s "
+                        "environment variable to manually specify a path to the "
+                        "notebook that you're currently using.",
+                        ENV_ZENML_NOTEBOOK_PATH,
```
avishniakov, line 74: Minor: we mostly use `f-string` in the code.

src/zenml/utils/notebook_utils.py
```diff
@@ -0,0 +1,249 @@
...
+    global _ACTIVE_NOTEBOOK_PATH
+
+    if _ACTIVE_NOTEBOOK_PATH == _UNINITIALIZED:
+        import ipynbname
+
+        from zenml.utils import source_utils
+
+        if path := os.environ.get(ENV_ZENML_NOTEBOOK_PATH, None):
+            logger.info(
+                "Using notebook path from environment variable: %s", path
```
avishniakov, line 56: Minor: we mostly use `f-string` in the code.

src/zenml/utils/notebook_utils.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if _ACTIVE_NOTEBOOK_PATH:
+            relative_path = os.path.relpath(
+                _ACTIVE_NOTEBOOK_PATH, source_utils.get_source_root()
+            )
+            if relative_path.startswith(os.path.pardir):
+                logger.warning(
+                    "Notebook path %s is not in source root %s, ignoring it..",
+                    _ACTIVE_NOTEBOOK_PATH,
+                    source_utils.get_source_root(),
```
avishniakov, line 86: Minor: we mostly use `f-string` in the code.

src/zenml/utils/notebook_utils.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    notebook_json = load_notebook(notebook_path)
+
+    for cell in notebook_json["cells"]:
+        if cell["id"] == cell_id:
+            if cell["cell_type"] != "code":
+                logger.warning(
+                    "Trying to extract code from notebook cell, but the cell "
+                    "type is %s. Continuing anyway..",
+                    cell["cell_type"],
```
avishniakov, line 165: Minor: we mostly use `f-string` in the code.

src/zenml/utils/notebook_utils.py
```diff
@@ -0,0 +1,249 @@
...
+                )
+
+            # TODO: Should we exclude lines starting with special symbols here?
+            # Or whould that mess with some code?
```
avishniakov, line 169: Special like magic cells, for instance?

src/zenml/materializers/base_materializer.py
schustmi: Done

src/zenml/utils/notebook_utils.py
schustmi: Yep exactly. I guess lines that start with `%` or `!`, but that could in theory remove some python lines right?

avishniakov: I can do this
```python
%load_ext autoreload

import this
```
So I would be cautious with this. Maybe just throw out the cells with `%%` in the beginning?
But again:
```bash
%%sh

python3 -c "import this"
```

Can we just keep them?

schustmi: Yeah combined cells are tricky. How about something like this?

```
for line in cell:
  if line.startswith("%") or line.startswith("!"):
    logger.warning("Remove all non-python code from your cells")
```

avishniakov: Should be good

src/zenml/utils/notebook_utils.py
avishniakov: I did some extra reading on this and have to admit that this is desirable approach 🙂 
In this case it would make sense to refactor a bit, since we use f-strings heavily in logging.

avishniakov (approved):

PR #2909, opened 2024-08-06
Dynamic (templated) names for model versions

avishniakov:
## Describe changes
This PR gives the ability to name Model Versions using `{date}` and `{time}` placeholders to use for automated runs with meaningful naming or for scheduling purposes.
`Model(version="my_model_version_{date}_{time}",...)` will be translated into a meaningful name upon pipeline execution.

This PR also has some minor improvements/refactors to ensure that the Model Version is created at a proper time and uses the time of the run as a filler.

*Results of testing in Pro*:
- Display of ModelVersions associated with Runs is partially working: visible in Pipeline Run, not visible in the Step Run. My best guess: is that the code is relying on the config states, which are not the valid source of truth in the templated world anymore, so we would need a change here. Data is available in the respective resources of the Run Responses.
- Scheduling with templated names works as expected.
- Side bug: none of my AWS resources were connected to a connector after creating via UI, something to look into separately.

## Pre-requisites

## Types of changes


avishniakov:
@coderabbitai review

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -88,6 +88,7 @@ class BaseOrchestrator(StackComponent, ABC):
     """
 
     _active_deployment: Optional["PipelineDeploymentResponse"] = None
+    supports_scheduling: ClassVar[bool] = False
```
schustmi, line 91: This seems like something that should be a property on the `BaseOrchestratorFlavor` instead? It might be useful to get this information without needing all the dependencies for a specific orchestrator installed.

src/zenml/models/v2/core/model_version.py
```diff
@@ -343,6 +343,7 @@ def to_model_class(
             version=self.name,
             was_created_in_this_run=was_created_in_this_run,
             suppress_class_validation_warnings=suppress_class_validation_warnings,
+            model_version_id=self.id,
         )
         mv.model_version_id = self.id
```
schustmi, line 348: Probably not necessary anymore?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -423,10 +459,10 @@ def _prepare(
                     step_run=step_run,
                     step_source=self._step.spec.source,
                 )
-                if self._step.config.model:
+                if model:
                     orchestrator_utils._link_pipeline_run_to_model_from_context(
                         pipeline_run_id=step_run.pipeline_run_id,
-                        model=self._step.config.model,
+                        model=model,
```
schustmi, line 465: Maybe this question is stupid, but do we still need this `ModelVersionPipelineRunRequest`? It seems like this would now just be an `UPDATE PipelineRun` that sets `PipelineRun.configured_model_version_id`?

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -115,6 +115,11 @@ class PipelineRunRequest(WorkspaceScopedRequest):
         default=None,
         title="Tags of the pipeline run.",
     )
+    configured_model_version_id: Optional[UUID] = Field(
```
schustmi, line 118: Are there any other ways that a model version can be linked to a pipeline run other than being explicitly configured? Do we count it for example if someone tries to load an artifact from the model version inside the step?

src/zenml/new/pipelines/pipeline.py
```diff
@@ -761,6 +768,13 @@ def _run(
                         "`zenml up`."
                     )
 
+                if self.configuration.model:
+                    logs_of_model = self.configuration.model._prepare_model_version_inside_run(
```
schustmi, line 772: Maybe the method name is wrong, but at this point we aren't actually inside the run yet right?

src/zenml/new/steps/step_context.py
```diff
@@ -177,18 +185,34 @@ def model(self) -> "Model":
...
+                model = self.pipeline_run.config.model
         else:
             raise StepContextError(
                 f"Unable to get Model in step `{self.step_name}` of pipeline "
                 f"run '{self.pipeline_run.id}': it was not set in `@step` or `@pipeline`."
             )
 
         # warm-up the model version
-        model._get_or_create_model_version()
+        if shall_warm_up:
```
schustmi, line 210: Why does this need to happen here? Can't we use e.g. code in the step launcher to always load/create the model version that is configured for a step before the step even launches, which means in all other places we can simply assume if the `StepRunResponse` has no model associated with it, there is none?

schustmi (changes requested):
I tried to review this but I feel I lack the understanding how the whole model linkage to pipeline/step runs is supposed to work. @avishniakov can you maybe explain that before I continue?

- I can configure a pipeline/step to be associated with a certain model by putting it in the decorator/config file
- These will be linked to my pipeline run/step run
- Are there other ways that we link models and pipeline/step runs?

avishniakov:
> I tried to review this but I feel I lack the understanding how the whole model linkage to pipeline/step runs is supposed to work. @avishniakov can you maybe explain that before I continue?
> 
> * I can configure a pipeline/step to be associated with a certain model by putting it in the decorator/config file
> * These will be linked to my pipeline run/step run
> * Are there other ways that we link models and pipeline/step runs?

You nailed it! 

If you put it in one of the decos - it will be associated with the corresponding run (and also linked to a pipeline run - this is still a separate entity). With these changes, theoretically, it should be possible to sunset the pipeline run model version links table (to be validated and realigned with the model).

If you put model definitions elsewhere, e.g. in step codes - no link will be created. This is by design as of now. Subject to change though, while it might make sense not only to track the links, but also to keep track of link flavor (read, write, own, etc.).

src/zenml/models/v2/core/model_version.py
avishniakov: I don't think so, this one is there to keep track of the already fetched/created models in a Model class itself. Used to ensure that we don't create the second model version in the second step and so on. I will revalidate this once again though.

src/zenml/orchestrators/step_launcher.py
avishniakov: This is the idea I also shared in the general comment below. Let me check if I can get rid of the MV PR links table completely.

src/zenml/models/v2/core/pipeline_run.py
avishniakov: As of now - we don't. Only what is in Decorator counts

src/zenml/models/v2/core/model_version.py
schustmi: Actually my comment was just to the line `mv.model_version_id = self.id` which seems to be passed in the `__init__` call above

avishniakov: Indeed, thanks!

src/zenml/orchestrators/base_orchestrator.py
bcdurak: To be more exact, I think this should be on the `BaseOrchestratorConfig` as a `@property` if possible. 

We have similar propertis like `is_remote`, `is_valid` in our base `StackComponentConfig` and `is_synchronous` in our `BaseOrchestratorConfig`.

bcdurak (changes requested):
Adding a few more comments on top of Michael's review.

src/zenml/orchestrators/step_launcher.py
avishniakov: So the problem with flattening it down at the moment is the following: it is possible to have N:M relationship for Pipeline Runs vs Model Versions. This happens, if I set one Model Version in pipeline decorator and some other in some step decorators. This is not reflected in the state of the Pipeline Run, but can be awkwardly reconstructed back by looking up in Step/Pipeline Runs.
This is in theory doable, but I would not take such a rework in the scope of this ticket.

src/zenml/orchestrators/base_orchestrator.py
avishniakov: @bcdurak , thanks, I'll move it there, but what are the benefits of having it as a property and not a class variable there?

src/zenml/models/v2/core/pipeline_run.py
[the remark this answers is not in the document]
avishniakov: Hm, it will be the same, but the `model_version` is a populated `ModelVersionResponse` using the `configured_model_version_id` to lookup for the model version. Is this somehow wrong?

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -98,6 +108,7 @@ class VertexOrchestrator(ContainerizedOrchestrator, GoogleCredentialsMixin):
     """Orchestrator responsible for running pipelines on Vertex AI."""
 
     _pipeline_root: str
+    supports_scheduling: ClassVar[bool] = True
```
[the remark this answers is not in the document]
avishniakov: Addressed

src/zenml/new/steps/step_context.py
avishniakov: Addressed

src/zenml/new/pipelines/pipeline.py
avishniakov: Addressed

src/zenml/orchestrators/base_orchestrator.py
avishniakov: Addressed

avishniakov:
CI is not progressing much now, but can you re-review please?

src/zenml/model/model.py
```diff
@@ -819,3 +830,81 @@ def __hash__(self) -> int:
...
+
+        logs = ""
+
+        # copy Model instance to prevent corrupting configs of the
+        # subsequent runs, if they share the same config object
+        self_copy = self.model_copy()
+
+        # in case request is within the step and no self-configuration is provided
+        # try reuse what's in the pipeline run first
+        if step_run is None and pipeline_run.model_version is not None:
```
schustmi, line 861: I'm assuming we never get here with a `step_run` that already has a `model_version`, correct?

src/zenml/new/steps/step_context.py
```diff
@@ -178,18 +186,21 @@ def model(self) -> "Model":
             StepContextError: If the `Model` object is not set in `@step` or `@pipeline`.
         """
         if self.step_run.config.model is not None:
-            model = self.step_run.config.model
+            if self.step_run.model_version:
+                model = self.step_run.model_version.to_model_class()
+            else:
+                model = self.step_run.config.model
```
schustmi, line 192: This still seems a little fishy to me. We should not need the config fallback here anymore right?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -113,6 +115,14 @@ class PipelineRunSchema(NamedSchema, table=True):
         ondelete="SET NULL",
         nullable=True,
     )
+    configured_model_version_id: UUID = build_foreign_key_field(
```
schustmi, line 118: Just to keep this in line with all our other schemas/models, this can simply be `model_version_id` right? Same for the relationship below and also the `StepRunSchema`

schustmi (changes requested):
Nice, it already seems much less complex now! Just left a few more small suggestions but for the most part looks good

src/zenml/new/pipelines/pipeline.py
```diff
@@ -1314,17 +1326,9 @@ def with_options(
 
         pipeline_copy = self.copy()
 
-        pipeline_copy._parse_config_file(
-            config_path=config_path,
-            matcher=inspect.getfullargspec(self.configure)[0]
-            + ["model_version"],  # TODO: deprecate `model_version` later on
+        pipeline_copy._reconfigure_from_file_with_overrides(
```
schustmi, line 1329: I'm still not entirely sure how this whole config file logic works/is supposed to work:
- We apply the config file to the pipeline here
- We also include it in run_args just below, which means that it will get re-applied once `Pipeline._run(...)` is getting called

Maybe what would help me is if you could point out some errors in how I think this behaves:
- I call `pipeline.with_options(config_path="config_a.yaml", enable_cache=...)`: This applies the config file, and overwrites the value with the value for `enable_cache` that was passed in code.
-  I call `pipeline.with_options(config_path="config_b.yaml", enable_cache=...)` again with a different config file. Again, the config file will get applied and the caching config overwritten.
- I now run the pipeline. This applies the config_b file again, which overwrites the value given by `enable_cache` in code in the step before this.

src/zenml/model/model.py
avishniakov: Yep, we hit this code once at the step launch.

src/zenml/new/steps/step_context.py
avishniakov: Yep, simplified

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
avishniakov: Renamed

src/zenml/new/pipelines/pipeline.py
avishniakov: TBH, this is also confusing to me. I would keep it clean and forbid overriding the values, so we don't land in such ethical discussions.
As of now we do it for the pipeline parameters, but not for configurations.

I think that what you described is valid, but I never tested anything that complex before.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -141,6 +151,9 @@ class PipelineRunSchema(NamedSchema, table=True):
     step_runs: List["StepRunSchema"] = Relationship(
         sa_relationship_kwargs={"cascade": "delete"},
     )
+    configured_model_version: "ModelVersionSchema" = Relationship(
```
schustmi, line 154:
```suggestion
    model_version: "ModelVersionSchema" = Relationship(
```

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -147,6 +157,9 @@ class StepRunSchema(NamedSchema, table=True):
             "primaryjoin": "StepRunParentsSchema.child_id == StepRunSchema.id",
         },
     )
+    configured_model_version: "ModelVersionSchema" = Relationship(
```
schustmi, line 160:
```suggestion
    model_version: "ModelVersionSchema" = Relationship(
```

src/zenml/new/pipelines/pipeline.py
schustmi: I see. Let's leave it as is in the current PR, but this is something that we should maybe revisit and clean up so we know what exactly is going on at some point 👍

schustmi (changes requested):

schustmi (approved):

src/zenml/models/v2/core/pipeline_run.py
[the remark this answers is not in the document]
bcdurak: I would not say that it is wrong but it feels like the effort is duplicated. In a regular case (where the model version is in the `Resources`), you would always have access to the id of the model version (regardless of hydration) through `pipeline_run.model_version.id` or vice versa.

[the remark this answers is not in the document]
bcdurak: It is also fine to leave it like this btw. If we want we can improve this later on, resolving the issue here.

bcdurak (approved):

PR #2913, opened 2024-08-07
Added last timestamp to zenserver

htahir1:
## Describe changes
I added a system endpoint that simply tracks the last user activity in the server. This is useful in various scenarios, 

## Pre-requisites

## Types of changes


src/zenml/zen_server/utils.py
```diff
@@ -570,3 +575,63 @@ def verify_admin_status_if_no_rbac(
...
+        "/system",
+        "/docs",
+        "/redoc",
+        "/openapi.json",
+    ]
+
+    # Check if the path starts with the user endpoint prefix
+    user_prefix = f"{API}{VERSION_1}{ACTIONS}"
+    if request.url.path.startswith(user_prefix):
+        return True
```
stefannica, line 605: What is this for ?

src/zenml/zen_server/zen_server_api.py
```diff
@@ -265,6 +298,12 @@ async def dashboard(request: Request) -> Any:
     return templates.TemplateResponse("index.html", {"request": request})
 
 
+@app.get("/system/last-activity")
+async def get_last_activity() -> Dict[str, str]:
+    """Get the timestamp of the last user activity."""
+    return {"last_user_activity": last_user_activity.isoformat()}
+
+
```
stefannica, line 306: I'll leave some corner-cases for you to consider here:

* when you have a server running multiple pod replicas, this will be less reliable, because calling the endpoint will only return the time when one particular pod has been accessed. But with larger time periods this should even out, so a few seconds or even hours of time difference between pods won't matter.
* the timestamp is reset whenever the pod is reset. This basically means that if even one of the zenml server pods is restarted during an observation time window, it will incorrectly appear as if it has seen user activity. Pods can be re-started purely for maintenance reasons, like pods being moved to another Kubernetes node to free up resources.

stefannica (approved):
Looks good and should work fairly well if you're not looking for extremely reliable results. I left some suggestions / food for thought.

src/zenml/zen_server/zen_server_api.py
htahir1: Thanks! I thought about both cases and I think that for our use-case this should be fine

/re replicas, it should even out, we're  going for a 12 hour horizon anyway
/re restarts, yeah it will just pick up the latest timestamp. this shouldn't effect a lot of things in the grand scheme of things.

src/zenml/zen_server/utils.py
htahir1: so all user requests go through /api/v1/.. however i had ACTIONS in there by mistake. fixed it now

src/zenml/zen_server/zen_server_api.py
```diff
@@ -159,6 +164,34 @@ async def set_secure_headers(request: Request, call_next: Any) -> Any:
...
+    Args:
+        request: The incoming request object.
+        call_next: A function that will receive the request as a parameter and
+            pass it to the corresponding path operation.
+
+    Returns:
+        The response to the request.
+    """
+    global last_user_activity
```
avishniakov, line 182: Not a big fan of this approach. Can we use DB for that? The overhead should be quite minimal, just ensure you write in async mode.

avishniakov (commented):
With DB storage we can close the edge cases raised by @stefannica .

src/zenml/zen_server/utils.py
avishniakov: I think the question is, why do we need this, if in this case none of the below will be met, so True will be returned anyways.

htahir1: i just wnated to be sure somehow ... but maybe I was overthinking it

src/zenml/zen_server/zen_server_api.py
htahir1: @avishniakov why are you not a big fan? Why do i need to do a DB write? whats the argument against this in memory?

avishniakov: As raised by @stefannica, memory is not syncing across nodes and is tending to errors on restarts. Both arguments make this variable a candidate for permanent storage with sync across nodes.

For your case specifically, imagine a server running 1 node and outing every 1 hour due to a memory leak or whatever. For you this server will be forever active, while it should be shut down after 12 hours. Is this fine with you? A DB is a simple improvement which can solve it in one shot and with minimal overhead, IMO.

htahir1: alright agreed

src/zenml/zen_server/zen_server_api.py
```diff
@@ -159,6 +172,44 @@ async def set_secure_headers(request: Request, call_next: Any) -> Any:
...
+            f"An unexpected error occurred while checking user activity: {e}"
+        )
+    if (
+        (datetime.now(UTC) - last_user_activity_reported).total_seconds()
+        > DEFAULT_ZENML_SERVER_REPORT_USER_ACTIVITY_TO_DB_SECONDS
+    ):
+        from zenml.models import ServerSettingsUpdate
+
+        last_user_activity_reported = datetime.now(UTC)
+        zen_store().update_server_settings(
```
schustmi, line 209: This means users could use the endpoint to set this value to something in the future, correct? Maybe this should use an internal zen store method like `update_last_user_activity` which only exists on the SQLZenStore?

avishniakov: Make sense, will look into

src/zenml/zen_server/zen_server_api.py
```diff
@@ -159,6 +172,44 @@ async def set_secure_headers(request: Request, call_next: Any) -> Any:
...
+        if is_user_request(request):
+            last_user_activity = datetime.now(UTC)
+    except Exception as e:
+        logger.debug(
+            f"An unexpected error occurred while checking user activity: {e}"
+        )
+    if (
+        (datetime.now(UTC) - last_user_activity_reported).total_seconds()
+        > DEFAULT_ZENML_SERVER_REPORT_USER_ACTIVITY_TO_DB_SECONDS
+    ):
```
stefannica: Brilliant ! With this, you avoid writing to the database too often.

src/zenml/zen_server/zen_server_api.py
```diff
@@ -265,6 +316,20 @@ async def dashboard(request: Request) -> Any:
...
+    Returns:
+        A JSON with last user activity.
+    """
+    return {
+        "last_user_activity": zen_store()
+        .get_server_settings()
+        .last_user_activity.isoformat()
+    }
+
+
```
stefannica, line 334: I will only make one suggestion here: instead of creating yet another endpoint, you could just add the `last_user_activity` field to the server info, which is already a non-authorized endpoint.

stefannica (approved):

src/zenml/zen_server/zen_server_api.py
htahir1: @stefannica wouldnt that report back to the telemetry needlessly? not sure it matters

schustmi:
@avishniakov FYI: Just merged one of my PRs to dev which contains a DB migration as well, so the downstream revision in the migration here has to be updated

src/zenml/zen_server/zen_server_api.py
avishniakov: Reworked

src/zenml/zen_server/zen_server_api.py
avishniakov: @stefannica @htahir1 shall I do the change?

avishniakov: I think the suggestion is very good actually. I will apply now.

htahir1: ok then lets go for it @avishniakov

schustmi (approved):

PR #2921, opened 2024-08-08
Allow non-ascii characters in requests

schustmi:
## Describe changes
Don't fail if non-ascii characters are in the request body.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/rest_zen_store.py
+++ b/src/zenml/zen_stores/rest_zen_store.py
@@ -13,6 +13,7 @@
 #  permissions and limitations under the License.
 """REST Zen Store implementation."""
 
+import json
 import os
 import re
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -275,6 +276,21 @@
...
+    We need to use this when sending data to fastapi, otherwise it fails to
+    parse the request body.
+
+    Args:
+        json_string: A json string
+
+    Returns:
+        The input string with only ascii characters.
+    """
+    return json.dumps(json.loads(json_string), ensure_ascii=True)
```
strickvl, line 291: CleanShot 2024-08-08 at 13 50 50@2x

Is this what we want? Should we warn people when they're doing this?

schustmi: If you load this into a pydantic model, it parses it back into café:

Screenshot 2024-08-08 at 13 54 25

schustmi: I also tried with pydantic1, where `model.json()` did the ascii encoding automatically and we never had issues there right?

strickvl: Sorry I should have been clearer. I wanted to say that this seems like it'd be unexpected for these values to be transformed this way. Wondering if we want an approach that replaces the letters maybe with some `?` character or something instead?

```python
ascii_text = text.encode('ascii', errors='replace').decode('ascii')
print(ascii_text)  # Output: caf?
```

There's also a bunch of packages that handle this, obv, like `unidecode`. Probably overkill here, but it'd at least turn `café` into `cafe`.

strickvl (approved):

src/zenml/zen_stores/rest_zen_store.py
schustmi: This is only for requests we send to our own endpoints, in which the request will get converted to a pydantic model again, which means `café` will stay `café`.

AlexejPenner (approved):
:seal:

Tested for ö in step comments and in model metadata. Seems to work well.

PR #2928, opened 2024-08-11
Bugfix for artifacts coming from a different artifact store

bcdurak:
## Describe changes
I implemented a bugfix for loading artifacts from a different artifact store than the active artifact store.

## Pre-requisites

## Types of changes


src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +308,95 @@ def _get_model_versions_from_artifacts(
...
+        Args:
+            exc_type: The class of the exception
+            exc_value: The instance of the exception
+            traceback: The traceback of the exception
+        """
+        active_artifact_store_response = Client().get_stack_component(
+            component_type=StackComponentType.ARTIFACT_STORE,
+            name_id_or_prefix=self.active_artifact_store_id,
+        )
+        _ = StackComponent.from_model(active_artifact_store_response)
```
avishniakov, line 402: Why do we need this? Is this a hack to overwrite the creds in fileio? If so, maybe we can have something cleaner, since this is just calls with some unexpected byproduct, IMO.

src/zenml/orchestrators/step_runner.py
```diff
@@ -447,14 +447,22 @@ def _load_input_artifact(
             # we use the datatype of the stored artifact
             data_type = source_utils.load(artifact.data_type)
```
avishniakov, line 449: Shall the `_load_artifact_from_uri` also be changed?

avishniakov (changes requested):

src/zenml/orchestrators/utils.py
bcdurak: The goal here is to recreate the active artifact store, which in turn registers the correct filesystem (with the proper set of credentials) in the filesystem registry. I agree that the behavior might not be very obvious at first glance but I would not categorize it as a hack though.

src/zenml/orchestrators/step_runner.py
bcdurak: As far as I can see, there is no need. `_load_artifact_from_uri` accepts an artifact store instance as an input and passes it to the constructor of the materializer. 

Any materializer that uses `self._artifact_store` will have access to it and any materializer that uses `fileio` will have access to the correct filesystem because the artifact store is already initialized at this point.

src/zenml/artifacts/utils.py
```diff
@@ -710,12 +710,19 @@ def _get_artifact_store_from_response_or_from_active_stack(
...
-            logger.warning(
-                "Unable to restore artifact store while trying to load artifact "
-                "`%s`. If this artifact is stored in a remote artifact store, "
-                "this might lead to issues when trying to load the artifact.",
-                artifact.id,
+        except KeyError:
+            raise RuntimeError(
+                "Unable to fetch the artifact store with id: "
+                f"'{artifact.artifact_store_id}'. Check whether the artifact"
+                "store still exists and you have to right permissions to "
```
schustmi, line 717:
```suggestion
                "store still exists and you have permissions to "
```

src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +308,95 @@ def _get_model_versions_from_artifacts(
...
+    will be restored.
+    """
+
+    def __init__(self, target_artifact_store_id: Optional[UUID]) -> None:
+        """Initialization of the context manager.
+
+        Args:
+            target_artifact_store_id: the ID of the artifact store to load.
+        """
+        self.active_artifact_store_id = Client().active_stack.artifact_store.id
```
schustmi, line 338: This ID should only be stored once we call enter, it might change in the meantime.

src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +308,95 @@ def _get_model_versions_from_artifacts(
...
+        self.target_artifact_store_id = target_artifact_store_id
+
+    def __enter__(self) -> "BaseArtifactStore":
+        """Instantiate the target artifact store and return it.
+
+        Returns:
+            the python object
```
schustmi, line 345: No

src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +308,95 @@ def _get_model_versions_from_artifacts(
...
+        """
+        try:
+            if self.target_artifact_store_id is not None:
+                if (
+                    self.active_artifact_store_id
+                    != self.target_artifact_store_id
+                ):
+                    get_logger(__name__).warning(
```
schustmi, line 353: Why do we need a warning for this? Isn't this the whole purpose of this context manager to do exactly that?

src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +308,95 @@ def _get_model_versions_from_artifacts(
...
+                    StackComponent.from_model(artifact_store_model_response),
+                )
+            else:
+                return Client().active_stack.artifact_store
+
+        except KeyError:
+            raise RuntimeError(
+                "Unable to fetch the artifact store with id: "
+                f"'{self.target_artifact_store_id}'. Check whether the artifact"
+                "store still exists and you have to right permissions to "
```
schustmi, line 374:
```suggestion
                "store still exists and you have permissions to "
```

src/zenml/orchestrators/utils.py
schustmi: I think a comment to explain would be nice at least. Or maybe instead of storing the artifact store ID, just store the object and call `self.artifact_store._register()` instead?

src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +308,95 @@ def _get_model_versions_from_artifacts(
             else:
                 break
     return models
+
+
+class artifact_store_handler:
```
schustmi, line 313: I think the name does not signal at all what this does. In what way does it handle my artifact store for me?

Maybe `register_artifact_store_filesystem`?

schustmi (changes requested):

src/zenml/artifacts/utils.py
bcdurak: Fixed it.

src/zenml/orchestrators/utils.py
bcdurak: Fixed it.

src/zenml/orchestrators/utils.py
bcdurak: Maybe, using a warning here is a bit of an overkill. However, since this is a rare occasion that might lead to some interesting behavior, I would at least throw a debug message here.

src/zenml/orchestrators/utils.py
bcdurak: Solid point. Renamed the context manager.

src/zenml/orchestrators/utils.py
bcdurak: Removed the `self.active_artifact_store_id` altogether now. Using the client instead.

src/zenml/orchestrators/utils.py
bcdurak: I added a comment on why we do this. As for the `__exit__`, I switched to `Client().active_stack..artifact_store._register()` instead.

bcdurak: Is it at all possible for the `Client().active_stack.artifact_store` to change between the `__enter__` and `__exit__` @schustmi ?

schustmi: Very unlikely but sure it's possible. But actually you probably want to register the active artifact store in that case right, not the old one?

bcdurak: Makes sense. It's a good thing that's exactly what the code does right now then 😄.

src/zenml/orchestrators/utils.py
```diff
@@ -305,3 +310,94 @@ def _get_model_versions_from_artifacts(
...
+
+        Args:
+            exc_type: The class of the exception
+            exc_value: The instance of the exception
+            traceback: The traceback of the exception
+        """
+        # As we exist the handler, we have to re-register the filesystem
+        # that belongs to the active artifact store as it may have been
+        # overwritten.
+        if ENV_ZENML_SERVER not in os.environ:
```
schustmi, line 402: I like this check to avoid using the client, but shouldn't this also be used in the enter call where we compare to the active client stack?

avishniakov (changes requested):
I think we cannot release it without a single test case. @bcdurak can you come up with some where the stack switch happens to validate that it is working as expected?

src/zenml/orchestrators/utils.py
bcdurak: I don't think so. In the `__enter__` call, both artifact stores (target and active) are initiated properly, meaning that they go through their own `__init__` which executes this check before `_register`ing in any case.

src/zenml/orchestrators/utils.py
bcdurak: I am not exactly sure if this `No` was referring to the implementation or the docstring. In any case, let me know if it is resolved.

src/zenml/orchestrators/utils.py
schustmi: Ok maybe more specific: Why did you add this `if ENV_ZENML_SERVER not in os.environ:` check here?

schustmi (approved):
I think the test doesn't really cover what it should, but if you think a real test is too complicated i'm okay with it

avishniakov (approved):

PR #2931, opened 2024-08-14
Lightening studio orchestrator

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
+
+---
+description: Orchestrating your pipelines to run on Lightning AI.
+---
+
+
+# Lightning AI Orchestrator
+
+[Lightning AI](https://lightning.ai/) is a platform that simplifies the development and deployment of AI applications. The Lightning AI orchestrator is an integration provided by ZenML that allows you to run your pipelines on Lightning AI's infrastructure, leveraging its scalable compute resources and managed environment.
```
htahir1, line 9: Lightning AI studio right?

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+* You are looking for a fast and easy way to run your pipelines on GPU instances.
+
+* you're already using Lightning AI for your machine learning projects.
+
+* you want to leverage Lightning AI's managed infrastructure for running your pipelines.
+
+
+* you're looking for a solution that simplifies the deployment and scaling of your ML workflows.
+
+
```
htahir1, line 28: extra lines here and above

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+* you want to take advantage of Lightning AI's optimizations for machine learning workloads.
+
+
+## How to deploy it
+
+To use the Lightning AI orchestrator, you need to have a Lightning AI account and the necessary credentials. You don't need to deploy any additional infrastructure, as the orchestrator will use Lightning AI's managed resources.
```
htahir1, line 34: link to lightning ai studio

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+
+
+## How to use it
+
+To use the Lightning AI orchestrator, you need:
+
+*   The ZenML `lightning` integration installed. If you haven't done so, run
+
+    ```shell
+    zenml integration install lightning
```
htahir1, line 44: rather use a pip install here

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+
+```python
+from zenml.integrations.lightning.flavors.lightning_orchestrator_flavor import LightningOrchestratorSettings
+
+
+lightning_settings = LightningOrchestratorSettings(
+    main_studio_name="my_studio",
+    machine_type="cpu",
+    async_mode=True,
+    custom_commands=["pip install -r requirements.txt", "do something else"]
```
htahir1, line 95: i think this needs a seperate paragraph explaining what happens

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+        "orchestrator.lightning": lightning_settings
+    }
+)
+def my_pipeline():
+    ...
+```
+
+
+{% hint style="info" %}
+ZenML will build a wheel that includes all the files within you zenml repository and upload it to the lightning studio.
```
htahir1, line 109: this isnt happening anymore right? the `zenml init` doesn't need to be where the pipeline is and the wheel isn't created either

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+You can now run any ZenML pipeline using the Lightning AI orchestrator:
+
+```shell
+python file_that_runs_a_zenml_pipeline.py
+```
+
+
+### Lightning AI UI
+
+Lightning AI provides its own UI where you can monitor and manage your running applications, including the pipelines orchestrated by ZenML.
```
htahir1, line 123: screenshots?

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+Check out the [SDK docs](https://sdkdocs.zenml.io/latest/integration_code_docs/integrations-lightning/#zenml.integrations.lightning.flavors.lightning_orchestrator_flavor.LightningOrchestratorSettings) for a full list of available attributes and [this docs page](../../how-to/use-configuration-files/runtime-configuration.md) for more information on how to specify settings.
+
+
+To use GPUs with the Lightning AI orchestrator, you need to specify a GPU-enabled machine type in your settings:
+
+```python
+lightning_settings = LightningOrchestratorSettings(
+    machine_type="gpu",
```
htahir1, line 179: give the list of options currently available?

docs/book/component-guide/orchestrators/lightning.md
```diff
@@ -0,0 +1,187 @@
...
+Lightning AI provides its own UI where you can monitor and manage your running applications, including the pipelines orchestrated by ZenML.
+For any runs executed on Lightning AI, you can get the URL to the Lightning AI UI in Python using the following code snippet:
+
+```python
+from zenml.client import Client
+
+pipeline_run = Client().get_pipeline_run("<PIPELINE_RUN_NAME>")
+orchestrator_url = pipeline_run.run_metadata["orchestrator_url"].value
+```
+
```
htahir1, line 132: i think we need a pragraph about how zenml spins up the studio, and what commands it runs. also we should talk about async mode and what happens then. finally we should talk about reuse of studios and the ability to even point to a base studio

htahir1 (changes requested):
Left many  comments on docs first :-) Otherwise its great!

src/zenml/integrations/lightning/orchestrators/lightning_orchestrator.py
```diff
@@ -0,0 +1,592 @@
...
+                #    "orchestrator."
+                # )
+
+            return True, ""
+
+        return StackValidator(
+            custom_validation_function=_validate_remote_components,
+        )
+
+    def _get_lightning_client(
```
schustmi, line 101: This does not get a lightning client, it sets lighning env variables

src/zenml/integrations/lightning/orchestrators/utils.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+    )
+
+    # Remove empty items and duplicates
+    requirements = sorted(set(filter(None, requirements)))
+
+    return requirements
+
+
+def download_and_extract_code(code_path: str, extract_dir: str) -> None:
```
schustmi, line 72: Seems like this is unused? If I missed it somewhere, it should probably use the function from `code_utils` instead.

src/zenml/utils/code_utils.py
```diff
@@ -168,55 +169,69 @@ def write_archive(
...
-    artifact_store = Client().active_stack.artifact_store
-
     with tempfile.NamedTemporaryFile(
         mode="w+b", delete=False, suffix=".tar.gz"
     ) as f:
         code_archive.write_archive(f)
 
         hash_ = hashlib.sha1()  # nosec
 
+        f.seek(0)
```
schustmi, line 188: Was this a bug with the hash computation before or why is this `seek` needed?

src/zenml/utils/code_utils.py
```diff
@@ -168,55 +169,69 @@ def write_archive(
...
+def zip_and_hash_code(code_archive: CodeArchive) -> Tuple[str, str]:
+    """Zip the code archive and compute its hash.
 
     Args:
-        code_archive: The code archive to upload.
+        code_archive: The code archive to zip and hash.
 
     Returns:
-        The path where to archived code is uploaded.
+        A tuple containing the temporary file path and the computed hash.
```
schustmi, line 179: The second item of this tuple is not the computed hash, but instead a filename. I think returning two filenames is not intuitive at all what they're supposed to be.

As far as I can see, this function is used below and in the lightning orchestrator, which actually doesn't even use the hash.

I would suggest we just use 
```python
with tempfile.NamedTemporaryFile(
        mode="w+b", delete=False, suffix=".tar.gz"
    ) as f:
        code_archive.write_archive(f)
```
in the lightning orchestrator (and also in the function below), I don't think that's worth having in a separate helper function.

I still like your idea of extracting the hash computation though, maybe we can make this `compute_file_hash(file) -> str`?

schustmi (changes requested):
Didn't have time to look into the core parts of the PR yet, so just a few initial comments for now

src/zenml/utils/code_utils.py
safoinme: it was suggested by claude auto-complete

src/zenml/utils/code_utils.py
safoinme: I think it's just doc string error later on i was also using the filename as second output

src/zenml/utils/code_utils.py
schustmi: That doesn't mean you shouldn't think about it or verify that it makes sense. In this case, the seeking already happens at the end of `code_archive.write_archive()` which means it's not necessary here.

src/zenml/utils/code_utils.py
schustmi: Still I think it would be nicer to have a helper function that computes just the hash of a file instead of having both the archive writing and hash computation combined in one function.
If we had a util function to compute just the file hash, we have other places in the code where we could reuse it.

If you still think that we should not duplicate this code

```python
with tempfile.NamedTemporaryFile(
        mode="w+b", delete=False, suffix=".tar.gz"
    ) as f:
        code_archive.write_archive(f)
```
maybe we create a second helper function which is called `write_archive_to_temporary_file(...) -> str` which does this and returns the filename?

schustmi (changes requested):

src/zenml/integrations/lightning/orchestrators/lightning_orchestrator_entrypoint_configuration.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import TYPE_CHECKING, List, Set
+
+if TYPE_CHECKING:
+    from uuid import UUID
+
+RUN_NAME_OPTION = "run_name"
+DEPLOYMENT_ID_OPTION = "deployment_id"
+
+
+class LightningOrchestratorEntrypointConfiguration:
```
schustmi, line 25: Why is this not inheriting and specifying the deployment ID again?

src/zenml/integrations/lightning/orchestrators/lightning_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,307 @@
...
+    main_studio = Studio(name=main_studio_name)
+    if pipeline_settings.machine_type:
+        main_studio.start(Machine(pipeline_settings.machine_type))
+    else:
+        main_studio.start()
+
+    logger.info("Main studio started.")
+    logger.info("Uploading code to main studio the code path: %s", filename)
+    main_studio.upload_file(
+        "/teamspace/studios/this_studio/.lightning_studio/.studiorc",
```
schustmi, line 145: I have not checked this, but there must be some overlap between this code and the code in the orchestrator (some commands that are run in the studio, ...). We should refactor this to use some common util functions in the future.

schustmi (approved):

htahir1 (approved):

PR #2937, opened 2024-08-19
Introduce new quickstart with a focus on Stack switching

AlexejPenner:
## Describe changes
I wrote a new quickstart project with a focus on a simple text-to-text problem and the main focus on switching stack from local execution to orchestration on one of the three hyperscalers.

## Pre-requisites

## Types of changes


release-cloudbuild.yaml
```diff
@@ -89,6 +89,71 @@ steps:
...
+        docker build \
+          --build-arg ZENML_VERSION=$TAG_NAME \
+          --build-arg PYTHON_VERSION=3.11 \
+          --build-arg CLOUD_PROVIDER=aws \
+          --target client \
+          -f docker/zenml-quickstart.Dockerfile . \
+          -t $$USERNAME/zenml-public-pipelines:quickstart-$TAG_NAME-py3.11-aws
+        
+        # Push to aws - do we have credentials here???
+        docker tag $$USERNAME/zenml:$TAG_NAME-py3.11 339712793861.dkr.ecr.eu-central-1.amazonaws.com/zenml-public-pipelines:quickstart-$TAG_NAME-py3.11-aws
```
avishniakov, line 130: I doubt you will get creds here, a secret Env would be needed here and in the Azure below. Better check with infra folks for greater details.

.github/workflows/update-templates-to-examples.yml
```diff
@@ -206,11 +206,11 @@ jobs:
           ref: ${{ github.event.pull_request.head.ref }}
       - name: Check-out fresh Starter template
         run: |
-          rm -rf examples/quickstart
-          mkdir -p examples/quickstart
-          printf 'info@zenml.io' | zenml init --path examples/quickstart --template starter --template-with-defaults
+          rm -rf examples/mlops_starter
```
avishniakov, line 209: I still see only `examples/quickstart` in this branch + the tag is not modified above - is this intended? The tag shall also be updated in the CLI base.

avishniakov (changes requested):

.github/workflows/update-templates-to-examples.yml
AlexejPenner: basically I want to write the old quickstart to mlops_starter and the new quickstart is fully written within this repo

examples/quickstart/run.py
```diff
@@ -128,125 +75,27 @@ def main(
...
-
-        # Run the pipeline
-        inference_configured(**run_args_inference)
-        logger.info("Inference pipeline finished successfully!")
+    run_args_train = {}
+
+    pipeline_args = {}
+    if no_cache:
+        pipeline_args["enable_cache"] = False
+    pipeline_args["config_path"] = os.path.join(
```
schustmi, line 92: There is no config file for the `custom` orchestration environment, which means that will fail with a `FileNotFoundError` at some point I think.

examples/quickstart/run.py
```diff
@@ -128,125 +75,27 @@ def main(
...
+        config_folder, f"training_{orchestration_environment}.yaml"
+    )
+    english_translation_pipeline.with_options(**pipeline_args)(
+        model_type=model_type, **run_args_train
+    )
+    logger.info("Training pipeline finished successfully!\n\n")
```
schustmi, line 98: Someone once ranted that we shouldn't log that a pipeline run finished if it actually didn't, so maybe this message should only be displayed for local/sync orchestrators?

schustmi (changes requested):

src/zenml/integrations/huggingface/materializers/huggingface_t5_materializer.py
```diff
@@ -0,0 +1,107 @@
...
+                    fileio.makedirs(dst)
+                    for subfile in fileio.listdir(src):
+                        subsrc = os.path.join(src, subfile)
+                        subdst = os.path.join(dst, subfile)
+                        fileio.copy(subsrc, subdst)
+                else:
+                    fileio.copy(src, dst)
+
+            # Load the model or tokenizer from the temporary directory
+            if data_type == T5ForConditionalGeneration:
```
schustmi, line 71: I think the recommended way to compare this would be `data_type is T5ForConditionalGeneration`. But maybe even better, can we use `issubclass(data_type, T5ForConditionalGeneration)`. Or another suggestion, how about `data_type.from_pretrained(temp_dir)` as it's the same for all of them anyway?

schustmi (changes requested):

.github/workflows/update-templates-to-examples.yml
avishniakov: You'd better `git mv` it in this case 🙂 
Now I see some issues with `git diff` updating quickstart (old). I created folder manually and checking if this helps

release-cloudbuild.yaml
avishniakov: @AlexejPenner you are pushing it all now to docker registry, right?

avishniakov (approved):

schustmi (approved):

htahir1 (approved):

src/zenml/integrations/huggingface/__init__.py
```diff
@@ -47,6 +48,26 @@ def activate(cls) -> None:
         from zenml.integrations.huggingface import materializers  # noqa
         from zenml.integrations.huggingface import services
 
+    @classmethod
+    def get_requirements(cls, target_os: Optional[str] = None) -> List[str]:
```
AlexejPenner, line 52: @avishniakov how does this look?

src/zenml/integrations/huggingface/__init__.py
```diff
@@ -47,6 +48,26 @@ def activate(cls) -> None:
...
+        requirements = cls.REQUIREMENTS
+
+        # In python 3.8 higher transformers version lead to other packages breaking
+        if sys.version_info.minor > 8:
+            requirements += ["transformers"]
+        else:
+            requirements += ["transformers<=4.31"]
+
+        return requirements
+
```
AlexejPenner, line 70: @avishniakov How does this look

PR #2939, opened 2024-08-21
Fixed optional typing on non-serializable types

AlexejPenner:
## Describe changes
This code:
```
from typing import Optional

from sklearn.neighbors import NearestNeighbors
import numpy as np


@step
def one():
    X = np.array([[1, 1], [2, 2]])
    return NearestNeighbors().fit(X)


@step
def two(a: Optional[int] = None, nn: Optional[NearestNeighbors] = None):
    pass


@pipeline()
def simple_pipeline():
    nn = one()
    two(nn=None, a=None)


simple_pipeline()
```

 Was failing with the following error message:

```
│    404 │   def _unknown_type_schema(self, obj: Any) -> CoreSchema:                               │
│ ❱  405 │   │   raise PydanticSchemaGenerationError(                                              │
│    406 │   │   │   f'Unable to generate pydantic-core schema for {obj!r}. '                      │
│    407 │   │   │   'Set `arbitrary_types_allowed=True` in the model_config to ignore this error  │
│    408 │   │   │   ' or implement `__get_pydantic_core_schema__` on your type to fully support   │
PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'sklearn.neighbors._unsupervised.NearestNeighbors'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or 
implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on 
`<some type>` otherwise to avoid infinite recursion.

For further information visit https://errors.pydantic.dev/2.7/u/schema-for-unknown-type
``` 

`def two(a: Optional[int] = None, nn: Optional[NearestNeighbors] = None):`

While primitive types allowed for optional typing, non-serializable types like NearestNeighbors lead to pydantic errors, as the utility functions were only checking serializable types (as the assumption probably was that at this point we are handling only parameters, not artifacts).

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/steps/entrypoint_function_utils.py
+++ b/src/zenml/steps/entrypoint_function_utils.py
@@ -25,9 +25,11 @@ from typing import (
     Sequence,
     Type,
     Union,
+    get_args,
+    get_origin,
 )
 
-from pydantic import ConfigDict, ValidationError, create_model
+from pydantic import ValidationError, create_model
 
 from zenml.constants import ENFORCE_TYPE_ANNOTATIONS
 from zenml.exceptions import StepInterfaceError
@@ -185,7 +187,6 @@ class EntrypointFunctionDefinition(NamedTuple):
             )
 
         parameter = self.inputs[key]
-
         if isinstance(
             value,
             (
@@ -235,17 +236,40 @@ class EntrypointFunctionDefinition(NamedTuple):
             parameter: The function parameter for which the value was provided.
             value: The input value.
         """
-        config_dict = ConfigDict(arbitrary_types_allowed=False)
+        annotation = parameter.annotation
 
-        # Create a pydantic model with just a single required field with the
-        # type annotation of the parameter to verify the input type including
-        # pydantics type coercion
-        validation_model_class = create_model(
-            "input_validation_model",
-            __config__=config_dict,
-            value=(parameter.annotation, ...),
-        )
-        validation_model_class(value=value)
+        # Handle Optional types
+        origin = get_origin(annotation)
+        if origin is Union:
+            args = get_args(annotation)
+            if type(None) in args:
+                if value is None:
+                    return  # None is valid for Optional types
+                # Remove NoneType from args as this case is handled from here
+                args = tuple(arg for arg in args if arg is not type(None))
+                annotation = args[0] if len(args) == 1 else Union[args]
+
+        # Handle None values for non-Optional types
+        if value is None and annotation is not type(None):
+            raise ValueError(f"Expected {annotation}, but got None")
+
+        # Use Pydantic for all types to take advantage of its coercion abilities
+        try:
+            config_dict = {"arbitrary_types_allowed": True}
+            validation_model_class = create_model(
+                "input_validation_model",
+                __config__=type("Config", (), config_dict),
+                value=(annotation, ...),
+            )
+            validation_model_class(value=value)
+        except ValidationError as e:
+            raise ValueError(f"Invalid input: {e}")
+        except Exception:
+            # If Pydantic can't handle it, fall back to isinstance
+            if not isinstance(value, annotation):
+                raise TypeError(
+                    f"Expected {annotation}, but got {type(value)}"
+                )
 
 
 def validate_entrypoint_function(
--- a/tests/unit/steps/test_base_step_new.py
+++ b/tests/unit/steps/test_base_step_new.py
@@ -13,7 +13,7 @@
 #  permissions and limitations under the License.
 import sys
 from contextlib import ExitStack as does_not_raise
-from typing import Dict, List, Tuple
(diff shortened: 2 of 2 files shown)


src/zenml/steps/entrypoint_function_utils.py
```diff
@@ -235,17 +236,40 @@ def _validate_input_value(
...
+        try:
+            config_dict = {"arbitrary_types_allowed": True}
+            validation_model_class = create_model(
+                "input_validation_model",
+                __config__=type("Config", (), config_dict),
+                value=(annotation, ...),
+            )
+            validation_model_class(value=value)
+        except ValidationError as e:
+            raise ValueError(f"Invalid input: {e}")
```
schustmi, line 266: Why? This validation error was caught in other places to raise a much better error message then you're giving here

src/zenml/steps/entrypoint_function_utils.py
```diff
@@ -235,17 +236,40 @@ def _validate_input_value(
...
-            "input_validation_model",
-            __config__=config_dict,
-            value=(parameter.annotation, ...),
-        )
-        validation_model_class(value=value)
+        annotation = parameter.annotation
+
+        # Handle Optional types
+        origin = get_origin(annotation)
+        if origin is Union:
```
schustmi, line 243: Is there no better way to leverage pydantic for this case? This seems like very messy code, and I'm not even sure why your example failed when using pydantic

src/zenml/steps/entrypoint_function_utils.py
```diff
@@ -235,17 +236,40 @@ def _validate_input_value(
...
+            args = get_args(annotation)
+            if type(None) in args:
+                if value is None:
+                    return  # None is valid for Optional types
+                # Remove NoneType from args as this case is handled from here
+                args = tuple(arg for arg in args if arg is not type(None))
+                annotation = args[0] if len(args) == 1 else Union[args]
+
+        # Handle None values for non-Optional types
+        if value is None and annotation is not type(None):
```
schustmi, line 253: What is this for? This would mean someone would explicitly have a `a: type(None)` annotation right, or how else would we get into this if condition?

schustmi (changes requested):

schustmi (changes requested):
Also, this really screams for a test case

AlexejPenner:
> Also, this really screams for a test case

already had some, are any obvious ones missing?

src/zenml/steps/entrypoint_function_utils.py
```diff
@@ -235,17 +235,23 @@ def _validate_input_value(
...
+        # Use Pydantic for all types to take advantage of its coercion abilities
+        try:
+            config_dict = ConfigDict(arbitrary_types_allowed=True)
+            validation_model_class = create_model(
+                "input_validation_model",
+                __config__=type("Config", (), config_dict),
+                value=(annotation, ...),
+            )
+            validation_model_class(value=value)
+        except Exception:
```
schustmi, line 249: Should this be 
```python
try:
  ...
except ValidationError:
  raise
except Exception:
  ...
```
instead? In case pydantic fails with a regular validation error, we actually want to raise that because it means the types don't match right?

schustmi (changes requested):

src/zenml/steps/entrypoint_function_utils.py
```diff
@@ -235,17 +235,23 @@ def _validate_input_value(
...
+            # If Pydantic can't handle it, fall back to isinstance
+            if not isinstance(value, annotation):
```
schustmi, line 251: Also this might fail with the raw annotation:
```python
isinstance(int, Union[int, str])

# TypeError: Subscripted generics cannot be used with class and instance checks
```

schustmi (changes requested):

schustmi (commented):
Actually I have a more general question regarding this PR:

This is not a json-serializable object, which means by our definition it can never be a parameter. This means this whole PR is related to a step input, which is of a non-json serializable type, for which there is a `None` default value and no artifact is being passed to it. Is that correct?

If yes, there might be other simpler solutions to prevent this case from even getting to this parameter validation method.

schustmi:
I'm still not entirely sure how to handle this. On the one hand, it would be nice to have your code work as is. On the other hand, if someone now we're to switch the default value to something Non-Null, we would fail and say a parameter has to be JSON-serializable, which also seems weird.

In case we want this though, I think we could achieve this by having
```python
arbitrary_types_allowed = value is None
config_dict = ConfigDict(arbitrary_types_allowed=arbitrary_types_allowed)
```
as the first two lines in `_validate_input_value(...)`.

PR #2942, opened 2024-08-22
Enable running a single step on the active stack

schustmi:
## Describe changes

This PR adds functionality to run a step on the active stack.
To make this work, we dynamically create a pipeline with just this single step and then run this single step pipeline.

## Pre-requisites

## Types of changes


htahir1 (changes requested):
Docs please <3

htahir1 (changes requested):
Great one! Just a few things docs should mention:

* ZENML_RUN_SINGLE_STEPS_WITHOUT_STACK how that works -> You can say this as a note while mentioning the `entrypoint()` thing
* We should mention unlsited runs somewhere. Maybe add a page for them or somehow explain them in these pages?

docs/book/user-guide/starter-guide/create-an-ml-pipeline.md
```diff
@@ -194,10 +194,10 @@ def svc_trainer(
 ```
 
 {% hint style="info" %}
-If you want to run the step function outside the context of a ZenML pipeline, all you need to do is call the step function outside of a ZenML pipeline. For example:
+If you want to run just a single step on your ZenML stack, all you need to do is call the step function outside of a ZenML pipeline. For example:
```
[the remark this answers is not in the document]
schustmi: Mhm i'm not sure, mentioning two many alternatives here seems like it would just confuse people in the starter guide, which is why I just opted for the default here. Do you think having explanations for both (and the difference) is necessary here?

docs/book/how-to/build-pipelines/run-an-individual-step.md
```diff
@@ -0,0 +1,70 @@
+# Run an individual step on your stack
+
+If you want to run just an invividual step on your stack, you can simply call the step
+as you would with a normal Python function. ZenML will internally create a pipeline with just your step
+and run it on the active stack.
+
+{% hint style="info" %}
+The pipeline run that will happen to execute your step will be `unlisted`, which means it
+will not be associated with any pipeline.
```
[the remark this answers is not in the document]
schustmi: This won't be available in OSS and also not in cloud for the time being. Do you still want to add it already?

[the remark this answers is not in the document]
htahir1: In the OSS i would expect the pipelines and runs tabs to be there... so id merge this in for now!

docs/book/user-guide/starter-guide/create-an-ml-pipeline.md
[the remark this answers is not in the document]
htahir1: Hmm i dont get it. the text says "If you want to run just a single step on your ZenML stack, all you need to do is call the step function outside of a ZenML pipeline. For example:" ... but this is wrong right? this will run a pipeline run instead?

docs/book/how-to/build-pipelines/run-an-individual-step.md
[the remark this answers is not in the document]
schustmi: The runs tab is part of the template PR right? Which means the porting to OSS hasn't even started yet

htahir1 (approved):

src/zenml/steps/utils.py
```diff
@@ -464,3 +469,88 @@ def log_step_metadata(
...
+        # Make sure the orchestrator runs sync so we stream the logs
+        key = settings_utils.get_stack_component_setting_key(orchestrator)
+        pipeline_settings[key] = BaseSettings(synchronous=True)
+
+    @pipeline(enable_cache=False, settings=pipeline_settings)
+    def single_step_pipeline() -> None:
+        __step(**inputs)
+
+    run = single_step_pipeline.with_options(unlisted=True)()
+    run = wait_for_pipeline_run_to_finish(run.id)
```
avishniakov, line 539: This call has a timeout of 64s if I read this correctly. This is way too low, IMO. Can we at least bump it up and make it configurable?

src/zenml/steps/utils.py
```diff
@@ -464,3 +469,88 @@ def log_step_metadata(
...
+    orchestrator = Client().active_stack.orchestrator
+
+    pipeline_settings: Any = {}
+    if "synchronous" in orchestrator.config.model_fields:
+        # Make sure the orchestrator runs sync so we stream the logs
+        key = settings_utils.get_stack_component_setting_key(orchestrator)
+        pipeline_settings[key] = BaseSettings(synchronous=True)
+
+    @pipeline(enable_cache=False, settings=pipeline_settings)
+    def single_step_pipeline() -> None:
```
avishniakov, line 535: Maybe call it the same as the step?

tests/unit/materializers/test_materializer_registry.py
```diff
@@ -29,7 +29,7 @@ def some_step() -> MyFloatType:
         return MyFloatType(3.0)
 
     with does_not_raise():
-        some_step()()
+        some_step().entrypoint()
```
avishniakov, line 32: Why `entrypoint` is used here and `call_entrypoint` in the next file? They are all units, so I would expect the same to be used. Does param validation bring any harm here?

avishniakov (changes requested):

src/zenml/steps/utils.py
schustmi: It just increases the sleep duration up to 64 seconds but has no actual timeout

src/zenml/steps/utils.py
schustmi: I like that idea 👍

tests/unit/materializers/test_materializer_registry.py
schustmi: Does not bring any harm I think, I'll change them all to `call_entrypoint`

avishniakov (approved):

PR #2943, opened 2024-08-23
Make `run_with_accelerate` a pythonic decorator

avishniakov:
## Describe changes
This PR fixes the problem of calling `run_with_accelerate` as a python decorator directly on a function (e.g. `@run_with_accelerate...`) to make it usable with remote orchestrators like Skypilot.
As a side effect, it also forbids the functional use of this decorator, as this will not be reflected in the `Source` on remote orchestrator later on.

## Pre-requisites

## Types of changes


schustmi (approved):

src/zenml/integrations/huggingface/steps/accelerate_runner.py
```diff
@@ -65,100 +66,99 @@ def training_pipeline(some_param: int, ...):
...
+                "in a functional way with steps, please apply decoration "
+                "directly to a step instead. This behavior will be also "
+                "allowed in future, but now it faces technical limitations.\n"
+                "Example (allowed):\n"
+                f"@{run_with_accelerate.__name__}(...)\n"
+                f"def {step_function.name}(...):\n"
+                "    ...\n"
+                "Example (not allowed):\n"
+                "def my_pipeline(...):\n"
+                f"    run_with_accelerate({step_function.name},...)(...)\n"
```
bcdurak, line 145: I am not too familiar with `accelerate` myself, so I might be missing something here, but if this `RuntimeError` is true, we might want to change the docstring of the `run_with_accelerate` function, as it seems like the example that is given there is not allowed anymore.

avishniakov: Exactly the same is written on L143 🙂 `Example (not allowed):\n`

bcdurak: I was talking about L55 😄 

```python
@pipeline
def training_pipeline(some_param: int, ...):
    run_with_accelerate(training_step, num_processes=4)(some_param, ...)
```

Isn't this the not-allowed case?

avishniakov: Yeah, indeed. Good catch 👍🏼

PR #2944, opened 2024-08-23
Allow functional `run_with_accelerate`

avishniakov:
## Describe changes
This PR updates the structure we use in BaseStep and Source to allow for dynamic decorating of the steps in pipelines with specially backed decorators. Decoration with any deco is still not supported, just with specific ones.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/source.py
+++ b/src/zenml/config/source.py
@@ -69,6 +69,9 @@ class Source(BaseModel):
     attribute: Optional[str] = None
     type: SourceType
 
+    dynamic_decorator: Optional["Source"] = None
+    dynamic_decorator_kwargs: Optional[Dict[str, Any]] = {}
+
     @classmethod
     def from_import_path(
         cls, import_path: str, is_module_path: bool = False
--- a/src/zenml/constants.py
+++ b/src/zenml/constants.py
@@ -501,3 +501,8 @@ ZENML_PRO_CONNECTION_ISSUES_SUSPENDED_PAUSED_TENANT_HINT = (
     "Organization page. If the tenant is PAUSED you can `Resume` it via UI "
     "and try again."
 )
+
+# Step decorators common objects used in setattr/getattr calls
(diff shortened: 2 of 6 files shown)


src/zenml/config/source.py
```diff
@@ -69,6 +69,11 @@ class Source(BaseModel):
     attribute: Optional[str] = None
     type: SourceType
 
+    dynamic_decorator_module: Optional[str] = None
+    dynamic_decorator_attribute: Optional[str] = None
+    dynamic_decorator_type: Optional[SourceType] = None
+    dynamic_decorator_kwargs: Optional[Dict[str, Any]] = {}
```
avishniakov, line 75: Not a big fan of this part - it should have been a Source instance with Kwargs aside, but this leads to circular issues of Pydantic, while it tries to dump Source.

avishniakov:
@bcdurak @schustmi can you have a look please? Shall it be improved and merged or shall we completely drop this and keep only two options for the users:
```python
@run_with_accelerate
@step
def step_1():
    ...

@step
def step_2():
    ...

step_2_accelerated = run_with_accelerate(step_2,...)
```

While wrapping a step inside a pipeline would not be allowed still.

src/zenml/config/source.py
schustmi: Yeah agreed, I also don't really like that such a specific feature (including some config options for the decorator 😅) is now on this super basic class which is essentially just representing a string like "my_module.SomeClass".

It would be nice to handle this in a generic way, not just for decorators on steps, but have a way for objects to modify the behavior of source resolving of themselves/other objects.

schustmi: But even if we had it generic solution, I'm still not sure where we would store the config for this. For me, it seems like something that should be in the `PipelineDeployment` model. Maybe when the decorator is applied, it could add its configuration dynamically to the step configuration somehow?

schustmi (changes requested):
Left some questions

avishniakov:
Closing this as depriotorized

PR #2948, opened 2024-08-25
Modal Step Operator

strickvl:
Seeking an earlyish review before I write the docs (in case anything major changes). This probably isn't the most elegant way of doing this, but it works at least. It supports both CPU and GPU-based workloads, and specifying CPU and memory for the instance on which the step will run.

See [Demo tenant](https://cloud.zenml.io/organizations/fc992c14-d960-4db7-812e-8f070c99c6f0/tenants/8a462fb6-b1fe-48df-9677-edc76bc8352d/runs/3d4d676c-12b6-4350-8030-a7efd5ad5040?tab=configuration) for the run details. Screenshot from Modal dashboard:

CleanShot 2024-08-25 at 17 18 35@2x

A simple dev pipeline to test this:

```python
from pathlib import Path

import zenml
from zenml.config import DockerSettings
from zenml.integrations.modal.flavors.modal_step_operator_flavor import (
    ModalStepOperatorSettings,
)

zenml_git_root = Path(zenml.__file__).parents[2]

docker_settings = DockerSettings(
    dockerfile=str(zenml_git_root / "docker" / "zenml-dev.Dockerfile"),
    build_context_root=str(zenml_git_root),
    python_package_installer="uv",
)

modal_settings = ModalStepOperatorSettings(gpu="H100")


@step
def hello_world() -> int:
    print("Hello world!")
    return 3


@step(
    step_operator="modal",
    settings={"step_operator.modal": modal_settings},
)
def hello_world_on_modal(some_number: int) -> int:
    manipulated_number = some_number * 3
    print(f"Hello I'm inside a container! {manipulated_number}")
    return manipulated_number


@step
def final_step(some_manipulated_number: int) -> None:
    print(f"Final number: {some_manipulated_number}")


@pipeline(settings={"docker": docker_settings}, enable_cache=False)
def hello_world_pipeline():
    some_val = hello_world()
    manipulated_number = hello_world_on_modal(some_val)
    final_step(manipulated_number)


if __name__ == "__main__":
    hello_world_pipeline()
```

## Remaining tasks

- [x] Write docs
- [x] Update integrations page

## Summary by CodeRabbit

- **New Features**
	- Introduced Modal integration for ZenML, enabling cloud-native step execution.
	- Added new modules and functionality related to modal operations.
	- Enhanced configuration and operational settings for Modal step operators.

- **Bug Fixes**
	- Improved error handling for Docker credentials and Modal application exits.

- **Documentation**
	- Updated documentation to reflect new Modal integration capabilities.


strickvl:
@coderabbitai review

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,208 @@
...
+    print("Executing {info.run_name} step...")
+
+if __name__ == "__main__":
+    run_step()
+"""
+            tmp.write(modal_code)
+            tmp_filename = tmp.name
+
+        # Create a Click context object
+        ctx = click.Context(run)
```
schustmi, line 192: I would suggest using normal subprocess calls instead of click for this. Once we implement a slimmer client version that will be installed in the docker images, click will not be part of that package which would lead to issues here.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,208 @@
...
+        # Set the necessary context parameters
+        ctx.ensure_object(dict)
+        ctx.obj["detach"] = False
+        ctx.obj["show_progress"] = True
+        ctx.obj["interactive"] = False
+        ctx.obj["env"] = None
+
+        try:
+            run.main(args=[tmp_filename], prog_name="modal run", obj=ctx.obj)
+        except SystemExit as e:
```
schustmi, line 203: We should remove the temp file here as it contains the docker credentials, right?

schustmi (changes requested):

schustmi:
Some answers for your general questions/comments:
- The settings should be visible on the DAG visualizer in the `Stack` card. Can you point me to a run with some settings so I can verify that it's not showing up.
- How is the orchestrator url related to this PR? It's just a step operator right, and there is no code in your step operator that would even set this metadata right?

I would suggest the following:
- We introduce another metadata constant:
`METADATA_STEP_OPERATOR_URL = "step_operator_url"`
- We implement the `get_step_run_metadata(...)` method in the modal step operator and include the modal URL for the constant above. We can then later expose this on the dashboard and also implement it for other step operators

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,208 @@
...
+            mode="w", delete=False, suffix=".py"
+        ) as tmp:
+            modal_code = f"""import modal
+
+my_secret = modal.Secret.from_dict({{
+    'REGISTRY_USERNAME': '{docker_username}',
+    'REGISTRY_PASSWORD': '{docker_password}',
+}})
+
+zenml_image = modal.Image.from_registry(tag='{image_name}', secret=my_secret).env({environment}).pip_install("pydantic~={major}.{minor}").run_commands('{" ".join(entrypoint_command)}')
```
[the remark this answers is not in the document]
strickvl: Unfortunately not. They said they're working on this etc, but at the moment the current behaviour is to pull the image, then install [this](https://github.com/modal-labs/modal-client/tree/main/modal/requirements) on top. At the moment it only breaks Pydantic, hence the installation on top I do, but agreed it's not super stable. I think I'll mention this in the docs.

strickvl:
> Some answers for your general questions/comments:
> 
> * The settings should be visible on the DAG visualizer in the `Stack` card. Can you point me to a run with some settings so I can verify that it's not showing up.
> * How is the orchestrator url related to this PR? It's just a step operator right, and there is no code in your step operator that would even set this metadata right?
> 
> I would suggest the following:
> 
> * We introduce another metadata constant:
>   `METADATA_STEP_OPERATOR_URL = "step_operator_url"`
> * We implement the `get_step_run_metadata(...)` method in the modal step operator and include the modal URL for the constant above. We can then later expose this on the dashboard and also implement it for other step operators

Ah yes you're right. Actually because of the way their CLI works (and the fact that we're now using subprocess), it's even a bit harder to get the URL for the specific run. Since we don't yet have it for other step operators, I think we can just leave it for now maybe.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,215 @@
...
+    @property
+    def validator(self) -> Optional[StackValidator]:
+        """Get the stack validator for the Modal step operator.
+
+        Returns:
+            The stack validator.
+        """
+
+        def _validate_remote_components(stack: "Stack") -> Tuple[bool, str]:
+            # ... validation logic ...
```
schustmi, line 78: Missing

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,215 @@
...
+                f"memory={resource_settings.get_memory(ByteUnit.MB)}"
+            )
+
+        decorator_args_str = ", ".join(decorator_args)
+
+        with tempfile.NamedTemporaryFile(
+            mode="w", delete=False, suffix=".py"
+        ) as tmp:
+            modal_code = f"""import modal
+
```
kramstrom, line 181:
```suggestion

        import modal

        my_secret = modal.Secret.from_dict(
            {
                "REGISTRY_USERNAME": docker_username,
                "REGISTRY_PASSWORD": docker_password,
            }
        )

        zenml_image = (
            modal.Image.from_registry(tag=image_name, secret=my_secret)
            .env(environment)
            .pip_install(f"pydantic~={major}.{minor}")
            .run_commands(" ".join(entrypoint_command))
        )

        resource_settings = info.config.resource_settings

        sb = modal.Sandbox(
            "bash",
            "-c",
            cmd, # TODO : Add cmd
            image=zenml_image,
            gpu=f":{resource_settings.gpu_count}",
            cpu=resource_settings.cpu_count,
            memory=resource_settings.get_memory(ByteUnit.MB),
            cloud=settings.cloud,
            region=settings.region,
        )

        sb.wait()

        stdout = sb.stdout()
        stderr = sb.stderr()
        returncode = sb.returncode()

        return stdout, stderr, returncode

```

kramstrom: Could the rest of this function be replaced with a modal sandbox?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,215 @@
...
+                ["modal", "run", tmp_filename],
+                check=True,
+                capture_output=True,
+                text=True,
+            )
+            logger.info("Modal app completed successfully.")
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Modal app exited with code {e.returncode}.")
+            logger.error(e.stdout)
+            raise
```
schustmi, line 213: Maybe we should have a custom exception here that explains that our step operator failed?

```python
raise RuntimeError(f"Model step operator failed: {str(e)}")
```
or something like that?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,215 @@
...
+
+if __name__ == "__main__":
+    run_step()
+"""
+            tmp.write(modal_code)
+            tmp_filename = tmp.name
+
+        # Run the Modal app using subprocess
+        try:
+            subprocess.run(
```
schustmi, line 203: I think it would be nice to stream the logs of this to make it easier to see what's happening for the user?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,215 @@
...
+            raise ValueError(
+                "No Container registry found in the stack. "
+                "Please add a container registry and ensure "
+                "it is correctly configured."
+            )
+
+        if docker_creds := stack.container_registry.credentials:
+            docker_username, docker_password = docker_creds
+        else:
+            raise ValueError(
```
schustmi, line 144: IMO `RuntimeError` is more fitting here

schustmi (changes requested):

src/zenml/integrations/modal/step_operators/modal_step_operator.py
kramstrom: Or actually using ephemeral apps would probably be better than sandbox in your case:

```py
        import modal

        my_secret = modal.Secret.from_dict(
            {
                "REGISTRY_USERNAME": docker_username,
                "REGISTRY_PASSWORD": docker_password,
            }
        )

        zenml_image = (
            modal.Image.from_registry(tag=image_name, secret=my_secret)
            .env(environment)
            .pip_install(f"pydantic~={major}.{minor}")
            .run_commands(" ".join(entrypoint_command))
        )

        resource_settings = info.config.resource_settings

        app = modal.App(
            "example-pandas",
            image=zenml_image,
            gpu=f":{resource_settings.gpu_count}",
            cpu=resource_settings.cpu_count,
            memory=resource_settings.get_memory(ByteUnit.MB),
            cloud=settings.cloud,
            region=settings.region,
        )

        with app.run():
            # run your step code here!
            # run_step()
            pass

```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: The logs do stream

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Yeah makes sense

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Sandbox has the advantage of not installing a modal layer of requirements on top of the image, but the disadvantage of not having streaming logs or any observability into what's going on in the image. I couldn't figure out how to get the logs in the dashboard, even. Is there a way?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,217 @@
...
+                ),
+            )
+            .env(environment)
+            .run_commands(" ".join(entrypoint_command))
+        )
+
+        resource_settings = info.config.resource_settings
+        gpu_str = settings.gpu
+        if resource_settings.gpu_count is not None:
+            gpu_str += f":{resource_settings.gpu_count}"
```
schustmi, line 196: `gpu_str` is coming from the settings and might be optional right? In that case, the `+=` will fail I think?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,217 @@
...
+        async def run_sandbox():
+            with modal.enable_output():
+                async with app.run():
+                    await modal.Sandbox.create.aio(
+                        "bash",
+                        "-c",
+                        " ".join(entrypoint_command),
+                        image=zenml_image,
+                        gpu=gpu_str,
+                        cpu=resource_settings.cpu_count or None,
```
schustmi, line 209: the `or None` is not necessary I think, `cpu_count` will already be `None` if it's not set.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,217 @@
...
+                        memory=resource_settings.get_memory(ByteUnit.MB)
+                        or None,
+                        cloud=settings.cloud or None,
```
schustmi, line 212:
```suggestion
                        cloud=settings.cloud,
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,217 @@
...
+                        region=settings.region or None,
```
schustmi, line 213:
```suggestion
                        region=settings.region,
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,217 @@
...
+                async with app.run():
+                    await modal.Sandbox.create.aio(
+                        "bash",
+                        "-c",
+                        " ".join(entrypoint_command),
+                        image=zenml_image,
+                        gpu=gpu_str,
+                        cpu=resource_settings.cpu_count or None,
+                        memory=resource_settings.get_memory(ByteUnit.MB)
+                        or None,
```
schustmi, line 211:
```suggestion
                        memory=resource_settings.get_memory(ByteUnit.MB),
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -0,0 +1,217 @@
...
+
+        app = modal.App(f"zenml-{info.run_name}-{info.step_run_id}")
+
+        async def run_sandbox():
+            with modal.enable_output():
+                async with app.run():
+                    await modal.Sandbox.create.aio(
+                        "bash",
+                        "-c",
+                        " ".join(entrypoint_command),
```
schustmi, line 206: Why do we need to specify the entrypoint both here as well as earlier when we define the image?

schustmi (changes requested):

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: But I have a conditional around it to catch it? Shouldn't that be enough?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: That's a very good question for which I have no good answer. Maybe I'll test removing one or the other of them. Also maybe @kramstrom has an answer for that?

kramstrom: You shouldn't need both, in the `run_commands` for the image you specify the commands to run when building the image (think docker's `RUN ...` statements)

On this line you specify the command to run in the sandbox

strickvl: Removing the command from when the image is created makes it fail. I'm guessing
that the command on the sandbox is actually redundant, but we need it because of
the way they create the Sandbox.

In any case, Modal team are reworking their API around Sandbox creation, so this
code will be updated in the coming weeks anyway, I think.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
schustmi: Yes now it's fine. When I commented there was a conditional around the gpu_count but not the actual gpu_str.

Now one last question: If someone just specifies the `gpu_count`, this string will be `:2` for example. Is that a valid value for modal, or should it only be appended if the type was specified?

```python
gpu_str = settings.gpu
if gpu_str and resource_settings.gpu_count is not None:
  gpu_str += f":{resource_settings.gpu_count}"
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: @kramstrom how would I remove one of them, then? I tried removing the `run_commands` line from the image definition above, but then it broke / stopped working.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Yes good point. I should check for that. Thanks!

src/zenml/integrations/modal/step_operators/modal_step_operator.py
kramstrom: what commands are you running in the entrypoint here?

strickvl: @kramstrom 'python -m zenml.entrypoints.entrypoint --entrypoint_config_source zenml.step_operators.step_operator_entrypoint_configuration.StepOperatorEntrypointConfiguration --deployment_id 86e4cbb8-cc5d-45b2-9969-e6956e200ee1 --step_name hello_world_on_modal --step_run_id b8f5ca23-2947-4463-be2c-2410655db816' something like this is the `entrypoint_command`... it's built at the same time as the image is built. so they should go together.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Done

schustmi (approved):

src/zenml/integrations/modal/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+from zenml.stack import Flavor
+
+MODAL_STEP_OPERATOR_FLAVOR = "modal"
+
+
+class ModalIntegration(Integration):
+    """Definition of Modal integration for ZenML."""
+
+    NAME = MODAL
+    REQUIREMENTS = ["modal>=0.64.49,<1"]
```
avishniakov, line 32: Is it safe enough to give that freedom, assuming things can break easily before 1.x?

avishniakov (approved):
LGTM, just a tiny nit which can be ignored

src/zenml/integrations/modal/__init__.py
strickvl: Yeah I wasn't sure. I also don't want to pin it to a specific version...

avishniakov: Let's hope for the best for now and will see how often it could be breaking due to interface changes 🙂

strickvl: Yeah you can see that they're releasing LOTS of patch updates: https://pypi.org/project/modal/#history

PR #2951, opened 2024-08-26
Add additional filtering options

schustmi:
## Describe changes
This PR adds some additional filtering options that we want to use in the dashboard.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -973,6 +973,7 @@ def filter_and_paginate(
             ValueError: if the filtered page number is out of bounds.
             RuntimeError: if the schema does not have a `to_model` method.
         """
+        query = query.distinct()
```
AlexejPenner, line 976: just for me to understand, why was this necessary only now?

AlexejPenner (commented):
just two questions

src/zenml/models/v2/base/filter.py
```diff
@@ -790,6 +790,67 @@ def _define_bool_filter(
...
+            # a ValueError. In this case, we already know that the filter
+            # will not produce any result and can simply ignore it.
+            pass
+
+        filter_ = self._define_filter(
+            column="name", value=value, operator=operator
+        )
+        conditions.append(filter_.generate_query_conditions(table=table))
+
+        return or_(False, *conditions)
```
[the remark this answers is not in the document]
schustmi: Yes that's correct. Initially my method was more flexible and the conditions array could potentially be empty, which is why I added this safety `False` here so SQLAlchemy doesn't crash. I'll remove it

src/zenml/zen_stores/sql_zen_store.py
schustmi: Tbh I don't really know. I encountered duplicate results once I allowed filtering tags by prefix (e.g. `tag="startswith:t"`), and if one entity contained multiple tags that matched this filter it appeared more than once in the result.

AlexejPenner (approved):
:seal:

AlexejPenner:
@schustmi Anything in particular I should look at again?

schustmi:
@AlexejPenner all of it ;)
I'm guessing the filtering pipelines by latest run status is the only major change that happened since your last review

src/zenml/models/v2/core/run_template.py
```diff
@@ -390,4 +409,37 @@ def get_custom_filters(
...
+                == PipelineDeploymentSchema.id,
+                PipelineDeploymentSchema.pipeline_id == PipelineSchema.id,
+                self.generate_name_or_id_query_conditions(
+                    value=self.pipeline,
+                    table=PipelineSchema,
+                ),
+            )
+            custom_filters.append(pipeline_filter)
+
+        if self.stack:
```
AlexejPenner, line 433:
```suggestion
        if self.stack is not None:
```

Shouldn't this be the same as the two above?

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -682,6 +708,70 @@ def get_custom_filters(
...
+                CodeReferenceSchema.code_repository_id
+                == CodeRepositorySchema.id,
+                self.generate_name_or_id_query_conditions(
+                    value=self.code_repository,
+                    table=CodeRepositorySchema,
+                ),
+            )
+            custom_filters.append(code_repo_filter)
+
+        if self.model:
```
AlexejPenner, line 754: Any reason these are not also `is not None` - what am I missing??

AlexejPenner (approved):
Two small nits for consistency - else :seal:

src/zenml/models/v2/core/pipeline_run.py
schustmi: Tbh i don't think it matters much for these, just empty strings would be treated differently. I guess for empty strings we don't actually want to include them in filters right? Every string in python starts/endswith/contains the empty string, which basically means the filter will always be true?

schustmi: I removed the `is not None` for all of them to disregard empty strings

AlexejPenner (approved):
:seal: - Happy to see filtering get better :)

PR #2953, opened 2024-08-27
Dependency cleanup and Python 3.12 support

bcdurak:
### Major changes

- Cleaning up the dependencies:
    - Removed `click-params`, `httplib2`, `pyparsing` as main dependencies.
    - Moved `secure` to the `server` extra
    - Separated `numpy` and `pandas` to implementations to their own integrations.
    - Added a limited version of `maison` to solve a problem with the `yamlfix` installation.
    - Upgraded `pydantic` to `2.8.*`
- Upgrading `zenml` to Python 3.12. The new version is also added to all relevant CI tests. 

### Minor changes

- Minor formatting/linting changes in some of the integrations.
- Reworked one of the unit tests to use the `PydanticMaterializer` instead of the `PandasMaterializer`.
- Limited the `pandas` version in the `deepchecks` integration to `<2.1.0`.
- Added placeholder materializers for the `PandasMaterializer` and the `NumpyMaterializer`.
- `mlflow` integration slightly reworked. The extra deployment option `mlserver` and `mlserver-mlflow` only installed <3.12 now.
- Some of our examples use the package: `skimage`. Previously, we were able to access this package (for instance in `mlflow` and `seldon` example tests) because `deepchecks` was installed in the same environment and `deepchecks` required it (because we installed all the integration together). However, on 3.12, `deepchecks` integration is still not supported -> it is not installed -> there is no `skimage` -> `mlflow` and `seldon` example tests fail. The package itself is around 14 MB so I added it to the `sklearn` integration.

### Related PRs
- [mlstacks](https://github.com/zenml-io/mlstacks/pull/162)
- [Starter template](https://github.com/zenml-io/template-starter/pull/22)
- [NLP template](https://github.com/zenml-io/template-nlp/pull/23)
- [E2E batch template](https://github.com/zenml-io/template-e2e-batch/pull/40)
- [LLM template](https://github.com/zenml-io/template-llm-finetuning/pull/7)

### Remaining TODOs
- ~~Investigate if we should implement a migration script to change the materializer entries to newer paths.~~
- ~~Investigate Windows testing on 3.12.~~
- ~~Add warnings/errors for integrations that are not yet supported in 3.12.~~.
- ~~Ignore the integration tests on 3.12 for integrations that are not yet supported in 3.12.~~.
- ~~Revert the changes in the CI.~~
- ~~Create new versions of all templates.~~
- ~~Investigate `httplib2` and `pyparsing` issue.~~

## Pre-requisites

## Types of changes


src/zenml/materializers/__init__.py
```diff
@@ -29,8 +29,6 @@
 from zenml.materializers.structured_string_materializer import (
     StructuredStringMaterializer,
 )
-from zenml.materializers.numpy_materializer import NumpyMaterializer
```
schustmi, line 32: We need to keep both those files around for backwards compatibility with older artifacts.

src/zenml/integrations/deepchecks/__init__.py
```diff
@@ -38,10 +38,13 @@ class DeepchecksIntegration(Integration):
     REQUIREMENTS = [
         "deepchecks[vision]==0.8.0",
         "torchvision>=0.14.0",
-        "pandas<2.0.0",
         "opencv-python==4.5.5.64",  # pin to same version
         "opencv-python-headless==4.5.5.64",  # pin to same version
         "tenacity!=8.4.0",  # https://github.com/jd/tenacity/issues/471
+        # Deepchecks materializer inherits from the PandasMaterializer, thus
```
schustmi, line 44: Why can't we actually import the pandas integration here then and put it's requirements?

pyproject.toml
```diff
@@ -237,6 +235,7 @@ dev = [
     "mkdocs-awesome-pages-plugin",
     "mkdocstrings",
     "mike",
+    "maison",
```
schustmi, line 238: What is this?

schustmi (changes requested):

src/zenml/materializers/__init__.py
bcdurak: I know, but we have to be careful keeping the Materializers in this `__init__` file because it is one of the things that we import all the time and the integrations might not be installed. We need to find a graceful way to handle this.

src/zenml/integrations/deepchecks/__init__.py
bcdurak: Should be theoretically possible. However, I am not 100% sure yet on how we will handle the PandasIntegration, so I will comeback to this.

src/zenml/materializers/__init__.py
schustmi: Doesn't have to be in the `__init__` I think, just the `pandas/numpy_materializer.py` files

bcdurak: Since the comment was on the `__init__`, I wanted to mention it just in case.

pyproject.toml
bcdurak: I was planning to dig a bit deeper into this. As I was playing around with the dependencies yesterday, I removed some unused requirements and I think, the relaxed requirements led to a version of `maison` greater than 2.0, which ultimately forces `yamlfix` to fail.

I also saw, as of last week, that we apply a similar fix when we use `install-zenml-dev` script. Perhaps @avishniakov can shed more light on this as he must have bumped into a similar error [here](https://github.com/zenml-io/zenml/blob/bda5e888ea6e323d25a2c72deba19c81e4c7201c/scripts/install-zenml-dev.sh#L63).

bcdurak:
Btw, @schustmi, this PR is still early in development. I was just using it to test things on the CI. Will update you as things move forward.

src/zenml/integrations/deepchecks/__init__.py
bcdurak: Btw, just as an additional note. `deepchecks` integration is a special case here. Much like our other integrations, [their requirements at 0.8.0 (the version that we use in ZenML)](https://github.com/deepchecks/deepchecks/blob/baea2cb313567fbd34d02c67980145e32ead4b15/requirements/requirements.txt#L1) actually allow us to use pandas v2. However, their code is trying to import [a function](https://github.com/deepchecks/deepchecks/blob/baea2cb313567fbd34d02c67980145e32ead4b15/deepchecks/utils/features.py#L21) that was eventually removed in `pandas>=2.1.0`, which leads to an error.

I will try to upgrade our `deepchecks` integration as the issue is resolved in their latest version.

src/zenml/integrations/huggingface/__init__.py
```diff
@@ -27,6 +27,7 @@ class HuggingfaceIntegration(Integration):
     """Definition of Huggingface integration for ZenML."""
 
     NAME = HUGGINGFACE
+
     REQUIREMENTS_IGNORED_ON_UNINSTALL = ["fsspec"]
```
avishniakov, line 31:
```suggestion
    REQUIREMENTS_IGNORED_ON_UNINSTALL = ["fsspec", "pandas"]
```

src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -0,0 +1,192 @@
...
+            self.pyarrow_exists = True
+        except ImportError:
+            self.pyarrow_exists = False
+            logger.warning(
+                "By default, the `PandasMaterializer` stores data as a "
+                "`.csv` file. If you want to store data more efficiently, "
+                "you can install `pyarrow` by running "
+                "'`pip install pyarrow`'. This will allow `PandasMaterializer` "
+                "to automatically store the data as a `.parquet` file instead."
+            )
```
avishniakov, line 66: Why not offer other options like `fastparquet`?

src/zenml/materializers/pandas_materializer.py
```diff
@@ -11,182 +11,27 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 #  or implied. See the License for the specific language governing
 #  permissions and limitations under the License.
-"""Materializer for Pandas."""
+"""Implementation of ZenML's pandas materializer."""
```
avishniakov, line 14: Doesn't feel accurate, since you just import it here.

src/zenml/zen_server/utils.py
```diff
@@ -29,7 +29,7 @@
 )
 from urllib.parse import urlparse
 
-import secure
+# import secure
```
avishniakov, line 32:
```suggestion
```

tests/integration/integrations/deepchecks/data_validators/test_deepchecks_data_validator.py
```diff
@@ -12,18 +12,26 @@
...
+
 from zenml.enums import StackComponentType
-from zenml.integrations.deepchecks import DEEPCHECKS_DATA_VALIDATOR_FLAVOR
-from zenml.integrations.deepchecks.data_validators import (
-    DeepchecksDataValidator,
-)
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 25:
```suggestion
    sys.version_info.minor == 12,
```

tests/integration/integrations/deepchecks/materializers/test_deepchecks_dataset_materializer.py
```diff
@@ -13,19 +13,28 @@
...
+import pytest
 
 from tests.unit.test_general import _test_materializer
-from zenml.integrations.deepchecks.materializers.deepchecks_dataset_materializer import (
-    DeepchecksDatasetMaterializer,
-)
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 25:
```suggestion
    sys.version_info.minor == 12,
```

tests/integration/integrations/deepchecks/materializers/test_deepchecks_result_materializer.py
```diff
@@ -13,47 +13,53 @@
...
-    """Fixture to get a check result."""
     train_dataset = Dataset(
         pd.DataFrame(data={"x": [1, 2, 3, 4, 5, 6, 7, 8, 9]}), label=None
     )
     test_dataset = Dataset(pd.DataFrame(data={"x": [1, 2, 3]}), label=None)
     return DatasetSizeComparison().run(train_dataset, test_dataset)
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 52:
```suggestion
    sys.version_info.minor == 12,
```

tests/integration/integrations/deepchecks/materializers/test_deepchecks_result_materializer.py
```diff
@@ -62,10 +68,20 @@ def test_deepchecks_dataset_materializer_with_check_result(
         )
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 72:
```suggestion
    sys.version_info.minor == 12,
```

tests/integration/integrations/deepchecks/test_validation_checks.py
```diff
@@ -11,15 +11,21 @@
...
+import sys
 
-from zenml.integrations.deepchecks.validation_checks import (
-    DeepchecksValidationCheck,
-)
+import pytest
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 20:
```suggestion
    sys.version_info.minor == 12,
```

tests/integration/integrations/tensorflow/materializers/test_keras_materializer.py
```diff
@@ -12,16 +12,25 @@
...
+import pytest
 
 from tests.unit.test_general import _test_materializer
-from zenml.integrations.tensorflow.materializers.keras_materializer import (
-    KerasMaterializer,
-)
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 23:
```suggestion
    sys.version_info.minor == 12,
```

tests/integration/integrations/tensorflow/materializers/test_tf_dataset_materializer.py
```diff
@@ -12,16 +12,25 @@
...
+import pytest
 
 from tests.unit.test_general import _test_materializer
-from zenml.integrations.tensorflow.materializers.tf_dataset_materializer import (
-    TensorflowDatasetMaterializer,
-)
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor >= 12,
```
avishniakov, line 23:
```suggestion
    sys.version_info.minor == 12,
```

avishniakov (changes requested):

pyproject.toml
bcdurak: I am resolving this then.

src/zenml/materializers/numpy_materializer.py
```diff
@@ -11,236 +11,27 @@
...
+# of the old artifact version entries.
+
+from zenml.exceptions import IntegrationError
+
+try:
+    from zenml.integrations.numpy.materializers.numpy_materializer import (  # noqa
+        NumpyMaterializer,
+    )
+except (ImportError, ModuleNotFoundError) as e:
+    raise IntegrationError(
```
schustmi, line 30: IMO this should be an `ImportError` as that's the reason for the failure, which is very different from an issue with one of our integrations

src/zenml/materializers/pandas_materializer.py
```diff
@@ -11,182 +11,27 @@
...
-
-    ASSOCIATED_TYPES: ClassVar[Tuple[Type[Any], ...]] = (
-        pd.DataFrame,
-        pd.Series,
+try:
+    from zenml.integrations.pandas.materializers.pandas_materializer import (  # noqa
+        PandasMaterializer,
+    )
+except (ImportError, ModuleNotFoundError) as e:
+    raise IntegrationError(
```
schustmi, line 30: Same as above, IMO wrong error type

src/zenml/zen_server/utils.py
schustmi, line 32: Why was this import moved inside the function?

schustmi (changes requested):

src/zenml/zen_server/utils.py
bcdurak: The `zenml.zen_server.utils` gets imported from numerous modules, whereas the `secure` package is only used by a function that is used by the `[server]` extra. Moving this inside the function allows me to move the `secure` dependency to the extra.

src/zenml/integrations/huggingface/__init__.py
bcdurak: Good catch! Fixed this.

src/zenml/integrations/pandas/materializers/pandas_materializer.py
bcdurak: I have not changed the implementation itself. Just moved it to an integration. If we want to offer other options, I would handle this in a separate PR.

src/zenml/materializers/pandas_materializer.py
bcdurak: Makes sense. Changed it accordingly.

src/zenml/materializers/numpy_materializer.py
bcdurak: Changed it!

src/zenml/materializers/pandas_materializer.py
bcdurak: Changed it!

tests/integration/integrations/deepchecks/materializers/test_deepchecks_result_materializer.py
```diff
@@ -13,47 +13,53 @@
...
         pd.DataFrame(data={"x": [1, 2, 3, 4, 5, 6, 7, 8, 9]}), label=None
     )
     test_dataset = Dataset(pd.DataFrame(data={"x": [1, 2, 3]}), label=None)
     return DatasetSizeComparison().run(train_dataset, test_dataset)
 
 
+@pytest.mark.skipif(
+    sys.version_info.minor == 12,
+    reason="The deepchecks integrations is not yet supported on 3.12.",
+)
```
stefannica, line 54: Why did you have to upgrade deepchecks if it's not even supported on 3.12 ?

scripts/install-zenml-dev.sh
```diff
@@ -28,7 +28,9 @@ parse_args () {
 
 install_zenml() {
     # install ZenML in editable mode
-    uv pip install $PIP_ARGS -e ".[server,templates,terraform,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,dev,mlstacks]"
+    uv pip install $PIP_ARGS -e ".[server,templates,terraform,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,dev]"
+    # TODO: Remove and add the extra back in
+    uv pip install $PIP_ARGS git+https://github.com/zenml-io/mlstacks.git@feature/upgrade-python-3.12
```
stefannica, line 33: Please remember to remove this

src/zenml/integrations/deepchecks/__init__.py
```diff
@@ -36,21 +35,42 @@ class DeepchecksIntegration(Integration):
...
         "torchvision>=0.14.0",
-        "pandas<2.0.0",
         "opencv-python==4.5.5.64",  # pin to same version
         "opencv-python-headless==4.5.5.64",  # pin to same version
         "tenacity!=8.4.0",  # https://github.com/jd/tenacity/issues/471
+        # TODO: Fix the explanation here
+        # Normally, the deepchecks integrations requires pandas to work.
+        # However, their version 0.8.0 is using a pandas function which got
+        # removed at pandas 2.1.0. Until we can upgrade the deepchecks
+        # requirement, we have to limit pandas to <2.1.0.
```
stefannica, line 47: Why did you have to upgrade deepchecks to >=0.18.0 then ?

stefannica (changes requested):
Overall looks good, but I had some nits and questions here and there.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8744,6 +8744,10 @@ def _get_resource_references(
                 # For Python versions <3.9, leave out the third parameter to
                 # _evaluate
                 target_schema = schema_ref._evaluate(vars(zenml_schemas), {})
+            elif sys.version_info >= (3, 12, 4):
```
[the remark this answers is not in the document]
bcdurak: > With CPython 3.12.4/3.13, the function signature of FutureRef._evaluate changed such that recursive_guard is no longer a positional argument; it is now keyword only.

As mentioned [here in this PR](https://github.com/pydantic/pydantic/pull/9612), the signature of `ForwardRef._evaluate` has changed. As we were handling the call ourselves here, I had to adapt our code. `pydantic` had a similar issue (when paired with Python 3.12.4 and above) and it seems like they solved it on 2.7.4. This is why I also bumped our `pydantic` dependency.

src/zenml/integrations/deepchecks/validation_checks.py
```diff
@@ -176,9 +176,6 @@ class DeepchecksDataIntegrityCheck(DeepchecksValidationCheck):
     TABULAR_FEATURE_LABEL_CORRELATION = source_utils.resolve(
         tabular_checks.FeatureLabelCorrelation
     ).import_path
-    TABULAR_IDENTIFIER_LEAKAGE = source_utils.resolve(
-        tabular_checks.IdentifierLeakage
-    ).import_path
```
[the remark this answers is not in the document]
bcdurak: You are absolutely right. As we discussed, I will create a ticket for updating the `deepchecks` integration fully.

tests/integration/integrations/deepchecks/materializers/test_deepchecks_result_materializer.py
bcdurak: It is due to the `pandas` upgrade. The older version of `deepchecks` that we were using (regardless of the python version) had an uncapped `pandas` dependency. Because of this, even when you do `pip install deepchecks`, it ends up installing a version of `pandas` that is not compatible with this older version of `deepchecks`.

scripts/install-zenml-dev.sh
bcdurak: Thank you for the reminder. I had to add one more of this to the `small-checks` workflow in the CI, unfortunately. We have to keep them until we can release a new version of `mlstacks`. I will keep this thread unresolved as a reminder to myself.

src/zenml/integrations/deepchecks/__init__.py
bcdurak: This was a leftover message which I now removed. I hope it is clear from the comment above why we had to upgrade the `deepchecks` dependency.

stefannica (approved):

src/zenml/zen_server/utils.py
schustmi: I'm still not sure about this. `zenml.zen_server.utils` should only be used in the server, so it should assume that the server extra is installed. If that is not the case, we should instead fix our endpoints IMO

bcdurak: > zenml.zen_server.utils should only be used in the server.

Overall, I agree with this statement. However, it is used by quite a lot of files at this time. While a significant chunk of them are endpoints, the rest consist mostly of some type-checking stuff, some CLI and`plugin_registry` code. Instead of checking whether a rework would be necessary on all of these files, moving the import inside the function (and moving the dependency) seemed like a more efficient solution here unless you see a problem that it would create due to this move.

schustmi: It does not create a problem, it's still the wrong solution IMO. `Any` returns are always bad for type checking, and in this case we could avoid them quite easy I think: Just moving the two secure header functions into a separate file, which will then only get imported by the server.

bcdurak: I liked the idea of separating them into a different file 👍  Implemented it, feel free to check.

src/zenml/zen_server/zen_server_api.py
```diff
@@ -78,15 +78,17 @@
     webhook_endpoints,
     workspaces_endpoints,
 )
+from zenml.zen_server.secure_headers import (
+    initialize_secure_headers,
+    secure_headers,
+)
```
avishniakov, line 84: Nice move 👍🏼

avishniakov (approved):

schustmi (approved):

scripts/install-zenml-dev.sh
bcdurak: @stefannica Just so you know, this is going to be a bit tricky as I can not remove these lines with this PR. 

Right now, the changes that I made are not compatible with the latest release of `mlstacks`(0.9.0). If I remove them now, all the subsequent CI runs on other PRs will fail. 

We can potentially release a new version of `mlstacks` as a solution but if we do that `mlstacks` will be broken until we release a new version of `zenml`. (There won't be any `zenml` versions which can work with this new `mlstacks` release.)

So, my proposal would be something like this:

1. I merge this PR while keeping `install-zenml-dev` and `small-checks` pointing to my upgrade branch on the `mlstacks` repo.
2. I create a follow-up PR which actually reverts them back.
3. Just before we release a new version of `zenml`, we do the `mlstacks` release, merge the PR from point 2 and proceed with the release.

This needs to be communicated to the product team as well, I will take over that.

PR #2966, opened 2024-08-29
Added some adjustments for colab

AlexejPenner:
## Describe changes
Colab needs to restart after pip installation of requirements.

Also changed:

`python run.py --model_type=t5-small --orchestration_environment aws` to 
`python run.py --model_type=t5-small --config_path configs/training_aws.yaml` to make it more clear that this does no stack switching, instead its simply about the config file. But the best thing is, you don't need to specify the config file at all.

Added docker settings to the default config file to allow stack switching with 0 change.

## Pre-requisites

## Types of changes


htahir1 (commented):
One thing that really sort of sucks in colab right now as well on top of this change is that there is a pip install zenml command which restarts the environment and then a git clone command and pip install -r requirements.txt command that again restarts the env and then there is a third block that restarts the notebook explcitiy! maybe we can put it all in one block?

AlexejPenner:
> One thing that really sort of sucks in colab right now as well on top of this change is that there is a pip install zenml command which restarts the environment and then a git clone command and pip install -r requirements.txt command that again restarts the env and then there is a third block that restarts the notebook explcitiy! maybe we can put it all in one block?

Yeah was thinkgin about this as well, just wondering, if we should then already ask for thetype of cloud at the very top

htahir1:
@AlexejPenner good idea!

htahir1 (approved):
Hope its tested!

AlexejPenner:
The current setup suddenly fails in colab after requirements.txt files have been installed:

zenml connect -> `Error : Failed to execute cell. Could not send execute message to runtime: [object Event].`

Investigating this

AlexejPenner:
> Hope its tested!

I will do one final sweep over all 12 supported permutations before clicking merge

so far this is promissing

AlexejPenner:
@htahir1 what do you think? merge to main and rebase of develop on top of this?

htahir1:
@AlexejPenner  sure why not

examples/quickstart/configs/training_azure.yaml
```diff
@@ -1,8 +1,12 @@
...
 # Environment configuration
 settings:
   docker:
     parent_image: "zenmldocker/zenml-public-pipelines:quickstart-0.65.0-py3.11-azure"
     skip_build: True
+    # requirements: requirements.txt  # Uncomment this to add your own dependencies
+
 # Uncomment the following two lines to specify the accelerator for your skypilot vm orchestrator
 #  orchestrator.vm_azure:
 #    instance_type: 'Standard_NC6'
```
stefannica, line 12: FYI, skypilot is no longer the only orchestrator option for Azure, nor is it deployed by default anymore with the Azure one-click and terraform stack deployments. It was replaces by the AzureML orchestrator. If you need an equivalent configuration for AzureML, let me know.

examples/quickstart/run.py
```diff
@@ -74,22 +70,32 @@ def main(
...
     """
-    config_folder = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)),
-        "configs",
-    )
-
+    client = Client()
     run_args_train = {}
 
+    crf = client.active_stack.container_registry.flavor
```
stefannica, line 79: The container registry can be missing from the active stack. How is mypy not even complaining here ?

examples/quickstart/run.py
```diff
@@ -74,22 +70,32 @@ def main(
...
+
+    if not config_path:
+        if crf == "aws" and asf == "s3" and orchf == "sagemaker":
+            config_path = "configs/training_aws.yaml"
+        elif crf == "gcp" and asf == "gcp" and orchf == "vertex":
+            config_path = "configs/training_gcp.yaml"
+        elif crf == "azure" and asf == "azure" and orchf == "azureml":
+            config_path = "configs/training_azure.yaml"
+        else:
+            config_path = "configs/training_default.yaml"
```
stefannica, line 95: Thank you, this auto-detection logic really hits the sweet spot and keeps things simple for users!

examples/quickstart/run.py
```diff
@@ -53,14 +51,12 @@
...
-    type=click.Choice(["local", "aws", "gcp", "azure"], case_sensitive=False),
-    default="local",
-    help="Choose the orchestration environment.",
+    "--config_path",
+    help="Choose the configuration file.",
 )
 def main(
     model_type: str,
-    orchestration_environment: str,
+    config_path: str,
```
stefannica, line 59: If this argument is optional, shouldn't this be `Optional[str] = None` instead ?

examples/quickstart/run.py
```diff
@@ -74,22 +70,32 @@ def main(
...
+    if not config_path:
+        if crf == "aws" and asf == "s3" and orchf == "sagemaker":
+            config_path = "configs/training_aws.yaml"
+        elif crf == "gcp" and asf == "gcp" and orchf == "vertex":
+            config_path = "configs/training_gcp.yaml"
+        elif crf == "azure" and asf == "azure" and orchf == "azureml":
+            config_path = "configs/training_azure.yaml"
+        else:
+            config_path = "configs/training_default.yaml"
+    print(f"Using {config_path} to configure the pipeline run.")
```
stefannica, line 96: While reviewing this, I had another idea: you should warn people if they use the aws/gcp/azure config files explicitly with a stack that doesn't match and if they haven't set `skip_build: False`. This will keep them from running into fuzzy errors later on because some integration is missing.

examples/quickstart/run.py
stefannica: On the other hand, I would recommend not even checking the container registry here at all, because the parent container image is only relevant for infrastructure-related stack components that will have code running inside the container itself: artifact stores, orchestrators and step operators.

examples/quickstart/run.py
```diff
@@ -74,22 +70,32 @@ def main(
...
+
     pipeline_args = {}
     if no_cache:
         pipeline_args["enable_cache"] = False
-    pipeline_args["config_path"] = os.path.join(
-        config_folder, f"training_{orchestration_environment}.yaml"
-    )
+
+    if not config_path:
+        if crf == "aws" and asf == "s3" and orchf == "sagemaker":
```
stefannica, line 88: I would also allow `local` orchestrator here as a valid option and ignore the container registry (see above). You might even want to check that the step operator, if present, is also `sagemaker`.

examples/quickstart/run.py
```diff
@@ -74,22 +70,32 @@ def main(
...
+            config_path = "configs/training_aws.yaml"
+        elif crf == "gcp" and asf == "gcp" and orchf == "vertex":
+            config_path = "configs/training_gcp.yaml"
```
stefannica, line 91: same here: allow `local` orchestrator here as a valid option, ignore the container registry and check the step operator, if present.

examples/quickstart/run.py
```diff
@@ -74,22 +70,32 @@ def main(
...
+        elif crf == "azure" and asf == "azure" and orchf == "azureml":
```
stefannica, line 92: same here

stefannica (changes requested):
Love the new feel ! Much easier to work with.

examples/quickstart/run.py
AlexejPenner: good catch

examples/quickstart/run.py
AlexejPenner: not sure I catch your drift, where should we allow local orchestrator - that one should use the default config

stefannica: I was thinking of cases in which someone uses the sagemaker step operator with a local orchestrator for example, but you already covered this.

stefannica (approved):

PR #2978, opened 2024-09-02
Prevent extra attributes in component configs

schustmi:
## Describe changes
Prevent extra attributes in the config classes of our stack components. Previously, this was defined in the `StackComponentConfig` parent class, but due to the inheritance order lots of our stack components allowed extra attributes. This PR fixes that by making sure extras are not allowed during stack component config validation.
- I also thought about changing the inheritance order of all config classes. However I think it is very unreliable and might lead to newly created stack components having the same issues again. It would also lead to crashes for existing stack components that were registered with a "faulty" config, which my solution does not. It will only fail if users now try to update such a faulty component, at which point they will have to remove the extra attributes.
- Potential issues with my implementation: If some stack component would require `extra="allow"` in their configuration for some reason, this would now be ignored and extras will not be allowed. I can't think of any case where this would be necessary though.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/stack/utils.py
+++ b/src/zenml/stack/utils.py
@@ -75,7 +75,16 @@ def validate_stack_component_config(
             return None
         raise
 
-    configuration = flavor_class.config_class(**configuration_dict)
+    config_class = flavor_class.config_class
+    # Make sure extras are forbidden for the config class. Due to inhertance
+    # order, some config classes allow extras by accident which we patch here.
+    validation_config_class = type(
+        config_class.__name__,
+        (config_class,),
+        {"model_config": {"extra": "forbid"}},
+    )
+    configuration = validation_config_class(**configuration_dict)
+
     if not configuration.is_valid:
         raise ValueError(
             f"Invalid stack component configuration. Please verify "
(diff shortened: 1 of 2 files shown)


src/zenml/stack/utils.py
```diff
@@ -75,7 +75,16 @@ def validate_stack_component_config(
             return None
         raise
 
-    configuration = flavor_class.config_class(**configuration_dict)
+    config_class = flavor_class.config_class
+    # Make sure extras are forbidden for the config class. Due to inheritance
```
bcdurak, line 79: As you mentioned in the description, for whatever reason, if a user sets the `extra` to `allow`, perhaps we can use a warning or a debug message here.

bcdurak (approved):
Looks good!

If I understood correctly, people will still be able to use their `Stack` and `StackComponent` objects in Python with "faulty" configs for the time being. This is just blocking them from creating new "faulty" components or updating them in a bad way.  We should maybe look into incorporating this into our design at some point in the future. 

Added some small comments, feel free to ignore.

src/zenml/stack/utils.py
schustmi: The problem is, I can't even detect that right? Whether someone did it explicitly or whether it happened due to inheritance of `BaseSettings`

bcdurak: True. Feel free to skip it.

PR #2982, opened 2024-09-04
Small dependency and docs updates

strickvl:
Includes updates for:

- Prodigy (added a link to docs) 
- Feast (opened up the dependency pin for the main package)
- I updated the Tensorflow dataset materializer as the method we were using to save and load was deprecated
- Fixed the Label Studio docs which were missing a `port` config setting as well as confusing around the use of `authentication_secret`

All were tested with Python 3.12 and work fine. (Tensorflow doesn't work with Python 3.12, so this was tested with 3.11).


avishniakov (approved):

bcdurak (commented):
Looks good but I would run the `ci-slow` before merging this.

PR #2984, opened 2024-09-04
Improve ResourceSettings flexibility

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/resource_settings.py
+++ b/src/zenml/config/resource_settings.py
@@ -16,10 +16,20 @@
 from enum import Enum
 from typing import Optional, Union
 
-from pydantic import Field, NonNegativeInt, PositiveFloat
-from pydantic_settings import SettingsConfigDict
+from pydantic import ConfigDict, Field, NonNegativeInt, PositiveFloat
 
 from zenml.config.base_settings import BaseSettings
+from zenml.utils.enum_utils import StrEnum
+
+# Vertex step operator
+# ['ACCELERATOR_TYPE_UNSPECIFIED', 'NVIDIA_TESLA_K80', 'NVIDIA_TESLA_P100', 'NVIDIA_TESLA_V100', 'NVIDIA_TESLA_P4', 'NVIDIA_TESLA_T4', 'NVIDIA_TESLA_A100', 'NVIDIA_A100_80GB', 'NVIDIA_L4', 'NVIDIA_H100_80GB', 'TPU_V2', 'TPU_V3', 'TPU_V4_POD', 'TPU_V5_LITEPOD']
+
+
+class AcceleratorType(StrEnum): ...
+
+
+# Or MachineType?
+class InstanceType(StrEnum): ...
 
 
 class ByteUnit(Enum):
@@ -73,6 +83,14 @@ class ResourceSettings(BaseSettings):
     gpu_count: Optional[NonNegativeInt] = None
     memory: Optional[str] = Field(pattern=MEMORY_REGEX, default=None)
 
+    accelerator: Optional[Union[AcceleratorType, str]] = Field(
+        union_mode="left_to_right", default=None
+    )
+    accelerator_count: Optional[NonNegativeInt] = None
+    instance_type: Optional[Union[InstanceType, str]] = Field(
+        union_mode="left_to_right", default=None
+    )
+
     @property
     def empty(self) -> bool:
         """Returns if this object is "empty" (=no values configured) or not.
@@ -115,9 +133,8 @@ class ResourceSettings(BaseSettings):
             # Should never happen due to the regex validation
             raise ValueError(f"Unable to parse memory unit from '{memory}'.")
 
-    model_config = SettingsConfigDict(
+    model_config = ConfigDict(
         # public attributes are immutable
         frozen=True,
-        # prevent extra attributes during model initialization
-        extra="forbid",
+        extra="allow",
     )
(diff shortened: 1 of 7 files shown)


src/zenml/config/resource_settings.py
```diff
@@ -16,10 +16,20 @@
...
 from typing import Optional, Union
 
-from pydantic import Field, NonNegativeInt, PositiveFloat
-from pydantic_settings import SettingsConfigDict
+from pydantic import ConfigDict, Field, NonNegativeInt, PositiveFloat
 
 from zenml.config.base_settings import BaseSettings
+from zenml.utils.enum_utils import StrEnum
+
+# Vertex step operator
```
bcdurak, line 24: Probably makes sense to remove this one here, right?

src/zenml/integrations/gcp/flavors/vertex_step_operator_flavor.py
```diff
@@ -28,12 +30,21 @@
...
-class VertexStepOperatorSettings(BaseSettings):
+class VertexStepOperatorResources(StackComponentResourceSettings):
+    accelerator: Optional[str] = None
+    accelerator_count: int = 0
+    instance_type: str = "n1-standard-4"
+    boot_disk_size_gb: int = 100
+    boot_disk_type: str = "pd-ssd"
+
+
+class VertexStepOperatorSettings(VertexStepOperatorResources):
```
bcdurak, line 47: Little typo down below (happens twice) "Check out out this table"

bcdurak: Might be nice to update the docstring as well.

src/zenml/config/stack_component_resource_settings.py
```diff
@@ -0,0 +1,27 @@
+from typing import Set
+
+from zenml.config.base_settings import BaseSettings
+
+
+class StackComponentResourceSettings(BaseSettings):
```
bcdurak, line 6: I am a bit confused here. Initially, I thought the implementation for the stack component resource settings would also inherit from the `ResourceSettings`. I am probably missing out on something here. Could you perhaps explain the relationship between the `ResourceSettings` and `StackComponentResourceSettings`?

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
```diff
@@ -220,19 +209,17 @@ def launch(
...
             "display_name": info.run_name,
             "job_spec": {
                 "worker_pool_specs": [
                     {
                         "machine_spec": {
-                            "machine_type": settings.machine_type,
-                            "accelerator_type": settings.accelerator_type,
-                            "accelerator_count": accelerator_count
-                            if settings.accelerator_type
+                            "machine_type": settings.instance_type,
```
bcdurak, line 219: Just a personal note here (feel free to ignore): I am still a bit unsure about this approach here. For instance, the configuration here is called `machine_type` in the Vertex nomenclature and we are calling it `instance_type` now. It is not too bad but it feels a bit unintuitive for me.

src/zenml/config/stack_component_resource_settings.py
schustmi: I wanted to make it inherit from that, but that would defeat the purpose because then they'd inherit even the attributes that are not possible to specify for this component.

Or do you think it's better to make them inherit from it, and the have a class var or something that defines which are ignored?

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
schustmi: I have the same concern actually and also talked to @htahir1, and I think we're going with a unified name for now

htahir1: What would be an alternative?

schustmi: We use the name that the tool (Vertex, Sagemaker, Skypilot, ...) uses. From what I've seen, it's "machine_type" for vertex, "size" for azure (which is just unintuitive) and instance_type for the rest

htahir1: but then you can specify "size" for vertex right? or "machine_type" for azure?

schustmi: No, in this case this would not be a general item on `ResourceSettings` (-> no autocomplete) but would be one of the extras that users could specify (still using `ResourceSettings`, just no explicit attribute)

htahir1: In this case, id go with clear docs and a unified attribute. WDYT @bcdurak ?

htahir1: yes lets do a unified attribute and clear docs

src/zenml/config/stack_component_resource_settings.py
schustmi: @bcdurak Any opinion on this? It's either:
- We inherit from `ResourceSettings`, and somehow have a class var or similar that defines which resource settings are not available. In that case, we can even make it fail with some validator if one of these is passed using this explicit class?
- We don't inherit from it. I guess we save this class var, and also users would not get autocomplete for attributes that are not available.

schustmi: Some more points:
- Inheritance: If a new attribute gets added to resource settings, we need to add the "not supported flags" to all orchestrators/step operators.
- No inheritance: We need to redefine all resource settings attributes on all the subclasses that support them

PR #2991, opened 2024-09-09
Remove references to workspaces from docs

strickvl:
I removed all the remaining references to workspaces from our documentation.


htahir1 (approved):
Great - most of these are logs though. Maybe you can take out the source logs from the code as well?

schustmi (approved):

PR #3003, opened 2024-09-11
Allow env variable placeholders in configurations

schustmi:
## Describe changes

Allows environment variables placeholders in the pipeline/step configurations in code as well as in config files.
The format will be `${ENV_VARIABLE_NAME}` and will be replaced at compilation time. This makes sure it is only happening client side and can't be used to read server environment variables when running a pipeline from the server.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/env_utils.py
+++ b/src/zenml/utils/env_utils.py
@@ -14,7 +14,16 @@
 """Utility functions for handling environment variables."""
 
 import os
-from typing import Dict, List, Optional, cast
+import re
+from typing import Any, Dict, List, Optional, TypeVar, cast
+
+from zenml.logger import get_logger
+from zenml.utils import string_utils
+
+logger = get_logger(__name__)
+
+V = TypeVar("V", bound=Any)
+ENV_VARIABLE_PLACEHOLDER_PATTERN = re.compile(pattern=r"\$\{([a-zA-Z0-9_]+)\}")
 
 ENV_VAR_CHUNK_SUFFIX = "_CHUNK_"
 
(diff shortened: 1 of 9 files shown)


schustmi:
@htahir1 This works, but made me wonder: Do we want this just for the config file, or for any value configured on a pipeline? For example, should these placeholders also be allowed for stack component settings defined on the step decorator? Or really just the config file?

src/zenml/utils/env_utils.py
```diff
@@ -99,3 +108,36 @@ def reconstruct_environment_variables(
...
+        key = match.group(1)
+        if value := os.getenv(key):
+            return value
+        else:
+            logger.debug(
+                "Unable to substitute environment variable placeholder %s "
+                "because the environment variable is not set, using an empty "
+                "string instead.",
+                key,
+            )
```
avishniakov, line 133: Hm, here I have doubts... I know that Unix systems operate like that, but I don't like that, take `rm -rf $I_DONT_EXISTS` as an example - bad things might happen :)
Maybe raising and letting the user sort this out is safer?

avishniakov (changes requested):
One question besides my comments: this works not only for `extra`, right? You used extra all over the place, but from the code it seems to be anywhere allowed, right?

src/zenml/utils/env_utils.py
schustmi: Don't have a strong opinion on this personally, but in most cases this will be harmless and worst case lead to a failed pipeline run, correct?

schustmi:
@avishniakov Yes correct, this works for all configs inside the yaml file, the pipeline and step decorators or values passed to `.configure(...)` or `.with_options(...)`

src/zenml/utils/env_utils.py
avishniakov: Yes, but it will be harder to debug at the end, IMO. I will give you a green light here, but I would love to see an explicit raise here 🙂

avishniakov (approved):

src/zenml/utils/env_utils.py
schustmi: Raised a `KeyError` now but made it configurable on the util function at least in case we decide to change it at some point.

PR #3005, opened 2024-09-11
Enhancing the orchestrator UX for major cloud providers

bcdurak:
## Describe changes

This PR aims to solve two problems at the same time:

1. Provide more meaningful URLs (Pipeline UIs and collection of Logs), possibly even before the first step starts. 
2. Implement a way to update the status of a run

### For task number 1 (URLs)

- Adjusted the `run_utils.deploy_pipeline` and `Stack.deploy_pipeline` calls to pass the `placeholder_run` forward.
- Adjusted the `BaseOrchestrator.run` to accept the placeholder run. During this call, we fetch everything that `prepare_or_run_pipeline` `yield`s and we publish it as metadata.
- Adjusted the `azureml`, `sagemaker` and `vertex` orchestrators to use their respective clients (and created `Job` objects) to `yield` their `orchestrator_run_id`s, `orchestrator_log_url` and `orchestrator_url`s preemptively if possible. This helps us to display the URLs, even before the first step starts.

### For task number 2 (Pipeline Status Refresh)

- Added three new extras to the ZenML package: `sagemaker`, `azureml` and `vertex`.
- Adjusted the `base.Dockerfile` and `zenml-server-dev.Dockerfile` to install these new extras as well.
- Added a `fetch_status` method to the `BaseOrchestrator` implementation, which on default throws a `NotImplementedError`.
- Added a `refresh_status` to the `PipelineRunResponse` model which makes use of the aforementioned `fetch_status` method.
- Added the corresponding `fetch_status` implementation for `vertex`, `sagemaker`, and `azureml` orchestrators.
- Added a new endpoint `runs/<RUN_ID>/refresh` which refreshes the status of a pipeline run if possible.
- Parameterized the old `Get Run` endpoint with a `refresh_status` flag that will return a response with a refreshed status if possible.
- Added a corresponding CLI command (`zenml pipeline runs refresh <RUN_ID>`) to refresh the status of a pipeline manually.

## Docs Updates

- Fixed the order of instructions in the Azure Guide page.
- Added new images and instructions for role assignments on Azure.
- Fixed a command in the GCP terraform deployment.
- Added the new `refresh` CLI command to the CLI docs

## TODOs

- Adjust the pipeline run cleanup on the placeholder run to happen at a later point: Handling this on a different PR

## Pre-requisites
- [X] If my changes require changes to the dashboard, these changes are communicated/requested. **(NOT YET DONE)**

## Types of changes


bcdurak:
I do have a question for both of you @schustmi @stefannica. If you check the changes in the `pyproject.toml`, you will see that I have added three new extras, `sagemaker`, `vertex`, and `azureml` respectively. We use these extras to install the requirements when we set up a server (like we do with the other cloud requirements). 

Now, for `GCP` for instance, we have `connectors-gcp`, `secrets-gcp`, `gcsfs`, and now the `vertex` as extras to the `zenml` package. When we build a server image, we install all of these extras together. At this point, I was considering whether it would be cleaner to merge them under a single `gcp` extra or use `zenml` to install the `gcp` integration. What do you think?

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -267,3 +284,26 @@ def get_run_status(
...
+    run_id: UUID,
+    _: AuthContext = Security(authorize),
+) -> None:
+    """Refreshes the status of a specific pipeline run.
+
+    Args:
+        run_id: ID of the pipeline run to refresh.
+    """
+    # Verify access to the run
+    run = verify_permissions_and_get_entity(
```
schustmi, line 304: This should probably also check permissions for the orchestrator?

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -308,6 +310,52 @@ def produced_artifact_versions(self) -> List["ArtifactVersionResponse"]:
...
+        orchestrator = cast(
+            BaseOrchestrator,
+            StackComponent.from_model(component_model=component_model),
+        )
+
+        # Fetch the status
+        status = orchestrator.fetch_status(run=self)
+
+        # If it is different from the current status, update it
+        if status != self.status:
```
schustmi, line 347: I think we could instead check if the run is in initializing/running, and only update the status in this case?
If the run is already in completed/failed, there is no refreshing to be done. Additionally, it might even run into some weird edge cases like a run that failed on the orchestrator side but actually finished successfully for ZenML (maybe something during the teardown failed)

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -107,13 +113,24 @@ def get_run(
...
         hydrate: Flag deciding whether to hydrate the output model(s)
             by including metadata fields in the response.
+        refresh_status: Flag deciding whether we should try to refresh
+            the status of the pipeline run using its orchestrator.
 
     Returns:
         The pipeline run.
     """
-    return verify_permissions_and_get_entity(
+    run = verify_permissions_and_get_entity(
```
schustmi, line 122: Same with the permissions for the orchestrator

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -434,3 +480,202 @@ def get_pipeline_run_metadata(
         )
         run_metadata[METADATA_ORCHESTRATOR_URL] = Uri(orchestrator_logs_url)
         return run_metadata
+
+    @staticmethod
+    def _dissect_pipeline_execution_arn(
```
schustmi, line 485: I think this is duplicated with the function at the top of this file?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -434,3 +480,202 @@ def get_pipeline_run_metadata(
...
+        """Generate run metadata based on the generated Sagemaker Execution.
+
+        Args:
+            execution: The corresponding _PipelineExecution object.
+
+        Yields:
+            A dictionary of metadata related to the pipeline run.
+        """
+        # Metadata
+        metadata: Dict[str, MetadataType] = dict()
```
schustmi, line 580:
```suggestion
        metadata: Dict[str, MetadataType] = {}
```

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -434,3 +480,202 @@ def get_pipeline_run_metadata(
...
+        if status in ["Executing", "Stopping"]:
+            return ExecutionStatus.RUNNING
+        elif status in ["Stopped", "Failed"]:
+            return ExecutionStatus.FAILED
+        elif status in ["Succeeded"]:
+            return ExecutionStatus.COMPLETED
+        else:
+            raise ValueError("Unknown status for the pipeline execution.")
+
+    def generate_metadata(
```
schustmi, line 568: Feel free to ignore, but to me these methods don't generate something, they instead compute something.

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
```diff
@@ -428,3 +439,145 @@ def get_pipeline_run_metadata(
...
+        """Generate run metadata based on the generated AzureML PipelineJob.
+
+        Args:
+            job: The corresponding PipelineJob object.
+
+        Yields:
+            A dictionary of metadata related to the pipeline run.
+        """
+        # Metadata
+        metadata: Dict[str, MetadataType] = dict()
```
schustmi, line 529:
```suggestion
        metadata: Dict[str, MetadataType] = {}
```

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -778,3 +797,174 @@ def _configure_container_resources(
...
+        """Generate run metadata based on the corresponding Vertex PipelineJob.
+
+        Args:
+            job: The corresponding PipelineJob object.
+
+        Yields:
+            A dictionary of metadata related to the pipeline run.
+        """
+        # Metadata
+        metadata: Dict[str, MetadataType] = dict()
```
schustmi, line 886:
```suggestion
        metadata: Dict[str, MetadataType] = {}
```

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -169,29 +171,41 @@ def run(
...
         """
         self._prepare_run(deployment=deployment)
 
         environment = get_config_environment_vars(deployment=deployment)
 
         try:
-            result = self.prepare_or_run_pipeline(
-                deployment=deployment, stack=stack, environment=environment
-            )
+            if result := self.prepare_or_run_pipeline(
```
schustmi, line 189: The name could be clearer IMO, maybe something like `metadata` or `metadata_iterator`?

schustmi: One other thing that we should check here in case you didn't yet: Other orchestrators should not return anything from this method anymore, otherwise it would fail in the iterating here.

schustmi (changes requested):
One general comment: I remember we talked about only deleting the placeholder run in case it failed during stack validation or other pre-flight checks, but not once it actually got to the orchestrator. I can't seem to find anything implemented in this PR that does this, is this something that we decided to not do for now?

docker/zenml-server-dev.Dockerfile
```diff
@@ -53,14 +53,14 @@ COPY src/zenml/__init__.py ./src/zenml/
 # NOTE: we need to uninstall zenml at the end to remove the incomplete
 # installation
 RUN pip install --upgrade pip \
-  && pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure] \
+  && pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex] \
```
stefannica, line 56: How much larger does this make the container image? When I suggested we do this, I meant doing it for ZenML Pro images only. Doing this for all server images might upset users if it adds significant bulk to the container image.

docker/base.Dockerfile
```diff
@@ -76,7 +76,7 @@ RUN if [ "$ZENML_NIGHTLY" = "true" ]; then \
       PACKAGE_NAME="zenml"; \
     fi \
     && pip install --upgrade pip \
-    && pip install "${PACKAGE_NAME}[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure]${ZENML_VERSION:+==$ZENML_VERSION}" \
+    && pip install "${PACKAGE_NAME}[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex]${ZENML_VERSION:+==$ZENML_VERSION}" \
```
stefannica, line 79: Why do you need to install these in the client image ? You already get them when users install integrations required for the orchestrator in their stack, right ? Installing these for everyone is not only unnecessary but also makes the images bigger (something are users are already complaining about).

stefannica (dismissed):

stefannica:
> I do have a question for both of you @schustmi @stefannica. If you check the changes in the `pyproject.toml`, you will see that I have added three new extras, `sagemaker`, `vertex`, and `azureml` respectively. We use these extras to install the requirements when we set up a server (like we do with the other cloud requirements).
> 
> Now, for `GCP` for instance, we have `connectors-gcp`, `secrets-gcp`, `gcsfs`, and now the `vertex` as extras to the `zenml` package. When we build a server image, we install all of these extras together. At this point, I was considering whether it would be cleaner to 

> merge them under a single `gcp` extra

Please don't merge them. Some of them are overlapping, yes, but they generally are used for different purposes.

> or use `zenml` to install the `gcp` integration. What do you think?

This would be even worse, as running `zenml integration install` is a sure way to introduce pip conflicts in other places.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -434,3 +480,202 @@ def get_pipeline_run_metadata(
...
+            region_name, pipeline_name, execution_id = (
+                dissect_pipeline_execution_arn(pipeline_execution.arn)
+            )
+
+            # Get the Sagemaker session
+            session = pipeline_execution.sagemaker_session
+
+            # List the Studio domains and get the Studio Domain ID
+            # TODO: Solve this with the config
+            domains_response = session.sagemaker_client.list_domains()
```
schustmi, line 618: What's the issue with using any domain? That it won't work? Or lead them to a broken link?
Somehow not a huge fan of having a config attribute just to make a link work

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: I was considering doing the check here but our RBAC implementation changed my mind. AFAIK, if I don't have access to the stack or the orchestrator within, our RBAC system would make sure that these entity is removed from the `RunResponse`, right? This way, I can check for these cases in the root `RunResponse.refresh_run_status()` method. This way I am also saving 2 calls (for the stack RBAC check and the orchestrator RBAC check). What do you think?

src/zenml/models/v2/core/pipeline_run.py
bcdurak: Good idea. I added the condition.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: Absolutely. I moved the method outside but forgot to remove the one in the class. Removed it now, thanks.

src/zenml/zen_server/routers/runs_endpoints.py
schustmi: Ah I think I get what you mean. All submodels from the response without access will be removed yes, but they will just not have a body and will not be entirely gone.
Additionally, I think the orchestrator is not even part of the run response, and gets loaded via some hydration calls -> RBAC is not used in these cases

docker/zenml-server-dev.Dockerfile
bcdurak: @htahir1, @schustmi, and I had a look into this and discussed the potential outcomes. The size of our server image goes up from 242 MB to 336 MB. We all deemed it OK. What's your take on this?

Also, keep in mind, If we do it for ZenML Pro images only, we will not be able to refresh the status of the runs from the server side for our OSS users.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: It would work as long as they have at least one domain. In this case, we take the first domain and then use it to create the link. I am not sure how people are scoping domains though, I lack the Sagemaker UX knowledge here a bit. I thought a config attribute would help them to specify which domain to use but I am also OK with leaving it like this.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: Nice idea, changed them to `compute_XXXX`.

docker/base.Dockerfile
bcdurak: This is for building the server image with the `server-builder`, isn't it? Maybe I am missing something here.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: Yes, I have checked this before and all orchestrators seem to be fine. Applied the renaming suggestion 👍

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: I see. I implemented it accordingly as an additional check for the `orchestrator`. 

PS: The `stack` is already covered as it is a part of the `RunResponse`.

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: Again, the comment above. Resolving this for now.

bcdurak:
@schustmi as a response to your general comment: You are right, I haven't implemented the change in the cleanup logic just yet due to time limitations. I am not sure if I can squeeze it right now but in the worst case, I will handle it in a separate PR.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
schustmi: I guess right now it would just fail if they have 0 domains, if that's a possibility?

Don't have a real opinion here though, do whatever you think makes sense, but if you don't know what makes sense, maybe leave it out for now and see if people complain?

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -307,6 +309,62 @@ def produced_artifact_versions(self) -> List["ArtifactVersionResponse"]:
...
+                "The orchestrator that this pipeline run response was "
+                "executed with has been deleted."
+            )
+
+        orchestrator = cast(
+            BaseOrchestrator,
+            StackComponent.from_model(component_model=orchestrator_list[0]),
+        )
+
+        if self.status in [
```
schustmi, line 348: How about we move this up even above the orchestrator loading?

schustmi (approved):

docker/zenml-server-dev.Dockerfile
htahir1: I mean lets just keep it and see if people complain about the size? why should we add more complexity by splitting up the logic here

src/zenml/models/v2/core/pipeline_run.py
bcdurak: On it.

docker/zenml-server-dev.Dockerfile
stefannica: Yeah, 90 MB increase is probably ok

docker/base.Dockerfile
stefannica: You're right, I misread this, I apologize.

PR #3006, opened 2024-09-12
Allow running a single step from the same cell it was defined in

schustmi:
## Describe changes

If a step was defined and called in the same notebook cell, it would fail to execute as a remote one step pipeline. The reason for this was the notebook cell code being imported in the container, which would start code to run the step which would then internally fail. We just stop this early by using the environment variable that is set to prevent the same thing happening for module-level pipeline calls.

## Pre-requisites

## Types of changes


strickvl (approved):
Looks good, but hope we can get tests in for these kinds of things in the future.

PR #3012, opened 2024-09-13
`darglint` becomes `pydoclint`

strickvl:
Darglint is slow + has been archived as a project for a while.

[`pydoclint`](https://github.com/jsh9/pydoclint) is a replacement, with a few extra tricks around docstring validation. Best of all, it's much faster than darglint (which we had to give its own CI runners, such was the sluggishness), so I've added this back into the linting script (where it used to live, once upon a time).

The `ruff` team / community [are working](https://github.com/astral-sh/ruff/issues/458) to integrate pydoclint into `ruff` proper, but the issue has been open for a while and it doesn't quite yet work the way we want (despite being available on the `preview` feature flag).


src/zenml/integrations/kaniko/image_builders/kaniko_image_builder.py
```diff
@@ -345,7 +345,7 @@ def _delete_pod(self, pod_name: str) -> None:
             pod_name: Name of the Pod to delete.
 
         Raises:
-            subprocess.CalledProcessError: If the kubectl call to delete
+            e: If the kubectl call to delete
```
schustmi, line 348: That's quite a downgrade here in information I'd say. Any way we can add an ignore here and still have the actual error type?

strickvl: Yeah so `pydoclint` native doesn't actually allow these custom ignores at all, unless you use it in ['flake8 plugin mode'](https://jsh9.github.io/pydoclint/how_to_ignore.html). We don't want to add flake8 back in, so I'd say that isn't really an option. `ruff` is slowly adding it in (i.e. use pydoclint as a backend to perform the checks for `DOC-...` error codes) but it is only available in `preview` model and it doesn't currently work too well. So I'd say we can either leave it this way + upgrade later when ruff support is finally stable, or we can just hold off completely until ruff support comes. I think it'd probably be a few months realistically before that's there, going off the pace of the comments etc on the GH issue.

strickvl:
Closing this PR for now until https://github.com/astral-sh/ruff/issues/12434 is implemented on the Ruff side.

PR #3015, opened 2024-09-16
Add the ability to use Github repos that contain symlinks.

aiakide:
## Describe changes
I had noticed that my GitHub repo was having problems encoding files. 
Bildschirmfoto 2024-09-10 um 18 22 10

As it turned out, the GitHub API also provides symlinks as files when they are checked into the repo. 
These files have no encoding information (no decoded_content). 

I added the functionality that symlinks present in a GitHub repo are also created in the local copy of the repo. The symlink information is collected first and only after all files have been downloaded from the GitHub API, the symlinks are created. 

It seems that this error only occurs with the GitHub API. The response of the GitLab API also contains content and encoding information for symlinks. 

```json
{
  "file_name": "latest",
  "file_path": "logs/scheduler/latest",
  "size": 10,
  "encoding": "base64",
  "content_sha256": "c1e9c1c135c34aa0eace8cbc57559af9a250f4fb93882b4bb9bf59c8dd73dfea",
  "ref": "main",
  "blob_id": "2d1f1c07a2151a06900dc48cf22c7d5072057cdb",
  "commit_id": "c41843c07b1499bfea65d2a417d47a472873247d",
  "last_commit_id": "e05ba60199ec18f43691952d4aff4b5da6dfc7f4",
  "execute_filemode": false,
  "content": "MjAyNC0wOS0wOQ=="
}
```

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/github/code_repositories/github_code_repository.py
+++ b/src/zenml/integrations/github/code_repositories/github_code_repository.py
@@ -15,7 +15,7 @@
 
 import os
 import re
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 import requests
 from github import Github, GithubException
@@ -151,7 +151,7 @@ class GitHubCodeRepository(BaseCodeRepository):
             raise RuntimeError("Invalid repository subdirectory.")
 
         os.makedirs(directory, exist_ok=True)
-
+        tmp_symlinks: List[Tuple[str, str]] = []
         for content in contents:
             local_path = os.path.join(directory, content.name)
             if content.type == "dir":
@@ -160,12 +160,25 @@ class GitHubCodeRepository(BaseCodeRepository):
                     directory=local_path,
                     repo_sub_directory=content.path,
                 )
+            elif content.type == "symlink":
+                symlink_content = self.github_repo.get_contents(
+                    content.path, ref=commit
+                )
+                symlink_target = symlink_content.raw_data["target"]
+                tmp_symlinks.append((local_path, symlink_target))
+                # As it cannot be assumed at this point that the targets of the
+                # symlink already exist, the symlinks are first collected here and processed later.
             else:
                 try:
                     with open(local_path, "wb") as f:
                         f.write(content.decoded_content)
                 except (GithubException, IOError) as e:
                     logger.error("Error processing %s: %s", content.path, e)
+        for symlink in tmp_symlinks:
+            symlink_source, symlink_target = symlink
+            create_symlink_in_local_repo_copy(
+                symlink_source=symlink_source, symlink_target=symlink_target
+            )
 
     def get_local_context(self, path: str) -> Optional[LocalRepositoryContext]:
         """Gets the local repository context.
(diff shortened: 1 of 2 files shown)


src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -202,3 +215,55 @@ def check_remote_url(self, url: str) -> bool:
...
+        FileNotFoundError: Informs that the target directory specified by
+            `symlink_target` does not exist.
+        PermissionError: Informs that there are insufficient permissions to
+            create the symbolic link.
+        NotImplementedError: Informs that symbolic links are not supported on
+            the current operating system.
+        OSError: Any other OS-related errors that occur.
+    """
+    try:
+        if os.path.exists(symlink_source):
```
schustmi, line 242: For this to happen, there would have to be another file in the GH repo with the same name as the symlink file, correct? Is that actually possible?

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -202,3 +215,55 @@ def check_remote_url(self, url: str) -> bool:
             return True
 
         return False
+
+
+def create_symlink_in_local_repo_copy(
+    symlink_source: str, symlink_target: str
+) -> None:
+    """This function attempts to create a symbolic link at `local_path` that points to `symlink_target`.
```
schustmi, line 223:
```suggestion
    """This function attempts to create a symbolic link at `symlink_source` that points to `symlink_target`.
```

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -202,3 +215,55 @@ def check_remote_url(self, url: str) -> bool:
...
+
+    If a file or directory already exists at `local_path`, it will
```
schustmi, line 225:
```suggestion
    If a file or directory already exists at `symlink_source`, it will
```

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -202,3 +215,55 @@ def check_remote_url(self, url: str) -> bool:
             return True
 
         return False
+
+
+def create_symlink_in_local_repo_copy(
```
schustmi, line 220: I think it would be nice to have a check whether the symlink target is actually part of the repository contents. In other cases, I feel like we should not create the symlink at all.

schustmi (changes requested):

src/zenml/integrations/github/code_repositories/github_code_repository.py
aiakide: This can occur if an error occurs in the further course of cloning the repository and another attempt is made to clone the repository to the same target directory. However, this is only the case if you clone to a permanent directory, but not when creating the Docker images. At least that's what I would assume.

aiakide: You could also solve this behavior with an exception. It could then look like this. 
```python 

    try:
        os.symlink(src=symlink_src, dst=symlink_dst)

    except FileExistsError as e:
        logger.debug("The symbolic link already exists. %s",e)
        
      ....
```

src/zenml/integrations/github/code_repositories/github_code_repository.py
aiakide: I have renamed the attributes to `symlink_dst` and `symlink_src`, corresponding to the attribute name of `os.path.symlink`.

src/zenml/integrations/github/code_repositories/github_code_repository.py
aiakide: What do you mean exactly? 
Currently the function is only called when a symlink is found in the repo content.

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -202,3 +215,51 @@ def check_remote_url(self, url: str) -> bool:
...
+    symlink_dst: str, symlink_src: str
+) -> None:
+    """This function attempts to create a symbolic link at `symlink_dst` that points to `symlink_src`.
+
+    If a file or directory already exists at `symlink_dst`, it will
```
schustmi, line 225: This is not true anymore I think?

src/zenml/integrations/github/code_repositories/github_code_repository.py
```diff
@@ -202,3 +215,51 @@ def check_remote_url(self, url: str) -> bool:
...
+    be removed before the symbolic link is created.
+
+    Args:
+        symlink_dst: The path where the symbolic link should be created.
+        symlink_src: The path that the symbolic link should point to.
+
+    Raises:
```
schustmi, line 232: To pass our CI, we'll need to remove those as they're actually not raised anymore, but a warning is logged instead.

schustmi (changes requested):

src/zenml/integrations/github/code_repositories/github_code_repository.py
schustmi: What I meant here is: The symlink source could be a path which is outside of the contents of the repository, for example like this:

```bash
echo outside > outside.txt
mkdir repo
cd repo
git init
ln -s ../outside.txt outside_symlink
```

In most cases, when downloading the symlink the source will not exist, but in some other cases it might lead to unexpected behaviour -> I would suggest we just prevent this case alltogether and don't try to create a symlink if the file is outside of the directory we download. What do you think?

schustmi:
Hey @aiakide, are you still planning on getting this PR merged?

schustmi:
@aiakide Well I don't want to give you any arbitrary deadline, but we're going to do the next release some time next week, would be nice to have this in obviously :)

For the testing, I think we've solved that on the other PR, correct?

schustmi:
Closing this in favor of #3332

PR #3020, opened 2024-09-17
Linking registered models in MLflow with the corresponding MLflow run

aiakide:
## Describe changes

To link a registered model in MLflow with the corresponding MLflow run, the parameter `mlflow_run_id` must be included in the metadata argument of the `mlflow_register_model_step`. 

The MLflow run can be traced back to a ZenML pipeline run. This is not a necessary change. However, it is more convenient to have a reference to the MLflow run for the registered models as well, as this makes it possible to view parameters or metrics, for example.

I have adjusted the `mlflow_register_model_step` so that the `mlflow_run_id` is added by default if it is not already present in the metadata. 

The following screenshots show the effects in the MLFlow user interface. 
Bildschirmfoto 2024-09-17 um 17 19 31
Bildschirmfoto 2024-09-17 um 17 19 55

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/mlflow/steps/mlflow_registry.py
+++ b/src/zenml/integrations/mlflow/steps/mlflow_registry.py
@@ -146,9 +146,8 @@ def mlflow_register_model_step(
         metadata.zenml_pipeline_run_uuid = pipeline_run_uuid
     if metadata.zenml_workspace is None:
         metadata.zenml_workspace = zenml_workspace
-    if metadata.model_extra:
-        if metadata.model_extra.get("mlflow_run_id", None) is None:
-            metadata.model_extra["mlflow_run_id"] = mlflow_run_id
+    if metadata.model_extra.get("mlflow_run_id", None) is None:
+        metadata.model_extra["mlflow_run_id"] = mlflow_run_id
 
     # Register model version
     model_version = model_registry.register_model_version(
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/mlflow/steps/mlflow_registry.py
```diff
@@ -146,6 +146,11 @@ def mlflow_register_model_step(
         metadata.zenml_pipeline_run_uuid = pipeline_run_uuid
     if metadata.zenml_workspace is None:
         metadata.zenml_workspace = zenml_workspace
+    if metadata.model_extra and "mlflow_run_id" in metadata.model_extra:
```
schustmi, line 149: I'm a little confused by this change, can you explain the intended behaviour of this code?

To me it seems like:
- We first check whether the `mlflow_run_id` exists in the `model_extra` dict
- If it doesn't exist, we do nothing. Why don't we add the key in this case?
- If it does exist, we have another `if` condition, but both the `if` and `else` branch do exactly the same thing: They overwrite the key with the mlflow run id.

It seems like both the `if` and `else` condition do the same thing: Overwrite `mlflow_run_id` in the `model_extra` dict, no matter if it's `None` or not.

And why do we need the actual check ahead if the key even exists and not just always set it?

schustmi (changes requested):

schustmi (approved):

src/zenml/integrations/mlflow/steps/mlflow_registry.py
schustmi: Yep looks good now!

safoinme (approved):

PR #3027, opened 2024-09-19
Leverage warm pools for AWS Sagemaker

avishniakov:
## Describe changes
I implemented support for [Sagemaker Warm Pools](https://docs.aws.amazon.com/sagemaker/latest/dg/train-warm-pools.html) to improve the runtime of the Sagemaker orchestrator.
Observed speed-up can be estimated as up to 45% for the step to be provisioned and started. The exact start-up time heavily depends on the base Docker image being used, instance types, and the composition of the steps.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
+++ b/src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
@@ -102,7 +102,7 @@ class SagemakerOrchestratorSettings(BaseSettings):
     max_runtime_in_seconds: int = 86400
     tags: Dict[str, str] = {}
     keep_alive_period_in_seconds: Optional[int] = 300  # 5 minutes
-    use_training_step: bool = True
+    use_training_step: Optional[bool] = None
 
     processor_args: Dict[str, Any] = {}
     estimator_args: Dict[str, Any] = {}
--- a/src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
+++ b/src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
@@ -30,6 +30,9 @@ from zenml.utils.secret_utils import SecretField
 if TYPE_CHECKING:
     from zenml.integrations.aws.orchestrators import SagemakerOrchestrator
 
+DEFAULT_TRAINING_INSTANCE_TYPE = "ml.m5.xlarge"
+DEFAULT_PROCESSING_INSTANCE_TYPE = "ml.t3.medium"
+
 
 class SagemakerOrchestratorSettings(BaseSettings):
     """Settings for the Sagemaker orchestrator.
@@ -42,6 +45,13 @@ class SagemakerOrchestratorSettings(BaseSettings):
         max_runtime_in_seconds: The maximum runtime in seconds for the
             processing job.
         processor_tags: Tags to apply to the Processor assigned to the step.
+        keep_alive_period_in_seconds: The time in seconds after which the
+            provisioned instance will be terminated if not used. This is only
+            applicable for TrainingStep type and it is not possible to use
+            TrainingStep type if the `output_data_s3_uri` is set to Dict[str, str].
+        use_training_steps_where_possible: Whether to use the TrainingStep
+            type if possible. It is not possible to use TrainingStep type
+            if the `output_data_s3_uri` is set to Dict[str, str].
         processor_args: Arguments that are directly passed to the SageMaker
             Processor for a specific step, allowing for overriding the default
             settings provided when configuring the component. See
@@ -74,11 +84,13 @@ class SagemakerOrchestratorSettings(BaseSettings):
                     Data must be available locally in /opt/ml/processing/output/data/<ChannelName>.
     """
 
-    instance_type: str = "ml.t3.medium"
+    instance_type: Optional[str] = None
     processor_role: Optional[str] = None
     volume_size_in_gb: int = 30
     max_runtime_in_seconds: int = 86400
     processor_tags: Dict[str, str] = {}
+    keep_alive_period_in_seconds: Optional[int] = 300  # 5 minutes
+    use_training_steps_where_possible: bool = True
 
     processor_args: Dict[str, Any] = {}
     input_data_s3_mode: str = "File"
--- a/src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
+++ b/src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
@@ -15,7 +15,7 @@
 
 from typing import TYPE_CHECKING, Any, Dict, Optional, Type, Union
 
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from zenml.config.base_settings import BaseSettings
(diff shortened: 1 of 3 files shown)


src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -316,8 +315,9 @@ def prepare_or_run_pipeline(
                     )
 
             # Construct S3 outputs from container for step
+            only_processing_step_possible = False
             outputs = None
-
+            output_path = None
```
schustmi, line 320: Seems like this is not used?

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -74,11 +84,13 @@ class SagemakerOrchestratorSettings(BaseSettings):
                     Data must be available locally in /opt/ml/processing/output/data/<ChannelName>.
     """
 
-    instance_type: str = "ml.t3.medium"
+    instance_type: Optional[str] = None
     processor_role: Optional[str] = None
     volume_size_in_gb: int = 30
     max_runtime_in_seconds: int = 86400
     processor_tags: Dict[str, str] = {}
```
schustmi, line 91: This and `processor_args` should maybe be renamed/deprecated now that its more generic?

schustmi (changes requested):

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
avishniakov: Used in `sagemaker.estimator.Estimator` constructor.

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -30,6 +30,9 @@
 if TYPE_CHECKING:
     from zenml.integrations.aws.orchestrators import SagemakerOrchestrator
 
+DEFAULT_TRAINING_INSTANCE_TYPE = "ml.m5.xlarge"
```
bcdurak, line 33: Might be nice to mention these default types in the docs as well.

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -42,6 +45,13 @@ class SagemakerOrchestratorSettings(BaseSettings):
         max_runtime_in_seconds: The maximum runtime in seconds for the
             processing job.
         processor_tags: Tags to apply to the Processor assigned to the step.
+        keep_alive_period_in_seconds: The time in seconds after which the
+            provisioned instance will be terminated if not used. This is only
+            applicable for TrainingStep type and it is not possible to use
+            TrainingStep type if the `output_data_s3_uri` is set to Dict[str, str].
+        use_training_steps_where_possible: Whether to use the TrainingStep
```
bcdurak, line 52: As far as I can see, this parameter can be used both on the pipeline-level and on the step-level. While the naming makes sense on the pipeline level, on the step-level it is a bit unintuitive IMO. How about we rename it to `use_training_step`?

I realize that some steps might not be conducted as a training step and they have to be conducted as a `ProcessingStep` because of this condition: `elif isinstance(step_settings.output_data_s3_uri, dict):`. We can check and correct this behaviour through a pydantic validator in the step settings.

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
avishniakov: Reworked as suggested

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
avishniakov: Thanks,  I made it via validator, but want to clarify something. @schustmi also pulling you in.

A user does the following:
1. Set `use_training_step` to True on a step and to False on a pipeline => outcome? I assume 'True'
2. Set `use_training_step` to False on a step and to True on a pipeline => outcome? I assume 'False'
3. Nothing set in step => use pipeline
4. Nothing set in pipeline => use step

Speaking in code
```python
use_training_step_in_step = step_settings.model_dump(exclude_unset=True).get("use_training_step",None)
if use_training_step_in_step is not None:
    use_train_step = use_training_step_in_step
else:
    use_train_step = self.config.use_training_step
```

I implemented it like this for now, LMK is this is lame.

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
avishniakov: done

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
schustmi: The settings get applied like this:
- If a value is set on the pipeline, it gets applied to all steps.
- If a value is set for a specific step, it overwrites the generic value from the pipeline.

So 1-3 of your points are definitely correct, the fourth one I think as well. The step one is always used first, only if not it will fallback to the pipeline one.

One thing to keep in mind here: Your `exclude_unset` filter here will probably not work. This is because we write the full settings (including fields with unset/default values) to our DB. The reason for this is that we can then display it in the dashboard, and it might also prevent issues if default values were to change between different releases.
This however means that `use_training_step` will always be `True` in case not specified by the user, and it will never fall back to the config. To achieve that, you'd probably have to make it an `Optional[bool] = None` field use use the `None` value to check if it was set.

schustmi:
```python
use_training_step = step_settings.use_training_step
if use_training_step is None:
  use_training_step = self.config.use_training_step
if use_training_step is None:
  use_training_step = True # Some default value if nothing is specified anywhere
```

Something like this

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -74,23 +96,70 @@ class SagemakerOrchestratorSettings(BaseSettings):
...
-    output_data_s3_mode: str = "EndOfJob"
+    output_data_s3_mode: str = DEFAULT_OUTPUT_DATA_S3_MODE
     output_data_s3_uri: Optional[Union[str, Dict[str, str]]] = Field(
         default=None, union_mode="left_to_right"
     )
 
+    processor_role: Optional[str] = None
+    processor_tags: Dict[str, str] = {}
+    _deprecation_validator = deprecation_utils.deprecate_pydantic_attributes(
+        "processor_role", "processor_tags"
```
schustmi, line 123: If you make this a tuple `("processor_tags", "tags")`, the value will be automatically migrated to the new attribute. Seems like something that could be useful in this case?

avishniakov: Wow, cool! Reworked

schustmi (approved):

PR #3034, opened 2024-09-23
Remove Python 3.8 support

strickvl:
Python 3.8 is EOL on October 1st, so we're removing support for it.


htahir1:
@bcdurak

schustmi (approved):

avishniakov (approved):
There are some Reds related to DAG on CI, doubt that it is related, but something to take into account @bcdurak , if we take this into the release today

PR #3041, opened 2024-09-24
Use a generic OAuth2 client credentials flow to login to the Cloud API

stefannica:
## Describe changes
Currently, the ZenML Pro tenants are hard-coded to using Auth0 to fetch the M2M tokens needed to access the ZenML Pro API. This PR changes that and makes the implementation generic, allowing ZenML Pro tenants to connect to any authentication server that supports the OAuth2 client credentials grant, not just Auth0.

This allows ZenML Pro tenants to connect straight to the ZenML Pro API itself to fetch M2M tokens through the client credentials grant instead of Auth0.

This is paired with https://github.com/zenml-io/zenml-cloud-api/pull/223 which adds OAuth2 client credentials grant support to the ZenML Pro API. 

Note: no configuration changes are needed from the ZenML Pro deployments themselves, because they will be using the same client ID and client secret to connect to the ZenML PRo API. However, as a future security hardening measure, each tenant will be given its own client ID (probably the same as the tenant ID) and client secret.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/cloud_utils.py
+++ b/src/zenml/zen_server/cloud_utils.py
@@ -1,6 +1,7 @@
 """Utils concerning anything concerning the cloud control plane backend."""
 
 import os
+from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, Optional
 
 import requests
@@ -19,11 +20,8 @@ class ZenMLCloudConfiguration(BaseModel):
     """ZenML Pro RBAC configuration."""
 
     api_url: str
-
     oauth2_client_id: str
     oauth2_client_secret: str
-    oauth2_audience: str
-    auth0_domain: str
 
     @field_validator("api_url")
     @classmethod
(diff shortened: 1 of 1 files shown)


schustmi (approved):

src/zenml/zen_server/cloud_utils.py
```diff
@@ -196,29 +199,49 @@ def _fetch_auth_token(self) -> str:
...
+            self._token is not None
+            and self._token_expires_at is not None
+            and datetime.now(timezone.utc) + timedelta(minutes=5)
+            < self._token_expires_at
+        ):
+            return self._token
+
         # Get an auth token from auth0
-        auth0_url = f"https://{self._config.auth0_domain}/oauth/token"
+        login_url = f"{self._config.api_url}/auth/login"
```
avishniakov, line 211: Maybe we need to keep the full URL configurable, so one can set it to an older value as well?
I mean now we are locking ourselves into in-house implementation, which is not a bad thing per se.  So you can also ignore this.

avishniakov (approved):

src/zenml/zen_server/cloud_utils.py
stefannica: I agree, but I'm trying to keep the current tenant environment variables still valid for now, and the API URL we already have in its base form.

To support fully configurable URLs, I would have to make changes to the tenant deployment configurations while also supporting backwards-compatibility for older tenants, which makes this more difficult.

PR #3053, opened 2024-10-04
`register_artifact ` function

avishniakov:
## Describe changes
I implemented `register_artifact ` function to support use-cases when users want to just link already existing data inside the artifact store scope manually saved by some other tools.

Docs section has quite some examples of the usage in various scenarios.

Example use-case: Pytorch-Lightning has the ability to store checkpoints on a given remote storage, so there is no need to do `save_artifact` in the end epoch callback or similar, but rather just link existing checkpoint to be available as a ZenML artifact for the future use.

## Pre-requisites

## Types of changes


docs/book/user-guide/starter-guide/manage-artifacts.md
```diff
@@ -301,6 +301,40 @@ Even if an artifact is created externally, it can be treated like any other arti
...
+# store is s3://my_bucket, otherwise the linking
+# going to fail with an error.
+
+from zenml.client import Client
+from zenml import link_folder_as_artifact
+from pytorch_lightning import Trainer
+
+# Define the model and fit it
+model = ...
+trainer = Trainer(default_root_dir="s3://my_bucket/my_model_data/")
```
htahir1, line 323: might make sense to somehow use some internal methods to get the bucket from the artifact store, like active_stack.artifact_store.uri because then it can work in the local case as well without hardcoding?

docs/book/user-guide/starter-guide/manage-artifacts.md
```diff
@@ -301,6 +301,40 @@ Even if an artifact is created externally, it can be treated like any other arti
...
+try:
+    trainer.fit(model)
+finally:
+    # We now link those checkpoints in ZenML as an artifact
+    # This will create a new artifact version
+    link_folder_as_artifact(folder_uri="s3://my_bucket/my_model_data/ckpts", name="my_model_ckpts")
```
htahir1, line 329: cant we also have callbacks to do this one by one in case training gets interrupted? might make sense to have `link_file_as_artifact` for that

htahir1: How will versioning work in this case?

htahir1 (changes requested):
Left some comments as most people are out today :-)

docs/book/user-guide/starter-guide/manage-artifacts.md
avishniakov: Reworked

docs/book/user-guide/starter-guide/manage-artifacts.md
avishniakov: In this case, the versioning would not work too much, but I added an extra example to this section.
In fact, `link_file_as_artifact` would mean the need to move data around, since our materialized relies on folders, AFAIK. I would wait for @schustmi to confirm on Monday, but even with the current approach on folders, I'm rather satisfied with how this experience could look like under a Callback of Lightning directly.

avishniakov:
@wjayesh since we had a word about this on Discord, you might want to look into this before we merge as well.

src/zenml/artifacts/load_directory_materializer.py
```diff
@@ -0,0 +1,80 @@
...
+import tempfile
+from pathlib import Path
+from typing import Any, ClassVar, Tuple, Type
+
+from zenml.enums import ArtifactType
+from zenml.io import fileio
+from zenml.materializers.base_materializer import BaseMaterializer
+
+
+class LoaderDirectoryMaterializer(BaseMaterializer):
```
schustmi, line 26: If I remember correctly, I implemented a `DirectoryMaterializer` which works to both save and load directories when an artifact was of type `Path`. Any specific reason we do not implement the `save(...)` method for this materializer as well?

I might be wrong here, but because this materializer has the `pathlib.Path` as it's associated type, it will even be registered for this type in the materializer registry in case it gets imported somehow? Which would then break all returns of type `pathlib.Path`.

src/zenml/artifacts/load_directory_materializer.py
```diff
@@ -0,0 +1,80 @@
...
+    def load(self, data_type: Type[Any]) -> Any:
+        """Copy the artifact files to a local temp directory.
+
+        Args:
+            data_type: Unused.
+
+        Returns:
+            Path to the local directory that contains the artifact files.
+        """
+        directory = tempfile.mkdtemp(prefix="zenml-artifact")
```
schustmi, line 48: I'm guessing we could add a check that this is indeed a directory, not a file? Otherwise the copying will fail/have unexpected behaviour.

schustmi (changes requested):

src/zenml/artifacts/load_directory_materializer.py
avishniakov: The `save` was not implemented to protect from misuse of this specific materializer by intent. And this implementation echos the `DirectoryMaterializer` you created in projects.

Yes, it works with `pathlib.Path`, shall I create a child class from `pathlib.Path` to protect this case?

docs/book/user-guide/starter-guide/manage-artifacts.md
schustmi: I'm guessing the callback will create files with the epoch name in the checkpoints directory? In that case, the repeated calls to this method would create artifact versions, but they would all point to the same directory which will have all the checkpoint files inside?

Currently when ZenML creates an artifact, we usually point the path to a directory. However we might be able to workaround that by simply creating an artifact which points to a specific file, as long as the materializer handles it correctly? Not sure if there are additional checks and whether we get the exact checkpoint file URL inside the handler, but worth a try I guess.

If that's not possible, it's somehow not that useful that actually call that at the end of every epoch, right? Just doing it once before the training call would actually be good enough I assume?

avishniakov: I depicted how this shall be used in the example in docs, but you are all right - pointing to a file is more intuitive for the checkpointing case (and for others maybe). I will try to switch gears here and point to the file instead.

src/zenml/artifacts/load_directory_materializer.py
schustmi: I see, I think my preferred option would be somehow allowing special materializers without associated types that don't get picked up as default for anything in the registry. I'm not sure however how much effort this is.

Something I think we should do in any case, even if we keep it as-is: The `NotImplementedError` that gets raised in the `save(...)` method should explain that this is a special materializer and should not be used for regular artifacts.

docs/book/user-guide/starter-guide/manage-artifacts.md
avishniakov: I made it available for single files as well

src/zenml/artifacts/load_directory_materializer.py
avishniakov: I made it through the `PreexistingArtifactPath` derived from the `Path`, otherwise logic of input artifacts blows up. Please check the recent changes again 🙂

src/zenml/artifacts/utils.py
```diff
@@ -272,29 +245,113 @@ def _create_version() -> Optional[ArtifactVersionResponse]:
...
+        version: Union[int, str],
+    ) -> Optional[ArtifactVersionResponse]:
+        artifact_version = ArtifactVersionRequest(
+            artifact_id=artifact.id,
+            version=version,
+            tags=tags,
+            type=ArtifactType.DATA,
+            uri=folder_or_file_uri,
+            materializer=source_utils.resolve(PreexistingDataMaterializer),
+            data_type=source_utils.resolve(Path),
```
schustmi, line 323: This should be `PreexistingArtifactPath` I think?

src/zenml/artifacts/load_directory_materializer.py
```diff
@@ -0,0 +1,93 @@
...
+import tempfile
+from pathlib import Path
+from typing import Any, ClassVar, Tuple, Type
+
+from zenml.enums import ArtifactType
+from zenml.io import fileio
+from zenml.materializers.base_materializer import BaseMaterializer
+
+
+class PreexistingArtifactPath(type(Path())):  # type: ignore[misc]
```
schustmi, line 26: Does `class PreexistingArtifactPath(Path):` not work for some reason?

src/zenml/artifacts/load_directory_materializer.py
```diff
@@ -0,0 +1,93 @@
...
+    def save(self, data: Any) -> None:
+        """Store the directory in the artifact store.
+
+        Args:
+            data: Path to a local directory to store.
+
+        Raises:
+            NotImplementedError: Always
+        """
+        raise NotImplementedError
```
schustmi, line 74: I still think this should have at least a short error message explaining to not use this materializer to save things, just in case

src/zenml/artifacts/utils.py
```diff
@@ -272,29 +245,113 @@ def _create_version() -> Optional[ArtifactVersionResponse]:
...
+            name=name,
+            response=response,
+            is_model_artifact=is_model_artifact,
+            is_deployment_artifact=is_deployment_artifact,
+        )
+
+    return response
+
+
+def link_existing_data_as_artifact(
```
schustmi, line 258: Maybe we could shorten this to something like `register_artifact(...)`? Or do you think that would not be clear enough in what it does?

schustmi (changes requested):

src/zenml/artifacts/load_directory_materializer.py
avishniakov: Yeah, it fails to init some internal variables, like `_flavor`. I tried using `super().__init__`, but it didn't help.
Some random guys from the Internet suggest to subclass like this for `Path`.

src/zenml/artifacts/utils.py
avishniakov: Yeah, thanks for the catch

src/zenml/artifacts/utils.py
avishniakov: Doch, `register_artifact(...)` reflects it well, I will switch

src/zenml/artifacts/load_directory_materializer.py
avishniakov: Added

src/zenml/artifacts/utils.py
avishniakov: Renamed

schustmi (approved):

avishniakov:
@htahir1 would need you 🦭 to merge this sometime

htahir1 (changes requested):
Some minor do overs but almost there

bcdurak (changes requested):
Left some comments as well.

src/zenml/artifacts/load_directory_materializer.py
```diff
@@ -0,0 +1,98 @@
...
+import tempfile
+from pathlib import Path
+from typing import Any, ClassVar, Tuple, Type
+
+from zenml.enums import ArtifactType
+from zenml.io import fileio
+from zenml.materializers.base_materializer import BaseMaterializer
+
+
+class PreexistingArtifactPath(type(Path())):  # type: ignore[misc]
```
schustmi, line 26: @bcdurak and me just looked at this again and thought that having an extra type just for this was still a little weird. What do you think of using `Path` as the `ASSOCIATED_TYPE`, and setting `SKIP_REGISTRATION=True` for this materializer?

This way, it would not be used when users return an artifact of type `Path`, but it will be used to load artifacts that were created by the `register_artifact` function because the materializer is specified explicitly.

avishniakov: Yeah, this is great idea

src/zenml/artifacts/utils.py
```diff
@@ -577,6 +636,146 @@ def get_artifacts_versions_of_pipeline_run(
...
+        RuntimeError: If the artifact store already contains an artifact with
+            the given URI.
+    """
+    if artifact_store.exists(uri):
+        # This check is only necessary for manual saves as we already check
+        # it when creating the directory for step output artifacts
+        other_artifacts = Client().list_artifact_versions(uri=uri, size=1)
+        if other_artifacts and (other_artifact := other_artifacts[0]):
+            raise RuntimeError(
+                f"Cannot create new artifact {name} version with URI "
```
bcdurak, line 661: I forgot to mention this here. Since you moved the function out, we might want to rephrase the error message here. Previously this was only used in a context where we were trying to create an artifact, now it is a check that can be used in different settings as well.

avishniakov: I would rather disagree here: the context is still the same - we try to create AV in ZenML on a given URI and it fails to do so since URI is already taken. I'd prefer to keep this one as is.

docs/book/user-guide/starter-guide/manage-artifacts.md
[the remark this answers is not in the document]
avishniakov: Done so

docs/book/user-guide/starter-guide/manage-artifacts.md
```diff
@@ -301,6 +301,51 @@ Even if an artifact is created externally, it can be treated like any other arti
 It is also possible to use these functions inside your ZenML steps. However, it is usually cleaner to return the artifacts as outputs of your step to save them, or to use External Artifacts to load them instead.
 {% endhint %}
 
+## Linking data **not** produced by ZenML pipelines as a ZenML artifact
```
htahir1, line 304:
```suggestion
### Linking existing data as a ZenML artifact
```

src/zenml/__init__.py
```diff
@@ -36,6 +36,8 @@
     log_artifact_metadata,
     save_artifact,
     load_artifact,
+    register_artifact,
+
```
htahir1, line 40: extra line?

htahir1 (approved):
Almost there! but consider it approved from my side once you apply final changes :-) Great job!

src/zenml/__init__.py
avishniakov, line 40:
```suggestion
```

bcdurak (approved):

PR #3054, opened 2024-10-04
Fix model linkage for the lazy loading scenarios

avishniakov:
## Describe changes
I fixed how we were storing the lazy loader links to the model, expected to be read from the context (`version=None`) to achieve:
- Single (correct) creation of Model Versions during runs
- Correct linkage of the Model Versions to various response models for the Dashboard and not only for it

## Pre-requisites

## Types of changes


src/zenml/model/lazy_load.py
```diff
@@ -28,7 +29,23 @@ class ModelVersionDataLazyLoader(BaseModel):
     model version during runtime time of the step.
     """
 
-    model: Model
+    model_name: str
```
schustmi, line 32: I might be wrong here, but won't this break all existing runs in the DB, as they store these objects in the DB as part of the configuration?

src/zenml/orchestrators/input_utils.py
```diff
@@ -115,8 +124,8 @@ def resolve_step_inputs(
         if issue_found:
             raise ValueError(
                 "Cannot fetch requested information from model "
-                f"`{config_.model.name}` version "
-                f"`{config_.model.version}` given artifact "
+                f"`{config_.model_name}` version "
+                f"`{config_.model_version}` given artifact "
```
schustmi, line 128: This might be `None` I guess, which we could probably display in a better way in the logs?

src/zenml/orchestrators/input_utils.py
```diff
@@ -85,21 +85,30 @@ def resolve_step_inputs(
     for name, config_ in step.config.model_artifacts_or_metadata.items():
         issue_found = False
         try:
-            if config_.metadata_name is None and config_.artifact_name:
-                if artifact_ := config_.model.get_artifact(
+            context_model = config_._get_model_response(
```
schustmi, line 88: If I get this correctly, we assume here that:
- These model lazy loaders only work with the model of the pipeline context. So something like the following would fail:
```python
@pipeline
def my_pipeline():
  model = Model(name="my_model")
  lazy_artifact = model.get_artifact("my_artifact")
  my_step(lazy_artifact)
```
- The model must already be linked to the pipeline run, even if this is happening in the first step. Are we ensuring that already?

schustmi (changes requested):
I haven't tested it on your branch yet so I might be wrong, but I think there still is an issue linking the model to a pipeline run in the following case:

```python
@pipeline(model=Model(name="my_model", version="existing_version"))
def my_pipeline():
  ...
```

My guess is because the model version already exists, `Model._prepare_model_version_before_step_launch` probably does not link it to the step run/pipeline run.

avishniakov:
> I haven't tested it on your branch yet so I might be wrong, but I think there still is an issue linking the model to a pipeline run in the following case:
> 
> ```python
> @pipeline(model=Model(model_version_id=<EXISTING_ID>))
> def my_pipeline():
>   ...
> ```
> 
> My guess is because the model version already exists, `Model._prepare_model_version_before_step_launch` probably does not link it to the step run/pipeline run.

This is a heavy misuse of the Model class 🙂 I will put a security gate, so `model_version_id` is not set directly on class init. One should `Model(name="foo",version=42)` instead.
Aside, the `name` is mandatory, so your code will look like `model=Model(name="foo",model_version_id="5d58709b-a2c2-4771-911e-80f09868aa30")`, which makes it a bit awkward.

src/zenml/orchestrators/input_utils.py
schustmi: Another thing that I thought of here: What happens in the case where steps define a separate model from the pipeline, in that case we don't link it to a run, right? In that case we would also run into an issue here?

For example:
```python
@step(model=Model("producer_model"))
def producer_step() -> Annotated[int, ArtifactConfig(model_name="pipeline_model")]:
  ...

@step(model=Model("consumer_model"))
def consumer_step(input):
  ...

@pipeline(model=Model("pipeline_model"))
def my_pipeline():
  producer_step()
  lazy_artifact = get_pipeline_context().model.get_artifact("my_artifact")
  consumer_step(lazy_artifact, after="producer_step")
```

src/zenml/model/lazy_load.py
avishniakov: Added migration

src/zenml/orchestrators/input_utils.py
avishniakov: > ```python
> @step(model=Model("producer_model"))
> def producer_step() -> Annotated[int, ArtifactConfig(model_name="pipeline_model")]:
>   ...
> 
> @step(model=Model("consumer_model"))
> def consumer_step(input):
>   ...
> 
> @pipeline(model=Model("pipeline_model"))
> def my_pipeline():
>   producer_step()
>   lazy_artifact = get_pipeline_context().model.get_artifact("my_artifact")
>   consumer_step(lazy_artifact, after="producer_step")
> ```

You syntax is a bit off, but something like this will do:
```python
from typing import Annotated
from zenml import pipeline, step, Model, get_pipeline_context
from zenml.artifacts.artifact_config import ArtifactConfig


@step(model=Model(name="producer_model"))
def producer_step() -> Annotated[int, ArtifactConfig(name="my_artifact",model_name="pipeline_model",model_version="latest")]:
  return 1

@step(model=Model(name="consumer_model"))
def consumer_step(input:int):
  return input*2

@pipeline(model=Model(name="pipeline_model"), enable_cache=False)
def my_pipeline():
  producer_step()
  lazy_artifact = get_pipeline_context().model.get_artifact("my_artifact")
  consumer_step(lazy_artifact, after="producer_step")


if __name__ == "__main__":
    run = my_pipeline()
    print(run.model_version)
    print(run.config.model)
    ```

File: src/zenml/orchestrators/input_utils.py (hunk 2 above)
Comment (avishniakov) [reply src/zenml/orchestrators/input_utils.py]: Significantly reworked this part

**Comment by avishniakov** [comment]
@schustmi I made quite some improvements here and there. I would give it a solid testing on a cloud tenant tomorrow, but locally all looks solid. I also played with your snippets - they helped to catch a few things and fix them 👍🏼

File: src/zenml/model/model.py [hunk 4]
```diff
@@ -878,21 +887,24 @@ def _prepare_model_version_before_step_launch(
...
             # update the configured model version id in runs accordingly
             if step_run:
-                Client().zen_store.update_run_step(
+                client.zen_store.update_run_step(
                     step_run_id=step_run.id,
                     step_run_update=StepRunUpdate(
                         model_version_id=model_version_response.id
                     ),
                 )
+                step_run = client.zen_store.get_run_step(step_run.id)
```
schustmi, line 899: This is just to fetch the new response, correct? I think all our `update_...` methods in the zen store already return the updated model, so we can save this request (and same for the one below)

src/zenml/zen_stores/migrations/versions/1d8f30c54477_migrate_to_new_.py
```diff
@@ -0,0 +1,120 @@
...
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "1d8f30c54477"
+down_revision = "0.67.0"
+branch_labels = None
+depends_on = None
+
+
```
schustmi, line 21: I think theres one legacy column still missing in this migration: `PipelineRunSchema.pipeline_configuration`

schustmi (changes requested):

src/zenml/orchestrators/input_utils.py
schustmi: > > If I get this correctly, we assume here that:
> > 
> > * These model lazy loaders only work with the model of the pipeline context. So something like the following would fail:
> > 
> > ```python
> > @pipeline
> > def my_pipeline():
> >   model = Model(name="my_model")
> >   lazy_artifact = model.get_artifact("my_artifact")
> >   my_step(lazy_artifact)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * The model must already be linked to the pipeline run, even if this is happening in the first step. Are we ensuring that already?
> 
> Yes, it will fail. But it is intended. As a side effect, I made some changes to how we warm-up models, to force pipeline model to be materialized earlier (before even first step `_prepare` is called).

Is this something that we should try to catch and fail, or do we just leave it as is?

avishniakov: User gets this message in such cases: `ValueError: Failed to lazy load model version data in step ``consumer_step``: Lazy loading of the model failed, since the model version is not set in the pipeline context.`.
Shall it be more clear?

avishniakov: Other option could be to check if the version is set in Model init, if it is called from the pipeline context and early raise. Maybe this is better, since this setting is inefficient anyway.

avishniakov: I made changes which will show
```
ValidationError: 1 validation error for Model
  Value error, `version` must be set if you use the `Model` class directly in the pipeline body, otherwise, you can use `get_pipeline_context().model` to lazy load the current Model Version from the pipeline context. [type=value_error, 
input_value={'name': 'my_model'}, input_type=dict]
```

src/zenml/zen_stores/migrations/versions/1d8f30c54477_migrate_to_new_.py
avishniakov: I don't think so - this field has no Step config in it, so there is nothing to update, IMO.

src/zenml/orchestrators/input_utils.py
schustmi: Yes this early fail seems perfect, thanks!

src/zenml/zen_stores/migrations/versions/1d8f30c54477_migrate_to_new_.py
schustmi: Ah you're right, I was confusing it with the `Model` object that is defined in there, but that's not relevant. Sorry!

src/zenml/model/model.py
avishniakov: Thanks, switched

schustmi (approved):

PR #3060, opened 2024-10-07
Remove log versions from versioned buckets in S3

avishniakov:
## Describe changes
I implemented the removal of versions for log files in S3 versioned bucket scenarios to prevent storage requirements from blowing up. It is still not advised to use versioning here, but if this is required for artifacts, then the logging at least would not be an issue.
Users have to bear in mind the performance impact of this since the versions would be looked up and deleted, which takes some time still.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/artifact_stores/base_artifact_store.py
+++ b/src/zenml/artifact_stores/base_artifact_store.py
@@ -475,6 +475,17 @@ class BaseArtifactStore(StackComponent):
 
         default_filesystem_registry.register(filesystem_class)
 
+    def _keep_only_latest_file_version(self, path: PathType) -> None:
+        """Keep only the latest file version in the given path.
+
+        Method is useful for logs stored in versioned file systems
+        like AWS S3.
+
+        Args:
+            path: The path to the file.
+        """
+        return
+
 
 class BaseArtifactStoreFlavor(Flavor):
     """Base class for artifact store flavors."""
--- a/src/zenml/artifact_stores/base_artifact_store.py
+++ b/src/zenml/artifact_stores/base_artifact_store.py
@@ -475,8 +475,8 @@ class BaseArtifactStore(StackComponent):
 
         default_filesystem_registry.register(filesystem_class)
 
-    def _keep_only_latest_file_version(self, path: PathType) -> None:
-        """Keep only the latest file version in the given path.
+    def _remove_previous_file_versions(self, path: PathType) -> None:
+        """Remove all file versions but the latest in the given path.
 
         Method is useful for logs stored in versioned file systems
         like AWS S3.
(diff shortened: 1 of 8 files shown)


src/zenml/artifact_stores/base_artifact_store.py
```diff
@@ -475,6 +475,17 @@ def _register(self) -> None:
 
         default_filesystem_registry.register(filesystem_class)
 
+    def _keep_only_latest_file_version(self, path: PathType) -> None:
```
schustmi, line 478: Maybe calling this `remove_previous_file_versions(...)` would make it a little clearer? I was confused at first and thought this would return a boolean which indicates whether only the latest file version should be kept.

src/zenml/integrations/s3/flavors/s3_artifact_store_flavor.py
```diff
@@ -106,6 +113,40 @@ def _validate_client_kwargs(
...
 
+    @property
+    def bucket(self) -> str:
+        """The bucket name of the artifact store.
+
+        Returns:
+            The bucket name of the artifact store.
+        """
+        if self._bucket is None:
+            self._bucket = os.path.normpath(self.path).split(os.sep)[1]
```
schustmi, line 124: Will this work on windows, where `os.sep` is not `/`?

src/zenml/integrations/s3/flavors/s3_artifact_store_flavor.py
```diff
@@ -106,6 +113,40 @@ def _validate_client_kwargs(
...
+        return self._bucket
+
+    @property
+    def is_versioned(self) -> bool:
```
schustmi, line 128: I think this would be better on the artifact store object itself, not the config: Usually our convention is that the config objects work without importing the client libraries of the integration, so other developers might think that they can use this property but it will crash if the integration is not installed.

schustmi (changes requested):

src/zenml/integrations/s3/flavors/s3_artifact_store_flavor.py
avishniakov: Ouch, indeed

src/zenml/artifact_stores/base_artifact_store.py
avishniakov: renamed

src/zenml/integrations/s3/flavors/s3_artifact_store_flavor.py
avishniakov: Moved it and also moved it to the init method, so the user will see the warning message early and not somewhere in the middle of pipeline logs.

PR #3061, opened 2024-10-07
Fix infinite loop while fetching artifact store in logs storage class

avishniakov:
## Describe changes
I fixed how the active artifact is being fetched in `StepLogsStorage`. Before: it was fetched in each instance of this class, resulting in an infinite loop of logging in debug level, since the GET active stack method was logging a debug message resulting in yet another call of GET active stack and so on.
After: the active artifact store is passed from the step launcher/runner where it is already available.

Byproduct: improved runtime of the logging in general.

This PR also has another minor improvement in the Client lazy evaluation wrapper: before any function was inspected on each call, which is far from efficient since inspect is normally heavy, now the needed inspect info is fetched once on the wrapper application.

## Pre-requisites

## Types of changes


stefannica (approved):
Ah, yes, nothing better than an "infinite loop in debug mode" bug. The sheer amount of logs that you probably had to go through to figure this one out was enough to make me steer clear of it. You’ve got more patience than a cactus waiting for rain in the desert. Respect !

PR #3068, opened 2024-10-09
Client-side computation of cached steps

schustmi:
## Describe changes

This PR adds functionality to compute cached steps client-side. This reduces the time/money spent waiting for remote orchestrators, which can often take a long time to spin up a step container only for it to immediately shut down because the step was cached.

Some things were removed as part of this PR to make the model linking more consistent for cached and non-cached steps:
- The deprecated functionality of using an external artifact to load an artifact from a model version
- The ability to specify a model version for a specific step output in the `ArtifactConfig`

## Pre-requisites

## Types of changes


src/zenml/orchestrators/step_runner.py
```diff
@@ -244,6 +244,11 @@ def run(
                     link_step_artifacts_to_model(
                         artifact_version_ids=output_artifact_ids
                     )
+                    # TODO: Why do we link the pipeline run to the model of the
```
schustmi, line 247: @avishniakov

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -0,0 +1,466 @@
...
+            # - Get rid of having a separate model for an output. Not sure what
+            #   the use case of that actually is and whether it warrants all
+            #   this complexity.
+            utils._link_cached_artifacts_to_model(
+                model_from_context=model,
+                step_run=step_run_request,
+                step_source=step_run.spec.source,
+            )
+            if model:
+                # TODO: When a step is not cached, we parse the output artifact
```
schustmi, line 298: @avishniakov

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -0,0 +1,466 @@
...
+            # state
+            pipeline_run = Client().get_pipeline_run(pipeline_run.id)
+
+            step_model, step_run = prepare_step_run_model(
+                step_run=step_run, pipeline_run=pipeline_run
+            )
+
+            model = step_model or pipeline_model
+
+            # TODO: This function also imports the step code, which means it
```
schustmi, line 281: @avishniakov

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -0,0 +1,466 @@
...
+        cache_enabled = utils.is_setting_enabled(
+            is_enabled_on_step=step.config.enable_cache,
+            is_enabled_on_pipeline=self.deployment.pipeline_configuration.enable_cache,
+        )
+
+        if cache_enabled:
+            if cached_step_run := cache_utils.get_cached_step_run(
+                cache_key=cache_key
+            ):
+                # TODO: If the step is cached, do we also want to include
```
schustmi, line 134: @avishniakov

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -0,0 +1,466 @@
...
+
+        cache_key = cache_utils.generate_cache_key(
+            step=step,
+            input_artifact_ids=input_artifact_ids,
+            artifact_store=self.stack.artifact_store,
+            workspace_id=Client().active_workspace.id,
+        )
+        request.cache_key = cache_key
+
+        # TODO: Doing this here means this will always fail when running a
```
schustmi, line 111: @htahir1

htahir1: @schustmi i think not storing the code + docstring isn't really an option as its a very nice feature people like already... any other options here?

schustmi: I can't think of anything other than what I mentioned as the alternative here, but I don't think that's an option to be honest for the sake of our request sizes. Additionally, it would be very nice to implement this feature in a way that we don't have to import the step client-side to keep our options open in the future (e.g. start a pipeline without having the requirements installed)

schustmi: I can try to hack together something that checks whether the run is started from a template, and in that case look for the docstring/source code in the steps of the run from which the template was created. This will be very hacky and also won't work anymore as soon as we allow specifying different code versions when executing a template.

src/zenml/orchestrators/step_run_utils.py
schustmi: For this I think that using the inputs of the cached step seems better than our current solution.
We already treat the outputs in a way that we reuse the outputs of the cached steps, including dynamic artifacts. If the new step execution would have behaved exactly the same as the cached step (which is sort of the idea of the entire caching mechanism), it would have also dynamically loaded these inputs.

src/zenml/orchestrators/step_runner.py
schustmi: For this my current understanding is:
1. Output artifacts might specify a custom model in their `ArtifactConfig`, and we want to link those potential models to the pipeline run in which they were used
2. External artifacts can also reference a model to load a value: This seems to be a deprecated feature though, which was replaced by using either a model lazy loader or client lazy loader. If this is correct, can I also remove this as part of my deprecation cleanup?
3. Regular input artifacts can not specify a custom model, and the model which they're associated with is already linked to the pipeline run because they were an output of a previous step.
4. This is the missing piece here I think: When someone calls `load_artifact(...)` within a step, I think we do not link that artifact to the active model version. If you can confirm this and agree that we should add it, I'll add the logic to `load_artifact(...)`

src/zenml/orchestrators/step_run_utils.py
schustmi: For this and the other comment regarding output artifact configs, I think I don't really get the value of having a custom model version for a specific step output. Do we have any specific use-case in mind where this would be required?

If no, my gut feeling would be to just remove the option entirely, which would simplify things. Other than that, I guess we have to store the `ArtifactConfig` in the `StepConfiguration` and use it to link the artifacts to their models.

src/zenml/orchestrators/step_runner.py
avishniakov: 1. Yes
2. Can be removed
3. Yes
4. Yes, but it was intentionally made so. The thinking process is like the inputs are not really owned by a model and those links are more like an ownership thing. At least it was. If we agree it is logical to link all those artifacts - we can do it, no objections.

src/zenml/orchestrators/step_run_utils.py
avishniakov: Agree, good to have them here

src/zenml/orchestrators/step_run_utils.py
avishniakov: I would say the latest. Don't think that people are using that. Have no proofs though.

src/zenml/orchestrators/step_run_utils.py
avishniakov: Yes, let's just remove it. To the facts: I can imagine it is just a not implemeted use case, no good logic behind. I checked the code and cannot find better explanation.

src/zenml/orchestrators/step_runner.py
schustmi: Regarding 4.: I guess this is being consistent with how other step inputs work: None of them get explicitly linked to the model version.

Our model versions now track essentially the artifacts produced by the pipeline/step run that they're configured for. I'm still not entirely sure whether also linking all inputs makes sense, but for now I think we can keep it as is.

src/zenml/orchestrators/step_run_utils.py
schustmi: I've removed the option to specify the model version on the artifact config, and also stored the artifact config in the DB so it can be used to link the artifacts correctly as `model_artifact` or `deployment_artifact`

src/zenml/orchestrators/step_run_utils.py
schustmi: Implemented the fallback from the template run if the step importing fails.

docs/book/user-guide/starter-guide/cache-previous-executions.md
```diff
@@ -22,6 +22,9 @@ ZenML understands that nothing has changed between subsequent runs, so it re-use
 
 In ZenML, caching is enabled by default. Since ZenML automatically tracks and versions all inputs, outputs, and parameters of steps and pipelines, steps will not be re-executed within the **same pipeline** on subsequent pipeline runs as long as there is **no change** in the inputs, parameters, or code of a step.
 
+If you run a pipeline without a schedule, ZenML will be able to compute the cached steps on your client machine. This means steps that these steps don't have to be executed by your [orchestrator](../../component-guide/orchestrators/orchestrators.md), which can save time and money when you're executing your pipelines remotely. If you always want your orchestrator
```
AlexejPenner, line 25:
```suggestion
If you run a pipeline without a schedule, ZenML will be able to compute the cached steps on your client machine. This means that these steps don't have to be executed by your [orchestrator](../../component-guide/orchestrators/orchestrators.md), which can save time and money when you're executing your pipelines remotely. If you always want your orchestrator
```

AlexejPenner (approved):
:seal:

src/zenml/artifacts/artifact_config.py
```diff
@@ -69,63 +58,28 @@ def my_step() -> Annotated[
...
-    model_config = ConfigDict(protected_namespaces=())
-
-    @model_validator(mode="after")
-    def artifact_config_validator(self) -> "ArtifactConfig":
-        """Model validator for the artifact config.
+    @model_validator(mode="before")
+    @classmethod
+    @before_validator_handler
+    def _remove_old_attributes(cls, data: Dict[str, Any]) -> Dict[str, Any]:
+        """Remove old attributes that are not used anymore.
```
avishniakov, line 68: Nice way to deprecate things!

src/zenml/artifacts/external_artifact_config.py
```diff
@@ -32,26 +31,23 @@ class ExternalArtifactConfiguration(BaseModel):
...
-        if self.version and self.model:
-            raise ValueError(
-                "Cannot provide both `version` and `model` when "
-                "creating an external artifact."
-            )
-        return self
+        data.pop("name", None)
+        data.pop("version", None)
+        data.pop("model", None)
+        return data
```
avishniakov, line 50: It looks like older pipelines might fail silently downstream because of this.

src/zenml/model/utils.py
```diff
@@ -152,16 +73,49 @@ def log_model_metadata(
     mv.log_metadata(metadata)
 
 
+def link_artifact_version_to_model_version(
+    artifact_version: ArtifactVersionResponse,
+    model_version: ModelVersionResponse,
+    artifact_config: Optional[ArtifactConfig] = None,
+) -> None:
+    """Link an artifact version to a pipeline version.
```
avishniakov, line 81:
```suggestion
    """Link an artifact version to a model version.
```

src/zenml/model/utils.py
```diff
@@ -152,16 +73,49 @@ def log_model_metadata(
...
+            model_version=model_version.id,
+            is_model_artifact=is_model_artifact,
+            is_deployment_artifact=is_deployment_artifact,
+        )
+    )
+
+
 def link_artifact_to_model(
-    artifact_version_id: UUID,
+    artifact_version: "ArtifactVersionResponse",
```
avishniakov, line 110:
```suggestion
    artifact_version: ArtifactVersionResponse,
```

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -0,0 +1,566 @@
...
+            invocation_id: The invocation ID for which to create the request.
+
+        Returns:
+            The step run request.
+        """
+        return StepRunRequest(
+            name=invocation_id,
+            pipeline_run_id=self.pipeline_run.id,
+            deployment=self.deployment.id,
+            status=ExecutionStatus.RUNNING,
```
avishniakov, line 85: Why not `INITIALIZING`?

src/zenml/orchestrators/step_runner.py
```diff
@@ -522,7 +519,7 @@ def _store_output_artifacts(
             The IDs of the published output artifacts.
         """
         step_context = get_step_context()
-        output_artifacts: Dict[str, UUID] = {}
+        output_artifacts: Dict[str, ArtifactVersionResponse] = {}
```
avishniakov, line 522: It looks like `ArtifactVersionResponse` is behind `TYPE_CHECKING` thus this shall fail.

src/zenml/steps/step_context.py
```diff
@@ -169,36 +172,24 @@ def model(self) -> "Model":
         """Returns configured Model.
 
         Order of resolution to search for Model is:
-            1. Model from @step
-            2. Model from @pipeline
+            1. Model from the step
+            2. Model from the pipeline
```
avishniakov, line 176:
```suggestion
            1. Model from the step context
            2. Model from the pipeline context
```

avishniakov (commented):
Left a few comments, but nothing critical - well done!

src/zenml/orchestrators/step_run_utils.py
schustmi: Somehow we never used that state for steps and only for pipelines. I guess there are only a few cases:
- If something fails, we change the state to failed.
- If everything goes right, the step status changes to running in less than 1 second after this request is made. So we would only need to update the status immediately anyway, which I think we can avoid by just setting it to running from the start?

src/zenml/orchestrators/step_runner.py
schustmi: It somehow does not fail, but I don't get why 🤔
I moved it out of type checking in any case

src/zenml/artifacts/external_artifact_config.py
schustmi: In which case would they fail?

I'm guessing this would only be possible if upgrading the server to the new release while still having an old scheduled pipeline running that uses a Docker image of an older version? The server would then remove those attributes before sending the response.

Would that case be fixed if I just don't remove this data here and add `extra=allow` in the pydantic config?

avishniakov: To me, it would be better just to give some instructions on how to fix this and write that deprecation clearly in the new release notes. I don't ask for any technical solution here - these guys should have been gone for a while already.

src/zenml/orchestrators/step_run_utils.py
avishniakov: Ok, let's keep it.

avishniakov (approved):
🦭

src/zenml/artifacts/external_artifact_config.py
schustmi: The problem here is that these attributes will already be removed on the server side, which means the warning/instructions will be very hidden. I instead added some instructions on the runtime error below that happens if the ID is missing.

PR #3072, opened 2024-10-10
Multiple output versions for a step outputs

avishniakov:
## Describe changes
Highlights:
- `StepRunResponse.outputs` is now `Dict[str, List["ArtifactVersionResponse"]]` to support multiple versions of the same artifact
- `StepNodeDetails.outputs` is now `Dict[str, List[str]]` for the same reason
- `type` is removed from `StepOutputSchema` and now resides in the `ArtifactVersionSchema` directly
- Old types are `DEFAULT` and `MANUAL` appended with new types of artifacts: `EXTERNAL` and `PREEXISTING` for `ExternalArtifact` and `register_artifact` respectively.
- `ArtifactVersionResponse/Request` now expects `save_type`

It is very shaky for the frontend, so I would keep it for a while until we align with @Cahllagerfeld 

P.S. docs to follow after some reviews feedback being collected.

## Pre-requisites

## Types of changes


schustmi:
I have more of a general question:
Your current implementation solves how an artifact was saved, but how do we expose the way it was loaded? E.g. even if an artifact is a regular step output, I can load it either the "normal" way by defining it as a step input, or manually by calling `load_artifact(...)`.

avishniakov:
> I have more of a general question: Your current implementation solves how an artifact was saved, but how do we expose the way it was loaded? E.g. even if an artifact is a regular step output, I can load it either the "normal" way by defining it as a step input, or manually by calling `load_artifact(...)`.

True, I didn't touch inputs here at all, since it was not in the scope of the ticket. I would prefer to handle it separately.
Do you refer to the DAG and how we define which input type is it? It would be hell complex design/FE wise, honestly speaking.

I can do this crazy thing and I have no clue how it would be shown:
```python
@step
def step_1()->Annotated[int,"my_int"]:
    return 42

@pipeline
def pipe_1():
    step_1()
    step_2(Client().get_artifact_version("my_int"), after=["step_2"])
```
So it is a Schrödinger-Artifact now: it is DEFAULT and LAZY_LOADED at the same time...

schustmi:
I see, makes sense let's handle that separately.
I think the problem is mostly on how to visualize what type of input it is right? Could just be a different colored line depending on the type type of input, but I'm sure Zuri will come up with something much better 😄
In general I think it's both, it's a regular output for the first step, but for the second step it's a coincidence that it was generated in the same run, and the general intention of the user was to load an artifact using the lazy loading, not using the regular way to pass artifacts, otherwise they would have done that ;)

schustmi:
That's also why the input type can't be stored in the artifact table, one artifact might be different types of inputs for different steps/runs or even the same exact step

avishniakov:
> That's also why the input type can't be stored in the artifact table, one artifact might be different types of inputs for different steps/runs or even the same exact step

Yep, I also realized that during development - this will stay separately as is it is now, but we would need to enrich it for more types beyond default and manual, IMO.

src/zenml/artifacts/utils.py
```diff
@@ -100,6 +101,7 @@ def save_artifact(
     is_model_artifact: bool = False,
     is_deployment_artifact: bool = False,
     manual_save: bool = True,
+    is_external_artifact: bool = False,
```
schustmi, line 104: Instead of passing `manual_save` and `is_external_artifact` here, can't we just simply pass the type already and default it to `manual`?

src/zenml/enums.py
```diff
@@ -38,11 +38,15 @@ class StepRunInputArtifactType(StrEnum):
     MANUAL = "manual"  # manually loaded via `zenml.load_artifact()`
 
 
-class StepRunOutputArtifactType(StrEnum):
-    """All possible types of a step run output artifact."""
+class ArtifactSaveType(StrEnum):
+    """All possible method types of how artifact versions can be saved."""
 
     DEFAULT = "default"  # output of the current step
```
schustmi, line 44: Not part of your changes, but what do you think of renaming this to something like `STEP_OUTPUT`? If I was a user and read `save_type=default` on an `ArtifactVersionResponse`, I would have no clue what this is referring to I think

src/zenml/models/v2/core/step_run.py
```diff
@@ -123,6 +123,7 @@ class StepRunRequest(WorkspaceScopedRequest):
 class StepRunUpdate(BaseModel):
     """Update model for step runs."""
 
+    # TODO: extend this to List[UUID]
```
schustmi, line 126: TODO

src/zenml/orchestrators/input_utils.py
```diff
@@ -64,7 +64,12 @@ def resolve_step_inputs(
             )
 
         try:
-            artifact = step_run.outputs[input_.output_name]
+            if len(step_run.outputs[input_.output_name]) > 1:
```
schustmi, line 67: This is the exact case the we talked about this morning in the design meeting I think.

I think we should be able to figure out which of the versions is the output that was returned by the step, and use that here instead of failing.

This is only possible if the following code does not work:
```python
@step
def my_step() -> Tuple[Annotated[int, "shared_name"], Annotated[int, "shared_name"]]:
  ...
```

src/zenml/orchestrators/utils.py
```diff
@@ -233,11 +233,12 @@ def _link_cached_artifacts_to_model(
...
             artifact_config_ = ArtifactConfig(name=output_name_)
 
-        link_artifact_config_to_model(
-            artifact_config=artifact_config_,
-            model=model_from_context,
-            artifact_version_id=output_id,
-        )
+        for output_id in output_ids:
+            link_artifact_config_to_model(
+                artifact_config=artifact_config_,
```
schustmi, line 238: The artifact config that is passed here is only valid for artifacts that were returned as the step output, not for dynamic ones saved using `save_artifact`. Are those already included in `step_run.outputs` at this point? And if yes, I would assume they're already linked right?

src/zenml/zen_stores/migrations/versions/1cb6477f72d6_move_artifact_save_type.py
```diff
@@ -0,0 +1,83 @@
...
+        batch_op.add_column(sa.Column("save_type", sa.TEXT(), nullable=True))
+
+    # Step 2: Move data from step_run_output_artifact.type to artifact_version.save_type
+    op.execute("""
+        UPDATE artifact_version
+        SET save_type = step_run_output_artifact.type
+        FROM step_run_output_artifact
+        WHERE artifact_version.id = step_run_output_artifact.artifact_id
+    """)
+    op.execute("""
```
schustmi, line 32: Aren't the ones that have no entry in the `step_run_output_artifact` table external artifacts? At long as no step runs were deleted at least?

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -223,12 +224,15 @@ def to_model(
             for artifact in self.input_artifacts
         }
 
-        output_artifacts = {
-            artifact.name: artifact.artifact_version.to_model(
-                pipeline_run_id_in_context=self.pipeline_run_id
+        output_artifacts: Dict[str, Set["ArtifactVersionResponse"]] = {}
```
schustmi, line 227: Why is this a set here instead of a list?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8063,26 +8062,14 @@ def update_run_step(
             existing_step_run.update(step_run_update)
             session.add(existing_step_run)
 
-            # Update the output artifacts.
-            for name, artifact_version_id in step_run_update.outputs.items():
+            # Update the artifacts.
+            for name, artifact_version_id in list(
+                step_run_update.outputs.items()
+            ) + list(step_run_update.saved_artifact_versions.items()):
```
schustmi, line 8068: Do we even need the `saved_artifact_versions` now? I guess this was just used to differentiate the output type for now, but this is not necessary anymore?

schustmi (changes requested):

src/zenml/zen_stores/migrations/versions/1cb6477f72d6_move_artifact_save_type.py
avishniakov: Hm, yeah, sounds reasonable.

src/zenml/orchestrators/input_utils.py
avishniakov: Adjusted, so it works for cases like
```python
from zenml import step, pipeline, save_artifact

from zenml.models.v2.core.pipeline_run import PipelineRunResponse


@step(enable_cache=False)
def producer_step() -> (
    Annotated[int, "shared_name"]
):
    save_artifact(41, "shared_name")
    return 42

@step(enable_cache=False)
def consumer_step(shared_name: int, expected: int):
    assert shared_name == expected

@pipeline
def my_pipeline():
    a = producer_step()
    consumer_step(a,42)
    consumer_step(Client().get_artifact_version("shared_name",1),41, after=["producer_step"])


if __name__ == "__main__":
    try:
        Client().delete_artifact("shared_name")
    except KeyError:
        pass
    prr: PipelineRunResponse = my_pipeline()
    assert set([a.load() for a in prr.steps["producer_step"].outputs["shared_name"]]) == {41, 42}
```

src/zenml/orchestrators/input_utils.py
```diff
@@ -64,10 +65,13 @@ def resolve_step_inputs(
             )
 
         try:
-            artifact = step_run.outputs[input_.output_name]
+            os_ = step_run.outputs[input_.output_name]
+            artifact = [
+                o for o in os_ if o.save_type == ArtifactSaveType.STEP_OUTPUT
+            ].pop()
```
schustmi, line 71: This will not raise a `KeyError` but an index error. I think we should also check that `len(list) < 2` just to be sure, in case we ever allow returning two regular artifacts with the same name, to avoid any ambiguity here.

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -554,13 +554,14 @@ def link_output_artifacts_to_model_version(
...
-
-        link_artifact_version_to_model_version(
-            artifact_version=output_artifact,
-            model_version=model_version,
-            artifact_config=artifact_config,
-        )
+    for output_name, output_artifacts in artifacts.items():
+        for output_artifact in output_artifacts:
+            artifact_config = None
+            if output_config := output_configurations.get(output_name, None):
```
schustmi, line 560: We need to make sure that this function is only being called with regular step output artifacts, as otherwise the artifact config does not apply

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -223,12 +224,15 @@ def to_model(
...
-        output_artifacts = {
-            artifact.name: artifact.artifact_version.to_model(
-                pipeline_run_id_in_context=self.pipeline_run_id
+        output_artifacts: Dict[str, List["ArtifactVersionResponse"]] = {}
+        for artifact in self.output_artifacts:
+            if artifact.name not in output_artifacts:
+                output_artifacts[artifact.name] = []
+            output_artifacts[artifact.name].append(
+                artifact.artifact_version.to_model(
+                    pipeline_run_id_in_context=self.pipeline_run_id
```
schustmi, line 233: This param can be removed, does not seem to be used anymore.

schustmi (changes requested):

src/zenml/enums.py
```diff
@@ -38,11 +38,15 @@ class StepRunInputArtifactType(StrEnum):
...
-class StepRunOutputArtifactType(StrEnum):
-    """All possible types of a step run output artifact."""
+class ArtifactSaveType(StrEnum):
+    """All possible method types of how artifact versions can be saved."""
 
-    DEFAULT = "default"  # output of the current step
+    STEP_OUTPUT = "default"  # output of the current step
     MANUAL = "manual"  # manually saved via `zenml.save_artifact()`
+    PREEXISTING = "preexisting"  # register via `zenml.register_artifact()`
+    EXTERNAL = (
```
schustmi, line 47: I'm now wondering about all these types: Are they really different from one another? What differentiates an `ExternalArtifact` from one saved with `save_artifact` for example? My current gut feeling is to even get rid of `ExternalArtifact` entirely and replace it by telling people to call `save_artifact` instead.

Also, why do we need to differentiate between `PREEXISTING` and `MANUAL` here, isn't this essentially the same thing as well?

src/zenml/orchestrators/step_run_utils.py
schustmi: This is currently not the case

schustmi: I fixed this, but one thing we don't (and never did): If a manual save was a model artifact for the original step run, it will not be a model artifact in the cached step run.

src/zenml/models/v2/core/step_run.py
```diff
@@ -331,7 +330,7 @@ def inputs(self) -> Dict[str, "ArtifactVersionResponse"]:
         return self.get_body().inputs
 
     @property
-    def outputs(self) -> Dict[str, "ArtifactVersionResponse"]:
+    def outputs(self) -> Dict[str, List["ArtifactVersionResponse"]]:
```
schustmi, line 333: I'm also wondering whether we might want to separate this into `regular outputs` from the step, and manual ones (using `save_artifact` and `register_artifact`)

src/zenml/enums.py
avishniakov: Yes, we can consider deprecating the `ExternalArtifact`, but this is not related to this ticket, I assume.

Regarding `PREEXISTING` and `MANUAL`, I would still keep it separate, since the `PREEXISTING` hacks how we store the data (paths ZenML creates are not respected anymore), which might be dangerous to manipulate during some housekeeping runs. Not the case as of now, but it will be safer to split those for future maintainability, IMO.

src/zenml/orchestrators/step_run_utils.py
avishniakov: I pushed a test for this case, which is skipped now. We will handle this separately, as discussed.

avishniakov:
There is still some unrelated error on mac 3.9, but I will merge this in as is.
```
Using Python 3.9.20 environment at /Users/runner/hostedtoolcache/Python/3.9.20/x64
  × No solution found when resolving dependencies:
  ╰─▶ Because torch==2.4.0 has no wheels with a matching Python implementation
      tag and vllm>=0.6.0 depends on torch==2.4.0, we can conclude that
      vllm>=0.6.0 cannot be used.
      And because only the following versions of vllm are available:
          vllm<=0.6.0
          vllm==0.6.1
          vllm==0.6.1.post1
          vllm==0.6.1.post2
          vllm==0.6.2
          vllm==0.6.3
          vllm==0.6.3.post1
      and you require vllm>=0.6.0, we can conclude that your requirements
      are unsatisfiable.
```

PR #3086, opened 2024-10-15
Fix S3 ArtifactStore auth issue

avishniakov:
## Describe changes
Fixes the authentication issued introduced with https://github.com/zenml-io/zenml/pull/3060

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/s3/artifact_stores/s3_artifact_store.py
+++ b/src/zenml/integrations/s3/artifact_stores/s3_artifact_store.py
@@ -131,7 +131,7 @@ class S3ArtifactStore(BaseArtifactStore, AuthenticationMixin):
             **kwargs: Additional keyword arguments.
         """
         super().__init__(*args, **kwargs)
-        self.__boto3_bucket = None
+        self._boto3_bucket_holder = None
 
         # determine bucket versioning status
         self.is_versioned = False
@@ -489,8 +489,8 @@ class S3ArtifactStore(BaseArtifactStore, AuthenticationMixin):
         Returns:
             The boto3 bucket object.
         """
-        if self.__boto3_bucket and not self.connector_has_expired():
-            return self.__boto3_bucket
+        if self._boto3_bucket_holder and not self.connector_has_expired():
+            return self._boto3_bucket_holder
 
         key, secret, token, region = self.get_credentials()
         s3 = boto3.resource(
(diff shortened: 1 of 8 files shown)


bcdurak (approved):

stefannica (approved):

src/zenml/integrations/s3/artifact_stores/s3_artifact_store.py
```diff
@@ -467,10 +465,27 @@ def _remove_previous_file_versions(self, path: PathType) -> None:
...
+        """
+        key, secret, token, region = self.get_credentials()
+        s3 = boto3.resource(
+            "s3",
+            aws_access_key_id=key,
+            aws_secret_access_key=secret,
+            aws_session_token=token,
+            region_name=region,
+        )
+        return s3.Bucket(self.config.bucket)
```
stefannica, line 491: I had a hunch and asked ChatGPT about it. Here's what it said, but I couldn't verify it yet:

In your current implementation, every time _boto3_bucket() is called, it will create a new S3 connection via boto3.resource. This can lead to inefficiencies, especially in scenarios where the function is called frequently. To avoid creating a new connection on every call, you can optimize it by caching or reusing the S3 resource object.

One of the simplest ways to avoid creating new connections repeatedly is to store the boto3.resource object as an instance variable (in case you're working within a class) or as a global variable. This way, you only create the S3 resource once and reuse it across multiple calls.

```
    def __init__(self):
        self._s3_resource = None  # Cache the S3 resource object

    def _boto3_bucket(self) -> Any:
        """Get the boto3 bucket object.
        Returns:
            The boto3 bucket object.
        """
        if self._s3_resource is None:  # Create the resource only once
            key, secret, token, region = self.get_credentials()
            self._s3_resource = boto3.resource(
                "s3",
                aws_access_key_id=key,
                aws_secret_access_key=secret,
                aws_session_token=token,
                region_name=region,
            )
        return self._s3_resource.Bucket(self.config.bucket)
```


However, it gets a bit more complicated now, because the service connector credentials themselves can expire, so you'd need to refresh the resource object if that happens:

```
    def _boto3_bucket(self) -> Any:
        """Get the boto3 bucket object.
        Returns:
            The boto3 bucket object.
        """
        # Refresh the bucket resource also if the connector has expired
        if self._s3_resource and not self.connector_has_expired():
            return self._s3_resource.Bucket(self.config.bucket)

        key, secret, token, region = self.get_credentials()
        self._s3_resource = boto3.resource(
            "s3",
            aws_access_key_id=key,
            aws_secret_access_key=secret,
            aws_session_token=token,
            region_name=region,
        )
        return self._s3_resource.Bucket(self.config.bucket)
```

stefannica (changes requested):

avishniakov:
@stefannica I made the requested changes and tested this in different conditions: with new perms present or without for the legacy connector - all seems good. Shall we merge it? CI is a bit shaky, but I can merge it like this - there are only docker failures not related to the PR.

stefannica (approved):

PR #3093, opened 2024-10-17
Scalability improvements

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -3918,7 +3918,7 @@ class Client(metaclass=ClientMetaClass):
         workspace_id: Optional[Union[str, UUID]] = None,
         user_id: Optional[Union[str, UUID]] = None,
         model_version_id: Optional[Union[str, UUID]] = None,
-        num_outputs: Optional[Union[int, str]] = None,
+        model: Optional[Union[UUID, str]] = None,
         hydrate: bool = False,
     ) -> Page[StepRunResponse]:
         """List all pipelines.
@@ -3939,10 +3939,10 @@ class Client(metaclass=ClientMetaClass):
             deployment_id: The id of the deployment to filter by.
             original_step_run_id: The id of the original step run to filter by.
             model_version_id: The ID of the model version to filter by.
+            model: Filter by model name/ID.
             name: The name of the step run to filter by.
             cache_key: The cache_key of the run to filter by.
             status: The name of the run to filter by.
-            num_outputs: The number of outputs for the step run
             hydrate: Flag deciding whether to hydrate the output model(s)
                 by including metadata fields in the response.
 
@@ -3968,7 +3968,7 @@ class Client(metaclass=ClientMetaClass):
             workspace_id=workspace_id,
             user_id=user_id,
             model_version_id=model_version_id,
-            num_outputs=num_outputs,
+            model=model,
         )
         step_run_filter_model.set_scope_workspace(self.active_workspace.id)
         return self.zen_store.list_run_steps(
--- a/src/zenml/models/v2/core/step_run.py
+++ b/src/zenml/models/v2/core/step_run.py
@@ -14,7 +14,7 @@
 """Models representing steps runs."""
 
 from datetime import datetime
-from typing import TYPE_CHECKING, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, ClassVar, Dict, List, Optional, Union
 from uuid import UUID
 
 from pydantic import BaseModel, ConfigDict, Field
@@ -33,6 +33,8 @@ from zenml.models.v2.base.scoped import (
 from zenml.models.v2.core.model_version import ModelVersionResponse
 
 if TYPE_CHECKING:
+    from sqlalchemy.sql.elements import ColumnElement
+
     from zenml.models.v2.core.artifact_version import ArtifactVersionResponse
     from zenml.models.v2.core.logs import (
         LogsRequest,
@@ -490,6 +492,11 @@ class StepRunResponse(
 class StepRunFilter(WorkspaceScopedFilter):
     """Model to enable advanced filtering of step runs."""
 
+    FILTER_EXCLUDE_FIELDS: ClassVar[List[str]] = [
+        *WorkspaceScopedFilter.FILTER_EXCLUDE_FIELDS,
+        "model",
+    ]
+
     name: Optional[str] = Field(
         default=None,
         description="Name of the step run",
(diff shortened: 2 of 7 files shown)


src/zenml/models/v2/core/step_run.py
```diff
@@ -543,8 +550,42 @@ class StepRunFilter(WorkspaceScopedFilter):
     )
     model_version_id: Optional[Union[UUID, str]] = Field(
         default=None,
-        description="Model version associated with the pipeline run.",
+        description="Model version associated with the step run.",
         union_mode="left_to_right",
     )
+    model: Optional[Union[UUID, str]] = Field(
```
bcdurak, line 556: I would propose renaming this attribute to something like `model_name_or_id`. I think it much more intuitive that way and also, `model` is a protected namespace by Pydantic. We are already bypassing that for some `model_XXX` attributes by using a custom `ConfigDict` but this one feels like the craziest 😄

src/zenml/cli/stack_components.py
```diff
@@ -139,7 +139,7 @@ def describe_stack_component_command(name_id_or_prefix: str) -> None:
             if component_.connector:
                 # We also need the flavor to get the connector requirements
                 connector_requirements = client.get_flavor_by_name_and_type(
-                    name=component_.flavor, component_type=component_type
```
bcdurak, line 142: At this point in the code, we have a `component` already, and it has the `flavor` in its `resources`. You can use it to fetch the connector requirements without making an extra hydrated `list_flavors` call.

bcdurak: Taking it even one step further if you move `connector_type`, `connector_resource_type`, `connector_resource_id_attr` from `FlavorResponseMetadata` to `FlavorResponseBody`, you can do this without making any extra calls.

src/zenml/cli/stack_components.py
```diff
@@ -1774,11 +1774,11 @@ def connect_stack_component_with_service_connector(
 
     try:
         flavor_model = client.get_flavor_by_name_and_type(
-            name=component_model.flavor, component_type=component_type
```
bcdurak, line 1777: Exactly the same as above.

src/zenml/client.py
```diff
@@ -2083,7 +2094,7 @@ def update_stack_component(
 
             validated_config = validate_stack_component_config(
```
bcdurak, line 2095: I realize this function is used in different contexts, but for instance, here, we already have a component model that has a flavor model in its resource but we pass the flavor name and end up querying for the flavor again.

src/zenml/models/v2/core/component.py
```diff
@@ -359,6 +362,15 @@ def connector(self) -> Optional["ServiceConnectorResponse"]:
         """
         return self.get_metadata().connector
 
+    @property
+    def flavor(self) -> Optional["FlavorResponse"]:
```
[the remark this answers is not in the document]
bcdurak: Added a few comments regarding this.

bcdurak (changes requested):
A huge fan of the `Clean Up PRs` series by @schustmi ❤️

src/zenml/models/v2/core/component.py
```diff
@@ -187,7 +185,7 @@ class ComponentResponseBody(WorkspaceScopedResponseBody):
     type: StackComponentType = Field(
         title="The type of the stack component.",
     )
-    flavor: str = Field(
```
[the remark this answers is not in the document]
schustmi: Yeah I had to keep it in for exactly this reason. For our system to work without this, we need some way to specify additional hydrations like `hydrate=["components.flavor"]` or similar, but that's going to be a bigger change I think

src/zenml/models/v2/core/step_run.py
schustmi: I think the UX with this renaming would be considerably worse: Instead of doing `zenml pipeline runs list --model=<MODEL_NAME> --user=<USER_ID>`, we would have `zenml pipeline runs list --model_name_or_id=<MODEL_NAME> --user_name_or_email_or_id=<USER_NAME>`, which is why we opted for the short names here. In my opinion, ideally these CLI options would be decoupled from the filter models, but right now that is not the case.
We have these shortcut names (including `model` on other filters already, so for consistency I'd keep them like this, is that okay?


(Regarding pydantic, I think `BaseModel.model` is actually more unlikely to be used by them compared to some other fields we have :P)

src/zenml/models/v2/core/component.py
```diff
@@ -232,6 +230,11 @@ class ComponentResponseMetadata(WorkspaceScopedResponseMetadata):
 class ComponentResponseResources(WorkspaceScopedResponseResources):
```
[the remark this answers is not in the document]
schustmi: Wouldn't it be possible to just replace `flavor` with `flavor_name` in the `ANALYTICS_FIELDS` to fix this?

src/zenml/cli/stack_components.py
schustmi: I used `component_.flavor` to avoid the flavor call 👍

For the other change, I'd say this is such a niche command/use case that we don't need to optimize for, and these service connector attributes are actually quite huge in some cases, so having them in every component response does not make much sense IMO. WDYT?

src/zenml/cli/stack_components.py
schustmi: Fixed 👍

src/zenml/cli/stack_components.py
bcdurak: Yeah, maybe you are right. I remember that I was getting way too hyped about reducing the number of calls, got caught up a bit 😄 I think as long as we use the flavor properly, it's all good. 👍

src/zenml/client.py
schustmi: Updated the function to allow a flavor response in cases where we already have it

src/zenml/models/v2/core/step_run.py
bcdurak: > (Regarding pydantic, I think BaseModel.model is actually more unlikely to be used by them compared to some other fields we have :P)

I am very curious about how this will turn out in the future. 😄

src/zenml/client.py
bcdurak: Nice 👍

src/zenml/models/v2/core/component.py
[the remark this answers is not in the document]
schustmi: Ah I see, I didn't think of that. I'll do it with a custom method 👍

src/zenml/analytics/utils.py
```diff
@@ -101,11 +102,13 @@ class analytics_disabler:
 
     def __init__(self) -> None:
         """Initialization of the context manager."""
-        self.original_value = os.environ.get(ENV_ZENML_ANALYTICS_OPT_IN)
+        self.original_value: Optional[bool] = None
 
     def __enter__(self) -> None:
         """Disable the analytics."""
-        os.environ[ENV_ZENML_ANALYTICS_OPT_IN] = str(False)
+        global _INTERNAL_DISABLE_ANALYTICS
```
schustmi, line 109: @bcdurak I had to make some changes to this analytics disable context manager, because it lead to issues with the default stack registration. The following was happening:
- The `SQLZenStore` was initialized, was caused the flavors to be registered.
- The flavor registration caused an analytics call
- The analytics call initialized the zen store of the global config
- The zen store initialization tried to register the default stack
- `orchestrator_schema.to_model()` failed because the flavor was not registered yet

I broke this loop by disabling the analytics before it even gets to the global config. Does this make sense or do I mess stuff up in other places by doing this?

bcdurak: This is a crazy connection but I can not think of any problems that this change would cause 👍

bcdurak (approved):

PR #3096, opened 2024-10-17
Simplify Metadata handling

AlexejPenner:
## Describe changes

The fact that run metadata is stored in a separate table with a row per key-value pair is business logic. I don't see the benefit of exposing this to the user/client. As a user when I run model_version.run_metadata i expect to get a simple
metadata dictionary, instead I get a fulminant RunMetadataResponseBody.

On top of that, the run_metadata lead to bloating of the api response sizes.

Example response body:
[response_1729171061701.json](https://github.com/user-attachments/files/17429160/response_1729171061701.json)

Time to list 100 hydrated model versions with the state on develop:

image

Time to list 100 hydrated model versions after this change:

image

## Breaking Change

All code that accesses run_metadata now needs to be changed:

```
metadata_value = current_model_version.run_metadata["key"].value
```

--> 

```
metadata_value = current_model_version.run_metadata["key"]
```

Metadata is now a familiar python dictionary.

## Pre-requisites

## Types of changes


src/zenml/cli/model.py
```diff
@@ -61,9 +61,7 @@ def _model_version_to_print(
 ) -> Dict[str, Any]:
     run_metadata = None
     if model_version.run_metadata:
-        run_metadata = {
-            k: v.value for k, v in model_version.run_metadata.items()
-        }
+        run_metadata = {k: v for k, v in model_version.run_metadata.items()}
```
schustmi, line 64:
```python
run_metadata = model_version.run_metadata
```

I don't think this needs the if anymore, it's just the same dict

src/zenml/models/v2/core/model_version.py
```diff
@@ -304,7 +302,7 @@ def description(self) -> Optional[str]:
         return self.get_metadata().description
 
     @property
-    def run_metadata(self) -> Optional[Dict[str, "RunMetadataResponse"]]:
+    def run_metadata(self) -> Optional[Dict[str, MetadataType]]:
```
schustmi, line 305: This is not actually optional, it has a default value of `{}` in the metadata

src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -351,7 +352,9 @@ def to_model(
                 artifact_store_id=self.artifact_store_id,
                 producer_step_run_id=producer_step_run_id,
                 visualizations=[v.to_model() for v in self.visualizations],
-                run_metadata={m.key: m.to_model() for m in self.run_metadata},
+                run_metadata={
+                    m.key: json.loads(m.value) for m in self.run_metadata
```
schustmi, line 356: Do we maybe need the pydantic json encoder here so this works with `UUID` and `datetime`? Just in case this ends up in some of the metadata dictionaries?

schustmi: Same applies to all the other schemas as well

src/zenml/zen_stores/schemas/run_metadata_schemas.py
```diff
@@ -111,42 +104,42 @@ class RunMetadataSchema(BaseSchema, table=True):
...
-                resource_type=MetadataResourceTypes(self.resource_type),
-                stack_component_id=self.stack_component_id,
-            )
-
-        return RunMetadataResponse(
-            id=self.id,
-            body=body,
-            metadata=metadata,
-        )
+    # def to_model(
```
schustmi, line 107: Remove

schustmi (changes requested):
A more general comment: Do you think it makes sense to remove these `run_metadata` attributes on all the responses to just `metadata`? Especially `run.run_metadata` seems very weird, and now that we're already breaking this by changing the type, might be a good opportunity to take a look at the names as well?

schustmi (approved):

bcdurak (approved):

PR #3099, opened 2024-10-18
Better input artifacts typing

avishniakov:
## Describe changes
I implemented helpers to define what is the actual input artifact type in a Step. It can be:
- STEP_OUTPUT for regular case
- EXTERNAL for ExternalArtifact case
- MANUAL for `load_artifact` or `Client().get_artifact_version` inside a step
- LAZY_LOADED for lazy load cases

`StepRunResponse` is now equipped with `input_types` for DAG.

I also extended `Client().get_artifact_version` to link the artifact back to the step as input of MANUAL type.

## Pre-requisites

## Types of changes


src/zenml/orchestrators/step_run_utils.py
```diff
@@ -236,6 +254,39 @@ def _try_to_get_docstring_and_source_code_from_template(
...
+            except RuntimeError:
+                return StepRunInputArtifactType.EXTERNAL
+
+        if artifact.save_type == ArtifactSaveType.MANUAL:
+            return StepRunInputArtifactType.LAZY_LOADED
+
+        if artifact.save_type == ArtifactSaveType.STEP_OUTPUT:
+            try:
+                # TODO: this fails for the cached steps, since artifact was not produced by that parent step,
+                # so STEP_OUTPUT become LAZY_LOADED by mistake. Any thoughts?
```
avishniakov, line 282: @schustmi do you have an idea about this case?

avishniakov:
@Cahllagerfeld this might be interesting for your work with DAG in future.

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -236,6 +247,30 @@ def _try_to_get_docstring_and_source_code_from_template(
...
+            return StepRunInputArtifactType.STEP_OUTPUT
+        elif input_name in step.config.external_input_artifacts:
+            return StepRunInputArtifactType.EXTERNAL
+        elif (
+            input_name in step.config.model_artifacts_or_metadata
+            or input_name in step.config.client_lazy_loaders
+        ):
+            return StepRunInputArtifactType.LAZY_LOADED
+        else:
+            return StepRunInputArtifactType.MANUAL
```
schustmi, line 272: This should not be possible, an artifact can't be manually loaded inside the step if the step didn't actually run yet. Probably best to fail here instead

src/zenml/client.py
```diff
@@ -4139,20 +4140,35 @@ def get_artifact_version(
...
-        return self._get_entity_version_by_id_or_name_or_prefix(
+
+        artifact = self._get_entity_version_by_id_or_name_or_prefix(
             get_method=self.zen_store.get_artifact_version,
             list_method=self.list_artifact_versions,
             name_id_or_prefix=name_id_or_prefix,
             version=version,
             hydrate=hydrate,
         )
+        try:
```
schustmi, line 4160: I'm wondering whether we should actually track this here? Is this really "loading" an input artifact if I just get the response to check some metadata? I guess I'm still using it in some way, even if I don't access the data?

In any case, having this here means we should remove the identical code in `load_artifact`, otherwise we just do the same exact call twice

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -236,6 +247,30 @@ def _try_to_get_docstring_and_source_code_from_template(
 
         return None, None
 
+    def _define_input_type(
```
schustmi, line 250: Maybe this is something that should happen server-side instead? All the information would be available there as well.

src/zenml/models/v2/core/step_run.py
```diff
@@ -156,6 +160,10 @@ class StepRunResponseBody(WorkspaceScopedResponseBody):
         title="The input artifact versions of the step run.",
         default_factory=dict,
     )
+    input_types: Dict[str, StepRunInputArtifactType] = Field(
```
schustmi, line 163: I'm wondering whether we can just include that in the `inputs` somehow? Maybe like this?

```python
class StepRunInputResponse(ArtifactVersionResponse):
  input_type: StepRunInputArtifactType
```

schustmi (changes requested):

src/zenml/client.py
avishniakov: I would say - yes, we should track this, even if one uses metadata only.
I removed the duplication from `load_artifact` funciton.

src/zenml/models/v2/core/step_run.py
avishniakov: reworked as suggested

schustmi (approved):

PR #3140, opened 2024-10-24
Lazy loading improvements

schustmi:
## Describe changes

This is still a draft which contains some questions that need to be answered first before any improvements can be implemented.

Current improvement ideas:
- Separate the object that does the client lazy loading evaluation from the object that we store in the configuration
  - The overwriting of `__getattr__` and `__call__` has some nasty interactions with pydantic methods, which have previously caused some bugs
- Make lazy loading an explicit behaviour (e.g. a `lazy: bool = False` argument on the client methods that support it). Right now it is very intransparent what's happening.
- In addition to the previous point, we could introduce explicit lazy responses for the client methods, which would only have the subset of methods that make sense for these lazy evaluations. This makes it less generic but also causes less issues at runtime.
- Maybe make lazy loadings for models also more transparent by having for example a `LazyModel` class that gets returned by `PipelineContext.model`?

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client_lazy_loader.py
+++ b/src/zenml/client_lazy_loader.py
@@ -208,6 +208,9 @@ def evaluate_all_lazy_load_args_in_client_methods(
                     with contextlib.suppress(ValueError):
                         kwargs[k] = ClientLazyLoader(**v).evaluate()
 
+                # Why do we check for `isinstance(v, ClientLazyLoader)` for the
+                # args but not for `kwargs`
+
             return func(*args_, **kwargs)
 
         return _inner
--- a/src/zenml/orchestrators/input_utils.py
+++ b/src/zenml/orchestrators/input_utils.py
@@ -140,6 +140,11 @@ def resolve_step_inputs(
             )
     for name, cll_ in step.config.client_lazy_loaders.items():
         value_ = cll_.evaluate()
+        # TODO: If we pass something as parameter here, the result is not stored
+        # anywhere. This means the user can not see what value was being passed
+        # into their step, and also not see where exactly this was coming from
+        # (e.g. which artifact/model version). Same thing applies to the model
+        # lazy loaders above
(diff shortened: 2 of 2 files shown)


src/zenml/client_lazy_loader.py
```diff
@@ -208,6 +208,9 @@ def _inner(*args: Any, **kwargs: Any) -> Any:
                     with contextlib.suppress(ValueError):
                         kwargs[k] = ClientLazyLoader(**v).evaluate()
 
+                # Why do we check for `isinstance(v, ClientLazyLoader)` for the
```
schustmi, line 211: This works as expected and evaluates at runtime because `intermediate` it passed as an arg:

```python
@pipeline(enable_cache=False, model=Model(name=MODEL_NAME))
def test_pipeline():
    intermediate = Client().get_model(MODEL_NAME).latest_version_id

    cll = (
        Client()
        .get_model_version(
            MODEL_NAME, intermediate
        )
        .get_artifact(ARTIFACT_NAME)
    )
```

If we pass it as a keyword argument instead, the `intermediate` lazy loader already gets evaluated at pipeline compilation time
```python
@pipeline(enable_cache=False, model=Model(name=MODEL_NAME))
def test_pipeline():
    intermediate = Client().get_model(MODEL_NAME).latest_version_id

    cll = (
        Client()
        .get_model_version(
            MODEL_NAME, model_version_name_or_number_or_id=intermediate
        )
        .get_artifact(ARTIFACT_NAME)
    )
```

schustmi: @avishniakov I tried to understand the reason for the `evaluate_all_lazy_load_args_in_client_methods` decorator, and the only thing I could come up with were these nested lazy loaders. Then I ran into this issue with the args/kwargs and was wondering what the expected behaviour would be in your opinion

PR #3154, opened 2024-10-28
Refactored workload API token management for better security and implemented generic API token dispenser

stefannica:
## Describe changes
This PR heavily reworks the current mechanism of generating API tokens used by pipeline workloads to authenticate to the ZenML server API to implement additional scoping and security validation checks for generated workload API tokens and to remove the previous time-based limitations. Here are some highlights:

* API keys are no longer exposed to the pipeline workload containers
* workload API tokens are no longer time-restricted and can be used to run long-running pipelines without having to reconfigure the default time expiration.
* workload API tokens can be scoped to a schedule, a pipeline run or a pipeline step run (useful for step operators). The validity of these tokens is tied to the lifecycle of the object(s) to which it is scoped:
  * schedule scoped API tokens can only be used to authenticate for as long as the schedule itself exists and is active
  * pipeline run scoped API tokens can only be used to authenticate for as long as the pipeline run exists and the pipeline run hasn't concluded (i.e. the pipeline run status is not FAILED or COMPLETED)
  * step run scoped API tokens can only be used to authenticated for as long as the pipeline run step exists and the step run hasn't concluded (i.e. the step run status is not FAILED or COMPLETED).
* the API key scope and authorized device scope are transferred from the API token used to authorize the client to the generated workload API tokens, just as before

In addition to this, the current restrictions concerning which types of API tokens can be used to generate workload API tokens have been removed. All types of accounts and API tokens (authorized devices, service accounts, dashboard session tokens) can be used to generate API tokens. Additional workload API tokens can be generated by pipeline workloads authenticated with workload API tokens and the scoping rules are applied incrementally:

* a schedule scoped workload API token can be used to generate workload API tokens for the same schedule and any pipeline run and step run
* a pipeline run scoped workload API token can be used to generate workload API tokens for the same pipeline run and any step run
* a step run scoped workload API token can only be used to generate workload API tokens for the same step run 

These improvements add a layer of security to workload API tokens that was previously only implemented via time-based expiration that was difficult to work with in the context of long-running pipelines:

* API tokens scoped for a schedule are only valid while the schedule is configured and active. If accidentally leaked, they can be invalidated by deactivating or deleting the schedule.
* API tokens scoped for a pipeline run are only valid while the pipeline is running. Once the pipeline is completed or failed, they are no longer valid.
* API tokens scoped for a step run are only valid while the step is running. Once the step is completed or failed, they are no longer valid.
* in addition to the above, all generated workload tokens inherit the API key scope or authorized device scope from the current authorization context. Therefore deleting, rotating or disabling an API key or authorized device automatically invalidates all workload API tokens generated while authenticated with that API key or authorized device.

The existing api_token endpoint can now also be used to issue generic API tokens than expire after one hour and can be used to make programmatic calls to the API for automation purposes.

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

- **New Features**
	- Introduced a new attribute for managing the lifetime of generic API tokens.
	- Enhanced OAuth login process with additional parameters for token generation.
	- Added a logging mechanism to warn about non-expiring API tokens.

- **Bug Fixes**
	- Improved security by ensuring appropriate token management based on authentication methods.

- **Documentation**
	- Updated documentation to include details about the new API token lifetime attribute.


htahir1:
@coderabbitai review

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -522,6 +529,20 @@ def api_token(
             detail="Not authenticated.",
         )
 
+    if not token.device_id and not token.api_key_id:
+        config = server_config()
+
+        # If not authenticated with a device or a service account, then a
+        # short-lived generic API token is returned.
+        return generate_access_token(
```
schustmi, line 537: I don't understand this early return here: Why is the verifying whether the user has permissions to create a pipeline run after this check? Is it not possible to request a pipeline scoped token without a device or API key?

My understanding would be that we want to maybe do this permission check if `pipeline_id` or `schedule_id` are set? Even that's not perfect in case of an unlisted pipeline, but covers more cases than the current implementation?

src/zenml/zen_server/routers/auth_endpoints.py
```diff
@@ -193,9 +193,12 @@
 
 def generate_access_token(
     user_id: UUID,
-    response: Response,
+    response: Optional[Response] = None,
     device: Optional[OAuthDeviceInternalResponse] = None,
     api_key: Optional[APIKeyInternalResponse] = None,
+    expires_in: Optional[int] = None,
+    pipeline_id: Optional[UUID] = None,
+    schedule_id: Optional[UUID] = None,
```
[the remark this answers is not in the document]
schustmi: This seems to be missing indeed

src/zenml/config/server_config.py
```diff
@@ -257,6 +260,10 @@ class ServerConfiguration(BaseModel):
     device_expiration_minutes: Optional[int] = None
     trusted_device_expiration_minutes: Optional[int] = None
 
+    generic_api_token_lifetime: int = (
```
schustmi, line 263: We can use `PositiveInt` from `pydantic` to do this validation for us automatically (or `NonNegativeInt` in case we want to allow 0 seconds?)

schustmi (changes requested):

htahir1:
@schustmi Small note: If we can get this PR over the line and give to aagya we might be able to do it before she leaves... not sure but just wanted to add that if you can "take this over" so to speak and its an easy win, id do that

schustmi:
Yeah I wanted to, but really unsure about the early return comment, I think we should probably wait for Stefan's input on this

src/zenml/orchestrators/utils.py
```diff
@@ -113,6 +115,17 @@ def get_config_environment_vars(
             # If a schedule is given, this is a long running pipeline that
             # should not have an API token that expires.
             expires_minutes = None
+            logger.warning(
```
schustmi, line 118: @stefannica maybe I'm missing something, but we show this warning message even if the user is currently authenticated with device/SA, no?

schustmi (changes requested):

src/zenml/zen_server/routers/auth_endpoints.py
stefannica: @schustmi , this endpoint is to be used in two distinct situations:

* from the dashboard, to generate 1 hour API tokens and display them to users (which is why we return early here)
* from the CLI, to generate these so called "workload" API tokens used by pipeline steps to connect to ZenML

The distinction is made by how the user is authenticated: if it's authenticated with an API key or through the web login (authorized device) flow, they are issued a "workload" API token (for which either a pipeline ID or schedule ID or both should theoretically be set, but I'll leave it to you to decide if we can also enforce that rule). If it's not an API key nor a device, it's a dashboard API token, in which case we issue a generic 1 hour token.

You can totally take this over and do whatever you like with this endpoint if you have better ideas for how we should make this available to users. My only concern is this: API tokens that are not tied to an API key nor an authorized device should be short lived, because they cannot be revoked in case of security incidents unless the entire user account is deactivated or deleted.

src/zenml/orchestrators/utils.py
stefannica: Not after merging in my other PR (zenml login).

stefannica:
@coderabbitai review

src/zenml/zen_server/auth.py
```diff
@@ -329,6 +330,84 @@ def authenticate_credentials(
...
+                ExecutionStatus.COMPLETED,
+            ]:
+                error = (
+                    f"The execution of step run "
+                    f"{decoded_token.step_run_id} has already concluded and "
+                    "API tokens scoped to it are no longer valid."
+                )
+                logger.error(error)
+                raise CredentialsNotValid(error)
+
```
schustmi, line 410: If I get this right, every time such a workload token is used to send a request to the server, this code is running, correct? Do you think this will cause any trouble at scale as this is running up to 3 additional database queries per request?

schustmi (commented):
Love the way this turned out, just one small comment from my side

src/zenml/zen_server/auth.py
stefannica: I guess I could implement a local time-based cache that would delay re-fetching the same object from the DB for up to 30 seconds. We should keep this contained to a limited number of entries though, otherwise the memory requirements will explode during high loads.

schustmi: Potentially yeah, but not necessary for now I hope. We're considerably reducing the amounts of server requests during pipeline execution in this and the next release, which means this hopefully doesn't matter as much

schustmi (approved):

src/zenml/zen_server/auth.py
stefannica: @schustmi I added a simple in-memory cache for objects that we want to cache in the ZenML server

src/zenml/zen_server/cache.py
```diff
@@ -0,0 +1,195 @@
...
+    automatically removes expired entries and evicts the oldest entry when
+    the cache reaches its maximum capacity.
+
+
+    Usage Example:
+
+        cache = MemoryCache()
+        uuid_key = UUID("12345678123456781234567812345678")
+
+        cached_or_real_object = cache.get_or_cache(
```
schustmi, line 65: I don't see the `get_or_cache` method anywhere

src/zenml/zen_server/cache.py
```diff
@@ -0,0 +1,195 @@
...
+    def _cleanup(self) -> None:
+        """Remove expired or excess entries."""
+        # Remove expired entries
+        keys_to_remove = [k for k, v in self.cache.items() if v.expired]
+        for k in keys_to_remove:
+            del self.cache[k]
+
+        # Ensure we don't exceed max capacity
+        while len(self.cache) > self.max_capacity:
+            self.cache.popitem(last=False)  # type: ignore[call-arg]
```
schustmi, line 134: Is it possible to annotate `self.cache: OrderedDict[...]` to get rid of this type ignore?

src/zenml/zen_server/cache.py
stefannica: Ah, you're right, this was left behind from an older implementation version.

src/zenml/zen_server/cache.py
stefannica: I read that `OrderedDict` isn't available as a type annotation in Python 3.8 and decided to use `Dict` instead. But then I realized we don't support Python 3.8 anymore so... yeah, sure :)

src/zenml/zen_server/cache.py
stefannica: done

src/zenml/zen_server/cache.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+F = Callable[[UUID], Optional[Any]]
```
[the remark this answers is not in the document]
stefannica: done

src/zenml/zen_server/cache.py
stefannica: done

schustmi (approved):

PR #3156, opened 2024-10-29
Enable cache precomputation for run templates

schustmi:
## Describe changes

Because no placeholder run was being passed when running a template, the client-side cache computation was not being considered. This PR fixes that by passing the placeholder run, and additionally improves the placeholder run cleanup behaviour and some return types/docstrings.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/pipelines/run_utils.py
+++ b/src/zenml/pipelines/run_utils.py
@@ -22,6 +22,7 @@ from zenml.models import (
     PipelineRunResponse,
     StackResponse,
 )
+from zenml.orchestrators.publish_utils import publish_failed_pipeline_run
 from zenml.orchestrators.utils import get_run_name
 from zenml.stack import Flavor, Stack
 from zenml.utils import code_utils, notebook_utils, source_utils
@@ -119,6 +120,7 @@ def deploy_pipeline(
     deployment: "PipelineDeploymentResponse",
     stack: "Stack",
     placeholder_run: Optional["PipelineRunResponse"] = None,
+    cleanup_placeholder_run: bool = False,
 ) -> None:
     """Run a deployment.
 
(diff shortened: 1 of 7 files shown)


src/zenml/pipelines/run_utils.py
```diff
@@ -119,41 +120,53 @@ def deploy_pipeline(
...
-            placeholder_run
-            and Client().get_pipeline_run(placeholder_run.id).status
-            == ExecutionStatus.INITIALIZING
-        ):
-            # The run hasn't actually started yet, which means that we
-            # failed during initialization -> We don't want the
-            # placeholder run to stay in the database
-            Client().delete_pipeline_run(placeholder_run.id)
-
+        # TODO: Ideally this would not run if the orchestrator already created a run on their end?
```
schustmi, line 168: @bcdurak I remember after your PR with the orchestrator URLs, we ran into a case where the placeholder run was deleted but we would have wanted it to still be there, so we could see the URLs?
If I remember correctly, it was a syncronous orchestrator that failed to e.g. download the Docker image for the first step, which would lead to a failure where the logs URL would have been interesting.

Any idea how to best implement this? The best option I can come up with would be to separate the orchestrator into two different methods:
- one that does everything until the job is submitted to the orchestrator
- another one that waits in case the orchestrator is running sync

After the first method completes, we don't delete the placeholder run anymore.

schustmi: Or another suggestion: If the run has any metadata associated with it (e.g. the orchestrator URLs), we don't delete it anymore but instead just mark it as failed. That might be a much easier and quicker solution than my other suggestion, and will in most cases achieve the same thing?

schustmi: @bcdurak As discussed, we now don't delete the placeholder run anymore.

src/zenml/zen_server/template_execution/runner_entrypoint_configuration.py
```diff
@@ -36,4 +34,9 @@ def run(self) -> None:
         stack = Client().active_stack
         assert deployment.stack and stack.id == deployment.stack.id
 
-        deploy_pipeline(deployment=deployment, stack=stack)
+        placeholder_run = get_placeholder_run(deployment_id=deployment.id)
```
bcdurak, line 37: Just a little bit unsure about this part. Is the relationship between a `deployment_id` and a `placeholder_run` always 1-to-1?

bcdurak (approved):
I like the changes with the cleanup and I am a bit surprised that our docstring checks did not fail with some of the things that you have fixed now. I left a small comment but otherwise it looks good. Feel free to merge.

src/zenml/zen_server/template_execution/runner_entrypoint_configuration.py
schustmi: It's 1 to 0 actually, which is why this function returns an optional placeholder run.

PR #3158, opened 2024-10-31
Add CLI command to export stack requirements

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/how-to/infrastructure-deployment/stack-deployment/export-stack-requirements.md
+++ b/docs/book/how-to/infrastructure-deployment/stack-deployment/export-stack-requirements.md
@@ -1,0 +1,13 @@
+---
+description: Export stack requirements
+---
+
+You can get the `pip` requirements of your stack by running the `zenml stack export-requirements <STACK-NAME>` CLI command.
+
+To install those requirements, it's best to write them to a file and then install them like this:
+```bash
+zenml stack export-requirements <STACK-NAME> --output-file stack_requirements.txt
+pip install -r stack_requirements.txt
+```
+
+<figure><img src="https://static.scarf.sh/a.png?x-pxid=f0b4f458-0a54-4fcd-aa95-d5ee424815bc" alt="ZenML Scarf"><figcaption></figcaption></figure>
--- a/docs/book/user-guide/production-guide/understand-stacks.md
+++ b/docs/book/user-guide/production-guide/understand-stacks.md
@@ -53,6 +53,8 @@ As you can see a stack can be **active** on your **client**. This simply means t
 {% endtab %}
 {% endtabs %}
 
+You can get the `pip` requirements of your stack by running the `zenml stack export-requirements <STACK-NAME>` CLI command.
+
 ## Components of a stack
 
 As you can see in the section above, a stack consists of multiple components. All stacks have at minimum an **orchestrator** and an **artifact store**.
--- a/docs/book/user-guide/production-guide/understand-stacks.md
+++ b/docs/book/user-guide/production-guide/understand-stacks.md
@@ -14,10 +14,12 @@ A `stack` is the configuration of tools and infrastructure that your pipelines c
 
 As visualized in the diagram above, there are two separate domains that are connected through ZenML. The left side shows the code domain. The user's Python code is translated into a ZenML pipeline. On the right side, you can see the infrastructure domain, in this case, an instance of the `default` stack. By separating these two domains, it is easy to switch the environment that the pipeline runs on without making any changes in the code. It also allows domain experts to write code/configure infrastructure without worrying about the other domain.
 
+{% hint style="info" %}
+You can get the `pip` requirements of your stack by running the `zenml stack export-requirements <STACK-NAME>` CLI command.
+{% endhint %}
+
 ### The `default` stack
 
-{% tabs %}
-{% tab title="CLI" %}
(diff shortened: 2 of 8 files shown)


htahir1 (changes requested):
Docs pplease :-)

schustmi:
@htahir1 Where? I don't think every random CLI command should have a separate docs page

src/zenml/cli/stack.py
```diff
@@ -1944,3 +1944,79 @@ def query_region(
...
+
+    try:
+        stack_model: "StackResponse" = Client().get_stack(
+            name_id_or_prefix=stack_name_or_id
+        )
+    except KeyError as err:
+        cli_utils.error(str(err))
+
+    stack_ = Stack.from_model(stack_model)
+    requirements = list(stack_.requirements())
```
bcdurak, line 2000: As far as I know, the `stack_.requirements` call here combines the requirements of this stack's components, which combines the requirements of its integrations. However in order to be able to do this, you initialize the `Stack` object.

Since you have the `stack_model` here, I think we can avoid initializing the `Stack` object here. Each component within this `stack_model` should already have an `integration` attribute which can help you get the proper requirements. Moreover, this way we could support the main use-case that this CLI function would help with: 

- I see a stack -> I want to use the stack -> I need to install the requirements for it first -> I do export-requirements and install it -> I use the stack.

Otherwise, you need to have them installed before exporting the requirements.

bcdurak (changes requested):

bcdurak:
@schustmi I am not sure how I feel about this myself, but do you think it might be nice to have a `zenml stack install-requirements` as well?

htahir1:
@schustmi It isn't a "random" CLI command (otherwise we wouldn't really need it). It's a CLI command that many of our users want and should be more aware off. I would even say we can upgrade the dashboard to include it instead of the zenml integration install we can ask them to do: `zenml stack export-requirements <STACK_NAME> | xargs pip install` or something. 

image

Just a cursory glance through the docs show me some places: 

- As a type hint here https://docs.zenml.io/user-guide/production-guide/understand-stacks
- As a new docs page in this section that talks about "How to manage stack requirements"  https://docs.zenml.io/how-to/stack-deployment

bcdurak:
@schustmi yeah, it did not feel right as I was writing that 😄

src/zenml/cli/stack.py
schustmi: Yeah I was just too lazy to do this, I now used the util method that we already have for this :)

bcdurak: Love it! It enables the use-case that I mentioned, which is very nice 👍

bcdurak (approved):
🦭

schustmi:
> @schustmi It isn't a "random" CLI command (otherwise we wouldn't really need it). It's a CLI command that many of our users want and should be more aware off. I would even say we can upgrade the dashboard to include it instead of the zenml integration install we can ask them to do: `zenml stack export-requirements <STACK_NAME> | xargs pip install` or something.
> 
> image
> Just a cursory glance through the docs show me some places:
> 
> * As a type hint here https://docs.zenml.io/user-guide/production-guide/understand-stacks
> * As a new docs page in this section that talks about "How to manage stack requirements"  https://docs.zenml.io/how-to/stack-deployment

I disagree with both the usefulness and especially the usefullness of a docspage. The dashboard command you linked does not work, because we almost always have other log messages that would mess with the command output, which makes it useless unless writing to a file.

htahir1:
@schustmi cant we make the implementation in a way that we can then pip it to xargs and a pip command? Surely that's the right UX here? even id use it all the time , because I could put it into poetry or uv or conda like that?
as for the dashboard, that's fine to ignore for now but it is the one remaining mention of zenml integration install, which you have already noted, sucks. 

As for the docs page, it is useless if not treated in the right way, ill leave some suggestions

docs/book/how-to/infrastructure-deployment/stack-deployment/export-stack-requirements.md
```diff
@@ -0,0 +1,13 @@
+---
+description: Export stack requirements
+---
+
+You can get the `pip` requirements of your stack by running the `zenml stack export-requirements <STACK-NAME>` CLI command.
```
htahir1, line 5: This chapter IMO is not about exporting stack requirements. It's about telling a story about:

- Why do we need to install stack requirements locally and in the execution environment?
- What does it mean to export requirements? (i.e. that these requirements are the suggested ranges of values that version of zenml is released with)
- A reason why someone would do this in a production environment
- What is the link between a requirements file generated by this file vs the stack integration installing dependencies that zenml already does when it builds a docker image

htahir1 (changes requested):

docs/book/user-guide/production-guide/understand-stacks.md
```diff
@@ -53,6 +53,8 @@ As you can see a stack can be **active** on your **client**. This simply means t
 {% endtab %}
 {% endtabs %}
 
+You can get the `pip` requirements of your stack by running the `zenml stack export-requirements <STACK-NAME>` CLI command.
```
htahir1, line 56: Put it in a hint maybe>?

htahir1 (approved):
Great work!!

PR #3162, opened 2024-10-31
Cleanup materializer temporary directories after step execution

schustmi:
## Describe changes
The temporary directory cleanup in our materializer was all over the place. This PR refactors it to use a common helper method. This helper method either
- cleans up the directory immediately after the context manager finishes
- cleans up the directory after the current step execution finishes. This is useful for materializers that copy data into a local directory and need it to stay on disk for lazy loading. Once the step execution finishes however, the data is not needed anymore as the step function ended.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/huggingface/materializers/huggingface_tf_model_materializer.py
+++ b/src/zenml/integrations/huggingface/materializers/huggingface_tf_model_materializer.py
@@ -63,7 +63,7 @@ class HFTFModelMaterializer(BaseMaterializer):
         Args:
             model: The TF Model to write.
         """
-        with self.get_temporary_directory(delete_at_exit=False) as temp_dir:
+        with self.get_temporary_directory(delete_at_exit=True) as temp_dir:
             model.save_pretrained(temp_dir)
             io_utils.copy_dir(
                 temp_dir,
--- a/src/zenml/integrations/huggingface/materializers/huggingface_tf_model_materializer.py
+++ b/src/zenml/integrations/huggingface/materializers/huggingface_tf_model_materializer.py
@@ -15,7 +15,6 @@
 
 import importlib
 import os
-from tempfile import TemporaryDirectory
 from typing import Any, ClassVar, Dict, Tuple, Type
 
 from transformers import (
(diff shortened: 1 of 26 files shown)


src/zenml/integrations/huggingface/materializers/huggingface_tf_model_materializer.py
```diff
@@ -46,30 +45,30 @@ def load(self, data_type: Type[TFPreTrainedModel]) -> TFPreTrainedModel:
...
-
-        config = AutoConfig.from_pretrained(temp_dir.name)
-        architecture = "TF" + config.architectures[0]
-        model_cls = getattr(
-            importlib.import_module("transformers"), architecture
-        )
-        return model_cls.from_pretrained(temp_dir.name)
+        with self.get_temporary_directory(delete_at_exit=False) as temp_dir:
+            io_utils.copy_dir(
+                os.path.join(self.uri, DEFAULT_TF_MODEL_DIR), temp_dir
```
bcdurak, line 50: I saw this pattern in some other materializers as well. Does the removal of `.name` lead to any different behavior or is it just the same?

src/zenml/materializers/base_materializer.py
```diff
@@ -326,3 +329,67 @@ def _extract_base_metadata(self, data: Any) -> Dict[str, "MetadataType"]:
...
+                used outside of the context of a step execution, the temporary
+                directory will not be deleted and the user is responsible for
+                deleting it themselves.
+
+        Yields:
+            Path to the temporary directory.
+        """
+        temp_dir = tempfile.mkdtemp(prefix="zenml-")
+
+        if delete_after_step_execution and not delete_at_exit:
```
bcdurak, line 356: So, if I get it correctly, if you set `delete_at_exit` to `True`, setting `delete_after_step_execution` to `True` or `False` will not have any effect on how it is being cleaned up. I was initially worried that both being `True` would lead us to try to delete the same directory twice and eventually a `FileNotFoundError` but I guess this is not the case here.

bcdurak (approved):
Such a great solution. Feel free to merge 🦭

PR #3164, opened 2024-10-31
Add artifact version batch request

schustmi:
## Describe changes
This PR implements an artifact version batch endpoint that is used to reduce the amount of requests needed to store the output artifacts at the end of a step.

## Pre-requisites

## Types of changes


src/zenml/zen_server/rbac/endpoint_utils.py
```diff
@@ -96,6 +96,48 @@ def verify_permissions_and_create_entity(
...
+
+    for request_model in batch:
+        if isinstance(request_model, UserScopedRequest):
+            if request_model.user != auth_context.user.id:
+                raise IllegalOperationError(
+                    f"Not allowed to create resource '{resource_type}' for a "
+                    "different user."
+                )
+
+    verify_permission(resource_type=resource_type, action=Action.CREATE)
```
bcdurak, line 130: I guess this is not a use case that we might bump into right now, but I am asking this with the future in mind. When we implement these batch-creation endpoints, what happens if only one entity fails during the creation (due to RBAC or any other reason like `EntityExists` errors) Should we fail the whole batch or only exclude the failing ones?

bcdurak (approved):
Overall, it looks good. Left a small comment regarding the batch creation logic.

src/zenml/artifacts/utils.py
```diff
@@ -82,20 +82,129 @@
...
+) -> List[ArtifactVisualizationRequest]:
+    """Save artifact visualizations.
+
+    Args:
+        data: The data for which to save the visualizations.
+        materializer: The materializer that should be used to generate and
+            save the visualizations.
+
+    Returns:
+        List of requests for the saved visualizations.
```
avishniakov, line 96: 1. I don't see that return is used in the code
2. If it is used, why we return Request for already existing entity - shall be Response, IMO

src/zenml/zen_server/rbac/endpoint_utils.py
avishniakov: For the creation, we only validate that the type is allowed to be created. I think what you have described can not happen.

avishniakov (approved):
Good addition for API. Only one minor comment from my end

src/zenml/artifacts/utils.py
schustmi: Unless I'm missing something, the return should be right below?
This function saves the visualizations into the artifact store, and returns the list of requests that reference the paths where they were stored. They get sent to the server as part of the `ArtifactVersionRequest(visualizations=...)`.

Maybe the function name should be different? `_save_artifact_visualizations_to_artifact_store`?

src/zenml/zen_server/rbac/endpoint_utils.py
schustmi: @bcdurak I think the comment on this line is very misleading 😅
As @avishniakov already pointed out, this check is only happening once, and it's happening before any item of the batch is even touched. But I think in general, yes it might happen that item 5 of the batch fails during it's creation.

In that case, we could:
- roll back all previous changes and return a failure for the request
- continue with the next item in the batch, and have the response flexible that it can indicate which items were created and which failed

Currently, we do none of that and simply loop over the batch and once we fail we fail, but keep the already created items in the DB.
I don't think that's ideal, but this PR simply mirrors the previous behaviour for artifact creation at step end, only that it happens with one request. Definitely something to improve on in the future!

src/zenml/artifacts/utils.py
avishniakov: You are right the return is used indeed. And I unrolled this logic better from the code and checked how it is used - all good

PR #3167, opened 2024-11-04
Prevent some race conditions

schustmi:
## Describe changes
This PR adds some unique constraint to avoid race conditions when creating entities, and also moves the retry logic for model version creation to the server.

## Pre-requisites

## Types of changes


src/zenml/exceptions.py
```diff
@@ -168,6 +168,10 @@ class EntityExistsError(ZenMLBaseException):
     """Raised when trying to register an entity that already exists."""
 
 
+class EntityCreationError(ZenMLBaseException):
+    """Raised when failing to create an entity."""
+
+
```
avishniakov, line 174: I bet we decided to stop creating custom exceptions. At least @stefannica always told me some for the core.

avishniakov (approved):
Only a minor comment, assuming tests will pass smoothly.

src/zenml/exceptions.py
schustmi: I talked to @stefannica and we agreed it makes sense in this case.

avishniakov: ok

PR #3182, opened 2024-11-08
New `log_metadata` function, new `oneof` filtering, additional `run_metadata` filtering

bcdurak:
## Describe changes

This PR features **three** main changes:

### The new `log_metadata` function

Now, all the metadata-creating functions are gathered under one method called `log_metadata`. It is possible to call this method with different inputs to log run metadata for artifact versions, model versions, steps, and runs.

### The `oneof` filtering

This allows us to filter entities using a new operator called `oneof`. You can possibly use this with IDs, tags, and all the other attributes like `PipelineRunFilter(tag='oneof:["cats", "dogs"]')`.

### The `run_metadata` filtering

The custom filtering logic for runs, steps, artifact-, and model versions has been improved. Now, you can filter these entities using filters like `PipelineRunFilter(run_metadata={"accuracy": "gt:0.85"})`.

#### Further changes

- The old metadata creation functions have been deprecated.

### The remaining TODOs

- Docs updates.
- Template updates.
- Project updates.

## Pre-requisites

## Types of changes


src/zenml/artifacts/utils.py
```diff
@@ -417,6 +417,10 @@ def log_artifact_metadata(
             called inside a step with a single output, or, if neither an
             artifact nor an output with the given name exists.
     """
+    logger.warning(
+        "The `log_artifact_metadata` function is deprecated and will soon be "
+        "removed. Please use `log_metadata` instead."
+    )
     try:
         step_context = get_step_context()
         in_step_outputs = (artifact_name in step_context._outputs) or (
```
avishniakov, line 426: Maybe use a call to your new `log_metadata` instead of old body?

src/zenml/model/utils.py
```diff
@@ -56,6 +56,11 @@ def log_model_metadata(
         ValueError: If no model name/version is provided and the function is not
             called inside a step with configured `model` in decorator.
     """
+    logger.warning(
+        "The `log_model_metadata` function is deprecated and will soon be "
+        "removed. Please use `log_metadata` instead."
+    )
+
     if model_name and model_version:
         from zenml import Model
```
avishniakov, line 66: Maybe use a call to your new `log_metadata` instead of old body?

src/zenml/models/v2/base/filter.py
```diff
@@ -190,6 +196,39 @@ def generate_query_conditions_from_column(self, column: Any) -> Any:
...
+                    return and_(
+                        numeric_column, numeric_column >= float(self.value)
+                    )
+                if self.operation == GenericFilterOps.LTE:
+                    return and_(
+                        numeric_column, numeric_column <= float(self.value)
+                    )
+            except Exception as e:
+                raise ValueError(
+                    f"Failed to compare the column to the numeric value: {e}"
```
avishniakov, line 230: This could be more clear: which column, what was the value, etc.

src/zenml/models/v2/base/filter.py
```diff
@@ -598,6 +640,23 @@ def _resolve_operator(value: Any) -> Tuple[Any, GenericFilterOps]:
...
+
+            if operator == operator.ONEOF:
+                try:
+                    value = json.loads(value)
+                    if not isinstance(value, list):
+                        raise ValueError(
+                            "When you are using the 'oneof:' filtering "
+                            "make sure that the provided value is a json "
+                            "formatted list."
+                        )
```
avishniakov, line 652:
```suggestion
                        raise ValueError
```
No need to duplicate the text

src/zenml/models/v2/base/filter.py
```diff
@@ -1043,6 +1107,38 @@ def _define_uuid_filter(
...
+            A Filter object.
+
+        Raises:
+            ValueError: If the value is not a proper value.
+        """
+        # For equality checks, ensure that the value is a valid UUID.
+        if operator == GenericFilterOps.ONEOF and not isinstance(value, list):
+            raise ValueError(
+                "If you are using `oneof:` as a filtering op, the value needs "
+                "to be a json formatted list string."
```
avishniakov, line 1131: Here and in all other places, maybe make this a constant in `filter.py` since it is quite repetitive, IMO?

src/zenml/steps/utils.py
```diff
@@ -442,6 +442,11 @@ def log_step_metadata(
...
             the function is not called from within a step.
     """
+    logger.warning(
+        "The `log_step_metadata` function is deprecated and will soon be "
+        "removed. Please use `log_metadata` instead."
+    )
+
     step_context = None
     if not step_name:
         with contextlib.suppress(RuntimeError):
```
avishniakov, line 452: Pls, replace the body with a new function call.

src/zenml/utils/metadata_utils.py
```diff
@@ -0,0 +1,330 @@
...
+from zenml.enums import MetadataResourceTypes
+from zenml.logger import get_logger
+from zenml.metadata.metadata_types import MetadataType
+from zenml.steps.step_context import get_step_context
+
+logger = get_logger(__name__)
+
+
+@overload
+def log_metadata(metadata: Dict[str, MetadataType]) -> None: ...
```
avishniakov, line 29: I'm curious what all those overloads give us here. Params are optional anyway...

avishniakov (changes requested):

src/zenml/utils/metadata_utils.py
bcdurak: You are right, in the original implementation, all the parameters are optional but the overloads here give us an advantage in two significant ways. 

- It helps us with `mypy` linting to see if the function is used in anyway that it shouldn't be.
- It helps the developer as well. If I write `log_metadata(metadata={}, artifact_name=..., step_name=...)`, my IDE warns me that this is not an allowed usage of the function.

src/zenml/models/v2/base/filter.py
bcdurak: Fixed this.

src/zenml/models/v2/base/filter.py
bcdurak: Fixed this as well.

src/zenml/models/v2/base/filter.py
```diff
@@ -211,6 +254,9 @@ def _remove_hyphens_from_value(cls, value: Any) -> Any:
         if isinstance(value, str):
             return value.replace("-", "")
 
+        if isinstance(value, list):
+            return [str(v).replace("-", "") for v in value]
```
schustmi, line 258: What if it's a list of `int`s?

src/zenml/models/v2/base/filter.py
```diff
@@ -598,6 +648,19 @@ def _resolve_operator(value: Any) -> Tuple[Any, GenericFilterOps]:
             ):
                 value = split_value[1]
                 operator = GenericFilterOps(split_value[0])
+
+            if operator == operator.ONEOF:
+                try:
+                    value = json.loads(value)
```
schustmi, line 654: Mhm I'm wondering whether json is the best format here? Wouldn't that be very weird as e.g. a query param in the URL? `/stacks?id=oneof:'["<ID>", "ID"]'`?

src/zenml/utils/metadata_utils.py
```diff
@@ -0,0 +1,339 @@
...
+            metadata=metadata,
+            resource_id=artifact_version_model.id,
+            resource_type=MetadataResourceTypes.ARTIFACT_VERSION,
+        )
+
+    # If every additional value is None, that means we are calling it bare bones
+    # and this call needs to happen during a step execution. We will use the
+    # step context to fetch the step, run and possibly the model version and
+    # attach the metadata accordingly.
+    elif all(
```
schustmi, line 271: I'm not sure that I like the current logic: If this is being called from a step and the user provides an artifact name, don't you think we should also link it to the pipeline run, step run and model version?

src/zenml/utils/metadata_utils.py
```diff
@@ -0,0 +1,339 @@
...
+        artifact_version: The version of the artifact.
+        model_version_id: The ID of the model version.
+        model_name: The name of the model.
+        model_version: The version of the model
+
+    Raises:
+        ValueError: If no identifiers are provided and the function is not
+            called from within a step.
+    """
+    # Initialize the client
```
schustmi, line 125:
```suggestion
```

src/zenml/utils/metadata_utils.py
```diff
@@ -0,0 +1,339 @@
...
+                resource_id=step_model.model_version.id,
+                resource_type=MetadataResourceTypes.MODEL_VERSION,
+            )
+
+    # If a step is identified by id, fetch it directly through the client,
+    # follow a similar procedure and log metadata for its pipeline and model
+    # as well.
+    elif step_id is not None:
+        step_model = client.get_run_step(step_run_id=step_id)
+        run_model = client.get_pipeline_run(
```
schustmi, line 158: This is just an unnecessary server call, why fetch the pipeline run if you only need its ID that you already have?

schustmi (changes requested):

src/zenml/utils/metadata_utils.py
bcdurak: Removed it.

src/zenml/utils/metadata_utils.py
bcdurak: Removed the redundant call.

src/zenml/models/v2/base/filter.py
```diff
@@ -190,6 +199,40 @@ def generate_query_conditions_from_column(self, column: Any) -> Any:
             return column.endswith(f"{self.value}")
         if self.operation == GenericFilterOps.NOT_EQUALS:
             return column != self.value
+        if self.operation == GenericFilterOps.ONEOF:
```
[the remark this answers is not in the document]
bcdurak: I added a pydantic `after` validator to do the check. 

Btw, we do the same check within `_resolve_operator` and `_define_uuid_filter` as well. It is a similar pattern for the other types too. Perhaps, it would be nice to optimize this logic.

src/zenml/utils/metadata_utils.py
schustmi: Yes 100% agreed

src/zenml/models/v2/base/filter.py
bcdurak: As we discussed, I will leave this as it is. But there is definitely room for improvement here.

avishniakov (approved):
My comments are addressed or rejected, so good to go from my end, as soon as tests are all green (I shipped fixes for all of them, besides mac 3.9 slow - ignore that one)

schustmi (approved):

src/zenml/utils/metadata_utils.py
bcdurak: This will be solved in a follow-up PR.

PR #3193, opened 2024-11-13
Follow-up on the `run_metadata` changes

bcdurak:
## Describe changes

This PR optimizes the way that we store run metadata related to different entities. (addresses [the review comment here](https://github.com/zenml-io/zenml/pull/3182#discussion_r1836301857))

In our old implementation, when someone calls `log_metadata`, it is possible that they attach the same metadata to different entities such as pipeline runs, step runs, and model versions. This process used to create X different entries with the same key-value pair in the metadata table.

In order to optimize this process, this PR separates the previous run metadata table into two tables: 

- One that holds the actual key-value pair.
- One that links this pair to different resources.

For this to work, I also implemented a new `RunMetadataRequest` model, that can hold more than one entity per key-value pair. In all the previous calls that used this request model, all affected models and schemas were adjusted accordingly.

## Other related changes

- [Preview for the docs changes](https://zenml-io.gitbook.io/barisky-docs/how-to/model-management-metrics/track-metrics-metadata)
- [Starter Template PR](https://github.com/zenml-io/template-starter/pull/27)
- [E2E Batch Template PR](https://github.com/zenml-io/template-e2e-batch/pull/47)
- [NLP Template PR](https://github.com/zenml-io/template-nlp/pull/27)
- [LLM Finetuning Template PR](https://github.com/zenml-io/template-llm-finetuning/pull/11)

## Remaining TODOs
- Add more tests (model_version_id logging)

## Pre-requisites

## Types of changes


bcdurak:
@htahir1 It would be great if you could also take a look at the docs changes.

docs/book/how-to/model-management-metrics/track-metrics-metadata/attach-metadata-to-steps.md
htahir1, line 1: As you deletd this you might want to fix other links + add a redirect in .gitbook.yaml

htahir1 (approved):
Well done! Just one small thing

docs/book/how-to/model-management-metrics/track-metrics-metadata/attach-metadata-to-a-model.md
```diff
@@ -1,62 +1,90 @@
 ---
-description: >-
-  Attach any metadata as key-value pairs to your models for future reference and
-  auditability.
+description: Learn how to attach metadata to a model.
 ---
 
 # Attach metadata to a model
```
avishniakov, line 5: This sections deserves an addition to explain people how they shall handle models from context (https://github.com/zenml-io/template-llm-finetuning/pull/11#discussion_r1843281476)

src/zenml/client.py
```diff
@@ -4435,23 +4435,17 @@ def _delete_artifact_from_artifact_store(
     def create_run_metadata(
         self,
         metadata: Dict[str, "MetadataType"],
-        resource_id: UUID,
-        resource_type: MetadataResourceTypes,
+        resources: List[Tuple[UUID, MetadataResourceTypes]],
```
avishniakov, line 4438: Maybe it deserves a proper class since this is user-facing and it is usually quite hard to understand those Tuple args, at least for someone like me.

src/zenml/client.py
```diff
@@ -4435,23 +4435,17 @@ def _delete_artifact_from_artifact_store(
...
         stack_component_id: Optional[UUID] = None,
     ) -> None:
         """Create run metadata.
 
         Args:
             metadata: The metadata to create as a dictionary of key-value pairs.
-            resource_id: The ID of the resource for which the
-                metadata was produced.
-            resource_type: The type of the resource for which the
+            resources: The ID and type of the resources for that the
```
avishniakov, line 4445:
```suggestion
            resources: The list of IDs and types of the resources for that the
```

src/zenml/models/v2/core/run_metadata.py
```diff
@@ -30,14 +30,12 @@
 class RunMetadataRequest(WorkspaceScopedRequest):
     """Request model for run metadata."""
 
-    resource_id: UUID = Field(
-        title="The ID of the resource that this metadata belongs to.",
-    )
-    resource_type: MetadataResourceTypes = Field(
-        title="The type of the resource that this metadata belongs to.",
+    resources: List[Tuple[UUID, MetadataResourceTypes]] = Field(
```
avishniakov, line 33: Same as above

avishniakov (changes requested):
I have requested some changes in the code and I will also chat with you today, since I have some metadata reusability concern to clarify.

avishniakov:
@bcdurak was nice to talk to you today, here is a brief summary of potential improvements with the current metadata:
- Metadata values duplication. It is not a problem for small entries, but it becomes heavy for long metadata entries since they will be logged again and again. It can be solved by normalizing the metadata table a bit and extracting the `value` to a separate table. Next, you would need to think about garbage collection to remove orphaned entries periodically. Also, this makes values immutable, obviously, since they can be used somewhere else.
- Metadata logged into the same key creates a new record, not removing the old one. **This one I see as critical for consistency.** I suggest you either overwrite the value for the existing key or store values as listed under the same key - design choice, IMO.
- Metadata logged inside steps is not logged on rerun in cached mode. This one is quite questionable if we should do this or not, but this would be a consistent UX in comparison to artifacts, for instance.

LMK, if you want to talk more.

P.S. @htahir1 FYI, this might be relevant for product road mapping and prioritization of tasks, so your input is warmly welcomed.

docs/book/how-to/model-management-metrics/track-metrics-metadata/attach-metadata-to-steps.md
bcdurak: Good catch, added it.

src/zenml/client.py
bcdurak: I also thought about this, and I was not so sure about which path to follow. On the one hand, it makes the code easier to read and follow, but on the other hand, it feels redundant to add a new class just to combine an ID and a type for an internal use case. I will try it out to see how it fits.

bcdurak: I guess, in this case, it is not fully internal as it is a `client` method, which makes it even worse to add a new class IMO.

src/zenml/models/v2/core/run_metadata.py
bcdurak: Closing this one, will follow the discussion above.

bcdurak:
@avishniakov Just had a discussion with @htahir1, here is what we decided to do:

1. Solving the `metadata values duplication` with the proposed structure is out of the scope of this ticket for now. We might solve this in a separate PR in the future.
2. Metadata logged into the same key creates a new record, not removing the old one. I have to discuss this with @znegrin but it is likely that we keep the same behaviour. In this case, I have to make sure that our users get the right value back, may it be a list of values or the latest value for that matter.
3. As for cached steps, this is a very valid concern. I will try to figure out a way to store if the metadata entry was created during a step execution and duplicate the same behavior.

src/zenml/client.py
avishniakov: Well, you can make it a private class method, if it is meant to be only used internally. In this case, it should be fine to keep awkward input parameters, since the "we told you not to use it" thing.

bcdurak:
@avishniakov I made some significant changes after our last discussion:

- Now, each time you call `log_metadata`, only one entity will be linked to the newly generated metadata entry.
- We now differentiate the metadata that is attached to a step during its execution with a `cached` flag. This helps us to create a new link if another step is using the cached execution of the original step and we link the original metadata to the new step. 
- I have added some additional utility classes as you suggested (`RunMetadataResource`) that are used in the process of generating run metadata.
- I have added two additional methods to the affected entity types. `fetch_metadata` returns the value of the latest entry for each key (we use it while building the response models) and `fetch_metadata_collection` returns a list for each key containing entries with a value and a timestamp (we can potentially use this in the future to create histograms, this is just the initial version).

I also updated all the docs pages accordingly.

@schustmi Would be nice if you could read the changes I described above before diving into the PR.

src/zenml/zen_server/routers/workspaces_endpoints.py
```diff
@@ -1009,24 +1009,21 @@ def create_run_metadata(
...
-        )
-        verify_permission_for_model(artifact_version, action=Action.UPDATE)
-    elif run_metadata.resource_type == MetadataResourceTypes.MODEL_VERSION:
-        model_version = zen_store().get_model_version(run_metadata.resource_id)
-        verify_permission_for_model(model_version, action=Action.UPDATE)
-    else:
-        raise RuntimeError(
-            f"Unknown resource type: {run_metadata.resource_type}"
-        )
+    for resource in run_metadata.resources:
```
schustmi, line 1012: There is `batch_verify_permissions_for_model` which does one RBAC server request instead of multiple, which would be helpful here I think

src/zenml/client.py
```diff
@@ -4438,23 +4438,20 @@ def _delete_artifact_from_artifact_store(
...
-                metadata was produced.
-            resource_type: The type of the resource for which the
+            resources: The list of IDs and types of the resources for that the
                 metadata was produced.
             stack_component_id: The ID of the stack component that produced
                 the metadata.
-
-        Returns:
-            None
+            cached: A flag indicating if the run metadata can be cached during
```
schustmi, line 4453: It doesn't feel right to have this as a public option in the client method somehow. If anything, this should be a server-internal flag that is somehow set from the context. I think we talked about it and it is used to differentiate metadata that was logged to a step after its execution, which should not be part of the metadata if we create a cached step from it, correct?

schustmi (changes requested):

docs/book/how-to/model-management-metrics/track-metrics-metadata/attach-metadata-to-a-model.md
```diff
@@ -1,62 +1,102 @@
...
+       log_metadata(
+           metadata={
+               "evaluation_metrics": {
+                   "accuracy": accuracy,
+                   "precision": precision,
+                   "recall": recall
+               }
+           },
+           model_name=step_context.model.name,
+           model_version=step_context.model.version,
```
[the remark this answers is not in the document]
bcdurak: Fixed it.

src/zenml/cli/base.py
```diff
@@ -79,19 +79,20 @@ def copier_github_url(self) -> str:
 ZENML_PROJECT_TEMPLATES = dict(
     e2e_batch=ZenMLProjectTemplateLocation(
         github_url="zenml-io/template-e2e-batch",
-        github_tag="2024.11.20",  # Make sure it is aligned with .github/workflows/update-templates-to-examples.yml
+        # TODO: Update to a newer date
```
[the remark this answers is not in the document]
bcdurak: Removing the TODOs now.

src/zenml/models/v2/core/run_metadata.py
```diff
@@ -13,35 +13,38 @@
...
         title="The metadata to be created.",
     )
     types: Dict[str, "MetadataTypeEnum"] = Field(
         title="The types of the metadata to be created.",
     )
+    cached: Optional[bool] = Field(
+        title="A flag indicating if the run metadata is cached through "
+        "a step execution.",
+        default=False,
+    )
```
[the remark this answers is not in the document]
bcdurak: This is needed to differentiate between the metadata which is linked to a step during its execution with a bare `log_metadata(metadata={})` call and any metadata which is linked to a step manually `log_metadata(metadata={}, step_id="xxx")`. 

Because if a new step is using the cached execution of this step, we want to include metadata coming from the first condition and exclude the metadata related to that second condition.

So if someone runs this pipeline:

```python
from zenml import log_metadata
from zenml import pipeline, step, Model

@step
def step1() -> None:
    log_metadata(metadata={"score": 0.22})


@step
def step2() -> None:
    log_metadata(metadata={"score": 0.65})
    log_metadata(metadata={"score": 0.55})


@pipeline
def metadata_test():
    step1()
    step2(after="step1")


run = metadata_test()
```

Then calls this:

```python
first_run = client.get_pipeline_run(...)
log_metadata(
    metadata={"score": 0.75, "accuracy": "apt"},
    step_id=first_run.steps["step2"].id
)
```

and then runs the same pipeline again and fetches the metadata of the second run (that is cached):

```python
second_run = client.get_pipeline_run(...)
print(run.run_metadata)
```

They should get: `{'step1::score': 0.22, 'step2::score': 0.55}`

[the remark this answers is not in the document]
bcdurak: I will add a test regarding this use-case as well.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8140,6 +8179,45 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
...
+                        RunMetadataResourceSchema.resource_id
+                        == step_run.original_step_run_id
+                    )
+                    .where(
+                        RunMetadataResourceSchema.resource_type
+                        == MetadataResourceTypes.STEP_RUN
+                    )
+                    .where(
+                        RunMetadataSchema.cached == True  # noqa: E712
+                    )
```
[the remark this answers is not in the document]
bcdurak: I explained the use-case down below. I hope it clears your question here as well. Briefly summarized, when we create a `StepRun` in our DB, if it is a cached step, we fetch the `cached` metadata of the original step, so we will link it to the new step as well (excluding any manually generated metadata through the `cached==True` flag).

src/zenml/utils/metadata_utils.py
```diff
@@ -103,6 +107,8 @@ def log_metadata(
     model_version_id: Optional[UUID] = None,
     model_name: Optional[str] = None,
     model_version: Optional[str] = None,
+    # Parameter to adjust whether we log to all related entities
+    log_related_entities: Optional[bool] = True,
```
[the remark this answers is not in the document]
bcdurak: This parameter is completely removed now. Oversight from my side that they were still in the main call.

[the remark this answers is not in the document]
bcdurak: Removed it now.

src/zenml/zen_stores/schemas/model_schemas.py
```diff
@@ -344,6 +348,41 @@ def from_request(
...
+
+        Returns:
+            a dictionary, where the key is the key of the metadata entry
+                and the values represent the latest entry with this key.
+        """
+        metadata_collection = self.fetch_metadata_collection()
+        return {
+            k: sorted(v, key=lambda x: x.created, reverse=True)[0].value
+            for k, v in metadata_collection.items()
+        }
```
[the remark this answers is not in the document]
bcdurak: Working on this now. Resolving the other comments to clean up the board a bit. Will update here once I fix it.

src/zenml/client.py
bcdurak: That's correct regarding the use-case. 

I also don't like the fact that it is a part of our public client, but unfortunately, we can not handle it from the server side at the moment. I will keep this on the backburner but for now I'm resolving this comment.

tests/integration/functional/model/test_model_version.py
```diff
@@ -107,10 +107,18 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+        model_name=model.name,
+        model_version=model.version,
+    )
     assert get_step_context().model.run_metadata["foo"] == "bar"
     log_metadata(
-        metadata={"foo": "bar"}, model_name=mdl_name, model_version="other"
+        metadata={"foo": "bar"},
+        model_name=mdl_name,
+        model_version="other",
     )
```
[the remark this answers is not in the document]
bcdurak: I have added this to the remaining TODOs in the description for now. I would like to add a few more anyhow.

src/zenml/zen_server/routers/workspaces_endpoints.py
bcdurak: Good catch, just fixed this.

docs/book/how-to/model-management-metrics/track-metrics-metadata/attach-metadata-to-an-artifact.md
```diff
@@ -1,68 +1,113 @@
...
+    pd.DataFrame, "processed_data"
+]:
     """Process a dataframe and log metadata about the result."""
-    # Perform processing on the dataframe...
     processed_dataframe = ...
 
     # Log metadata about the processed dataframe
-    log_artifact_metadata(
+    log_metadata(
         artifact_name="processed_data",
```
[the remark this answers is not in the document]
bcdurak: Can you elaborate a bit here?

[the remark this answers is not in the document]
avishniakov: I mean something like this was easy before, but not straightforward with your changes:
```python
@step
def my_step()->int:
    log_artifact_metadata(metadata={...})
    return 42
```

src/zenml/models/v2/core/run_metadata.py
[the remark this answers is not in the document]
avishniakov: I see, in this case, the name looks confusing to me. Maybe `cachable` would do a better job?

src/zenml/zen_stores/schemas/model_schemas.py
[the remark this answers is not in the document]
avishniakov: Isn't it possible just to keep it as simple as following, so you just define an interface and the implementation is in children?
```python
class SomeBaseClass(BaseSchema):

    run_metadata_resources: List["RunMetadataResourceSchema"]

    def fetch_metadata_collection(self) -> Dict[str, List[RunMetadataEntry]]:
        ...

    def fetch_metadata(self) -> Dict[str, MetadataType]:
        ...
```

avishniakov (commented):
I don't have anything new to add to the recent changes, but I replied to the unresolved comments form the past

src/zenml/zen_stores/schemas/model_schemas.py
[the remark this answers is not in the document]
bcdurak: This was unfortunately not possible as it yielded a ValueError: `<class 'list'> has no matching SQLAlchemy type`.  However, changing the line to `run_metadata_resources = Relationship()` helped me overcome the issue, so I have changed the implementation.

docs/book/how-to/model-management-metrics/track-metrics-metadata/attach-metadata-to-an-artifact.md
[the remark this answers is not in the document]
bcdurak: This has changed now. As I mentioned below, I have added two new parameters called `infer_artifact` and `infer_model` which makes it easier to use `log_metadata` within the step execution.

src/zenml/models/v2/core/run_metadata.py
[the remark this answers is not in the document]
bcdurak: I have changed the `cached` attribute with `publisher_step_id`. This is for the future where we can use the step execution id to figure out which models got tagged as well.

src/zenml/utils/metadata_utils.py
```diff
@@ -13,96 +13,118 @@
...
 ) -> None: ...
 
 
 @overload
 def log_metadata(
     *,
     metadata: Dict[str, MetadataType],
-    run_id_name_or_prefix: Union[UUID, str],
+    infer_artifact: bool = False,
+    artifact_name: Optional[str] = None,
```
avishniakov, line 82: Why do those 2 come together here?

```suggestion
    infer_artifact: bool = False,
```

src/zenml/utils/metadata_utils.py
```diff
@@ -13,96 +13,118 @@
...
     step_id: Optional[UUID] = None,
     step_name: Optional[str] = None,
     run_id_name_or_prefix: Optional[Union[UUID, str]] = None,
-    # Parameters to manually log metadata for artifacts
+    # Artifacts
     artifact_version_id: Optional[UUID] = None,
     artifact_name: Optional[str] = None,
     artifact_version: Optional[str] = None,
-    # Parameters to manually log metadata for models
+    infer_artifact: Optional[bool] = None,
```
avishniakov, line 122:
```suggestion
    infer_artifact: bool = False,
```
maybe

src/zenml/utils/metadata_utils.py
```diff
@@ -13,96 +13,118 @@
...
+    # Models
     model_version_id: Optional[UUID] = None,
     model_name: Optional[str] = None,
-    model_version: Optional[str] = None,
+    model_version: Optional[Union[ModelStages, int, str]] = None,
+    infer_model: Optional[bool] = None,
```
avishniakov, line 127:
```suggestion
    infer_model: bool = False,
```
maybe

src/zenml/zen_stores/schemas/utils.py
```diff
@@ -67,3 +70,44 @@ def get_page_from_list(
         total=total,
         items=page_items,
     )
+
+
+class RunMetadataInterface:
```
avishniakov, line 75: 👏🏼

tests/integration/functional/artifacts/test_utils.py
```diff
@@ -215,7 +215,11 @@ def artifact_multi_output_metadata_logging_step() -> (
         "description": "Blupus is great!",
         "metrics": {"accuracy": 0.9},
     }
-    log_metadata(metadata=output_metadata, artifact_name="int_output")
+    log_metadata(
+        metadata=output_metadata,
+        artifact_name="int_output",
+        infer_artifact=True,
```
avishniakov, line 222: Why `infer_artifact=True`? I might be misreading this, but you gave artifact name already, no?

avishniakov (approved):
I left some comments, nothing major, mostly around new flag for artifacts. If you consider this as not critical or not relevant, I will put my approval upfront to prevent PR being blocked from merging.

src/zenml/utils/metadata_utils.py
bcdurak: You can potentially have a step with multiple outputs, you name one of them through the `artifact_name` and expect the `infer_artifact` to look for it in the step execution.

tests/integration/functional/artifacts/test_utils.py
bcdurak: I also discussed with @schustmi. Eventually, we do **not** want to enable people to provide just a name to `log_metadata` for the latest version (with models or artifacts). So, if you provide just a name, it needs to be paired with `infer_artifact=True`, so we look for it in the output space.

src/zenml/utils/metadata_utils.py
bcdurak: changed it.

src/zenml/model/utils.py
```diff
@@ -52,29 +52,31 @@ def log_model_metadata(
...
-                "Model name and version must be provided unless the function is "
-                "called inside a step with configured `model` in decorator."
-            )
-        mv = step_context.model
-
-    mv.log_metadata(metadata)
+        raise ValueError(
+            "You can call `log_model_metadata` by either providing both "
+            "`model_name` and `model_version` or keeping both of them None."
+        )
```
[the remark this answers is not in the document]
bcdurak: Yes, if the user does not specify the name and the version, `log_metadata` will infer it.

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -222,10 +228,7 @@ def to_model(
             RuntimeError: If the step run schema does not have a deployment_id
                 or a step_configuration.
         """
-        run_metadata = {
-            metadata_schema.key: json.loads(metadata_schema.value)
-            for metadata_schema in self.run_metadata
-        }
+        run_metadata = self.fetch_metadata()
```
schustmi, line 231: Can we fetch this inside the `if`, this is only needed when `include_metadata` is True>

schustmi: Same for other the pipeline run

src/zenml/zen_stores/schemas/run_metadata_schemas.py
```diff
@@ -103,3 +78,63 @@ class RunMetadataSchema(BaseSchema, table=True):
...
+        target_column="id",
+        ondelete="CASCADE",
+        nullable=False,
+    )
+
+    # Relationship back to the base metadata table
+    run_metadata: RunMetadataSchema = Relationship(back_populates="resources")
+
+    # Relationship to link specific resource types
+    pipeline_run: List["PipelineRunSchema"] = Relationship(
```
schustmi, line 113:
```suggestion
    pipeline_runs: List["PipelineRunSchema"] = Relationship(
```

Same for all the others

src/zenml/zen_stores/migrations/versions/cc269488e5a9_separate_run_metadata.py
```diff
@@ -0,0 +1,140 @@
...
+                ),  # Changed to BIGINT
+            ),
+            resource_data,
+        )
+
+    # Drop the old `resource_id` and `resource_type` columns from `run_metadata`
+    op.drop_column("run_metadata", "resource_id")
+    op.drop_column("run_metadata", "resource_type")
+
+    # Add the cached column to the database table
```
schustmi, line 89:
```suggestion
```

schustmi: Wrong, and doesn't provide any info that isn't in the line of code below

src/zenml/zen_stores/migrations/versions/cc269488e5a9_separate_run_metadata.py
```diff
@@ -0,0 +1,140 @@
...
+                sa.Column(
+                    "run_metadata_id",
+                    sqlmodel.sql.sqltypes.GUID(),
+                    nullable=False,
+                ),  # Changed to BIGINT
+            ),
+            resource_data,
+        )
+
+    # Drop the old `resource_id` and `resource_type` columns from `run_metadata`
```
schustmi, line 85:
```suggestion
```

src/zenml/zen_stores/migrations/versions/cc269488e5a9_separate_run_metadata.py
```diff
@@ -0,0 +1,140 @@
...
+# revision identifiers, used by Alembic.
+revision = "cc269488e5a9"
+down_revision = "ec6307720f92"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Creates the 'run_metadata_resource' table and migrates data."""
+    # Create the `run_metadata_resource` table
```
schustmi, line 23:
```suggestion
```

src/zenml/zen_stores/migrations/versions/cc269488e5a9_separate_run_metadata.py
```diff
@@ -0,0 +1,140 @@
...
+        sa.Column("resource_type", sa.String(length=255), nullable=False),
+        sa.Column(
+            "run_metadata_id",
+            sqlmodel.sql.sqltypes.GUID(),
+            sa.ForeignKey("run_metadata.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+    )
+
+    # Migrate existing data from `run_metadata` to `run_metadata_resource`
```
schustmi, line 42:
```suggestion
```

src/zenml/zen_stores/migrations/versions/cc269488e5a9_separate_run_metadata.py
```diff
@@ -0,0 +1,140 @@
...
+    connection = op.get_bind()
+
+    # Fetch existing `run_metadata` data
```
schustmi, line 45:
```suggestion
```

src/zenml/utils/metadata_utils.py
```diff
@@ -114,150 +136,152 @@ def log_metadata(
...
+                "If you are using the `infer_model` option, the function must "
+                "be called inside a step with configured `model` in decorator."
+                "Otherwise, you can provide a `model_version_id` or a "
+                "combination of `model_name` and `model_version`."
             )
-        else:
-            step_context = None
-            with contextlib.suppress(RuntimeError):
-                step_context = get_step_context()
+        mv = step_context.model
```
schustmi, line 224: I think using `step_context.model_version` would be better here, you can log to the ID directly

src/zenml/utils/metadata_utils.py
```diff
@@ -114,150 +136,152 @@ def log_metadata(
...
+                "`log_metadata`, it must be called inside a step with outputs."
+                "Otherwise, you can provide a `artifact_version_id` or a "
+                "combination of `artifact_name` and `artifact_version`."
+            )
+
+        step_output_names = list(step_context._outputs.keys())
+
+        if artifact_name is not None:
+            # If a name provided, ensure it is in the outputs
+            assert artifact_name in step_output_names, (
```
schustmi, line 270: This should be a ValueError, assert is not something that should be used to verify conditions

src/zenml/utils/metadata_utils.py
```diff
@@ -13,96 +13,118 @@
...
+@overload
 def log_metadata(
+    *,
     metadata: Dict[str, MetadataType],
-    # Parameters to manually log metadata for steps and runs
+    infer_model: bool = False,
+) -> None: ...
+
+
+def log_metadata(
```
schustmi, line 112: I think the whole function could be a lot shorter if we just store the `RunMetadataResource.id` and `RunMetadataResource.type` and do one client call in the end?

schustmi (changes requested):

src/zenml/zen_stores/schemas/step_run_schemas.py
bcdurak: Good catch! Fixed this.

src/zenml/utils/metadata_utils.py
bcdurak: Changed it now.

src/zenml/utils/metadata_utils.py
bcdurak: Switched to it now.

src/zenml/utils/metadata_utils.py
bcdurak: Makes sense. I trimmed it a bit already but kept the `resources` parameter as a list and would rather not remove it right now. I am still a bit in between if we will ever tag multiple resources at the same time and if it is alright with you, I would keep it like this.

PR #3211, opened 2024-11-21
Various fixes and improvements

stefannica:
## Describe changes
This is a bundle of bug fixes and improvements that were collected during the load tests I have been performing on the ZenML server.

1. adding a new authorized API endpoint, `GET /ap1/v1/load-info`, that can be called to fetch some rudimentary information about the current state of the server: how many threads are currently running and how many database connections are established, used or in overflow, in the following format:

```
{
  "threads": 6,
  "db_connections_total": 13,
  "db_connections_active": 7,
  "db_connections_overflow": 5
}
```

2. some Kubernetes orchestrator improvements:

  * using a role binding instead of a cluster role binding when creating the service account used to run pipelines, because a role binding is less intrusive (only grants permissions in one namespace), whereas the cluster role grants permissions across the entire cluster
  * setting some _default_ CPU and memory resource requests for the orchestrator and step pods. It is recommended to have memory resource requests defined for pods, otherwise the Kubernetes scheduler can place them all on the same node where they can blow up and affect other pods or even take down the entire node itself (which is what I observed during my tests). 400 MiB is the minimum memory use for these pods.

3. catching `urllib3.HTTPError` when verifying the Kubernetes service connector credentials helps report better errors to the client in case of connection errors (e.g. missing or invalid certificates)

4. set `_warn_on_response_updates` to False on the `ServiceConnectorResponse` removes some pesky and very verbose warnings that show up in the server logs because the `type` field is intentionally overwritten in transit.

5. makes some Helm chart fixes and improvements

  * setting the `zenml.database.poolSize` and `zenml.database.maxOverflow` values is currently breaking the deployment, because these values are integer and not processed properly in the Helm chart secret definition
  * with this opportunity, I created a special template section for all `ZENML_STORE_*` environment variables instead of setting them individually all over the place, just like we do with server settings and secret store settings.

6. updated all our middlewares to be better behaved and catch all exceptions that the next middleware called with `call_next` might throw and return a response, instead of letting the exception slip. This is a recommendation that came up during load testing, as some of these errors were polluting the logs without revealing the real source of the problem.

7. this PR contains some fixes and improvements for the database backup/restore functionality

* when extracting table information from the database during the backup, ordering rows by anything other than their index is **extremely** slow with larger tables. E.g. fetching 50 pipeline metadata rows sorted by `created` date takes 50 seconds when that table has around 80k rows. For this reason, we always sort by the index columns (with the exception below)
* similarly, for self-referencing tables (step runs), during backup we had to sort the rows using their `created` timestamp, to avoid breaking the foreign key constraint later on when the database dump is restored. This is now replaced with a different strategy: we detect which tables have self-references and then we simply disable the constraint checks during the database restore operation only for these tables.
* increased batch size from 50 rows to 100 rows to speed up the backup

## Pre-requisites

## Types of changes


src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
```diff
@@ -300,13 +300,15 @@ def create_edit_service_account(
         cluster_role_binding_name: Name of the cluster role binding.
             Defaults to "zenml-edit".
     """
-    crb_manifest = build_cluster_role_binding_manifest_for_service_account(
+    crb_manifest = build_role_binding_manifest_for_service_account(
```
schustmi, line 303: Is this now a `rb_manifest` if it's not a cluster role binding anymore? Also the `cluster_role_binding_name` seems wrong?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -422,6 +423,37 @@ def prepare_or_run_pipeline(
             )
             return
 
+        # We set some default minimum resource requests for the orchestrator pod
```
schustmi, line 426: Can we extract that maybe into a method on the `KubernetesPodSettings`? Like `apply_default_requests()`, in order to avoid having this code around multiple times?

src/zenml/zen_server/routers/server_endpoints.py
```diff
@@ -71,6 +73,51 @@ def server_info() -> ServerModel:
...
+            db_connections_active=0,
+            db_connections_overflow=0,
+        )
+
+    from sqlalchemy.pool import QueuePool
+
+    # Get the number of active connections
+    pool = store.engine.pool
+    assert isinstance(pool, QueuePool)
+    idle_conn = pool.checkedin()
```
schustmi, line 108: I'm still not really sure whether this is actually computing what we think it is:

I'm going to paste the QueuePool Implementation here again:

```python
def checkedin(self) -> int:
        return self._pool.qsize()

    def overflow(self) -> int:
        return self._overflow if self._pool.maxsize else 0

    def checkedout(self) -> int:
        return self._pool.maxsize - self._pool.qsize() + self._overflow
```

The total connections right now are computed as follows:
```
con_in_pool + (pool_max_size - con_in_pool + overflow) + max(0, overflow)
= pool_max_size + overflow + max(0, overflow)
```

Is that really correct though?

`overflow` starts at `-pool_max_size` and gets incremented each time a connection is opened, and decreased each time a connection is closed (at least that's my understanding of it).

I think we need to consider two cases:
Case 1: The amount of connections are less or equal connections than `max_pool_size` -> `overflow <=0`
In this case, I think the computation for the total connections is fine.

Case 2: There are more connections than the pool size:
In this case, the total connections are `pool_max_size + overflow + max(0, overflow)` which just adds the overflow twice.

schustmi (changes requested):

src/zenml/zen_server/routers/server_endpoints.py
stefannica: that might explain why I've seen bigger connection numbers when overflow is reached

schustmi: Maybe `total` is just `checkedin() + checkedout()`?

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
stefannica: I did another round of renaming. Should be good now.

src/zenml/zen_server/routers/server_endpoints.py
stefannica: yes, quite right

PR #3245, opened 2024-12-05
Fix request model validation

schustmi:
## Describe changes
This PR contains two changes:
- Our `make_dependable` function that allows us to use filter models attributes as query parameters in FastAPI endpoints had a bug where it would throw a `500` HTTP error instead of the intended `422`, because there were some objects that are not JSON serializable in the `ValidationError.errors()` list that we passed on to the `HTTPError`. This PR fixes that by instead using our helper function to compute the error detail.
- When an invalid UUID was being passed for a `equals:<ID>` filter, we threw a ValueError. This did not match the behaviour when using e.g. a `oneof:[<ID>]` filter and leads to additional complexity in the frontend, which is why we removed this error. The filter with an invalid ID will now simply return an empty list.

## Pre-requisites

## Types of changes


bcdurak (commented):
> When an invalid UUID was being passed for a equals:<ID> filter, we threw a ValueError. This did not match the behaviour when using e.g. a oneof:[<ID>] filter and leads to additional complexity in the frontend, which is why we removed this error. The filter with an invalid ID will now simply return an empty list.

I would propose approaching this the other way around. If we throw an error when people use an invalid UUID on a field which is clearly UUID, it is much more helpful to the user (and to the frontend) IMO. So, instead of removing the check from `equals:`, I would propose adding it to `oneof:` as well and keep things consistent.

schustmi:
> > When an invalid UUID was being passed for a equals: filter, we threw a ValueError. This did not match the behaviour when using e.g. a oneof:[] filter and leads to additional complexity in the frontend, which is why we removed this error. The filter with an invalid ID will now simply return an empty list.
> 
> I would propose approaching this the other way around. If we throw an error when people use an invalid UUID on a field which is clearly UUID, it is much more helpful to the user (and to the frontend) IMO. So, instead of removing the check from `equals:`, I would propose adding it to `oneof:` as well and keep things consistent.

I could make the same argument for date filters before the year 2020 though, as ZenML didn't exist yet and therefore no pipeline run can be created before that. Or any name that contains invalid characters. Or a negative integer for a version number.

Especially for the filter implementation in the frontend, it's very weird if the user switches the operator from `contains` to `equals` and suddenly gets an error message coming from the backend I feel. I don't really see the need to have this inconsistency just for a single operator and type, maybe @Cahllagerfeld also some opinion here

bcdurak:
> > > When an invalid UUID was being passed for a equals: filter, we threw a ValueError. This did not match the behaviour when using e.g. a oneof:[] filter and leads to additional complexity in the frontend, which is why we removed this error. The filter with an invalid ID will now simply return an empty list.
> > 
> > 
> > I would propose approaching this the other way around. If we throw an error when people use an invalid UUID on a field which is clearly UUID, it is much more helpful to the user (and to the frontend) IMO. So, instead of removing the check from `equals:`, I would propose adding it to `oneof:` as well and keep things consistent.
> 
> I could make the same argument for date filters before the year 2020 though, as ZenML didn't exist yet and therefore no pipeline run can be created before that. Or any name that contains invalid characters. Or a negative integer for a version number.
> 
> Especially for the filter implementation in the frontend, it's very weird if the user switches the operator from `contains` to `equals` and suddenly gets an error message coming from the backend I feel. I don't really see the need to have this inconsistency just for a single operator and type, maybe @Cahllagerfeld also some opinion here

I see your point. However, I have to say, throwing an error if someone enters a name with an invalid character made sense to me intuitively. 😄 I guess, the main use case I was trying to avoid was this: Back in the day (not sure if this is still the case), some of our UUID filters would work with the `XXXX-XXXX-XXXXX-XXX` format whereas some of them would work with the `XXXXXXXXXXX` format. In cases like this, if I get empty results back, as a user, it would not be clear to me if I entered something wrong or if the results were just empty. That's why checking the format (without checking the content like the name or the date) seemed like a good idea.

I agree with you completely regarding the switch though from the frontend perspective. Maybe @stefannica and @AlexejPenner can chime in here as well. Regardless, I don't think this is a critical issue. I will approve for now, feel free to merge as you see fit.

bcdurak (approved):

PR #3254, opened 2024-12-09
Create model versions server-side to avoid race conditions

schustmi:
## Describe changes
If multiple steps ran in parallel, having the same `Model` specified for either multiple steps or the pipeline could create multiple numeric versions for that model, even though only one was intended. This happened because the model version creation was happening on the client side and there were no unique constraints in place to make sure this does not happen. This PR aims to fix this by creating the model versions on the server as part of the pipeline run or step run creation.

Some other small changes included in this PR:
- The `number` attribute from the model version request was removed. This should have never been an attribute on the request model, as the client is not allowed to set it anyway.
- Some attributes from the deployment schema are only accessed when they should be included in the response.

## Pre-requisites

## Types of changes


src/zenml/models/v2/core/model_version.py
```diff
@@ -77,10 +77,6 @@ class ModelVersionRequest(WorkspaceScopedRequest):
         default=None,
     )
 
-    number: Optional[int] = Field(
```
AlexejPenner, line 80: good catch

AlexejPenner (approved):
🦭

src/zenml/zen_stores/schemas/model_schemas.py
```diff
@@ -328,12 +374,17 @@ class ModelVersionSchema(NamedSchema, RunMetadataInterface, table=True):
 
     @classmethod
     def from_request(
-        cls, model_version_request: ModelVersionRequest
+        cls,
+        model_version_request: ModelVersionRequest,
+        model_version_number: int,
+        producer_run_id: Optional[UUID] = None,
```
bcdurak, line 380: Just leaving a note for the future (I do realize that we can not fix this with this PR): I can see why we need these two parameters here, but I am not a big fan of adding them here. The `from_request` is not defined as an `@abstractmethod` in our base schemas, so mypy is not complaining, but the conversion from a `Request` model to a `Schema` sometimes occurs fully in this method, sometimes fully in the `SQLZenStore`, sometimes partially in both places. I feel like this is just bad practice and makes it hard to read/fix the code. We can improve it in the future.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -10299,73 +10348,296 @@ def _get_next_numeric_version_for_model(
...
+            if model_version.name.isnumeric():
+                raise ValueError(
+                    "Can't create model version with custom numeric model "
+                    "version name."
+                )
 
-        has_custom_name = model_version.name is not None
-        if has_custom_name:
-            validate_name(model_version)
+            if str(model_version.name).lower() in ModelStages.values():
```
bcdurak, line 10614: Could this become potentially problematic if the `ModelStages` evolve into a wider selection of stages and there are already models with these entries in the DB?

bcdurak: Actually, if this check did not happen before, we might even run into this issue now.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -10299,73 +10348,296 @@ def _get_next_numeric_version_for_model(
...
+        else:
+            producer_run_id = pipeline_or_step_run.pipeline_run_id
+            step_run = pipeline_or_step_run.to_model(include_metadata=True)
+            configured_model = step_run.config.model
+            substitutions = step_run.config.substitutions
+
+        if not configured_model:
+            return None
+
+        model_request = ModelRequest(
```
bcdurak, line 10528: Just a small nitpick but I feel like this is a bit of an unusual interaction. Using a `Request` model to `create` something seems completely fine but when the `get` operation comes into the mix, it feels a bit odd to use the same `Request` models that we have used for the creation.

bcdurak (approved):
Thanks for the explanations before, it made the review process much easier. Leaving a few comments now, otherwise, LGTM.

src/zenml/zen_stores/sql_zen_store.py
schustmi: The check happened, it was just on the client-side and I now moved lots of the logic server-side. So if someone was trying to actively abuse it they could have such a name in their DB, but even then it doesn't really cause any problems. They just won't be able to fetch that model version by name, because we will always fetch it by stage instead.

src/zenml/zen_stores/schemas/model_schemas.py
schustmi: Yes 100%, the whole `from_request` methods I'm not a huge fan of I have to say.

PR #3262, opened 2024-12-12
Include user of latest run in pipeline response

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/models/v2/core/pipeline.py
+++ b/src/zenml/models/v2/core/pipeline.py
@@ -45,7 +45,6 @@ from zenml.models.v2.core.tag import TagResponse
 
 if TYPE_CHECKING:
     from zenml.models import PipelineRunResponse, UserResponse
-    from zenml.models.v2.core.pipeline_run import PipelineRunResponse
     from zenml.zen_stores.schemas import BaseSchema
 
     AnySchema = TypeVar("AnySchema", bound=BaseSchema)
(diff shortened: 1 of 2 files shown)


schustmi:
@bcdurak I have a question regarding the response here: Currently our `PipelineResponseBody` already contains the `latest_run_id` and `latest_run_status`, and this PR adds a `latest_run_user` to the resources.

In general, I would say this is all wrong and there should just be a `latest_run` in the resources and all the other fields should be removed. The only reason I did not implement it as part of this PR is that the `PipelineRunResponseBody` is huuuuge. It contains lots of other entities like the stack, build, code reference etc. All of this should ideally be moved into the resources of the run.

Now to my question: When do we do this? Should I do it as part of this PR? Or should I just put the `latest_run` into resources and we don't care about the responses getting a little bigger?

bcdurak:
@schustmi, I agree with your general take. I don't like the direction this is heading and this is definitely not the only model being affected. We have quite a few response models that refer to other responses in this partial fashion. 

Unfortunately, this is not something we can handle quickly as there are other interfaces like the dashboard that would have to change as well. Maybe, it is a good start to do the changes for the pipeline runs (if you have the bandwidth) and we can see how it goes from there.

bcdurak (approved):
Feel free to merge as is if you decide to.

PR #3265, opened 2024-12-15
Replace deprecated `datetime.utcnow()` with `datetime.now(timezone.utc)` 

aiakide:
## Describe changes
Since Pyhton 3.12, `datetime.utcnow()` is deprecated and causes a DeprecationWarning. 

According to the [Python Standard Library documentation](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcnow), it is recommended to replace `datetime.utcnow()` with `datetime.now(timezone.utc)`. 

## Pre-requisites
  - [X] [ZenML Docs](https://docs.zenml.io)

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8628,10 +8628,23 @@ def _update_pipeline_run_status(
...
                 ):
+                    # We need to ensure both datetimes are timezone-aware to avoid TypeError when subtracting
+                    # Now calculate the duration time
+                    if pipeline_run.start_time.tzinfo is None:
+                        pipeline_run.start_time = (
+                            pipeline_run.start_time.replace(
+                                tzinfo=timezone.utc
+                            )
+                        )
+                    if run_update.end_time.tzinfo is None:
```
schustmi, line 8643: Can this really happen, we just set the `run_update.end_time` a few lines above for a specific timezone?

aiakide: No, you're right. That can't happen. I hadn't seen that the `run_update_end_time` was set at this point.
I've adapted it.

schustmi (approved):

schustmi:
@aiakide Actually I have one found one more place that might be problematic if we use timezone-aware datetimes:
In `src/zenml/stack_deployments/stack_deployment.py::get_stack`, the `date_start` gets passed to filter the stacks, which fails if there is a timezone in that date.

Could you add the following code there as part of this PR?
```python
if date_start and date_start.tzinfo:
  date_start = date_start.astimezone(timezone.utc).replace(tzinfo=None)
```

This makes sure we convert the date to UTC first, and then remove the tz information so that the filtering does not fail.

aiakide:
@schustmi I've updated it. Thanks for the hint. :) 
By the way, is there any way to install all the integrations needed for the test? When I run the tests locally, the tests for the integrations fail (missing modules). The same for the mypy tests. 

I could not solve the problem with the CONTRIBUTING.md.

schustmi:
> @schustmi I've updated it. Thanks for the hint. :) By the way, is there any way to install all the integrations needed for the test? When I run the tests locally, the tests for the integrations fail (missing modules). The same for the mypy tests.
> 
> I could not solve the problem with the CONTRIBUTING.md.

@aiakide Thanks! I think running `bash scripts/install-zenml-dev.sh --system --integrations yes` should install all integrations necessary for the tests

PR #3266, opened 2024-12-16
Fixed and improved sorting

bcdurak:
## Describe changes
- Fixed sorting for MySQL
- Added tests
- Added sorting by latest version for models and artifacts
- Fixed sorting by tags

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/models/v2/base/scoped.py
+++ b/src/zenml/models/v2/base/scoped.py
@@ -466,6 +466,11 @@ class WorkspaceScopedFilter(UserScopedFilter):
 class WorkspaceScopedTaggableFilter(WorkspaceScopedFilter):
     """Model to enable advanced scoping with workspace and tagging."""
 
+    FILTER_EXCLUDE_FIELDS: ClassVar[List[str]] = [
+        *WorkspaceScopedFilter.FILTER_EXCLUDE_FIELDS,
+        "tag",
+    ]
+
     tag: Optional[str] = Field(
         description="Tag to apply to the filter query.", default=None
     )
--- a/src/zenml/models/v2/base/scoped.py
+++ b/src/zenml/models/v2/base/scoped.py
@@ -245,6 +245,8 @@ class UserScopedFilter(BaseFilter):
                 UserSchema, getattr(table, "user_id") == UserSchema.id
             )
 
+            query = query.add_columns(UserSchema.name)
+
             if operand == SorterOps.ASCENDING:
                 query = query.order_by(asc(column))
             else:
@@ -448,6 +450,8 @@ class WorkspaceScopedFilter(UserScopedFilter):
                 WorkspaceSchema,
                 getattr(table, "workspace_id") == WorkspaceSchema.id,
             )
+
+            query = query.add_columns(WorkspaceSchema.name)
 
             if operand == SorterOps.ASCENDING:
                 query = query.order_by(asc(column))
--- a/src/zenml/models/v2/base/scoped.py
+++ b/src/zenml/models/v2/base/scoped.py
@@ -466,16 +466,6 @@ class WorkspaceScopedTaggableFilter(WorkspaceScopedFilter):
         description="Tag to apply to the filter query.", default=None
     )
 
-    FILTER_EXCLUDE_FIELDS: ClassVar[List[str]] = [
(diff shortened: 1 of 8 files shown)


src/zenml/models/v2/base/scoped.py
```diff
@@ -245,6 +245,8 @@ def apply_sorting(
                 UserSchema, getattr(table, "user_id") == UserSchema.id
             )
 
+            query = query.add_columns(UserSchema.name)
```
schustmi, line 248: Can you explain why this is necessary?

From what I read, this essentially means the query is now a
`select(original_query.output, user_name)` query? This would be very confusing IMO, that people using the filter model now suddenly get extra items in their query results

src/zenml/models/v2/base/scoped.py
```diff
@@ -524,79 +523,3 @@ def get_custom_filters(
             )
 
         return custom_filters
-
-    def apply_sorting(
```
schustmi, line 528: Why was the tag sorting removed now?

bcdurak: Unfortunately, the tests showed that the concatenation of the tags alphabetically did not work as intended and tags were sorted randomly. Until we could figure out a solution, I have decided to disable sorting by tags.

src/zenml/models/v2/base/scoped.py
bcdurak: You may have seen this pattern in different places in this PR. This is needed because eventually we need to order by this column and if we don't add this here, the entire thing fails with an `SQLOperationalError`, simply due to the fact that it can not find the column in the initial selection.

I realize this might be confusing at first glance, but as you can see from the code, I did not have to make any adjustments to the Schema -> Response conversion for the corresponding Schemas. This means that our developers won't have to manage it and also, since this is entirely isolated to our internal code, our users will not be affected by it either.

src/zenml/models/v2/base/scoped.py
schustmi: I think the time to figure out a solution is now, and as part of this PR :P

The frontend already has some many version flags to enable/disable features, and the idea behind implementing the sorting was to enable sorting by **all** columns that are displayed in the frontend tables. If we now leave out one column, this would be even more complexity in the frontend

src/zenml/models/v2/base/scoped.py
schustmi: Yep but I really don't get why, and I don't think we should rely on "this for some reason currently works so let's hope for the best" in our DB code.

```python
from sqlmodel import SQLModel, Field, create_engine, Session
from sqlalchemy import Column, select, ForeignKey


class TableA(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    a: str = Field(nullable=False)
    __tablename__ = "table_a"


class TableB(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    fk: int = Field(
        sa_column=Column(
            ForeignKey(
                f"table_a.id",
                name="fk",
            ),
            nullable=False,
        ),
    )
    b: int = Field(nullable=False)


engine = create_engine("sqlite:///:memory:")
SQLModel.metadata.create_all(engine)


with Session(engine) as session:
    a = TableA(a="1")
    session.add(a)
    session.commit()

    b1 = TableB(b=2, fk=a.id)
    session.add(b1)
    session.commit()

    b2 = TableB(b=3, fk=a.id)
    session.add(b2)
    session.commit()


with Session(engine) as session:
    query = select(TableB).distinct()
    query = query.join(TableA)
    query = query.add_columns(TableA.a).order_by(TableA.a)
    query = query.limit(5).offset(0)
    print(query)
    result = session.exec(query).all()

    print(result)
```

This is a simple example to show the behaviour of this `add_columns(...)` method, and if I run this I get the added columns in the result.

schustmi (changes requested):
Replied to the open comments, I think there is still some things we should finish/figure out before merging this.

src/zenml/models/v2/base/scoped.py
bcdurak: I fixed this one. Now, it looks a bit more complicated as I had to sort the tags before, but I am hopeful the new tests with the tag sorting will pass now.

schustmi (approved):

PR #3270, opened 2024-12-20
Fix the onboarding state to account for zenml login

stefannica:
## Describe changes
We used to rely on the "authorized device created" event to mark the "zenml connect" step in the onboarding state.
With "zenml login" taking the place of "zenml connect" and the new login experience, this is still valid for OSS servers, but the ZenML Pro servers no longer use "authorized devices" at the zenml server level, they use "authorized devices" at the ZenML Pro control plane level.

This PR fixes this by introducing:

* a custom `User-Agent` header value set for all the API requests made by _the ZenML Python client_ (to be able to differentiate ZenML client calls from other type of calls)
* a new probe that, when triggered, will update the "zenml login" onboarding state: whenever _a ZenML Python client_ successfully logs in to the server using the "external" authentication scheme

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/auth.py
+++ b/src/zenml/zen_server/auth.py
@@ -37,7 +37,12 @@ from zenml.constants import (
     LOGIN,
     VERSION_1,
 )
-from zenml.enums import AuthScheme, ExecutionStatus, OAuthDeviceStatus
+from zenml.enums import (
+    AuthScheme,
+    ExecutionStatus,
+    OAuthDeviceStatus,
+    OnboardingStep,
+)
 from zenml.exceptions import (
     AuthorizationException,
     CredentialsNotValid,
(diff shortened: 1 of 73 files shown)


src/zenml/zen_server/auth.py
```diff
@@ -761,6 +769,18 @@ def authenticate_external_user(external_access_token: str) -> AuthContext:
             )
             context.alias(user_id=external_user.id, previous_id=user.id)
 
+    # This is the best spot to update the onboarding state to mark the
+    # "zenml login" step as completed for ZenML Pro servers, because the
+    # user has just successfully logged in. However, we need to differentiate
+    # between web clients (i.e. the dashboard) and CLI clients (i.e. the
+    # zenml CLI).
+    user_agent = request.headers.get("User-Agent", "").lower()
+    logger.info(f"User agent: {user_agent}")
```
schustmi, line 778: Should this be a debug log instead?

schustmi (approved):

src/zenml/zen_server/auth.py
stefannica: It was merely a left-over debug log. Removed now.

PR #3271, opened 2024-12-20
Create Sagemaker pipeline schedules if specified

htahir1:
## Describe changes
Add schedule support for sagemaker. Tested manually.

## Pre-requisites

## Types of changes


src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -603,19 +793,31 @@ def compute_metadata(
...
+                {
+                    "schedule_rule_name": execution["rule_name"],
+                    "schedule_type": execution["schedule_type"],
+                    "schedule_expression": execution["schedule_expr"],
+                    "pipeline_name": execution["pipeline_name"],
+                }
+            )
+
+            if next_execution := execution.get("next_execution"):
+                metadata["next_execution_time"] = next_execution.isoformat()
```
htahir1, line 808: Question to @schustmi and @bcdurak , i made this so that some metadata is captured in the schedule case, but not all. Can you help me fix this? Its probably flawed logic as it doesn't show up on the dashboard at all

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -479,38 +475,232 @@ def prepare_or_run_pipeline(
...
+                    f"Missing required field for IAM policy creation: {e}. "
+                    f"Please ensure your execution role has sufficient permissions "
+                    f"to start pipeline executions."
+                )
+
+            # Create the EventBridge rule
+            events_client = session.boto_session.client("events")
+            rule_name = f"zenml-{deployment.pipeline_configuration.name}"
+
+            # Determine first execution time based on schedule type
```
schustmi, line 564: The same code is duplicated above

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -479,38 +475,232 @@ def prepare_or_run_pipeline(
...
+                # Convert local time to UTC for EventBridge
+                dt = deployment.schedule.run_once_start_time.astimezone(
+                    timezone.utc
+                )
+                schedule_expr = f"cron({dt.minute} {dt.hour} {dt.day} {dt.month} ? {dt.year})"
+                next_execution = deployment.schedule.run_once_start_time
+
+            # Create IAM policy for EventBridge
+            iam_client = session.boto_session.client("iam")
+            role_name = self.config.execution_role.split("/")[
```
schustmi, line 521: Isn't the execution role the one that should be assumed by the pipeline during execution, and not used for something around the pipeline execution like this scheduling? This seems more like it should be tied to the role of the service connector if applicable or solved in some other way.

docs/book/component-guide/orchestrators/sagemaker.md
```diff
@@ -339,4 +335,119 @@ This approach allows for more granular tagging, giving you flexibility in how yo
...
+    pass
+```
+
+When you deploy a scheduled pipeline, ZenML will:
+1. Create an EventBridge rule with the specified schedule
+2. Configure the necessary IAM permissions
+3. Set up the SageMaker pipeline as the target
+
+{% hint style="info" %}
+If you run the same pipeline with a schedule multiple times, the existing schedule will be updated with the new settings rather than creating a new schedule. This allows you to modify schedules by simply running the pipeline again with new schedule parameters.
```
schustmi, line 375: I would not do this to keep it consistent with all other orchestrators.

schustmi (changes requested):

schustmi:
@htahir1 A sagemaker pipeline has a `put_triggers(...)` method which internally already creates schedules AFAICS. Maybe something we should use instead of implementing it ourselves?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
htahir1: @schustmi ping on this

htahir1: @bcdurak

schustmi: This `compute_metadata` method gets called from `prepare_or_run_pipeline` and yielded as a result of it. This however only gets applied when there is already is a run (-> non-scheduled case), as otherwise there is no run to attach the metadata to.

For the scheduled case, there is `SagemakerOrchestrator.get_pipeline_run_metadata` which gets called at execution time. Ideally, you'd be able to return the metadata from this object.

Another change IMO would be to allow attaching metadata to schedules, which is actually where most of that metadata fits best IMO. However I would suggest doing that as part of the native schedule implementation, as right now it wouldn't lead to any visibility of that metadata anyway.

docs/book/component-guide/orchestrators/sagemaker.md
```diff
@@ -339,4 +335,138 @@ This approach allows for more granular tagging, giving you flexibility in how yo
...
+    },
+    {
+      "Effect": "Allow",
+      "Action": "iam:PassRole",
+      "Resource": "arn:aws:iam::*:role/*",
+      "Condition": {
+        "StringLike": {
+          "iam:PassedToService": "scheduler.amazonaws.com"
+        }
+      }
```
stefannica, line 432: The "client role" - the role used to create the pipeline and scheduler - is the one that needs to be able to "pass" the role to the scheduler role.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -479,38 +475,175 @@ def prepare_or_run_pipeline(
...
+            if self.config.scheduler_role is None:
+                logger.info(
+                    "No scheduler_role configured. Using service connector "
+                    "role to schedule pipeline."
+                )
+                sts = session.boto_session.client("sts")
+                try:
+                    service_connector_role_arn = sts.get_caller_identity()[
+                        "Arn"
+                    ]
```
stefannica, line 552: keep in mind that this doesn't necessarily comes from a service connector. The error messages should reflect that.

bcdurak: Fixing this as well.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -479,38 +475,175 @@ def prepare_or_run_pipeline(
...
+                    elif ":assumed-role/" in service_connector_role_arn:
+                        # Convert assumed-role ARN format to role ARN format
+                        # From: arn:aws:sts::123456789012:assumed-role/role-name/session-name
+                        # To: arn:aws:iam::123456789012:role/role-name
+                        service_connector_role_arn = re.sub(
+                            r"arn:aws:sts::(\d+):assumed-role/([^/]+)/.*",
+                            r"arn:aws:iam::\1:role/\2",
+                            service_connector_role_arn,
+                        )
+                except Exception:
```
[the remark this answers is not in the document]
bcdurak: Added this now.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: @schustmi I have done most of the things that you described here. I have added the functionality of generating the pipeline run metadata through the `SagemakerOrchestrator.get_pipeline_run_metadata`. That means, we are able to get "almost" the same metadata for the scheduled pipeline runs. The only exception is the Studio URL, because the executor role does not have access to list domains and fetch their IDs. (Btw, I believe we can clean this up even further now, but I would rather handle this in a separate PR.)

As for schedules, I have added the option to attach metadata to them. Much like how we handle the step metadata within the context of pipeline run metadata, I am using the schedule within the deployment to display it on the pipeline run page. It is not a very generalized solution at the moment, but it works well with the Sagemaker case.

bcdurak:
@htahir1 @schustmi Just for your info, the quickstart is running without a hiccup on default stacks.

docs/book/component-guide/orchestrators/azureml.md
```diff
@@ -213,10 +213,18 @@ its [JobSchedules](https://learn.microsoft.com/en-us/azure/machine-learning/how-
...
+from zenml import pipeline
 from zenml.config.schedule import Schedule
 
+@pipeline
+def my_pipeline():
+    ...
+    
 # Run a pipeline every 5th minute
-pipeline.run(schedule=Schedule(cron_expression="*/5 * * * *"))
+my_pipeline = my_pipeline.with_options(
```
schustmi, line 224: 😅

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
```diff
@@ -184,6 +184,7 @@ class SagemakerOrchestratorConfig(
 
     Attributes:
         execution_role: The IAM role ARN to use for the pipeline.
+        scheduler_role: The IAM role ARN to use for the scheduler.
```
schustmi, line 187: We had some issues ourselves trying to figure out what this role is used for, and what permissions it needs. Maybe it makes sense to have a little more explanation here?

src/zenml/models/v2/core/schedule.py
```diff
@@ -136,6 +137,11 @@ class ScheduleResponseMetadata(WorkspaceScopedResponseMetadata):
     orchestrator_id: Optional[UUID]
     pipeline_id: Optional[UUID]
 
+    run_metadata: Dict[str, MetadataType] = Field(
+        title="Metadata associated with this step run.",
```
schustmi, line 141:
```suggestion
        title="Metadata associated with this schedule.",
```

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -269,6 +269,13 @@ def fetch_metadata_collection(self) -> Dict[str, List[RunMetadataEntry]]:
             for k, v in step_metadata.items():
                 metadata_collection[f"{s.name}::{k}"] = v
 
+        # Fetch the metadata related to the schedule of this run
+        if self.deployment is not None:
+            if schedule := self.deployment.schedule:
+                schedule_metadata = schedule.fetch_metadata_collection()
+                for k, v in schedule_metadata.items():
+                    metadata_collection[f"schedule:{k}"] = v
```
schustmi, line 277: What if I call a step `schedule`? 🤔

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -523,11 +722,28 @@ def get_pipeline_run_metadata(
...
-            "pipeline_execution_arn": pipeline_execution_arn,
+            "pipeline_execution_arn": execution_arn,
         }
 
+        zenml_client = Client()
+
+        if deployment_id := zenml_client.get_pipeline_run(
+            run_id
+        ).deployment_id:
+            deployment = zenml_client.get_deployment(deployment_id)
```
schustmi, line 735: Instead of doing two (hydrated) requests here, you could extend `self.get_settings(...)` to allow inputting a `PipelineConfiguration`, and then fetch that from the step context instead.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -588,13 +804,13 @@ def fetch_status(self, run: "PipelineRunResponse") -> ExecutionStatus:
 
     def compute_metadata(
         self,
-        execution: Any,
+        execution_arn: Any,
```
schustmi, line 807: Isn't this a `Optional[str]` now?

schustmi (changes requested):

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
bcdurak: You are absolutely right. A full explanation with the roles might be a bit convoluted here though. I am thinking of adding a link to the correct docs section. WDYT?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: There is one little difference. If there were a step called `schedule`, the corresponding metadata entries in this model would start with `schedule::` not `schedule:`.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: Agreed with `str` but why do you think we need to add the `Optional` here?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: Not a bad idea. Initially, I was a bit hesitant to change a function that is widely used in other parts of the codebase as well, but adding an additional optional variable shouldn't hurt. I would probably still keep the current code, in case someone wants to call this method outside of a step though. WDYT?

schustmi: As long as you don't remove functionality from it and just add to it, should be good :)
Mhm I guess the current code could be a fallback in case the `get_step_context` call fails? But this method is not supposed to be called outside of a step execution anyway, so I would even take it out to be honest.

src/zenml/integrations/aws/flavors/sagemaker_orchestrator_flavor.py
schustmi: Yep sounds good, short explanation and link to docs?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
schustmi: The first line in the function is a check `if execution_arn`, that's why I thought so

bcdurak: Yeah, I saw that too, and that if condition was redundant. I am adjusting the code accordingly.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: Fixed this as well.

schustmi (approved):

bcdurak (approved):
Merging this now.

PR #3288, opened 2025-01-15
Allow custom log formats

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/orchestrators/utils.py
```diff
@@ -40,6 +45,14 @@
 if TYPE_CHECKING:
     from zenml.artifact_stores.base_artifact_store import BaseArtifactStore
 
+ENV_VARS_TO_PROPAGATE = [
```
schustmi, line 48: @stefannica Do you think this is a reasonable change, to automatically forward these logging related environment variables to the Docker containers that run steps? Or should I revert this, or add another environment variable/setting somewhere that allows specifying which environment variables get forwarded?

src/zenml/logger.py
```diff
@@ -144,14 +145,26 @@ def set_root_verbosity() -> None:
...
 
 
+def get_formatter() -> logging.Formatter:
+    """Get a configured logging formatter.
+
+    Returns:
+        The formatter.
+    """
+    if log_format := os.environ.get(ENV_ZENML_LOGGING_FORMAT, None):
+        return logging.Formatter(fmt=log_format)
```
francoisserra, line 155: @schustmi  Do you think we can also add a standard Python [entry-point](https://packaging.python.org/en/latest/specifications/entry-points/) logic, in order to be able to inject custom implementation of the Logger? 
It is a clean way to provide extensibility in a plugin fashion, just by pip-installing dependencies in the runtime.
Here is an example of [how MLFlow does its plugin system](https://mlflow.org/docs/latest/plugins.html#defining-a-plugin).

schustmi: I think there are many ways to extend this (this was supposed to just be a quick way to have some basic configurability). There's also the logging file config which we could give access to I assume (https://docs.python.org/3/library/logging.config.html#logging-config-fileformat), but I'm curious to hear what kind of plugin you'd like, both specifically for this logging feature as well as in general in ZenML :)

francoisserra: My company uses a custom JSON log format to ingest in the central logging infrastructure. Some of the fields can come from environment variables, and some can be dynamically added based on the context (e.g., if there is an HTTP context, we add the `correlation_id`). We can also reformat the exception stack trace.
We use the `json-logging` lib to do this, thus we need to provide our implementation of the CustomLogger class.

src/zenml/orchestrators/utils.py
stefannica: The concern I have is that some of these client-side env vars may override whatever is set in the container images or docker settings that the users define for containerized pipelines. How would a user change the logging verbosity for a containerized pipeline before this PR ? And would that still work ?

schustmi: Yep that is the case with the current implementation. Previously, the only way for users to configure the logging behaviour for steps in container images was by setting these environment variables in the Docker settings, which means they will be included in the Dockerfile.

My changes in this PR essentially pass the environment as runtime options using the orchestrator, which will override the values present in the Docker image. So this would lead to different behaviour than previous versions of ZenML in cases where users have local settings that differ from the environment variables they set in the Docker settings. Which I definitely agree is not the best solution.
I see a few ways to change this:
- Remove the automatic forwarding of any environment variable. This means that the logging format will only be applied client-side, and if wanted users would also have to set it on the Docker settings to apply it to step logs.
- Only forward my new environment variable for the log format. This wouldn't change any existing behaviour, but would be very inconsistent in the way we treat these (logging) environment variables.
- Allow users to configure which environment variables from the client environment get forwarded (probably via a different environment variable or a pipeline configuration?). We could then still decide whether we want to default this to nothing, only the log format, or all logging environment variables

What do you think?

src/zenml/logger.py
francoisserra: @schustmi here's the pseudo-code / sequence we use when setting up  the log facility:

- Define a set of loggers to skip
- Define CustomLogger class (json_logging lib)
- Define function to get root logger names
- Define function to filter logger names based on the set of loggers to skip
- Define function to attach logger with handlers
- Define function to set up logger with parameters
- Clear existing handlers and set up new ones
- Attach loggers for main package and additional loggers if autodiscovery is enabled
- Set log level and propagation for the main logger

src/zenml/orchestrators/utils.py
schustmi: @stefannica @bcdurak After a discussion with @htahir1 we decided to keep the new environment variable consistent with all the existing ones, and revisit this environment forwarding when we take a closer look at environments.

docs/book/how-to/control-logging/set-logging-format.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+The logging format must use the `%`-string formatting style. Check out [this page](https://docs.python.org/3/library/logging.html#logrecord-attributes) for all available attributes.
+
+Note that setting this on the [client environment](../pipeline-development/configure-python-environments/README.md#client-environment-or-the-runner-environment) (e.g. your local machine which runs the pipeline) will **not automatically change the log format on remote pipeline runs**. That means setting this variable locally with only effect pipelines that run locally.
+
+If you wish to configure it also for [remote pipeline runs](../../user-guide/production-guide/cloud-orchestration.md), you can set the `ZENML_LOGGING_FORMAT` environment variable in your pipeline runs environment as follows:
+
+```python
+docker_settings = DockerSettings(environment={"ZENML_LOGGING_FORMAT": "%(asctime)s %(message)s"})
```
bcdurak, line 20: I like including the required imports in code blocks like this. It would help our users to copy it right away, while helping us to maintain it a bit better.

docs/book/reference/environment-variables.md
```diff
@@ -15,6 +15,14 @@ export ZENML_LOGGING_VERBOSITY=INFO
 
 Choose from `INFO`, `WARN`, `ERROR`, `CRITICAL`, `DEBUG`.
 
+## Logging format
+
+```bash
+export ZENML_LOGGING_FORMAT=%(asctime)s %(message)s
```
bcdurak, line 21:
```suggestion
export ZENML_LOGGING_FORMAT='%(asctime)s %(message)s'
```

src/zenml/constants.py
```diff
@@ -154,6 +155,7 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 ENV_ZENML_BACKUP_SECRETS_STORE_PREFIX = "ZENML_BACKUP_SECRETS_STORE_"
 ENV_ZENML_SKIP_PIPELINE_REGISTRATION = "ZENML_SKIP_PIPELINE_REGISTRATION"
 ENV_AUTO_OPEN_DASHBOARD = "AUTO_OPEN_DASHBOARD"
+ENV_ZENML_AUTO_OPEN_DASHBOARD = "ZENML_AUTO_OPEN_DASHBOARD"
```
bcdurak, line 158: I guess this is a new version of the (now deprecated) "AUTO_OPEN_DASHBOARD" variable. I just realized that we didn't even have the old one in the docs, so it might be nice to add that to the environment variables page as well.

src/zenml/utils/dashboard_utils.py
```diff
@@ -157,8 +158,17 @@ def show_dashboard(url: str) -> None:
 
     elif environment in (EnvironmentType.NATIVE, EnvironmentType.WSL):
         if constants.handle_bool_env_var(
-            constants.ENV_AUTO_OPEN_DASHBOARD, default=True
+            constants.ENV_AUTO_OPEN_DASHBOARD, default=False
```
bcdurak, line 161: With the current state, if they use the old variable and set it to `False`, this check will still return `True`. Here, I would still expect the old one to work, while throwing a warning.

Alternatively, if they configured the old one to be `True` and the new one to be `False`, this will still open the dashboard. 

Not sure, if this is the intended behavior, that's why I just wanted to note it down here. (IMO, if the new one is set, I would expect it to take precedence.)

src/zenml/constants.py
schustmi: I would keep the docs limited to the important env variables tbh. This is a niche env var, and everytime it actually affects you (when you login), the log message already tells you about it.

src/zenml/utils/dashboard_utils.py
schustmi: Previously, the dashboard automatically openend unless you set the legacy environment variable to `False`. IMO, setting any of these environment variable to `True` should be a no-op. It's the default behaviour anyway.

With the current logic, there is no precedence: If any of the two are set to `False`, the dashboard will not open automatically. If the old one is used, it will warn that it's deprecated.

WDYT?

docs/book/how-to/control-logging/set-logging-format.md
schustmi: I just copied this from a different docs page, I'll add it.

htahir1 (approved):

bcdurak (approved):

PR #3292, opened 2025-01-17
Pyproject.toml support for DockerSettings

schustmi:
## Describe changes

This PR adds the following functionality:
- support for `pyproject.toml` files to be specified for installing requirements in Docker images.
ZenML will extract the requirements from the pyproject.toml using one of the default or a custom user-provided command.
- install packages from a `requirements.txt` or `pyproject.toml` file in the source root if no requirements or pyproject file was specified in the `DockerSettings`.
- an attribute to disable the above automatic installation.
- a way to specify a command that installs a local python package.

## Pre-requisites

## Types of changes


schustmi:
@htahir1 Any ideas for the general docs structure of this? Or do we just leave it as-is for now?

htahir1:
So I think the docs need to change in these places too:

- In production guide on [this page](https://docs.zenml.io/user-guide/production-guide/configure-pipeline)

> Now that we have our pipeline up and running in the cloud, you might be wondering how ZenML figured out what sort of dependencies to install in the Docker image that we just ran on the VM. The answer lies in the [runner script we executed (i.e. run.py)](https://github.com/zenml-io/zenml/blob/main/examples/quickstart/run.py#L215), in particular, these lines:

- Probably i'd include the default behavior of trying to find requirements.txt and pyproject.toml implicitly here in the top page: https://docs.zenml.io/how-to/customize-docker-builds

- Probably makes sense to add the implicit behavior here as we mention a lot about allow_download_from_code_repository and allow_download_from_artifact_store? https://docs.zenml.io/how-to/customize-docker-builds/which-files-are-built-into-the-image

- Maybe we need to add it somewhere herE? https://docs.zenml.io/how-to/pipeline-development/configure-python-environments/handling-dependencies#use-a-tool-like-pip-compile-for-reproducibility Not sure exactly

- Finally, I would include it here as a note that pyproject.toml behavior is not recommended but if you include it this is what happens: https://docs.zenml.io/how-to/project-setup-and-management/setting-up-a-project-repository/set-up-repository

src/zenml/config/docker_settings.py
```diff
@@ -91,12 +91,19 @@ class DockerSettings(BaseSettings):
...
     Depending on the configuration of this object, requirements will be
     installed in the following order (each step optional):
-    - The packages installed in your local python environment
+    - The packages installed in your local python environment (extracted using
+      `pip freeze`)
     - The packages required by the stack unless this is disabled by setting
-      `install_stack_requirements=False`.
+      `install_stack_requirements=False`
     - The packages specified via the `required_integrations`
+    - The packages defined inside a pyproject.toml file given by the
```
bcdurak, line 99: I think there is a misalignment with the docs here. In the docs, we say the priority is as follows:

1. Packages in the local env 
2. Packages for the stack
3. Required Integrations
4. Requirements (could be a list or a path to a requirements.txt)
5. Pyproject.toml next

In the docstring here, 4 and 5 are switched.

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -437,8 +443,32 @@ def gather_requirements_files(
             - Packages installed in the local Python environment
             - Requirements defined by stack integrations
             - Requirements defined by user integrations
+            - Requirements exported from a pyproject.toml
```
bcdurak, line 446: Same issue with the order as above.

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -437,8 +443,32 @@ def gather_requirements_files(
...
             - User-defined requirements
         """
+        implicit_requirements = False
+        pyproject_path = docker_settings.pyproject_path
+        requirements = docker_settings.requirements
+
+        if not any(
+            [
+                docker_settings.replicate_local_python_environment,
+                docker_settings.required_integrations,
```
bcdurak, line 456: Does this mean that adding an extra integration (or a package) to be installed disables the discovery of the implicit requirements? So, if I was not using any of these settings before and now want to add `sklearn` to my docker settings, suddenly the `requirements.txt` will not be detected. I am not sure if this is the optimal behavior here.

src/zenml/config/docker_settings.py
```diff
@@ -192,8 +209,10 @@ class DockerSettings(BaseSettings):
     )
     python_package_installer_args: Dict[str, Any] = {}
     replicate_local_python_environment: Optional[
-        Union[List[str], PythonEnvironmentExportMethod]
+        Union[List[str], PythonEnvironmentExportMethod, bool]
     ] = Field(default=None, union_mode="left_to_right")
+    pyproject_path: Optional[str] = None
+    pyproject_export_command: Optional[List[str]] = None
```
bcdurak, line 215: As far as I can see, it is possible to set this as follows: `DockerSettings(pyproject_export_command=[......])` without an explicit `pyproject_path` as long as there is a `pyproject.toml` file to work with, in the current source root. But as soon as someone puts an integration, requirement or the replicate flag, this becomes unused.

bcdurak: Perhaps a validator can be helpful to detect this early.

src/zenml/config/docker_settings.py
bcdurak: I see that the docstring here is correct and the docs need to be updated.

bcdurak (commented):
Leaving some initial comments.

src/zenml/utils/pipeline_docker_image_builder.py
schustmi: Correct. I'm not super convinced by it myself to be honest, but what would be a better alternative in your opinion?

src/zenml/config/docker_settings.py
schustmi: What would you validate here? I don't see the problem of this being unused if there is no `pyproject.toml` to be honest

src/zenml/config/docker_settings.py
schustmi: You're right, I fixed the error in the docs.

In general, there are two scenarios:
- You don't explicitly specify any requirements using docker settings: In this case we first look for a `requirements.txt` (-> highest priority), and if it doesn't exist we look for a `pyproject.toml`
- You do explicitly specify some requirements (using e.g. `DockerSettings.required_integrations` or `DockerSettings.requirements`): In this case the order in the docstring is the one that is being used. Requirements being last means it is the highest priority (it will override values from all previous steps), which is in line with the first case.

Do you think these make sense?

src/zenml/utils/pipeline_docker_image_builder.py
bcdurak: Yeah, I guess I do not have a super clear alternative either. 

The reason why I was a bit hesitant here was this. We have a default behavior that decides whether we look for these files or not. By adding an additional integration, (that I intuitively thought would not change this behavior), I change the way that it looks for these files. I guess the `replicate_local_python_environment` flag at least allows people to control the behavior. Perhaps we can write a sentence about this in the docs?

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -68,6 +68,12 @@
 }
 UV_DEFAULT_ARGS = {"no-cache-dir": None}
 
+# TODO: these don't actually install any extras. Should we include all extras? No extras? Exclude dev extras?
```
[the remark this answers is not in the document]
bcdurak: Is this still relevant for this PR?

src/zenml/config/docker_settings.py
bcdurak: I would validate the following two scenarios:
- If `pyproject_export_command` is set, there needs to be a `pyproject.toml` in the directory.
- If `required_integrations` are set, `pyproject_export_command ` is unused. 

However, both are not so critical, I just thought early detection might help us improve the UX a bit.

src/zenml/config/docker_settings.py
bcdurak: I left a comment about the confusion that I had with the interaction between the implicit and explicit definition of requirements but outside of that, the explanation looks good here.

avishniakov:
@michael-zenml @htahir1 May I add my 5 cents here? Does adding `pyproject.toml` support mean that it will also rely on lock file or it will lock each time? In our setting it will be crucial to have it bounded to locks, since normally locks are also committed for consistency and reproducibility.
Aside from that - great addition to have in ZenML!

avishniakov:
A small companion PR to that Poetry activity: https://github.com/zenml-io/zenml/pull/3470

schustmi:
@avishniakov Users will have full control to specify a command that will be run to extract those requirements from the `pyproject.toml`. This will depend on what package manager you're using, for example in the case of `uv`, you could specify `uv export --frozen` which exports the requirements from the current lock file. Does that clarify, or do you think we need additional changes in this PR to make it work for your case?

avishniakov:
> @avishniakov Users will have full control to specify a command that will be run to extract those requirements from the `pyproject.toml`. This will depend on what package manager you're using, for example in the case of `uv`, you could specify `uv export --frozen` which exports the requirements from the current lock file. Does that clarify, or do you think we need additional changes in this PR to make it work for your case?

Yeah, if we can do something like `poetry export --without-hashes --format=requirements.txt` to specify - just good.

stefannica (changes requested):

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -434,8 +440,32 @@ def gather_requirements_files(
...
+            root = source_utils.get_source_root()
+            requirements_path = os.path.join(root, "requirements.txt")
+            pyproject_file_path = os.path.join(root, "pyproject.toml")
+
+            if os.path.exists(requirements_path):
+                implicit_requirements = True
+                requirements = requirements_path
+            elif os.path.exists(pyproject_file_path):
+                implicit_requirements = True
+                pyproject_path = pyproject_file_path
```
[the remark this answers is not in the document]
schustmi: You mean `True`, to disable the automatic behavior? I guess we want this to be the default going forward, so at some point this will have to be a breaking change? I'm not sure whether this should be now or some time in the future, @htahir1 maybe has an opinion?

src/zenml/config/docker_settings.py
```diff
@@ -149,13 +177,16 @@ class DockerSettings(BaseSettings):
...
         install_stack_requirements: If `True`, ZenML will automatically detect
             if components of your active stack are part of a ZenML integration
             and install the corresponding requirements and apt packages.
             If you set this to `False` or use custom components in your stack,
             you need to make sure these get installed by specifying them in
             the `requirements` and `apt_packages` attributes.
+        local_project_install_command: Command to install a local project in
+            the Docker image. This is run after the code files are copied into
+            the image, and it is therefore only possible when code is included
+            in the image, not downloaded at runtime.
```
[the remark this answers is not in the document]
schustmi: Yep agreed, I think we even have a ticket for this to enable it in a more generic way (and only for certain orchestrators), but this is out of scope for now and therefore the above is true at least for the current implementation

src/zenml/config/docker_settings.py
```diff
@@ -305,6 +334,55 @@ def _validate_skip_build(self) -> "DockerSettings":
...
+    @model_validator(mode="after")
+    def _deprecate_replicate_local_environment_commands(
+        self,
+    ) -> "DockerSettings":
+        """Deprecates some values for `replicate_local_python_environment`.
+
+        Returns:
+            The validated settings values.
+        """
+        if isinstance(self.replicate_local_python_environment, (str, list)):
```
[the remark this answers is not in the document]
schustmi: Yep true, applied it.

src/zenml/utils/pipeline_docker_image_builder.py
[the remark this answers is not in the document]
stefannica: I meant `True`, of course.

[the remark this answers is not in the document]
htahir1: I guess it isnt much of a breaking change but we can add it to the breaking change section in the release notes.... I would make it default tbh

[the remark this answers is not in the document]
htahir1: @schustmi i wouldnt worry so much about it - whats the worst that could happen tbh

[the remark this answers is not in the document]
schustmi: @htahir1 It potentially breaks CI (or any automated) pipelines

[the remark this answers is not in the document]
htahir1: @schustmi hmm but if we set this to false nobody will ever find out about it... but ur right... I'm unsure... maybe lets flip it for now but then we need docs to explain how to flip it back

stefannica (approved):

PR #3315, opened 2025-01-24
Schedule timezone fixes

schustmi:
## Describe changes

This PR contains some fixes to the timezones when using schedules:
- The user-facing `Schedule` object that is used to define schedules will treat any timezone unaware datetime as if it was in the local timezone. This enables users to write something like this `my_pipeline.with_options(Schedule(start_time=datetime.now())` which will create a schedule that starts immediately. To have more control over it, the `Schedule` object also allows users to pass timezone aware datetime, which will respect the given timezone.

- The `ScheduleRequest` object behaves a little differently: When it is passed a timezone aware datetime, it converts it to UTC timezone and then makes it timezone unaware. If it is passed a naive datetime, it simply assumes it is in UTC time. This means when we create this object internally as part of a pipeline run, we already have timezone aware datetimes and convert them to UTC. If someone were to externally call the `POST /schedules` endpoint and they pass naive datetimes, we treat them as UTC similar to all other endpoints.

## Pre-requisites

## Types of changes


src/zenml/config/schedule.py
```diff
@@ -43,17 +52,44 @@ class Schedule(BaseModel):
...
+
+        Returns:
+            A timezone aware datetime or None.
+        """
+        if value and value.tzinfo is None:
+            assert info.field_name
+            logger.warning(
+                "Your schedule `%s` is missing a timezone. It will be treated "
+                "as a datetime in your local timezone."
+            )
```
stefannica, line 89:
```suggestion
            logger.warning(
                f"Your schedule `{info.field_name}` is missing a timezone. It will be treated "
                "as a datetime in your local timezone."
            )
```

stefannica (approved):
One tiny adjustment, otherwise looks good

src/zenml/models/v2/core/schedule.py
```diff
@@ -46,15 +46,36 @@ class ScheduleRequest(WorkspaceScopedRequest):
...
+    run_once_start_time: Optional[datetime] = None
 
     orchestrator_id: Optional[UUID]
     pipeline_id: Optional[UUID]
 
+    @field_validator(
+        "start_time", "end_time", "run_once_start_time", mode="after"
+    )
+    @classmethod
+    def _ensure_tzunaware_utc(
```
bcdurak, line 62: I know we are not handling the update logic of schedules fully, but ideally, the same validator should be applied to the update model too right?

src/zenml/config/schedule.py
schustmi: We're sort of following this style guide for our code base, but even more so, the argument of filtering log messages makes a lot of sense I think. Right now we're just logging to stdout, but we've had multiple discussions already and I'm sure at some point we will allow customizing some logging backends where ZenML logs to. So we should just write our logs with these placeholders, to be prepared for when (not if) this is being implemented.

src/zenml/models/v2/core/schedule.py
schustmi: I have another branch (workspace-related) that removes all these options from the schedule update model. This should have never existed in the first place, schedules are just read-only currently in ZenML until we implement native schedules.

src/zenml/config/schedule.py
schustmi: Yep that I fixed, thanks 🙏

stefannica: "Your schedule `%s` is missing a timezone" comment is missing a schedule :smile:

bcdurak: image

src/zenml/models/v2/core/schedule.py
bcdurak: Alright, feel free to resolve then.

bcdurak (approved):

Issue #3268, opened 2024-12-18
[BUG]: Helm chart does not allow mounting TLS certificates


mazay:
### System Information

Helm chart version `0.71.0`

### What happened?

The Helm chart does not support mounting TLS certificates for using MySQL SSL encrypted connections.

The docs suggest adding certificates to the chart root, which is not feasible - we should not repackage the helm chart to enable such simple functionality.

Alternatively, the app could use the CA certificates already available on the pod, eg. `/etc/ssl/certs/ca-certificates.crt` when `?ssl=true` or a similar argument is present in the DB connection URL.

At the moment we had to make SSL encryption optional and sacrifice the security.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

htahir1:
@stefannica wdyt? We might consider customizing the Helm chart to mount the existing CA certificates directory from the pod into the application container. This would allow the application to use the pre-existing certificates without needing to repackage the Helm chart.

stefannica:
Giving this a try here: https://github.com/zenml-io/zenml/issues/3268

stefannica:
@mazay this should have been fixed with https://github.com/zenml-io/zenml/pull/3320. Please confirm and then close this issue when convenient.

mazay:
@stefannica thanks. Do you plan on releasing this?

stefannica:
> [@stefannica](https://github.com/stefannica) thanks. Do you plan on releasing this?

It will be in the next ZenML release, of course. I can't back-port it however, because it relies on some code changes too.

mazay:
@stefannica ok, we use helm artifact for deployment. So I can't test it until it's released

stefannica:
> [@stefannica](https://github.com/stefannica) ok, we use helm artifact for deployment. So I can't test it until it's released

@mazay the new Helm chart version (0.74.0) has just been published: https://artifacthub.io/packages/helm/zenml/zenml

stefannica:
I'll go ahead and close this issue, as the helm chart was released. Please feel free to re-open it if it doesn't solve your problem.

PR #3320, opened 2025-01-25
ZenML Helm chart improvements

stefannica:
## Describe changes
Implement a few Helm chart improvements:

* injecting custom CA certificates via inline values or Kubernetes secrets (implemented using init containers and copying the global CA certificate bundle from `/etc/ssl/certs/ca-certificates.crt` to a mounted location, then appending custom CA certs to it).
* HTTP proxy settings
* improved database SSL options by allowing certificates to be configured via inline values or Kubernetes secrets instead of files that had to be copied inside the helm chart folder
* allow to configure the database password by passing a Kubernetes secret instead of making it part of the DB URL

This PR also updates the self-hosted ZenML Pro documentation to provide some more insights and to document the GCP container registry support (rendered here: https://zenml-io.gitbook.io/stefan-docs/getting-started/zenml-pro/self-hosted)

Fixes https://github.com/zenml-io/zenml/issues/3268.

## Pre-requisites

## Types of changes


src/zenml/zen_server/deploy/helm/templates/custom-ca-configmaps.yaml
```diff
@@ -0,0 +1,20 @@
+{{- range $ca := .Values.zenml.certificates.customCAs }}
+{{- range $cm := list (dict "suffix" "") (dict "suffix" "-db-migration" "isDbMigration" true) }}
+---
+apiVersion: v1
+kind: ConfigMap
```
mazay, line 5: `Secret` is more suitable for storing certs

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
```diff
@@ -21,6 +21,21 @@ spec:
...
     spec:
+      {{- if or .Values.zenml.certificates.customCAs .Values.zenml.certificates.secretRefs }}
+      initContainers:
+        - name: update-ca-certificates
+          image: "{{ .Values.zenml.image.repository }}:{{ .Values.zenml.image.tag | default .Chart.Version }}"
+          imagePullPolicy: {{ .Values.zenml.image.pullPolicy }}
+          command: ["/bin/sh", "-c"]
+          args: ["update-ca-certificates && cp -r /etc/ssl/certs/* /updated-certs"]
+          securityContext:
+            runAsUser: 0  # Run as root
```
mazay, line 32: this is a security concern, also I don't think it's a good idea - the image should be immutable, to have predictable results each time you deploy an app

as long as images are built every few months there should be no need to update CAs on the container

src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
```diff
@@ -45,6 +60,13 @@ spec:
             - name: zenml-config
               mountPath: /zenml/.zenconfig
           env:
+            - name: REQUESTS_CA_BUNDLE
+              value: /etc/ssl/certs/ca-certificates.crt
+            - name: SSL_CERT_FILE
+              value: /etc/ssl/certs/ca-certificates.crt
+            # Force Node.js to use system CAs
```
mazay, line 67: how to make it use system CAs?

docs/book/getting-started/deploying-zenml/deploy-with-helm.md
```diff
@@ -658,4 +658,99 @@ podSecurityContext:
...
+Then update your values.yaml to use your custom images:
+```yaml
+zenml:
+  image:
+    repository: your-registry/zenml-server-with-certs
+    tag: your-tag
+```
+
+This chart offers a multitude of configuration options. For detailed
+information, check the default [`values.yaml`](values.yaml) file.
```
schustmi, line 721: Should this also link to https://artifacthub.io/packages/helm/zenml/zenml?modal=values?

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
```diff
@@ -101,6 +120,20 @@ spec:
               value: {{ $value | quote }}
             {{- end }}
             {{- end }}
+            {{- if .Values.zenml.proxy.enabled }}
+            - name: HTTP_PROXY
+              value: {{ .Values.zenml.proxy.httpProxy | quote }}
+            - name: HTTPS_PROXY
+              value: {{ .Values.zenml.proxy.httpsProxy | quote }}
+            - name: NO_PROXY
+              value: {{ include "zenml.noProxyList" . | quote }}
```
schustmi, line 129: Maybe I'm missing something, but I can't find where the `noProxyList` is defined. It should be a combination of `{{ .Values.zenml.proxy.noProxy }}` and `{{ .Values.zenml.proxy.additionalNoProxy }}`, correct?

schustmi (changes requested):

src/zenml/zen_server/deploy/helm/templates/custom-ca-configmaps.yaml
stefannica: Duly noted, will switch to secret.

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
stefannica: Yes, good point, I think I missed porting this over from the ZenML Pro chart.

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
mazay: > As mentioned in the documentation update changes in this same PR, dynamically updating system-wide CA certificates when the Helm chart is installed requires root access. However, this is isolated to the init container, while the workload container continues to run as a regular user. There is a workaround specified where you can inject CA certificates directly in the container image(s) at build time.

running as root in production environments is considered bad practice (even for `initContainers`) and may be forbidden which would lead to failed or infinitely pending pods

> Could you please clarify what you think is not a good idea? The Helm chart doesn't modify the container image. It only modifies the contents of the mounted volumes. The container image itself is in fact immutable by design (i.e. it is mounted as a read-only filesystem).

altering installed packages (I assume that's what `update-ca-certificates` does) is bad practice, same goes for altering stuff in the `/etc`

> Including CAs directly in the container image at build time is indeed a valid alternative that countless organizations use to adjust external container images to their internally generated SSL/TLS certificates. However, there's also the case of generating per-application CA and/or server self-signed certificates, in which the feature in this PR is useful.

this is true for home-brew images/projects but I have never seen companies rebuilding third-party images for adding CAs nor I consider it being a good practice

I think custom certificates should be mounted in a separate location and application should (conditionally?) load extra certificates from that location

I'm sorry, but I think this PR may be solving my initial issue but it most certainly brings extra issues that make it unusable in strict secure environments

> In the issue you raised (https://github.com/zenml-io/zenml/issues/3268), you said you need to use the CA certificate in the container to be able to use SSL for your MySQL connections. Can you please explain why you think this PR is not addressing that use-case?

I think the bigger problem is that application does not use CAs already present on the image either

I can use SequelAce or cli myslq client without importing any extra certificates, the clients are smart enough to recognize that the DB instance requires `tls` encryption and use certificates already existing on my laptop

ZenML doesn't even try using system certs, instead if I set `?ssl=true` it fails since I didn't provide cert paths and only way to add custom certs was to repackage the helm chart

that said, I think there are 2 parts of this issue:
- ZenML does not support MySQL TLS encryption
- ZenML does chart does not support mounting custom certs

stefannica: @mazay alright, thank you for pushing back on the root user access required for the init containers. I have reworked the solution and I think you'll find that the new approach is far better and matches all your requirements:

* still uses init containers, but no need for root access
* installing individual CA certificates system wide in the container via `zenml.certificates` also works for the MySQL database (I tested this)
* just in case, I also reworked all the SSL options under `zenml.database` to be able to take in certificates either as inline values or secret references. The `sslCa` option in particular is not really needed, as the same can be achieved by using `zenml.certificates`, but it's there nonetheless

src/zenml/zen_server/deploy/helm/templates/server-deployment.yaml
stefannica: @mazay so to clarify: the `pymysql` SQLAlchemy driver that we use in ZenML will respect the `REQUESTS_CA_BUNDLE` value. This means that we can set `REQUESTS_CA_BUNDLE` to point to the CA certificate bundle that we prepare by appending custom CA certs to the system one provided by the container image in `/etc/ssl/certs/ca-certificates.crt`. We wouldn't need to configure `zenml.database.sslCa` anymore (i.e. we won't need to point the MySQL URL to the CA certificate file).

src/zenml/zen_server/deploy/helm/values.yaml
```diff
@@ -956,6 +1016,35 @@ zenml:
       generateCerts: false
       secretName: zenml-tls-certs
 
+  # Configuration for custom CA certificates.
+  certificates:
+
+    # Direct injection of CA certificates in PEM format
+    customCAs: []
```
wjayesh, line 1023: naming this customCAs is a bit redundant as this block is all about custom CAs anyways. having this variable here and then a secretrefs next doesn't clearly tell you that they both point to custom CAs.
maybe we can use `inline` ?

schustmi (approved):

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
```diff
@@ -95,6 +181,27 @@ spec:
             - name: {{ $k }}
               value: {{ $v | quote }}
             {{- end }}
+            {{- if .Values.zenml.proxy.enabled }}
+            - name: HTTP_PROXY
+              value: {{ .Values.zenml.proxy.httpProxy | quote }}
+            - name: HTTPS_PROXY
+              value: {{ .Values.zenml.proxy.httpsProxy | quote }}
```
wjayesh, line 188: for my understanding, these env variables are consumed by python libraries we use, automatically?

wjayesh (changes requested):
This is great! finally, we get rid of the repackaging menace when it comes to using files in the chart 🙏

src/zenml/zen_server/deploy/helm/values.yaml
```diff
@@ -272,11 +272,71 @@ zenml:
...
+    # NOTE: setting any of the SSL certificate options (sslCa, sslCert, sslKey)
+    # will automatically set this option to true.
+    ssl: false
+
+    # The MySQL SSL CA certificate. Required for SSL enabled authentication if
+    # the CA certificate is not part of the certificates shipped with the
+    # container image. Can be provided as an inline value or as a Kubernetes
+    # secret reference.
+    #
+    # NOTE: it is recommended to use the `zenml.certificates` option instead.
```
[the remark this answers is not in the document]
stefannica: Using `zenml.certificates` is recommended because that's the way of configuring CA certificates globally, for all services in the container (database included). Configuring the CA certificate database here works only for the database.

I'll add this as an additional explanation here.

src/zenml/zen_server/deploy/helm/templates/server-db-job.yaml
stefannica: exactly !

wjayesh (approved):

PR #3329, opened 2025-01-29
Fixing the CI with the new `huggingface-hub` version

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/integrations/huggingface/services/huggingface_deployment.py
```diff
@@ -138,30 +139,67 @@ def inference_client(self) -> InferenceClient:
...
             hf_endpoint = create_inference_endpoint(
                 name=self._generate_an_endpoint_name(),
-                repository=self.config.repository,
-                framework=self.config.framework,
-                accelerator=self.config.accelerator,
-                instance_size=self.config.instance_size,
-                instance_type=self.config.instance_type,
-                region=self.config.region,
-                vendor=self.config.vendor,
+                repository=validated_config["repository"],
```
schustmi, line 189: Why are these attributes optional in the config if they're not actually optional when creating the inference endpoint? It feels like we could just fail much ealier and avoid this hack with dicts that prevent any type checking if we just made them required? Maybe @safoinme has an answer?

schustmi (changes requested):

src/zenml/integrations/huggingface/services/huggingface_deployment.py
bcdurak: I am also not sure myself. This was an external contribution and I am not sure what will happen for the people who already use this integration if I change optional values into required values in this configuration. As the call only happens in the `provision` call, I decided to go with this quick and dirty approach. As you said, perhaps @safoinme can shed more light on this.

src/zenml/integrations/huggingface/services/huggingface_deployment.py
```diff
@@ -253,15 +298,13 @@ def predict(self, data: "Any", max_new_tokens: int) -> "Any":
             )
         if self.prediction_url is not None:
             if self.hf_endpoint.task == "text-generation":
-                result = self.inference_client.task_generation(
```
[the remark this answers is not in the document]
schustmi: They probably introduced typing in their newest release, which means before the `inference_client` was simply `Any` which means calling any method on it won't lead to a mypy error

src/zenml/integrations/huggingface/services/huggingface_deployment.py
schustmi: It seems the base class with the optionals is also used for the component config (but then simply not used from what I can see), so I guess we need to leave it optional here and refactor at some point..

schustmi (approved):

PR #3335, opened 2025-02-03
Improve queries for pipelines, run templates, models and artifacts

schustmi:
## Describe changes
- Each time a pipeline was fetched from the DB, we also fetched all the runs for this pipeline in order to include the status/ID of the latest run in the model.
- Each time a run template was fetched from the DB, we also fetched all the runs for this template in order to include the status/ID of the latest run in the model.
- Each time an artifact was fetched from the DB, we also fetched all the versions for this artifact in order to include the ID/name of the latest version in the model.
- Each time a model was fetched from the DB, we also fetched all the versions for this model in order to include the ID/name of the latest version in the model.

As the runs and artifact/model versions can get quite a lot, this is very inefficient given that we're only interested in a single element (=the latest one). This PR fixes that by only querying for the latest run/version.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/schemas/artifact_schemas.py
+++ b/src/zenml/zen_stores/schemas/artifact_schemas.py
@@ -17,7 +17,8 @@ from typing import TYPE_CHECKING, Any, List, Optional
 from uuid import UUID
 
 from pydantic import ValidationError
-from sqlalchemy import TEXT, Column, UniqueConstraint
+from sqlalchemy import TEXT, Column, UniqueConstraint, desc, select
+from sqlalchemy.orm import object_session
 from sqlmodel import Field, Relationship
 
 from zenml.config.source import Source
@@ -90,6 +91,32 @@ class ArtifactSchema(NamedSchema, table=True):
         ),
     )
 
+    @property
+    def latest_version(self) -> Optional["ArtifactVersionSchema"]:
+        """Fetch the latest version for this artifact.
+
+        Raises:
+            RuntimeError: If no session for the schema exists.
+
+        Returns:
+            The latest version for this artifact.
+        """
+        if session := object_session(self):
+            return (
+                session.exec(
+                    select(ArtifactVersionSchema)
+                    .where(ArtifactVersionSchema.artifact_id == self.id)
+                    .order_by(desc(ArtifactVersionSchema.created))
+                    .limit(1)
+                )
+                .scalars()
+                .one_or_none()
+            )
+        else:
+            raise RuntimeError(
+                "Missing DB session to fetch latest run for pipeline."
+            )
+
     @classmethod
     def from_request(
         cls,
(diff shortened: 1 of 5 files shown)


src/zenml/zen_stores/schemas/artifact_schemas.py
```diff
@@ -90,6 +91,32 @@ class ArtifactSchema(NamedSchema, table=True):
...
+    def latest_version(self) -> Optional["ArtifactVersionSchema"]:
+        """Fetch the latest version for this artifact.
+
+        Raises:
+            RuntimeError: If no session for the schema exists.
+
+        Returns:
+            The latest version for this artifact.
+        """
+        if session := object_session(self):
```
stefannica, line 104: I know it is convenient to use a function like this to fetch the session out of thin air, but this goes against some design principles and breaks the current order of things. For one, nobody would suspect that you have a dependency here on a DB session if they follow the code bread crumbs, this is completely hidden from a module dependency perspective.

I highly recommend that you used one of two alternative approaches, to make this dependency more obvious and to keep the design aligned to the current status quo:

First, make `latest_version` a function instead of a property and have it take in the DB session as an argument (let's call it `get_latest_version` instead). Then you have one of three choices:

1. modify `to_model` to also take in the DB session as a mandatory argument, and make sure to pass it in all calls made to `to_model` from the SQL Zen Store
2. keep the `to_model` signature untouched, but call `get_latest_version` directly from SQL Zen Store *in addition to* `to_model`. Or, to make it clear that this is mandatory, you could even add `latest_version` as a mandatory argument to `to_model` and set it in SQL Zen Store to the result you get from calling the `get_latest_version` function.
3. in addition to 2. above, move `get_latest_version` to SQL Zen Store, which is where it should belong to in the first place. The schema files shouldn't have any SQL statements in them. They're just there to define schemas after all. We have a method called `_count_entity` that follows almost the same logic.

stefannica (changes requested):
The change looks good from a functional perspective, but it is breaks some design considerations. Please see my comment for some ways to improve this.

src/zenml/zen_stores/schemas/artifact_schemas.py
stefannica: Update: you explained that this is exactly how lazy loading works (hidden from the user and relies on the active DB session that created the schema object). In that case, this is fine. But please to check that:

1. this is indeed tied to the object and represents the DB session that created it
2. this is thread safe

stefannica (approved):

src/zenml/zen_stores/schemas/artifact_schemas.py
schustmi: I verified that this is indeed always the session object that was used to create/read the schema, and also ran some stress tests on our server without issues.

PR #3336, opened 2025-02-03
Runtime environment variables specification

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/utils/env_utils.py
```diff
@@ -152,3 +153,36 @@ def _substitution_func(v: str) -> str:
     return string_utils.substitute_string(
         value=value, substitution_func=_substitution_func
     )
+
+
+@contextlib.contextmanager
+def temporary_environment(environment: Dict[str, str]) -> Iterator[None]:
```
stefannica, line 159: This temporary context will only affect the current process. I think users may want to set pipeline-level environment variables that will have a global effect in the environment where the pipeline steps are running. That means that these variables should be injected somehow into `orchestrator.prepare_or_run_pipeline` and `step_operator.launch`. Is this possible ?

stefannica (changes requested):

src/zenml/utils/env_utils.py
schustmi: @stefannica That was actually my initial intention for this PR, however there are some complications: `orchestrator.prepare_or_run_pipeline` currently accepts a dictionary of environment variables. To make this work with our per-step configurations, we'd have to either change the public API of the orchestrator and pass different environments for all steps. Or we could expect/make it a convention that each orchestrator should update this dict with the step-specific environment configuration that is defined in the config. Or I guess we could make this a pipeline-only configuration, and don't allow setting specific environment variables for a steps, but I'm not a huge fan of this to be honest.

schustmi: WDYT, maybe the `convention` solution is the easiest way to go?
The orchestrator implementations would look something like this:
```
def prepare_or_run_pipeline(...):
  for step in deployment.steps():
    full_env = { **environment, **step.config.environment}
    ...
```

stefannica: This can work, yeah :smiling_face_with_tear:

htahir1 (changes requested):
Docs please! :D

src/zenml/utils/env_utils.py
schustmi: Alright, I'll adjust it then. Our local orchestrator currently ignores these environment variables. Do you think we should also set them for consistency? If yes, does that also include the auth ones that are coming from `prepare_or_run_pipeline`?

src/zenml/orchestrators/step_runner.py
```diff
@@ -186,11 +187,25 @@ def run(
...
+                step_config=step_run.config, stack=self._stack
+            )
+            step_environment.update(secret_environment)
+
             step_failed = False
             try:
-                return_values = step_instance.call_entrypoint(
-                    **function_params
-                )
+                with env_utils.temporary_environment(step_environment):
```
schustmi, line 212: TODO: Also wrap this around failure hook

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6840,6 +6859,97 @@ def _delete_backup_secret_values(
...
+            resource_type = resource_types[type(resource)]
+
+            query = select(SecretResourceSchema).where(
+                SecretResourceSchema.resource_id == resource.id,
+                SecretResourceSchema.resource_type == resource_type.value,
+                SecretResourceSchema.secret_id == secret_schema.id,
+            )
+            secret_resource = session.exec(query).first()
+            if secret_resource:
+                session.delete(secret_resource)
```
stefannica: I've seen this done in the past: first, fetching an entry, then deleting it if it exists. I think you can compress it into just one delete operation and save one roundtrip to the DB.

src/zenml/zen_stores/migrations/versions/a71cca37e42d_add_stack_component_env_and_secrets.py
```diff
@@ -0,0 +1,74 @@
...
+
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a71cca37e42d"
+down_revision = "0.84.0"
```
stefannica: Seriously outdated, please rebase.

src/zenml/utils/secret_utils.py
```diff
@@ -182,3 +187,54 @@ def is_clear_text_field(field: "FieldInfo") -> bool:
             )
 
     return False
+
+
+def resolve_and_verify_secrets(
+    secrets: List[Union[str, "UUID"]],
+    zen_store: Optional["ZenStoreInterface"] = None,
```
stefannica: This `zen_store` argument is never actually used by callers. I imagine you wanted to pass the zen_store object here when running from a server context, to avoid using `Client()` inside the server. I think you should just get rid of it, or simply detect in this function whether you're running in a server (and use `zenml.zen_server.utils.zen_store()`) or not (and use Client()). But generally, I don't think anyone cared whether Client() is created inside the server so far, and there were no problems AFAIK.

src/zenml/cli/stack.py
```diff
@@ -217,6 +217,14 @@ def stack() -> None:
...
     required=False,
 )
+@click.option(
+    "--secret",
+    "secrets",
+    help="Secret to attach to the stack.",
+    type=str,
+    required=False,
+    multiple=True,
+)
```
stefannica, line 227: Have you thought about adding a CLI option to set environment variables at stack level ?

src/zenml/cli/stack_components.py
```diff
@@ -232,6 +232,14 @@ def generate_stack_component_register_command(
...
         type=str,
     )
+    @click.option(
+        "--secret",
+        "secrets",
+        help="Secrets to attach to the component.",
+        type=str,
+        required=False,
+        multiple=True,
+    )
```
stefannica, line 242: Same here: have you thought about adding a CLI option to set environment variables at stack component level (update too) ?

src/zenml/config/step_configurations.py
```diff
@@ -303,10 +308,16 @@ def apply_pipeline_configuration(
...
+                    "environment",
+                    "secrets",
                 },
                 exclude_none=True,
             )
 
+            original_values["secrets"] = pipeline_values.get(
+                "secrets", []
+            ) + original_values.get("secrets", [])
+
```
stefannica, line 385: Here you are treating secrets separately, because they are not scalars: step-level secrets are added on top of pipeline secrets.

Suggestion: remove duplicates after merging (unless you do that somewhere else, like at processing time).
Question: shouldn't you also treat environment variables separately, and basically merge step env vars on top of pipeline env vars? It looks like step env vars are simply replacing ALL pipeline env vars currently.

stefannica: The question I've answered myself: the `get_step_environment` utility applies all of the priority rules you defined (`Component < Stack < Pipeline < Step`), which is why you need the original step env vars to be stored in the step configuration and not the merged ones.

src/zenml/config/compiler.py
```diff
@@ -651,3 +671,37 @@ def convert_component_shortcut_settings_keys(
...
+                and value != environment[key_without_prefix]
+            ):
+                logger.warning(
+                    "Got multiple values for environment variable `%s`.",
+                    key_without_prefix,
+                )
+            else:
+                environment[key_without_prefix] = value
+
+    return environment
```
stefannica, line 712: Can you please explain or document this in the PR ? I don't understand what this part does or why it's needed.

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -374,12 +390,17 @@ def configure(
             # merges dicts
             tags = self._configuration.tags + tags
 
+        if merge and secrets and self._configuration.secrets:
+            secrets = self._configuration.secrets + list(secrets)
+
```
stefannica, line 407: Just a tiny nit that I find it awkward that secrets are merged here, but environment variables are merged elsewhere. I'm sure there's a reason you chose to process these separately, it just feels strange from the POV of someone reviewing the code.

src/zenml/orchestrators/step_runner.py
```diff
@@ -186,11 +187,25 @@ def run(
...
+            # orchestrator. But for some orchestrators, this is not possible and
+            # we therefore make sure to set them here so they're at least
+            # available for the user code.
+            step_environment = env_utils.get_step_environment(
+                step_config=step_run.config, stack=self._stack
+            )
+            secret_environment = env_utils.get_step_secret_environment(
+                step_config=step_run.config, stack=self._stack
+            )
+            step_environment.update(secret_environment)
```
stefannica, line 208: If I understand this correctly:

* the non-secret environment variables are usually already set by the orchestrator even before any code is launched (e.g. as container env vars).
* you're setting the non-secret environment variables here anyway, just in case that didn't happen (i.e. for some orchestrators that don't support "container"-level env vars or similar)
* the secret env vars are NOT set by the orchestrator at all
* you're also setting the secret env vars here for that same reason

The limitations that come from this are:
* if non-secret env vars need to be set before the step runner is reached, or if they need to be set in a way that extends onto sub-processes, they won't be available in some orchestrators
* secret env vars will never be available for sub-processes or for code that executes before the step runner is reached

This is a limitation that should be documented.

src/zenml/utils/env_utils.py
stefannica: Yes, I think the local orchestrator should not ignore these env vars. As for the auth ones (i.e. the "base" env vars), I think you can ignore them.

stefannica (changes requested):
Some nits and questions, but this already looks amazing and is something I can reuse for Deployments, some of which support setting secrets separately from env vars at the infra level.

src/zenml/cli/stack_components.py
schustmi: I have, and I have no idea why I haven't implemented it yet 😅 I'll add it

src/zenml/cli/stack.py
schustmi: Will add this as well

src/zenml/zen_stores/migrations/versions/a71cca37e42d_add_stack_component_env_and_secrets.py
schustmi: Done

src/zenml/pipelines/pipeline_definition.py
schustmi: I have made some changes to the merging logic anyway which should make this cleaner. But in general, we have a helper function that recursively merged dictionaries. But for lists, we don't do that by default, which is why this manual code had to be added here.

src/zenml/zen_stores/sql_zen_store.py
schustmi: Done

src/zenml/utils/secret_utils.py
schustmi: Actually my current PR was missing that part in the run templates, which converts (potential) secret names to IDs. I'd keep this pattern with the zen_store to be honest, we use it in lots of other places where we access stuff server-side.

src/zenml/config/step_configurations.py
schustmi: Two things here:
- Environment also gets merged here, but because it's a dict, our util function handles it automatically, which it doesn't for lists.
- The deduplication happens at processing time

src/zenml/orchestrators/step_runner.py
schustmi: Yes all of that is correct. As soon as the general idea of this PR is approved, I'll write the docs and will make sure to include a point on this.

src/zenml/pipelines/pipeline_definition.py
schustmi: As mentioned in another comment, environment variables get automatically merged as they're a dictionary, and our utility function to merge dicts recursively already handles this.
The same thing applies to tags, which are also a list and are handled separately just above this.

Not sure if there's any easy/good way to handle this more generically, what do you think?

(All of this applies to the three places where we merge: here, similar code in the step class, and when merging step + pipeline config)

src/zenml/config/compiler.py
schustmi: I'm not even sure if we want to keep this tbh 😅

Essentially: If your current environment where you're running a pipeline in has some env variables set with prefix `__ZENML__`, we will automatically add them to the environment of the pipeline (after removing the prefix), which means they will be available in the execution environment (e.g. in the container running a step in kubernetes).

e.g. `__ZENML__TEST=1 python run.py` means the environment variable `TEST` will be set to value `1` wherever I run the pipeline.

WDYT, worth keeping or remove it? Maybe also @htahir1

src/zenml/utils/secret_utils.py
schustmi: Correction: I actually handled the secrets for run templates already, but in a much nicer way. So I did indeed not need this anymore and removed it :)

src/zenml/config/compiler.py
stefannica: My first thought is this pattern is too weird to be taken seriously by anyone. And now that you have several other really nice ways of passing in env variables, nobody is going to use something that feels as weird as this and is also problematic from a reproducibility POV.

src/zenml/pipelines/pipeline_definition.py
stefannica: > environment variables get automatically merged as they're a dictionary

This was a key piece of information I didn't have before, but now it makes sense that they are separate. I'm probably missing some details, but an easy/good way to handle this more generically could be that our `pydantic_utils.update_model` utility also handles merging lists, not just dictionaries, so they're all processed in the same place rather than separately. But doing this might have backwards-incompatible repercussions, if there are other pipeline/step settings that are lists.

schustmi: Yes exactly, this is also used to merge all settings, which includes all settings of all integration flavors, many of which have lists that shouldn't (at least for backwards compatibility, not sure if semantically) be merged.

stefannica (approved):

Issue #3341, opened 2025-02-05
[BUG]: Model.get_artifact not loading the latest version


pierre-godard:
### System Information

ZENML_LOCAL_VERSION: 0.73.0
ZENML_SERVER_VERSION: 0.73.0
ZENML_SERVER_DATABASE: mysql
ZENML_SERVER_DEPLOYMENT_TYPE: other
ZENML_CONFIG_DIR: /home/pierre/.config/zenml
ZENML_LOCAL_STORE_DIR: /home/pierre/.config/zenml/local_stores
ZENML_SERVER_URL: http://localhost:8080
PYTHON_VERSION: 3.12.4
ENVIRONMENT: native
SYSTEM_INFO: {'os': 'linux', 'linux_distro': 'ubuntu', 'linux_distro_like': 'debian', 'linux_distro_version': '20.04'}

### What happened?

I ran multiple times a pipeline to train a model. It looks something like this:

```python
from typing import Annotated
import pandas as pd
from zenml import ArtifactConfig, Model, pipeline, step
from sklearn.base import BaseEstimator

my_model = Model(name="my_model", version="1.2.3")

@step
def my_estimator_fit_step(
    X: pd.DataFrame, y: pd.DataFrame
) -> Annotated[BaseEstimator, ArtifactConfig(name="my_estimator", artifact_type=ArtifactType.MODEL)]:
    # Define and fit my_estimator
    return my_estimator


@pipeline(model=my_model)
def fit_pipeline():
    # Data preparation
    my_estimator = my_estimator_fit_step(X_train, y_train)
```

Then I have another pipeline that uses this model by loading it from the `my_model` object:

```python
@pipeline
def predict_pipeline():
    # Data preparation
    my_estimator = my_model.get_model_artifact("my_estimator")
    # Use the estimator to generate predictions
```

After many runs (10+ without cache) of `fit_pipeline` under the same "my_model" version (in this example `1.2.3`), I would expect `my_model` to load the latest fit version of "my_estimator". Instead, I get the model artifact with the version name `"9"`.

This version name is given automatically and is auto-incremental. I noticed that it is a string and after some investigation of the source code, I found the following in `zenml/models/v2/core/model_version.py:ModelVersionResponse._get_linked_object`:
```python
if version is None:
    version = max(collection[name].keys())
```
[...]

PR #3343, opened 2025-02-05
Get the latest artifact of a model by creation date instead of version name

pierre-godard:
## Describe changes
Solves the issue in https://github.com/zenml-io/zenml/issues/3341.

## Pre-requisites

## Types of changes


src/zenml/models/v2/core/model_version.py
```diff
@@ -449,7 +449,10 @@ def _get_linked_object(
         if name not in collection:
             return None
         if version is None:
-            version = max(collection[name].keys())
+            return max(
+                collection[name].values(),
+                key=lambda a: client.get_artifact_version(a).created,
```
schustmi, line 454: As you correctly mentioned, this will now run N requests to fetch all the versions first.
I would suggest a different way, which would require a little more work:
- Use `Client().list_artifact_versions(sort_by="desc:created", size=1, name=artifact_name, model_version_id=self.id)`. This should fetch you the latest created version for that artifact that is associated with a model version.
- The method above also allows filtering by `type`, which would be relevant when `_get_linked_object` is being called from `get_model_artifact(...)` instead of just `get_artifact(...)`. So you'd also have to refactor `get_linked_object` to optionally accept a type parameter that it filters for.

Does that make sense? Let me know if I should explain more

pierre-godard: Perfectly makes sense, thank you for the detailed answer! It feels much better

src/zenml/models/v2/core/model_version.py
```diff
@@ -527,11 +524,7 @@ def get_deployment_artifact(
         Returns:
             Specific version of the deployment artifact or None
         """
-        return self._get_linked_object(
-            self.deployment_artifact_ids,
-            name,
-            version,
-        )
+        return self._get_linked_object(name, version, ArtifactType.SERVICE)
```
pierre-godard, line 527: From the documentation I understood that "deployment artifacts" correspond to `ArtifactType.SERVICE` (the term "deployment" being deprecated). Is that correct in this context as well?

schustmi: Yep that is correct 👍

src/zenml/models/v2/core/model_version.py
```diff
@@ -438,19 +438,25 @@ def _get_linked_object(
...
         from zenml.client import Client
 
-        client = Client()
+        artifact_versions = Client().list_artifact_versions(
+            sort_by="desc:created",
+            size=1,
+            name=name,
+            version=version,
+            model_version_id=self.id,
+            type=type,
```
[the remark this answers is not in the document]
pierre-godard: Good to know! I just added it

schustmi (approved):

PR #3357, opened 2025-02-11
Improve source root docs/logs when running a pipeline

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/how-to/project-setup-and-management/setting-up-a-project-repository/set-up-repository.md
+++ b/docs/book/how-to/project-setup-and-management/setting-up-a-project-repository/set-up-repository.md
@@ -98,14 +98,13 @@ Collect all your notebooks in one place.
 
 #### .zen
 
-By running `zenml init` at the root of your project, you define the project scope for ZenML. In ZenML terms, this will be called your "source's root". This will be used to resolve import paths and store configurations.
-
-Although this is optional, it is recommended that you do this for all of your
-projects. This is especially important if you are using Jupyter noteeboks in
-your project as these require you to have initialized a `.zen` file.
+By running `zenml init` at the root of your project, you define the project scope for ZenML. In ZenML terms, this will be called your "source root". This will be used to resolve import paths and store configurations.
+- When running Jupyter notebooks, it is required that you have a `.zen` directory initialized in one of the parent directories of your notebook.
+- When running regular Python scripts, it is still **highly** recommended that you have a `.zen` directory initialized in the root of your project. If that is not the case,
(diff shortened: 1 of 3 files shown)


schustmi:
@stefannica I already added a log that tells users to check whether their source root is correct before uploading the code. Do you think we should have an additional warning that we print when the `.zen` directory is not the same as the implicit source root (the parent directory of the `__main__` module file)? Or only if there are a certain number of directory levels between them?

strickvl:
@schustmi I saw that thread and I was wondering whether we need to model / explain (at least very briefly) the existence / use of `zenml init` already in the starter guide on [this page](https://docs.zenml.io/user-guide/starter-guide/create-an-ml-pipeline) somewhere?

schustmi:
> @schustmi I saw that thread and I was wondering whether we need to model / explain (at least very briefly) the existence / use of `zenml init` already in the starter guide on [this page](https://docs.zenml.io/user-guide/starter-guide/create-an-ml-pipeline) somewhere?

@strickvl I'm not sure if that might be too early and complicated on this page. It's not really relevant for local runs anyway, and even in remote runs from a single file it's not really necessary as it should work as expected. Especially given that the user probably hasn't used ZenML before, and therefore doesn't have any orphan `.zen` directories on their machine. WDYT?

strickvl:
> > @schustmi I saw that thread and I was wondering whether we need to model / explain (at least very briefly) the existence / use of `zenml init` already in the starter guide on [this page](https://docs.zenml.io/user-guide/starter-guide/create-an-ml-pipeline) somewhere?
> 
> @strickvl I'm not sure if that might be too early and complicated on this page. It's not really relevant for local runs anyway, and even in remote runs from a single file it's not really necessary as it should work as expected. Especially given that the user probably hasn't used ZenML before, and therefore doesn't have any orphan `.zen` directories on their machine. WDYT?

Yeah I wasn't even thinking to go into that detail. Just to say something simple like "oh and do a `zenml init` at the root of your repository... it might well save you some pain in the future" and then link out to the main place we talk about why this is a thing.

schustmi:
@strickvl Added a small mention

stefannica (approved):

stefannica:
> @stefannica I already added a log that tells users to check whether their source root is correct before uploading the code. Do you think we should have an additional warning that we print when the `.zen` directory is not the same as the implicit source root (the parent directory of the `__main__` module file)? Or only if there are a certain number of directory levels between them?

I think what you did should be enough.

PR #3359, opened 2025-02-13
Enable Weave integration in Wandb settings

htahir1:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/integrations/wandb/__init__.py
```diff
@@ -30,7 +30,7 @@ class WandbIntegration(Integration):
     """Definition of Plotly integration for ZenML."""
 
     NAME = WANDB
-    REQUIREMENTS = ["wandb>=0.12.12", "Pillow>=9.1.0"]
+    REQUIREMENTS = ["wandb>=0.12.12", "Pillow>=9.1.0", "weave>=0.51.33"]
```
strickvl, line 33:
```suggestion
    REQUIREMENTS = ["wandb>=0.12.12,<1.0.0", "Pillow>=9.1.0", "weave>=0.51.33,<1.0.0"]
```

Maybe?

strickvl (changes requested):
Just one small comment about the dependencies that I noticed, but otherwise lgtm. I'll test this now on my end.

strickvl:
Tested it with [this file](https://gist.github.com/strickvl/1f06131b3841fccc4b3fd4c5ae37f893) and while it does indeed initialise a weave project, it doesn't actually track any of the LLM calls inside. I think I'd assume that the ZenML integration would work a little bit like the `autolog` functionality of the main wandb package where you don't have to manually track things. [The docs](https://weave-docs.wandb.ai/quickstart) show `@weave.op()` in use which is maybe what we need to automatically set up to wrap each step?

strickvl (changes requested):
Docs should be clear(er) on exactly what will get logged with this weave addition. i.e. either llm calls are autologged or they're not.

And as mentioned in my earlier comment, I think probably we need to set this up to handle autologging (or show in the docs how someone would manually log their llm calls with this integration).

htahir1:
@strickvl should be way better now

strickvl:
> @strickvl should be way better now

I'm not sure that this current implementation is the way we want to enable tracing of LLM calls within zenml.

Will see if I can take a deeper look when back, but at first glance we're now putting the responsibility on the user to add the `@weave.op()` decorator. Shouldn't we enable tracing automatically by basically calling / using the decorator within the ZenML step decorator (if someone had it enabled). I looked into that previously when I was seeing if we could enable the same for braintrust and all the other tracing providers. IMO we should find a way to autolog/autotrace whatever llm calls happen in the step *without* forcing the user to manually add this extra decorator. I discussed it previously with @schustmi how we might want to do this.

The key thing is if someone switches their experiment tracker out they shouldn't have to then remove all these `@weave.op()` calls from their zenml code, just like we do with all other components.

htahir1:
@strickvl i actually disagree... weave.op itself has many parameters that will change over time and we'll have to constantly expose them to the user. its not much effort to just put weave.op at the top of your step. i dislike abstractions for the sake of abstractions here

strickvl:
But then we lose the 'write your code once and keep it the same when you switch out our stack component' completely. At that point then I don't know what the benefit of the weave integration is.

htahir1:
I know but tbh thats hard anyway .. we can rediscuss when we do more integrations

src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
```diff
@@ -131,25 +127,50 @@ def cleanup_step_run(self, info: "StepRunInfo", step_failed: bool) -> None:
...
             name=run_name,
             tags=tags,
-            settings=settings,
+            settings=settings.settings,
         )
+
+        if settings and settings.enable_weave:
+            import weave
+
+            logger.info("Initializing weave")
```
schustmi, line 159: Maybe this log could be in the `if self.config.project_name` to not clutter the logs?

src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
```diff
@@ -131,25 +127,50 @@ def cleanup_step_run(self, info: "StepRunInfo", step_failed: bool) -> None:
...
+            info: Step run information.
             run_name: Name of the wandb run to create.
             tags: Tags to attach to the wandb run.
-            settings: Additional settings for the wandb run.
         """
         logger.info(
             f"Initializing wandb with entity {self.config.entity}, project "
             f"name: {self.config.project_name}, run_name: {run_name}."
         )
+        settings = cast(
```
schustmi, line 145: This returns a non-optional, so you can get rid of the `if settings and ...` part in those conditions.

src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
```diff
@@ -131,25 +127,50 @@ def cleanup_step_run(self, info: "StepRunInfo", step_failed: bool) -> None:
...
+                weave.init(project_name=self.config.project_name)
+            else:
+                logger.info(
+                    "Weave enabled but no project_name specified. "
+                    "Skipping weave initialization."
+                )
+        elif settings and not settings.enable_weave:
+            import weave
+
+            logger.info("Disabling weave")
```
schustmi, line 171: This log should also only happen when we actually do it I think

schustmi (changes requested):

docs/book/component-guide/experiment-trackers/wandb.md
```diff
@@ -196,13 +197,109 @@ def my_step(
...
+    return "Step with Weave disabled"
+```
+
+#### Weave Initialization Behavior
+
+When using Weave with ZenML, there are a few important behaviors to understand:
+
+1. If `enable_weave=True` and a `project_name` is specified in your W&B experiment tracker, Weave will be initialized with that project name.
+2. If `enable_weave=True` but no `project_name` is specified, Weave initialization will be skipped.
+3. If `enable_weave=False` (explicit disabling), Weave will be disabled with `settings={"disabled": True}`.
```
schustmi, line 290: This also requires a project name to work (Not sure why but at least in code that's what's needed right now)

src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
```diff
@@ -131,25 +127,49 @@ def cleanup_step_run(self, info: "StepRunInfo", step_failed: bool) -> None:
...
+
+            if self.config.project_name:
+                logger.info("Initializing weave")
+                weave.init(project_name=self.config.project_name)
+            else:
+                logger.info(
+                    "Weave enabled but no project_name specified. "
+                    "Skipping weave initialization."
+                )
+        elif settings and not settings.enable_weave:
```
schustmi, line 167:
```suggestion
        else:
```

htahir1:
@schustmi done

strickvl (approved):

PR #3372, opened 2025-02-24
Pending Changes for Active Stack

preyasshah9:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/config/pipeline_run_configuration.py
```diff
@@ -46,6 +47,7 @@ class PipelineRunConfiguration(
     steps: Dict[str, StepConfigurationUpdate] = {}
     settings: Dict[str, SerializeAsAny[BaseSettings]] = {}
     tags: Optional[List[str]] = None
+    stack: Optional[Stack] = None
```
schustmi, line 50: This should be an `Optional[str]` instead. The `Stack` object is not something that can be defined in a yaml file, which is usually how users specify attributes of this run configuration class. The value for `stack` will be the name or ID of a stack, which will then be passed to `temporary_active_stack`.

schustmi (changes requested):
There is one major part missing here still: When running/building a pipeline for which this config attribute is set, we actually need to use the `temporary_active_stack` context manager to switch the stack for the run/build.

This will be in `PipelineDefinion.build(...)` and `PipelineDefinition._run(...)`.

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -565,19 +566,26 @@ def build(
...
+                matcher=list(PipelineRunConfiguration.model_fields.keys()),
             )
+            run_config = PipelineRunConfiguration(**_from_config_file)
 
-            return build_utils.create_pipeline_build(
-                deployment=deployment,
-                pipeline_id=pipeline_id,
-                code_repository=code_repository,
-            )
+            with temporary_active_stack(stack_name_or_id=run_config.stack):
```
preyasshah9, line 575: Read the value of stack from the `PipelineRunConfiguration` and if available use that within the `temporary_active_stack` context manager. Same below.

preyasshah9:
@schustmi Hi Michael,

Thank you for the feedback on PR. I have processed through code and tried to address your comment to best of my understanding. Let me know if this sounds good!
Let me know if this sounds good. If you have any pointers on how to test this change specifically, I would really appreciate this.

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -565,19 +566,26 @@ def build(
             if settings:
                 compile_args["settings"] = settings
 
-            deployment, _, _ = self._compile(**compile_args)
-            pipeline_id = self._register().id
-
-            local_repo = code_repository_utils.find_active_code_repository()
-            code_repository = build_utils.verify_local_repository_context(
-                deployment=deployment, local_repo_context=local_repo
+            _from_config_file = self._parse_config_file(
```
schustmi, line 569: Not a huge fan of you having to do this manually before the compilation step, but I also don't see any easy way around this. I'll refactor this (the entire config file logic is very messy right now) after this PR is merged.

schustmi (changes requested):
One thing I forgot: Could you also add this logic to temporarily set a different stack on the `PipelineDefinition.create_run_template` method?

As for testing this:
- The test should first register a second stack (should probably just be default local orchestrator/artifact store) 
- The active stack should be the default one
- We then configure any pipeline (just a single-step, do nothing pipeline should be fine) to run on a different stack.
- The above bullet point can be done either via a config file, or by passing it directly to `test_pipeline.with_options(stack=...)`. These should probably be two different test cases.
- It should then run the pipeline `run = test_pipeline()`, and check whether the run ran on the configured stack. `assert run.stack.id == ...`
- We can then also check that the stack gets correctly reset to the previously active stack: `assert Client().active_stack.id == ...`

I think examples for most of these can be found in `tests/unit/pipelines/test_base_pipeline.py` (e.g. running a pipeline, configuring from a file). Let me know if I can help in any other way

schustmi:
> > PipelineDefinition.create_run_template
> 
> When `create_run_template` function is called, it is already within the context manager of `temporary_active_stack`:
> 
> https://github.com/zenml-io/zenml/blob/0d059d3a54d7ab86b2e3057319fb75f2e6c25451/src/zenml/cli/pipeline.py#L363
> 
> I can also add a context manager but I believe this case is already handled.

Yep, but that is only available when using the CLI to create a run template, but not when doing this from purely in code. I think the better solution would be to remove the context manager from the CLI function, and instead pass the `--stack` CLI option via the newly introduced `PipelineRunConfiguration.stack` attribute if set.

preyasshah9:
@schustmi Added some unit tests as per your comment. When I ran locally, I did not see issues related to the tests I added. I don't think unit tests are auto-executed by CI here. I executed tests on my branch using: `bash scripts/test-coverage-xml.sh`

preyasshah9:
> > > PipelineDefinition.create_run_template
> > 
> > 
> > When `create_run_template` function is called, it is already within the context manager of `temporary_active_stack`:
> > https://github.com/zenml-io/zenml/blob/0d059d3a54d7ab86b2e3057319fb75f2e6c25451/src/zenml/cli/pipeline.py#L363
> > 
> > I can also add a context manager but I believe this case is already handled.
> 
> Yep, but that is only available when using the CLI to create a run template, but not when doing this from purely in code. I think the better solution would be to remove the context manager from the CLI function, and instead pass the `--stack` CLI option via the newly introduced `PipelineRunConfiguration.stack` attribute if set.

One last question. I did pass the `stack` option to the `pipeline_instance.with_options`. However, passing it to the `create_run_template`, there are a couple of calls this function is making:

```
deployment = self._create_deployment(**self._run_args, skip_schedule_registration=True)
return Client().create_run_template(name=name, deployment_id=deployment.id, **kwargs)
```

Here the [create_run_template](https://github.com/zenml-io/zenml/blob/271076b49fa4fb6954cebf0a8b8e497bc4d2ae9a/src/zenml/client.py#L3455), but it is not possible (or required) to patch the `stack`.  Let me know if this needs to be updated.

However, `_create_deployment` seems an important place to pass the `stack` option. This also calls the `_compile` function. They both use the `active_stack`. But there is no easier way to pass it outside of the configuration file. 
https://github.com/zenml-io/zenml/blob/271076b49fa4fb6954cebf0a8b8e497bc4d2ae9a/src/zenml/pipelines/pipeline_definition.py#L1053.

Any suggestions?

preyasshah9:
@schustmi Friendly ping for the clarification :)

preyasshah9:
@schustmi It's been a while! Any feedback on the comment above?

schustmi:
@preyasshah9 Sorry for the late comment, I lost track of this among so many ongoing things.

The code in the `pipeline_definition.py` file on develop has changed a bit, can you resolve those merge conflicts before we have another look please.

In general, the `create_run_template` method when being called from a python file by the user (not using the CLI function) will still not respect the stack. So yes you're right we might need to implement this generically in the `_create_deployment` method somehow.

preyasshah9:
The merge conflicts have been resolved!

bcdurak:
Hey @preyasshah9, thank you for your contribution on this ticket. I'm closing it due to staleness, feel free to reopen it if you'd like to keep working on it.

PR #3379, opened 2025-02-25
Add SMTP Email Alerter integration

strickvl:
This commit adds a new SMTP Email Alerter integration, allowing users to send email notifications from within ZenML pipelines and steps.

Features:
- Support for both plain text and HTML-formatted emails
- Documentation with detailed examples and setup instructions
- Support for customizable email subjects and recipients
- Predefined step for easy integration into pipelines

🤖 Generated with Claude Code


src/zenml/integrations/smtp_email/alerters/smtp_email_alerter.py
```diff
@@ -0,0 +1,336 @@
...
+        will always return False and log a warning.
+
+        Args:
+            question: Not used.
+            params: Not used.
+
+        Returns:
+            Always False.
+        """
+        logger.warning(
```
schustmi, line 332: Any reason not to raise an exception here?

schustmi (changes requested):

strickvl:
@coderabbitai review

src/zenml/integrations/smtp_email/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+            List of new flavors defined by the SMTP Email integration.
+        """
+        from zenml.integrations.smtp_email.flavors import SMTPEmailAlerterFlavor
+
+        return [SMTPEmailAlerterFlavor]
+
+
+SMTPEmailIntegration.check_installation()
+
+# Import hooks AFTER integration registration to avoid circular imports
```
schustmi, line 50: Any reason we really need this code block? Seems like no other integration does this, and in the failure case it might even cause us to not import the integration as `__all__` is set to an empty list?

src/zenml/integrations/smtp_email/alerters/smtp_email_alerter.py
```diff
@@ -0,0 +1,398 @@
...
+                nor in the alerter component.
+        """
+        if params and not isinstance(params, BaseAlerterStepParameters):
+            raise RuntimeError(
+                "The config object must be of type `BaseAlerterStepParameters`."
+            )
+        if (
+            params
+            and isinstance(params, SMTPEmailAlerterParameters)
+            and hasattr(params, "recipient_email")
```
schustmi, line 105:
```suggestion
```

I think these are both not necessary as it's always true?

src/zenml/integrations/smtp_email/steps/smtp_email_alerter_post_step.py
```diff
@@ -0,0 +1,70 @@
...
+            in the alerter.
+        subject: Optional custom subject for the email.
+        include_pipeline_info: If True, includes pipeline and step information
+            in the email.
+
+    Raises:
+        RuntimeError: If no alerter is configured in the active stack.
+    """
+    alerter = Client().active_stack.alerter
+    if not alerter:
```
schustmi, line 48: Shouldn't we also check the alerter type here?

In general it feels kinda weird to me what is happening with alerters. Shouldn't there be **one** alerter post step, which works independently of the alerter type? This is the whole reason for stack switching.

I would suggest that we think of a generic format that get's passed to `alerter.post(...)`, and we can then get rid of all these specific steps. We can also avoid checks in the open AI hooks for example that handle the email alerter separately. The alerter itself should be responsible for the formatting, not some outside logic IMO.

src/zenml/integrations/smtp_email/alerters/smtp_email_alerter.py
```diff
@@ -0,0 +1,398 @@
...
+            logger.info("Email alert sent successfully")
+            return True
+
+        except Exception as e:
+            logger.error(f"Error sending email alert: {str(e)}")
+            return False
+
+    def ask(
+        self, question: str, params: Optional[BaseAlerterStepParameters] = None
+    ) -> bool:
```
schustmi, line 381:
```suggestion
    ) -> Never:
```

`from typing import Never`

src/zenml/integrations/smtp_email/alerters/smtp_email_alerter.py
```diff
@@ -0,0 +1,398 @@
...
+            )
+        except RuntimeError:
+            settings = None
+
+        if settings is not None and settings.include_html is not None:
+            return settings.include_html
+
+        return self.config.include_html
+
+    def _get_subject_prefix(self) -> str:
```
schustmi, line 162: All of these helper methods seem to be doing the same exact thing, can we merge that into a single method that takes the attribute name (and potentially a `fail_if_missing`) as arguments?

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
```diff
@@ -102,11 +270,112 @@ def openai_alerter_failure_hook_helper(
...
+Exception: ({type(exception).__name__}) {exception}
+
+OpenAI ChatGPT's suggestion (model = {model_name}) on how to fix it:
+{suggestion}
+
+{plain_traceback}
+"""
+            # For HTML email, create a custom HTML body with proper formatting
+            html_body = f"""
+            <html>
```
schustmi, line 296: I did not verify in detail, but is there a way to remove some duplication in all these html strings in various places of this PR?

strickvl: Done

src/zenml/integrations/smtp_email/steps/smtp_email_alerter_post_step.py
strickvl: I've heavily updated it based on this comment. Hope I haven't gone too far. There's no breaking changes, but it's def a fairly large change.

src/zenml/integrations/openai/hooks/open_ai_failure_hook.py
```diff
@@ -13,20 +13,180 @@
...
 
 from zenml import get_step_context
 from zenml.client import Client
+from zenml.hooks import get_failure_template
 from zenml.logger import get_logger
 
 logger = get_logger(__name__)
 
 
+def _format_markdown_for_html(markdown_text: str) -> str:
```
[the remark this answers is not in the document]
strickvl: Ok yes you might be right. This isn't the nicest code. I'll work on improving it.

strickvl:
@bugbot run

strickvl:
@bugbot run

strickvl:
@codex review

strickvl:
@codex review

htahir1:
Closing due to staleness - feel free to open again whenever you are ready :-)

PR #3386, opened 2025-02-28
Store build duration

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


bcdurak (approved):
Will approve this for now. Just out of curiosity, do we have to put the `duration` in the `Base` model instead of the `Request` model? I was wondering if this would cause any issues since the `Base` model is used in other contexts like the `build_utils` and the `PipelineRunConfiguration`.

schustmi:
> Will approve this for now. Just out of curiosity, do we have to put the `duration` in the `Base` model instead of the `Request` model? I was wondering if this would cause any issues since the `Base` model is used in other contexts like the `build_utils` and the `PipelineRunConfiguration`.

The `PipelineBuildBase` is used in a few places to parse yaml files of a build that users can pass when running a pipeline using the CLI. Most likely, the yaml file that they input there is the output from `zenml pipeline build` which I think will include the duration. So I put it into the base just to be safe and because I don't see any downsides to it.

PR #3387, opened 2025-03-03
Implement entrypoint typing

simonpernas:
## Describe changes
I implemented type hints for the entrypoint of functions decorated with `@step`. With my changes, `some_step.entrypoint` will have the same typing information as the undecorated `some_step` function. This is meant to help from a user point of view.

I discussed briefly with Michael Schuster on Slack, because I do not have a proper testing environment for ZenML. He advised me to create a PR nonetheless, but I fear there may be issues that I missed because I could not run all the tests. 

## Pre-requisites

## Types of changes


src/zenml/steps/base_step.py
```diff
@@ -211,20 +227,8 @@ def __init__(
...
-            *args: Positional arguments passed to the step.
-            **kwargs: Keyword arguments passed to the step.
-
-        Returns:
-            The output of the step.
-        """
-
     @classmethod
-    def load_from_source(cls, source: Union[Source, str]) -> "BaseStep":
+    def load_from_source(cls, source: Union[Source, str]) -> "BaseStep[F]":
```
schustmi, line 231: I think this should rather return `BaseStep[Callable[..., Any]]` which at least for me solved some type issues. As a more general question: Shouldn't we always use `BaseStep[Callable[..., Any]]` instead of `BaseStep[Any]` or is this already covered because `F` is bound to `Callable[..., Any]` anyway?

src/zenml/steps/base_step.py
```diff
@@ -91,10 +93,24 @@
 
 logger = get_logger(__name__)
 
-T = TypeVar("T", bound="BaseStep")
+T = TypeVar("T", bound="BaseStep[Any]")
+F = TypeVar("F", bound=Callable[..., Any])
 
 
-class BaseStep:
+class _AbstractEntrypoint(Protocol[F]):
```
schustmi, line 100: I'm guessing this was necessary to annotate the entrypoint? Do you think we could improve that even further by using https://docs.python.org/3/library/typing.html#typing.ParamSpec?

Without any testing, I was thinking of something like this:
```python
P = ParamSpec("P")
R = TypeVar("R")
F = TypeVar("F", bound=Callable[P, R])

class BaseStep(Generic[F]):
    @abstractmethod
    def entrypoint(self, *args: P.args, **kwargs: P.kwargs) -> R:
        pass
```

simonpernas: I did not think to use ParamSpec, but indeed it works here. Although not in the way you propose, because a TypeVar cannot be bound with another TypeVar (as far as I understand). But I think I came up with another cleaner solution that also addresses both your comments. Let me just verify that it actually works and passes MyPy tests

src/zenml/steps/base_step.py
simonpernas: With the version you commented on, if we tried to type a BaseStep with something that is not a Callable, like `BaseStep[int]`, MyPy would report an error like this:
 `Value of type variable "F" of "BaseStep" cannot be "int"  [type-var]`. 

But it may still better to write `BaseStep[Callable[..., Any]]`, at least for clarity. And anyway in the new version I get rid of this `F` entirely, and instead I use a `ParamSpec` as you suggested, giving typing like `BaseStep[P, R]`, or if we don't care about the precise typing, `BaseStep[..., Any]`

schustmi:
@simonpernas The CI is still failing, other than that looks good to me. Thanks for the PR!

schustmi:
@simonpernas I think the CI failures are due to using `from __future__ import annotations`. ZenML currently doesn't work with these future string annotations yet and requires the real classes. Until we rework that and add support for it, do you think there is a way to revert this change and make it work without this import?

simonpernas:
If I trust what I wrote in my commits, I added these future annotations because some of what I did does not work without them in Python 3.9, which the CI checks. 
I really can't answer for sure right now, but I will check and come back when/if I find a solution.

htahir1:
Closing due to lack of activity

PR #3404, opened 2025-03-13
Adding removing tags with various update models

bcdurak:
## Describe changes

Some of our update models such as the `PipelineUpdate`, `ArtifactUpdate` etc. have `add_tags` and `remove_tags` fields. Normally, these fields are annotated with `List[str]`. However, the dashboard uses stringified UUID in these fields to interact with these fields.

Previously, in our SQLZenStore logic, we followed the `Fetch First - If NotFoundError Create` logic. With the latest tag changes, this switched to `Create First - If IntegrationError Fetch`. While the UUIDs were ok with the first approach, now it creates a tag with the stringified UUID as the name. 

To counter that, I have modified the logic of `get or create tag` in the SqlZenStore. Since our get_tag endpoint also works with a `tag_name_or_id`, I made it impossible for people to create tags where the name is a UUID.

## Pre-requisites
- [X] **IMPORTANT**: I made sure that my changes are reflected properly in the following resources:
  - [X] Dashboard: Needs to be communicated to the frontend team.

## Types of changes


stefannica (approved):

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -11306,6 +11306,21 @@ def _attach_tags_to_resources(
 
         tag_schemas = []
         for tag in tags:
+            # Check if the tag is a string that can be converted to a UUID
+            if isinstance(tag, str):
+                try:
+                    tag_uuid = UUID(tag)
+                    tag_schema = self._get_schema_by_id(
```
schustmi, line 11313: This should not be inside the try-catch, who knows if it raises a random `ValueError` that you would not want to ignore.

schustmi (changes requested):

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Nice catch. I turned it into a `try-except-else` and the try clause only includes the UUID conversion.

PR #3405, opened 2025-03-13
Exclusive tag behaviour

bcdurak:
## Describe changes

After consideration, we have decided that the `exclusive` tags will only work with entities that support the functionality of `exclusive` tags.

That's why I adjusted the `update_tag` and `create_tag_resource` methods in our SQLZenStore. Now, you can not add a tag to a non-supported entity and you can not update a tag to be `exclusive` if it is already associated with an entity that is not supported.

Added some tests to cover different cases.

## Pre-requisites
- [X] **IMPORTANT**: I made sure that my changes are reflected properly in the following resources:
  - [X] Dashboard: Needs to be communicated to the frontend team.

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -11847,7 +11882,9 @@ def _create_tag_resource_schemas(
                                         )
                                     )
                     else:
-                        logger.debug(
+                        raise IllegalOperationError(
```
stefannica, line 11885: `IllegalOperationError` is used for authorization errors. This feels more like a `ValueError`.

stefannica: Could you also use `ValueError` instead of  `IllegalOperationError` in `update_tag` please ?

stefannica (changes requested):

stefannica (approved):

PR #3407, opened 2025-03-13
Rename workspace to project

stefannica:
## Describe changes
This PR is the final phase in implementing ZenML projects: the big rename from workspace to project and introducing "cute names for projects".

Renaming changes:

* renames the `workspace` DB table to `project`
* renames all `workspace_id` foreign key relationships to `project_id` in the DB
* also renames unique DB constraints
* implements a big DB migration script for all these changes
* renames `workspace` to `project` everywhere else: the domain models, the Zen stores, the Client, the CLI and the REST API endpoints
* keeps the `workspace` endpoints to preserve backwards-compatibility for the dashboard 

"Cute names" for projects:

* add "display name" to projects and populate it with the current "name" values
* add additional validation rules for the "name" field to allow it to be used as a short ID (slug) 

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/cbc6acd71f92_add_workspace_display_name.py
```diff
@@ -0,0 +1,58 @@
...
+    meta = sa.MetaData()
+    meta.reflect(bind=connection, only=("workspace",))
+    workspace_table = sa.Table("workspace", meta)
+
+    for workspace_id, workspace_name in connection.execute(
+        sa.select(
+            workspace_table.c.id,
+            workspace_table.c.name,
+        )
+    ):
```
schustmi, line 42: Should we also ensure that existing workspace names are lowercase and have no special characters?

src/zenml/zen_stores/migrations/versions/12eff0206201_rename_workspace_to_project.py
```diff
@@ -0,0 +1,768 @@
...
+        batch_op.create_foreign_key(
+            "fk_trigger_project_id_project",
+            "project",
+            ["project_id"],
+            ["id"],
+            ondelete="CASCADE",
+        )
+        batch_op.drop_column("workspace_id")
+
+    with op.batch_alter_table("user", schema=None) as batch_op:
```
schustmi, line 753: We drop this but don't seem to add a `default_project_id` as replacement?

src/zenml/models/v2/core/project.py
```diff
@@ -0,0 +1,203 @@
...
+            data: The values to validate.
+
+        Returns:
+            The validated values.
+        """
+        name = data.get("name")
+        display_name = data.get("display_name")
+
+        if not name and not display_name:
+            return data
```
schustmi, line 74: This will trigger a validation error which i guess is intended?

src/zenml/models/v2/core/project.py
```diff
@@ -0,0 +1,203 @@
...
+
+        if not name:
+            assert display_name
+
+            tenant_name = display_name.lower().replace(" ", "-")
```
schustmi, line 79: `project_name`

schustmi (changes requested):

src/zenml/zen_stores/migrations/versions/cbc6acd71f92_add_workspace_display_name.py
stefannica: This has a near-zero chance of happening. I already blocked the DB upgrade in the previous PR if there is more than one workspace present. And it's very improbable that someone would actually use a weird name for their default workspace, given that it wasn't possible to manage them with the CLI and dashboard before.

src/zenml/zen_stores/migrations/versions/12eff0206201_rename_workspace_to_project.py
stefannica: You're right. I don't really need this anymore. It was meant as a temporary measure for a quicker PoC, but now we decided to go with a fully updated dashboard which supports creating/switching between workspaces, so I dropped it.

src/zenml/models/v2/core/project.py
stefannica: Yes :smile: , I just pass it to Pydantic here.

src/zenml/models/v2/core/project.py
stefannica: done

src/zenml/zen_stores/migrations/versions/12eff0206201_rename_workspace_to_project.py
schustmi: I thought we maybe wanted to even have that as an option for the user to configure in the dashboard, to essentially select which project will be active by default each time they access a workspace. But I guess when we get to that point we can always implement that

schustmi (approved):

PR #3410, opened 2025-03-13
Fixing filter models with multiple inputs

bcdurak:
## Describe changes

This PR brings some fixes and improvements for handling filters with listed attributes. 

- Adjusted the TaggableFilters to accept `tags` as `API_MULTI_INPUT_PARAMS`. Now, the API actually accepts tags as parameters.
- Created a new interface called `RunMetadataAttachedFilter` for entity types to which we can attach metadata and cleaned up the old filter models accordingly.
- Adjusted the new filter to accept `List[str]` instead of `Dict[str, Any]` in order to enable its use on the CLI and the API.
- Added a test to cover the new changes.

## Pre-requisites

## Types of changes


src/zenml/models/v2/base/filter.py
```diff
@@ -768,9 +758,6 @@ def generate_custom_query_conditions_for_column(
         filter_ = FilterGenerator(table).define_filter(
             column=column, value=value, operator=operator
         )
-        if isinstance(filter_, StrFilter):
-            filter_.json_encode_value = json_encode_value
```
schustmi, line 772: That value seems unused now?

src/zenml/models/v2/base/scoped.py
```diff
@@ -583,3 +587,131 @@ def apply_sorting(
             return query
 
         return super().apply_sorting(query=query, table=table)
+
+
+class RunMetadataAttachedFilter(BaseFilter):
```
schustmi, line 592: I prefer something like `RunMetadataFilterMixin`, but feel free to ignore and keep as is.

src/zenml/client.py
```diff
@@ -3896,7 +3896,7 @@ def list_pipeline_runs(
         tag: Optional[str] = None,
         tags: Optional[List[str]] = None,
         user: Optional[Union[UUID, str]] = None,
-        run_metadata: Optional[Dict[str, Any]] = None,
+        run_metadata: Optional[List[str]] = None,
```
schustmi, line 3899: Does this change also require some updates to the docs?

src/zenml/models/v2/base/filter.py
```diff
@@ -248,31 +246,23 @@ def generate_query_conditions_from_column(self, column: Any) -> Any:
...
-        # Don't convert the value to a json string if the operation is contains
-        # because the quotes around strings will mess with the comparison
         if self.operation == GenericFilterOps.CONTAINS:
             return column.like(f"%{self.value}%")
-
-        json_value = (
-            json.dumps(self.value) if self.json_encode_value else self.value
-        )
-
         if self.operation == GenericFilterOps.STARTSWITH:
```
schustmi, line 253: Now thinking about this, if the column does not contain json encoded values, this would simply return false results, correct?

If this is simply a string column and I pass in `startswith:a`, it would give me results I do not expect. Maybe it is worth having this as a special handling only for certain columns, not the generic string filter.

src/zenml/models/v2/base/filter.py
```diff
@@ -248,31 +246,23 @@ def generate_query_conditions_from_column(self, column: Any) -> Any:
...
         if self.operation == GenericFilterOps.ONEOF:
             assert isinstance(self.value, list)
-            # Convert the list of values to a list of json strings
-            json_list = (
-                [json.dumps(v) for v in self.value]
-                if self.json_encode_value
-                else self.value
-            )
-            return column.in_(json_list)
+            return column.in_(self.value)
```
schustmi, line 249: `oneof` IMO should have the same logic as `equals`, just for multiple entries. The logic for `equals` however changed as part of this PR, so I think it should also be applied here?

schustmi (changes requested):

src/zenml/models/v2/base/filter.py
bcdurak: Yes, I forgot to remove it from the signature. Fixed it now.

src/zenml/models/v2/base/scoped.py
bcdurak: Not a bad idea, renamed it accordingly.

src/zenml/client.py
bcdurak: We didn't have any docs specifically pointing out how to filter by run_metadata. Since we are will be merging the pages really soon anyhow, I will handle that in an upcoming PR.

src/zenml/models/v2/base/filter.py
bcdurak: What do you mean by false results? 

A great example that I can think of, in terms of strings columns that do not contain json encoded values, is the stack name. If I do `zenml stack list --name=startswith:def`, it still returns the `default` stack. The potential problem is, if we ever create a stack with the name `"default"`, that would get return as well. Is that what you mean?

src/zenml/models/v2/base/filter.py
bcdurak: So, are you saying that we should add an `equals` condition (that we used down below) for each entry in the list and then combine them with an `or`?

src/zenml/models/v2/base/filter.py
schustmi: Yep exactly

src/zenml/models/v2/base/filter.py
schustmi: Yes

src/zenml/models/v2/base/filter.py
bcdurak: The only column that I can think of (where we have this type and value in separate columns, and every value is JSON encoded) is the value column in the RunMetadataSchema. Can you think of anything else?

Also, if I separate this case (and only add the `or_` conditions if we are dealing with that column), this would fix it for most filters. However, we will still run into the same issue with the run metadata filtering. At this moment in the code, all I am getting from the user is a string. If I ran two pipelines with `log_metadata("random_key": 1)` and `log_metadata("random_key": "1")` and people filter by `run_metadata=["random_key:startswith:1"]`, they will get both. Is that fine for you?

schustmi: No can't think of any other column, which makes this even more "important": We're essentially implementing the `StrFilter` (which is used for 100s of columns) to work correctly with 1 column instead of the 99 others.

I think the second case with `"1"` and `1` in this column, there isn't any quick solution to it so I think it should be fine

bcdurak: I have addressed these issues in my latest commit.

src/zenml/models/v2/base/filter.py
bcdurak: Adressed this as well.

schustmi (approved):

PR #3421, opened 2025-03-17
Pass API token as kubernetes secret

schustmi:
## Describe changes
Add option to pass the ZenML server API token as a Kubernetes secret instead of plain environment variable.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
@@ -130,11 +130,17 @@ def main() -> None:
 
         if orchestrator.config.pass_zenml_token_as_secret:
             env.pop("ZENML_STORE_API_TOKEN")
-            secret_name = orchestrator_run_id
-            pod_settings.env_from.append(
+            secret_name = orchestrator.get_token_secret_name(
+                deployment_config.id
+            )
+            pod_settings.env.append(
                 {
-                    "secretRef": {
-                        "name": secret_name,
+                    "name": "ZENML_STORE_API_TOKEN",
+                    "valueFrom": {
+                        "secretKeyRef": {
+                            "name": secret_name,
+                            "key": "token",
+                        }
                     },
                 }
             )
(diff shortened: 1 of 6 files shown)


schustmi:
@stefannica Are you okay with this approach in general? If yes I'd clean this up a little and write some docs for it

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -174,6 +192,20 @@ def run_step_on_kubernetes(step_name: str) -> None:
 
     logger.info("Orchestration pod completed.")
 
+    if (
```
schustmi, line 195: I only run this cleanup in case the orchestration pod exited succesfully. This is in order to prevent accidently deleting the secret while a step pod is still running, but will lead to some secrets being left over if the pipeline failed. @stefannica what do you think about this?

stefannica: If I'm reading the `ThreadedDagRunner` implementation correctly, this code is only reached when all of the pods have completed running (i.e. all of the threads have been "joined"). So you don't need this condition at all.

schustmi: Which condition? I'm not explicitly checking for success anywhere, but implicitly by putting the code where it is right now

stefannica: Oh, I see... then why don't you just use a try/finally around `ThreadedDagRunner.run()` ?

schustmi: That is precisely my question: This would handle also the case where a step is failing. There is however one other option that I've also seen sometimes: The dag runner fails with some random error, even though the step is actually still running. In that case, we would remove the secret already. Now that I'm typing it however, I realize that at this point the token would already be set as environment variable, which means we can delete it without disturbing the (potentially still active) step. I'll adjust it.

PR #3449, opened 2025-03-20
Stop CLI profiler running so much

strickvl:
This small update cancels running workflows of the CLI profiler on the same branch when you push new changes


.github/workflows/performance-profiling.yml
```diff
@@ -4,6 +4,10 @@ on:
   pull_request:
     types: [opened, synchronize, ready_for_review]
   workflow_dispatch:
+concurrency:
```
schustmi, line 7: Can you please also add a filter on which files trigger this workflow? Only files in `src/zenml` should be relevant I assume?

htahir1 (approved):

PR #3472, opened 2025-03-28
Upgrading pydantic to 2.11

bcdurak:
## Describe changes
| Command | develop Time (s) | feature/upgrading-pydantic-to-2.11 Time (s) | Difference | Status |
|---------|-------------------|-------------------|------------|--------|
| `zenml --help` | 3.007808 ± 0.012083 | 1.686373 ± 0.013748 | -1.321s | ✅ Improved |
| `zenml model list` | 4.211289 ± 0.025671 | 2.718185 ± 0.044351 | -1.493s | ✅ Improved |
| `zenml pipeline list` | 4.135158 ± 0.027532 | 2.768479 ± 0.028478 | -1.367s | ✅ Improved |
| `zenml stack --help` | 3.028235 ± 0.037987 | 1.722361 ± 0.051498 | -1.306s | ✅ Improved |
| `zenml stack list` | 5.056318 ± 2.079630 | 2.759821 ± 0.037323 | -2.296s | ✅ No longer slow |

This PR removes all `mlserver` related requirements from the MLflow integration because these are dragging in an older version of Pydantic. They are modeled as optional requirements going forward, that the user must install manually if needed.

🎤 💧 
## Pre-requisites

## Types of changes


.github/workflows/performance-profiling.yml
```diff
@@ -3,7 +3,7 @@ name: ZenML CLI Performance Profiling
 on:
   pull_request:
     types: [opened, synchronize, ready_for_review]
-    paths: [src/zenml]
+    paths-ignore: [docs/**, '**.md']
```
htahir1, line 6: Why? Wouldnt it make sense to only do it for src?

htahir1 (approved):
Just some questions otherwise LGTM if ur sure

.github/workflows/performance-profiling.yml
```diff
@@ -3,7 +3,7 @@ name: ZenML CLI Performance Profiling
 on:
   pull_request:
     types: [opened, synchronize, ready_for_review]
-    paths: [src/zenml]
+    paths: [src/zenml/*, pyproject.toml]
```
stefannica, line 6: this won't cut it, you need to cover all of the subpaths (i.e. `src/zenml/**`) according to https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

stefannica (changes requested):

stefannica (approved):

pyproject.toml
```diff
@@ -51,7 +51,7 @@ gitpython = "^3.1.18"
 packaging = ">=24.1"
 passlib = { extras = ["bcrypt"], version = "~1.7.4" }
 psutil = ">=5.0.0"
-pydantic = "~2.8"
+pydantic = "~2.11"
```
htahir1, line 54: can we do `>=2.x, <3`? @stefannica @bcdurak

stefannica: looking at the amount of manual fixes that Baris had to make to allow 2.11 to work, I would say no

bcdurak: Most of the changes are coming from linting, however, if you take a look at `models/__init__.py`, we had some changes related to the pydantic upgrade as well. I am not sure why this was not failing before, but if we would relax this 2 weeks ago, the release of pydantic 2.11 would make `zenml` fail completely.

.github/workflows/performance-profiling.yml
bcdurak: Fixed it. 😄

pyproject.toml
htahir1: @bcdurak can you try this out? it would be a big headache solver

stefannica: I'll try this out

stefannica: I had to remove this `mlserver` requirement from the mlflow integration. It's the only thing that was interfering with us relaxing the Pydantic version range, because it was explicitly installing an older version of Pydantic.

htahir1: But what impact does that have on mlflow examples @stefannica ?

stefannica: @htahir1 absolutely no impact on any of our examples. This `mlserver` feature is turned off by default and none of our examples are setting it explicitly. In fact, if you search `mlserver` in all of the zenml-io GitHub org, you'll not find one mention of it in a demo, example or even notebook.

htahir1: @stefannica hmm wonder why we even have this then. im just a bit concerned about it , maybe it isn't mentioned explicitly but used in the e2e example?

stefannica: `mlserver` is just one flavor of a local mlflow model server. The default one works just as well. I think this feature was never really used, even though it was available.

htahir1: ok then!

PR #3482, opened 2025-03-31
Add service annotations to helm chart

jsybel:
## Describe changes
Hello :wave: I added the option to add `annotations` in the Service Resource.
I tested it locally by using the helm template command.

## Pre-requisites
- [x] **IMPORTANT**: I made sure that my changes are reflected properly in the following resources: I don't think updates are needed to the docs - please let me know if I am mistaken!

## Types of changes

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -992,6 +992,7 @@ zenml:
   service:
     type: ClusterIP
     port: 80
+    annotations:
 
   ingress:
     enabled: true
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -992,7 +992,7 @@ zenml:
   service:
     type: ClusterIP
     port: 80
-    annotations:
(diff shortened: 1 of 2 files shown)


helm/values.yaml
```diff
@@ -992,6 +992,7 @@ zenml:
   service:
     type: ClusterIP
     port: 80
+    annotations:
```
stefannica, line 995:
```suggestion
    annotations: {}
```

I'm worried that a null default value might result in an error if this is not set.

stefannica (changes requested):
Thanks! Looks mostly good, with one exception: the default value should not be none.

helm/values.yaml
jsybel: Thanks @stefannica I have added your suggestion!

stefannica (approved):

PR #3497, opened 2025-04-04
Update step status if Kubernetes Orchestrator step pods fail

stefannica:
## Describe changes
The Kubernetes Orchestrator doesn't handle cases in which step pods are failing to start or crashing mid-way. As a result, the pipeline run is forever stuck in an "in progress" or "initializing" state.

This PR fixes that by making the following global modifications:

* passing the placeholder pipeline run to the orchestrator's `prepare_or_run_pipeline` function, which enables the orchestrator to extract the pipeline run ID
* rework the `ThreadedDagRunner` to keep track of failed steps and to call an optional finalizer when the pipeline run is complete

The Kubernetes Orchestrator uses these modifications to handle pod failures:

* passes the run ID to the orchestrator pod as an entrypoint arg, which uses it to update the pipeline run / steps status
* implements a `ThreadedDagRunner` finalizer that looks at all failed steps and updates the pipeline run / steps status if needed 

In addition to the above, new `KubernetesOrchestratorSettings` have been introduced to better control retries and timeouts:

* `pod_startup_timeout`: The maximum time to wait for a `Pending` step pod to start (in seconds). 
* `pod_failure_max_retries`: The maximum number of times to retry a step pod if the step Kubernetes pod fails to start
* `pod_failure_retry_delay`: The delay in seconds between pod failure retries and pod startup retries (in seconds)
* `pod_failure_backoff`: The backoff factor for pod failure retries and pod startup retries.

## Pre-requisites

## Types of changes


avishniakov:
@stefannica thanks a lot! I cannot properly test it atm, since I don't have a test server at hand, but I wanted to clarify 2 moments, before you merge this:
- Will it help with `retry` functionality? Can I configure the step to retry X times, if failed with this new functionality? Will it restart, if some Quota issues occurred or it was Evicted?
- How it will handle the long `Pending` pods? Maybe a wait timeout for pod to start shall be in place as well?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -162,24 +172,96 @@ def run_step_on_kubernetes(step_name: str) -> None:
...
+        # Fetch the pipeline run using any means possible.
+        list_args: Dict[str, Any] = {}
+        if args.run_id:
+            # For a run triggered outside of a schedule, we can use the
+            # placeholder run ID to find the pipeline run.
+            list_args = dict(id=UUID(args.run_id))
+        else:
+            # For a run triggered by a schedule, we can only use the
+            # orchestrator run ID to find the pipeline run.
+            list_args = dict(orchestrator_run_id=args.run_name)
```
schustmi, line 225: I'm not sure this `args.run_name` is the correct value to use. As far as I can see, it might get sanitized before being used as the pod name later and will therefore not always be the `orchestrator_run_id`.

However, this entrypoint file is actually responsible for setting the orchestrator run ID, and we currently use `orchestrator_run_id = socket.gethostname()` to do so. So I think it's best if we use that exact value again here to make sure this always works.

Orchestrator run ID in combination with the deployment ID is actually also covered by a unique constraint in the DB, so in theory we don't even need the passing of the placeholder run as this will work also in non-scheduled cases.

schustmi (changes requested):

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
stefannica: I tried to use `orchestrator_run_id` in one of my earlier attempts, but the `orchestrator_run_id` is not set for a placeholder run (unless at least one step has already run ??)

stefannica: I switched to using `orchestrator_run_id`

stefannica:
> @stefannica thanks a lot! I cannot properly test it atm, since I don't have a test server at hand, but I wanted to clarify 2 moments, before you merge this:
> 
> * Will it help with `retry` functionality? Can I configure the step to retry X times, if failed with this new functionality? Will it restart, if some Quota issues occurred or it was Evicted?
> * How it will handle the long `Pending` pods? Maybe a wait timeout for pod to start shall be in place as well?

@avishniakov I added configuration options for both these cases.

PR #3509, opened 2025-04-08
Fix security check for poetry

bcdurak:
## Describe changes
Solves the security issue:

> Issue: [B605:start_process_with_a_shell] Starting a process with a shell, possible injection detected, security issue.
   Severity: High   Confidence: High

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/integration.py
+++ b/src/zenml/cli/integration.py
@@ -14,6 +14,7 @@
 """Functionality to install or uninstall ZenML integrations via the CLI."""
 
 import os
+import subprocess
 import sys
 from typing import Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


schustmi (approved):

src/zenml/cli/integration.py
```diff
@@ -233,9 +234,9 @@ def export_requirements(
                 "and try again."
             )
         else:
-            formatted_reqs = " ".join([f'"{req}"' for req in requirements])
-            cmd = f"poetry add {formatted_reqs}"
-            os.system(cmd)
+            # Use subprocess.run with shell=False to avoid command injection
+            args = ["poetry", "add"] + requirements
```
schustmi, line 238: Shouldn't this still do the escaping?

bcdurak: No. Unlike os.system() or a subprocess.run("..", shell=True) (with a string input), we are using a list here with subprocess.run([..]). In such cases, each argument is passed directly without going through a shell. So, we would not have to handle any escaping here.

PR #3532, opened 2025-04-10
Update NumPy integration to work with 1.x and 2.x versions of the library

htahir1:
## Describe changes
Update NumPy version compatibility handling

## Pre-requisites

## Types of changes


strickvl:
Other places need testing, too:

These services use NumPy for array handling:
- `src/zenml/integrations/mlflow/services/mlflow_deployment.py`
- `src/zenml/integrations/databricks/services/databricks_deployment.py`
- `src/zenml/integrations/seldon/services/seldon_deployment.py`

And two other issues that were flagged based off the migration document:

### 1. Data Type Promotion Changes
NumPy 2.0 changes data type promotion rules as per NEP 50. The key changes:
- Scalar precision is now preserved consistently
- For floating-point values, this can lead to lower precision results when working with scalars
- For integers, errors or overflows are possible

#### Solution:
- Add explicit casting where needed in the `NumpyMaterializer`, particularly in metadata extraction
- Add tests that check both NumPy 1.x and 2.0 behavior
- Consider using `item()` to convert to Python scalars where appropriate

### 2. Default Integer Type Changes
NumPy 2.0 changes the default integer to 64-bit on all 64-bit systems, which might affect serialization/deserialization.

#### Solution:
- Test the materializer with integer arrays to ensure consistent behavior
- Consider explicitly specifying integer types where necessary

htahir1:
@strickvl i made some minor adjustments. the three integrations i don't really think are effected or need to be changed. the extract metadata , maybe

strickvl:
Based off the migration guide, two suggestions for improved backwards compatibility:

- Consider adding support for scalar precision preservation changes:

```python
# When converting between NumPy arrays and Python scalars
# NumPy 2.0 preserves precision more strictly
def _ensure_dtype_compatibility(arr):
   if IS_NUMPY_2:
         return arr  # NumPy 2.0 already preserves precision
   else:
         # For 1.x, explicitly preserve precision when needed
         return arr.astype(arr.dtype, copy=False)
```

- Add handling for the copy parameter changes in asarray and array creation:

```python
# For array creation functions that might be used
def _create_array(data, dtype=None):
   if IS_NUMPY_2:
         return np.asarray(data, dtype=dtype)
   else:
         # In NumPy 1.x, copy behavior is different
         return np.array(data, dtype=dtype, copy=False)
```

Otherwise LGTM.

htahir1:
@strickvl done

strickvl (approved):

src/zenml/integrations/evidently/__init__.py
```diff
@@ -56,6 +56,7 @@ class EvidentlyIntegration(Integration):
     REQUIREMENTS = [
         "evidently>=0.4.16,<=0.4.22",
         "tenacity!=8.4.0",  # https://github.com/jd/tenacity/issues/471
+        "numpy<2.0.0",  # evidently is not compatible with NumPy 2.0
```
bcdurak, line 59: Doesn't evidently have this built-in to their own set of dependencies? Why do we have to explicitly limit this here?

htahir1: the latest evidently probably supports this but the versions we have pinned here do not apparently pin numpy @bcdurak

bcdurak:
@htahir1 Do not merge this yet. Due to evidently and azure, the testing suite was still using Numpy<2.0. I am testing it with Numpy 2.2.5 and the materializer is not working.

PR #3539, opened 2025-04-14
Add Pandas custom data type error handling and logging

htahir1:
## Describe changes

This PR enhances the `PandasMaterializer` to better handle custom pandas data types. It adds detection, warning, and metadata collection for non-standard data types in DataFrames.

raises this error:

image

Tested with this pipeline:

```
import pandas as pd
import numpy as np
from datetime import datetime, timedelta
# import db_dtypes  # Import for DB date types

from zenml.config import DockerSettings
import site
import sys
from pathlib import Path
from typing import Any
import zenml
from zenml.config import DockerSettings


def get_docker_settings(skip_parent_build: bool = False, **kwargs: Any) -> DockerSettings:
    settings_kwargs = {
        "build_config": {
            "build_options": {
                "platform": "linux/amd64",
            }
        },
        "python_package_installer": "uv",
    }

    if not skip_parent_build:
        for path in site.getsitepackages() + [site.getusersitepackages()]:
            if Path(path).resolve() in Path(zenml.__file__).resolve().parents:
                # Not an editable install
                break
        else:
            # Editable install, add parent build for local files
            zenml_git_root = Path(zenml.__file__).parents[2]
            settings_kwargs.update(
                {
                    "dockerfile": str(zenml_git_root / "docker" / "zenml-dev.Dockerfile"),
                    "build_context_root": str(zenml_git_root),
                    "parent_image_build_config": {
                        "build_options": {
                            "platform": "linux/amd64",
                            "buildargs": {
                                "PYTHON_VERSION": f"{sys.version_info.major}.{sys.version_info.minor}"
                            },
                        }
                    },
                }
            )

    settings_kwargs.update(kwargs)
    return DockerSettings(**settings_kwargs)


@step(settings={"docker": get_docker_settings(requirements=["db_dtypes", "pandas", "numpy"], python_package_installer="uv")})
def generate_dummy_data() -> Annotated[pd.DataFrame, "dummy_data"]:
    """Generate dummy data with various data types including dbdate."""
    import db_dtypes
    # Create a base date and generate a range of dates
    base_date = datetime(2023, 1, 1)
    dates = [base_date + timedelta(days=i) for i in range(100)]
    
    # Generate random numerical data
    numeric_data = np.random.randn(100, 3)
    
    # Create categories
    categories = np.random.choice(['A', 'B', 'C'], size=100)
    
    # Create the DataFrame
    df = pd.DataFrame({
        'date': dates,                           # standard datetime column
        'value1': numeric_data[:, 0],            # float column
        'value2': numeric_data[:, 1],            # float column
        'value3': numeric_data[:, 2],            # float column
        'category': categories,                  # categorical column
        'integer': np.random.randint(0, 100, 100)  # integer column
    })
    
    # Convert the standard date column to dbdate explicitly
    df['db_date'] = pd.Series(dates, dtype='dbdate')
    
    print(f"Generated DataFrame with shape: {df.shape}")
    print(f"Data types: {df.dtypes}")
    
    return df


@step(settings={"docker": get_docker_settings(requirements=["pyarrow", "pandas", "numpy"], python_package_installer="uv")})
def process_data(df: pd.DataFrame) -> Annotated[pd.DataFrame, "processed_data"]:
    """Process the dummy data to demonstrate data manipulation with dbdate."""
    # Convert dbdate to datetime64 to use datetime accessors
    datetime_series = df['db_date'].astype('datetime64[ns]')
    
    # Add columns based on converted datetime
    df['year'] = datetime_series.dt.year
    df['month'] = datetime_series.dt.month
    df['day_of_week'] = datetime_series.dt.day_name()
    
    # Some simple aggregations
    grouped = df.groupby('month').agg({
        'value1': 'mean',
        'value2': 'sum',
        'integer': 'max'
    }).reset_index()
    
    print(f"Processed data shape: {grouped.shape}")
    return grouped


@pipeline(enable_cache=False)
def dummy_data_pipeline():
    """Pipeline that generates and processes dummy data including dbdate values."""
    data = generate_dummy_data()
    processed = process_data(data)
    return processed


if __name__ == "__main__":
    # Run the pipeline
    result = dummy_data_pipeline()
    
    # Fetch the last run
    
    pipeline = Client().get_pipeline("dummy_data_pipeline")
    last_run = pipeline.last_run

```

## Background
A user reported an issue where ZenML failed to load DataFrame artifacts in subsequent pipeline steps when the DataFrame contained custom data types from BigQuery (specifically `db_dtypes`). The user had to add `import db_dtypes` to their code even though they weren't explicitly using it.

The issue occurs because:
1. When data is loaded from BigQuery, custom data types are created
2. These are serialized to parquet by ZenML
3. When deserializing in later steps, the custom type libraries aren't loaded

## Changes
* Added detection of non-standard data types during serialization
* Enhanced error messages when custom type errors occur during deserialization
* Added warnings that suggest possible missing dependencies
* Included custom type information in artifact metadata
* Improved internal DataFrame handling for consistent type checking

## Testing
This change should be tested with DataFrames containing various custom data types, particularly those from:
- BigQuery (db_dtypes)
- GeoPandas
- Arrow
- Other pandas extension types

## Additional Notes
This is a non-breaking change that should improve the user experience when working with various data types in pandas, particularly when integrating with external data sources like BigQuery that use custom types.

## Pre-requisites

## Types of changes


src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -137,9 +194,30 @@ def save(self, df: Union[pd.DataFrame, pd.Series]) -> None:
...
+        try:
+            custom_dtypes = []
+
+            for col, dtype in df.dtypes.items():
+                dtype_str = str(dtype)
+                if not is_standard_dtype(dtype_str):
+                    custom_dtypes.append(f"{col}: {dtype_str}")
+
+            if custom_dtypes:
+                logger.warning(
```
schustmi, line 211: Not a fan of this. This logs a warning for users, even if they're well aware (as they should be) that they need these libraries in all places where they try to load the object.

src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -90,23 +121,49 @@ def load(self, data_type: Type[Any]) -> Union[pd.DataFrame, pd.Series]:
...
-                    "as a `.parquet` file, which requires `pyarrow` "
-                    "for reading, You can install `pyarrow` by running "
-                    "'`pip install pyarrow fastparquet`'."
+                with self.artifact_store.open(self.csv_path, mode="rb") as f:
+                    df = pd.read_csv(f, index_col=0, parse_dates=True)
+        except Exception as e:
+            # Check for common data type error patterns
+            error_str = str(e).lower()
+
+            is_dtype_error = (
```
schustmi, line 147: This is very flaky. The only reason that I can see to even have this code is that the pandas error is too bad, so we have to somehow try to improve it in our library?

src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -204,14 +282,46 @@ def extract_metadata(
...
+        try:
+            for col, dtype in df.dtypes.items():
+                dtype_str = str(dtype)
+                if not is_standard_dtype(dtype_str):
+                    col_name = "series" if is_series else str(col)
+                    custom_types[col_name] = dtype_str
+                    # Try to get module information if available
+                    try:
+                        module_name = dtype.type.__module__.split(".")[0]
+                        custom_types[f"{col_name}_module"] = module_name
```
schustmi, line 309: This is trying to extract the package, not the actual module in which the type is stored, correct?

schustmi (changes requested):

src/zenml/integrations/pandas/materializers/pandas_materializer.py
htahir1: Im not sure - we had the case where people actually did this without knowing

src/zenml/integrations/pandas/materializers/pandas_materializer.py
htahir1: its trying to extract the module_name

src/zenml/integrations/pandas/materializers/pandas_materializer.py
htahir1: hmm what would you suggest otherwise

src/zenml/integrations/pandas/materializers/pandas_materializer.py
schustmi: Then they wouldn't be able to load that dataframe in a different notebook either, it doesn't matter whether ZenML is involved. This is simply a user problem that they're not aware of what they're doing.

If we're being honest, reading this warning when running a pipeline, everyone will simply ignore it. The place where it will most likely fail will be when trying to read this artifact in a different pipeline than where it was produced, or outside of a pipeline. And I don't think people would then check the logs of the pipeline that produced the artifact. So IMO this has next to 0 impact and can simply be removed

src/zenml/integrations/pandas/materializers/pandas_materializer.py
schustmi: Don't know, open up a PR in the pandas repo to improve the error message. Not a ZenML concern IMO

src/zenml/integrations/pandas/materializers/pandas_materializer.py
schustmi: The module name is the full `__module__` before you're splitting it and only taking the first element. And usually (not always), this first element is the PyPI package which contains the module.
So just a naming thing, if you want to call it module just take `dtype.type.__module__`

src/zenml/integrations/pandas/materializers/pandas_materializer.py
htahir1: I mean i think its still nice to have it

schustmi: It will be:
- above the stack trace. The kind of user that runs into this error and isn't able to figure it out from the pandas exception will for sure not scroll up above and look at some warnings.
- flaky, if the wording of the error message changes at all, it will simply not work anymore.
- untested, which means we will never update this at all
To me this is simply dead code/tech debt without a purpose, but if you think it's helpful feel free to merge. I don't think we need any testing of this.

schustmi (approved):

PR #3602, opened 2025-04-23
Improve run template UX

schustmi:
## Describe changes
This PR contains some UX improvements for run templates, mostly aimed towards the UX when running a template from the dashboard:
- Exclude `None` and default (=empty) values from the run config template. This makes the text editor when running a template from the UI much more useful as it only contains relevant information.
- Only include configuration options for each step that were actually configured on the step, not if the step inherited those configurations from the pipeline.
  - Previously, a setting on the pipeline would be replicated and shown in the template UI for all steps.
  - This was because we stored the step configuration in the database with the pipeline configuration already applied. This PR changes that and instead stores both the raw, unmerged step configuration as well as the merged one in the database.
  - Caveat: Similar to our client pipeline configuration methods, we consider `None` values as not set and ignore them, which means there will be no way to "unset" a value when triggering a template. It is for example not possible to remove a failure hook source that was set for the run from which the template was created.
- Add an environment variable that can be used to configure the runner parent image.

## Pre-requisites

## Types of changes


src/zenml/config/step_configurations.py
```diff
@@ -311,3 +327,50 @@ class Step(StrictBaseModel):
...
+    @classmethod
+    @before_validator_handler
+    def _autocomplete_step_config_overrides(cls, data: Any) -> Any:
+        """Autocompletes the step config overrides.
+
+        Args:
+            data: The values dict used to instantiate the model.
+
+        Returns:
+            The values dict with the step config overrides autocompleted.
```
AlexejPenner, line 342:
```suggestion
    def _add_step_config_overrides_if_missing(cls, data: Any) -> Any:
        """Adds step config overrides to the dict if it doesn't exists.
        
        This serves to ensure backwards compatibility, as this field does not exist in old steps.

        Args:
            data: The values dict used to instantiate the model.

        Returns:
            The values dict with the step config overrides added.
```

AlexejPenner (commented):
just thought this was a bit confusing

stefannica (approved):

stefannica (approved):

PR #3606, opened 2025-04-25
Max parallelism option for kubernetes orchestrator

schustmi:
## Describe changes
This PR adds a setting for the kubernetes orchestrator that allows users to configure the maximum amount of parallel steps executing at any time.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/orchestrators/kubernetes.md
+++ b/docs/book/component-guide/orchestrators/kubernetes.md
@@ -130,6 +130,9 @@ The Kubernetes orchestrator will by default use a Kubernetes namespace called `z
 * `pod_failure_retry_delay`: The delay (in seconds) between retries to create a step pod that fails to start.
 * `pod_failure_max_retries`: The maximum number of retries to create a step pod that fails to start.
 * `pod_failure_backoff`: The backoff factor to use for retrying to create a step pod that fails to start.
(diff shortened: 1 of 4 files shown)


stefannica (approved):
It's a bit unclear how `parallel_node_startup_waiting_period` plays along with `max_parallelism` now, but the implementation looks good.

schustmi:
> It's a bit unclear how `parallel_node_startup_waiting_period` plays along with `max_parallelism` now, but the implementation looks good.

@stefannica Yes the `parallel_node_startup_waiting_period` is implemented a little strangely anyway 😅 Once a step finishes, it sleeps for this period between each parallel step that can now be run. But if two steps finish at the same time, multiple steps can still startup at exactly the same time anyway.

I think in general, the Threaded DAG runner class could be reworked to be much nicer and maybe also use less threads which are quite expensive and don't actually need to be active for the entire duration of a step run. But i'll leave that for another time :)

I have one remaining question for you regarding this PR: Do you think this should be
- a config option, which means it can only be specified when registering/updating the orchestrator, and therefore requires specific RBAC permissions
- a setting (which I currently implemented). I guess the upside of this is that you might want to have different values for this depending on the pipeline that you're running? But the downside is that any DS writing a pipeline could in theory set this to unlimited and crash the kubernetes cluster even though it was set up with a reasonable maximum value by the user that created the stack component.

PR #3627, opened 2025-04-30
Limit max concurrent template runs

schustmi:
## Describe changes
This PR limits the amount of concurrent template runs.

## Pre-requisites

## Types of changes


schustmi:
@stefannica I've changed this now to handle all the run template requests in a sequential manner with a configurable amount of worker threads. These threads are however additional threads on top of the ones that are created/used by fastapi. Do you think this is a good idea or should I try to get the worker threads from the same pool, while still somehow limiting the concurrency?

src/zenml/zen_server/template_execution/utils.py
```diff
@@ -241,13 +245,10 @@ def _task_with_analytics_and_error_handling() -> None:
...
-    if background_tasks:
-        background_tasks.add_task(_task_with_analytics_and_error_handling)
-    else:
-        # Run synchronously if no background tasks were passed. This is probably
-        # when coming from a trigger which itself is already running in the
-        # background
+    if sync:
         _task_with_analytics_and_error_handling()
+    else:
+        run_template_executor.submit(_task_with_analytics_and_error_handling)
```
stefannica, line 251: The number of threads is limited, but not the queue size. This means that you can still run into a situation where the server is flooded with requests and eventually croak due to OOM.

I recommend you also cap the queue size to match the thread pool size. This is a good way to create back-pressure and actually make the client wait (and/or eventually time out) if it's creating too many parallel run template requests instead of killing the server.

Yes, this means that FastAPI request threads will eventually be blocked waiting for the run template threads to complete. I think this is perfectly acceptable, as it gives server admins a knob that they can properly calibrate without compromising the server's reliability.

One more thing you can do is to make the submit requests cancellable: if the client times out, FastAPI also has the option to notify you that the request timed out (see https://fastapiexpert.com/blog/category/fastapi/#understanding-client-disconnection-in-fastapi). If you don't handle client disconnects for jobs like these, they will just keep on running. I.e. you can run `await request.is_disconnected()` to check if the request was cancelled before you execute anything in the worker thread.

stefannica (changes requested):

src/zenml/config/server_config.py
```diff
@@ -152,6 +153,9 @@ class ServerConfiguration(BaseModel):
             server.
         workload_manager_implementation_source: Source pointing to a class
             implementing the workload management interface.
+        max_concurrent_template_runs: The maximum number of concurrent template
+            runs that can be executed on the server. If not specified, the
+            default value of 1/4 of the thread pool size will be used.
```
stefannica, line 158:
```suggestion
        max_concurrent_template_runs: The maximum number of concurrent template
            runs that can be executed on the server.
```

The last part is no longer accurate, is it ?

src/zenml/zen_server/template_execution/utils.py
```diff
@@ -50,26 +51,85 @@
...
+        """
+        self._semaphore.acquire()
+        try:
+            future = self._executor.submit(fn, *args, **kwargs)
+        except Exception:
+            self._semaphore.release()
+            raise
+        else:
+            future.add_done_callback(lambda _: self._semaphore.release())
+            return future
```
stefannica, line 108: Why is this different than a simple try/finally ? What does `future.add_done_callback` actually do?

src/zenml/zen_server/template_execution/utils.py
```diff
@@ -50,26 +51,85 @@
...
+
+    def shutdown(self, **kwargs: Any) -> None:
+        """Shutdown the executor.
+
+        Args:
+            **kwargs: Keyword arguments to pass to the shutdown method of the
+                executor.
+        """
+        self._executor.shutdown(**kwargs)
```
stefannica, line 117: This isn't called anywhere. Are you missing the cleanup/deinitialization part ?

src/zenml/zen_server/utils.py
```diff
@@ -196,6 +200,36 @@ def initialize_workload_manager() -> None:
...
+        The run template executor.
+    """
+    global _run_template_executor
+    if _run_template_executor is None:
+        raise RuntimeError("Run template executor not initialized")
+
+    return _run_template_executor
+
+
+def initialize_run_template_executor() -> None:
```
stefannica, line 219: What about a de-initialize/cleanup function?

stefannica (changes requested):

src/zenml/zen_server/utils.py
schustmi: Added this in a the fastapi shutdown event

src/zenml/zen_server/template_execution/utils.py
schustmi: Added the call in the fastapi shutdown event handler

stefannica (approved):

PR #3653, opened 2025-05-08
Publish Codespace docker image with release

safoinme:
## Describe changes
This adds Github action and Dockerfile to build and push docker image for the codespace with release. This image builds on top of zenml lates docker image and add code-server dependency

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/publish_codespace_image.yml
+++ b/.github/workflows/publish_codespace_image.yml
@@ -1,0 +1,71 @@
+---
+name: Publish ZenML Codespace Docker image
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag (leave empty for latest)'
+        required: false
+        type: string
+        default: ''
+  push:
+    tags:
+      - '*'
+    branches:
+      - main
+      - develop
+  pull_request:
+    paths:
+      - 'docker/zenml-codespace.Dockerfile'
+
+jobs:
+  publish_codespace_image:
+    name: Build and publish zenml-codespace Docker image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    env:
+      ZENML_DEBUG: 1
+      ZENML_ANALYTICS_OPT_IN: false
+      PYTHONIOENCODING: utf-8
+    steps:
+      - uses: actions/checkout@v4.2.2
+      
+      - name: Determine version
+        run: |-
+          if [[ -n "${{ inputs.version }}" ]]; then
+            VERSION="${{ inputs.version }}"
+          elif [[ "${GITHUB_REF}" == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+          else
+            CURRENT_VERSION=$(cat src/zenml/VERSION)
+            DATE=$(date +"%Y%m%d")
+            VERSION="${CURRENT_VERSION}.dev${DATE}"
+          fi
+          VERSION=$(echo $VERSION | sed 's/[^a-zA-Z0-9._-]/-/g')
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+          echo "Determined VERSION=$VERSION"
+      
+      - name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


docker/zenml-codespace.Dockerfile
```diff
@@ -0,0 +1,57 @@
+ARG PYTHON_VERSION=3.11
+ARG ZENML_VERSION=latest
+
+# Use the official ZenML image as a base
+FROM zenmldocker/zenml:latest as base
+
+# Set user to root for installations
+USER root
```
schustmi, line 8: This never gets changed back.

docker/zenml-codespace.Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# Set environment variables (as in Modal example)
+ENV ENTRYPOINTD=""
+
+# Default working directory from base image is likely /app
+WORKDIR /app
+
+# Expose the default code-server port (optional, for documentation)
+EXPOSE 8080
+
+# Default command - using dumb-init is often good practice with containers
```
schustmi, line 56: To me it feels like this entire image is just a copy of some modal example? Wouldn't it be easier to simply use their image and install zenml in it?

.github/workflows/release.yml
```diff
@@ -89,9 +89,16 @@ jobs:
     needs: wait-for-package-release
     uses: ./.github/workflows/publish_docker_image.yml
     secrets: inherit
-  publish-helm-chart:
+  publish-codespace-image:
```
schustmi, line 92: I'm not sure why this needs to happen before the helm chart necessarily? As this doesn't actually block the OSS release, I would have this at the end of the GH action somewhere, where it failing doesn't actually cause some crucial release actions to not start.

.github/workflows/publish_codespace_image.yml
```diff
@@ -0,0 +1,71 @@
+---
+name: Publish ZenML Codespace Docker image
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag (leave empty for latest)'
```
schustmi, line 7: If I leave this empty, the tag will never be latest

.github/workflows/publish_codespace_image.yml
```diff
@@ -0,0 +1,71 @@
...
+        required: false
+        type: string
+        default: ''
+  push:
+    tags:
+      - '*'
+    branches:
+      - main
+      - develop
```
schustmi, line 16: I don't think we need this action to trigger automatically anywhere. It should build the image at release, and if anything there should be another option to trigger it manually I think

.github/workflows/publish_codespace_image.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      - name: Build and push
+        uses: docker/build-push-action@v5.0.0
+        with:
+          context: .
+          file: ./docker/zenml-codespace.Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
```
schustmi, line 65: So if it's a PR event, it should just build the docker image but not push it? 🤔

.github/workflows/publish_codespace_image.yml
```diff
@@ -0,0 +1,71 @@
...
+          tags: |
+            zenmldocker/zenml-codespace:${{ env.VERSION }}
+            ${{ github.event_name != 'pull_request' && github.ref == 'refs/heads/main' && 'zenmldocker/zenml-codespace:latest' || '' }}
```
schustmi, line 68: A push to main should not overwrite the latest image, this should happen only on release.

schustmi (changes requested):

bcdurak:
@safoinme just had an idea about this. I haven't taken a deep look at this, but if this pushes a set of Docker images with every release, you can possibly add this to [the cloudbuild configuration](https://github.com/zenml-io/zenml/blob/main/release-cloudbuild.yaml).

PR #3660, opened 2025-05-12
Limit concurrent pipeline runs

schustmi:
## Describe changes
Limit the amount of concurrent pipeline runs.

Caveat: Right now, we're not able to guarantee that we detect all failed runs if they fail on the orchestrator side without notifying the ZenML server. For that reason, we only count runs that started in the last 5 minutes to count towards the concurrent run limit.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/server_config.py
+++ b/src/zenml/config/server_config.py
@@ -156,6 +156,8 @@ class ServerConfiguration(BaseModel):
             implementing the workload management interface.
         max_concurrent_template_runs: The maximum number of concurrent template
             runs that can be executed on the server.
+        max_concurrent_pipeline_runs: The maximum number of concurrent pipeline
+            runs that can be executed on the server.
         pipeline_run_auth_window: The default time window in minutes for which
             a pipeline run action is allowed to authenticate with the ZenML
             server.
@@ -297,6 +299,7 @@ class ServerConfiguration(BaseModel):
     max_concurrent_template_runs: int = (
         DEFAULT_ZENML_SERVER_MAX_CONCURRENT_TEMPLATE_RUNS
     )
+    max_concurrent_pipeline_runs: Optional[int] = None
     pipeline_run_auth_window: int = (
         DEFAULT_ZENML_SERVER_PIPELINE_RUN_AUTH_WINDOW
     )
--- a/src/zenml/zen_server/routers/runs_endpoints.py
+++ b/src/zenml/zen_server/routers/runs_endpoints.py
@@ -89,8 +89,16 @@ def check_concurrent_run_limit() -> None:
             "You have reached the maximum number of concurrent runs. "
             "Please upgrade your subscription to increase the limit. "
             "If some of your runs are incorrectly still in running state even "
-            "though they've failed already, you can manually delete them by "
-            "running `zenml pipeline runs delete <ID>`."
+            "though they've failed already, you can free up resources by: "
+            "1. Running `zenml pipeline runs list --status=running` and/or "
+            "zenml pipeline runs list --status=initializing "
+            "to find stuck runs "
+            "2. Note the IDs of runs you want to remove from the list "
+            "3. Delete each run with `zenml pipeline runs delete <RUN_ID> -y` "
+            "4. To check if this resolved the issue, run your pipeline again "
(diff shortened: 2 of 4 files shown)


src/zenml/config/server_config.py
```diff
@@ -297,6 +299,7 @@ class ServerConfiguration(BaseModel):
     max_concurrent_template_runs: int = (
         DEFAULT_ZENML_SERVER_MAX_CONCURRENT_TEMPLATE_RUNS
     )
+    max_concurrent_pipeline_runs: Optional[int] = None
```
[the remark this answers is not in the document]
schustmi: If we decide to merge this, it will be set by the cloudapi. But I'm pretty sure we will not query stigg or the cloudapi constantly here to refresh the limit.

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -71,6 +72,28 @@
...
+
+    Raises:
+        SubscriptionUpgradeRequiredError: If the limit of concurrent runs is
+            exceeded.
+    """
+    max_concurrent_runs = server_config().max_concurrent_pipeline_runs
+
+    if (
+        max_concurrent_runs
+        and zen_store().count_active_runs() >= max_concurrent_runs
```
htahir1, line 86: i hope this query is optimized! :D

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -71,6 +72,28 @@
...
+    ):
+        raise SubscriptionUpgradeRequiredError(
+            "You have reached the maximum number of concurrent runs. "
+            "Please upgrade your subscription to increase the limit. "
+            "If some of your runs are incorrectly still in running state even "
+            "though they've failed already, you can manually delete them by "
+            "running `zenml pipeline runs delete <ID>`."
```
htahir1, line 93: so there is no way to fetch the active run id or something to make this easier?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4977,6 +4977,31 @@ def delete_event_source(self, event_source_id: UUID) -> None:
...
 
+    def count_active_runs(self) -> int:
+        """Count the number of active runs.
+
+        Returns:
+            The number of active runs.
+        """
+        # Because we can not guarantee that we detect all failed runs, we only
+        # count runs that have been running for less than 5 minutes.
+        ten_minutes_ago = utc_now(tz_aware=False) - timedelta(minutes=5)
```
htahir1, line 4988:
```suggestion
        timedelta = utc_now(tz_aware=False) - timedelta(minutes=5)
```

htahir1: Can this be fooled by having the client machine change times?

src/zenml/zen_server/routers/runs_endpoints.py
schustmi: There is no "the active run", this is a generic function and not all servers are having just a single concurrent run as limit. Even so, I would never give them the command here as they will just blindly execute it anyway, this way they have to actually look for the run and actively decide which run to delete.

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -71,6 +72,28 @@
...
+        )
```
htahir1, line 94:
```suggestion
        raise SubscriptionUpgradeRequiredError(
            "You have reached the maximum number of concurrent runs. "
            "Please upgrade your subscription to increase the limit. "
            "If some of your runs are incorrectly still in running state even "
            "though they've failed already, you can free up resources by: "
            "1. Running `zenml pipeline runs list --status=running` to find stuck runs "
            "2. Note the IDs of runs you want to remove from the list "
            "3. Delete each run with `zenml pipeline runs delete <RUN_ID> -y` "
            "4. To check if this resolved the issue, run your pipeline again "
            "For more details on pipeline management, see our documentation at "
            "https://docs.zenml.io/user-guides/best-practices/keep-your-dashboard-server-clean"
        )
```

PR #3668, opened 2025-05-14
Store validated config with converted types in DB

schustmi:
## Describe changes
When creating or updating a stack component, we stored the user-provided configuration in the database as a serialized JSON string. When using the CLI to register/update, this means all values were stored as strings, which lead to some validation errors in the frontend.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -2024,7 +2024,9 @@ class Client(metaclass=ClientMetaClass):
             name=name,
             type=component_type,
             flavor=flavor,
-            configuration=configuration,
+            configuration=validated_config.model_dump(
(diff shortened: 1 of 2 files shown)


bcdurak (approved):
This looks good. Maybe, just a quick idea. I saw that you made changes to the client and the related endpoints that cover almost all of the cases where we are using this `ComponentRequest` class. Wouldn't it be better to solve this over a pydantic validator or maybe over on the `SqlZenStore` side?

schustmi:
@bcdurak On the server side, we do not have access to custom flavors, which is why this is both on the client as well as the endpoints.
This should cover all the cases I think?

PR #3681, opened 2025-05-17
Fix `DockerServiceConnector._verify` assertion error when no `resource_id` is provided

strickvl:
## Summary
- Fixed bug in DockerServiceConnector._verify where it unconditionally asserts resource_id is not None
- Added unit tests to verify both with and without resource_id cases
- This error would occur when ServiceConnector.verify attempts to perform global verification without a specific resource ID

## Details
The DockerServiceConnector._verify method was asserting that resource_id is not None, but the parent ServiceConnector.verify method may call _verify with a None resource_id when it tests global authentication. This caused assertion errors during verification when no specific resource was being verified.

The fix modifies the method to only attempt to authorize the client when a resource ID is provided, avoiding the assertion error and allowing the method to gracefully handle global verification.

### DIFF
--- a/src/zenml/service_connectors/docker_service_connector.py
+++ b/src/zenml/service_connectors/docker_service_connector.py
@@ -374,8 +374,9 @@ class DockerServiceConnector(ServiceConnector):
                 f"\nSkipping Docker connector verification."
             )
         else:
-            assert resource_id is not None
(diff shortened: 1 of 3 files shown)


strickvl:
Closing as actually it isn't practically possible to breach this, as originally thought.

PR #3689, opened 2025-05-21
Reset active project after deleting local database

schustmi:
## Describe changes
This fixes the following case:
- Use ZenML with a local SQLite DB
- Delete the DB
- Run a pipeline without running any other previous command

The ID of the previously active project will still be stored in the global config, but won't exist anymore in the newly created database. This lead to an error that the project can not be found.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/global_config.py
+++ b/src/zenml/config/global_config.py
@@ -751,13 +751,15 @@ class GlobalConfiguration(BaseModel, metaclass=GlobalConfigMetaClass):
         Returns:
             The model of the active project.
         """
-        project_id = self.get_active_project_id()
+        # Initialize the store first. This might create a local database if it
+        # doesn't exist yet, which will then refresh our active project ID.
+        _ = self.zen_store
 
         if self._active_project is not None:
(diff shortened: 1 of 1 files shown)


schustmi:
@stefannica I couldn't find a different solution, but you're welcome to try 😄 

I think this is harmless though, if you look at the code:
- `self.get_active_project_id()`: This is either in the config file, which means it will return straight away. If it isn't, it initializes the ZenStore anyway which means my solution would not mess anything up.
- `self._active_project` can never be set without an initialized ZenStore, because where would the model be coming from.
- The next line already initializes `self.zen_store` anyway. This will also sanitize the config, which means after this line returns, the active project ID will be changed to the correct one. But the one passed to `self.zen_store.get_project(...)` is wrong.

bcdurak (approved):

stefannica:
I still am not convinced this is necessary, but I probably don't know how to reproduce it properly:

image

stefannica (approved):

PR #3690, opened 2025-05-21
Performance test utilities, stats and stress test pipeline updates

stefannica:
## Describe changes

* full round-trip client-server request IDs tracking 
* updated stress test pipeline
* FastAPI middleware with extensive debug statistics: request status, duration, thread count
* SQLZenStore DB statistics
* client request statistics
* FastAPI middleware to combat client retry storms by bailing early with 429 (rate-limiting) errors
* add POST to the list of requests that can be retried by the client in case of 429 errors
* Pro API connection statistics
* allow errors to happen while fetching Kubernetes logs
* improved user activity tracking: use lock and store in DB only when needed to reduce DB access

## Pre-requisites

## Types of changes


examples/stress-test/run.py
```diff
@@ -186,9 +311,58 @@ def load_test_pipeline(
         duration: The duration of the load test in seconds.
         sleep_interval: The interval to sleep between API calls in seconds.
     """
+    (
```
schustmi, line 314:
```python
result = init_step()
load_step(duration, sleep_interval, *result)
```

I think this also works and is a little more readable?

src/zenml/config/server_config.py
```diff
@@ -252,6 +253,11 @@ class ServerConfiguration(BaseModel):
             used.
         file_download_size_limit: The maximum size of the file download in
             bytes. If not specified, the default value of 2GB will be used.
+        thread_pool_size: The size of the thread pool for handling requests. If
+            not specified, the default value of 40 will be used.
+        server_request_timeout: The timeout for server requests in seconds. If
+            not specified, the default value of 15 seconds will be used. This
```
schustmi, line 259:
```suggestion
            not specified, the default value of 20 seconds will be used. This
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -412,6 +414,81 @@ def exponential_backoff_with_jitter(
...
+class Session(SqlModelSession):
+    """Session subclass that automatically tracks duration and calling context."""
+
+    def __enter__(self) -> "Session":
+        """Enter the context manager.
+
+        Returns:
+            The SqlModel session.
+        """
+        if logger.isEnabledFor(logging.DEBUG):
```
schustmi, line 426: Should we maybe wrap this whole thing in a `try...except` to make sure it never fails? Or do we want to catch failures in here anyway?

schustmi (changes requested):

src/zenml/zen_stores/sql_zen_store.py
stefannica: Ah, good catch !

stefannica: On second thought, I see no reason why exceptions here should be expected or ignored for that matter.

schustmi (approved):

PR #3701, opened 2025-05-24
Improve error message for duplicate pipeline run names

strickvl:
## Summary
- Improve the error message when users try to run a pipeline with a duplicate `run_name` 
- Handle duplicate run name errors directly in SQLZenStore where they occur
- Provide a user-friendly error message with actionable solutions
- Fix SQLAlchemy autoflush timing issue to ensure proper error handling

## Problem
When users run a pipeline with a duplicate run name (whether from a config file, programmatically, or any other method), they get a confusing database error. The error can come in two forms:

1. A raw SQL IntegrityError (when using REST API):
```
RuntimeError: (pymysql.err.IntegrityError) (1062, "Duplicate entry 'test_run_name-6e23c0466cc4411c8b9f75f0c8a1a818' for key 'pipeline_run.unique_run_name_in_project'")
```

2. An EntityExistsError with a technical message

Additionally, there was a subtle SQLAlchemy autoflush timing issue where IntegrityErrors were being raised during `_get_reference_schema_by_id` calls instead of during the explicit `session.commit()`, preventing proper error handling.

## Solution
This PR catches IntegrityError in SQLZenStore's `_create_run` method and provides a much more helpful error message with actionable solutions. It also fixes the SQLAlchemy autoflush issue to ensure errors are properly caught and handled.

### Before (Raw SQL Error)
```
RuntimeError: (pymysql.err.IntegrityError) (1062, "Duplicate entry 'my_run-6e23c0466cc4411c8b9f75f0c8a1a818' for key 'pipeline_run.unique_run_name_in_project'")
[SQL: INSERT INTO pipeline_run ...]
```

### After (User-Friendly Error)
```
Pipeline run name 'my_run' already exists in this project. Each pipeline run must have a unique name.

To fix this, you can:
1. Use a different run name
2. Use a dynamic run name with placeholders like: "my_run_{date}_{time}"
3. Remove the run name from your configuration to auto-generate unique names

For more information on run naming, see: https://docs.zenml.io/concepts/steps_and_pipelines/yaml_configuration#run-name
```

## Changes Made
- **Enhanced error handling** in SQLZenStore's `_create_run()` method to catch IntegrityError and provide user-friendly messages
- **Fixed SQLAlchemy autoflush issue** by wrapping logs processing in `session.no_autoflush` to prevent premature IntegrityError during reference lookups
- **Replaced mocked unit test** with proper integration test that actually runs real pipelines twice with the same name
- **Made error message more generic** - removed specific mention of "config file" since run names can be set in multiple ways
- **Updated integration tests** to verify the fix works end-to-end with real pipeline execution

## Test Plan
- [x] Integration test verifies duplicate run names are properly detected when running real pipelines
- [x] Integration test confirms EntityExistsError is raised (not raw IntegrityError) 
- [x] Integration test validates clear error message is shown to users
- [x] All existing tests continue to pass
- [x] Mypy type checking passes
- [x] Formatting and linting checks pass

## Technical Details
The key fix was addressing the SQLAlchemy autoflush behavior where `_get_reference_schema_by_id` calls would trigger early database flushes, causing IntegrityErrors to be raised before the explicit `session.commit()` and thus not being caught by the try/except block. Using `session.no_autoflush` ensures the error handling works as intended.


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5700,27 +5700,47 @@ def _create_run(
...
 
         try:
             session.commit()
-        except IntegrityError:
+        except IntegrityError as e:
             # We have to rollback the failed session first in order to
             # continue using it
             session.rollback()
+
+            # Check if this is a duplicate run name error
```
schustmi: There is a better place to detect this: This `_create_run` method is called from `get_or_create_run`, and if you scroll right to the bottom of that method you can find an explanation where it mentions name duplicates. If that works, we can avoid this ugly check for the constraint in the error message

schustmi (approved):

src/zenml/pipelines/run_utils.py
```diff
@@ -87,6 +87,7 @@ def create_placeholder_run(
         tags=deployment.pipeline_configuration.tags,
         logs=logs,
     )
+
```
strickvl:
```suggestion
```

schustmi (approved):

PR #3704, opened 2025-05-27
New Pro onboarding

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -9018,9 +9018,8 @@ class SqlZenStore(BaseZenStore):
                             OnboardingStep.PRODUCTION_SETUP_COMPLETED,
                         }
                     )
-                if (
-                    OnboardingStep.THIRD_PIPELINE_RUN
-                    not in self._cached_onboarding_state
+                if OnboardingStep.THIRD_PIPELINE_RUN not in (
+                    self._cached_onboarding_state or {}
                 ):
                     onboarding_state = self.get_onboarding_state()
                     if OnboardingStep.PIPELINE_RUN in onboarding_state:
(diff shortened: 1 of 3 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8596,6 +8596,19 @@ def _update_pipeline_run_status(
...
+                ):
+                    onboarding_state = self.get_onboarding_state()
+                    if OnboardingStep.PIPELINE_RUN in onboarding_state:
+                        completed_onboarding_steps.add(
+                            OnboardingStep.SECOND_PIPELINE_RUN
+                        )
+                    if OnboardingStep.SECOND_PIPELINE_RUN in onboarding_state:
+                        completed_onboarding_steps.add(
+                            OnboardingStep.THIRD_PIPELINE_RUN
+                        )
```
stefannica, line 9033: Two questions here:

* why do you use the cached onboarding state in the first condition but not the other two
* this will add "second pipeline run" twice when the third pipeline is run

stefannica (changes requested):

src/zenml/zen_stores/sql_zen_store.py
schustmi: - If the third run already happened (=in the cached state), we know that we don't need to track it anymore. This means it is not necessary to fetch the state from the DB (`self.get_onboarding_state()`) to sync with potential other server pods.
- If the third run didn't happen yet, we first load the most recent state, and then include the second/third run if necessary. The whole thing is a set, which means there should not be any duplicate entries.

All of this can still run into race conditions and might even not track the third run if two runs finish at exactly the same time, but as this is non-critical and only for displaying progress in the UI I think that should be okay?

stefannica (approved):

PR #3712, opened 2025-05-28
Add JAX array materializer

nicholasjng:
Previously, only NumPy arrays were understood by ZenML's materializer machinery. This, however, is a problem for people using other array libraries, such as JAX, whose `jax.Array` type resembles NumPy arrays very closely, and supports its entire API.

To support more array API types, implement a materializer built on the [Python array API standard](https://data-apis.org/array-api/latest/index.html). This standardizes the array API in question, and exposes the Python module containing the respective array library's implementation in the `__array_namespace__()` method of the array.

In particular, this immediately adds native support for JAX and MLX arrays to ZenML, since those two already implement the array API standard today.

--------------------

Consider the following mini pipeline:

```python
import jax
import jax.numpy as jnp


@step
def zeros() -> jax.Array:
    return jnp.zeros(5)

@step
def compute_sum(arr: jax.Array) -> float:
    return jnp.sum(arr).item()


@pipeline(enable_cache=False)
def basic_pipeline():
    arr = zeros()
    compute_sum(arr)


if __name__ == "__main__":
    basic_pipeline()
```

Before this change, we see (along with other output):

```console
[zeros] No materializer is registered for type <class 'jaxlib._jax.ArrayImpl'>, so the default Pickle materializer was used. Pickle is not production ready and should only be used for prototyping as the artifacts cannot be loaded when running with a different Python version. Please consider implementing a custom materializer for type <class 'jaxlib._jax.ArrayImpl'> according to the instructions at https://docs.zenml.io/how-to/handle-data-artifacts/handle-custom-data-types
```

After this change (I'll remove the debug log in the code later):

```
[compute_sum] Array type <class 'jax.Array'> processed by ArrayMaterializer
```

Left TODO are a few touch-ups as per the inline todo comments, and the dynamic associated types question.

As of now, the array API integration still has a hard dependency on NumPy, due to its use of `np.load()`.
Also, I didn't take the time to implement the whole materializer interface yet, and removed NumPy 1.x support in the first prototype.

Happy for opinions and discussion.


nicholasjng:
Additional information to consider:

* This can eventually replace the existing NumPy materializer, minus the existing backwards compatibility code.
* Compressed NumPy arrays (.npz files, saved with `np.savez_compressed`) could also be supported, which creates smaller archives for integer-valued arrays.

Interestingly, this integration is unlike others previously, since it can in principle deal with any array type coming from a library that implements the array API standard mentioned above. This is however hard for ZenML, which stores the types associated with a materializer in its `ASSOCIATED_TYPES` field, which is an immutable class variable attribute.

To fully exploit the array API polymorphism would require either 

a) adding all known array API implementers to the materializer's associated types, which is what is currently shown in this PR, but kind of against the spirit of the solution in the first place (ZenML is supposed to call this materializer on anything and anyone having a `__array_namespace__()` method), or

b) materializers being dispatched not by a type match, but rather a protocol match (in this case, something like

```python
class ArrayLike(Protocol):
    def __array_namespace__(self, *, version: str | None = None) -> types.ModuleType: ...
```

I don't know enough about ZenML to judge how easy it is to achieve b), but I guess method a) can work well enough until there are complaints.

htahir1:
Wonderful PR @nicholasjng <3 Let me discuss with the team as this has some bigger questions

src/zenml/integrations/array/materializer.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class ArrayMaterializer(BaseMaterializer):
+    """Materializer for array libraries conforming to the Python Array API standard.
+
+    This includes, but is not limited to NumPy, JAX, cupy, and MLX.
+    For more information, see https://data-apis.org/array-api/latest/index.html.
+    """
+
+    # TODO: How to reference associated types without having to install all of them?
```
htahir1: Isnt the answer to this that we should have a seperate integration for each? Like a jax integration and a mx integration?

nicholasjng: Not really imo - the idea is that when we encounter a type that implements the array API, we just defer to the array type constructor (jax.Array, torch.Tensor etc.) without having to maintain N integrations for each library, meaning less work for you guys.

But in this case, I think there is currently no way to statically know all array types beforehand. I can import-or-skip all popular(ish) array frameworks (TF, torch, JAX, numpy, ...), let me know if that works for you.

htahir1: @schustmi wdyt?

schustmi: I think adjusting the materializer registry to work with these protocol types is the real solution here, but this protocol should also include code to actually save and load that object to make it independent of any specific library.

That however is not anywhere a priority for us and I'm not sure how complicated it will be to implement. Feel free to give it a try though.

As for the current solution, I'm not a fan of it. It simply abuses the integration interface by specifying a subset of requirements, and doesn't even work in a generic way but only for a few hardcoded libraries. If anything now breaks in the imports of those other libraries, this might even cause problems for saving/loading numpy artifacts.

I think in it's current state, I'd rather have a `jax` integration if anyone needs that, instead of this.

nicholasjng:
Hey, I repushed the stuff as a proper JAX integration. I'm still of the opinion that support for generics is worth having in ZenML, but this is good enough for my momentary needs.

cc @schustmi

schustmi:
@nicholasjng I agree that generic support would for sure be nice, but we'll have to change our integration system first to make it work in a nice way. Thanks for the PR, looks good to me now!

PR #3716, opened 2025-05-28
Ability to stop pipelines early + Kubernetes Orchestrator implementation

bcdurak:
This PR introduces comprehensive pipeline stopping functionality to ZenML, with particular focus on the Kubernetes orchestrator implementation. The feature enables users to gracefully or forcefully terminate running pipeline executions through both CLI and programmatic interfaces.

### Pipeline Run Stop Interface
- Added `stop_run()` method to `PipelineRunResponse` model that delegates to the orchestrator's stop implementation
- Extended `ExecutionStatus` enum with `STOPPING` and `STOPPED` states to track termination lifecycle
- Implemented server-side REST API endpoint `/runs/{run_id}/stop` with graceful termination option
- Added CLI command `zenml pipeline runs stop <run_name_or_id>` with `--graceful` flag (defaults to force stop)

### Base Orchestrator Framework
- Extended `BaseOrchestrator` with `supports_cancellation` property and abstract `stop_run()` method
- Implemented status validation to prevent stopping finished or already-stopping pipelines
- Added automatic status transition to `STOPPING` before calling concrete orchestrator implementations
- Integrated keyboard interrupt handling for synchronous orchestrators

### Stop Mechanism
The Kubernetes orchestrator implements two stop modes:

1. **Force Stop (default):** Actively terminates all running step pods associated with the pipeline run
2. **Graceful Stop (--graceful flag):** Allows the orchestrator pod and running steps to complete naturally without intervention

### Technical Implementation
- Leverages Kubernetes label selectors (`zenml-orchestrator-run-id`) to identify related step pods
- Implements pod discovery and termination with configurable grace periods (30 seconds)

### Pipeline Cancellation Detection
- Added `check_pipeline_cancellation()` function in the Kubernetes orchestrator entrypoint
- Integrated cancellation checks into the `ThreadedDagRunner` via the `check_fn` parameter


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -255,6 +271,23 @@ def run(
...
+                    self.stop_run(placeholder_run)
+                    logger.info("Pipeline run stopped successfully.")
+                except NotImplementedError:
+                    logger.warning(
+                        "Stop functionality not implemented for this orchestrator. "
+                        "The pipeline may continue running in the background."
+                    )
+                except Exception as e:
+                    logger.error(f"Failed to stop pipeline run: {e}")
+            raise
```
stefannica: I wonder whether this is a good idea or existing users might be really upset over the fact that interrupting their client with CTRL+C stops their entire pipeline. How would I even kill the client early without stopping the entire pipeline ? WDYT ?

src/zenml/cli/pipeline.py
```diff
@@ -511,6 +511,59 @@ def list_pipeline_runs(**kwargs: Any) -> None:
...
 
+@runs.command("stop")
+@click.argument("run_name_or_id", type=str, required=True)
+@click.option(
+    "--graceful",
+    "-g",
+    is_flag=True,
+    default=False,
+    help="Use graceful shutdown (default is False).",
+)
```
stefannica, line 522: In my experience, the graceful stop option is usually the default one, while the non-graceful (forceful) one needs to be asked for explicitly.

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -389,38 +391,40 @@ def refresh_run_status(
...
+    Args:
+        run_id: ID of the pipeline run to stop.
+        graceful: If True, allows for graceful shutdown where possible.
+            If False, forces immediate termination. Default is False.
+    """
+    run = verify_permissions_and_get_entity(
+        id=run_id,
+        get_method=zen_store().get_run,
+        hydrate=True,
+    )
```
stefannica: This operation has a few more consequences than simply reading the run. I reckon you would need at least `write` permissions on the pipeline run in order to stop it. Maybe even read permissions on the orchestrator stack component. WDYT?

src/zenml/cli/pipeline.py
```diff
@@ -511,6 +511,59 @@ def list_pipeline_runs(**kwargs: Any) -> None:
...
+        run_utils.stop_run(run=run, graceful=graceful)
+        action = "Gracefully stopped" if graceful else "Force stopped"
+        cli_utils.declare(f"{action} pipeline run '{run.name}'.")
+    except NotImplementedError:
+        cli_utils.error(
+            "The orchestrator used for this pipeline run does not support "
+            f"{'gracefully' if graceful else 'forcefully'} stopping runs."
+        )
+    except Exception as e:
+        cli_utils.error(f"Failed to stop pipeline run: {e}")
```
stefannica, line 564: It would be a good idea to catch `IllegalOperationError` too, as it's thrown a lot if the pipeline run is not in the correct state.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -131,11 +133,79 @@ def __init__(
...
+                        "The execution was interrupted."
+                    )
+            except (RunStoppedException, RunInterruptedException):
+                raise
+            except Exception as e:
+                raise RunInterruptedException(str(e))
+
+        # Register handlers for common termination signals
+        signal.signal(signal.SIGTERM, signal_handler)
+        signal.signal(signal.SIGINT, signal_handler)
```
stefannica, line 209: You have to assume that for some orchestrators, the process calling this constructor may have already configured its own signal handlers. Instead of simply overwriting the previous handlers, you should save them and call them as part of your handler (i.e. practice signal handler chaining). If you don't, you might break some termination and cleanup functionality that the orchestrator relies on to keep track of its steps.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -131,11 +133,79 @@ def __init__(
...
+                        orchestrator_run_id=self._orchestrator_run_id,
+                        hydrate=False,
+                    )
+                    if pipeline_run_list.items:
+                        pipeline_run = pipeline_run_list.items[0]
+
+                if pipeline_run and pipeline_run.status in [
+                    ExecutionStatus.STOPPING,
+                    ExecutionStatus.STOPPED,
+                ]:
```
stefannica, line 180: shouldn't you publish this status update regardless of the pipeline run status ? What if one step gets killed without anyone stopping the parent pipeline run?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -259,6 +331,12 @@ def _bypass() -> None:
                                 force_write_logs=force_write_logs,
                             )
                             break
+                        except (
+                            RunStoppedException,
+                            RunInterruptedException,
+                            KeyboardInterrupt,
+                        ) as e:
+                            raise e
```
stefannica: If the step is interrupted while the pipeline run is still RUNNING (so not because the pipeline run itself was stopped), should we still retry the step here ?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -296,10 +374,11 @@ def _bypass() -> None:
                         artifacts=step_run.outputs,
                         tags=pipeline_run.config.tags,
                     )
-
+        except RunStoppedException:
+            logger.info(f"Pipeline run `{pipeline_run.name}` stopped.")
+            raise
         except:  # noqa: E722
             logger.error(f"Pipeline run `{pipeline_run.name}` failed.")
-            publish_utils.publish_failed_pipeline_run(pipeline_run.id)
```
stefannica, line 298: So if the step failed for any other reason, like an error in the custom user code, you would previously mark the pipeline as failed, but not anymore. Why ?

And it should have been the step being marked as failed, not the entire pipeline run, correct ?

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -859,12 +859,16 @@ def fetch_status(self, run: "PipelineRunResponse") -> ExecutionStatus:
...
-        # https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/PipelineState
-        if status in ["Executing", "Stopping"]:
+        # https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DescribePipelineExecution.html
+        if status == "Executing":
             return ExecutionStatus.RUNNING
-        elif status in ["Stopped", "Failed"]:
+        elif status == "Stopping":
+            return ExecutionStatus.STOPPING
+        elif status == "Stopped":
+            return ExecutionStatus.STOPPED
```
stefannica, line 862: Just checking: there is no implementation for `_stop` in this orchestrator, so these statuses would only be set if the user manually stopped the Sagemaker pipeline using the AWS console or CLI, correct ?

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
```diff
@@ -506,14 +506,16 @@ def fetch_status(self, run: "PipelineRunResponse") -> ExecutionStatus:
             return ExecutionStatus.INITIALIZING
         elif status in ["Running", "Finalizing"]:
             return ExecutionStatus.RUNNING
+        elif status == "CancelRequested":
+            return ExecutionStatus.STOPPING
+        elif status == "Canceled":
+            return ExecutionStatus.STOPPED
```
stefannica, line 521: Same here: there is no implementation for `_stop` in this orchestrator.

src/zenml/orchestrators/base_orchestrator.py
stefannica: One example: 
image

src/zenml/orchestrators/step_launcher.py
bcdurak: That's a good catch. 👍 I implemented the chaining.

Actually, with the same logic, some other tool can very well overwrite our signal handler too. Still, we are setting them up when we initialize the StepLauncher, so I guess it should not be a big issue.

src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
bcdurak: People can stop their pipeline on the orchestrator side, and refreshing the status will be more accurate in this case.

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
bcdurak: Same as above, so I am resolving this one for now.

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: I have extended the endpoint to include an "UPDATE" action check as well. The run that is forwarded to the utility function that comes after this check is already dehydrated at this point. There is a check there that checks the stack and if it doesn't exist (due to RBAC or a deletion), it warns the user accordingly. Do you think we should explicitly check for the RBAC case here as well or is this enough?

src/zenml/cli/pipeline.py
bcdurak: Isn't it already covered by the `except Exception as e` or am I missing something here?

src/zenml/cli/pipeline.py
bcdurak: Normally, I would agree with this. But upon a quick investigation, I have realized that most of our orchestrators do not have a feature to "gracefully" cancel runs. I thought it wouldn't make sense to make it the default behaviour when a vast majority of our integrations do not support it.

src/zenml/orchestrators/step_launcher.py
bcdurak: That's a good point. I think we should just keep the `RunStoppedException` here (which can only happen if the run was already in a `Stopping` state). Changing the code accordingly.

src/zenml/orchestrators/step_launcher.py
bcdurak: 1. I don't think that we need this here anymore. Every time a step's status gets updated, the server actually updates the status of the pipeline run accordingly anyway.
2. Since the exception is still raised, this step will still be marked as `Failed`.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: I implemented this because it was a part of the ticket that explicitly stated `Ctrl+C must stop running pods`. Perhaps we can discuss whether we should implement this or not. @schustmi any thoughts?

src/zenml/orchestrators/step_launcher.py
bcdurak: This is the special case. If your scenario happens, we raise a `RunInterruptedException` which ends up tagging the step as `Failed`.

src/zenml/cli/pipeline.py
```diff
@@ -511,6 +511,59 @@ def list_pipeline_runs(**kwargs: Any) -> None:
...
+        yes: If set, don't ask for confirmation.
+    """
+    # Ask for confirmation to stop run.
+    if not yes:
+        action = "gracefully stop" if graceful else "force stop"
+        confirmation = cli_utils.confirmation(
+            f"Are you sure you want to {action} pipeline run `{run_name_or_id}`?"
+        )
+        if not confirmation:
+            cli_utils.declare("Pipeline run stop canceled.")
```
schustmi: `stop canceled` seems a little weird? Maybe `Not stopping pipeline run` or something along those lines?

src/zenml/orchestrators/publish_utils.py
```diff
@@ -54,6 +55,34 @@ def publish_successful_step_run(
...
+    Returns:
+        The updated step run.
+    """
+    from zenml.client import Client
+
+    step_run = Client().zen_store.update_run_step(
+        step_run_id=step_run_id,
+        step_run_update=StepRunUpdate(
+            status=status,
+            end_time=end_time,
```
schustmi, line 85: The end time should only be set if publishing a state that also indicates that the step is finished, maybe that is something that this function could verify?

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -326,6 +338,8 @@ def run(
                             submission_result.wait_for_completion()
                         except BaseException as e:
                             raise RunMonitoringError(original_exception=e)
+
+        # TODO: Need to add the KeyboardInterrupt handling here.
```
schustmi: I agree with @stefannica here. If we use keyboard interrupts as a stopping signal for synchronous orchestrators, there is no way to stop the client without stopping the run as well, which I'm not a fan of.

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -141,6 +142,15 @@ def config(self) -> BaseOrchestratorConfig:
         """
         return cast(BaseOrchestratorConfig, self._config)
 
+    @property
```
schustmi: Do we actually need this property, or can we simply detect whether the orchestrator subclass implements the `_stop_run` method?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -642,3 +651,107 @@ def get_orchestrator_run_id(self) -> str:
...
+            raise ValueError(
+                "Cannot determine orchestrator run ID for the run. "
+                "Unable to stop the pipeline."
+            )
+
+        pods_stopped = []
+        errors = []
+
+        # Configure graceful termination settings
+        grace_period_seconds = (
```
schustmi: Why is this graceful, even though the user explicitly requested non-graceful termination?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -642,3 +651,107 @@ def get_orchestrator_run_id(self) -> str:
...
+                        f"Successfully initiated graceful stop of step pod: {pod.metadata.name}"
+                    )
+                except Exception as e:
+                    error_msg = (
+                        f"Failed to stop step pod {pod.metadata.name}: {e}"
+                    )
+                    logger.warning(error_msg)
+                    errors.append(error_msg)
+
+        except Exception as e:
```
schustmi, line 699: This exception only handles the `list_namespaced_pod` call, so I think the code would be much more readable if the `try..except` would be just around that code instead of this whole section of code.

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -157,19 +159,22 @@ def build_pod_manifest(
         containers=[container_spec],
         restart_policy="Never",
         image_pull_secrets=image_pull_secrets,
+        termination_grace_period_seconds=30,
```
schustmi: Maybe make this configurable in the function?

src/zenml/orchestrators/dag_runner.py
```diff
@@ -103,6 +108,7 @@ def __init__(
         self.reversed_dag = reverse_dag(dag)
         self.run_fn = run_fn
         self.finalize_fn = finalize_fn
+        self.check_fn = check_fn
```
schustmi: `check_fn` seems like a very unclear name, this is a function that checks whether the whole dag runner is cancelled or not, correct?

src/zenml/orchestrators/dag_runner.py
```diff
@@ -168,6 +174,14 @@ def _run_node(self, node: str) -> None:
         """
         self._prepare_node_run(node)
 
+        # Check if execution should continue (e.g., check for cancellation)
+        if self.check_fn and not self.check_fn():
```
schustmi: If this function is something that cancels the entire dag runner, wouldn't it make sense to simply run it once here, if it returns that it should be canceled, set a boolean and not rerun every time?

src/zenml/orchestrators/publish_utils.py
```diff
@@ -92,25 +119,78 @@ def publish_failed_pipeline_run(
...
-        or len(step_statuses) < num_steps
+        run_status == ExecutionStatus.RUNNING
+        or run_status == ExecutionStatus.INITIALIZING
     ):
-        return ExecutionStatus.RUNNING
+        if ExecutionStatus.FAILED in step_statuses:
+            return ExecutionStatus.FAILED
+        elif ExecutionStatus.STOPPED in step_statuses:
+            if all(status.is_finished for status in step_statuses):
+                return ExecutionStatus.STOPPED
```
schustmi: How can a step be in status `STOPPED` with a run not in status `STOPPING/STOPPED`?

src/zenml/utils/run_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    # Check if the stack is still accessible
+    if run.stack is None:
+        raise ValueError(
+            "The stack that this pipeline run response was executed on "
+            "is either not accessible or has been deleted."
+        )
+
+    # Check if pipeline can be stopped
+    if run.status == ExecutionStatus.COMPLETED:
```
schustmi, line 42: maybe check for `run.status.is_finished` instead?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8814,13 +8824,15 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
...
-                reference_id=step_run.original_step_run_id,
-                session=session,
-                reference_type="original step run",
-            )
+            # Validate pipeline status before creating step
+            if run.status in [
+                ExecutionStatus.STOPPING,
+                ExecutionStatus.STOPPED,
+            ]:
+                raise RunStoppedException(
```
schustmi: This needs to be mapped to an HTTP error code

src/zenml/orchestrators/step_launcher.py
```diff
@@ -131,11 +133,92 @@ def __init__(
...
+
+            try:
+                client = Client()
+                pipeline_run = None
+
+                if self._step_run:
+                    pipeline_run = client.get_pipeline_run(
+                        self._step_run.pipeline_run_id
+                    )
+                else:
```
schustmi, line 171: If `self._step_run` doesn't exist (which means the step was never created in the DB), why can't we simply raise the exception here? Currently, you're fetching the run to check is status, but then later you can't do anything with it as there is no step run to update.

src/zenml/orchestrators/base_orchestrator.py
stefannica: I know, but you're changing a behaviour that our users (including myself) were used to since day 1 :smile:

src/zenml/cli/pipeline.py
stefannica: Yest, but I meant catch it and show a different error message for those cases.

src/zenml/cli/pipeline.py
bcdurak: Agreed and fixed.

src/zenml/orchestrators/publish_utils.py
bcdurak: Makes sense. Added:

```python
if end_time is not None and not status.is_finished:
    raise ValueError("End time cannot be set for a non-finished step run.")
```

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: If this is not graceful, it sends out `SIGKILL`, and it doesn't give the pods any chance to react to the incoming signal through a signal handler.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: I use it for the following purpose:

1. `stop_run` does following two things: mark the pipeline `stopping` and actually stop the run through `_stop_run`. 
2. `stop_run` is a global implementation for all orchestrators that's why the `NotImplementedError` is in the `_stop_run`
3. The property sort of tells me if the `_stop_run` will throw a `NotImplementedError` before I actually call it, so that I do not update the status to be `stopping`

If you can think of a different way to handle this, I am open to suggestions.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Makes sense, rearranged the code.

src/zenml/orchestrators/publish_utils.py
bcdurak: Right now, there is no possible way, but I thought we might implement something in the future.

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
bcdurak: I adjusted this. Now, a possible value in the settings. It is only set here and we do not need to use it again when we call `delete_namespaced_pod`

src/zenml/utils/run_utils.py
bcdurak: Nope, theoretically, `Failed` pipelines are already deemed finished but people can still stop any running pods right?

src/zenml/zen_stores/sql_zen_store.py
bcdurak: You are right, however, now that you mention it, I implemented the `RunStoppedException` with the sole purpose of handling the shutdown of the pod. I think it is better to switch this to an `IllegalOperationError`. WDYT?

src/zenml/orchestrators/base_orchestrator.py
bcdurak: Should we make this configurable or completely remove this then?

src/zenml/orchestrators/dag_runner.py
bcdurak: Applied it.

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -88,6 +90,7 @@ class KubernetesOrchestratorSettings(BaseSettings):
     failed_jobs_history_limit: Optional[NonNegativeInt] = None
     ttl_seconds_after_finished: Optional[NonNegativeInt] = None
     prevent_orchestrator_pod_caching: bool = False
+    pod_stop_grace_period: PositiveInt = 30
```
schustmi, line 93: So I'm not allowed to set it to 0? :D

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -169,6 +170,7 @@ def pre_step_run(step_name: str) -> bool:
         "pipeline": kube_utils.sanitize_label(
             deployment.pipeline_configuration.name
         ),
+        "zenml-orchestrator-run-id": orchestrator_pod_name,
```
schustmi: I think with my change, every pod is already tagged with the `run.id`, which maybe is even nicer to use for you? That would also improve the error messages in the `KubernetesOrchestrator._stop_run`, which currently mention the `orchestrator_run_id` which most users won't be familiar with.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -330,6 +333,51 @@ def finalize_run(node_states: Dict[str, NodeStatus]) -> None:
...
+                hydrate=False,  # We only need status, not full hydration
+                project=deployment.project_id,
+                deployment_id=deployment.id,
+                **list_args,
+            )
+            if not len(pipeline_runs):
+                # No pipeline run found, assume we should continue
+                return True
+
+            pipeline_run = pipeline_runs[0]
```
schustmi: Also with my latest changes that are merged in here, you are guaranteed to have a `pipeline_run` already (defined somewhere above). I think in this function, you should just re-fetch this pipeline run, no need to run any `list_pipeline_runs(...)` calls

src/zenml/orchestrators/step_launcher.py
schustmi: I think we never needed this, but it was added as an additional safety mechanism if for some reason the step status update failed.

src/zenml/orchestrators/dag_runner.py
```diff
@@ -173,6 +181,17 @@ def _run_node(self, node: str) -> None:
...
+        # Check if execution should continue (e.g., check for cancellation)
+        if self.check_fn:
+            if self._stop_requested:
+                self._finish_node(node, cancelled=True)
+                return
+
+            if not self.check_fn():
+                self._stop_requested = True
+                self._finish_node(node, cancelled=True)
+                return
```
schustmi:
```python
self._stop_requested = self._stop_requested or not self._check_fn():
if self._stop_requested:
  self._finish_node(node, cancelled=True)
  return
```

Maybe like this is a little cleaner?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -259,7 +344,9 @@ def _bypass() -> None:
                                 force_write_logs=force_write_logs,
                             )
                             break
-                        except BaseException as e:  # noqa: E722
+                        except RunStoppedException as e:
+                            raise e
+                        except Exception as e:  # noqa: E722
```
schustmi: Why not `BaseException` anymore, which also covers keyboard interrupt?

schustmi (changes requested):

src/zenml/orchestrators/base_orchestrator.py
schustmi: IMO, I would keep the existing behaviour. If anything, we could check whether its a keyboard interrupt during the `waiting phase`, and then log a CLI command that tells the user that:
- their run is still going
- a CLI command to stop the run if they want to

src/zenml/orchestrators/base_orchestrator.py
schustmi: Yes, in `stop_run` you could check whether the orchestrator subclass overwrites the `_stop_run` method (in which case stopping  is supported. I do a similar thing in the `run(...)` method in this class, where I check whether a subclass overwrites the `submit_pipeline(...)` method.

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
bcdurak: Yes, I wanted to make sure that the pod shuts down gracefully. Otherwise, it is the SIGKILL again.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: I made the necessary changes for this one as well.

src/zenml/orchestrators/step_launcher.py
bcdurak: I would like to keep it like this for the following scenario:

- I run a pipeline.
- I canceled it gracefully.
- Run gets updated to `Stopping`.
- While it waits for the steps to finish, one of the running steps has failed.
- This should not mark the status of the pipeline run as `Failed`.

Once a pipeline is in the `Stopping` state, the only state that it can switch to is the `Stopped` state IMO.

src/zenml/orchestrators/step_launcher.py
bcdurak: I initially changed this because I was also dealing with the `Ctrl+C` case in this PR. Even though the `Ctrl+C -> Stop Run` case is removed for now, I think a `KeyboardInterrupt` should still stop the execution instead of going into the retries. WDYT?

src/zenml/orchestrators/base_orchestrator.py
bcdurak: That's a great idea. Implemented it.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: We are continuing the discussion [here](https://github.com/zenml-io/zenml/pull/3716#discussion_r2172244356). So, I am resolving this for now.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: Cool, I have adapted the approach.

src/zenml/orchestrators/dag_runner.py
bcdurak: In the docstring, I have described it as: 

```check_fn: A function that returns True if the run should continue after each step execution, False if it should stop (e.g., due to cancellation). If None, execution continues normally.```

It can potentially cover more than the cancellation case, thus the vague naming.

File: src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py (hunk 26 above)
Comment (bcdurak) [reply src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py]: Thanks for the assist. 😄 I really hated implementing that list call there.

File: src/zenml/orchestrators/step_launcher.py (hunk 23 above)
Comment (bcdurak) [reply src/zenml/orchestrators/step_launcher.py:171]: Done.

File: src/zenml/orchestrators/step_launcher.py (hunk 8 above)
Comment (schustmi) [reply src/zenml/orchestrators/step_launcher.py:298]: Yes, then that is something you should prevent from happening server side.

File: src/zenml/orchestrators/dag_runner.py (hunk 18 above)
Comment (schustmi) [reply src/zenml/orchestrators/dag_runner.py]: It checks whether the run should `continue` in your own words. If the run should not continue, by definition it got cancelled, no? Whether that is by the user or by anyone else

File: src/zenml/orchestrators/step_launcher.py (hunk 28 above)
Comment (schustmi) [reply src/zenml/orchestrators/step_launcher.py]: Mhm only the last retry executes the failure hook, I'm not sure how to best handle this tbh. I would just leave it as is as I haven't heard any complaints about this tbh.

File: src/zenml/zen_stores/sql_zen_store.py [hunk 29]
```diff
@@ -6120,6 +6120,14 @@ def update_run(
                 session=session,
             )
             session.refresh(existing_run)
+
+            if run_update.status is not None:
+                self._update_pipeline_run_status(
+                    pipeline_run_id=run_id,
+                    session=session,
+                )
+            session.refresh(existing_run)
```
schustmi, line 6128: Can we avoid this double refresh and just call it afterwards?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8835,6 +8844,16 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
                 reference_id=step_run.pipeline_run_id,
                 session=session,
             )
+
+            # Validate pipeline status before creating step
+            if run.status in [
+                ExecutionStatus.STOPPING,
+                ExecutionStatus.STOPPED,
+            ]:
+                raise IllegalOperationError(
```
schustmi, line 8852: I'm wondering whether a 403 HTTP error (which this is converted to) is the right thing here? Maybe a 400/422 would be better? Also curious what @stefannica thinks

src/zenml/orchestrators/step_launcher.py
bcdurak: Alright, I am switching it back to `BaseException`.

src/zenml/orchestrators/dag_runner.py
bcdurak: Yes, I renamed to `continue_fn`.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Leaving this as 403.

stefannica (approved):

PR #3719, opened 2025-06-02
Improved Kubernetes orchestrator pod caching

schustmi:
## Describe changes
This PR
- enables orchestrator pod caching for scheduled runs.
- delays the caching decision to run right before the actual step pod starts, not at orchestrator pod startup. This means the caching decision will be made with the latest state of all step runs available in the server, which might differ from the step runs available at orchestrator pod startup.
- adds metadata to the Kubernetes orchestrator.
- improves pod labels when using the Kubernetes orchestrator.

**Changes to what we consider placeholder runs:**
Previously, a run without an `orchestrator_run_id` in the database was considered a placeholder run. This PR changes that behaviour and instead uses the status of a run to decide whether the run is a placeholder run (runs with state `initializing` are placeholder runs). This change was necessary so that the orchestrator pod can create a pipeline run (with orchestrator run ID) for scheduled runs to enable the orchestrator pod caching.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5840,18 +5840,33 @@ def _replace_placeholder_run(
...
+            # In very rare cases, there can be multiple placeholder runs for
+            # the same deployment. By ordering by the orchestrator_run_id, we
+            # make sure that we use the placeholder run with the matching
+            # orchestrator_run_id if it exists, before falling back to the
+            # placeholder run without any orchestrator_run_id provided.
+            # Note: This works because both SQLite and MySQL consider NULLs
+            # to be lower than any other value. If we add support for other
+            # databases (e.g. PostgreSQL, which considers NULLs to be greater
+            # than any other value), we need to potentially adjust this.
+            .order_by(desc(PipelineRunSchema.orchestrator_run_id))
```
stefannica, line 5873: good to know !

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -562,13 +564,33 @@ def update_placeholder(
                 "placeholder run."
             )
 
+        if request.status == ExecutionStatus.INITIALIZING:
```
stefannica:
```suggestion
        if request.is_placeholder_request:
```

... for consistency ?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5951,27 +5969,31 @@ def get_or_create_run(
...
-                #     is the case if the user ran the pipeline on a schedule.
-                # (2) There was a placeholder run, but a previous pipeline run
-                #     already used it. This is the case if users rerun a
-                #     pipeline run e.g. from the orchestrator UI, as they will
-                #     use the same deployment_id with a new orchestrator_run_id.
-                # (3) A step of the same pipeline run already replaced the
-                #     placeholder run.
-                pass
+            if not pipeline_run.is_placeholder_request:
+                # Only run this is the request is not a placeholder run itself,
```
stefannica:
```suggestion
                # Only run this if the request is not a placeholder run itself,
```

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -152,6 +152,15 @@ def populate_request(
...
 
+                # As a last resort, we try to reuse the docstring/source code
+                # from the cached step run. This is part of the cache key
+                # computation, so it must be identical to the one we would have
+                # computed ourselves.
+                if request.source_code is None:
+                    request.source_code = cached_step_run.source_code
+                if request.docstring is None:
+                    request.docstring = cached_step_run.docstring
+
```
stefannica, line 164: Can you explain this part a bit ? This wasn't necessary before and the `_get_docstring_and_source_code` already attempts to fetch these from various sources (the step instance and run template). Why is it absolutely necessary now that these be populated in the request, if it wasn't necessary before ?

src/zenml/integrations/kubernetes/step_operators/kubernetes_step_operator.py
```diff
@@ -216,6 +214,10 @@ def launch(
             pod_settings=settings.pod_settings,
             env=environment,
             mount_local_stores=False,
+            labels={
+                "run_id": kube_utils.sanitize_label(str(info.run_id)),
```
stefannica, line 218: You are intentionally using the run_id here instead of the run_name, yes ? Why?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -282,16 +299,21 @@ def finalize_run(node_states: Dict[str, NodeStatus]) -> None:
 
                 # If steps failed for any reason, we need to mark the step run as
                 # failed, if it exists and it wasn't already in a final state.
+                step_runs = Client().list_run_steps(
+                    size=1,
+                    pipeline_run_id=pipeline_run.id,
+                    name=step_name,
+                )
```
stefannica: this will scale badly, I know it. For 1000 steps, it will make 1000 calls to the API. I think we might need a new endpoint that fetches the status of all steps in one go. Or maybe move this whole thing behind the REST API.

stefannica: On second thought, this whole thing feels like something that should happen server-side when the pipeline run status is updated... this whole thing needs to be reconsidered.

schustmi: My change here only improves upon the previous implementation, which fetched **all** steps of the pipeline run. My solution instead only fetches the steps for which the pod is actually in a failed state. However, I definitely agree that we could batch this into some requests, instead of fetching each step separately. I'll change it.

src/zenml/orchestrators/step_run_utils.py
schustmi: This was never necessary (both of those fields are optional). It just now occurred to me that we can simply reuse the code/docstring of the cached step if we can't figure it out in any other way, so I added it. Otherwise, all steps that got cached from the orchestrator pod would not have any code/docstring associated with them, as
- the code to import the step instance most likely doesn't exist in the orchestrator pod
- there is no run template involved

src/zenml/integrations/kubernetes/step_operators/kubernetes_step_operator.py
schustmi: No specific reason other than the ID being the thing that you'll usually use to fetch a run from ZenML I think. I'll also add the run name

schustmi: Done

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
schustmi: I'm now fetching only relevant steps, and doing so in batched chunks.

stefannica (approved):

PR #3721, opened 2025-06-02
Use the active artifact store as cache when loading artifacts

stefannica:
## Describe changes
Loading multiple artifacts can take a long time because every call to `artifact.load()` ends up re-initializing the Artifact Store where the artifact is store, which includes fetching an entirely new set of Service Connector client credentials.

A "gather results" step implementation like the following can end up calling the Service Connector client credentials endpoint tens of times, amounting to several minutes or even hours to complete the execution, if hundreds of artifacts are involved:

```
    results = []
    for step_name, step_info in steps.items():
        if not step_name.startswith("load_step"):
            continue

        # Get the output artifact
        output = step_info.outputs.get("output")
        if output:
            # Get the actual artifact data using the client
            artifact_data = output[0].load()
            results.append(artifact_data)
```

This PR makes a minor but powerful change: using the artifact store object in the active stack as a cache, if it happens to be the artifact store where the requested artifact is stored (which is 80% of the cases). This way, only one Artifact Store instance and set of Service Connector client credentials are required.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
+++ b/src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
@@ -20,24 +20,35 @@ from typing import (
     Any,
     Dict,
     Iterator,
+    List,
     Optional,
     Tuple,
     Type,
+    Union,
     cast,
 )
 from uuid import UUID
 
 import boto3
-import sagemaker
 from botocore.exceptions import WaiterError
+from sagemaker.estimator import Estimator
+from sagemaker.inputs import TrainingInput
 from sagemaker.network import NetworkConfig
-from sagemaker.processing import ProcessingInput, ProcessingOutput
+from sagemaker.processing import ProcessingInput, ProcessingOutput, Processor
+from sagemaker.session import Session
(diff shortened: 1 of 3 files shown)


src/zenml/integrations/aws/orchestrators/sagemaker_orchestrator.py
```diff
@@ -475,42 +505,56 @@ def prepare_or_run_pipeline(
                         )
                     )
 
-            # Convert environment to a dict of strings
-            environment = {
+            step_environment: Dict[str, Union[str, PipelineVariable]] = {
                 key: str(value)
-                if not isinstance(value, ExecutionVariable)
+                if not isinstance(value, PipelineVariable)
```
schustmi, line 510: Because we're adding the Execution variable only in the line after this, `environment` will be a `Dict[str, str]` which means this check isn't strictly necessary, correct?

stefannica: I would also assume so, but I need @bcdurak to confirm, since he introduced this change in https://github.com/zenml-io/zenml/pull/3380/files and I don't have the full story behind it.

PR #3723, opened 2025-06-04
Split run submission and monitoring

schustmi:
## Describe changes
This PR introduces a new method `submit_pipeline(...)` on the orchestrator interface as a replacement for the `prepare_or_run_pipeline(...)` method.

For most orchestrators, the `prepare_or_run_pipeline` method had two main responsibilities:
- Submitting the pipeline to the underlying orchestrator backend
- In case the orchestrator was configured to run synchronous, waiting (and potentially streaming logs) until the run finished

This caused some issues with error handling outside this method, as depending on which phase the orchestrator is in (submitting vs waiting), we want to treat errors differently:
- If there is an error during the submission phase, we want to mark the corresponding run as failed
- If there is an error during the waiting phase, we do not want to mark the run as failed as this might cause downstream problems if some steps of the run are still being executed (e.g. token expiration, status checks)

## Pre-requisites

## Types of changes


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -39,6 +52,20 @@
...
 
 
+class SubmissionResult:
+    """Result of submitting a pipeline run."""
+
+    def __init__(
+        self,
+        wait_for_completion: Optional[Callable[[], None]] = None,
+        run_metadata: Optional[Dict[str, MetadataType]] = None,
+        schedule_metadata: Optional[Dict[str, MetadataType]] = None,
```
stefannica: Can you make this a union, for clarity?

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -402,57 +401,29 @@ def _convert_to_custom_training_job(
...
-        `ContainerOp`, because using the `dsl.ContainerOp` class directly is
-        deprecated when using the Kubeflow SDK v2. The step entrypoint command
-        with the entrypoint arguments is the command that will be executed by
-        the container created using the previously created Docker image.
-
-        This callable is then compiled into a JSON file that is used as the
-        intermediary representation of the Kubeflow pipeline.
-
-        This file then is submitted to the Vertex AI Pipelines service for
-        execution.
```
stefannica, line 437: Why remove this whole thing ?

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -134,52 +152,47 @@ def get_orchestrator_run_id(self) -> str:
             The orchestrator run id.
         """
 
-    @abstractmethod
+    def submit_pipeline(
```
bcdurak, line 155: Should this be an `@abstractmethod`?

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -95,22 +128,7 @@ def supports_client_side_caching(self) -> bool:
...
-    -------------
-    The `run(...)` method is the entrypoint that is executed when the
-    pipeline's run method is called within the user code
-    (`pipeline_instance.run(...)`).
-
-    This method will do some internal preparation and then call the
-    `prepare_or_run_pipeline(...)` method. BaseOrchestrator subclasses must
-    implement this method and either run the pipeline steps directly or deploy
-    the pipeline to some remote infrastructure.
-    """
```
bcdurak, line 113: I would still keep an explanation here even though it changed a little bit.

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -134,52 +152,47 @@ def get_orchestrator_run_id(self) -> str:
...
         """
 
-    @abstractmethod
+    def submit_pipeline(
+        self,
+        deployment: "PipelineDeploymentResponse",
+        stack: "Stack",
+        environment: Dict[str, str],
+        placeholder_run: Optional["PipelineRunResponse"] = None,
+    ) -> Optional[SubmissionResult]:
```
bcdurak, line 161: I have a question here. Due to the older architecture, we were previously using a yield here. Thanks to that yield pattern, any metadata that was generated was immediately communicated to the server. But now, it is slightly different. Any error that happens during the submission of the pipeline will hold us back from sending that metadata to the server. This is probably not a big deal for most orchestrators but can you think of any examples where this can be an issue?

src/zenml/orchestrators/local/local_orchestrator.py
```diff
@@ -87,6 +96,7 @@ def prepare_or_run_pipeline(
             string_utils.get_human_readable_time(run_duration),
         )
         self._orchestrator_run_id = None
+        return None
```
bcdurak, line 99: I have seen this in a few more places. Why do we explicitly return a `None` here?

src/zenml/orchestrators/base_orchestrator.py
schustmi: I made this not an abstract method to keep backwards compatibility with existing custom orchestrators, which haven't migrated to the new method yet.

src/zenml/orchestrators/base_orchestrator.py
schustmi: It's gonna be removed anyway, right? No one should ever look at this method from now on, and all the existing ones have already implemented it. IMO this explanation was useless anyway, reading it does not benefit anyway anyone doesn't already know what they're doing :D

src/zenml/orchestrators/local/local_orchestrator.py
schustmi: Because `mypy` wants it that way 😄
But more generally, this returns an optional submission result, so it only makes sense to return one of the two options: `None` or `SubmissionResult(...)`

src/zenml/orchestrators/base_orchestrator.py
schustmi: No, all the orchestrator methods now return at essentially the same place as they yielded before, so there is no extra code being executed before sending the metadata to the server.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: Yes, perhaps this was meant to be more of a general statement, and I misplaced this comment. When @stefannica and I were taking a look at this PR last week, we saw that some of the docstrings shrank quite significantly during the changes (Vertex and Kubeflow are good examples of this). I thought that we might lose some relevant information there.

schustmi: I think I only removed docstrings that were either
- completely wrong, as they were copy pasted from a different orchestrator
- completely useless, as they were copy pasted from the base class (which as described above, is not only generic but also provides no actual guidelines on what to do)
- outdated, because it referenced some internal implementation details, which is not the purpose of a docstring.

src/zenml/integrations/lightning/orchestrators/lightning_orchestrator.py
```diff
@@ -194,81 +156,46 @@ def setup_credentials(self) -> None:
...
-                    "Property `schedule_timezone` must be set when passing "
-                    "`cron_expression` to a Lightning orchestrator."
-                    "Lightning orchestrator requires a Java Timezone ID to run the pipeline on schedule."
-                    "Please refer to https://docs.oracle.com/middleware/1221/wcs/tag-ref/MISC/TimeZones.html for more information."
-                )
+            logger.warning(
+                "Lightning Orchestrator currently does not support the "
+                "use of schedules. The `schedule` will be ignored "
+                "and the pipeline will be run immediately."
+            )
```
stefannica, line 191: :smile:

PR #3724, opened 2025-06-05
Tag improvements

bcdurak:
## Describe changes
- Fixed a problem while fetching entities that have the same exclusive tag on the SQLZenStore.
- Previously, you could possibly call `add_tags(tags=[...], pipeline=..., run=..., run_template=...)`, which would only affect the pipeline as it is the first if condition. I added assertions to enforce a proper usage, but tbh the solution still seems quite ugly to me. Really open to suggestions here. @schustmi If you greenlight this, I will make the same changes to the `metadata_utils` as well.

## Pre-requisites

## Types of changes


src/zenml/utils/tag_utils.py
```diff
@@ -220,44 +220,88 @@ def add_tags(
     resource_type = None
 
     # Tag a pipeline
-    if pipeline is not None:
```
schustmi, line 223: So this now silently fails/does nothing if I provide a pipeline and an artifact?

bcdurak: Actually, it would try to tag the artifact because I made a mistake while implementing the if conditions. Fixed it now.

strickvl (approved):

PR #3728, opened 2025-06-06
Dynamically import entrypoint in root __init__

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/__init__.py
+++ b/src/zenml/__init__.py
@@ -14,6 +14,7 @@
 """Initialization for ZenML."""
 
 import os
+from typing import Any
 
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


schustmi:
@stefannica Can you explain what the supposed circular import is here? If circular imports really were an issue here, it would not actually be running because they cause an exception.

I can try to explain the issue in more detail:
- For databricks, we essentially need the following: A `package_name` (=`zenml` in our case), as well as a string that points databricks to a function it can execute as the main entrypoint (we're passing `entrypoint.main`). What they do is essentially the following: 
  ```python
  result = importlib.import(package_name)
  for name in entrypoint.split("."):
    result = getattr(result, name)
  ```
  What this means for us is the we need a way to get to the entrypoint module from our root `__init__` module using `getattr(...)`.
- For most our other orchestrators, we run the following command in a Docker container: `python -m zenml.entrypoints.entrypoint ...`. When this is being executed, Python first imports the root `zenml/__init__.py` module. It then imports the `zenml/entrypoints/__init__.py` module. It then imports the `zenml/entrypoints/entrypoint.py` module. This last point is where we get the runtime warning, which tells us that the module it is trying to import (`zenml.entrypoints.entrypoint`) already exists in `sys.modules` as it was imported by the root `__init__` module already.

This PR implements a solution so that the first case is still possible (using the dynamic import behaviour) while making sure that our root `__init__` doesn't actually import the entrypoint module in the second case.

PR #3733, opened 2025-06-12
Add Modal orchestrator with step operator and orchestrator flavors

htahir1:
**New Features**
  * Introduced Modal orchestrator integration, enabling orchestration and scaling of ML pipelines on Modal's serverless cloud platform with support for both pipeline and per-step execution modes.
  * Added configuration options for resource allocation, authentication, environment selection, and execution modes for both orchestrator and step operator.

**Documentation**
  * Added comprehensive guides for the new Modal orchestrator and enhanced the Modal step operator documentation with setup, configuration, authentication, and usage best practices.
  * Updated component guide and table of contents to include Modal orchestrator details and clarified differences between orchestrator and step operator.

**Refactor**
  * Modularized step operator logic by moving validation, image management, and authentication into shared utility functions for improved maintainability.

**Tests**
  * Updated tests to align with the new parameter structure for GPU configuration utilities.

  Breaking Changes

  None - this is a new integration that doesn't affect existing functionality.

  Dependencies

  - Adds modal>=0.64.49,<1 requirement to Modal integration
  - No new dependencies for core ZenML

  ---
  Note: This orchestrator follows the same patterns as other ZenML orchestrators
  (GCP Vertex, Kubernetes) and integrates seamlessly with the existing ZenML
  stack architecture.

Note: I also updated the step operator logic to unify it

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

**New Features**
  * Introduced Modal orchestrator integration, enabling orchestration and scaling of ML pipelines on Modal's serverless cloud platform with configurable execution modes and resource settings.
  * Added comprehensive documentation for the Modal orchestrator, including setup, configuration, usage, troubleshooting, and best practices.

**Enhancements**
  * Expanded Modal step operator documentation with detailed setup instructions, authentication options, GPU support, and environment management.
  * Improved configuration options for the Modal step operator, including environment, timeout, and secure API token handling.

**Bug Fixes**
  * Updated internal references and tests to align with new configuration parameter structures.

**Refactor**
  * Modularized Modal integration by introducing utility functions for authentication, resource management, image handling, and stack validation.

**Documentation**
  * Updated and expanded documentation for Modal orchestrator and step operator, including a new entry in the component guide table of contents.


src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,642 @@
...
+                        # Skip logs that seem to be from much earlier runs
+                        if call_id and call_id in log_msg:
+                            # This log definitely belongs to our execution
+                            logger.info(f"{log_msg}")
+                        elif (
+                            log_age < 300
+                        ):  # Only show logs from last 5 minutes
+                            # This log is recent enough to likely be ours
+                            logger.info(f"{log_msg}")
+                        # Else: skip old logs
```
strickvl: This bit seems maybe like it could use more attention. Also are you sure that we wouldn't have time zone mismatches here etc?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,642 @@
...
+            pipeline_resource_dict = pipeline_resource_settings.model_dump()
+        else:
+            pipeline_resource_dict = pipeline_resource_settings
+        resource_settings = ResourceSettings.model_validate(
+            pipeline_resource_dict
+        )
+    else:
+        # Fallback to first step's resource settings if no pipeline-level resources
+        if deployment.step_configurations:
+            first_step = list(deployment.step_configurations.values())[0]
```
strickvl: what is the first step is unrepresentative, though?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+
+**Option 2: Using Modal API token (recommended for production)**
+
+```shell
+# Register the orchestrator with explicit credentials
+zenml orchestrator register <ORCHESTRATOR_NAME> \
+    --flavor=modal \
+    --token=<MODAL_TOKEN> \
+    --workspace=<MODAL_WORKSPACE> \
+    --synchronous=true
```
strickvl, line 106: I think this should use `--token-id` and `--token-secret` separately as per the code?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+
+# Fire-and-forget execution
+modal_settings = ModalOrchestratorSettings(
+    synchronous=False
+)
+```
+
+### Authentication with different environments
+
+For production deployments, you can specify different Modal environments:
```
strickvl, line 533: Maybe could have a little info box in this section (or maybe even above, linking down here) to say that you might want to have two different stacks, each associated with a different modal environment, one for prod and the other for development etc etc.

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+
+The Modal orchestrator is designed for speed and cost-effectiveness, running entire pipelines in single serverless functions to minimize cold starts and optimize resource utilization.
+
+{% hint style="warning" %}
+This component is only meant to be used within the context of a [remote ZenML deployment scenario](https://docs.zenml.io/getting-started/deploying-zenml/). Usage with a local ZenML deployment may lead to unexpected behavior!
+{% endhint %}
+
+## When to use it
+
+You should use the Modal orchestrator if:
```
strickvl, line 17: should we have something at the end of this section around the tradeoffs of using modal and when it might not be best suited etc?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+* The ZenML `modal` integration installed. If you haven't done so, run:
+  ```shell
+  zenml integration install modal
+  ```
+* [Docker](https://www.docker.com) installed and running.
+* A [remote artifact store](../artifact-stores/README.md) as part of your stack.
+* A [remote container registry](../container-registries/README.md) as part of your stack.
+* Modal CLI installed and authenticated:
+  ```shell
+  pip install modal
```
strickvl: do you need the pip install if you have the integration already installed?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+python file_that_runs_a_zenml_pipeline.py
+```
+
+### Modal UI
+
+Modal provides an excellent web interface where you can monitor your pipeline runs in real-time, view logs, and track resource usage.
+
+You can access the Modal dashboard at [modal.com/apps](https://modal.com/apps) to see your running and completed functions.
+
+### Configuration overview
```
strickvl, line 130: do you want to link to the step operator somewhere and explain at least why someone would want to use one over the other, and which is better for what etc?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+
+# Granular execution - each step separate (useful for debugging)
+modal_settings = ModalOrchestratorSettings(
+    execution_mode="per_step"
+)
+```
+
+### Using GPUs
+
+Modal makes it easy to use GPUs for your ML workloads. Use `ResourceSettings` to specify the number of GPUs and `ModalOrchestratorSettings` to specify the GPU type:
```
strickvl, line 421: Do we need to say something here about base images + ensuring that the base image has the relevant CUDA kernel and other things set up?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+@pipeline(
+    settings={
+        "orchestrator": modal_settings
+    }
+)
+def my_pipeline():
+    ...
+```
+
+This ensures your pipelines start executing immediately without waiting for container initialization.
```
strickvl: Do we need to say something about how this is implemented / what the cost implications of this are?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Getting help
+
+- Check the [Modal documentation](https://modal.com/docs) for platform-specific issues
+- Monitor your functions in the [Modal dashboard](https://modal.com/apps)
+- Use `zenml logs` to view detailed pipeline execution logs
+
+For more information and a full list of configurable attributes of the Modal orchestrator, check out the [SDK Docs](https://sdkdocs.zenml.io/latest/integration_code_docs/integrations-modal.html#zenml.integrations.modal.orchestrators).
+
+<figure><img src="https://static.scarf.sh/a.png?x-pxid=f0b4f458-0a54-4fcd-aa95-d5ee424815bc" alt="ZenML Scarf"><figcaption></figcaption></figure>
```
strickvl:
```suggestion
```

src/zenml/integrations/modal/__init__.py
```diff
@@ -11,25 +11,26 @@
...
+MODAL_ORCHESTRATOR_FLAVOR = "modal"
 MODAL_STEP_OPERATOR_FLAVOR = "modal"
 
 
 class ModalIntegration(Integration):
     """Definition of Modal integration for ZenML."""
 
     NAME = MODAL
-    REQUIREMENTS = ["modal>=0.64.49,<1"]
+    REQUIREMENTS = ["modal>=1"]
```
strickvl, line 33: Have you tested that this doesn't break the step operator?

src/zenml/integrations/modal/__init__.py
```diff
@@ -38,8 +39,15 @@ def flavors(cls) -> List[Type[Flavor]]:
...
+            ModalOrchestratorFlavor,
+            ModalStepOperatorFlavor,
+        )
 
-        return [ModalStepOperatorFlavor]
+        return [ModalOrchestratorFlavor, ModalStepOperatorFlavor]
+
+
+# Import utils module to make it discoverable by mypy
+from zenml.integrations.modal import utils  # noqa: F401
```
strickvl: ?

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,177 @@
...
+from typing import TYPE_CHECKING, Optional, Type
+
+from zenml.config.base_settings import BaseSettings
+from zenml.orchestrators import BaseOrchestratorConfig, BaseOrchestratorFlavor
+from zenml.utils.secret_utils import SecretField
+
+if TYPE_CHECKING:
+    from zenml.integrations.modal.orchestrators import ModalOrchestrator
+
+MODAL_ORCHESTRATOR_FLAVOR = "modal"
```
strickvl:
```suggestion
```

Instead import from the init where you've already defined this.

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,177 @@
...
+        return self.generate_default_sdk_docs_url()
+
+    @property
+    def logo_url(self) -> str:
+        """A url to represent the flavor in the dashboard.
+
+        Returns:
+            The flavor logo.
+        """
+        return "https://public-flavor-logos.s3.eu-central-1.amazonaws.com/orchestrator/modal.png"
```
strickvl, line 211: check that this exists? probably just can copy it over from the step operator logo we already have?

src/zenml/integrations/modal/flavors/modal_step_operator_flavor.py
```diff
@@ -36,20 +37,40 @@ class ModalStepOperatorSettings(BaseSettings):
...
         cloud: The cloud provider to use for the step execution.
+        environment: The Modal environment to use for the step execution.
+        timeout: Maximum execution time in seconds (default 24h).
     """
 
     gpu: Optional[str] = None
     region: Optional[str] = None
     cloud: Optional[str] = None
+    environment: Optional[str] = None
+    timeout: int = 86400  # 24 hours (Modal's maximum)
```
strickvl: Do we need / want to update the step operator docs with a section on environments etc like we have in the orchestrator docs?

src/zenml/integrations/modal/utils.py
strickvl: We should check / just make this really explicit in the docs that this is how we assume this

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,642 @@
...
+    # Wait a moment for the function to start before beginning log streaming
+    # This helps avoid capturing old logs from previous runs
+    time.sleep(1)
+
+    # Start log streaming in a separate thread
+    log_stream_active = threading.Event()
+    log_stream_active.set()
+    start_time = time.time()
+
+    def stream_logs() -> None:
```
strickvl: Function in function smells a bit wrong and also wondering if we should instead use their Python SDK to stream the logs? https://github.com/modal-labs/modal-client/blob/4177d0b994ac69e01ada7d7a96655c9dcaae570e/modal/cli/utils.py#L24

Possibly something for down the line, though the func-in-func seems off.

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,467 @@
...
+        ModalOrchestratorSettings,
+    )
+
+from zenml.integrations.modal.flavors.modal_orchestrator_flavor import (
+    ModalExecutionMode,
+)
+
+if TYPE_CHECKING:
+    from zenml.models import PipelineDeploymentResponse, PipelineRunResponse
+    from zenml.models.v2.core.pipeline_deployment import PipelineDeploymentBase
```
strickvl: combine the 'if TYPE_CHECKING' parts?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
...
+def second_step():
+    # Uses same pipeline-level resource configuration
+    ...
+```
+
+### Execution modes
+
+The Modal orchestrator supports two execution modes:
+
+1. **`pipeline` (default)**: Runs the entire pipeline in a single Modal function for maximum speed and cost efficiency
```
strickvl: Not sure I understand why this `pipeline` option is max speed. Isn't it running everything sequentially in the same container? Wouldn't running things in parallel in separate Modal function calls run faster?

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,177 @@
...
+    max_containers: Optional[int] = 10  # Allow up to 10 concurrent containers
+    execution_mode: ModalExecutionMode = (
+        ModalExecutionMode.PIPELINE
+    )  # Default to fastest mode
+    synchronous: bool = (
+        True  # Wait for completion (True) or fire-and-forget (False)
+    )
+    app_warming_window_hours: float = (
+        2.0  # Default 2-hour window for app reuse
+    )
```
strickvl: I think we don't state in the docs anywhere that there is a two-hour default set for this warm window. I think we should, esp just as a note for anyone doing dev work on their pipeline, but also maybe that for prod workflows (esp anything automated / triggers etc) you might want to lower this value quite a bit?

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,375 @@
+---
+description: Orchestrating your pipelines to run on Modal's serverless cloud platform.
+---
+
+# Modal Orchestrator
+
+Using the ZenML `modal` integration, you can orchestrate and scale your ML pipelines on [Modal's](https://modal.com/) serverless cloud platform with minimal setup and maximum efficiency.
+
+The Modal orchestrator is designed for speed and cost-effectiveness, running entire pipelines in single serverless functions to minimize cold starts and optimize resource utilization.
+
```
strickvl, line 10: Maybe some representative screenshot of the Modal UI in here to make the docs a bit friendlier?

htahir1: i think its fine without

src/zenml/integrations/modal/__init__.py
htahir1: have not checked, thats one thing id love your help on @strickvl

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
htahir1: exists

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+class ModalOrchestratorConfig(
+    BaseOrchestratorConfig, ModalOrchestratorSettings
+):
+    """Modal orchestrator config optimized for BLAZING FAST execution.
```
schustmi:
```suggestion
    """Modal orchestrator config.
```

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,177 @@
...
+        environment: Modal environment name (optional).
+
+    Note: If token_id and token_secret are not provided, falls back to
+    Modal's default authentication (~/.modal.toml).
+    """
+
+    token_id: Optional[str] = SecretField(default=None)
+    token_secret: Optional[str] = SecretField(default=None)
+    workspace: Optional[str] = None
+    environment: Optional[str] = None
```
schustmi: Already exists on superclass

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,177 @@
...
+        return True
+
+    @property
+    def is_synchronous(self) -> bool:
+        """Whether the orchestrator runs synchronous or not.
+
+        Returns:
+            True since the orchestrator waits for completion.
+        """
+        return True
```
schustmi: This is wrong. All other orchestrators have the synchronous attribute in the config and return that value here I think?

src/zenml/integrations/modal/flavors/modal_step_operator_flavor.py
```diff
@@ -36,20 +37,40 @@ class ModalStepOperatorSettings(BaseSettings):
...
+        environment: Modal environment name (optional).
+
+    Note: If token_id and token_secret are not provided, falls back to
+    Modal's default authentication (~/.modal.toml).
+    """
+
+    token_id: Optional[str] = SecretField(default=None)
+    token_secret: Optional[str] = SecretField(default=None)
+    workspace: Optional[str] = None
+    environment: Optional[str] = None
```
schustmi: Defined on super class

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,464 @@
...
+    Tuple,
+    Type,
+    cast,
+)
+from uuid import uuid4
+
+try:
+    import modal
+except ImportError:
+    modal = None  # type: ignore
```
schustmi: ? And then it crashes with a worse error below?

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,464 @@
...
+    Raises:
+        Exception: If step execution fails.
+    """
+    # Get pipeline run ID for debugging/logging
+    pipeline_run_id = os.environ.get("ZENML_PIPELINE_RUN_ID", "unknown")
+
+    logger.info(
+        f"Running step '{step_name}' remotely (pipeline run: {pipeline_run_id})"
+    )
+    sys.stdout.flush()
```
schustmi: ?

strickvl (approved):
LGTM, assuming tests pass.

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,464 @@
...
+    Args:
+        step_name: Name of the step to execute.
+        deployment_id: ID of the pipeline deployment.
+        orchestrator_run_id: ID of the orchestrator run.
+
+    Raises:
+        Exception: If step execution fails.
+    """
+    # Get pipeline run ID for debugging/logging
+    pipeline_run_id = os.environ.get("ZENML_PIPELINE_RUN_ID", "unknown")
```
schustmi: What is this, and who sets it?

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,464 @@
...
+        )
+        sys.stdout.flush()
+
+        # Create the entrypoint arguments
+        args = StepEntrypointConfiguration.get_entrypoint_arguments(
+            step_name=step_name, deployment_id=deployment_id
+        )
+
+        # Create the configuration and run the step
+        config = StepEntrypointConfiguration(arguments=args)
```
schustmi: This class is not meant to be used in that way, no idea if this will lead to issues

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,464 @@
...
+
+        Args:
+            deployment: The pipeline deployment.
+
+        Returns:
+            A list of Docker build configurations.
+        """
+        # Use the standard containerized orchestrator build logic
+        # This ensures ZenML builds the image with all pipeline code
+        return super().get_docker_builds(deployment)
```
schustmi: Remove

schustmi (changes requested):

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,464 @@
...
+
+        Raises:
+            RuntimeError: If Modal is not installed or if a step fails.
+            Exception: If pipeline execution fails.
+        """
+        _ = placeholder_run  # Mark as intentionally unused
+        if modal is None:
+            raise RuntimeError(
+                "Required dependencies not installed. Please install with: pip install modal"
+            )
```
safoinme:
```suggestion
        if modal is None:
            raise ImportError(
                "Modal is not installed. Please install with: pip install 'modal>=1.0'"
            )
```

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,695 @@
...
+            process = subprocess.Popen(
+                cmd,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                text=True,
+                bufsize=1,  # Line buffered
+                universal_newlines=True,
+            )
+
+            # Stream logs line by line
```
safoinme: I think it's probably better to use subproccess with context manager to avoid resource leak

```suggestion
            with subprocess.Popen(
                cmd,
                stdout=subprocess.PIPE,
                stderr=subprocess.STDOUT,
                text=True,
                bufsize=1,  # Line buffered
                universal_newlines=True,
            ) as process:
```

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: No this is intentional

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: agreed

src/zenml/integrations/modal/__init__.py
htahir1: Checked it and it works

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,389 @@
...
+        self._setup_modal_client()
+
+        # Generate orchestrator run ID and include pipeline run ID for isolation
+        orchestrator_run_id = str(uuid4())
+
+        environment[ENV_ZENML_MODAL_ORCHESTRATOR_RUN_ID] = orchestrator_run_id
+
+        # Pass pipeline run ID for proper isolation (following other orchestrators' pattern)
+        if placeholder_run:
+            environment["ZENML_PIPELINE_RUN_ID"] = str(placeholder_run.id)
```
htahir1, line 188: @schustmi can i do this? Does this work?

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,368 @@
...
+    if step_settings:
+        step_modal_settings = ModalOrchestratorSettings.model_validate(
+            step_settings.model_dump() if step_settings else {}
+        )
+        # Merge with pipeline-level settings
+        for key, value in step_modal_settings.model_dump(
+            exclude_unset=True
+        ).items():
+            if value is not None:
+                setattr(settings, key, value)
```
strickvl: I think with this implementation, settings made for one step would unintentionally affect subsequent steps? (because the pipeline-level settings object is changed in-place for each step).

So a settings override for a single step will persist for all subsequent steps. Maybe create a deep copy of the pipeline-level settings within run_step_on_modal before applying step-specific overrides?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,415 @@
...
+    Args:
+        image_name: Base Docker image name
+        environment: Environment variables
+
+    Returns:
+        Hash string to use as cache key
+    """
+    # Create deterministic string from inputs
+    cache_input = f"{image_name}|{sorted(environment.items())}"
+    return hashlib.sha256(cache_input.encode()).hexdigest()[:12]
```
strickvl: I think this function can be removed (and where it's called?). The `_generate_image_cache_key` function  is called, but its result is only logged and not used.

docs/book/component-guide/orchestrators/modal.md
```diff
@@ -0,0 +1,705 @@
...
+   - `execution_mode` - Execution strategy: "pipeline" (default) or "per_step"
+   - `max_parallelism` - Maximum concurrent steps (for "per_step" mode)
+   - `timeout` - Maximum execution time in seconds
+   - `synchronous` - Wait for completion (True) or fire-and-forget (False)
+
+{% hint style="info" %}
+**GPU Configuration**: Use `ResourceSettings.gpu_count` to specify how many GPUs you need, and `ModalOrchestratorSettings.gpu` to specify what type of GPU. Modal will combine these automatically (e.g., `gpu_count=2` + `gpu="A100"` becomes `"A100:2"`).
+{% endhint %}
+
+### Configuration Overview
```
strickvl: The docs have two sections with the header `### Configuration overview`...

strickvl (changes requested):

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,389 @@
...
+            deployment_id=str(deployment.id),
+            execution_mode=execution_mode,
+            run_id=run_id,
+        )
+        logger.info(f"Sandbox tags: {sandbox_tags}")
+
+        with modal.enable_output():
+            # Create sandbox with the entrypoint command
+            # Note: Modal sandboxes inherit environment from the image
+            sb = await modal.Sandbox.create.aio(
```
safoinme: why is async interface used to create sandboxes? instead of just `modal.Sandbox.create`

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
htahir1: Good one, fixed

src/zenml/integrations/modal/utils.py
htahir1: Done

docs/book/component-guide/orchestrators/modal.md
htahir1: Changed

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: - await modal.Sandbox.create.aio() properly integrates with the asyncio event loop
    - modal.Sandbox.create() would make blocking system calls that don't yield control back to the event

  In this specific case, since we're using asyncio.run() and only creating one sandbox at a time in
  pipeline mode, the practical difference might seem minimal. However:

  - Per-step mode: When creating multiple sandboxes concurrently, async is crucial
  - Future extensibility: Allows for potential concurrent operations
  - Modal's design: Modal's async API is optimized for better performance and proper resource management

strickvl (changes requested):
I think you need maybe to add modal to the full list of orchestrators in the docs page?

strickvl (approved):

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,389 @@
...
+        )
+        logger.info(f"Using execution mode: {execution_mode}")
+
+        # Get resource settings from pipeline configuration
+        resource_settings = get_resource_settings_from_deployment(
+            deployment, RESOURCE_SETTINGS_KEY
+        )
+
+        # Build Modal image
+        zenml_image = self._build_modal_image(deployment, stack, environment)
```
htahir1: Comment: Dont build each time, use metadata of build id to track whether this build is already in modal or not

htahir1:
@coderabbitai summary

htahir1:
@coderabbitai review

htahir1:
@coderabbitai review

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,169 @@
...
+            "A100"). Use ResourceSettings.gpu_count to specify the number of GPUs.
+        region: The region to use for the pipeline execution.
+        cloud: The cloud provider to use for the pipeline execution.
+        modal_environment: The Modal environment to use for the pipeline execution.
+        timeout: Maximum execution time in seconds (default 24h).
+        execution_mode: Execution mode - PIPELINE (fastest) or PER_STEP (granular).
+        max_parallelism: Maximum number of parallel sandboxes (for PER_STEP mode).
+        synchronous: Wait for completion (True) or fire-and-forget (False).
+    """
+
```
schustmi, line 62: I think having an option to specify the sandbox name would be nice?

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
```diff
@@ -0,0 +1,169 @@
...
+    """Modal orchestrator settings.
+
+    Attributes:
+        gpu: The type of GPU to use for the pipeline execution (e.g., "T4",
+            "A100"). Use ResourceSettings.gpu_count to specify the number of GPUs.
+        region: The region to use for the pipeline execution.
+        cloud: The cloud provider to use for the pipeline execution.
+        modal_environment: The Modal environment to use for the pipeline execution.
+        timeout: Maximum execution time in seconds (default 24h).
+        execution_mode: Execution mode - PIPELINE (fastest) or PER_STEP (granular).
```
schustmi: Needs more explanation IMO, also the naming I'm not sure about.
Both modes execute your pipeline, but `PIPELINE` uses a single sandbox and runs sequentially, while `PER_STEP` spins up a separate sandbox per step.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -91,40 +77,8 @@ def validator(self) -> Optional[StackValidator]:
...
-
-        return StackValidator(
-            required_components={
-                StackComponentType.CONTAINER_REGISTRY,
-                StackComponentType.IMAGE_BUILDER,
-            },
-            custom_validation_function=_validate_remote_components,
-        )
+        validator: StackValidator = create_modal_stack_validator()
+        return validator
```
schustmi:
```suggestion
        return create_modal_stack_validator()
```

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @property
+    def validator(self) -> Optional[StackValidator]:
+        """Ensures there is a container registry and artifact store in the stack.
+
+        Returns:
+            A `StackValidator` instance.
+        """
+        validator: StackValidator = create_modal_stack_validator()
+        return validator
```
schustmi:
```suggestion
        return create_modal_stack_validator()
```

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,206 @@
...
+                environment.
+            placeholder_run: An optional placeholder run for the deployment.
+
+        Raises:
+            Exception: If pipeline execution fails.
+
+        Returns:
+            None if the pipeline is executed synchronously, otherwise an iterator of metadata dictionaries.
+        """
+        if deployment.schedule:
```
schustmi: We fail earlier when trying to run with a schedule, so no need to have this warning

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # Get settings from pipeline configuration
+        settings = cast(
+            "ModalOrchestratorSettings", self.get_settings(deployment)
+        )
+
+        # Check execution mode
+        execution_mode = getattr(
+            settings, "execution_mode", ModalExecutionMode.PIPELINE
+        )
```
schustmi:
```suggestion
        execution_mode = settings.execution_mode
```

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,206 @@
...
+                    if placeholder_run
+                    else None,
+                    synchronous=synchronous,
+                )
+            )
+        except Exception as e:
+            logger.error(f"Pipeline execution failed: {e}")
+            raise
+
+        logger.info("✅ Pipeline execution completed successfully")
```
schustmi: This is only true if the orchestrator is running in sync mode

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,206 @@
...
+                "use of schedules. The `schedule` will be ignored "
+                "and the pipeline will be run immediately."
+            )
+
+        # Setup Modal authentication
+        self._setup_modal_client()
+
+        # Generate orchestrator run ID and include pipeline run ID for isolation
+        orchestrator_run_id = str(uuid4())
+        environment[ENV_ZENML_MODAL_ORCHESTRATOR_RUN_ID] = orchestrator_run_id
```
schustmi: There is no need to pass this to the entrypoint if it is simply a random UUID. In that case, we can simply generate it in the entrypoint itself.

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,206 @@
...
+        )
+
+        # Check execution mode
+        execution_mode = getattr(
+            settings, "execution_mode", ModalExecutionMode.PIPELINE
+        )
+        logger.info(
+            f"🚀 Executing pipeline with Modal ({execution_mode.lower()} mode)"
+        )
+
```
schustmi, line 198: If the execution mode is pipeline, we do not allow per-step images (as they cannot be used). I suggest you overwrite the `get_docker_builds(...)` method as follows:
- Call the super implementation
- Check if there is any build configuration in the result that has `key="orchestrator"` and a non-None `step_name`. In that case, raise an error.

schustmi (changes requested):

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: fixed

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: fixed

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: Done

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: Done

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
htahir1: Done

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
htahir1: sandboxes dont have names and the app names are equal to the pipeline names

htahir1 (commented):
@schustmi Fixed your comments

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
schustmi: I think this reply is on the wrong comment. What I meant was actually the app name. I would equate this with specifying a kubernetes namespace or pod name, both of which we allow users to specify but fallback to defaults like the pipeline name if no value is given.

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,258 @@
...
+        if mode == ModalExecutionMode.PIPELINE:
+            for build in builds:
+                if (
+                    build.key == ORCHESTRATOR_DOCKER_IMAGE_KEY
+                    and build.step_name is not None
+                ):
+                    raise ValueError(
+                        f"Per-step Docker settings are not supported in PIPELINE "
+                        f"execution mode. Step '{build.step_name}' has custom Docker "
+                        f"settings but will be ignored since the entire pipeline runs "
```
schustmi, line 119: It will not be ignored, but instead the run will fail right here with this exception

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,258 @@
...
+        # Create sandbox executor
+        executor = ModalSandboxExecutor(
+            deployment=deployment,
+            stack=stack,
+            environment=environment,
+            settings=settings,
+        )
+
+        # Determine if we should wait for completion
+        synchronous = (
```
schustmi: `synchronous = settings.synchronous`

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -231,7 +158,7 @@ async def run_sandbox() -> asyncio.Future[None]:
                         cloud=settings.cloud,
                         region=settings.region,
                         app=app,
-                        timeout=86400,  # 24h, the max Modal allows
+                        timeout=settings.timeout,
```
schustmi, line 161: Feel free to ignore as part of this PR as well, but seems like this could also be refactored to use the same code that runs a step of a pipeline in a new sandbox?

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+        ModalOrchestratorSettings,
+        orchestrator.get_settings(deployment),
+    )
+
+    # Setup Modal client
+    setup_modal_client(
+        token_id=orchestrator.config.token_id,
+        token_secret=orchestrator.config.token_secret,
+        workspace=orchestrator.config.workspace,
+        environment=orchestrator.config.modal_environment,
```
schustmi, line 326: The `modal_environment` is not a config option and can therefore be overwritten using the settings. So I'm guessing we should probably fetch this from the pipeline settings instead?

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,258 @@
...
+
+        return builds
+
+    def _setup_modal_client(self) -> None:
+        """Setup Modal client with authentication."""
+        setup_modal_client(
+            token_id=self.config.token_id,
+            token_secret=self.config.token_secret,
+            workspace=self.config.workspace,
+            environment=self.config.modal_environment,
```
schustmi, line 132: Same as for the entrypoint, this should be fetched from the settings instead.

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
htahir1: @schustmi ok added

src/zenml/integrations/modal/flavors/modal_orchestrator_flavor.py
htahir1: done

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    logger.info(f"🚀 Executing {len(pipeline_dag)} pipeline steps")
+
+    # Run using ThreadedDagRunner with optimized execution
+    ThreadedDagRunner(
+        dag=pipeline_dag,
+        run_fn=run_step_wrapper,
+        finalize_fn=finalize_wrapper,
+        max_parallelism=getattr(pipeline_settings, "max_parallelism", None),
```
schustmi, line 229: `pipeline_settings.max_parallelism`

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+    if deployment.build is None:
+        raise ValueError(
+            "Deployment build is None, cannot prepare image cache"
+        )
+
+    # Add pipeline-level image if needed
+    pipeline_image = ModalOrchestrator.get_image(deployment=deployment)
+    build_id = str(deployment.build.id)
+    pipeline_cache_key = (
+        f"{build_id}_pipeline_{str(hash(pipeline_image))[-8:]}"
```
schustmi: Can you explain the logic behind this weird way to get a cache key?

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+    # Collect all unique images needed across all steps
+    unique_images: Dict[str, str] = {}  # cache_key -> image_name
+
+    # Check if deployment has a build
+    if deployment.build is None:
+        raise ValueError(
+            "Deployment build is None, cannot prepare image cache"
+        )
+
+    # Add pipeline-level image if needed
```
schustmi: Why do we need this? We're already running in the sandbox started for the pipeline, from now on we will only need images for each step sandbox, no?

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+        args: Parsed command line arguments.
+        orchestrator_run_id: The orchestrator run ID.
+    """
+    try:
+        client = Client()
+        deployment = client.get_deployment(args.deployment_id)
+
+        # Fetch the pipeline run
+        list_args: Dict[str, Any] = {}
+        if args.run_id:
```
schustmi, line 251: If schedules are not supported, there will always be a placeholder run, which means you can make this argument non-optional and it's guaranteed to exist here.

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+        dag=pipeline_dag,
+        run_fn=run_step_wrapper,
+        finalize_fn=finalize_wrapper,
+        max_parallelism=getattr(pipeline_settings, "max_parallelism", None),
+    ).run()
+
+
+def finalize_run(
+    node_states: Dict[str, NodeStatus],
+    args: argparse.Namespace,
```
schustmi, line 235: Passing this just hides any potential linting issue. Passing the the `deployment_id` and `run_id` explicitly here is much better IMO

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+    """Finalize the run by updating step and pipeline run statuses.
+
+    Args:
+        node_states: The states of the nodes.
+        args: Parsed command line arguments.
+        orchestrator_run_id: The orchestrator run ID.
+    """
+    try:
+        client = Client()
+        deployment = client.get_deployment(args.deployment_id)
```
schustmi, line 247: We've fetched this before, no need to do it again

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+        pipeline_failed = False
+
+        for step_name, node_state in node_states.items():
+            if node_state != NodeStatus.FAILED:
+                continue
+
+            pipeline_failed = True
+
+            # Mark failed step runs as failed
+            step_run = pipeline_run.steps.get(step_name)
```
schustmi, line 276: Accessing this fetches **all** steps of the pipeline, which means if you have lots of failed nodes in a huge pipeline with 1000 steps, you're fetching up to 1.000.000 steps. Instead, this should be accessed once outside the for loop

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+from zenml.logger import get_logger
+from zenml.stack import Stack, StackValidator
+
+logger = get_logger(__name__)
+
+# Common environment variable for Modal orchestrator run ID
+ENV_ZENML_MODAL_ORCHESTRATOR_RUN_ID = "ZENML_MODAL_ORCHESTRATOR_RUN_ID"
+
+
+class ModalAuthenticationError(Exception):
```
schustmi: Maybe I missed it, but is this used anywhere?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+            )
+
+        # Only token ID provided
+        os.environ["MODAL_TOKEN_ID"] = token_id
+        logger.debug("Using platform token ID from config")
+        logger.warning(
+            "Only token ID provided. Make sure MODAL_TOKEN_SECRET is set "
+            "or platform authentication may fail."
+        )
+        logger.debug(f"Token ID starts with: {token_id[:5]}...")
```
schustmi, line 88: No need to log any information about tokens, not even a prefix IMO

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+        required_components={
+            StackComponentType.CONTAINER_REGISTRY,
+            StackComponentType.IMAGE_BUILDER,
+        },
+        custom_validation_function=_validate_remote_components,
+    )
+
+
+def get_resource_settings_from_deployment(
+    deployment: Any,
```
schustmi: Why `Any`?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+    """Extract resource settings from pipeline deployment.
+
+    Args:
+        deployment: The pipeline deployment.
+        resource_settings_key: Key to look for resource settings.
+
+    Returns:
+        ResourceSettings object with the configuration.
+    """
+    pipeline_resource_settings: Union[Dict[str, Any], Any] = (
```
schustmi: This will return you `ResourceSettings` as a type, nothing else.

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+    """Create a stack validator for Modal components.
+
+    Returns:
+        A StackValidator that ensures remote artifact store and container registry.
+    """
+
+    def _validate_remote_components(stack: "Stack") -> Tuple[bool, str]:
+        if stack.artifact_store.config.is_local:
+            return False, (
+                "Serverless components run code remotely and "
```
schustmi, line 300: `Modal components`?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+        pipeline_name: Name of the pipeline
+        deployment_id: ZenML deployment ID
+        execution_mode: Execution mode (PIPELINE or PER_STEP)
+        step_name: Step name (for PER_STEP mode)
+        run_id: Pipeline run ID
+
+    Returns:
+        Dictionary of tags for the sandbox
+    """
+    tags = {
```
schustmi, line 271: Not sure about modal, but usually tags are only allowed to have a certain format, and we should probably sanitize them?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+        environment=None,  # No environment variables for cached images
+    )
+
+    # Store the image in the app for future use
+    setattr(app, image_name_key, zenml_image)
+
+    return zenml_image
+
+
+def build_modal_image(
```
schustmi: This function is useless, it calls a "helper" with the same arguments.

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+    Args:
+        resource_settings: The resource settings.
+
+    Returns:
+        Tuple of (cpu_count, memory_mb) with validated values.
+    """
+    # Get CPU count with validation
+    cpu_count: Optional[int] = None
+    if resource_settings.cpu_count is not None:
+        cpu_count = int(resource_settings.cpu_count)
```
schustmi: Does modal not allow partial cpus, e.g. `1.5`?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+
+    return zenml_image
+
+
+def get_or_build_modal_image(
+    image_name: str,
+    stack: "Stack",
+    pipeline_name: str,
+    build_id: str,
+    app: Any,
```
schustmi: Why `Any`?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+
+    # Build new image using shared helper
+    zenml_image = _build_modal_image_from_registry(
+        image_name=image_name,
+        stack=stack,
+        environment=None,  # No environment variables for cached images
+    )
+
+    # Store the image in the app for future use
+    setattr(app, image_name_key, zenml_image)
```
schustmi: Ugly hack, why do we need to set this on the modal app?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+        stack: The ZenML stack containing container registry.
+        pipeline_name: The pipeline name for caching.
+        build_id: The build ID for the image key.
+        app: The Modal app to store/retrieve images.
+
+    Returns:
+        The configured Modal image.
+    """
+    # Try to get existing image from the app
+    image_name_key = f"zenml_image_{build_id}"
```
schustmi: How is this even supposed to work? This key is the same for **all** images in the build, which are certainly not referring to the same docker image.

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+        try:
+            memory_value = resource_settings.get_memory(ByteUnit.MB)
+            if memory_value is not None:
+                memory_mb = int(memory_value)
+                # Validate memory is reasonable
+                if memory_mb <= 0:
+                    logger.warning(f"Invalid memory {memory_mb}MB, ignoring.")
+                    memory_mb = None
+                elif memory_mb < 128:  # Less than 128MB seems too low
+                    logger.warning(
```
schustmi: Why would it not be intentional, the user explicitly specified this?

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+    # Create Modal secret for registry authentication
+    registry_secret = modal.Secret.from_dict(
+        {
+            "REGISTRY_USERNAME": docker_username,
+            "REGISTRY_PASSWORD": docker_password,
+        }
+    )
+
+    # Build Modal image from the ZenML-built image
+    logger.debug(f"Building Modal image from base: {image_name}")
```
schustmi: Duplicate

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+    logger.debug(f"Creating Modal image from base: {image_name}")
+    zenml_image = (
+        modal.Image.from_registry(
+            image_name, secret=registry_secret
+        ).pip_install("modal")  # Install Modal in the container
```
schustmi: `modal` is installed in the container, as part of the ZenML image build that happened for the stack.

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,490 @@
...
+            logger.warning(f"Failed to parse memory setting: {e}")
+            memory_mb = None
+
+    return cpu_count, memory_mb
+
+
+def _build_modal_image_from_registry(
+    image_name: str,
+    stack: "Stack",
+    environment: Optional[Dict[str, str]] = None,
```
schustmi, line 153: Didn't we say we don't build these into the image anymore?

If you do, your image reuse doesn't work, as these environment variables contain e.g. credentials to access the ZenML server. If they get reused, they point to potentially a wrong ZenML user, are credentials for the wrong user, ...

schustmi (changes requested):

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+    """Handles execution of ZenML pipelines and steps in Modal sandboxes."""
+
+    def __init__(
+        self,
+        deployment: "PipelineDeploymentResponse",
+        stack: "Stack",
+        environment: Dict[str, str],
+        settings: ModalOrchestratorSettings,
+        shared_image_cache: Optional[Dict[str, Any]] = None,
+        shared_app: Optional[Any] = None,
```
schustmi: No `Any`

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+        run_id: Optional[str] = None,
+        synchronous: bool = True,
+    ) -> None:
+        """Execute the entire pipeline in a single sandbox.
+
+        Args:
+            run_id: The pipeline run ID.
+            synchronous: Whether to wait for completion.
+        """
+        logger.debug("Executing entire pipeline in single sandbox")
```
schustmi, line 433: This is called no matter which mode the orchestrator is running in.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+
+        # Build entrypoint command
+        command = (
+            ModalOrchestratorEntrypointConfiguration.get_entrypoint_command()
+        )
+        from uuid import UUID
```
schustmi: Import at the top

src/zenml/integrations/modal/orchestrators/modal_orchestrator.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Determine if we should wait for completion
+        synchronous = (
+            settings.synchronous
+            if hasattr(settings, "synchronous")
+            else self.config.synchronous
+        )
+
+        # Submit the pipeline
+        run_id = str(placeholder_run.id) if placeholder_run else None
```
schustmi: You're converting this to a string here, only to convert it back to a UUID in the `executor.execute_pipeline` method.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+        else:
+            # Create Modal app for this pipeline
+            self.app_name = get_modal_app_name(settings, deployment)
+            self.app = modal.App.lookup(
+                self.app_name,
+                create_if_missing=True,
+                environment_name=settings.modal_environment,
+            )
+
+    def _build_entrypoint_command(
```
schustmi: 10 lines of code for a `+`? Can be removed I thjink

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+            Complete command without environment variables.
+        """
+        return base_command + args
+
+    # ---------------------------------------------------------------------
+    # Resource utilities
+    # ---------------------------------------------------------------------
+
+    @staticmethod
+    def _to_resource_settings(data: Any | None) -> ResourceSettings:
```
schustmi: This whole method is insanity, and something similar also exists in the utils file. The input to this isn't `Any`, you know exactly what it could be and if we know this we don't have to write most of this code here.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+            return ResourceSettings.model_validate(data)
+        except Exception as e:
+            logger.warning(
+                "Failed to validate resource settings %s – %s. Using default.",
+                data,
+                e,
+            )
+            return ResourceSettings()
+
+    def _get_step_settings(self, step_name: str) -> ModalOrchestratorSettings:
```
schustmi: No. The settings you receive from the deployment are already merged, nothing of this happens in orchestrator-specific code. Remove this entirely.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+                    if value is not None:
+                        pipeline_settings_dict[key] = value
+
+        # Create merged settings from the combined dictionary
+        merged_settings = ModalOrchestratorSettings.model_validate(
+            pipeline_settings_dict
+        )
+        return merged_settings
+
+    def _create_environment_secret(self) -> Optional[Any]:
```
schustmi: No `Any`.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+
+        # Create secret from environment variables
+        # Modal handles efficiency internally
+        # Cast to Dict[str, str | None] to match Modal's expected type
+        env_dict: Dict[str, Optional[str]] = {
+            k: v for k, v in self.environment.items()
+        }
+        return modal.Secret.from_dict(env_dict)
+
+    def _get_resource_settings(
```
schustmi, line 115: Method 17 to get resource settings -> No

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+            return ResourceSettings()
+
+        # Pipeline-level: delegate to existing util (already returns RS)
+        resource_settings = get_resource_settings_from_deployment(
+            self.deployment, RESOURCE_SETTINGS_KEY
+        )
+        logger.debug("Using pipeline-level resource settings")
+        return resource_settings
+
+    def _get_resource_config(
```
schustmi, line 155: I've seen something similar in the utils file

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
schustmi: Also, `Any` already includes `None`, and we do not use `|` in the type annotations yet. Not sure if our lowest supported python version even supports it yet, but the codebase is not using it.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+        self, step_name: Optional[str] = None
+    ) -> tuple[Optional[str], Optional[int], Optional[int]]:
```
schustmi, line 157: Uppercase in type annotations for consistency

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+                else:
+                    logger.warning(
+                        "Deployment build is None, cannot store image ID"
+                    )
+            else:
+                logger.warning("Image not hydrated after sandbox creation")
+        except Exception as e:
+            logger.warning(f"Failed to store image ID: {e}")
+
+    def _get_image_name(self, step_name: Optional[str] = None) -> str:
```
schustmi: This method is a 100% equivalent to `ModalOrchestrator.get_image`. Remove

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+        """
+        # Use build ID and step name to create unique cache key
+        # Include a hash of the image name for uniqueness
+        if self.deployment.build is None:
+            raise ValueError(
+                "Deployment build is None, cannot generate cache key"
+            )
+
+        build_id = str(self.deployment.build.id)
+        image_hash = str(hash(image_name))[-8:]  # Last 8 chars of hash
```
schustmi, line 416: I've seen this code in lots of other places. Remove in all except one.

src/zenml/integrations/modal/orchestrators/modal_orchestrator_entrypoint.py
```diff
@@ -0,0 +1,389 @@
...
+    shared_executor = ModalSandboxExecutor(
+        deployment=deployment,
+        stack=active_stack,
+        environment=environment,
+        settings=pipeline_settings,
+        shared_image_cache=shared_image_cache,
+        shared_app=shared_app,
+    )
+
+    def run_step_wrapper(step_name: str) -> None:
```
schustmi, line 199: It's a wrapper around a wrapper around a wrapper. Simplify

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+        command = StepEntrypointConfiguration.get_entrypoint_command()
+        args = StepEntrypointConfiguration.get_entrypoint_arguments(
+            step_name=step_name, deployment_id=self.deployment.id
+        )
+        entrypoint_command = self._build_entrypoint_command(command, args)
+
+        # Execute step sandbox
+        await self._execute_sandbox(
+            entrypoint_command=entrypoint_command,
+            mode="PER_STEP",
```
schustmi, line 467: Use enum

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+            )
+
+        build_id = str(self.deployment.build.id)
+        image_hash = str(hash(image_name))[-8:]  # Last 8 chars of hash
+        if step_name:
+            return f"{build_id}_{step_name}_{image_hash}"
+        else:
+            return f"{build_id}_pipeline_{image_hash}"
+
+    async def execute_pipeline(
```
schustmi, line 422: This is more like `start_pipeline_sandbox` right?

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+            f"Resource config for {step_name or 'pipeline'}: "
+            f"GPU={gpu_values}, CPU={cpu_count}, Memory={memory_mb}MB"
+        )
+
+        return gpu_values, cpu_count, memory_mb
+
+    def _prepare_modal_api_params(
+        self,
+        entrypoint_command: List[str],
+        image: Any,
```
schustmi, line 200: No `Any`

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+                # Stream output for better user experience
+                async for line in sb.stdout:
+                    print(line, end="")
+                await sb.wait.aio()
+            else:
+                logger.debug("Sandbox started asynchronously")
+
+            # Store the image ID for future caching after sandbox creation
+            # The image should be hydrated after being used in sandbox creation
+            await self._store_image_id(zenml_image)
```
schustmi: Should probably happen before the sync waiting?

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+
+    async def _store_image_id(self, zenml_image: Any) -> None:
+        """Store the image ID for future caching after sandbox creation.
+
+        Args:
+            zenml_image: The Modal image that was used.
+        """
+        try:
+            # After sandbox creation, the image should be hydrated
+            zenml_image.hydrate()
```
schustmi: No zenml image has this method, probably because this is a modal image instead. No `Any` type annotations.

src/zenml/integrations/modal/orchestrators/modal_sandbox_executor.py
```diff
@@ -0,0 +1,683 @@
...
+                deployment_id=self.deployment.id,
+                run_id=run_id_uuid,
+            )
+        )
+        entrypoint_command = self._build_entrypoint_command(command, args)
+
+        # Execute pipeline sandbox
+        await self._execute_sandbox(
+            entrypoint_command=entrypoint_command,
+            mode="PIPELINE",
```
schustmi, line 447: Actually the mode here can be either pipeline or step, because this launches a sandbox that either
- runs the entire pipeline
- only orchestrator the steps
So we should rather pass the actual mode here, or nothing at all. This is anyway only used for tags.

schustmi (changes requested):

htahir1:
Closing due to staleness

PR #3735, opened 2025-06-12
Kubernetes Orchestrator Refresh Functionality

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
@@ -206,8 +206,9 @@ def get_pod(
         The found pod object. None if it's not found.
     """
     try:
-        return core_api.read_namespaced_pod(name=pod_name, namespace=namespace)
-    except k8s_client.rest.ApiException as e:
+        pod = core_api.read_namespaced_pod(name=pod_name, namespace=namespace)
+        return cast(k8s_client.V1Pod, pod)
+    except ApiException as e:
         if e.status == 404:
             return None
         raise RuntimeError from e
@@ -581,3 +582,54 @@ def get_pod_owner_references(
     return cast(
         List[k8s_client.V1OwnerReference], pod.metadata.owner_references
     )
+
+
+def compute_step_pod_name(orchestrator_run_id: str, step_name: str) -> str:
+    """Compute the step pod name for a given orchestrator run and step.
+
+    Args:
+        orchestrator_run_id: The orchestrator run ID.
+        step_name: The name of the step.
+
+    Returns:
+        The computed step pod name.
+    """
+    # Step pods follow the pattern: {orchestrator_run_id}-{step_name}
+    # We need to sanitize the step name to ensure it's valid for Kubernetes
+    sanitized_step_name = sanitize_label(step_name)
+    return f"{orchestrator_run_id}-{sanitized_step_name}"
+
+
+def get_step_pod_statuses(
+    core_api: k8s_client.CoreV1Api,
+    namespace: str,
+    orchestrator_run_id: str,
+    step_names: List[str],
+) -> Dict[str, Optional[PodPhase]]:
+    """Get the status of multiple step pods.
+
+    Args:
+        core_api: The Kubernetes Core API client.
+        namespace: The Kubernetes namespace.
+        orchestrator_run_id: The orchestrator run ID.
+        step_names: List of step names to check.
+
+    Returns:
+        Dictionary mapping step names to their pod phases (None if pod not found).
+    """
+    step_statuses = {}
+
+    for step_name in step_names:
+        pod_name = compute_step_pod_name(orchestrator_run_id, step_name)
+        pod = get_pod(core_api, pod_name, namespace)
+
+        if pod and pod.status and pod.status.phase:
+            try:
+                step_statuses[step_name] = PodPhase(pod.status.phase)
+            except ValueError:
+                # Handle unknown pod phases
+                step_statuses[step_name] = PodPhase.UNKNOWN
+        else:
+            step_statuses[step_name] = None
+
+    return step_statuses
--- a/src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
@@ -34,7 +34,16 @@ Adjusted from https://github.com/tensorflow/tfx/blob/master/tfx/utils/kube_utils
 import enum
 import re
 import time
-from typing import Any, Callable, Dict, List, Optional, TypeVar, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    TypeVar,
+    cast,
+)
 
 from kubernetes import client as k8s_client
 from kubernetes import config as k8s_config
@@ -47,7 +56,13 @@ from zenml.integrations.kubernetes.orchestrators.manifest_utils import (
     build_service_account_manifest,
 )
 from zenml.logger import get_logger
+from zenml.orchestrators.utils import get_orchestrator_run_name
 from zenml.utils.time_utils import utc_now
+
+if TYPE_CHECKING:
+    from zenml.integrations.kubernetes.orchestrators.kubernetes_orchestrator import (
+        KubernetesOrchestratorSettings,
+    )
 
 logger = get_logger(__name__)
 
@@ -584,52 +599,33 @@ def get_pod_owner_references(
     )
 
 
-def compute_step_pod_name(orchestrator_run_id: str, step_name: str) -> str:
-    """Compute the step pod name for a given orchestrator run and step.
+def compute_step_pod_name(
+    step_name: str,
+    orchestrator_pod_name: str,
+    namespace: str,
+    settings: "KubernetesOrchestratorSettings",
+) -> str:
+    """Compute the step pod name for a given step.
 
     Args:
-        orchestrator_run_id: The orchestrator run ID.
         step_name: The name of the step.
+        orchestrator_pod_name: The name of the orchestrator pod.
+        namespace: The namespace in which the pod will be created.
+        settings: The settings for the orchestrator.
 
     Returns:
         The computed step pod name.
     """
-    # Step pods follow the pattern: {orchestrator_run_id}-{step_name}
-    # We need to sanitize the step name to ensure it's valid for Kubernetes
-    sanitized_step_name = sanitize_label(step_name)
-    return f"{orchestrator_run_id}-{sanitized_step_name}"
+    prefix = settings.pod_name_prefix
+    if prefix and not orchestrator_pod_name.startswith(prefix):
+        max_length = calculate_max_pod_name_length_for_namespace(
+            namespace=namespace
+        )
+        pod_name_prefix = get_orchestrator_run_name(
+            pipeline_name=prefix, max_length=max_length
+        )
+        pod_name = f"{pod_name_prefix}-{step_name}"
+    else:
+        pod_name = f"{orchestrator_pod_name}-{step_name}"
 
-
-def get_step_pod_statuses(
-    core_api: k8s_client.CoreV1Api,
-    namespace: str,
-    orchestrator_run_id: str,
-    step_names: List[str],
-) -> Dict[str, Optional[PodPhase]]:
-    """Get the status of multiple step pods.
-
-    Args:
-        core_api: The Kubernetes Core API client.
-        namespace: The Kubernetes namespace.
-        orchestrator_run_id: The orchestrator run ID.
-        step_names: List of step names to check.
-
-    Returns:
-        Dictionary mapping step names to their pod phases (None if pod not found).
-    """
-    step_statuses = {}
-
-    for step_name in step_names:
-        pod_name = compute_step_pod_name(orchestrator_run_id, step_name)
-        pod = get_pod(core_api, pod_name, namespace)
-
-        if pod and pod.status and pod.status.phase:
-            try:
-                step_statuses[step_name] = PodPhase(pod.status.phase)
-            except ValueError:
-                # Handle unknown pod phases
-                step_statuses[step_name] = PodPhase.UNKNOWN
-        else:
-            step_statuses[step_name] = None
-
-    return step_statuses
+    return sanitize_pod_name(pod_name=pod_name, namespace=namespace)
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
@@ -743,30 +743,23 @@ class KubernetesOrchestrator(ContainerizedOrchestrator):
             )
 
     def fetch_status(
-        self, run: "PipelineRunResponse", fetch_steps: bool = False
+        self, run: "PipelineRunResponse", include_steps: bool = False
     ) -> Tuple[ExecutionStatus, Optional[Dict[str, ExecutionStatus]]]:
         """Refreshes the status of a specific pipeline run.
 
         Args:
             run: The run that was executed by this orchestrator.
-            fetch_steps: If True, also fetch the status of individual steps.
+            include_steps: If True, also fetch the status of individual steps.
 
         Returns:
             A tuple of (pipeline_status, step_statuses_dict).
-            If fetch_steps is False, step_statuses_dict will be None.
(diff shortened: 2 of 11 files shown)


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -388,11 +389,21 @@ def _cleanup_run(self) -> None:
         """Cleans up the active run."""
         self._active_deployment = None
 
-    def fetch_status(self, run: "PipelineRunResponse") -> ExecutionStatus:
+    def fetch_status(
+        self, run: "PipelineRunResponse", include_steps: bool = False
+    ) -> Tuple[
+        Optional[ExecutionStatus], Optional[Dict[str, ExecutionStatus]]
```
schustmi, line 450: My retry changes make it so that a step can be retried multiple times as part of one run. I'm not sure how that will be mapped on the orchestrator side, but maybe this will have to be a `Dict[str, List[ExecutionStatus]]` eventually

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -199,28 +199,15 @@ def get_run(
...
-                        "with, is deleted."
-                    )
-            else:
-                raise RuntimeError(
-                    f"The stack, the run '{run.id}' was executed on, is deleted."
-                )
-
-            run = run.refresh_run_status()
+            logger.warning(
+                "DEPRECATED: Refreshing the status through the ZenML server "
```
schustmi: If you're suggesting a replacement endpoint, then I guess refreshing the status through the server is not deprecated at all, but rather on this endpoint?

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
```diff
@@ -581,3 +597,35 @@ def get_pod_owner_references(
     return cast(
         List[k8s_client.V1OwnerReference], pod.metadata.owner_references
     )
+
+
+def compute_step_pod_name(
```
schustmi: This will have to change after the step retries is merged, which switches to kubernetes jobs to run steps.

src/zenml/utils/run_utils.py
```diff
@@ -72,3 +73,73 @@ def stop_run(run: PipelineRunResponse, graceful: bool = False) -> None:
...
+    # Fetch the status (and optionally step statuses)
+    pipeline_status, step_statuses = orchestrator.fetch_status(
+        run=run, include_steps=include_step_updates
+    )
+
+    client = Client()
+
+    # Update step statuses
+    if include_step_updates:
+        assert step_statuses is not None
```
schustmi: This means if someone is refreshing with step status for any orchestrator other than kubernetes, they will get an internal server error?

Asserts are not something to use ever, unless you are 100% sure that value is not None. Which in this case I don't think is the case, rather it will be None in most cases

src/zenml/utils/run_utils.py
```diff
@@ -72,3 +73,73 @@ def stop_run(run: PipelineRunResponse, graceful: bool = False) -> None:
...
+        BaseOrchestrator,
+        StackComponent.from_model(component_model=orchestrator_list[0]),
+    )
+
+    # Fetch the status (and optionally step statuses)
+    pipeline_status, step_statuses = orchestrator.fetch_status(
+        run=run, include_steps=include_step_updates
+    )
+
+    client = Client()
```
schustmi: The `Client` class should not be used on the server. In other functions that are used both client as well as server-side, we usually have an argument called `zen_store` that can be passed, and only if that is not set we fall back to the client.

src/zenml/utils/run_utils.py
```diff
@@ -72,3 +73,73 @@ def stop_run(run: PipelineRunResponse, graceful: bool = False) -> None:
...
+
+    # Update step statuses
+    if include_step_updates:
+        assert step_statuses is not None
+
+        for step_name, step_status in step_statuses.items():
+            # If the new status is different from the current status, update it
+            if step_status != run.steps[step_name].status:
```
schustmi: If you look at run.steps, this simply fetches all the steps of the pipeline run. For a pipeline with 1000 steps, you're fetching 1000 * 1000 steps in this simple for loop.

src/zenml/utils/run_utils.py
```diff
@@ -72,3 +73,73 @@ def stop_run(run: PipelineRunResponse, graceful: bool = False) -> None:
...
+        run=run, include_steps=include_step_updates
+    )
+
+    client = Client()
+
+    # Update step statuses
+    if include_step_updates:
+        assert step_statuses is not None
+
+        for step_name, step_status in step_statuses.items():
```
schustmi: I don't think this logic works. What if the step is `Retried` in the ZenML server, but that is not something that you can detect from the orchestrator? If a step is already in a finished state, it should IMO not be updated anymore.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -771,6 +771,218 @@ def _stop_run(
...
+                "Cannot determine orchestrator run ID for the run. "
+                "Unable to fetch the status."
+            )
+
+        # Check the orchestrator pod status (only if run is not finished)
+        if not run.status.is_finished:
+            orchestrator_pod_phase = self._check_pod_status(
+                pod_name=orchestrator_run_id,
+            )
+            pipeline_status = self._map_pod_phase_to_execution_status(
```
schustmi, line 821: The orchestrator pod can finish without error (which gets mapped to a `success` state for the pipeline run here) but still have lots of failed steps, as the orchestrator pod simply catches the exceptions in those step pods. Now we would be updating a pipeline run to a state which it very much isn't in, no?

schustmi (changes requested):

src/zenml/utils/run_utils.py
bcdurak: Moving the run.steps outside of the for loop.

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: Yeah, the phrasing is off. I fixed it.

src/zenml/utils/run_utils.py
bcdurak: If someone tries to include_step_updates with any other orchestrator than the Kubernetes orchestrator, it should fail in my opinion. 

Even on Kubernetes, if there is no step to update, `step_statuses` will be an empty dict instead of None.

But I agree on the `AssertionError`.

src/zenml/utils/run_utils.py
bcdurak: > If a step is already in a finished state, it should IMO not be updated anymore.

In the Kubernetes orchestrator, we enforce this in the `_fetch_step_status`. Do you think we should also enforce this here?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: For me to update the status of the run, the run needs to be in a non-finished state. So, that means I can't update a run that has `failed` back to `running` or `success`, and the run status will be updated as soon as a step fails.

I guess the following scenario might still be possible if the events happen in the following order:

- (process 1) I click the refresh button. 
- (process 1) It fetches the run that is in a `running` state.
- (process 2) A step fails.
- (process 2) It updates itself, thus the run, to `failed`.
- (process 3) Orchestrator finishes.
- (process 1) It uses the client to take a look at the orchestrator state that is now finished.
- (process 1) Tries to update the status from `failed` to `success`

Even in this case, the `_update_pipeline_run_status` will update the run back to `failed`.

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
bcdurak: Switched this to work with K8s Jobs.

src/zenml/utils/run_utils.py
bcdurak: Done, adjusted the endpoints accordingly as well.

src/zenml/orchestrators/base_orchestrator.py
bcdurak: I am resolving this as this is deemed out of the scope of this PR.

src/zenml/utils/run_utils.py
bcdurak: Enforced here now.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -234,6 +234,10 @@ def run_step_on_kubernetes(step_name: str) -> None:
                 pod_name, namespace=namespace
             )
 
+            # Add step name to labels so both pod and job have consistent labeling
```
schustmi, line 237: Maybe we can call them `base_labels` and `step_labels` instead? Right now the names are a little misleading IMO, as `step_labels` is used for both the pod and the job, and `step_pod_labels` is only used as the generic base for all steps

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
```diff
@@ -587,6 +603,38 @@ def get_pod_owner_references(
     )
 
 
+def compute_step_pod_name(
```
schustmi: I think this is not needed anymore?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -785,6 +785,253 @@ def _stop_run(
...
+            # List events related to the pod
+            events = self._k8s_core_api.list_namespaced_event(
+                namespace=self.config.kubernetes_namespace,
+                field_selector=f"involvedObject.name={pod_name}",
+                limit=15,
+            )
+
+            # Look for failure/eviction/deletion related events
+            failure_reasons = [
+                "Evicted",
```
schustmi: How did you get this list? Is this a complete list of failure events? And isn't there a better way to get only failure events? `type!=Normal` for example?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -785,6 +785,253 @@ def _stop_run(
...
+            try:
+                return kube_utils.PodPhase(pod.status.phase)
+            except ValueError:
+                # Handle unknown pod phases
+                logger.warning(
+                    f"Unknown pod phase for pod {pod_name}: {pod.status.phase}"
+                )
+                return kube_utils.PodPhase.UNKNOWN
+        else:
+            # Pod is missing - check events to determine if this indicates failure or unknown status
```
schustmi: The pod is not missing here, it might also just not have a status.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -785,6 +785,253 @@ def _stop_run(
...
+                    return ExecutionStatus.FAILED
+
+        # Check job status fields
+        if job.status:
+            if job.status.active and job.status.active > 0:
+                return ExecutionStatus.RUNNING
+            elif job.status.succeeded and job.status.succeeded > 0:
+                return ExecutionStatus.COMPLETED
+            elif job.status.failed and job.status.failed > 0:
+                return ExecutionStatus.FAILED
```
schustmi: This is wrong, as the job will retried failures. I think in general, just checking the conditions of the job should be fine, no need to also check the status. The job might also be running even if active is `0`, if it's in the process of spinning up a new pod.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -785,6 +785,253 @@ def _stop_run(
...
+
+        # Return None if no clear status - don't update
+        return None
+
+    def _fetch_step_status(
```
schustmi: statuses?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -785,6 +785,253 @@ def _stop_run(
...
+        self, run: "PipelineRunResponse", orchestrator_run_id: str
```
schustmi: I don't think this needs the orchestrator run id?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -785,6 +785,253 @@ def _stop_run(
...
+            # Extract step name from job labels
+            if not job.metadata or not job.metadata.labels:
+                continue
+
+            step_name = job.metadata.labels.get("step_name")
+            if not step_name:
+                continue
+
+            # Check if this step is already finished
+            step_response = run.steps.get(step_name)
```
schustmi: This is again fetching all the steps every time

src/zenml/utils/run_utils.py
```diff
@@ -72,3 +75,84 @@ def stop_run(run: PipelineRunResponse, graceful: bool = False) -> None:
...
+        zen_store = Client().zen_store
+    else:
+        zen_store = zen_store
+
+    # Update step statuses
+    if include_step_updates:
+        if step_statuses:
+            current_steps = run.steps
+            for step_name, step_status in step_statuses.items():
+                if current_steps[step_name].status.is_finished:
```
schustmi, line 145: `current_steps` might not always contain that step. For example: The job already exists, which means your logic in the KubernetesOrchestrator returns something for that step. But the job has not yet reached the point in code where it registered the step in the server yet. So in that case, we do not want to update anything as that will fail.

schustmi (changes requested):

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: Makes sense, switching them to `base_labels` and `step_labels`.

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
bcdurak: Removed.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Removed the comment and adjusted the message.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Renamed

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Moved it outside.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Correct, removed unused input.

src/zenml/utils/run_utils.py
bcdurak: Added an additional check. If the `step_name` is not in the run.steps yet, we skip.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Removed status-based checks as discussed.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
bcdurak: Removed the event-based updates as discussed.

src/zenml/utils/run_utils.py
schustmi: I think this check should be here in the generic code, not in the Kubernetes implementation. It simply means, anyone implementing this for an orchestrator has more ways to fuck up.

bcdurak: Good point, added it.

PR #3737, opened 2025-06-13
Add VSCode tutorial pipeline testing to CI

strickvl:
Adds a new CI workflow that tests all tutorial pipelines from the `zenml-io/vscode-tutorial-extension` repository to ensure ZenML core changes don't break user-facing tutorial examples.

This provides early detection of breaking changes that would affect the tutorial pipelines that users follow when learning ZenML.

View a successful run of these tests here: https://github.com/zenml-io/zenml/actions/runs/15637115894/job/44055030989. (This took just over 2 mins to run end-to-end).

## Key Features
- **Standalone workflow**: Can be called from ci-slow or run independently via GitHub Actions UI
- **Regression testing**: Tests ZenML from current branch (not PyPI) for accurate testing
- **Comprehensive coverage**: Runs all 10 tutorial pipelines with proper error handling
- **Resilient**: Includes retry logic for network issues and timeout protection per pipeline
- **Well documented**: Updates both tests/README.md and CLAUDE.md with testing strategy

## Implementation Details
- **New workflow**: `.github/workflows/vscode-tutorial-pipelines-test.yml`
- **Test script**: `scripts/test-tutorial-pipelines.sh` with validation and reporting
- **CI integration**: Added to `ci-slow.yml` as `vscode-tutorial-pipelines-test` job
- **Environment**: Python 3.12 with UV package manager for fast dependency resolution
- **Error handling**: Comprehensive validation of repo structure and meaningful failure reporting

## Files Changed
- `.github/workflows/vscode-tutorial-pipelines-test.yml` (new standalone workflow)
- `scripts/test-tutorial-pipelines.sh` (new test execution script)
- `.github/workflows/ci-slow.yml` (added job call)
- `tests/README.md` (updated testing strategy documentation)
- `CLAUDE.md` (updated CI documentation)

## Test Plan
- [x] YAML syntax validation passes
- [x] Formatting with `scripts/format.sh` passes
- [x] Workflow includes proper error handling and resilience features
- [x] Script validates tutorial repository structure
- [x] Documentation accurately reflects new testing approach

The workflow will be triggered when the `run-slow-ci` label is applied to PRs, ensuring tutorial pipeline compatibility is verified before merging changes that could break user examples.


.github/workflows/vscode-tutorial-pipelines-test.yml
```diff
@@ -0,0 +1,96 @@
...
+              break
+            elif [ $i -lt 3 ]; then
+              echo "Clone attempt $i failed, retrying in 5 seconds..."
+              sleep 5
+            else
+              echo "Failed to clone tutorial repository after 3 attempts"
+              exit 1
+            fi
+          done
+      - name: Create virtual environment
```
bcdurak, line 74: Does this mean that we are using a singular venv for all the pipelines? In that case, we would also have to merge all the requirements which might become an issue down the road.

bcdurak: Also, if any of the pipelines depend on previous runs, artifacts, etc., running everything on the same server might create unforeseen circumstances.

strickvl: They're all separate pipelines. I mean, all these tutorial pipelines get run with the same env, one after another in the Codespace, so it shouldn't be a problem here.

strickvl:
> I have a general question about this. How do we version the changes in the tutorials? If something breaks in ZenML, the tutorial will break. Let's say we fix it in the other repo, which will run it out-of-sync with the latest ZenML release for a while until we make the next release. Even when we release, the pipelines will be incompatible with earlier versions of ZenML. Do we have any precautions for it?

The main thing this is trying to ward against is someone making a breaking change in the ZenML core API and not realising that it's also broken the tutorial pipelines.

The way we version changes to the tutorial pipelines is that it has it's own repo and people can make changes / PRs there etc and fix / update to work with the latest versions of ZenML.

I don't think there's a problem with the pipelines being incompatible with earlier versions of ZenML. These tutorial pipelines will only ever be run with the latest release of ZenML. And any changes made to the tutorials repo won't be reflected in the Codespaces environment until someone rebuilds the extension + pushes a new Docker image. 

But maybe @safoinme has some thoughts on this as well.

scripts/test-tutorial-pipelines.sh
```diff
@@ -0,0 +1,102 @@
...
+    echo "⚠️  WARNING: utils.py not found in tutorial repository root"
+    echo "   Pipelines may fail to import utils module"
+fi
+
+echo "📋 Discovering pipeline files..."
+
+# Define all pipeline files to test
+PIPELINE_FILES=(
+    "pipelines/caching/cache_pipeline.py"
+    "pipelines/fanOut/fan_pipeline.py"
```
safoinme: Can we make this somehow dynamic? so if new pipelines were added

scripts/test-tutorial-pipelines.sh
```diff
@@ -0,0 +1,102 @@
...
+FAILED_PIPELINES=()
+
+# Run each pipeline
+for pipeline in "${PIPELINE_FILES[@]}"; do
+    if [ -f "$pipeline" ]; then
+        echo ""
+        echo "🔍 Testing pipeline: $pipeline"
+        echo "----------------------------------------"
+        
+        if timeout 300 env PYTHONPATH="$(pwd):${PYTHONPATH:-}" python "$pipeline"; then
```
safoinme: can we create readonly variables on top so changes can be easier to track 

readonly PIPELINE_TIMEOUT=300

safoinme: can we create readonly variables on top so changes can be easier to track 

readonly PIPELINE_TIMEOUT=300

safoinme (approved):

scripts/test-tutorial-pipelines.sh
strickvl: Done

scripts/test-tutorial-pipelines.sh
strickvl: Done

bcdurak:
> > I have a general question about this. How do we version the changes in the tutorials? If something breaks in ZenML, the tutorial will break. Let's say we fix it in the other repo, which will run it out-of-sync with the latest ZenML release for a while until we make the next release. Even when we release, the pipelines will be incompatible with earlier versions of ZenML. Do we have any precautions for it?
> 
> The main thing this is trying to ward against is someone making a breaking change in the ZenML core API and not realising that it's also broken the tutorial pipelines.
> 
> The way we version changes to the tutorial pipelines is that it has it's own repo and people can make changes / PRs there etc and fix / update to work with the latest versions of ZenML.
> 
> I don't think there's a problem with the pipelines being incompatible with earlier versions of ZenML. These tutorial pipelines will only ever be run with the latest release of ZenML. And any changes made to the tutorials repo won't be reflected in the Codespaces environment until someone rebuilds the extension + pushes a new Docker image.
> 
> But maybe @safoinme has some thoughts on this as well.

I see your point. However, it will be impossible to ward against someone breaking ZenML. Those types of changes are inevitable. 😄 

I guess if the compatibility with the earlier versions is not an issue, and more importantly, any changes made to the tutorials repo isn't reflected until someone rebuilds, this approach should be fine.

.github/workflows/vscode-tutorial-pipelines-test.yml
bcdurak: Not really sure how we plan to move forward with adding more pipelines here. But just as a warning, if two pipelines have conflicting dependencies in the future, this will become a problem.

For now, we can keep it like this, though.

PR #3758, opened 2025-06-20
Threadsafe RestZenStore

schustmi:
## Describe changes
This PR adds a reentrant lock around the `session` and (re-)authentication code of the `RestZenStore` to make it useable from multiple threads.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/rest_zen_store.py
+++ b/src/zenml/zen_stores/rest_zen_store.py
@@ -274,6 +274,7 @@ from zenml.utils.networking_utils import (
     replace_localhost_with_internal_hostname,
 )
 from zenml.utils.pydantic_utils import before_validator_handler
+from zenml.utils.time_utils import utc_now
 from zenml.zen_server.exceptions import exception_from_response
 from zenml.zen_stores.base_zen_store import BaseZenStore
 
@@ -443,6 +444,7 @@ class RestZenStore(BaseZenStore):
     _session: Optional[requests.Session] = None
     _server_info: Optional[ServerModel] = None
     _session_lock: RLock = PrivateAttr(default_factory=RLock)
+    _last_authenticated: Optional[datetime] = None
 
     # ====================================
     # ZenML Store interface implementation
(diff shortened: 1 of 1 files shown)


src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -4442,62 +4455,68 @@ def _request(
...
-                    )
-                    raise CredentialsNotValid(
-                        "The current credentials are no longer valid. Please "
-                        "log in again using 'zenml login'."
-                    ) from e
+                with self._session_lock:
+                    if self._last_authenticated != last_authenticated:
+                        # Another thread has re-authenticated since the last
+                        # request. We simply retry the request.
+                        continue
```
schustmi, line 4462: @stefannica I switched this to a timestamp as discussed. I have one very theoretical case where this solution does not work as intended, but IMO this can be ignored:

Let's imagine the following scenario: 2 threads, the store currently has `self._api_token = None` and the `credentials_store` contains an invalid (not expired!) token.

- Both threads send a request at the same time
- Both threads get a `CredentialsNotValid`
- `thread_a` enters the critical section. It enters the `if self._api_token is None` condition, and authenticates. This simply fetches the outdated token from the credentials store and sets it on the store.
- `thread_b` enters the critical section. It detects that another thread ran the auth code already.
- Both threads retry the request
- Both threads again get a `CredentialsNotValid`
- `thread_a` enters the critical section. It enters the `elif not re_authenticated:` condition and authenticates. This uses some credentials to fetch a new token. Let's assume (and this is very theoretical) that this token is for some reason invalid.
- `thread_b` enters the critical section. It detects that another thread ran the auth code already.
- Both threads retry the request
- Both threads again get a `CredentialsNotValid`
- `thread_a` enters the critical section. It has `re_authenticated=True`, which means it raises an exception.
- `thread_b` enters the critical section. It enters the `elif not re_authenticated:` condition and authenticates. **Ideally, this would immediately also raise an exception and not try to authenticate again, as the same code already failed to generate valid credentials when `thread_` tried it.**

If more threads were involved, all of them would run into the last bullet point eventually, which means that each thread separately tries to authenticate at some point.

One potential solution for this would be to clear the credentials that were used to generate the token (e.g. the pro credentials/api key/username+pw), but not sure if that makes sense?

stefannica: I think the current solution is enough for the most cases.

stefannica (approved):

PR #3761, opened 2025-06-23
Helm chart: Add option to create PVC for local configuration path

jsuchome:
This enables persistence for deployment with local sqlite database.

## Describe changes

Added `.database.persistence` section to the helm chart. If enabled, PVC will be created for mounting `zenml-config` so that when the pod restarts, data stored in local db (users, stack components etc.) are not lost

The option is disabled by default to keep the default behavior unchanged.

## Pre-requisites

## Types of changes

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -268,10 +268,17 @@ zenml:
   enableImplicitAuthMethods: false
 
   # MySQL database configuration. If not set, a local sqlite database will be
-  # used, which will not be persisted across pod restarts.
+  # used. To ensure data are not lost, check the persistence part.
   # NOTE: the certificate files need to be copied in the helm chart folder and
   # the paths configured here need to be relative to the root of the helm chart.
   database:
+
+    # If set to true, path where the local database is created will be
+    # mounted as a persistent volume so the data is not lost on pod restarts.
+    persistence:
+      enabled: false
+      size: 1Gi
+      storageClassName: ""
 
     # The database URL. If not set, a local sqlite database will be used, which
     # will not be persisted across pod restarts. The URL can contain the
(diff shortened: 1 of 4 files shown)


helm/values.yaml
```diff
@@ -306,11 +306,18 @@ zenml:
...
 
   # MySQL database configuration. If not set, a local sqlite database will be
-  # used, which will not be persisted across pod restarts.
+  # used. To ensure data are not lost, check the persistence part.
   # NOTE: the certificate files need to be copied in the helm chart folder and
   # the paths configured here need to be relative to the root of the helm chart.
   database:
 
+    # If set to true, path where the local database is created will be
+    # mounted as a persistent volume so the data is not lost on pod restarts.
```
stefannica, line 315: Could you please add a warning here about running with more than 2 replicas, saying that you can’t share a volume across pods on different nodes unless it supports ReadWriteMany access mode (which most don't)?

stefannica (approved):
Great addition, thanks for the contribution !

helm/values.yaml
jsuchome: What about forcing it? Either way:

1. replicaCount > 1 -> do not allow pvc creation
2. persistence enabled -> force 1 replica

Which reminds me, you are already creating pvc for some db-backup, right? There does not seem to be any check for number of replicas in this case

jsuchome: Oh, now I see, it's actually always `replicas: 1`  when database url is not provided, persistence or not

jsuchome: https://github.com/zenml-io/zenml/blob/main/helm/templates/server-deployment.yaml#L9-L15

jsuchome: So I did something else: make sure that `persistence` is only used for local db setup. Then the link above makes sure that this also means 1 replica. What do you think?

stefannica: Great idea, thanks !

stefannica (approved):

PR #3762, opened 2025-06-23
Server-side parent step computation

schustmi:
## Describe changes
This PR adds logic to compute the parent steps server-side. This means the client does not need to fetch all upstream steps anymore to create a `StepRunRequest`, which should improve performance especially when having large `fan-in` steps.
## Pre-requisites

## Types of changes


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8779,19 +8779,18 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
...
 
             step_model = step_schema.to_model(include_metadata=True)
 
+            for upstream_step in step_model.spec.upstream_steps:
+                self._set_run_step_parent_step(
+                    child_step_run=step_schema,
+                    parent_step_name=upstream_step,
+                    session=session,
+                )
+
```
stefannica, line 8793: This is the only change I didn't understand. Did you move this after `to_model` to keep the step model clean of upstream steps?

stefannica (approved):

src/zenml/zen_stores/sql_zen_store.py
schustmi: I only did that for an easy way to get the `upstream_steps`, which are already parsed as part of the `spec` of the model, but would be a raw json string when fetching from the schema. This step_model here anyway is not the final one, but only an intermediate one used for the same purpose already (easy access to parsed json fields)

PR #3770, opened 2025-06-24
Hide service connector secrets as internal implementation details

### LINKED ISSUE #3747: Service Connector credentials being exposed to user without proper permissions

_No response_

### System Information

All zenml versions <= 0.83.0

### What happened?

I have a user workspace role with the following permissions assigned:
Service Connector: Read/Client
Secrets: read/read secret value
I have found out that the Service Connector is creating a "service" secret for its' needs. Via this secret user can still view the credentials for the service connector, even if on the Connectors page they are not viewed.

Since this goes in disagreement with the configured permissions, I tend to label this as a bug.

Suggestions for a fix: limit the view of the "service" secrets for the users without Read secret value on Service Connector object.

### Reproduction steps

1.
2.
3.
...

### Relevant log output

```shell

```

### Code of Conduct

stefannica:
## Describe changes
Fixes #3747.

ZenML Service Connectors rely on ZenML secrets to store the sensitive attributes in their configuration. These secrets are managed internally by ZenML (created/updated/deleted) and are used solely for the purpose of storing Service Connector credentials, but they are still visible to the users as regular secrets and can be managed as such.

This PR reworks these secrets as Service Connector implementation details that are completely hidden from the users:

* no longer visible as regular secrets (i.e. accessing them directly via the API no longer works)
* any functionality that previously needed to access these secrets to fetch the credentials from the client side is replaced by the `expand_secrets` flag in the `get_service_connector` and `list_service_connector` Client, REST API endpoints and Zen store functions.

## Implementation Details

* a new `internal` attribute is added to the secrets SQL table that is set only for service connector secrets (a DB migration script is included)
* the regular secrets API operations (get/list/update/delete) can no longer access internal secrets
* the service connector configuration uses a new type for its configuration that extends from `Dict[str, Any]` and provides functionality required to mix SecretStr and regular values together as dictionary values. This replaces the `secret` and `secret_id` attributes that were previously present in service connector models and used to access the secrets directly from the client.

## Pre-requisites

## Types of changes


src/zenml/models/v2/core/service_connector.py
```diff
@@ -38,6 +44,107 @@
 
 logger = get_logger(__name__)
 
+# ------------------ Configuration Model ------------------
+
+
+class ServiceConnectorConfiguration(Dict[str, Any]):
```
schustmi, line 50: When this is serialized/deserialized, does it preserve the `SecretStr` type in the dictionary?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6743,6 +6743,38 @@ def _delete_backup_secret_values(
...
+
+        try:
+            # Set the secret values in the configured secrets store
+            self._set_secret_values(
+                secret_id=new_secret.id, values=secret.secret_values
+            )
+        except:
+            # If setting the secret values fails, delete the secret from the
+            # database.
+            with Session(self.engine) as session:
```
schustmi: can't we reuse the passed `session` for this?

src/zenml/models/v2/core/service_connector.py
schustmi: It must be right, otherwise I don't really understand how for example the client can send a config dictionary to the server, and the server then understands which one of those to store as a secret?

schustmi: Ah nvm, this seems to be handled in `_validate_and_configure_resources`, which we then need to make sure is called everytime we receive a service connector requests/response somewhere?

stefannica: yes, exactly, `_validate_and_configure_resources` is called in a bunch of places exactly for this reason.

src/zenml/zen_stores/sql_zen_store.py
stefannica: done

src/zenml/models/v2/core/service_connector.py
stefannica: in fact, I added more `_validate_and_configure_resources` calls in the REST zen store to make sure the models are properly validated after they come through the wire.

schustmi (approved):

PR #3771, opened 2025-06-24
Use container registry for authenticating kubernetes image pull

AlexejPenner:
## Describe changes
Claude and I implemented the missing feature that allows ZenML to make appropriate credentials available to the kubernetes pods for image pulling.

I have validated that this works for my specific usecase - I'm happy to iterate on this and to discuss decisions.

I am also not sure how this will affect legacy customers that might be setting image pull secrets explicitly

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/container_registries/base_container_registry.py
+++ b/src/zenml/container_registries/base_container_registry.py
@@ -273,9 +273,8 @@ class BaseContainerRegistry(AuthenticationMixin):
         if not registry_uri.startswith(("http://", "https://")):
             domain = registry_uri.split("/")[0]
             return f"https://{domain}"
-        
-        return registry_uri
 
+        return registry_uri
 
 
 class BaseContainerRegistryFlavor(Flavor):
--- a/src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
@@ -386,6 +386,7 @@ def create_secret(
     namespace: str,
     secret_name: str,
     data: Dict[str, Optional[str]],
+    secret_type: str = "Opaque",
 ) -> None:
     """Create a Kubernetes secret.
 
@@ -394,10 +395,13 @@ def create_secret(
         namespace: The namespace in which to create the secret.
         secret_name: The name of the secret to create.
         data: The secret data.
+        secret_type: The secret type.
     """
     core_api.create_namespaced_secret(
         namespace=namespace,
-        body=build_secret_manifest(name=secret_name, data=data),
+        body=build_secret_manifest(
+            name=secret_name, data=data, secret_type=secret_type
+        ),
     )
 
 
@@ -416,10 +420,20 @@ def update_secret(
         data: The secret data. If the value is None, the key will be removed
             from the secret.
     """
+    # For updates, we only patch the data field, not the type (which is immutable)
+    import base64
+
+    encoded_data = {
+        key: base64.b64encode(value.encode()).decode() if value else None
+        for key, value in data.items()
+    }
+
+    patch_body = {"data": encoded_data}
+
     core_api.patch_namespaced_secret(
         namespace=namespace,
         name=secret_name,
-        body=build_secret_manifest(name=secret_name, data=data),
+        body=patch_body,
     )
 
 
@@ -428,6 +442,7 @@ def create_or_update_secret(
     namespace: str,
     secret_name: str,
     data: Dict[str, Optional[str]],
+    secret_type: str = "Opaque",
 ) -> None:
     """Create a Kubernetes secret if it doesn't exist, or update it if it does.
 
@@ -437,17 +452,35 @@ def create_or_update_secret(
         secret_name: The name of the secret to create or update.
         data: The secret data. If the value is None, the key will be removed
             from the secret.
+        secret_type: The secret type.
 
     Raises:
         ApiException: If the secret creation failed for any reason other than
             the secret already existing.
     """
     try:
-        create_secret(core_api, namespace, secret_name, data)
+        create_secret(core_api, namespace, secret_name, data, secret_type)
     except ApiException as e:
         if e.status != 409:
             raise
-        update_secret(core_api, namespace, secret_name, data)
+
+        # Check if the existing secret has a different type
+        try:
+            existing_secret = core_api.read_namespaced_secret(
+                name=secret_name, namespace=namespace
+            )
+            if existing_secret.type != secret_type:
+                # Delete the secret if the type is different (type is immutable)
+                delete_secret(core_api, namespace, secret_name)
+                create_secret(
+                    core_api, namespace, secret_name, data, secret_type
+                )
+            else:
+                # Same type, just update the data
+                update_secret(core_api, namespace, secret_name, data)
+        except ApiException:
+            # If we can't read the existing secret, try to update anyway
+            update_secret(core_api, namespace, secret_name, data)
 
 
 def create_or_update_secret_from_manifest(
@@ -466,30 +499,37 @@ def create_or_update_secret_from_manifest(
     """
     namespace = secret_manifest["metadata"]["namespace"]
     secret_name = secret_manifest["metadata"]["name"]
-    
+
     # Extract data from manifest - handle both 'data' and 'stringData' fields
     secret_data = {}
-    
+
     # Handle base64-encoded 'data' field
     if "data" in secret_manifest:
         import base64
+
         for key, encoded_value in secret_manifest["data"].items():
             if encoded_value is not None:
                 # Decode base64 data back to string
-                secret_data[key] = base64.b64decode(encoded_value).decode('utf-8')
+                secret_data[key] = base64.b64decode(encoded_value).decode(
+                    "utf-8"
+                )
             else:
                 secret_data[key] = None
-    
+
     # Handle plain text 'stringData' field
     if "stringData" in secret_manifest:
         secret_data.update(secret_manifest["stringData"])
-    
+
+    # Extract secret type from manifest, default to "Opaque" if not specified
+    secret_type = secret_manifest.get("type", "Opaque")
+
     # Use the existing create_or_update_secret function
     create_or_update_secret(
         core_api=core_api,
         namespace=namespace,
         secret_name=secret_name,
         data=secret_data,
+        secret_type=secret_type,
     )
 
 
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
@@ -239,6 +239,15 @@ def main() -> None:
 
         logger.info(f"Container registry: {active_stack.container_registry}")
 
+        # Get registry credentials for imagePullSecrets
+        registry_credentials = None
+        if pipeline_settings.auto_generate_image_pull_secrets:
+            registry_credentials = (
+                orchestrator.get_kubernetes_image_pull_secret_data(
+                    active_stack.container_registry
+                )
+            )
+
         # Define Kubernetes pod manifest and any required secrets.
         pod_manifest, secret_manifests = build_pod_manifest(
             pod_name=pod_name,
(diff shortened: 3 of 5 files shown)


src/zenml/integrations/kubernetes/service_connectors/kubernetes_service_connector.py
```diff
@@ -274,9 +274,9 @@ def _connect_to_resource(
             assert isinstance(cfg, KubernetesTokenConfig)
 
             if cfg.token:
-                k8s_conf.api_key["authorization"] = (
-                    cfg.token.get_secret_value()
-                )
+                k8s_conf.api_key[
+                    "authorization"
+                ] = cfg.token.get_secret_value()
```
stefannica: unrelated changes like this one tells me that either there's a new formatter version released, or that you're using a really old one. You should delete and re-create your virtual environment, just to be safe.

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -33,6 +34,156 @@
...
+
+        # Normalize registry URI for consistency
+        if registry_uri.startswith("https://"):
+            registry_uri = registry_uri[8:]
+        elif registry_uri.startswith("http://"):
+            registry_uri = registry_uri[7:]
+
+        credentials.append((registry_uri, username, password))
+
+    return credentials
```
stefannica: this feels like something that should be part of the base container registry class, not a kubernetes utility module

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -33,6 +34,156 @@
...
+    if registry_uri in ("docker.io", "index.docker.io"):
+        registry_server = "https://index.docker.io/v1/"
+    else:
+        # Extract just the domain part for better image matching
+        if registry_uri.startswith(("http://", "https://")):
+            registry_server = registry_uri
+        else:
+            # For URIs like 'registry.onstackit.cloud/zenml', use just the domain
+            domain = registry_uri.split("/")[0]
+            registry_server = f"https://{domain}"
```
stefannica: Same here: this should be part of the base container registry class, maybe merge it into `_get_container_registry_credentials`.

The rule of thumb is basically this:

* if the code changes from one container registry flavor to another, it belongs in the container registry base class, or better yet, the base class + implementations, if it's something that a subclass might override
* if the code is specific to kubernetes, it belongs here

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -33,6 +34,156 @@
...
+    # Get container registry from stack
+    container_registry = stack.container_registry
+    if not container_registry:
+        return credentials
+
+    # Try to get credentials from the container registry
+    registry_credentials = container_registry.credentials
+    if registry_credentials:
+        username, password = registry_credentials
+
```
stefannica: I think you should check the connector first, as it always takes precedence over credential attributes. E.g. someone might configure credential attrs and then later on link the container registry stack component to a service connector and forget about deleting the attrs. This is common practice in all stack components.

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -33,6 +34,156 @@
...
+    Args:
+        namespace: The Kubernetes namespace to create secrets in.
+
+    Returns:
+        Tuple of (secret_manifests, local_object_references) where:
+        - secret_manifests: List of Kubernetes secret manifests to create
+        - local_object_references: List of V1LocalObjectReference objects for imagePullSecrets
+    """
+    credentials = _get_container_registry_credentials()
```
stefannica: There is a cleaner way to do this from a design perspective. Let me explain the thought process:

* first, you have to ask yourself "what is the main area of concern for this utility python module ?". The answer is "building k8s manifests and nothing else". This should make it oblivious to things like stacks and ZenML Client(s). Basically, if your design is a tree of components linked by dependencies, this module should be a low-level leaf: it should know nothing about anything else in your design and therefore has no external dependencies (manifested here as imports). Or if you use the layered design / onion analogy, this is an outermost layer, with no dependencies. 
* next, you realize this new utility function needs container registry credentials. So how do you get them here ?
    * option 1. (currently implemented): call out to the container registry and fetch the credentials. But this makes the module dependent on inner details like the Client or active stack and that spaghettifies the design and its dependency tree and we want to avoid that.
    * option 2. (recommended): use a "dependency injection" - have the other inner components that use this utility module do the grunt work of extracting the container credentials and injecting them into this function.

stefannica (changes requested):
Loving this, because it (almost) hits all the right notes:

* docker credentials might be temporary, especially if they are coming from service connectors. Your solution always fetches the latest credentials and patches the existing secret instead of configuring the secret once and then forgetting about it. This is one of the challenges I feared we'd hit with this feature and you nailed it.
* you might have different credentials for different pipelines, so you're not attaching the credentials to the service account, you're attaching them directly to the pods. This offers a nice isolation where secrets are only used by the pods that need them.
* the secrets are only needed when the pod starts, then you can forget about them. This is what's still missing: these secrets will pile up (each of them uses the pipeline run ID/name, right) until you have thousands of secrets. You need to clean up the secrets that are no longer in use - e.g. the k8s orchestrator pod could clean up after itself.

Aside from this, there are some design concerns that Claude didn't get right.

stefannica:
> I am also not sure how this will affect legacy customers that might be setting image pull secrets explicitly

You need to do the research to figure this out.

docs/book/component-guide/service-connectors/connector-types/docker-service-connector.md
```diff
@@ -91,9 +91,5 @@ The 'dockerhub' Docker Service Connector connector was used to successfully conf
 
 The Docker Service Connector can be used by all Container Registry stack component flavors to authenticate to a remote Docker/OCI container registry. This allows container images to be built and published to private container registries without the need to configure explicit Docker credentials in the target environment or the Stack Component.
 
-{% hint style="warning" %}
```
schustmi, line 94: This warning is still valid for every orchestrator apart from the Kubernetes one for which you now implemented it. So either this gets implemented for all orchestrators as part of this PR, or this warning needs to stay.

src/zenml/integrations/kubernetes/orchestrators/kube_utils.py
```diff
@@ -450,6 +450,39 @@ def create_or_update_secret(
         update_secret(core_api, namespace, secret_name, data)
 
 
+def create_or_update_secret_from_manifest(
```
schustmi, line 486: There already is a `create_secret` function in this file, can you either use that or adjust it so we have less duplicated code?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -207,8 +207,39 @@ def run_step_on_kubernetes(step_name: str) -> None:
...
+                        )
+                        logger.debug(f"Created imagePullSecret {secret_name}")
+                    except Exception as create_e:
+                        logger.warning(
+                            f"Failed to create imagePullSecret {secret_name}: {create_e}"
+                        )
+                else:
+                    logger.warning(
+                        f"Failed to check for existing secret {secret_name}: {e}"
+                    )
```
schustmi: If I understand it correctly, this secret never gets cleaned up? We're already using a kubernetes secret for the ZenML token which already has some cleanup logic, it probably makes sense to also apply this here?

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
```diff
@@ -108,7 +259,9 @@ def build_pod_manifest(
     env: Optional[Dict[str, str]] = None,
     mount_local_stores: bool = False,
     owner_references: Optional[List[k8s_client.V1OwnerReference]] = None,
-) -> k8s_client.V1Pod:
+    auto_generate_image_pull_secrets: bool = True,
```
schustmi, line 382: I don't think this is the right place to put this. The purpose of this function is to create a configurable pod manifest. In my opinion, you should be able to pass previously generated image pull secrets into this function, they should not be created as part of it.

schustmi (changes requested):
On a more general note, your change means that the kubernetes orchestrator and step operator are now broken for all users where the service account does not have permissions to create secrets. This feature should probably be configurable, whether its enabled by default is debatable I think.

schustmi:
Adding on top of what @stefannica already mentioned: Those Docker credentials are short-lived and expire eventually. Currently, you're creating them client-side, which means they should be valid at least until the orchestrator pod spins up. However, let's say I have a sequential pipeline with three steps, each of them taking 20 hours to complete. When we get to running the third step, those credentials might already be expired. I think the solution for that would be to create a new (or refresh the existing) secret right before each step pod gets created.

src/zenml/integrations/kubernetes/service_connectors/kubernetes_service_connector.py
AlexejPenner: i was using an old formatter - it changed too many files - I missed this one when reverting those changes

stefannica:
> On a more general note, your change means that the kubernetes orchestrator and step operator are now broken for all users where the service account does not have permissions to create secrets. This feature should probably be configurable, whether its enabled by default is debatable I think.

@schustmi don't we already need permission to create secrets to store the ZenML API token as a secret ?

schustmi:
> > On a more general note, your change means that the kubernetes orchestrator and step operator are now broken for all users where the service account does not have permissions to create secrets. This feature should probably be configurable, whether its enabled by default is debatable I think.
> 
> @schustmi don't we already need permission to create secrets to store the ZenML API token as a secret ?

For the same reason as I mentioned here, I made this an optional feature and even disabled it by default. So for most people (which haven't explicitly enabled this), their service accounts for now do not require those permissions.

src/zenml/container_registries/base_container_registry.py
```diff
@@ -116,6 +113,13 @@ def credentials(self) -> Optional[Tuple[str, str]]:
                     connector.config.password.get_secret_value(),
                 )
 
+        # Fall back to direct authentication secrets
+        secret = self.get_typed_authentication_secret(
+            expected_schema_type=BasicAuthSecretSchema
+        )
+        if secret:
+            return secret.username, secret.password
+
```
AlexejPenner, line 122: @stefannica according to your comment the connector should be checked first, but in the credentials implementation it was the other way around - let me know if I am missing something or if it made sense to flip here

src/zenml/integrations/kubernetes/orchestrators/manifest_utils.py
AlexejPenner: agreed 100%

AlexejPenner: i found the local. path function as well there that violates these design principles so I will be fixing that too

src/zenml/container_registries/dockerhub_container_registry.py
```diff
@@ -13,16 +13,34 @@
...
+        DockerHub requires authentication against the specific
+        'https://index.docker.io/v1/' endpoint regardless of how
+        the registry URI is configured.
+
+        Returns:
+            The DockerHub registry server URI.
+        """
+        return "https://index.docker.io/v1/"
+
+
```
AlexejPenner, line 43: I'm a bit unsure on this one as container registries seem like the ugly duckling among our stack components with very little actual implementation of the base classes.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -339,6 +356,16 @@ def finalize_run(node_states: Dict[str, NodeStatus]) -> None:
             except k8s_client.rest.ApiException as e:
                 logger.error(f"Error cleaning up secret {secret_name}: {e}")
 
+        # Clean up old imagePullSecrets to prevent accumulation
+        # Only clean up for non-scheduled runs to avoid interfering with running schedules
```
AlexejPenner, line 391: @schustmi you mentioned that htese are temporary - what would I need to do for scheduled pipelines?

AlexejPenner:
I tested it on my use-case and things worked - once we both have some time I would take you up on your offer to test together in more detail @stefannica

htahir1:
Closing due to staleness - feel free to open again whenever you are ready :-)

AlexejPenner:
@stefannica let me know when you have some time to test this together with me

PR #3774, opened 2025-06-25
Easier option to enable a step operator/experiment tracker for a step

schustmi:
## Describe changes
This PR makes it easier to specify a step operator/experiment tracker for a step. Instead of requiring the name of the step operator/experiment tracker in the configuration (which is already implemented that way to facilitate multiple components of a type in a stack), this PR allows users to simply pass `step_operator=True` or `experiment_tracker=True`. The step operator/experiment tracker of the active stack will then be selected automatically.

## Pre-requisites

## Types of changes


schustmi:
> Great one but requires docs changes :-)! Maybe around here https://docs.zenml.io/concepts/steps_and_pipelines/configuration#direct-component-assignment

Yep didn't have time yesterday, I updated your page and also all the other ones I could find to use this simpler method.

src/zenml/config/step_configurations.py
```diff
@@ -145,8 +145,8 @@ class StepConfigurationUpdate(StrictBaseModel):
     enable_artifact_metadata: Optional[bool] = None
     enable_artifact_visualization: Optional[bool] = None
     enable_step_logs: Optional[bool] = None
-    step_operator: Optional[str] = None
-    experiment_tracker: Optional[str] = None
+    step_operator: Optional[Union[bool, str]] = None
```
bcdurak, line 148: Are we planning to remove the `str` option in the future? 

If so, should we add a deprecation warning somewhere if people still use it?

bcdurak: On the other hand, if we allow multiple instances of the same type of component in the future, it might make sense to keep the `str` option as well.

bcdurak (approved):
Looks good overall, left a tiny comment.

PR #3778, opened 2025-06-25
Orchestrator pod logs for Kubernetes

bcdurak:
## Describe changes
Additionally, adds functionality for pipeline runs to feature more than 1 log response.

## Pre-requisites

## Types of changes


src/zenml/logging/step_logging.py
```diff
@@ -584,3 +587,60 @@ def wrapped_flush(*args: Any, **kwargs: Any) -> Any:
...
+        # Fetch the active stack
+        client = Client()
+        active_stack = client.active_stack
+
+        # Configure the logs
+        logs_uri = prepare_logs_uri(
+            artifact_store=active_stack.artifact_store,
+        )
+
+        logs_context = PipelineLogsStorageContext(
```
schustmi, line 639: This is missing a check whether logging is disabled via the env variable or the setting, no?

src/zenml/models/v2/core/logs.py
```diff
@@ -35,6 +35,11 @@ class LogsRequest(BaseRequest):
 
     uri: str = Field(title="The uri of the logs file")
 
+    title: Optional[str] = Field(
```
schustmi: We use `name` in 100 other place already, why not here?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -241,6 +241,7 @@ def launch(self) -> None:
 
             logs_model = LogsRequest(
                 uri=logs_uri,
+                descriptor="step",
```
schustmi: This is called `title` now

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -856,6 +856,7 @@ def _run(
 
                 logs_model = LogsRequest(
                     uri=logs_uri,
+                    descriptor="client",
```
schustmi: `title`

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -461,19 +463,30 @@ def run_logs(
...
-        raise KeyError("No logs available for this pipeline run")
-
-    return fetch_logs(
-        zen_store=store,
-        artifact_store_id=logs.artifact_store_id,
-        logs_uri=logs.uri,
-        offset=offset,
-        length=length,
-    )
+            if title is None or title == "workload":
```
schustmi: I would call this `runner` instead of `workload`.

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -461,19 +463,30 @@ def run_logs(
...
+                    workload_id=deployment.id
+                )
+                logs.append(("workload", workload_logs))
+
+    if run.log_collection:
+        for log_entry in run.log_collection:
+            if title is None or log_entry.title == title:
+                logs.append(
+                    (
+                        log_entry.title if log_entry.title else "",
```
schustmi: `log_entry.title or ""`

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -158,9 +158,9 @@ class PipelineRunSchema(NamedSchema, RunMetadataInterface, table=True):
             overlaps="run_metadata",
         ),
     )
-    logs: Optional["LogsSchema"] = Relationship(
+    logs: List["LogsSchema"] = Relationship(
```
schustmi, line 161: `Optional`

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6109,6 +6110,30 @@ def update_run(
...
+                    log_entry = LogsSchema(
+                        uri=log_request.uri,
+                        title=log_request.title,
+                        pipeline_run_id=existing_run.id,
+                        artifact_store_id=log_request.artifact_store_id,
+                    )
+                    session.add(log_entry)
+
+                session.commit()
+                session.refresh(existing_run)
```
schustmi: These refresh calls are DB calls each time, and are not necessary in this case I think. Can you try without it and see if it still shows up in the response of the update endpoint?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8890,6 +8915,7 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
 
                 log_entry = LogsSchema(
                     uri=step_run.logs.uri,
+                    descriptor=getattr(step_run.logs, "descriptor", "step"),
```
schustmi: `title`

schustmi (changes requested):

src/zenml/models/v2/core/logs.py
bcdurak: As we discussed, I am renaming this to `source`.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: The relationship between the two will change with this PR, won't it?

src/zenml/logging/step_logging.py
bcdurak: Added the missing check.

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: Renamed it.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Good catch, thanks :)

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
schustmi: Ah you're right, my bad

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -537,7 +537,7 @@ def to_model(
                 if self.model_version
                 else None,
                 tags=[tag.to_model() for tag in self.tags],
```
schustmi, line 547: Should we also add the `client` one in the response as `log=...` for backwards compatibility? And then remove it in a few versions

bcdurak: Good idea. One little detail to note down here: The `source` entries are not unique, meaning there could potentially be more than 1 entry with `client` as a source. So, since we are just doing this for backwards compatibility, I have used the first one in the list if there is more than one.

schustmi: Yeah seems reasonable

src/zenml/orchestrators/step_launcher.py
```diff
@@ -241,6 +241,7 @@ def launch(self) -> None:
 
             logs_model = LogsRequest(
                 uri=logs_uri,
+                source="orchestrator",
```
schustmi: I think in other places we call this environment where we're collecting logs "execution environment". Do you think it therefore makes sense to also call this the "execution" logs? The source for the logs is the step execution I guess?

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -436,22 +436,21 @@ def stop_run(
 @async_fastapi_endpoint_wrapper
 def run_logs(
     run_id: UUID,
+    source: Optional[str] = None,
     offset: int = 0,
     length: int = 1024 * 1024 * 16,  # Default to 16MiB of data
     _: AuthContext = Security(authorize),
-) -> str:
+) -> List[Tuple[str, str]]:
```
schustmi: This is now a breaking change and makes old clients incompatible with this server version, if they call this endpoint. Is that any concern you think?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -531,13 +531,22 @@ def to_model(
 
         resources = None
         if include_resources:
+            # Add the client logs as "logs" if they exist, for backwards compatibility
+            # TODO: This will be safe to remove in future releases (>0.84.0).
+            client_logs = [
+                log_entry
+                for log_entry in self.logs
+                if log_entry.source == "client"
```
schustmi, line 539: The DB migration to set this for old versions is still missing right?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8889,6 +8913,7 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
 
                 log_entry = LogsSchema(
                     uri=step_run.logs.uri,
+                    source=getattr(step_run.logs, "source", "step"),
```
schustmi: Why the getattr here?

schustmi (changes requested):

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: You are right, this is a breaking change. However, as far as I know, the only place where "we" are using this is the dashboard, and Julian is already aware of the changes. I am not sure if there are any external users who explicitly use this endpoint over the API.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Removed.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: You can see the migration script now.

src/zenml/orchestrators/step_launcher.py
bcdurak: Just checked the docs, you are completely right. However, to be completely honest, I think the word "execution" is way more confusing than "orchestrator" here.

schustmi: How can this be more confusing than using the same source as the logs coming from the orchestrator pod? 😄

src/zenml/zen_stores/migrations/versions/85289fea86ff_adding_source_to_logs.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    # Set source to "client" where pipeline_run_id is present but step_run_id is null
+    connection.execute(
+        sa.text("""
+            UPDATE logs 
+            SET source = 'client' 
+            WHERE pipeline_run_id IS NOT NULL AND step_run_id IS NULL
+        """)
+    )
```
schustmi, line 46: These two statements do not guarantee that all logs now have a source, and therefore the migration might fail later when making that column non-nullable. I would suggest simply using a `WHERE source IS NULL` in the second one, as there isn't really any other case anyway

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6109,6 +6110,29 @@ def update_run(
...
+                    # Create the log entry
+                    log_entry = LogsSchema(
+                        uri=log_request.uri,
+                        source=log_request.source,
+                        pipeline_run_id=existing_run.id,
+                        artifact_store_id=log_request.artifact_store_id,
+                    )
+                    session.add(log_entry)
+
+                session.commit()
```
schustmi: This might now run into `IntegrityErrors` that we should probably catch and ignore, or convert to a better error for the user? Probably in this case it should simply raise an `EntityExistsError` which returns a 409 conflict HTTP error

bcdurak: Good catch. There were three instances where the `LogsSchema` was being used: creating a run, updating a run, and creating a step. While the first and the third scenarios are very unlikely to yield `IntegrityError`s, I wrapped all three of them to catch any errors.

src/zenml/zen_stores/migrations/versions/85289fea86ff_adding_source_to_logs.py
bcdurak: You are right. I think this is theoretically impossible anyhow, because we remove the log entry in case the pipeline or the step gets removed. So, I can not think of a scenario, but I will modify the second one. 

Keep in mind, in this case, we can possibly create two entries that are "client-NULL-NULL"  and "client-NULL-NULL" in terms of source-pipeline_id-step_id. AFAIK, this should not break the unique constraint between these three values on MySQL or SQLite.

src/zenml/orchestrators/step_launcher.py
bcdurak: This is completely on me 😅 I thought we were discussing the LogsRequest within the orchestrator scope, and I was completely confused by your request.

You are right, let's name it "execution".

schustmi (approved):

PR #3779, opened 2025-06-26
Add field descriptions to config for forms

AlexejPenner:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/stack/flavor.py
+++ b/src/zenml/stack/flavor.py
@@ -233,31 +233,17 @@ class Flavor:
         """Generate SDK docs url for a flavor.
 
         Returns:
-            The complete url to the zenml SDK docs
+            The complete url to the zenml SDK docs pointing to component configuration
         """
         from zenml import __version__
 
         base = f"https://sdkdocs.zenml.io/{__version__}"
-
-        component_type = self.type.plural
-
-        if "zenml.integrations" in self.__module__:
-            # Get integration name out of module path which will look something
-            #  like this "zenml.integrations.<integration>....
-            integration = self.__module__.split(
-                "zenml.integrations.", maxsplit=1
-            )[1].split(".")[0]
-
-            return (
-                f"{base}/integration_code_docs"
-                f"/integrations-{integration}/#{self.__module__}"
-            )
-
-        else:
-            return (
-                f"{base}/core_code_docs/core-{component_type}/"
-                f"#{self.__module__}"
-            )
+        
+        # Get the config class name to point to the specific config documentation
+        config_class_name = self.config_class.__name__
+        
+        # Point to the specific config class documentation page
+        return f"{base}/component_configs/{config_class_name}.html"
 
 
 def validate_flavor_source(
--- a/src/zenml/artifact_stores/base_artifact_store.py
+++ b/src/zenml/artifact_stores/base_artifact_store.py
@@ -33,7 +33,7 @@ from typing import (
     cast,
 )
 
-from pydantic import model_validator
+from pydantic import Field, model_validator
 
 from zenml.constants import ENV_ZENML_SERVER
 from zenml.enums import StackComponentType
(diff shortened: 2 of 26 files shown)


src/zenml/stack/flavor.py
```diff
@@ -233,31 +233,32 @@ def generate_default_sdk_docs_url(self) -> str:
...
+        impl_class = self.implementation_class
+        module_name = impl_class.__module__
+
+        # Convert module path to docs URL format
+        # e.g., zenml.orchestrators.local.local_orchestrator -> core_code_docs/core-orchestrators
+        if module_name.startswith("zenml.integrations."):
+            # For integrations: zenml.integrations.kubernetes -> integrations-kubernetes
+            integration_name = module_name.split(".")[2]
+            docs_path = (
+                f"integration_code_docs/integrations-{integration_name}.html"
```
schustmi: Where is this `.html` suffix coming from?

src/zenml/stack/flavor.py
```diff
@@ -233,31 +233,32 @@ def generate_default_sdk_docs_url(self) -> str:
...
             )
-
+            anchor = f"#zenml.integrations.{integration_name}-modules"
```
schustmi: Where is this `-modules` coming from suddenly?

src/zenml/stack/flavor.py
```diff
@@ -233,31 +233,32 @@ def generate_default_sdk_docs_url(self) -> str:
...
-            #  like this "zenml.integrations.<integration>....
-            integration = self.__module__.split(
-                "zenml.integrations.", maxsplit=1
-            )[1].split(".")[0]
-
-            return (
-                f"{base}/integration_code_docs"
-                f"/integrations-{integration}/#{self.__module__}"
+        # Get the implementation class module and name
+        impl_class = self.implementation_class
```
schustmi: Why does this need to import the implementation class? Server-side, this will for most integrations simply fail with an `ImportError` because the dependencies are not installed.

src/zenml/stack/flavor.py
AlexejPenner: 1) the new stuff is not for the sdk docs - its so we can have on hover tool tips in the creation forms

2) I am equally confused why the sdk docs suddenly changed to have this suffix - maybe I introduced this with my PR that fixed the PR - maybe something in our toolchain for sdk docs changed

src/zenml/stack/flavor.py
AlexejPenner: same as my notes above

src/zenml/stack/flavor.py
AlexejPenner: i think this one I need to get rid of, At some point i needed to resort to thid for some sdk docs experiments without a thought to the implications and forgot to revbert it

AlexejPenner: ah no, actually just oversight on my end

src/zenml/stack/flavor.py
AlexejPenner: https://sdkdocs.zenml.io/0.83.0/integration_code_docs/integrations-kubernetes.html#zenml.integrations.kubernetes-modules check this out - it scrolls me to the right place in the loooon gpage

schustmi: Yes, but check out this (the previous code): https://sdkdocs.zenml.io/0.83.0/integration_code_docs/integrations-kubernetes#zenml.integrations.kubernetes.flavors.kubernetes_orchestrator_flavor

It doesn't just scroll you to a random modules section of the page, but instead to exactly the module that you're interested in.

You can even point it to a specific class if you want that instead, e.g. https://sdkdocs.zenml.io/0.83.0/integration_code_docs/integrations-kubernetes#zenml.integrations.kubernetes.flavors.KubernetesOrchestratorSettings

schustmi (changes requested):
@AlexejPenner See my latest reply regarding the sdk docs link.

Also, my main question has not been answered yet: Currently, our SDK docs include the `Attributes` for the Settings/Config classes like this:

Screenshot 2025-07-09 at 10 25 50

How does this work with your changes? I assume it doesn't at all, which means there will be no mention of any attribute in the SDK docs. IMO that's unacceptable, as this is the core information that we want here actually. We need to find a way how to either parse these field descriptions into the docstring in our sdk docs helper or solve this otherwise.

schustmi (approved):

PR #3789, opened 2025-07-01
Step retries

schustmi:
## Describe changes

This PR enables step retries in the backend, and implements a generic in-process solution for all orchestrators as well as a more sophisticated solution for the Kubernetes orchestrator.

### Step retries
- A pipeline run can now contain multiple step runs with the same name, but they need to have different versions. This version gets incremented each time a new step gets registered.
- When a step fails, and the version has not yet reached the retry limit yet, the step status gets changed to `RETRYING` instead of `FAILED`.
- When a new version for a step run gets registered, the status of all previous versions gets set to `RETRIED`
- The DAG visualizer for now only contains the latest version of the step runs and does not show the previous, retried versions.

### Generic in-process orchestrator retries
We previously had an implementation that retried **just the user code** in the same process if the user code failed. This PR changes that so we retry not just the user code, but also the code leading up to this (cache computation, creating the step, ...).

### Kubernetes-specific retries
The Kubernetes orchestrator does not handle the retries in-process, but instead the implementation switches from using raw pods to using jobs. These jobs are configured to handle the retry behaviour according to the user configuration.

### Future improvements
- Create step run earlier in the entrypoint
- Use different exit code
- Limit backoff manually while monitoring job
- Enable step logs earlier

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/orchestrators/kubernetes.md
+++ b/docs/book/component-guide/orchestrators/kubernetes.md
@@ -479,7 +479,7 @@ docs page.
 
 ## Best practices for highly parallel pipelines
 
-If you're trying to run pipelines with mutliple parallel steps, there are some configuration options that you can tweak to ensure the best possible performance:
+If you're trying to run pipelines with multiple parallel steps, there are some configuration options that you can tweak to ensure the best possible performance:
 - Ensure you enable [retries for your steps](../../how-to/steps-pipelines/advanced_features.md#automatic-step-retries) in case something doesn't work
 - Add a `backoff_limit_margin` to deal with unexpected Kubernetes evictions/preemptions
 - Limit the amount of maximum parallel steps using the `max_parallelism` setting


htahir1 (changes requested):
One case not handled is if you do `zenml login` once and authenticate through ZenML Pro, then do zenml login again it doesnt show the questionnaire anymore. Is this what we want? 


image

The underlying reason is that `zenml login` is also used to switch around between workspaces once authenticated to pro organization

so you can do: `zenml login internal` and `zenml login demo-tenant` to switch between workspaces.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -253,34 +255,81 @@ def run_step_on_kubernetes(step_name: str) -> None:
...
+        job_manifest = build_job_manifest(
+            job_name=pod_name,
+            pod_template=pod_template_manifest_from_pod(pod_manifest),
+            backoff_limit=backoff_limit,
+            ttl_seconds_after_finished=settings.ttl_seconds_after_finished,
+            active_deadline_seconds=settings.active_deadline_seconds,
+            pod_failure_policy=pod_failure_policy,
+        )
+
+        kube_utils.create_job(
```
schustmi: @stefannica Do you think we should also wrap these `create_job` and `wait_for_job_to_finish` calls in some retries? Seems like under load, even these requests to fetch pod logs and create jobs might run into issues

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -253,34 +255,81 @@ def run_step_on_kubernetes(step_name: str) -> None:
             labels=step_pod_labels,
         )
 
-        kube_utils.create_and_wait_for_pod_to_start(
-            core_api=core_api,
-            pod_display_name=f"pod for step `{step_name}`",
-            pod_name=pod_name,
-            pod_manifest=pod_manifest,
+        retry_config = step_config.retry
+        backoff_limit = retry_config.max_retries if retry_config else 0
```
schustmi: @stefannica Do you think simply adding a few retries to this (or even making this `+X tries` configurable on the orchestrator) is a good idea? Worst case, we retry a few more times, and all those tries will be rejected by the server saying "Retry limit reached for step". But it would increase the likelihood of us making sure that the step gets retried its configured maximum.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
stefannica: Tough call. I don't really know the reasons why creating a job might fail. After all, it's not a workload, like a pod, it's more like a secret: It shouldn't fail if the cluster doesn't have resources. There are very few reasons why creating a job might fail and it doesn't have anything to do with how many pods or nodes there are.

On the other hand, we recently learned that the Kubernetes python client doesn't retry calls that are retry-worthy, like 400 errors. So yeah, I would retry these too, as long as retries are idempotent (which they should be, because this is a declarative resource).

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
stefannica: I was thinking about suggesting this same thing when we were talking about this: retrying this a few more times in addition to its configured maximum to ensure that we reach the target maximum or retries. I'm 100% for doing this, with a configurable parameter controlling it.

schustmi: I added this with a configurable parameter that defaults to `0`.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
schustmi: Implemented, but currently not limited to any error codes. Is that something that you think we should add?

htahir1 (approved):

bcdurak (approved):
I have tested the stopping functionality and it seems to be working 👍

PR #3790, opened 2025-07-01
New zenml login and CLI style improvements

znegrin:
## Describe changes
I implemented an interactive login menu for the `zenml login` command to achieve a better user experience by providing clear login options instead of automatically defaulting to ZenML Pro.
When users run `zenml login` without arguments, they now see a styled interactive menu with three options:

1. Login to a local server (`zenml login --local`)
2. Login to ZenML Pro (https://cloud.zenml.io)
3. Login to a custom cloud server (prompts for URL)

This also includes CLI styling improvements across multiple files for better visual consistency and remove unnecessary colors to make it more legible.

## Pre-requisites

## Types of changes


htahir1 (changes requested):
One case not handled is if you do `zenml login` once and authenticate through ZenML Pro, then do zenml login again it doesnt show the questionnaire anymore. Is this what we want? 


image

The underlying reason is that `zenml login` is also used to switch around between workspaces once authenticated to pro organization

so you can do: `zenml login internal` and `zenml login demo-tenant` to switch between workspaces.

znegrin:
> One case not handled is if you do `zenml login` once and authenticate through ZenML Pro, then do zenml login again it doesnt show the questionnaire anymore. Is this what we want?

Right now how it works is that if you do `zenml login` again and you are already logged to ZenML Pro, it goes automatically to the selection screen again, which kind of makes sense as it is the expected behavior to switch to a new workspace, right?

src/zenml/cli/login.py
```diff
@@ -46,6 +49,57 @@
...
+            default="2"
+        )
+        
+        if choice == "1":
+            return "local"
+        elif choice == "2":
+            return "pro" 
+        elif choice == "3":
+            return "cloud"
+
```
htahir1: might want to turn this into an Enum (there is a enums.py)

src/zenml/cli/login.py
```diff
@@ -895,20 +951,67 @@ def login(
...
+                restart=restart,
+            )
+        elif login_method == "pro":
+            # Connect to ZenML Pro
+            connect_to_pro_server(
+                api_key=api_key_value,
+                pro_api_url=pro_api_url,
+                verify_ssl=verify_ssl,
+            )
+        elif login_method == "cloud":
```
htahir1: use enums here as i mentioned above

src/zenml/cli/login.py
```diff
@@ -895,20 +951,67 @@ def login(
...
+            # Get custom server URL from user
+            console.print()
+            server_url = Prompt.ask(
+                "[bold purple]Enter the ZenML server URL[/bold purple]",
+                default="https://"
```
htahir1: the default here is wrong.. it should somehow validate that it starts with "http" or something and not exit until such an address is provided

htahir1 (changes requested):
ok i agree. left some more comments

src/zenml/cli/login.py
znegrin: Replaced it with (http/https) as those are the ones it is validating right now. Also changed the color of the text.

htahir1 (approved):
For me this looks good. but now we need @bcdurak to take a look too

htahir1:
Closes #3791

src/zenml/cli/utils.py
```diff
@@ -186,6 +197,40 @@ def warning(
...
+        styled_text = Text()
+        styled_text.append("✔ ", style="green bold")  # Explicit green for checkmark
+        styled_text.append(text, style=style)
+        
+        console.print(styled_text, **kwargs)
+    else:
+        # If checkmark already exists, apply success style to entire text
+        base_style = zenml_style_defaults["success"]
+        style = Style.chain(base_style, Style(bold=bold, italic=italic))
+        console.print(text, style=style, **kwargs)
```
bcdurak:
```suggestion
    base_style = zenml_style_defaults["success"]
    style = Style.chain(base_style, Style(bold=bold, italic=italic))
    console.print(text, style=style, **kwargs)
```
I would not forcefully add a checkmark if the original message does not include it. Otherwise, I love the new success option.

src/zenml/console.py
```diff
@@ -18,11 +18,15 @@
...
-    "info": Style(color="cyan", dim=True),
+    "info": Style(color="white", dim=True),
     "warning": Style(color="yellow"),
     "danger": Style(color="red", bold=True),
     "title": Style(color="cyan", bold=True, underline=True),
     "error": Style(color="red"),
+    "success": Style(color="green"),
+    "repr.str": Style(color="white", dim=False),
+    "repr.uuid": Style(color="magenta", dim=False),
+    "tenant": Style(color="white", dim=False, bold=False),
```
bcdurak: There are already a few instances where we used tenant (or better yet, workspace) names. However, this has been used only once. For consistency and to clean up, I would remove this line from here. (The line in the `/src/zenml/cli/login.py` needs to be removed as well.)

```suggestion
```

src/zenml/cli/login.py
```diff
@@ -424,7 +495,7 @@ def connect_to_pro_server(
         )
 
     cli_utils.declare(
-        f"Connecting to ZenML Pro server: {server.name} [{str(server.id)}] "
+        f"Connecting to ZenML Pro server: [tenant]{server.name}[/tenant] [{str(server.id)}] "
```
bcdurak:
```suggestion
        f"Connecting to ZenML Pro server: {server.name} [{str(server.id)}] "
```

src/zenml/logger.py
```diff
@@ -77,7 +78,11 @@ def format(self, record: logging.LogRecord) -> str:
         if ZENML_LOGGING_COLORS_DISABLED:
             # If color formatting is disabled, use the default format without colors
             formatter = logging.Formatter(self.format_template)
-            return formatter.format(record)
+            formatted_message = formatter.format(record)
+            # Add warning icon for warning messages even when colors are disabled
+            if record.levelno == LoggingLevels.WARN.value:
+                formatted_message = "⚠ " + formatted_message
+            return formatted_message
```
bcdurak:
```suggestion
            return formatter.format(record)
```

Similar to the checkmark, I would not alter the message with a warning sign here.

src/zenml/logger.py
```diff
@@ -87,12 +92,16 @@ def format(self, record: logging.LogRecord) -> str:
             )
             formatter = logging.Formatter(log_fmt)
             formatted_message = formatter.format(record)
+            
+            # Add warning icon for warning messages
+            if record.levelno == LoggingLevels.WARN.value:
+                formatted_message = self.yellow + "⚠ " + self.reset + formatted_message
```
bcdurak:
```suggestion
```

Same as above.

src/zenml/cli/login.py
schustmi: `tenant` should be `workspace` I assume?

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -882,9 +882,12 @@ def _run(
                 if run:
                     run_url = dashboard_utils.get_run_url(run)
                     if run_url:
-                        logger.info(
-                            f"Dashboard URL for Pipeline Run: {run_url}"
-                        )
+                        # Display the dashboard URL directly without panel to avoid wrapping
```
schustmi: This `zenml.console` is a CLI class and should not be used in the regular python client code.

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -882,9 +882,12 @@ def _run(
...
+                        from zenml.console import console
+                        
+                        console.print()
+                        console.print(f"Dashboard URL: [bright_blue underline]{run_url}[/bright_blue underline]", overflow="ignore", crop=False)
+                        console.print()
```
bcdurak:
```suggestion
                        logger.info(
                            f"Dashboard URL for Pipeline Run: {run_url}"
                        )
```

I would revert back to the logger here. We do not use the console outside of the CLI.

bcdurak (changes requested):

src/zenml/cli/utils.py
znegrin: @bcdurak maybe we can add the checkmark directly to specific messages where it makes sense and remove this logic?

znegrin: Fixed in a commit, adding the checkmark to the affected messages directly instead of using the logic.

src/zenml/cli/login.py
znegrin: @bcdurak I fixed it, replacing the tenant labels with just two '' so it looks white in the CLI as other strings.

znegrin: I removed it and replaced it with `'{server.name}'` so it is formated like other server urls or names

znegrin:
@bcdurak @htahir1 @schustmi I think I covered all the issues and also passed the scripts for the tests. I hope it is all ok now. Please let me know if there's something else I can fix or do in a better way, as I am more used to work with frontend stuff, and it is my first time working on changes for a CLI interface. Thank you!

bcdurak (approved):

PR #3793, opened 2025-07-01
Add support for external service accounts

stefannica:
## Describe changes
Extend the external authenticator support to differentiate between external service accounts and external user accounts. This paves the way forward to a migration path from workspace-level service accounts to ZenML Pro global service-accounts.

* add ability to configure avatars for users and service accounts (used to sync ZenML Pro user account avatars).
* add ability to configure display names for service accounts (used to sync ZenML Pro service account display names ).
* existing users and service accounts are "adopted" if they exist and have the same username, to allow migrating from workspace-level service accounts to ZenML Pro global service accounts.
* update ZenML Pro self-hosted docs to replace `email` with `username`
* prevent most actions to be used with external service accounts (update/delete and create/update/rotate API keys)

## Pre-requisites

## Types of changes


src/zenml/models/v2/core/service_account.py
```diff
@@ -51,6 +52,10 @@ class ServiceAccountRequest(BaseRequest):
         title="A description of the service account.",
         max_length=TEXT_FIELD_MAX_LENGTH,
     )
+    external_user_id: Optional[UUID] = Field(
```
schustmi, line 72: This should probably not be something that regular users can set in this request (or the update model below)?

src/zenml/zen_server/auth.py
```diff
@@ -732,19 +737,84 @@ def authenticate_external_user(
...
+                        f"Service account with username {external_user.username} "
+                        f"already exists with a different external user ID: "
+                        f"{service_account.external_user_id}."
+                    )
+                # Update the user information
+                user = store.update_service_account(
+                    service_account_name_or_id=service_account.name,
+                    service_account_update=ServiceAccountUpdate(
+                        name=external_user.username,
+                        active=True,
```
schustmi: This made me think of something. What if someone disables a Pro service account, this will not get propagated here right? Is that an issue?

Also, should we prevent creating API keys for external service accounts?

src/zenml/zen_server/auth.py
```diff
@@ -732,19 +737,84 @@ def authenticate_external_user(
...
+                        external_user_id=external_user.id,
+                    ),
+                ).to_user_model()
+            except KeyError:
+                pass
+        else:
+            # Try finding a user with the same username
+            try:
+                user = store.get_user(user_name_or_id=external_user.username)
```
schustmi: Aren't we already in a catch of the `KeyError` because this user doesn't exist?

src/zenml/models/v2/core/service_account.py
stefannica: Yeah, good point. I moved this attribute to a `ServiceAccountInternalRequest` separate model class.

src/zenml/zen_server/auth.py
stefannica: > This made me think of something. What if someone disables a Pro service account, this will not get propagated here right? Is that an issue?

Short answer: no, this is not a problem. This is the same as disabling or deleting a ZenML Pro user: the workspace API token is valid for an hour, after which the ZenML Pro credentials are exchanged for a new workspace API token. If the user or service account is no longer active or is deleted, the user won't be allowed to re-authenticate to the ZenML Pro API  to gain access to the workspace.

> Also, should we prevent creating API keys for external service accounts?

Good catch !

src/zenml/zen_server/auth.py
stefannica: Yes, but the last time we tried looking the user up by its external user ID, here we're trying to find a user that has the same name. This check is useful for migrations of OSS servers to ZenML Pro: if the server contains user or service accounts with the same username as those that coming from ZenML Pro, we no longer throw an error, but instead we adopt the existing user.

src/zenml/models/v2/core/service_account.py
```diff
@@ -77,6 +87,10 @@ class ServiceAccountUpdate(BaseUpdate):
         title="Whether the service account is active or not.",
         default=None,
     )
+    external_user_id: Optional[UUID] = Field(
```
schustmi: Same as above, not for regular users. If it's to complicated to extend this model though, feel free to ignore

src/zenml/zen_server/auth.py
schustmi: Another point, when "converting" an existing service account to an external ones, right now existing API keys would still stay active, correct? Should we also delete them?

schustmi (changes requested):

src/zenml/models/v2/core/service_account.py
stefannica: I moved this attribute to a ServiceAccountInternalUpdate separate model class.

src/zenml/zen_server/auth.py
stefannica: I thought about keeping the API keys active for a while longer. At least until we deprecate and remove the workspace service accounts for ZenML Pro use. In fact, the API keys can stay there after deprecation and we'll just disable authenticating with them and managing them in ZenML Pro tenants.

stefannica: I just pushed a change that prevents using workspace API keys to login with a service account that has been "converted" (i.e. is an external service account now).

schustmi (approved):

PR #3812, opened 2025-07-07
Adding verbosity levels to log messages in storage

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/logger.py
+++ b/src/zenml/logger.py
@@ -51,10 +51,7 @@ class CustomFormatter(logging.Formatter):
     reset: str = "\x1b[0m"
 
     format_template: str = (
-        "%(asctime)s - %(name)s - %(levelname)s - %(message)s (%("
-        "filename)s:%(lineno)d)"
-        if LoggingLevels[ZENML_LOGGING_VERBOSITY] == LoggingLevels.DEBUG
-        else "%(message)s"
+        "[%(levelname)s] %(message)s (%(name)s:%(filename)s:%(lineno)d)"
     )
 
     COLORS: Dict[LoggingLevels, str] = {
--- a/src/zenml/logger.py
+++ b/src/zenml/logger.py
@@ -45,7 +45,7 @@ class LogCollectorRegistry:
 
     def __new__(cls) -> "LogCollectorRegistry":
         """Create a new instance of the LogCollectorRegistry.
-            
+
         Returns:
             The singleton instance of the LogCollectorRegistry.
         """
@@ -111,27 +111,21 @@ class LogCollectorRegistry:
 
     def _parse_and_clean_message(self, message: str) -> str:
         """Parse and clean ZenML log format tokens from message.
-        
+
         Args:
             message: The original message that may contain log level tokens.
-            
+
         Returns:
             The cleaned message with log level tokens and location info removed.
         """
         clean_message = message
-        
+
         # Try to remove log level tokens
         for level in LoggingLevels:
             level_token = f"[{getLevelName(level.value)}] "
             if level_token in clean_message:
                 clean_message = clean_message.replace(level_token, "")
-        
-        # Remove location information in format (module:file.py:line) from the end
-        # Pattern matches: (anything:anything.py:digits) at the very end
-        # The key is the .py: pattern which is unique to location info
-        location_pattern = r'\s*\([^)]*\.py:\d+\)\s*$'
-        clean_message = re.sub(location_pattern, '', clean_message)
-        
+
         return clean_message
 
     def _wrapped_write(self, is_stderr: bool = False) -> Callable:
--- a/src/zenml/logger.py
+++ b/src/zenml/logger.py
@@ -17,7 +17,8 @@ import logging
 import os
 import re
 import sys
-from typing import Any, Dict
+from logging import getLevelName
+from typing import Any, Callable, Dict, List, Optional
 
 from rich.traceback import install as rich_tb_install
 
@@ -34,6 +35,172 @@ from zenml.enums import LoggingLevels
 ZENML_LOGGING_COLORS_DISABLED = handle_bool_env_var(
     ENV_ZENML_LOGGING_COLORS_DISABLED, False
 )
+
+
+class LogCollectorRegistry:
+    """Singleton registry for log collectors."""
+
+    _instance: Optional["LogCollectorRegistry"] = None
+    _initialized: bool = False
+
+    def __new__(cls) -> "LogCollectorRegistry":
+        """Create a new instance of the LogCollectorRegistry.
+            
+        Returns:
+            The singleton instance of the LogCollectorRegistry.
+        """
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
+    def __init__(self) -> None:
+        """Initialize the LogCollectorRegistry.
+
+        Returns:
+            The singleton instance of the LogCollectorRegistry.
+        """
+        if not self._initialized:
+            self.collectors: List[Callable[[str, bool], Any]] = []
+            self.original_stdout_write: Optional[Callable] = None
+            self.original_stderr_write: Optional[Callable] = None
+            self.original_stdout_flush: Optional[Callable] = None
+            self.original_stderr_flush: Optional[Callable] = None
+            self._stdout_stderr_wrapped: bool = False
+            self._initialized = True
+
+    def add_collector(self, collector: Callable[[str, bool], Any]) -> None:
+        """Add a collector to handle log messages.
+
+        Args:
+            collector: A callable that takes (message: str, is_stderr: bool) and processes it.
+        """
+        # Setup stdout/stderr wrapping on first registration
+        if not self._stdout_stderr_wrapped:
+            self._setup_stdout_stderr_wrapping()
+
+        if collector not in self.collectors:
+            self.collectors.append(collector)
+
+    def remove_collector(self, collector: Callable[[str, bool], Any]) -> None:
+        """Remove a collector from the registry.
+
+        Args:
+            collector: The collector to remove.
+        """
+        if collector in self.collectors:
+            self.collectors.remove(collector)
+
+    def _default_collector(self, message: str, is_stderr: bool) -> Any:
+        """Default collector that writes to original stdout/stderr.
+
+        Args:
+            message: The message to write.
+            is_stderr: Whether this is stderr output.
+
+        Returns:
+            The result of the original write method.
+        """
+        # Parse out log level tokens before writing to console
+        clean_message = self._parse_and_clean_message(message)
+
+        if is_stderr and self.original_stderr_write:
+            return self.original_stderr_write(clean_message)
+        elif not is_stderr and self.original_stdout_write:
+            return self.original_stdout_write(clean_message)
+        return None
+
+    def _parse_and_clean_message(self, message: str) -> str:
+        """Parse and clean ZenML log format tokens from message.
+        
+        Args:
+            message: The original message that may contain log level tokens.
+            
+        Returns:
+            The cleaned message with log level tokens and location info removed.
+        """
+        clean_message = message
+        
+        # Try to remove log level tokens
+        for level in LoggingLevels:
+            level_token = f"[{getLevelName(level.value)}] "
+            if level_token in clean_message:
+                clean_message = clean_message.replace(level_token, "")
+        
+        # Remove location information in format (module:file.py:line) from the end
+        # Pattern matches: (anything:anything.py:digits) at the very end
+        # The key is the .py: pattern which is unique to location info
+        location_pattern = r'\s*\([^)]*\.py:\d+\)\s*$'
+        clean_message = re.sub(location_pattern, '', clean_message)
+        
+        return clean_message
+
+    def _wrapped_write(self, is_stderr: bool = False) -> Callable:
+        """Create a wrapped write function.
+
+        Args:
+            is_stderr: Whether this is for stderr.
+
+        Returns:
+            The wrapped write function.
+        """
+
+        def wrapped_write(*args: Any, **kwargs: Any) -> Any:
+            message = args[0]
+            result = None
+
+            # Call all collectors in sequence
+            for collector in self.collectors:
+                try:
+                    result = collector(message, is_stderr)
+                except Exception as e:
+                    # If a collector fails, log the error but continue
+                    print(f"Error in log collector: {e}", file=sys.stderr)
+
+            return result
+
+        return wrapped_write
+
+    def _wrapped_flush(self, is_stderr: bool = False) -> Callable:
+        """Create a wrapped flush function.
+
+        Args:
+            is_stderr: Whether this is for stderr.
+
+        Returns:
+            The wrapped flush function.
+        """
+
+        def wrapped_flush(*args: Any, **kwargs: Any) -> Any:
+            original_flush = (
+                self.original_stderr_flush
+                if is_stderr
+                else self.original_stdout_flush
+            )
+            if original_flush:
+                return original_flush(*args, **kwargs)
+            return None
+
+        return wrapped_flush
+
+    def _setup_stdout_stderr_wrapping(self) -> None:
+        """Set up stdout and stderr wrapping."""
+        if not self._stdout_stderr_wrapped:
+            # Store original methods
+            self.original_stdout_write = getattr(sys.stdout, "write")
+            self.original_stdout_flush = getattr(sys.stdout, "flush")
+            self.original_stderr_write = getattr(sys.stderr, "write")
+            self.original_stderr_flush = getattr(sys.stderr, "flush")
+
+            # Add default collector directly to avoid recursion
+            self.collectors.append(self._default_collector)
+
+            # Wrap stdout and stderr
+            setattr(sys.stdout, "write", self._wrapped_write(is_stderr=False))
+            setattr(sys.stdout, "flush", self._wrapped_flush(is_stderr=False))
+            setattr(sys.stderr, "write", self._wrapped_write(is_stderr=True))
+            setattr(sys.stderr, "flush", self._wrapped_flush(is_stderr=True))
+
+            self._stdout_stderr_wrapped = True
 
 
 class CustomFormatter(logging.Formatter):
@@ -149,7 +316,7 @@ def get_formatter() -> logging.Formatter:
         The formatter.
     """
     if log_format := os.environ.get(ENV_ZENML_LOGGING_FORMAT, None):
-        return logging.Formatter(fmt=log_format)
+        return logging.Formatter(fmt=f"[%(levelname)s] {log_format}")
     else:
         return CustomFormatter()
 
@@ -191,6 +358,9 @@ def init_logging() -> None:
     console_handler = logging.StreamHandler(sys.stdout)
     console_handler.setFormatter(get_formatter())
     logging.root.addHandler(console_handler)
+
+    # Initialize the singleton registry (wrapping happens on first registration)
+    LogCollectorRegistry()
 
     # Enable logs if environment variable SUPPRESS_ZENML_LOGS is not set to True
     suppress_zenml_logs: bool = handle_bool_env_var(
--- a/src/zenml/logging/step_logging.py
+++ b/src/zenml/logging/step_logging.py
@@ -20,7 +20,7 @@ import sys
 import time
 from contextvars import ContextVar
 from types import TracebackType
-from typing import Any, Callable, List, Optional, Type, Union
+from typing import Any, Callable, List, Optional, Tuple, Type, Union
 from uuid import UUID, uuid4
 
 from zenml import get_step_context
@@ -34,6 +34,7 @@ from zenml.constants import (
     ENV_ZENML_DISABLE_STEP_NAMES_IN_LOGS,
     handle_bool_env_var,
 )
+from zenml.enums import LoggingLevels
 from zenml.exceptions import DoesNotExistException
 from zenml.logger import get_logger
 from zenml.logging import (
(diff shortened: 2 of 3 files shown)


src/zenml/logger.py
```diff
@@ -50,12 +50,20 @@ class CustomFormatter(logging.Formatter):
...
+
+        Args:
+            record: The log record to format.
+
+        Returns:
+            The format template string.
+        """
+        # Only include location info for DEBUG level
+        if get_logging_level() == LoggingLevels.DEBUG:
+            return "[%(levelname)s] %(message)s (%(name)s:%(filename)s:%(lineno)d)"
```
schustmi: Why did you remove the time here? I think when setting the level to debug, it's very important to also have these printed to the console

src/zenml/logging/step_logging.py
```diff
@@ -33,12 +32,15 @@
...
+    ENV_ZENML_CAPTURE_PRINTS,
     ENV_ZENML_DISABLE_PIPELINE_LOGS_STORAGE,
     ENV_ZENML_DISABLE_STEP_NAMES_IN_LOGS,
+    ENV_ZENML_STORAGE_LOGGING_VERBOSITY,
     handle_bool_env_var,
 )
 from zenml.exceptions import DoesNotExistException
-from zenml.logger import get_logger
+
+# Removed get_logger import to avoid circular dependency
```
schustmi: This is something that should be commented on the GH PR as an explanation, not in code right?

src/zenml/logging/step_logging.py
```diff
@@ -279,11 +392,7 @@ def write(self, text: str) -> None:
 
         if not self.disabled:
             # Add timestamp to the message when it's received
```
schustmi: Not happening anymore

src/zenml/logging/step_logging.py
```diff
@@ -472,28 +583,68 @@ def __init__(
...
+        self.artifact_store_handler.setFormatter(storage_formatter)
+
+        # Add handler to root logger
+        root_logger = logging.getLogger()
+        root_logger.addHandler(self.artifact_store_handler)
+
+        # Set root logger level to minimum of console and storage levels
+        # This ensures records can reach both handlers
+        self.original_root_level = root_logger.level
+        storage_level = getattr(
```
schustmi: Can crash if set to invalid value. I think similar code is also written above, maybe useful to have this in a helper function and fall back to info?

src/zenml/logging/step_logging.py
```diff
@@ -472,28 +583,68 @@ def __init__(
...
+                ENV_ZENML_STORAGE_LOGGING_VERBOSITY, "INFO"
+            ).upper(),
+        )
+        console_level = root_logger.level
 
-        self.stderr_write = getattr(sys.stderr, "write")
-        self.stderr_flush = getattr(sys.stderr, "flush")
+        min_level = min(storage_level, console_level)
+        if min_level < root_logger.level:
+            root_logger.setLevel(min_level)
```
schustmi, line 777: If someone adds separate handlers to the root logger, they will still only get the correct messages as the handlers themselves specify a level, correct?

src/zenml/logging/step_logging.py
```diff
@@ -472,28 +583,68 @@ def __init__(
...
+            ENV_ZENML_DISABLE_STEP_NAMES_IN_LOGS, default=None
+        )
+
+        if step_names_disabled is None:
+            # If env var is not set, use the prepend_step_name setting
+            # Pipeline context: prepend_step_name=False -> show step names (False)
+            # Step context: prepend_step_name=True -> don't show step names (True)
+            step_names_in_console.set(self.prepend_step_name)
+        else:
+            # If env var disables step names, always set to False
```
schustmi: What this comment describes is not happening here, if `step_names_disabled` is `False`, you always set it to True instead of respecting `self.prepend_step_name`

src/zenml/logging/step_logging.py
```diff
@@ -511,15 +662,33 @@ def __exit__(
             exc_val: The instance of the exception
             exc_tb: The traceback of the exception
 
-        Restores the `write` method of both stderr and stdout.
+        Removes the handler from loggers and context variables.
         """
+        # Remove handler from root logger and restore original level
+        if self.artifact_store_handler:
+            root_logger = logging.getLogger()
+            root_logger.removeHandler(self.artifact_store_handler)
```
schustmi: Should have a check or catch in error if something removed the handler before.

src/zenml/logging/step_logging.py
```diff
@@ -52,15 +54,126 @@
...
+logger = logging.getLogger(__name__)
 
+# Context variables
 redirected: ContextVar[bool] = ContextVar("redirected", default=False)
+step_names_in_console: ContextVar[bool] = ContextVar(
+    "step_names_in_console", default=False
+)
+
+# Context variables for active handlers
+active_stdout_handlers: ContextVar[List[logging.Handler]] = ContextVar(
```
schustmi: Why do we need these separate if we only ever include the artifact store handlers in both, and the separation is already happening with the log level?

Shouldn't this simply be a list of artifact store handlers, that we call from the wrapped print?

src/zenml/logger.py
bcdurak: I did for the previous approach. Now the console and storage are two separate methods, I am reverting the changes here, as you suggested.

src/zenml/logging/step_logging.py
bcdurak: Remnants of the old solution and some Claude comments 🤦 Removing them.

src/zenml/logging/step_logging.py
bcdurak: You are right, the separation is happening on the `LogRecord` level. 

To keep it a bit cleaner, I am reducing this down to a single list as you suggested.

src/zenml/logging/step_logging.py
bcdurak: removed.

src/zenml/logging/step_logging.py
bcdurak: Yeah, the `__init__` of the `ArtifactStoreHandler` was already covering this case but this one didn't. Since they check the same, I will implement a helper method.

bcdurak: I have checked how we are handling the actual logging verbosity and made this one parallel to it as well. That means, if I set it to an invalid value, the code will fail just like the regular verbosity.

src/zenml/logging/step_logging.py
bcdurak: The root logger will only process the messages if its level is set to the lowest level that its handlers are processing.

So, I made this refer to all the handlers now that are on the root logger.

src/zenml/logging/step_logging.py
bcdurak: Oh, I see how this can be confusing. I always thought these two would be on the same level. So I set the env variable to not disable it, that would be the end of it. I guess you can tell ZenML not to disable it over the env variable but yet still disable it over the parameter.

bcdurak: So I am adjusting it to this, let me know if this is what you had in mind:

- We don't show the step names only if `someone disables it through the env variable` or if `they explicitly put that setting to False in their config`

src/zenml/logging/step_logging.py
bcdurak: Added an additional if condition.

src/zenml/logging/step_logging.py
schustmi: Yes, but let's say the root logger is set to debug, but a user registers a handler with level info. Will their handler be triggered by the root logger?

src/zenml/logging/step_logging.py
```diff
@@ -52,15 +52,117 @@
 from zenml.utils.time_utils import utc_now
 from zenml.zen_stores.base_zen_store import BaseZenStore
 
-# Get the logger
 logger = get_logger(__name__)
 
+# Context variables
 redirected: ContextVar[bool] = ContextVar("redirected", default=False)
+step_names_in_console: ContextVar[bool] = ContextVar(
```
schustmi: Just for me to understand: This is only being used by the Formatter of the console handler, correct?

src/zenml/logging/step_logging.py
schustmi: I thought:
- Env variable is set to `True` -> Never include them
- Env variable is anything else (unset, False, invalid value) -> Do what `prepend_step_name` configures

schustmi (changes requested):

src/zenml/logging/step_logging.py
bcdurak: Yes

src/zenml/logging/step_logging.py
bcdurak: I have adjusted it already.

src/zenml/logging/step_logging.py
bcdurak: Yes. In your example, it works without a problem. Since the root logger is set to debug, it will process all messages with level info. 

However, the opposite is a bit more complicated. If the level of the root logger is set to info and someone appends a handler with level debug, they will have to adjust the level of the root logger:

```python
import logging

class TestHandler(logging.Handler):
    def emit(self, record):
        print(f"Handler received: {record.levelname} - {record.getMessage()}")

root_logger = logging.getLogger()
root_logger.setLevel(logging.INFO)  # Root logger at INFO

user_handler = TestHandler()
user_handler.setLevel(logging.DEBUG)  # Handler at DEBUG
root_logger.addHandler(user_handler)

root_logger.debug("This is a DEBUG message")
root_logger.info("This is an INFO message") 
```

This will only output: `Handler received: INFO - This is an INFO message` even though the handler is set to `DEBUG`.

bcdurak:
> One more thought I had: Now that we're not wrapping `stderr` anymore, exceptions probably won't show up in the logs anymore, correct?

Yes and no. We usually wrap the step and pipeline executions in try-except clauses and we capture any exceptions that happen through the regular `logging.error()` calls. It is not as detailed as the actual traceback, but it is at least visible through the logs. Would that be enough in your opinion? Otherwise, we can log more when an error occurs as well.

src/zenml/logging/step_logging.py
schustmi: Yes that makes sense to me, the root logger should dictate the behaviour 👍

schustmi:
> > One more thought I had: Now that we're not wrapping `stderr` anymore, exceptions probably won't show up in the logs anymore, correct?
> 
> Yes and no. We usually wrap the step and pipeline executions in try-except clauses and we capture any exceptions that happen through the regular `logging.error()` calls. It is not as detailed as the actual traceback, but it is at least visible through the logs. Would that be enough in your opinion? Otherwise, we can log more when an error occurs as well.

All good :) I already created #3838 which tracks this more explicitly

src/zenml/constants.py
```diff
@@ -218,6 +220,16 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
         ENV_ZENML_LOGGING_VERBOSITY, default="INFO"
     ).upper()
 
+
+if IS_DEBUG_ENV:
```
schustmi: Can we move this in the same `if` as the regular logging verbosity above?

src/zenml/logging/step_logging.py
```diff
@@ -53,15 +53,117 @@
...
+        # Convert print arguments to message
+        message = " ".join(str(arg) for arg in args)
+
+        # Determine if this should go to stderr or stdout based on file argument
+        file_arg = kwargs.get("file", sys.stdout)
+
+        # Call active handlers first (for storage)
+        if message.strip():
+            if file_arg == sys.stderr:
+                handlers = logging_handlers.get()
```
schustmi: The same here and in the else

src/zenml/logging/step_logging.py
```diff
@@ -473,28 +572,56 @@ def __init__(
...
+        # Set the step names context variable
+        step_names_disabled = handle_bool_env_var(
+            ENV_ZENML_DISABLE_STEP_NAMES_IN_LOGS, default=False
+        )
 
-        setattr(sys.stderr, "write", self._wrap_write(self.stderr_write))
-        setattr(sys.stderr, "flush", self._wrap_flush(self.stderr_flush))
+        if step_names_disabled or not self.prepend_step_name:
+            # Step names are disabled through the env or they are disabled in the config
+            step_names_in_console.set(False)
```
schustmi, line 792: What if two get registered, the first one has prepend_step_name, the second one doesn't. The second one would incorrectly reset the value from the first one, no?

src/zenml/logging/step_logging.py
```diff
@@ -529,70 +670,8 @@ def __exit__(
...
-        """
-
-        def wrapped_flush(*args: Any, **kwargs: Any) -> Any:
-            output = method(*args, **kwargs)
-            self.storage.save_to_file()
-            return output
-
-        return wrapped_flush
+        # Reset the step names context to default
+        step_names_in_console.set(False)
```
schustmi: What if another context is registered that still wants to have step names in the console?

schustmi (changes requested):
On a more general note: Now that we're not wrapping `sys.stdout/stderr` anymore, any custom logging framework that doesn't use `logging` under the hood will not be captured, right? Have you thought about this/implications?

src/zenml/constants.py
bcdurak: Done.

src/zenml/logging/step_logging.py
bcdurak: Fixed.

src/zenml/logging/step_logging.py
bcdurak: Theoretically, yes. However, I don't think this poses a problem at this stage. For this to happen, you would have to init two contexts. The only instances where we init these contexts are: `pipeline._run()`, `step_runner.run()`, and `step_launcher.launch()`. The pipeline is the outer one, and this parameter is always set to `False`. As you suggested in the other thread, we are not utilizing two separate `run` or `launch` calls.

src/zenml/logging/step_logging.py
bcdurak: Since this is also related to the thread above, I am closing this for now.

bcdurak:
@schustmi Since we weren't dealing with the `stderr` anymore, I didn't want to rely on any error logging messages and I have added [this](https://github.com/zenml-io/zenml/pull/3812/commits/4cbd79e124977fb57b510df1de8e1fab3908b322) now. Take a look, let me know what you think.

src/zenml/logging/step_logging.py
schustmi: Yes I agree that this will not happen, but I still think this is just bad/wrong code that could easily be fixed, and who knows where else this context will be used (maybe we allow users to create their own context to log into separate files for specific things?)

src/zenml/logging/step_logging.py
```diff
@@ -512,15 +646,44 @@ def __exit__(
             exc_val: The instance of the exception
             exc_tb: The traceback of the exception
 
-        Restores the `write` method of both stderr and stdout.
+        Removes the handler from loggers and context variables.
         """
-        self.storage.save_to_file(force=True)
+        # Write about the exception if it exists
```
schustmi:
```suggestion
```

src/zenml/logging/step_logging.py
```diff
@@ -512,15 +646,44 @@ def __exit__(
...
+        if exc_type is not None:
+            # Write the exception and its traceback to the logs
+            self.artifact_store_handler.emit(
+                logging.LogRecord(
+                    name="exception",
+                    level=logging.ERROR,
+                    pathname="",
+                    lineno=0,
+                    msg="Exception occurred during execution.",
```
schustmi: during what execution?

src/zenml/logging/step_logging.py
```diff
@@ -53,15 +53,124 @@
...
+            timestamp = utc_now().strftime("%Y-%m-%d %H:%M:%S")
+            message = (
+                f"[{timestamp} UTC] [{record.levelname}] {record.getMessage()}"
+            )
+
+            # Include exception information if present
+            if record.exc_info:
+                import traceback
+
+                exception_text = "".join(
```
schustmi: Have you checked what the end result of this will be? If it's the traceback, shouldn't it be separated by newlines instead?

src/zenml/logging/step_logging.py
```diff
@@ -512,15 +646,44 @@ def __exit__(
...
         """
-        self.storage.save_to_file(force=True)
+        # Write about the exception if it exists
+        if exc_type is not None:
+            # Write the exception and its traceback to the logs
+            self.artifact_store_handler.emit(
+                logging.LogRecord(
+                    name="exception",
+                    level=logging.ERROR,
+                    pathname="",
```
schustmi, line 821: I guess all these parameters (lineno, pathname) we ignore anyway in our artifact store handler, which is why we don't care about figuring them out here?

src/zenml/logging/step_logging.py
bcdurak: Removed.

src/zenml/logging/step_logging.py
bcdurak: Yes. The `format_exception` already returns a list of strings that end with `\n`. This is how it currently looks:

Screenshot 2025-07-29 at 17 12 37

src/zenml/logging/step_logging.py
bcdurak: At this point, it's either the pipeline or step execution, but I wanted to keep it vague. The `exc_info` will explain the details anyhow. You can see how it plays out [here](https://github.com/zenml-io/zenml/pull/3812#discussion_r2240276337).

src/zenml/logging/step_logging.py
bcdurak: Yes.

src/zenml/logging/step_logging.py
bcdurak: I have made some adjustments to this. Every time we `__enter__` the context, I save the original state of the parameter and restore it as we `__exit__` it.

src/zenml/logging/step_logging.py
```diff
@@ -635,70 +695,47 @@ def __exit__(
...
+                    lineno=0,
+                    msg="An exception has occurred.",
+                    args=(),
+                    exc_info=(exc_type, exc_val, exc_tb) if exc_val else None,
+                )
+            )
 
-        # Step 2: Shutdown thread (it will automatically drain queue and merge files)
+        # Remove handler from root logger and restore original level
+        if self.artifact_store_handler:
```
schustmi: This is always True?

src/zenml/logging/step_logging.py
```diff
@@ -635,70 +695,47 @@ def __exit__(
...
+            root_logger = logging.getLogger()
+            # Check if handler is still in the root logger before removing
+            if self.artifact_store_handler in root_logger.handlers:
+                root_logger.removeHandler(self.artifact_store_handler)
+            # Restore original root logger level
+            if hasattr(self, "original_root_level"):
```
schustmi: This always exists, and also needs to `hasattr` check

src/zenml/logging/step_logging.py
```diff
@@ -635,70 +695,47 @@ def __exit__(
...
+                root_logger.setLevel(self.original_root_level)
+
+        # Remove handler from context variables
+        handlers = logging_handlers.get().copy()
+        if self.artifact_store_handler in handlers:
+            handlers.remove(self.artifact_store_handler)
+        logging_handlers.set(handlers)
```
schustmi: This is a potential cause for race conditions. Multitple threads could be removing their own handler in parallel, which would cause some handler to be left over even if all should have been removed.

src/zenml/logger.py
```diff
@@ -193,9 +340,27 @@ def init_logging() -> None:
...
+        console_handler = logging.StreamHandler(sys.stdout)
+        console_handler.setFormatter(get_formatter())
+        # Set console handler level explicitly to console verbosity
+        console_handler.setLevel(get_logging_level().value)
+        root_logger.addHandler(console_handler)
+
+    # Initialize global print wrapping
+    try:
+        setup_global_print_wrapping()
+    except ImportError:
```
schustmi: Why would there be an import error here?

schustmi (changes requested):

src/zenml/logging/step_logging.py
bcdurak: Removing the if.

src/zenml/logging/step_logging.py
bcdurak: I have implemented a `ContextVarList` that protects it against race conditions while `append`ing and `remove`ing with a `lock`.

src/zenml/logger.py
bcdurak: Removed the try-except.

src/zenml/logging/step_logging.py
bcdurak: Refactored the check.

src/zenml/utils/context_utils.py
```diff
@@ -0,0 +1,56 @@
...
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Context variable utilities."""
+
+import threading
+from contextvars import ContextVar
+from typing import Any, List, Optional
+
+
+class ContextVarList:
```
schustmi: This could be a generic class with the list item type, would remove the need for all the `Any` and enable type checking.

src/zenml/utils/context_utils.py
```diff
@@ -0,0 +1,56 @@
...
+            if item not in current_list:
+                new_list = current_list + [item]
+                self._context_var.set(new_list)
+
+    def remove(self, item: Any) -> None:
+        """Thread-safely remove an item from the list."""
+        with self._lock:
+            current_list = self.get()
+            if item in current_list:
+                new_list = [x for x in current_list if x != item]
```
schustmi: This will do an `x is item` because the ArtifactStoreHandler doesn't implement any comparison logic. Probably what you intended, but could be clearer by just using `is` directly?

src/zenml/utils/context_utils.py
bcdurak: Added the generic type.

src/zenml/utils/context_utils.py
bcdurak: Fixed.

PR #3838, opened 2025-07-14
Collect step run exception info

schustmi:
## Describe changes

This PR implements tracking of exceptions that happen during the step execution. In addition to storing the traceback information, we also try to detect which line in the user step code was the reason that caused this exception, so we can highlight this line (and show the exception inline) in the step code in the dashboard.

## Pre-requisites

## Types of changes


src/zenml/orchestrators/publish_utils.py
```diff
@@ -59,13 +65,15 @@ def publish_step_run_status_update(
     step_run_id: "UUID",
     status: "ExecutionStatus",
     end_time: Optional[datetime] = None,
+    exception_info: Optional[ExceptionInfo] = None,
```
bcdurak, line 68: Is there any possibility that we would have exception info for a successful step run?

src/zenml/orchestrators/publish_utils.py
```diff
@@ -102,6 +111,7 @@ def publish_failed_step_run(step_run_id: "UUID") -> "StepRunResponse":
         step_run_id=step_run_id,
         status=ExecutionStatus.FAILED,
         end_time=utc_now(),
+        exception_info=step_exception_info.get(),
```
bcdurak, line 114: In `publish_failed_step_run`, we fetch it from the context variable, but in `publish_successful_step_run`, we had it as an input. Any reason why they are different?

src/zenml/orchestrators/step_runner.py
```diff
@@ -193,6 +200,24 @@ def run(
                 )
             except BaseException as step_exception:  # noqa: E722
                 step_failed = True
+
+                exception_info = exception_utils.collect_exception_information(
```
bcdurak, line 204: Right now, this covers the `step_instance.call_entrypoint(...)` call. Maybe we can keep parallel to the scope of the `logs_context`. WDYT?

src/zenml/orchestrators/publish_utils.py
bcdurak: Just checked this, couldn't find an instance where this input was used.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -212,6 +213,7 @@ def launch(self) -> None:
         Raises:
             RunStoppedException: If the pipeline run is stopped by the user.
         """
+        publish_utils.step_exception_info.set(None)
```
bcdurak, line 217: Is there any way this `launch`method is called by two parallel running processes? As we discussed yesterday, this might create a situation where they overwrite each other.

bcdurak: Reminds me of [this](https://github.com/zenml-io/zenml/pull/3812#discussion_r2215597418) as well.

bcdurak (commented):
I have two general questions as well. 

- What do you think about adding this to the metadata of the step run? I am not sure if we should add a DB column just for the exception info.
- Would we need to add something similar for the pipeline runs as well?

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -101,6 +102,14 @@ class StepRunSchema(NamedSchema, RunMetadataInterface, table=True):
             nullable=True,
         )
     )
+    exception_info: Optional[str] = Field(
```
bcdurak, line 105: I am not even sure if we would ever reach 16MBs of traceback, but if we ever do, this might lead to a DB error. We can perhaps do some form of cropping to avoid that.

src/zenml/orchestrators/publish_utils.py
schustmi: A successful step run should never have any exception info, and in that function it is also not an input.

src/zenml/orchestrators/step_launcher.py
schustmi: No, otherwise we would have bigger issues with the step context and everything around that. But even if it was, that's the whole reason this is a `ContextVar`, it would work even with multiple threads.

src/zenml/zen_stores/schemas/step_run_schemas.py
schustmi: I'm sure that will never happen, but I'll add a limit to it :)

src/zenml/orchestrators/publish_utils.py
schustmi: Two things:
- This is not a function for successful step runs, it's a status update
- Of course its used, you even commented below on the function that uses it.

src/zenml/orchestrators/step_runner.py
schustmi: I want to actually run this right after the user code, as otherwise we might include some stuff from the failure hook etc which is not really the actual cause for the error that caused the step to fail.

bcdurak (approved):

PR #3845, opened 2025-07-16
Changes to the fetch logs endpoints

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ zenml = "zenml.cli.cli:cli"
 alembic = { version = ">=1.8.1,<=1.15.2" }
 bcrypt = { version = "4.0.1" }
 click = "^8.0.1,<=8.2.1"
-cloudpickle = ">=2.0.0,<3"
+cloudpickle = ">=2.0.0"
 distro = "^1.6.0"
 docker = "~7.1.0"
 gitpython = "^3.1.18"
--- a/src/zenml/pipelines/build_utils.py
+++ b/src/zenml/pipelines/build_utils.py
@@ -710,6 +710,9 @@ def should_upload_code(
         # already be included.
         return False
 
+    if build.contains_code:
+        return False
+
     for step in deployment.step_configurations.values():
         docker_settings = step.config.docker_settings
 


src/zenml/logging/step_logging.py
```diff
@@ -86,13 +125,96 @@ def emit(self, record: logging.LogRecord) -> None:
...
+                    filename=record.filename,
+                    lineno=record.lineno,
+                )
+                json_line = log_record.model_dump_json(exclude_none=True)
+                self.storage.write(json_line)
+            else:
+                # Message is too large, split into chunks and emit each one
+                chunks = self._split_to_chunks(message)
+                for i, chunk in enumerate(chunks):
+                    chunk_message = f"{chunk} (chunk {i + 1}/{len(chunks)})"
```
schustmi: Why is this necessary? And if so, shouldn't the chunk be stored in the `LogEntry` somewhere instead of the message?

src/zenml/logging/step_logging.py
```diff
@@ -152,121 +311,164 @@ def prepare_logs_uri(
...
+def fetch_log_records(
     zen_store: "BaseZenStore",
     artifact_store_id: Union[str, UUID],
     logs_uri: str,
     offset: int = 0,
-    length: int = 1024 * 1024 * 16,  # Default to 16MiB of data
-    strip_timestamp: bool = False,
-) -> str:
-    """Fetches the logs from the artifact store.
+    count: int = MAX_LOG_ENTRIES,  # Number of entries to return
```
schustmi: Why is this not configurable on the endpoint?

bcdurak: As far as I can see, it is already configurable on the endpoint. Or am I missing something?

src/zenml/logging/step_logging.py
bcdurak: Previously, we had a size limit to the return of the endpoint. Right now, we have a count limit. In order to have a controllable size as well, I wanted to split the messages in chunks. 

Having them interpreted as separate `LogEntry`s also made it easier to handle them on the background thread. 

I am open to change it though.

src/zenml/logging/step_logging.py
schustmi: You are absolutely right, I just saw the constant and somehow missed that's it an endpoint arg, sorry :D

src/zenml/logging/step_logging.py
schustmi: While the chunking might be useful in general, I think the current solution isn't ideal. What if the log message was some structured string, and now parsing it fails because of those chunk inserts that the user never asked for? I think we shouldn't modify the log messages and appending some `chunk 1/X` at the end, instead store the `chunk` in the log entry object. For short messages, there will only ever be a single chunk, while for longer messages there might be multiple. I think that would also require an `id` for the log entry, so we can match those chunks to the same message. The frontend should worry about displaying this information (e.g. display `...` or add a `fetch remaining message` options), but nothing that the backend cares about IMO

bcdurak: Added `chunk_index`, `chunk_count_in_entry` and `entry_id` to the `LogEntry` object. These are only relevant if the message is larger than the `MAX_MESSAGE_SIZE`.

src/zenml/logging/step_logging.py
```diff
@@ -62,6 +68,52 @@
...
+    )
+    chunk_index: Optional[int] = Field(
+        default=None,
+        description="The index of the chunk in the log entry",
+    )
+    chunk_count_in_entry: Optional[int] = Field(
+        default=None,
+        description="The total number of chunks in the log entry",
+    )
+    entry_id: Optional[UUID] = Field(
```
schustmi: I think I would just make this non-optional and call it `id`, that way it aligns with all our other models?

src/zenml/logging/step_logging.py
```diff
@@ -62,6 +68,52 @@
...
+        description="The name of the file that generated this log entry",
+    )
+    lineno: Optional[int] = Field(
+        default=None, description="The fileno that generated this log entry"
+    )
+    chunk_index: Optional[int] = Field(
+        default=None,
+        description="The index of the chunk in the log entry",
+    )
+    chunk_count_in_entry: Optional[int] = Field(
```
schustmi: Somehow seems a little verbose, what about just `chunk_count` or `total_chunks`?

src/zenml/logging/step_logging.py
```diff
@@ -152,121 +326,164 @@ def prepare_logs_uri(
...
+        # For any other errors during streaming, fall back to empty result
+        logger.warning(f"Error streaming logs from {logs_uri}: {e}")
+        return []
+
+
+def _stream_logs_line_by_line(
+    zen_store: "BaseZenStore",
+    artifact_store_id: Union[str, UUID],
+    logs_uri: str,
+) -> Any:
```
schustmi: Needs a type annotation

src/zenml/logging/step_logging.py
```diff
@@ -152,121 +326,164 @@ def prepare_logs_uri(
...
-                        break
 
-                if not ret:
-                    raise DoesNotExistException(
-                        f"Folder '{logs_uri}' is empty in artifact store "
-                        f"'{artifact_store.name}'."
-                    )
-                return "".join(ret)
+            # Sort files to read them in order
+            files.sort()
```
schustmi, line 397: What is this sorting by?

src/zenml/logging/step_logging.py
```diff
@@ -86,13 +138,98 @@ def emit(self, record: logging.LogRecord) -> None:
...
+        start = 0
+        for i in range(chunk_count):
+            # Calculate the end position for this chunk
+            end = min(start + MAX_MESSAGE_SIZE, len(message_bytes))
+
+            # Extract the chunk and decode it
+            chunk_bytes = message_bytes[start:end]
+
+            # Handle potential UTF-8 boundary issues
+            try:
```
schustmi: I think this would be much simpler without this `try`, the while loop should already handle it?

src/zenml/logging/step_logging.py
```diff
@@ -86,13 +138,98 @@ def emit(self, record: logging.LogRecord) -> None:
...
+                # If we cut in the middle of a UTF-8 character, back off until we find a valid boundary
+                while len(chunk_bytes) > 0:
+                    try:
+                        chunk_text = chunk_bytes.decode("utf-8")
+                        break
+                    except UnicodeDecodeError:
+                        chunk_bytes = chunk_bytes[:-1]
+                        end -= 1
+                else:
+                    # If we can't decode anything, skip this chunk
```
schustmi: If I understand correctly, you're not skipping anything here right? At this point, end is already set to `start` (due to all the substractions), so the same chunk will simply be retried and skipped again and again?

src/zenml/logging/step_logging.py
bcdurak: Sure, I will change it. Would you make the `chuck_index` and `total_chunks` non-optional as well?

src/zenml/logging/step_logging.py
bcdurak: Renamed it.

src/zenml/logging/step_logging.py
bcdurak: Added `Iterator[str]`

src/zenml/logging/step_logging.py
bcdurak: We used the same method previously. Since, in the case of immutable filesystems, the filenames are timestamped, we sort them by name before reading.

src/zenml/logging/step_logging.py
schustmi: Yeah I'd say so, seems to be the simplest and doesn't hurt us right?

bcdurak: Already done.

src/zenml/logging/step_logging.py
bcdurak: Yes, there were some additional problems here as well. I refactored the while thing.

src/zenml/logging/step_logging.py
```diff
@@ -62,6 +68,52 @@
...
+    )
+    chunk_index: int = Field(
+        default=0,
+        description="The index of the chunk in the log entry",
+    )
+    total_chunks: int = Field(
+        default=1,
+        description="The total number of chunks in the log entry",
+    )
+    entry_id: UUID = Field(
```
schustmi:
```suggestion
    id: UUID = Field(
```

src/zenml/logging/step_logging.py
```diff
@@ -86,13 +138,98 @@ def emit(self, record: logging.LogRecord) -> None:
...
+                    module=record.module,
+                    filename=record.filename,
+                    lineno=record.lineno,
+                )
+                json_line = log_record.model_dump_json(exclude_none=True)
+                self.storage.write(json_line)
+            else:
+                # Message is too large, split into chunks and emit each one
+                chunks = self._split_to_chunks(message)
+                chunk_id = uuid4()
```
schustmi:
```suggestion
                entry_id = uuid4()
```

src/zenml/logging/step_logging.py
```diff
@@ -86,13 +138,98 @@ def emit(self, record: logging.LogRecord) -> None:
...
+        start = 0
+
+        while start < len(message_bytes):
+            # Calculate the end position for this chunk
+            end = min(start + MAX_MESSAGE_SIZE, len(message_bytes))
+
+            # If we're not at the end of the message, find a safe UTF-8 boundary
+            if end < len(message_bytes):
+                # Back up to find a valid UTF-8 character boundary
+                while end > start and (message_bytes[end] & 0x80) != 0:
```
schustmi: What is this `& 0x80 != 0` doing exactly?

src/zenml/logging/step_logging.py
```diff
@@ -86,13 +138,98 @@ def emit(self, record: logging.LogRecord) -> None:
...
+                    # Check if this byte is the start of a UTF-8 character
+                    if (message_bytes[end] & 0xC0) != 0x80:
+                        break
+                    end -= 1
+
+            # Extract and decode the chunk
+            chunk_bytes = message_bytes[start:end]
+            try:
+                chunk_text = chunk_bytes.decode("utf-8")
```
schustmi: I don't think this handles the case where the entire chunk is not valid, right? In this case, end will be equal to start, and `chunk_text` will be an empty string. And then, it will stay in this loop forever?

src/zenml/logging/step_logging.py
bcdurak: It checks whether this byte is a part of a non-English character (multi-byte, starting with bit 1) or not. In other words, 0x80 checks if the first bit is 1 and if it is, it means that the point where we split the chunk belongs to a multi-byte character.

src/zenml/logging/step_logging.py
bcdurak: Changed the naming.

src/zenml/logging/step_logging.py
bcdurak: Changed it.

src/zenml/logging/step_logging.py
bcdurak: Yes, I have changed it to `start + 1`, so the `chunk_bytes` will have at least one byte to decode.

If the chunk is not valid, this would then lead to a decode error and we would decode the whole by replacing all the errors.

```python
            except UnicodeDecodeError:
                # Advance to where the next iteration would normally start
                end = min(start + MAX_MESSAGE_SIZE, len(message_bytes))
                
                # This should rarely happen with proper boundary detection
                # but handle it gracefully by skipping to the next chunk boundary
                chunks.append(
                    message_bytes[start:end].decode("utf-8", errors="replace")
                )
```

src/zenml/logging/step_logging.py
```diff
@@ -86,13 +138,99 @@ def emit(self, record: logging.LogRecord) -> None:
...
+        start = 0
+
+        while start < len(message_bytes):
+            # Calculate the end position for this chunk
+            end = min(start + MAX_MESSAGE_SIZE, len(message_bytes))
+
+            # If we're not at the end of the message, find a safe UTF-8 boundary
+            if end < len(message_bytes):
+                # Back up to find a valid UTF-8 character boundary
+                while end > start + 1 and (message_bytes[end] & 0x80) != 0:
```
schustmi: Is there any difference to doing this compared to the previous logic that tried to decode and caught the exception? To me, that was understandable and made sense, while this implementation is hard to understand and verify that is works

src/zenml/logging/step_logging.py
bcdurak: Now that I think about this again, I have changed the logic here slightly. Previously, we were first checking whether it is a multi-byte character or not and then check whether it was the continuation byte. Now, I just directly check if it is the continuation byte as the first check is redundant.

src/zenml/logging/step_logging.py
bcdurak: Not sure how computationally expensive it is to try to decode the whole string over and over again, but I do agree this was more complicated. I have changed it again. LMK what you think.

schustmi: Could be yes, but this will anyway not be a real concern as the amount of log messages >10KB will be 0 in 99.99% of cases I hope

bcdurak:
@schustmi I am fine with waiting for the frontend as long as I got the ✅ 😅

bcdurak:
@schustmi I have changed a few things in this PR after your last review:

- There was a redundant formatting logic. Removed it from the artifact store handler.
- There was an error with the interaction between Enum input values and FastAPI in the endpoints. Thus, the filter parameter (to decide the minimum log level) for the endpoint is an int now. (This behaviour is also consistent with how we handle other Enum parameter like status in similar endpoints.)
- I have introduced pagination to the logs. Now, we don't load and send the entire `.log` file (or up to 16mb or so) but we paginate. This was a bit more complicated as we need count how many entries there are. Especially with filters or searches, you need to go through the entirety of the logs to figure out how many matching entries there are. I investigated if this would create any hits in terms of performance, however, the results were quite promising. I ran our stress test pipeline on the classic Kubernetes stack and it generated 200+ MB of logs (and roughly 70000 entries). Loading and paginating them properly takes roughly only 4-5 secs. There is one slight hiccup however, our original `Page` object is tied to `BaseResponse`s for some reason, so I have used a minimal model with the same fields to make it easier for Julian.
- Finally, I have implemented an additional download endpoint. Since we implemented pagination for the previous endpoint, it wasn't exactly suited for the download functionality. Now, we have a separate endpoint that yields a streaming result. I believe you have implemented a similar endpoint with a `FileResponse` before. This is sort of a similar endpoint with a streaming output. One question however: You implemented a token parameter for yours, and I am not sure if I need a similar approach with my endpoint.

Let me know what you think.

schustmi:
> * There was an error with the interaction between Enum input values and FastAPI in the endpoints. Thus, the filter parameter (to decide the minimum log level) for the endpoint is an int now. (This behaviour is also consistent with how we handle other Enum parameter like status in similar endpoints.)

Can you point me towards one of these endpoints where the status (or other enums) are treated as ints? As all our enums are actually using strings as their enum values, using those strings seems more intuitive to me.

> * I have introduced pagination to the logs. Now, we don't load and send the entire `.log` file (or up to 16mb or so) but we paginate. This was a bit more complicated as we need count how many entries there are. Especially with filters or searches, you need to go through the entirety of the logs to figure out how many matching entries there are. I investigated if this would create any hits in terms of performance, however, the results were quite promising. I ran our stress test pipeline on the classic Kubernetes stack and it generated 200+ MB of logs (and roughly 70000 entries). Loading and paginating them properly takes roughly only 4-5 secs. There is one slight hiccup however, our original `Page` object is tied to `BaseResponse`s for some reason, so I have used a minimal model with the same fields to make it easier for Julian.
- *Only 4-5secs*: this seems like quite a lot to me? 🤔 Once I try to search for something different, it will take another 4-5 seconds before it displays anything, because it has to find all the occurences to figure out the pages, right? In other tools, I've seen that they at least display the initial page/results quickly, and then in the background fetch the total pages etc async. Maybe that is something we also want to consider here.
- Why don't we simply make the `Page` object work with any `BaseModel` instead then? I think there is no limitation that would prevent you from simply changing one type annotation and it would work, instead of copying the class?

> * Finally, I have implemented an additional download endpoint. Since we implemented pagination for the previous endpoint, it wasn't exactly suited for the download functionality. Now, we have a separate endpoint that yields a streaming result. I believe you have implemented a similar endpoint with a `FileResponse` before. This is sort of a similar endpoint with a streaming output. One question however: You implemented a token parameter for yours, and I am not sure if I need a similar approach with my endpoint.

Yes you will probably need the same token approach, due to the CSRF issues that @Cahllagerfeld mentioned in the daily today. The frontend will open the download in a new tab to use the browser native download functionality, but at the point our authentication does not work anymore because we're not on the correct domain anymore. Which is why the download endpoint itself needs to be unauthenticated, and there is a second authenticated endpoint which generates a temporary token that can be used to access the logs endpoint.

bcdurak:
> Can you point me towards one of these endpoints where the status (or other enums) are treated as ints? As all our enums are actually using strings as their enum values, using those strings seems more intuitive to me.

Oh, maybe I misspoke a bit here. What I meant was, they are treated as strings. Because if I do `level: LoggingLevels` as an input typehint on the endpoint, the swagger docs show the possible values (from the enum) correctly, however the input is not handled properly.

> Only 4-5secs: this seems like quite a lot to me? 🤔 Once I try to search for something different, it will take another 4-5 seconds before it displays anything, because it has to find all the occurences to figure out the pages, right? In other tools, I've seen that they at least display the initial page/results quickly, and then in the background fetch the total pages etc async. Maybe that is something we also want to consider here.

I think the main bulk wasn't even the processing. When I tried it locally with the same file, it was quite fast, however, if you do it with a remote artifact store with a connector it takes some time to open the file. I have never implemented an endpoint that returns a part of the result first and then fetches the rest but we can definitely look into it. If it is alright, I would investigate this on a separate PR though.

> Why don't we simply make the Page object work with any BaseModel instead then? I think there is no limitation that would prevent you from simply changing one type annotation and it would work, instead of copying the class?

I had the same idea. I took a look at our current `Page` object and couldn't see anything that would tie to the base responses immediately. I will try to change this if you also dont see any red flags.

@schustmi

src/zenml/zen_server/auth.py
```diff
@@ -1045,56 +1046,85 @@ def generate_access_token(
...
     )
 
 
-def verify_artifact_download_token(
-    token: str, artifact_version_id: UUID
+def verify_download_token(
+    token: str,
+    download_type: DownloadType,
+    resource_id: UUID,
+    extra_data: Optional[Dict[str, Any]] = None,
```
schustmi: The name seems a little off IMO, maybe `extra_claims`?

src/zenml/enums.py
```diff
@@ -53,6 +53,14 @@ class ArtifactSaveType(StrEnum):
     )
 
 
+class DownloadType(StrEnum):
+    """All possible types of downloadable content."""
+
+    ARTIFACT = "artifact"
```
schustmi: This is an artifact version right?

schustmi (changes requested):

src/zenml/zen_server/auth.py
bcdurak: Renamed it.

src/zenml/enums.py
bcdurak: Makes sense. Renamed it.

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -244,26 +261,111 @@ def get_step_status(
...
+    """
+    step = zen_store().get_run_step(step_id, hydrate=True)
+    pipeline_run = zen_store().get_run(step.pipeline_run_id)
+    verify_permission_for_model(pipeline_run, action=Action.READ)
+
+    store = zen_store()
+
+    # Verify that logs are available for this step
+    if step.logs is None:
+        raise HTTPException(
```
schustmi: raise `KeyError`

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -244,26 +261,111 @@ def get_step_status(
...
+            zen_store=store,
+            artifact_store_id=step.logs.artifact_store_id,
+            logs_uri=step.logs.uri,
+            page=page,
+            count=count,
+            level=level,
+            search=search,
+            from_position=parsed_from_position,
+        )
+    except ValueError as e:
```
schustmi: No need, this happens autoamtically.

schustmi: In general, this is for all endpoints. Don't raise HTTP exceptions, just raise regular exceptions as we already convert them.

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -244,26 +261,111 @@ def get_step_status(
...
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+@async_fastapi_endpoint_wrapper
+def get_step_logs_info(
+    step_id: UUID,
+    page_size: int = DEFAULT_PAGE_SIZE,
+    level: int = LoggingLevels.INFO.value,
+    search: Optional[str] = None,
+    _: AuthContext = Security(authorize),
+) -> LogInfo:
```
schustmi: Sometimes its LogInfo, sometimes page information

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -244,26 +261,111 @@ def get_step_status(
...
+    from_position: Optional[int] = None,
     _: AuthContext = Security(authorize),
-) -> str:
-    """Get the logs of a specific step.
+) -> List[LogEntry]:
+    """Get log entries for efficient pagination.
+
+    This endpoint supports optimized navigation modes:
+
+    - page=1: Returns the latest entries (first page)
```
schustmi: First page = latest entries? I thought first page are the first entries?

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -244,26 +261,111 @@ def get_step_status(
...
+    - page=N without from_position: Requires using page index endpoint first
```
schustmi: So this will fail then?

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -244,26 +261,111 @@ def get_step_status(
...
+    - with from_position: Direct navigation to the page starting from position
```
schustmi: Do we need anything other than this? page=1 is simply `from_position=0`, no?

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -429,20 +450,30 @@ def stop_run(
 def run_logs(
     run_id: UUID,
     source: str,
-    offset: int = 0,
-    length: int = 1024 * 1024 * 16,  # Default to 16MiB of data
+    page: int = 1,
+    count: int = DEFAULT_PAGE_SIZE,
+    level: int = LoggingLevels.INFO.value,
+    search: Optional[str] = None,
+    seek_file_index: Optional[int] = None,
```
schustmi: completely different param names than the other endpoint..

src/zenml/logging/step_logging.py
```diff
@@ -62,6 +76,99 @@
...
+
+class PageIndexResponse(BaseModel):
+    """Response containing pagination index for efficient navigation."""
+
+    total_entries: int = Field(description="Total number of log entries")
+    total_pages: int = Field(description="Total number of pages")
+    page_size: int = Field(description="Number of entries per page")
+    page_positions: List[PageIndexEntry] = Field(
+        description="Start positions for each page"
+    )
```
schustmi: Where are those classes used?

schustmi (changes requested):

src/zenml/zen_server/routers/steps_endpoints.py
bcdurak: Fixed.

src/zenml/zen_server/routers/steps_endpoints.py
bcdurak: Done.

src/zenml/zen_server/routers/steps_endpoints.py
bcdurak: It is. Since I changed the approach, I forgot to adjust the docstring.

src/zenml/zen_server/routers/steps_endpoints.py
bcdurak: You are right. I wanted to rename this, but forgot to do it. Done now.

src/zenml/zen_server/routers/steps_endpoints.py
bcdurak: Fixed the docstring.

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: Fixed.

src/zenml/logging/step_logging.py
bcdurak: Remnants of the old approach. Removed now.

src/zenml/zen_server/routers/steps_endpoints.py
bcdurak: You are right. In most cases, page 1 is `from_position=0`. However, when it comes to the runs and the corresponding runner logs, we do not support from_position just yet. That's why I wanted to support both over the API.

PR #3851, opened 2025-07-17
Fail on certain container waiting reasons during job monitoring

schustmi:
## Describe changes
When a job starts its pods, the containers sometimes get stuck in a waiting state with reasons like
- not being able to fetch the image
- the image name being invalid
- some configuration error, e.g. a missing secret or configmap

When this happens, the job would be stuck indefinitely, as this doesn't cause the pod to fail (it just stays in pending state).
This PR monitors the container states of pods created by a job, and fails the job if any container is in a (configurable) failed state.

## Pre-requisites

## Types of changes


src/zenml/integrations/azure/service_connectors/azure_service_connector.py
```diff
@@ -1490,7 +1490,8 @@ def _list_aks_clusters(
                 aks_clusters = [
                     cluster
                     for cluster in aks_clusters
-                    if self._get_resource_group(cluster.id)
+                    if cluster.id
```
schustmi, line 1493: @stefannica I think there was a new release of some azure library that caused a few linting issues in this file. I tried to solve them, but probably best for you to have a look whether that's actually the correct implementation.

stefannica (approved):

PR #3855, opened 2025-07-18
Solving the race conditions for logging

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/logging/step_logging.py
+++ b/src/zenml/logging/step_logging.py
@@ -18,6 +18,7 @@ import os
 import re
 import sys
 import time
+import threading
 from contextlib import nullcontext
 from contextvars import ContextVar
 from types import TracebackType
@@ -269,6 +270,9 @@ class PipelineLogsStorage:
         # Immutable filesystems state
         self.last_merge_time = time.time()
 
+        # Lock for shared state
+        self.lock = threading.Lock()
+
     def write(self, text: str) -> None:
         """Main write method.
 
@@ -278,14 +282,15 @@ class PipelineLogsStorage:
         if text == "\n":
             return
 
-        if not self.disabled:
-            # Add timestamp to the message when it's received
-            timestamp = utc_now().strftime("%Y-%m-%d %H:%M:%S")
-            formatted_message = (
-                f"[{timestamp} UTC] {remove_ansi_escape_codes(text)}"
-            )
-            self.buffer.append(formatted_message.rstrip())
-            self.save_to_file()
+        with self.lock:
+            if not self.disabled:
+                # Add timestamp to the message when it's received
+                timestamp = utc_now().strftime("%Y-%m-%d %H:%M:%S")
+                formatted_message = (
+                    f"[{timestamp} UTC] {remove_ansi_escape_codes(text)}"
+                )
+                self.buffer.append(formatted_message.rstrip())
+                self._save_to_file()
 
     @property
     def _is_write_needed(self) -> bool:
@@ -312,6 +317,15 @@ class PipelineLogsStorage:
 
     def save_to_file(self, force: bool = False) -> None:
         """Method to save the buffer to the given URI.
+
+        Args:
+            force: whether to force a save even if the write conditions not met.
+        """
+        with self.lock:
+            self._save_to_file(force)
+
+    def _save_to_file(self, force: bool = False) -> None:
+        """Internal method to save the buffer to the given URI (no locking).
 
         Args:
             force: whether to force a save even if the write conditions not met.
@@ -400,7 +414,7 @@ class PipelineLogsStorage:
             and time.time() - self.last_merge_time > self.merge_files_interval
         ):
             try:
-                self.merge_log_files()
+                self._merge_log_files()
             except (OSError, IOError) as e:
                 logger.error(f"Error while trying to roll up logs: {e}")
             finally:
@@ -414,12 +428,21 @@ class PipelineLogsStorage:
         Args:
             merge_all_files: whether to merge all files or only raw files
         """
+        with self.lock:
+            self._merge_log_files(merge_all_files)
+
+    def _merge_log_files(self, merge_all_files: bool = False) -> None:
+        """Internal method to merge log files (no locking).
+
+        Args:
+            merge_all_files: whether to merge all files or only raw files
+        """
         if self.artifact_store.config.IS_IMMUTABLE_FILESYSTEM:
             merged_file_suffix = "_merged"
             files_ = self.artifact_store.listdir(self.logs_uri)
             if not merge_all_files:
                 # already merged files will not be merged again
-                files_ = [f for f in files_ if merged_file_suffix not in f]
+                files_ = [f for f in files_ if merged_file_suffix not in str(f)]
             file_name_ = self._get_timestamped_filename(
                 suffix=merged_file_suffix
             )
--- a/src/zenml/logging/step_logging.py
+++ b/src/zenml/logging/step_logging.py
@@ -17,8 +17,8 @@ import logging
 import os
 import re
 import sys
-import time
 import threading
+import time
 from contextlib import nullcontext
 from contextvars import ContextVar
 from types import TracebackType
@@ -442,7 +442,9 @@ class PipelineLogsStorage:
             files_ = self.artifact_store.listdir(self.logs_uri)
             if not merge_all_files:
                 # already merged files will not be merged again
-                files_ = [f for f in files_ if merged_file_suffix not in str(f)]
+                files_ = [
+                    f for f in files_ if merged_file_suffix not in str(f)
+                ]
             file_name_ = self._get_timestamped_filename(
                 suffix=merged_file_suffix
             )


src/zenml/logging/step_logging.py
```diff
@@ -269,6 +270,9 @@ def __init__(
         # Immutable filesystems state
         self.last_merge_time = time.time()
 
+        # Lock for shared state
+        self.lock = threading.Lock()
```
schustmi: I think just to be save we should use an `RLock` here

bcdurak: Makes sense, switched it.

bcdurak:
@schustmi I am running the stress test now. Will merge if everything is in the green.

src/zenml/logging/step_logging.py
```diff
@@ -313,6 +320,15 @@ def _get_timestamped_filename(self, suffix: str = "") -> str:
     def save_to_file(self, force: bool = False) -> None:
         """Method to save the buffer to the given URI.
 
+        Args:
+            force: whether to force a save even if the write conditions not met.
+        """
+        with self.buffer_lock:
+            self._save_to_file(force)
```
schustmi: If you're calling `_save_to_file` here inside the buffer lock, you keep this (supposedly short-term) lock until the entire IO operation has finished, no?

schustmi (changes requested):
In general this is written in a way that makes it very hard to grasp what is happening, as there is so much locking happening in so many places.

Wouldn't it be much simpler like this? Obviously very simplified:
```python
def write(...):
  buffer_to_write = []
  
  with self.buffer_lock:
    self.buffer.append(...)
    if self.should_write():
      buffer_to_write = self.buffer.copy()
      self.buffer = []
      self.last_merge_time = time.time()
  
  if buffer_to_write:
    self._write_data()
```

And then have the `io_lock` in `_write_data`? Of course we would still need to make sure the threads enter the `io_lock` in the correct order, and this doesn't cover the merging yet. But I feel like it would be much clearer, wdyt?

src/zenml/logging/step_logging.py
bcdurak: Yeah I am aware. I have been trying to fix this but haven't pushed my changes yet.

bcdurak:
> In general this is written in a way that makes it very hard to grasp what is happening, as there is so much locking happening in so many places.
> 
> Wouldn't it be much simpler like this? Obviously very simplified:
> 
> ```python
> def write(...):
>   buffer_to_write = []
>   
>   with self.buffer_lock:
>     self.buffer.append(...)
>     if self.should_write():
>       buffer_to_write = self.buffer.copy()
>       self.buffer = []
>       self.last_merge_time = time.time()
>   
>   if buffer_to_write:
>     self._write_data(buffer_to_write)
> ```
> 
> And then have the `io_lock` in `_write_data`? Of course we would still need to make sure the threads enter the `io_lock` in the correct order, and this doesn't cover the merging yet. But I feel like it would be much clearer, wdyt?

Exactly what I have been struggling with. This way, it would have been much simpler but you pointed out the main problem. If there are two threads that finishes the buffer process (and they want to write the buffer to the file), it is possible for the for the order to get messed up. I am trying a solution where the process picks up the `io_lock` right before releasing the `buffer_lock` (if they need to merge or write). So far, it seems to work but performance-wise it wasn't that different from using a single lock.

While I was playing around with it, I have discovered other problems as well:
- We protect our system against any logs that happen during the `save_to_file` call but we do not currently protect it during the `write` call and the `merge` calls (I guess it is very unlikely for the `merge` to recursively call itself).
- Additionally, (and also more importantly) if we separate the locks, the `io_lock` would currently set the `self.disabled` to True (because we don't want to deal with any log messages coming from the artifact store implementations). However, while the IO operations are happening within the thread that acquired the `io_lock`, it would make the other threads unable to use the `write` functionality of the `storage` instance because it's disabled.

bcdurak:
@schustmi I have pushed the idea that I had in mind. The code is still in a bit of a dirty state, but feel free to take a look.

bcdurak:
@schustmi, I revamped the whole thing. Feel free to take a look.

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
```diff
@@ -272,7 +272,7 @@ def launch(
                 "ExperimentName": settings.experiment_name,
                 "TrialName": sanitized_training_job_name,
             }
-        info.force_write_logs()
+
```
schustmi: Can you explain why this isn't needed anymore?

src/zenml/logging/__init__.py
```diff
@@ -18,7 +18,7 @@
 STEP_LOGS_STORAGE_INTERVAL_SECONDS: int = 15
 
 # How many messages to buffer before uploading logs to the artifact store
-STEP_LOGS_STORAGE_MAX_MESSAGES: int = 100
+STEP_LOGS_STORAGE_MAX_MESSAGES: int = 1000
```
schustmi: In constants you define `1000` again as the default when reading the env variable, but somehow don't use this one?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -286,27 +285,10 @@ def launch(self) -> None:
...
-
-                        def _bypass() -> None:
-                            return None
-
-                        force_write_logs = _bypass
                     self._run_step(
                         pipeline_run=pipeline_run,
                         step_run=step_run,
-                        force_write_logs=force_write_logs,
+                        force_write_logs=lambda: None,
```
schustmi: If nothing is using this, we can simply remove it from that class?

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,225 @@ def __init__(
...
+
+        Args:
+            timeout: Maximum time to wait for thread shutdown.
+        """
+        if self.log_storage_thread and self.log_storage_thread.is_alive():
+            # First, prevent new items from being queued
+            self.shutdown_requested = True
+
+            # Then signal the worker to begin graceful shutdown
+            self.shutdown_event.set()
```
schustmi: Why can't this both be done with the same event?

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,225 @@ def __init__(
...
+
+            # Wait for thread to finish (it will drain the queue automatically)
+            self.log_storage_thread.join(timeout=timeout)
+            if self.log_storage_thread.is_alive():
+                # Thread didn't shutdown gracefully, but we can't do much more
+                # The non-daemon thread will still be waited for by Python
+                pass
```
schustmi: These lines do what exactly? Seems like nothing :D

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,225 @@ def __init__(
...
+            )
+            self.log_storage_thread.start()
+
+    def _log_storage_worker(self) -> None:
+        """Log storage thread worker that processes the log queue."""
+        while not self.shutdown_event.is_set():
+            all_buffers = []
+            try:
+                # Wait for first item in queue or timeout to check shutdown
+                try:
```
schustmi: Do we need to use the buffer_lock here to avoid any race conditions?

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,225 @@ def __init__(
...
+            except Exception as e:
+                self.thread_exception = e
+                logger.error("Error in log storage thread: %s", e)
+            finally:
+                # Always mark all queue tasks as done
+                for _ in all_buffers:
+                    self.log_queue.task_done()
+
+        # Shutdown requested - drain remaining queue items in batches
+        while True:
```
schustmi: This is exactly the same code as above..

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,225 @@ def __init__(
...
-            # Add timestamp to the message when it's received
-            timestamp = utc_now().strftime("%Y-%m-%d %H:%M:%S")
-            formatted_message = (
-                f"[{timestamp} UTC] {remove_ansi_escape_codes(text)}"
-            )
-            self.buffer.append(formatted_message.rstrip())
-            self.save_to_file()
+        # Check if log storage thread encountered an exception
+        if self.thread_exception:
+            return
```
schustmi: We should restart the thread here or something, no? Even in your current code, the thread had an exception that you caught, so its still running just fine.

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,225 @@ def __init__(
...
+                formatted_message = (
+                    f"[{timestamp} UTC] {remove_ansi_escape_codes(text)}"
+                )
+                self.buffer.append(formatted_message.rstrip())
+
+                # Check if the buffer needs to be queued for processing
+                buffer_to_queue = None
+                if self._is_write_needed:
+                    buffer_to_queue = self.buffer.copy()
+                    self.buffer = []
```
schustmi: What I still don't really get: Why do we need this buffer at all, and can't simply put the elements in the queue and let the log thread handle the logic?

schustmi (changes requested):

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
bcdurak: As far as I can see, this was initially implemented so that we can [dump all the logs from the client side (that launches the step) in the right order.](https://github.com/zenml-io/zenml/pull/2788). However, this approach still has some problems, especially for immutable artifact stores. For instance, both the environment that launches the step and the environment that executes it can initiate merging/clean up logic at the same time. Additionally, it was only adopted by a handful of step operators, leaving out the rest of the integrations. I thought, the ultimate solution here would be to separate the logs for different `source`s, just like we did with the pipeline response. However, I would rather fix this in a follow-up PR.

bcdurak: For the time being, I am going to implement a flush functionality and add this method back.

src/zenml/logging/__init__.py
bcdurak: This is only defined in the `logging/__init__.py`. In the constants, we have the `LOGS_STORAGE_MAX_QUEUE_SIZE` coming for the env variable. Maybe that created the confusion?

src/zenml/logging/step_logging.py
bcdurak: I don't think so. All the buffers here are already committed to the queue and will not change. Since there is a single thread consuming from the queue, I don't think we need a lock here. Maybe I am missing something?

src/zenml/orchestrators/step_launcher.py
bcdurak: Adding this back in for the time being. Explanation [here](https://github.com/zenml-io/zenml/pull/3855/files#r2235308173).

src/zenml/logging/step_logging.py
bcdurak: I refactored it.

src/zenml/logging/step_logging.py
bcdurak: Remnants of some prompts. Cleaned them up.

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
schustmi: I like the different log sources, I think that would solve the issue. For now, removing this would cause the following, which I think was also part of this initial ticket:
- step container logs a few messages, but they don't get flushed
- step container launches the step operator container
- step operator container logs enough messages to flush. These now get written before the logs in the initial bullet point.

src/zenml/logging/__init__.py
schustmi: Ah yes that confused me, sorry! Just weird that they're in separate places IMO

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
bcdurak: Yes, that's why I added it back in. Once we allow multiple log sources for steps, we can remove this again.

src/zenml/logging/step_logging.py
bcdurak: Adjusted it accordingly.

src/zenml/logging/step_logging.py
schustmi: I guess there might be some small inconsistencies depending on when threads try to put stuff in the queue, e.g.
- the log thread starts fetching things from the queue
- some other threads puts stuff in the queue
- without locks, that item now might or might not get picked up by the log thread

Probably not an issue though

bcdurak: Yeah, I don't think this would be an issue at the moment.

src/zenml/logging/step_logging.py
```diff
@@ -252,40 +260,192 @@ def __init__(
...
+                self.log_queue.task_done()
+
+    def _log_storage_worker(self) -> None:
+        """Log storage thread worker that processes the log queue."""
+        # Process the log queue until shutdown is requested
+        while not self.shutdown_event.is_set():
+            self._process_log_queue()
+
+        # Shutdown requested - drain remaining queue items and merge log files
+        while True:
```
schustmi: Why do we need to have the `while True` here? Isn't the `_process_log_queue` already fetching the entire queue?

bcdurak: Good catch, removed it.

src/zenml/logging/step_logging.py
bcdurak: I have adjusted it according to our discussion.

src/zenml/logging/step_logging.py
bcdurak: I switched to a single queue method. Feel free to check it out.

src/zenml/integrations/aws/step_operators/sagemaker_step_operator.py
bcdurak: After switching to single queue, I still need to adjust it. I am currently on it.

bcdurak: Done adjusting.

src/zenml/logging/step_logging.py
```diff
@@ -232,93 +235,209 @@ def _read_file(
...
+        finally:
+            # Always mark all queue tasks as done
+            for _ in messages:
+                self.log_queue.task_done()
+
+    def _log_storage_worker(self) -> None:
+        """Log storage thread worker that processes the log queue."""
+        # Process the log queue until shutdown is requested
+        while not self.shutdown_event.is_set():
+            self._process_log_queue()
```
schustmi, line 348: One last thing: This log thread would now constantly upload log messages, even if just a single message is in the queue right? There is no sleep, this thread would constantly do work if there are any messages in the queue? Also, the write interval is not respected at all currently?

bcdurak: I mean, the write interval dictates how long the `log_queue.get(...)` waits for a message (in the case of an empty) before checking if the thread is shutdown. Maybe it makes more sense to set this to a constant value and use the write interval as a sleep mechanism. However, in this case, I would keep this value relatively low. Because, any consecutive write operation will start after the IO operation time + waiting time of the previous iteration. In which case, I am afraid that the queue might fill up easier and we might block main thread.

bcdurak: Reduced the get timeout to `1` and set the write interval (sleep time) to `2` seconds (configurable).

schustmi (approved):

PR #3856, opened 2025-07-22
Pipeline snapshots

schustmi:
## Describe changes

This PR prepares the removal of run templates, and instead exposes `PipelineSnapshots` (former `PipelineDeployments`) as a user-facing concept as a replacement.

Run templates are already a very thin wrapper (name, description, tags) around the underlying pipeline snapshot. Instead of having this wrapper entity, this PR exposes the snapshot itself as something that can be triggered from the server and viewed in the dashboard. To achieve feature parity with run templates, these snapshots can now have version names and tags. One difference to run templates is that these snapshots will require a unique version name per pipeline, instead of the globally unique name that run templates require.

**Implementation decisions:**
- Unlisted pipeline runs do not exist anymore. Every pipeline run is now associated with a pipeline, and deleting the pipeline also deletes all its runs.
- ~~Creating versioned snapshots (which means giving them a version name) will only be possible if this deployment can actually be triggered. This means when running with a Pro server, and only for containerized remote stacks without custom components.~~ As we're going to use this entity for multiple purposes (not just triggering runs from the server), we don't prevent their creation anymore.

**Frontend changes:**
- (Required): Require confirmation when deleting a pipeline, and also mention that this will now delete all runs and deployments associated with the pipeline.
- (Required): Fix the DAG. The response may now contain nodes of type `triggered_run`, which cause the current implementation to fail. We should instead display them and link to the triggered pipeline run.
- (Optional): Switch to displaying pipeline snapshots and not run templates. This is not required immediately as this PR is implemented in a backwards compatible way, but ideally would happen in the same release still.
- (Optional): On the snapshot detail, the code section should show `Client().trigger_pipeline(snapshot_id=...)` instead of `Client().trigger_pipeline(template_id=...)`

**Changes that need to be mentioned in the release notes**:
- This version is not compatible with older versions.
- Run templates are deprecated and replaced by pipeline snapshots.
- The `StackComponent.prepare_pipeline_deployment(...)` method has been removed, and is not being called anymore.
- The `ZENML_SERVER_MAX_CONCURRENT_TEMPLATE_RUNS` env variable has been renamed to `ZENML_SERVER_MAX_CONCURRENT_SNAPSHOT_RUNS`. The legacy one has been deprecated but is still supported.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/pipeline.py
+++ b/src/zenml/cli/pipeline.py
@@ -857,13 +857,6 @@ def create_pipeline_snapshot(
     required=False,
     help="Path to configuration file for the run.",
 )
-@click.option(
-    "--stack",
-    "stack_name_or_id",
-    type=str,
-    required=False,
-    help="Name or ID of the stack for which to find a snapshot.",
-)
 def trigger_snapshot(
     snapshot_id: str,
     config_path: Optional[str] = None,
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -3535,7 +3535,7 @@ class Client(metaclass=ClientMetaClass):
             wait_for_pipeline_run_to_finish,
         )
 
-        if Counter([snapshot_id, template_id, pipeline_name_or_id])[None] != 1:
+        if Counter([snapshot_id, template_id, pipeline_name_or_id])[None] != 2:
             raise RuntimeError(
                 "You need to specify exactly one of snapshot, template or "
                 "pipeline to trigger."


src/zenml/cli/pipeline.py
```diff
@@ -364,6 +369,101 @@ def create_run_template(
...
+        cli_utils.declare(
+            "You can now trigger this deployment from the dashboard or by "
+            "calling `Client().trigger_deployment"
+            f"(deployment_id={deployment.id})`"
+        )
+    else:
+        cli_utils.declare(
+            "This deployment can not be triggered because the orchestrator "
+            "is not running remotely with Docker images."
+        )
```
AlexejPenner: Does this mean above it will actually create som deployment instance but then tell the user its unusable

src/zenml/models/v2/core/pipeline_deployment.py
```diff
@@ -213,6 +303,34 @@ def get_hydrated_version(self) -> "PipelineDeploymentResponse":
...
+            the value of the property.
+        """
+        return self.get_body().version
+
+    @property
+    def runnable(self) -> bool:
+        """The `runnable` property.
+
+        Returns:
+            the value of the property.
```
AlexejPenner: I feel we could elaborate more here, if I know nothing and try to inspect what it means to be runnable, I would love to actually be told somewhere.

src/zenml/models/v2/core/step_run.py
```diff
@@ -615,6 +619,15 @@ def model_version(self) -> Optional[ModelVersionResponse]:
...
         return self.get_resources().model_version
 
+    @property
+    def triggered_runs(self) -> List[UUID]:
+        """The `triggered_runs` property.
+
+        Returns:
+            the value of the property.
+        """
+        return self.get_resources().triggered_runs
```
AlexejPenner: is there any pagination of this? Potentially this could be a huge list

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -1553,6 +1551,31 @@ def create_run_template(
             name=name, deployment_id=deployment.id, **kwargs
         )
 
+    def deploy(
+        self,
+        name: str,
```
AlexejPenner, line 1584: do we need to force people to give the deployment a name, or can we choose to numerically name these by default?

src/zenml/cli/pipeline.py
schustmi: Yes that is correct. Do you think we should try to prevent this? Maybe the user also just wanted to use this to create a named reference to see some code version related to it? (unlikely for sure)

src/zenml/models/v2/core/step_run.py
schustmi: Yes you're right, I just made it a list of `UUIDs` instead of the actual objects to decrease the potential response size. But we might not even need this property on the response at all, I'll make this fetchable using `Client().list_pipeline_runs(triggered_by_step_run=<ID>)` or something along those lines.

src/zenml/pipelines/pipeline_definition.py
schustmi: For now I've made it explicit only, as I'm not really sure what to call these if given. For example:
- What would be the version name if I already have two deployments with names `A` and `B`? Would it be `1` or `3`?
- What would be the version name if I already have two deployments with names `1` and `3`?

AlexejPenner: good question - I guess this means either we allow explicit naming or we completely take charge of implicit naming

Maybe to turn around the question, whtas the point of allowing explicit naming, what will these names be used for

I want to avoid:

production_deployment_final_final_2

src/zenml/cli/pipeline.py
AlexejPenner: hmm, as it stands I think preventing from the get go might be the better choice

src/zenml/pipelines/pipeline_definition.py
schustmi: Well these names will be either specified when triggering a deployment, but also will be shown in some list as the available deployments for a specific pipeline. I think having the option to define names is just nice for users, as they can use something more expressive than just numbers, e.g. `2025-07-28` as the name of a version.

I guess we could also have a check that does not allow numeric version names when defining them explicitly, not sure thats a good idea though

schustmi: See  the PR description on how this will be handled.

src/zenml/models/v2/core/pipeline_deployment.py
AlexejPenner: HAHA just resolved it uncommented, sneaky - I still think this docstring is useless, just added to pass linting - what does it mean to be runnable? If I am in my IDE and i do pipeline_deploymnet.runnable - i would love to be told what it means to be runnable

src/zenml/zen_server/routers/pipeline_deployments_endpoints.py
```diff
@@ -223,6 +240,7 @@ def get_deployment(
         get_method=zen_store().get_deployment,
         hydrate=hydrate,
         step_configuration_filter=step_configuration_filter,
+        include_config_schema=include_config_schema,
```
AlexejPenner: is this for simplifying the frontend?

src/zenml/zen_stores/template_utils.py
```diff
@@ -27,11 +27,61 @@
...
+            raise ValueError(
+                "Unable to create deployment versions as a component of the "
+                "associated stack has no flavor."
+            )
+
+        if component.flavor_schema.is_custom:
+            raise ValueError(
+                "Unable to create deployment versions as a component of the "
+                "associated stack has a custom flavor."
+            )
```
AlexejPenner: Is there any plan to deal with this in the future, even with a workaround?

src/zenml/models/v2/core/pipeline_deployment.py
schustmi: Yes this is just for linting, same as the 1000 other model properties where we do the same thing. This is better than having to manage the same explanation twice, and that can be found in the pedantic model field that this property references.

src/zenml/zen_server/routers/pipeline_deployments_endpoints.py
schustmi: Not so much about simplifying, but including this schema is quite expensive and not necessary anywhere unless trying to run the deployment. The deployment itself gets fetched in quite a few places, which is why I added this.

src/zenml/zen_stores/template_utils.py
schustmi: Workaround - not possible I think.
Idea how to solve it exists, but I don't think the implementation will happen anytime soon

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4881,96 +4972,208 @@ def create_deployment(
...
+                schema_class=PipelineSnapshotSchema,
+                session=session,
+            )
+
+            session.delete(snapshot)
+
+            # We set the reference of all snapshots to this snapshot to null
+            # manually as we can't have a foreign key there to avoid a cycle
+            snapshots = session.exec(
+                select(PipelineSnapshotSchema).where(
```
stefannica, line 5159: Is this based on an actual error that you got during a DB upgrade ? I'm pretty sure that I'm accounting for self-references in the DB backup/restore process, if that's all that you're worried about here:

image

schustmi: Yep, back when implementing run templates, self-references caused an issue in the DB backup/restore. I haven't actually checked this time, I'll give it another try

PR #3874, opened 2025-08-01
Different pipeline execution modes

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
@@ -402,6 +402,63 @@ def main() -> None:
 
                 raise
 
+        def stop_step(step_name: str) -> None:
+            """Stop a step.
+
+            Args:
+                step_name: Name of the step.
+            """
+            # Use both run_id and step_name for precise targeting
+            label_selector = (
+                f"run_id={kube_utils.sanitize_label(str(pipeline_run.id))},"
+                f"step_name={kube_utils.sanitize_label(step_name)}"
+            )
+    
+            try:
+                jobs = batch_api.list_namespaced_job(
+                    namespace=namespace,
+                    label_selector=label_selector,
+                )
+            except Exception as e:
+                logger.warning(f"Failed to find jobs for step {step_name}: {e}")
+                return
+
+            if not jobs.items:
+                logger.debug(f"No jobs found for step {step_name}")
+                return
+
+            for job in jobs.items:
+                # Check if job is already completed or failed - skip deletion
+                job_finished = False
+                if job.status.conditions:
+                    for condition in job.status.conditions:
+                        if (
+                            condition.type in ["Complete", "Failed"] 
+                            and condition.status == "True"
+                        ):
+                            job_finished = True
+                            break
+                
+                if job_finished:
+                    logger.debug(
+                        f"Job {job.metadata.name} for step {step_name} already finished, skipping deletion"
+                    )
+                    continue
+
+                # Delete the running/pending job
+                try:
+                    batch_api.delete_namespaced_job(
+                        name=job.metadata.name,
+                        namespace=namespace,
+                        propagation_policy="Foreground",
+                    )
+                    logger.info(
+                        f"Successfully stopped step job: {job.metadata.name}"
+                    )
+                except Exception as e:
+                    logger.warning(f"Failed to stop step job {job.metadata.name}: {e}")
+
+            
         def finalize_run(node_states: Dict[str, NodeStatus]) -> None:
             """Finalize the run.
 
@@ -499,6 +556,7 @@ def main() -> None:
                 continue_fn=check_pipeline_cancellation,
                 parallel_node_startup_waiting_period=parallel_node_startup_waiting_period,
                 max_parallelism=pipeline_settings.max_parallelism,
+                stop_fn=stop_step,
             ).run()
             logger.info("Orchestration pod completed.")
         finally:
--- a/src/zenml/orchestrators/local/local_orchestrator.py
+++ b/src/zenml/orchestrators/local/local_orchestrator.py
@@ -17,6 +17,7 @@ import time
 from typing import TYPE_CHECKING, Dict, Optional, Type
 from uuid import uuid4
 
+from zenml.enums import ExecutionMode
 from zenml.logger import get_logger
 from zenml.orchestrators import (
     BaseOrchestrator,
@@ -76,8 +77,26 @@ class LocalOrchestrator(BaseOrchestrator):
         self._orchestrator_run_id = str(uuid4())
         start_time = time.time()
 
+        # Get execution mode for failure handling
+        execution_mode = deployment.pipeline_configuration.execution_mode
+
+        if execution_mode is None:
+            execution_mode = ExecutionMode.CONTINUE_ON_FAILURE
+
+        # Track failed steps
+        failed_steps = []
+
         # Run each step
         for step_name, step in deployment.step_configurations.items():
+            if any(fs in step.spec.upstream_steps for fs in failed_steps):
+                continue
+
+            if (
+                execution_mode == ExecutionMode.STOP_ON_FAILURE
+                and len(failed_steps) > 0
+            ):
+                continue
+
             if self.requires_resources_in_orchestration_environment(step):
                 logger.warning(
                     "Specifying step resources is not supported for the local "
@@ -86,9 +105,13 @@ class LocalOrchestrator(BaseOrchestrator):
                     step_name,
                 )
 
-            self.run_step(
-                step=step,
-            )
+            try:
+                self.run_step(step=step)
+            except Exception as e:
+                failed_steps.append(step_name)
+
+                if execution_mode == ExecutionMode.FAIL_FAST:
+                    raise e
 
         run_duration = time.time() - start_time
         logger.info(
@@ -192,3 +215,19 @@ class LocalOrchestratorFlavor(BaseOrchestratorFlavor):
             The implementation class for this flavor.
         """
         return LocalOrchestrator
+
+    @property
+    def supported_execution_modes(self):
+        """Returns the supported execution modes for this flavor.
+
+        Returns:
+            A tuple of supported execution modes.
+        """
+        from zenml.enums import ExecutionMode
+
+        # Local orchestrator supports all execution modes
+        return (
+            ExecutionMode.FAIL_FAST,
+            ExecutionMode.STOP_ON_FAILURE,
+            ExecutionMode.CONTINUE_ON_FAILURE,
+        )
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -123,6 +123,7 @@ from zenml.enums import (
     ArtifactSaveType,
     AuthScheme,
     DatabaseBackupStrategy,
+    ExecutionMode,
     ExecutionStatus,
     LoggingLevels,
     MetadataResourceTypes,
@@ -8932,6 +8933,30 @@ class SqlZenStore(BaseZenStore):
                     f"Cannot create step '{step_run.name}' for pipeline in "
                     f"{run.status} state. Pipeline run ID: {step_run.pipeline_run_id}"
                 )
+
+            # Handle FAILED status based on execution mode
+            if run.status == ExecutionStatus.FAILED:
+                # Get execution mode from pipeline configuration
+                pipeline_configuration = (
+                    PipelineConfiguration.model_validate_json(
+                        run.deployment.pipeline_configuration
+                    )
+                )
+
+                if pipeline_configuration.execution_mode is None:
+                    execution_mode = ExecutionMode.CONTINUE_ON_FAILURE
+
+                # Only CONTINUE_ON_FAILURE allows new steps when pipeline is FAILED
+                if (
+                    pipeline_configuration.execution_mode
+                    != ExecutionMode.CONTINUE_ON_FAILURE
+                ):
+                    raise IllegalOperationError(
+                        f"Cannot create step '{step_run.name}' for pipeline in "
+                        f"FAILED state with execution mode {execution_mode}. "
+                        f"Pipeline run ID: {step_run.pipeline_run_id}"
+                    )
+
             self._get_reference_schema_by_id(
                 resource=step_run,
                 reference_schema=StepRunSchema,
--- a/src/zenml/config/pipeline_configurations.py
+++ b/src/zenml/config/pipeline_configurations.py
@@ -22,6 +22,7 @@ from zenml.config.constants import DOCKER_SETTINGS_KEY
 from zenml.config.retry_config import StepRetryConfig
 from zenml.config.source import SourceWithValidator
 from zenml.config.strict_base_model import StrictBaseModel
+from zenml.enums import ExecutionMode
 from zenml.model.model import Model
 from zenml.utils.tag_utils import Tag
 from zenml.utils.time_utils import utc_now
@@ -42,6 +43,7 @@ class PipelineConfigurationUpdate(StrictBaseModel):
     enable_artifact_visualization: Optional[bool] = None
     enable_step_logs: Optional[bool] = None
     enable_pipeline_logs: Optional[bool] = None
+    execution_mode: Optional[ExecutionMode] = None
     settings: Dict[str, SerializeAsAny[BaseSettings]] = {}
     tags: Optional[List[Union[str, "Tag"]]] = None
     extra: Dict[str, Any] = {}
--- a/src/zenml/config/pipeline_run_configuration.py
+++ b/src/zenml/config/pipeline_run_configuration.py
@@ -24,6 +24,7 @@ from zenml.config.schedule import Schedule
 from zenml.config.source import SourceWithValidator
 from zenml.config.step_configurations import StepConfigurationUpdate
 from zenml.config.strict_base_model import StrictBaseModel
+from zenml.enums import ExecutionMode
 from zenml.model.model import Model
 from zenml.models import PipelineBuildBase
 from zenml.utils import pydantic_utils
@@ -41,6 +42,7 @@ class PipelineRunConfiguration(
     enable_artifact_visualization: Optional[bool] = None
     enable_step_logs: Optional[bool] = None
     enable_pipeline_logs: Optional[bool] = None
+    execution_mode: Optional[ExecutionMode] = None
     schedule: Optional[Schedule] = None
     build: Union[PipelineBuildBase, UUID, None] = Field(
         default=None, union_mode="left_to_right"
--- a/src/zenml/constants.py
+++ b/src/zenml/constants.py
@@ -525,7 +525,7 @@ LOGS_STORAGE_QUEUE_TIMEOUT = handle_int_env_var(
 )
 # Log batch-write and merge windows in seconds.
 LOGS_WRITE_INTERVAL_SECONDS = handle_int_env_var(
-    ENV_ZENML_LOGS_WRITE_INTERVAL_SECONDS, default=2
+    ENV_ZENML_LOGS_WRITE_INTERVAL_SECONDS, default=1
 )
 LOGS_MERGE_INTERVAL_SECONDS = handle_int_env_var(
     ENV_ZENML_LOGS_MERGE_INTERVAL_SECONDS, default=10 * 60
--- a/src/zenml/enums.py
+++ b/src/zenml/enums.py
@@ -114,7 +114,7 @@ class ExecutionStatus(StrEnum):
         Returns:
             Whether the execution status refers to a failed execution.
         """
-        return self in {ExecutionStatus.FAILED}
+        return self in {ExecutionStatus.FAILED, ExecutionStatus.FAILING}
 
 
 class LoggingLevels(Enum):
@@ -126,6 +126,14 @@ class LoggingLevels(Enum):
     INFO = logging.INFO
     DEBUG = logging.DEBUG
     CRITICAL = logging.CRITICAL
+
+
+class ExecutionMode(StrEnum):
+    """Enum that represents the execution mode of a pipeline run."""
+
+    FAIL_FAST = "fail_fast"
+    STOP_ON_FAILURE = "stop_on_failure"
+    CONTINUE_ON_FAILURE = "continue_on_failure"
 
 
 class StackComponentType(StrEnum):
--- a/src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
+++ b/src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
@@ -376,3 +376,19 @@ class KubernetesOrchestratorFlavor(BaseOrchestratorFlavor):
         )
 
         return KubernetesOrchestrator
+
+    @property
+    def supported_execution_modes(self):
+        """Returns the supported execution modes for this flavor.
+
+        Returns:
+            A tuple of supported execution modes.
+        """
+        from zenml.enums import ExecutionMode
+
+        # Kubernetes orchestrator supports all execution modes
+        return (
+            ExecutionMode.FAIL_FAST,
+            ExecutionMode.STOP_ON_FAILURE,
+            ExecutionMode.CONTINUE_ON_FAILURE,
+        )
--- a/src/zenml/orchestrators/base_orchestrator.py
+++ b/src/zenml/orchestrators/base_orchestrator.py
@@ -34,7 +34,7 @@ from zenml.constants import (
     ENV_ZENML_PREVENT_CLIENT_SIDE_CACHING,
     handle_bool_env_var,
 )
-from zenml.enums import ExecutionStatus, StackComponentType
+from zenml.enums import ExecutionMode, ExecutionStatus, StackComponentType
 from zenml.exceptions import RunMonitoringError, RunStoppedException
 from zenml.logger import get_logger
 from zenml.metadata.metadata_types import MetadataType
@@ -444,10 +444,29 @@ class BaseOrchestrator(StackComponent, ABC):
             deployment: The deployment to prepare.
         """
         self._active_deployment = deployment
+        self._validate_execution_mode()
 
     def _cleanup_run(self) -> None:
         """Cleans up the active run."""
         self._active_deployment = None
+
+    def _validate_execution_mode(self) -> None:
+        """Validate that the requested execution mode is supported.
+
+        This base implementation logs the execution mode being used.
+        Individual orchestrator implementations can override this method
+        to add specific validation.
+        """
+        if not self._active_deployment:
+            return
+
+        execution_mode = (
+            self._active_deployment.pipeline_configuration.execution_mode
+        )
+        if execution_mode is None:
+            logger.debug("No execution mode specified, using default behavior")
+        else:
+            logger.info(f"Using execution mode: {execution_mode}")
 
     def fetch_status(
         self, run: "PipelineRunResponse", include_steps: bool = False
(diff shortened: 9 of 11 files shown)


src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -505,6 +505,67 @@ def start_step_job(node: Node) -> NodeStatus:
...
+            """Stop a step.
+
+            Args:
+                node: The node to stop.
+            """
+            step_name = node.id
+
+            label_selector = (
+                f"run_id={kube_utils.sanitize_label(str(pipeline_run.id))},"
+                f"step_name={kube_utils.sanitize_label(node.id)}"
```
schustmi, line 527: After this sanitization of the step name, they might not be unique anymore, which is why this might find multiple steps. I've introduced annotations with the step name that should be used to verify this

src/zenml/orchestrators/local/local_orchestrator.py
```diff
@@ -76,8 +77,39 @@ def submit_pipeline(
...
         # Run each step
         for step_name, step in deployment.step_configurations.items():
+            if (
+                execution_mode == ExecutionMode.STOP_ON_FAILURE
+                and len(failed_steps) > 0
+            ):
+                logger.warning(
+                    "Stopping pipeline run due to failure in step(s) %s and "
+                    "execution mode %s.",
+                    ", ".join(failed_steps),
```
schustmi, line 98: There can never be multiple steps right?

src/zenml/orchestrators/local/local_orchestrator.py
```diff
@@ -76,8 +77,39 @@ def submit_pipeline(
...
+                    execution_mode,
+                )
+                continue
```
schustmi, line 102: Why continue, this should break?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -182,7 +182,15 @@ def signal_handler(signum: int, frame: Any) -> None:
                             status=ExecutionStatus.STOPPED,
                             end_time=utc_now(),
                         )
-                    raise RunStoppedException("Pipeline run in stopped.")
+                    raise RunStoppedException("Pipeline run is stopped.")
+                elif step_run := pipeline_run.steps.get(self._step_name, None):
```
schustmi: Fetching all steps, unnecessary

schustmi (changes requested):

src/zenml/orchestrators/local/local_orchestrator.py
bcdurak: Fixed the message.

src/zenml/orchestrators/step_launcher.py
bcdurak: Switched to client.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: Fixed it.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -505,6 +505,77 @@ def start_step_job(node: Node) -> NodeStatus:
...
+                            job_finished = True
+                            break
+
+                if job_finished:
+                    logger.debug(
+                        f"Job {job.metadata.name} for step {step_name} already finished, skipping deletion"
+                    )
+                    continue
+
+                step_name_annotation = job.metadata.annotations.get(
```
schustmi: Shouldn't this be above the checks if it's completed? Otherwise there might be misleading logs

src/zenml/orchestrators/local/local_orchestrator.py
```diff
@@ -76,8 +81,39 @@ def submit_pipeline(
...
+        # Track failed steps
+        failed_steps: List[str] = []
+
         # Run each step
         for step_name, step in deployment.step_configurations.items():
+            if (
+                execution_mode == ExecutionMode.STOP_ON_FAILURE
+                and len(failed_steps) > 0
+            ):
+                logger.warning(
```
schustmi, line 95: One more UX question: In the case of the `STOP_ON_FAILURE` execution mode, would you expect an exception or not?

Currently, I could run my pipeline like this `my_pipeline()` and it runs just fine without an exception, which seems a little weird? Or maybe I'm just biased by the way it previously worked?

(Actually the same thing applied to `CONTINUE_ON_ERROR`)

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
```diff
@@ -153,8 +153,39 @@ def submit_pipeline(
...
+                execution_mode == ExecutionMode.STOP_ON_FAILURE
+                and len(failed_steps) > 0
+            ):
+                logger.warning(
+                    "Stopping pipeline run due to failure in step(s) %s and "
+                    "execution mode %s.",
+                    ", ".join(failed_steps),
+                    execution_mode,
+                )
+                continue
```
schustmi: = only one step, + should be `break`, similar to local orchestrator

schustmi (changes requested):

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: Makes sense, relocated it.

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
bcdurak: Done.

src/zenml/orchestrators/local/local_orchestrator.py
bcdurak: I have added a `RuntimeError` at the end of the pipeline run, if there are any failed steps.

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
```diff
@@ -153,8 +153,39 @@ def submit_pipeline(
         environment[ENV_ZENML_LOCAL_STORES_PATH] = local_stores_path
         start_time = time.time()
 
+        # Get execution mode for failure handling
```
schustmi:
```suggestion
```

I'll suggest removing all of those comments if they provide no value and just make it more effort to read the code. In general, if a variable is called `execution_mode`, I think its pretty clear that this is fetching the execution mode 😄

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
```diff
@@ -153,8 +153,39 @@ def submit_pipeline(
...
+        execution_mode = deployment.pipeline_configuration.execution_mode
+
+        # List of steps that failed
```
schustmi:
```suggestion
```

src/zenml/orchestrators/local/local_orchestrator.py
```diff
@@ -76,8 +80,39 @@ def submit_pipeline(
         self._orchestrator_run_id = str(uuid4())
         start_time = time.time()
 
+        # Get execution mode for failure handling
```
schustmi:
```suggestion
```

src/zenml/orchestrators/local/local_orchestrator.py
```diff
@@ -76,8 +80,39 @@ def submit_pipeline(
...
+        execution_mode = deployment.pipeline_configuration.execution_mode
+
+        # Track failed steps
```
schustmi:
```suggestion
```

schustmi (changes requested):

schustmi (approved):

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -9348,6 +9356,25 @@ def update_run_step(
                         session=session,
                     )
 
+            # If the step failed, we need to stop all other steps in the pipeline.
+            # This is only relevant in FAIL_FAST mode.
+            if step_run_update.status == ExecutionStatus.FAILED:
```
schustmi: I think this same logic must also be in `create_run_step`, as we sometimes publish a failed step immediately.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -9348,6 +9356,25 @@ def update_run_step(
...
+                execution_mode = None
+                if deployment := existing_step_run.deployment:
+                    config = PipelineConfiguration.model_validate_json(
```
schustmi: `PipelineRunSchema.get_pipeline_configuration()` already does this for you, might be nicer to use

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Added this.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Thanks, used this.

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -470,6 +473,15 @@ def end_time(self) -> Optional[datetime]:
         """
         return self.get_metadata().end_time
 
+    @property
+    def in_progress(self) -> Optional[bool]:
```
schustmi: non-optional

src/zenml/client.py
```diff
@@ -3979,6 +3979,7 @@ def list_pipeline_runs(
         model: Optional[Union[UUID, str]] = None,
         stack: Optional[Union[UUID, str]] = None,
         stack_component: Optional[Union[UUID, str]] = None,
+        in_progress: Optional[bool] = None,
```
schustmi, line 3982: docstring

src/zenml/orchestrators/step_launcher.py
```diff
@@ -182,7 +182,17 @@ def signal_handler(signum: int, frame: Any) -> None:
                             status=ExecutionStatus.STOPPED,
                             end_time=utc_now(),
                         )
-                    raise RunStoppedException("Pipeline run in stopped.")
+                    raise RunStoppedException("Pipeline run is stopped.")
+                elif step_run := client.get_run_step(
+                    self._step_run.id, hydrate=False
+                ):
+                    if step_run.status in ExecutionStatus.STOPPING:
```
schustmi: This maybe does what you intend, but `==` is the correct syntax

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
```diff
@@ -205,7 +234,17 @@ def submit_pipeline(
                     logger.info(line.strip().decode())
             except ContainerError as e:
                 error_message = e.stderr.decode()
-                raise RuntimeError(error_message)
+                failed_steps.append(step_name)
+                if execution_mode == ExecutionMode.FAIL_FAST:
```
schustmi, line 252: Doesn't this also apply to `STOP_ON_FAILURE`?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -514,6 +514,77 @@ def start_step_job(node: Node) -> NodeStatus:
...
+                        propagation_policy="Foreground",
+                    )
+                    logger.info(
+                        f"Successfully stopped step job: {job.metadata.name}"
+                    )
+                except Exception as e:
+                    logger.warning(
+                        f"Failed to stop step job {job.metadata.name}: {e}"
+                    )
+
```
schustmi, line 589: There should be a `break` after the deletion I assume?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8969,12 +8971,11 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
             self._set_request_user_id(request_model=step_run, session=session)
 
             # Check if the pipeline run exists
-            run = self._get_reference_schema_by_id(
-                resource=step_run,
-                reference_schema=PipelineRunSchema,
-                reference_id=step_run.pipeline_run_id,
-                session=session,
-            )
+            run = session.exec(
```
schustmi: By replacing this method in all the places, we lose all the nice error messages and explanations, not a good idea IMO.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8994,18 +9008,6 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
             )
             step_config = run.get_step_configuration(step_name=step_run.name)
 
-            # Release the read locks of the previous two queries before we
-            # try to acquire more exclusive locks
-            session.commit()
```
schustmi: Removing these lines will cause a deadlock (hence the comment that tells you to release the lock early before trying to acquire more/other locks). This is the entire point why this isn't as easy. It won't work if we just acquire the run lock, then do a bunch of other stuff and then very late release it again. It will always cause deadlocks with other parallel calls.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -9676,8 +9736,6 @@ def _update_pipeline_run_status(
 
         pipeline_run.update(run_update)
         session.add(pipeline_run)
-        # Commit so that we release the lock on the pipeline run.
```
schustmi: There's a reason this comment is here, and you can't simply delete this commit without causing issues. This is why it's important to remove any garbage AI comment. If there is a comment in code, there is a good reason it means something useful. If this was a regular "I didn't think about it but just put it there" commit, the comment wouldn't be necessary.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -9620,30 +9686,24 @@ def _set_run_step_output_artifact(
...
 
         # Make sure we start with a fresh transaction before locking the
         # pipeline run
-        session.commit()
-        pipeline_run = session.exec(
-            select(PipelineRunSchema)
-            .with_for_update()
-            .where(PipelineRunSchema.id == pipeline_run_id)
-        ).one()
-        step_run_statuses = session.exec(
```
schustmi: There is a reason that I wrote this query instead of using `pipeline_run.step_runs`: It's much quicker, it doesn't load the entire row but just the status that we need.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -9081,6 +9083,29 @@ def create_run_step(self, step_run: StepRunRequest) -> StepRunResponse:
...
+            # If the created step is in failed state, we need to stop all
+            # other steps in the pipeline. This is only relevant in
+            # FAIL_FAST mode.
+            if step_schema.status == ExecutionStatus.FAILED:
+                execution_mode = (
+                    run.get_pipeline_configuration().execution_mode
+                )
+
+                if execution_mode == ExecutionMode.FAIL_FAST:
+                    for step in run.step_runs:
```
schustmi: A separate query is much more efficient than looping over all steps which always loads the full rows.

schustmi (changes requested):

src/zenml/client.py
bcdurak: already added.

src/zenml/models/v2/core/pipeline_run.py
bcdurak: done.

src/zenml/orchestrators/step_launcher.py
bcdurak: done.

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
bcdurak: I have rearranged the logic.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: I am not sure. Just for my understanding, ideally, there will be only one job per step here, right?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -676,6 +712,7 @@ def update_placeholder(
         self.orchestrator_run_id = request.orchestrator_run_id
         self.orchestrator_environment = orchestrator_environment
         self.status = request.status.value
+        self.in_progress = not request.status.is_finished
```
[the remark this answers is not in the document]
bcdurak: Tuple?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
schustmi: Exactly, but `jobs` might contain multiple. You then, in the beginning of the for-loop, filter out all the "wrong" candidates. But if you already found one, we're sure that there will be no other match. So it's inefficient to keep looping over the (potentially thousand) jobs in the list, only to check their annotations and then `continue`

bcdurak: Ah, that makes sense. I forgot about the trick with that we did with the `STEP_NAME_ANNOTATION_KEY` for a second. Adding the break now.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -599,6 +628,11 @@ def update(self, run_update: "PipelineRunUpdate") -> "PipelineRunSchema":
             The updated `PipelineRunSchema`.
         """
         if run_update.status:
+            self.status = run_update.status.value
```
schustmi: I think even this line here should be removed. The run status update should always happen via the SQLZenStore._update_run_status method anyway, no?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -613,6 +647,8 @@ def update(self, run_update: "PipelineRunUpdate") -> "PipelineRunSchema":
                 if run_update.status_reason:
                     self.status_reason = run_update.status_reason
 
+            self.in_progress = self._check_if_run_in_progress()
```
schustmi, line 648: Somehow I'm still thinking we need to optimize this and not call it as often, because the _check_if_run_in_progress is essentially the most DB heavy operation that we're doing in all of the server. For example, do we need to call this every time, or is it enough to call it if the pipeline run status changes?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -691,3 +728,128 @@ def is_placeholder_run(self) -> bool:
             ExecutionStatus.INITIALIZING.value,
             ExecutionStatus.PROVISIONING.value,
         }
+
+    def _check_if_run_in_progress(self) -> bool:
+        """Checks whether the run is in progress.
+
+        Returns:
+            A tuple containing:
```
schustmi: Not a tuple

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -691,3 +728,128 @@ def is_placeholder_run(self) -> bool:
...
+
+                    # Create Step objects from configurations
+                    pipeline_configuration = self.get_pipeline_configuration()
+                    steps = []
+                    for step_config in step_configurations:
+                        step = Step.from_dict(
+                            data=json.loads(step_config.config),
+                            pipeline_configuration=pipeline_configuration,
+                        )
+                        steps.append(step)
```
schustmi: You implemented a helper function for this I think?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -691,3 +728,128 @@ def is_placeholder_run(self) -> bool:
...
+
+                    # Build a reverse DAG (step_name -> list of downstream steps)
+                    reverse_dag: Dict[str, List[str]] = {}
```
schustmi: This logic should be in (some) helper functions, and tested.

schustmi (changes requested):

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: Fixed.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: Yes, using it now. Also renamed it to `get_steps` because we already do the configuration to step conversion there.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: Not really, at least in the current version. Let's say I am trying to stop a pipeline. It sets the pipeline status to `STOPPING` manually, not through the steps. In that moment, the pipeline could have 3 `SUCCESSFUL` steps and 3 steps that are not started yet. In that moment, without feeding the `new_status` to this function, it is impossible to handle it.

schustmi: I guess what I was trying to say is this: Even in SQLZenStore.update_run(...) we eventually call `_update_pipeline_run_status(...)`, which sets the status again. So why do we have to do it in the first place via the schema?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: This only gets called, if the `run_update` has a `status` and that new `status` is failed. Is that what you mean?

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Reverted the changes here.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Reverted the changes here.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Reverted the changes here.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Reverted the changes here.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Reverted the changes here.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: I've put them in utility functions in `run_utils` and implemented unit and integrations tests.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: I am removing this line because it was logically wrong anyhow.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -514,6 +514,78 @@ def start_step_job(node: Node) -> NodeStatus:
...
+                try:
+                    batch_api.delete_namespaced_job(
+                        name=job.metadata.name,
+                        namespace=namespace,
+                        propagation_policy="Foreground",
+                    )
+                    logger.info(
+                        f"Successfully stopped step job: {job.metadata.name}"
+                    )
+                    break
```
schustmi, line 583: I guess ideally the break would also happen in the `except` case? 😄

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -691,3 +727,83 @@ def is_placeholder_run(self) -> bool:
...
+
+    def _check_if_run_in_progress(self) -> bool:
+        """Checks whether the run is in progress.
+
+        Returns:
+            A flag to indicate whether the run is in progress.
+        """
+        run_status = ExecutionStatus(self.status)
+
+        # Unfinished states are in progress
```
schustmi: Can we please remove all comments in this method. I couldn't find a single useful one that isn't explained by the next line in code.

src/zenml/utils/run_utils.py
```diff
@@ -165,3 +166,42 @@ def refresh_run_status(
...
+def build_dag_with_downstream_steps(steps: List[Step]) -> Dict[str, Set[str]]:
+    """Build DAG with downstream steps from a list of steps.
+
+    Args:
+        steps: The list of steps.
+
+    Returns:
+        The DAG with downstream steps.
+    """
+    dag: Dict[str, Set[str]] = {step.config.name: set() for step in steps}
```
schustmi: `step.config.name` is not a unique ID of the step, and is not what is used in `upstream_steps` either. It should never be used, I'll add a comment in another place on what to use instead.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -691,3 +727,83 @@ def is_placeholder_run(self) -> bool:
...
+                    # Get the statuses of all steps that have run so far
+                    steps_statuses = {
+                        s.name: ExecutionStatus(s.status)
+                        for s in self.step_runs
+                    }
+
+                    # Check if all non-skipped steps are finished
+                    for step in steps:
+                        # If it is a failed step or a downstream step of a failed step, continue
+                        if step.config.name in steps_to_skip:
```
schustmi: Again, this should never be used. Instead, use `StepConfigurationSchema.name` as the unique step ID.

This means your `get_steps` helper method should either
- return a list of `StepConfigurationSchema`
- return a `Dict[str, Step]` which has the ID as the key

src/zenml/utils/run_utils.py
```diff
@@ -165,3 +166,42 @@ def refresh_run_status(
         )
 
     return run
+
+
+def build_dag_with_downstream_steps(steps: List[Step]) -> Dict[str, Set[str]]:
```
schustmi: I think this can be simply `build_dag(...)`, no?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -691,3 +727,83 @@ def is_placeholder_run(self) -> bool:
...
+                ExecutionMode.FAIL_FAST,
+                ExecutionMode.STOP_ON_FAILURE,
+            ]:
+                return False
+
+            # Continue on failure is in progress
+            elif execution_mode == ExecutionMode.CONTINUE_ON_FAILURE:
+                if self.deployment:
+                    # Get all steps
+                    steps = self.get_steps()
```
schustmi: We can make this whole method much more efficient by loading `deployment_schema.pipeline_spec` instead of all the separate step configurations. This spec contains all the information to build the DAG

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: Hahaha sure, just so you know, these comments are not AI-generated. I usually write them down as pseudo-code to keep my sanity in check 😄

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: Done.

src/zenml/utils/run_utils.py
bcdurak: Renamed.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: I am now using the pipeline spec in the deployment (along with the `pipeline_parameter_name`) as discussed. I've resolved the comments above as well.

schustmi (approved):

AlexejPenner:
@bcdurak where is the docs for this ?

bcdurak:
@AlexejPenner Already backported to the release, you can find it [here](https://docs.zenml.io/concepts/steps_and_pipelines/advanced_features#execution-modes).

PR #3900, opened 2025-08-25
Caching by value

schustmi:
## Describe changes
This PR
- adds an interface that allows caching based on artifact values (instead of artifact ids)
- implements this interface for some (mostly primitive) artifact types stored by the some materializers
- implements cache policies to enable/disable this behavior

## Pre-requisites

## Types of changes


src/zenml/config/pipeline_configurations.py
```diff
@@ -51,6 +52,7 @@ class PipelineConfigurationUpdate(StrictBaseModel):
     parameters: Optional[Dict[str, Any]] = None
     retry: Optional[StepRetryConfig] = None
     substitutions: Dict[str, str] = {}
+    cache_policy: Optional[CachePolicyWithValidator] = None
```
bcdurak, line 55: I was wondering whether we should add the same attribute to the `PipelineConfiguration` with a default value. If you remember, we had a similar chat regarding the execution modes and ended up adding it to the `PipelineConfiguration` as well, while keeping it optional here.

bcdurak: Now that I see it, I think you already followed a similar pattern for the steps.

src/zenml/models/v2/core/artifact_version.py
```diff
@@ -201,6 +205,10 @@ class ArtifactVersionResponseBody(ProjectScopedResponseBody):
         title="ID of the artifact store in which this artifact is stored.",
         default=None,
     )
+    content_hash: Optional[str] = Field(
```
bcdurak, line 209: Once a hash is computed by the materializer, the resulting string gets generated and stored in the DB without any size limitations. Especially with custom materializers, people might implement something that returns a huge string and since it is a part of the body, perhaps we can add a limitation at some point?

bcdurak (commented):
The changes look good, leaving a few small comments.

src/zenml/config/pipeline_configurations.py
schustmi: They are very different, and what I implemented makes sense for my case:
- execution mode is required for each pipeline
- cache policy is required for each step
- cache policy is not required (doesn't even make sense) for a pipeline, and configuring it on the pipeline simply applies it to all steps

src/zenml/models/v2/core/artifact_version.py
schustmi: Hmm yeah I guess, I'll add it

src/zenml/config/pipeline_configurations.py
bcdurak: Yeah, that makes sense. And you already cover the default values with the steps. All good here then.

src/zenml/models/v2/core/artifact_version.py
schustmi: Done

bcdurak (approved):

PR #3901, opened 2025-08-25
Agent guide docs

htahir1:
## Describe changes
Added an agent guide

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/user-guide/agent-guide/README.md
+++ b/docs/book/user-guide/agent-guide/README.md
@@ -11,7 +11,9 @@ Transform your agent development from experimental scripts into systematic, prod
 
 ## Why pipelines for agents (for ML/AI engineers)
 
-Agents evolve quickly: prompts change, tools are added/removed, and behavior shifts in production. A pipeline-first approach gives you reproducibility, lineage, and safe iteration: you version prompts and configs, deploy with confidence, and evaluate continuously on real traces. ZenML lets you apply the same rigor you use for classical ML to agents—regardless of whether you use an agent framework or direct LLM calls. The benefit is speed with control: faster iteration loops, clear governance, and a shared operating model across teams.
+Agents evolve quickly: prompts change, tools are added/removed, and behavior shifts in production. A pipeline-first approach gives you reproducibility, lineage, and safe iteration: you version prompts and configs, deploy with confidence, and evaluate continuously on real traces. 
+
+ZenML lets you apply the same rigor you use for classical ML to agents—regardless of whether you use an agent framework or direct LLM calls. The benefit is speed with control: faster iteration loops, clear governance, and a shared operating model across teams.
 
 Common pitfalls of ad‑hoc agent work today:
 - Experiment across notebooks and frameworks without shared structure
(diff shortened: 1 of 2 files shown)


strickvl (commented):
## High-level comments

- somewhere you should make the distinction / mention what's the difference between LLMOps guide + agent guide. (maybe also at the top of both of those you should say 'oh, there's also this other guide which covers x and y and its purpose is z which you might want to look at also… and also link to it at the end of the respective guides)
- I think at a high level everything checks out (apart from the conditional routing etc in chapter 1), but as is this will raise many many questions, so IMO it needs to be fully expanded. (like the other guides)

# chapter 1: dev workflow

- the dev workflow where you compare two different agents IMO is not how people are developing. Rather people are comparing two different versions of the same agents, but with different prompts. I think that's something that def needs changing in the 'development & experimentation' section.
- the `complete_agent_development_pipeline` doesn't make sense to me. I'd just delete that whole "Complete Development Pipeline" section. I don't know that anyone would ever (want to) do it that way. Also this guide doesn't need that.
- For my taste, that 'development and experimentation' section is sort of the real meat of the guide and we don't really get into the details of the 'why' you might want to do one vs the other etc. These guides were always sort of educational as well as practical, so somehow I'd expect more prose amidst the code snippets. After all, this section is sort of the big sell… + must answer the question why should I use ZenML when experimenting on my agents instead of just doing it outside ZenML.

## Chapter 2: production & deployment

- we sort of miss a step here where we're logging things to the model control plane, since in the 'deploying your agent' it assumes that we've already done that?
- again here I don't think it's a realistic pattern that we're deciding in a conditional which config to use ('direct' or 'framework' etc.) That would have happened much earlier in the process.
- I think I'd want to see more of a story around 'versions' here. You're doing it in code, but it's somehow a bit implicit / not foregrounded?
- The bit about a/b testing + traffic_split was maybe an interesting idea, but it feels a bit half-baked since the exact implementation of that is hidden away.
- At the end you say 'the production logs + traces you're collecting are useful' etc, but AFAICS we didn't talk about that yet at all… 

## Chapter 3: evaluation

- For me, the high-level code examples make sense but for many people I can see them as prompting more questions than they answer. I.e. people will want the details. "how do you write an llm as judge eval" / 'what model to use' / 'how to create a test set' etc..
- I think maybe this page is fine for now, but going forward I think we're going to have to write up all the details of our take on this stuff as we'll be asked these questions repeatedly. (or point them to somewhere else that does answer them)

## Language use

- I think we can be a bit more deliberate about how / when we use agents vs workflows. e.g. in very first sentence of the guide, I read "Transform your agent development from experimental scripts into systematic, production-ready workflows using ZenML as your agent development platform." which somehow muddies the water a bit. (unless that's the point? if so, not sure that's the best way to handle that. IMO we should just lean into the word 'agent')
- Not sure 'chapter' is the right word, since the sections aren't labelled with numbers…
- The section "What ZenML tracks automatically" is maybe a bit misleading. I mean we don't track any of that 'automatically'. We only track it if you hook it up to a step output or input, right?

## Nits

- just start with python 3.10 instead of 3.9+ since 3.9 is not long for this world (deprecated next month)
- In the 'What You'll Learn' in the README.md, the 'After chapter 1' etc stuff doesn't get properly laid out / rendered in gitbook. Needs fixing.
- It's not "the OpenAI SDK" but "the OpenAI Agents SDK"
- update 'gpt-4' anywhere / everywhere to 'gpt-5'
- obviously the images need zurification

PR #3905, opened 2025-08-26
Bump allowed versions of click

strickvl:
Our upper ceiling on the versions of `click` we allow means that we can't, for example, install google's ADK agent framework in the same environment as ZenML.

Following how we did it in #3445 I'm just bumping the upper ceiling of click to allow for more wiggle room.


schustmi (approved):

strickvl:
Merging this in but letting CI continue to run in case any errors pop up ahead of the release.

PR #3913, opened 2025-09-01
Remove Google ADK from agent examples list

strickvl:
Google ADK requires `cloudpickle>3` whereas `zenml>0.84.3` only supports `cloudpickle<3` so removing this (again) for now.


htahir1:
id keep it in and just wait for next week where itll wrk

strickvl:
No longer needed as we now support `cloudpickle>3` following #3914 being merged in.

PR #3912, opened 2025-08-27
Add Google ADK agent features and update documentation

strickvl:
## Summary
- Add new Google ADK (Agent Development Kit) integration example for Gemini-powered agents
- Update all agent framework examples to use consistent DockerSettings with UV package installer
- Add proper environment variable handling for API keys across all examples

## Key Changes
- **New Google ADK Integration**: Complete example showcasing Gemini-powered agents with tool calling capabilities
- **Standardized Docker Configuration**: All agent examples now use UV package installer for faster dependency resolution
- **Environment Variable Management**: Consistent API key handling across all agent framework integrations
- **Documentation Updates**: Updated main README to include Google ADK in the agent frameworks table

## Test Plan
- [ ] Verify Google ADK example runs successfully with valid API key
- [ ] Confirm all existing agent examples still work with new DockerSettings
- [ ] Test UV package installer performance improvements

## Files Modified
- `examples/agent_framework_integrations/README.md` - Added Google ADK entry
- `examples/agent_framework_integrations/google_adk/` - New complete integration example
- Multiple `run.py` files - Added consistent DockerSettings configuration

This enhances the agent framework integrations with a new powerful option while standardizing the development experience across all examples.

PR #3914, opened 2025-09-01
Allow cloudpickle>3.x

strickvl:
Looking at the [cloudpickle release notes](https://github.com/cloudpipe/cloudpickle/blob/master/CHANGES.md) it doesn't seem that going from 2.x -> 3.x was the result of breaking changes.

Loosening this cloudpickle dependency would mean we could install Google ADK in the same env as ZenML (see #3913 and #3912 for more on my attempts to do this). I tested and it is currently possible to install `zenml>=0.84.3` together with `google-adk>=1.6.1` when we relax this cloudpickle requirement.


htahir1:
@strickvl the only thing we should test is if we use cloud pickle 2.0 and try to load an artifact with 3.0 what happens. As this might break previous pipeline run artifacts?

strickvl:
> @strickvl the only thing we should test is if we use cloud pickle 2.0 and try to load an artifact with 3.0 what happens. As this might break previous pipeline run artifacts?

Tested this. Ran a pipeline with `int`, `pd.DataFrame` and `bool` outputs where `cloudpickle` 2.x was installed (and an S3 artifact store). Then I loaded the artifacts with 3.x installed and they all loaded fine. I'll merge this in once the tests pass.

PR #3922, opened 2025-09-03
Airflow 3

schustmi:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/airflow/orchestrators/dag_generator.py
+++ b/src/zenml/integrations/airflow/orchestrators/dag_generator.py
@@ -225,7 +225,7 @@ else:
         dag_id=dag_config.id,
         is_paused_upon_creation=False,
         tags=dag_config.tags,
-        schedule_interval=dag_config.schedule,
+        schedule=dag_config.schedule,
         start_date=dag_config.start_date,
         end_date=dag_config.end_date,
         catchup=dag_config.catchup,


src/zenml/integrations/airflow/orchestrators/dag_generator.py
```diff
@@ -225,7 +225,7 @@ def get_kubernetes_pod_operator_init_kwargs(
         dag_id=dag_config.id,
         is_paused_upon_creation=False,
         tags=dag_config.tags,
-        schedule_interval=dag_config.schedule,
+        schedule=dag_config.schedule,
```
bcdurak, line 228: Do changes like this break our support for Airflow 2? Same question about the changes in the docs as well (like the pydantic update).

schustmi: Yes, this is a breaking change

htahir1: @schustmi in this case can we make it so that there is an if-else condition for this? Airflow 2.0 is really still popular

schustmi: We don't do this for any integration. It would be very easy if our integrations were separate packages. I can implement it here as an exception, but not a fan.

htahir1: yes i think here we need an exception @schustmi but not a general rule

schustmi: Actually just checked and this arg already exists in airflow 2.4 which we previously supported. So it should continue working as-is.

My docs changes only define a recommended version, so everything should be good as-is I think.

PR #3923, opened 2025-09-04
Prevent stopping runs without orchestrator run id

schustmi:
## Describe changes
When stopping a run that does not yet have an orchestrator run id, our orchestrators don't actually know what to delete. Additionally, when the first step container is in the process of starting while the run is being stopped, the following happens:
- The step starts, and tries to replace the placeholder run (which essentially means setting the orchestrator run id)
- There is no placeholder run anymore (status = `INITIALIZING`), as the status of the run was already updated to `STOPPING`.
- The step therefore assumes there is no placeholder run, and instead creates a new run.
- There are now two runs
  - one in `STOPPING` state without any steps
  - another one in `RUNNING` state with a single step, which will (depending on the orchestrator) never finish

This PR fixes this by not allowing runs without an orchestrator run ID to be cancelled.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
@@ -57,7 +57,11 @@ from zenml.integrations.kubernetes.orchestrators.manifest_utils import (
 )
 from zenml.logger import get_logger
 from zenml.logging.step_logging import setup_orchestrator_logging
-from zenml.models import PipelineDeploymentResponse, PipelineRunResponse
+from zenml.models import (
+    PipelineDeploymentResponse,
+    PipelineRunResponse,
+    PipelineRunUpdate,
+)
 from zenml.orchestrators import publish_utils
 from zenml.orchestrators.step_run_utils import (
     StepRunRequestFactory,
@@ -257,7 +261,12 @@ def main() -> None:
     else:
         orchestrator_run_id = orchestrator_pod_name
         if args.run_id:
-            pipeline_run = client.get_pipeline_run(args.run_id)
+            pipeline_run = client.zen_store.update_run(
+                run_id=args.run_id,
+                run_update=PipelineRunUpdate(
+                    orchestrator_run_id=orchestrator_run_id
+                ),
+            )
         else:
             pipeline_run = create_placeholder_run(
                 deployment=deployment,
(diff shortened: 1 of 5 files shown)


bcdurak (approved):
The changes look good. 

Just one question out of curiosity, do we even have disable the button on the frontend as you suggested in the TODOs? As far as I can see, we update the pipeline run to have an orchestrator ID before any step can possibly run and claim it, right? Your solution makes it impossible for the scenario (that you described in the PR) to happen.

schustmi:
> The changes look good.
> 
> Just one question out of curiosity, do we even have disable the button on the frontend as you suggested in the TODOs? As far as I can see, we update the pipeline run to have an orchestrator ID before any step can possibly run and claim it, right? Your solution makes it impossible for the scenario (that you described in the PR) to happen. Maybe I am missing something.

Yes we do. First of all, we only claim it once the orchestrator pod starts, which might be 5 mins after you ran the pipeline. Second of all, this is just for the Kubernetes orchestrator and doesn't happen for any other orchestrator.

bcdurak:
> Yes we do. First of all, we only claim it once the orchestrator pod starts, which might be 5 mins after you ran the pipeline. Second of all, this is just for the Kubernetes orchestrator and doesn't happen for any other orchestrator.

I mean, in those 5 minutes (for any orchestrator), even if they try to stop, they will not be able to, since the run wouldn't be claimed by any orchestrator or step environment by this point. They will have no orchestrator run id and now it will impossible to stop a run a without the orchestrator run ID. But you are also right, I guess UX-wise this would be a nice touch as well.

schustmi:
> > Yes we do. First of all, we only claim it once the orchestrator pod starts, which might be 5 mins after you ran the pipeline. Second of all, this is just for the Kubernetes orchestrator and doesn't happen for any other orchestrator.
> 
> I mean, in those 5 minutes (for any orchestrator), even if they try to stop, they will not be able to, since the run wouldn't be claimed by any orchestrator or step environment by this point. They will have no orchestrator run id and now it will impossible to stop a run a without the orchestrator run ID. But you are also right, I guess UX-wise this would be a nice touch as well.

Yes, but the stopping will put the status of the run as `stopping`. Which will cause the steps to create a second run, because no placeholder run for it exists anymore. Which is the whole point of this PR, to prevent this.

bcdurak:
> Yes, but the stopping will put the status of the run as `stopping`. 

Isn't this impossible without the run having an `orchestrator_run_id` now? 😅

schustmi:
> > Yes, but the stopping will put the status of the run as `stopping`.
> 
> Isn't this impossible without the run having an `orchestrator_run_id` now? 😅

I don't think I get the question. Yes it's not possible via the backend anymore, as I raise an exception there.
But your initial question was regarding the frontend button. Why would I not disable it if I already know that clicking it will raise an exception?

PR #3924, opened 2025-09-04
Add provisioning execution status

schustmi:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -596,8 +599,19 @@ def update(self, run_update: "PipelineRunUpdate") -> "PipelineRunSchema":
             The updated `PipelineRunSchema`.
         """
         if run_update.status:
-            self.status = run_update.status.value
-            self.end_time = run_update.end_time
+            if (
+                run_update.status == ExecutionStatus.PROVISIONING
```
bcdurak, line 603: As far as I can see:

- A run can only be in the `initializing` state when it is created as a placeholder run.
- A run can only be updated to the `provisioning` state from the `initializing` state by the orchestrator after submitting.

With that in mind, should we maybe implement a control mechanism here, making it impossible for our users to update running, failed pipelines (or any other states like that) to `initializing` or `provisioning`?

schustmi: Correct. I guess my PR introduced the `provisioning` state and also handles the case (= this can only be set if the run is still in initializing state, not from any other place).

I don't think it's necessary to handle the `initializing` state, as we never update a run to this status.

schustmi:
> One additional question. Do you think it makes sense to implement some sort of execution status pydantic model where we can combine the status with the status reason? This way we wouldn't have to do the same effort for the steps (or any other entity that would use the `ExecutionStatus`) in the future?

If I was writing this from scratch, yes I think it makes sense. Currently, the status is used in hundreds of places in both frontend and backend, so I would say its surely not worth the effort. Also, we often fetch the status column of run/steps to e.g. compute the updated run status, so keeping that as light as possible without requiring any json parsing etc makes sense I think.

PR #3959, opened 2025-09-15
Snapshot UX improvements

schustmi:
## Describe changes
Misc UX improvements for snapshots.
- Easier fetching using client method
- More filter options
- Use consistent naming of `Running a snapshot`

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -3306,69 +3306,69 @@ class Client(metaclass=ClientMetaClass):
                 hydrate=hydrate,
                 include_config_schema=include_config_schema,
             )
-        else:
-            list_kwargs: Dict[str, Any] = {
-                "named_only": None,
-                "project": project,
-                "hydrate": hydrate,
-                "size": 1,
-            }
 
-            # First, try to get the snapshot by name
-            snapshots = self.list_snapshots(
-                name=str(name_id_or_prefix),
-                pipeline=pipeline_name_or_id,
-                **list_kwargs,
-            )
-            if snapshots.total == 1:
-                snapshot = snapshots.items[0]
-            elif snapshots.total == 0:
-                # No name matches. If the user provided a pipeline, we assume
-                # they want to fetch the snapshot by name and fail. Otherwise,
-                # we try to fetch by ID prefix.
-                if pipeline_name_or_id or not allow_prefix_match:
+        list_kwargs: Dict[str, Any] = {
+            "named_only": None,
+            "project": project,
+            "hydrate": hydrate,
+            "size": 1,
+        }
+
+        # First, try to get the snapshot by name
+        snapshots = self.list_snapshots(
+            name=str(name_id_or_prefix),
+            pipeline=pipeline_name_or_id,
+            **list_kwargs,
+        )
+        if snapshots.total == 1:
+            snapshot = snapshots.items[0]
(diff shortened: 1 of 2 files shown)


src/zenml/client.py
```diff
@@ -3262,68 +3261,114 @@ def delete_trigger(
 
     def get_snapshot(
         self,
-        id_or_prefix: Union[str, UUID],
+        name_id_or_prefix: Union[str, UUID],
+        *,
+        pipeline_name_or_id: Optional[Union[str, UUID]] = None,
         project: Optional[Union[str, UUID]] = None,
+        include_config_schema: Optional[bool] = None,
+        allow_prefix_match: bool = True,
```
stefannica, line 3269: Just FYI introducing this with True default value changes the existing default behavior for all `get_snapshot` invocations throughout all existing code, because previously a fixed match would've been expected.

I checked all calls and, with a couple of exceptions, it's used only with UUIDs.

src/zenml/client.py
```diff
@@ -3262,68 +3261,114 @@ def delete_trigger(
...
+                hydrate=hydrate,
+                include_config_schema=include_config_schema,
+            )
+        else:
+            list_kwargs: Dict[str, Any] = {
+                "named_only": None,
+                "project": project,
+                "hydrate": hydrate,
+                "size": 1,
+            }
```
stefannica:
```suggestion
        list_kwargs: Dict[str, Any] = {
            "named_only": None,
            "project": project,
            "hydrate": hydrate,
            "size": 1,
        }
```

no else needed after return

src/zenml/utils/tag_utils.py
```diff
@@ -306,7 +306,7 @@ def add_tags(
             infer_artifact,
         ]
     ):
-        snapshot_model = client.get_snapshot(id_or_prefix=snapshot)
+        snapshot_model = client.get_snapshot(name_id_or_prefix=snapshot)
```
stefannica:
```suggestion
        snapshot_model = client.get_snapshot(name_id_or_prefix=snapshot, allow_prefix_match=False)
```

The current expectation here is that of a full match or nothing.

src/zenml/utils/tag_utils.py
```diff
@@ -674,7 +674,7 @@ def remove_tags(
...
```
stefannica:
```suggestion
        snapshot_model = client.get_snapshot(name_id_or_prefix=snapshot, allow_prefix_match=False)
```

The current expectation here is that of a full match or nothing.

src/zenml/client.py
```diff
@@ -3447,7 +3492,7 @@ def update_snapshot(
 
     def delete_snapshot(
         self,
-        id_or_prefix: str,
+        id_or_prefix: Union[str, UUID],
```
stefannica: this should be `name_id_or_prefix`, no ?

stefannica (changes requested):

src/zenml/client.py
schustmi: Yep, also adjusted it in the update method

src/zenml/client.py
schustmi: I removed the indentation of the rest of the method

stefannica (approved):

PR #3972, opened 2025-09-18
Track pipeline output spec and schema

schustmi:
## Describe changes
Stores the pipeline outputs as well as a JSON schema of the pipeline outputs as part of the compiled pipeline spec.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/compiler.py
+++ b/src/zenml/config/compiler.py
@@ -27,7 +27,7 @@ from typing import (
     Type,
 )
 
-from pydantic import ConfigDict, create_model, BaseModel
+from pydantic import BaseModel, ConfigDict, create_model
 
 from zenml import __version__
 from zenml.config.base_settings import BaseSettings, ConfigurationLevel
@@ -653,7 +653,11 @@ class Compiler:
             )
             for output_artifact in pipeline._output_artifacts
         ]
-        output_schema = compute_pipeline_output_schema(pipeline=pipeline)
+        try:
+            output_schema = compute_pipeline_output_schema(pipeline=pipeline)
+        except Exception as e:
+            logger.warning("Failed to compute pipeline output schema: %s", e)
+            output_schema = None
 
         return PipelineSpec(
             steps=step_specs,
@@ -666,7 +670,7 @@ class Compiler:
 
 def compute_pipeline_output_schema(
     pipeline: "Pipeline",
-) -> Optional[Dict[str, Any]]:
+) -> Dict[str, Any]:
     """Computes the pipeline output schema.
 
     Args:
@@ -675,9 +679,6 @@ def compute_pipeline_output_schema(
     Returns:
         The pipeline output schema.
     """
-    if not pipeline._output_artifacts:
-        return None
-
     output_model_class: Type[BaseModel] = create_model(
         f"{pipeline.name}_output",
         __config__=ConfigDict(arbitrary_types_allowed=True),
--- a/src/zenml/config/compiler.py
+++ b/src/zenml/config/compiler.py
@@ -689,7 +689,7 @@ def compute_pipeline_output_schema(
         )
 
     output_model_class: Type[BaseModel] = create_model(
-        f"{pipeline.name}_output",
+        "PipelineOutput",
         __config__=ConfigDict(arbitrary_types_allowed=True),
         **{
             _get_schema_output_name(output_artifact): (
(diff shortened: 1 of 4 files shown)


src/zenml/config/compiler.py
```diff
@@ -643,12 +645,50 @@ def _compute_pipeline_spec(
...
+        The pipeline output schema.
+    """
+    if not pipeline._output_artifacts:
+        return None
+
+    output_model = create_model(
+        f"{pipeline.name}_output",
+        __config__=ConfigDict(arbitrary_types_allowed=True),
+        **{
+            output_artifact.output_name: (
```
schustmi: @stefannica @safoinme How should we handle the case where a pipeline returns multiple step outputs, all of which have the same name?
- Should the pipeline output names be something like `<step-name>_<output_name>`?
- Should I just add suffixes to avoid duplicates, e.g. `output`, `output_2`, ...
- Something else?

src/zenml/config/compiler.py
```diff
@@ -643,12 +646,50 @@ def _compute_pipeline_spec(
...
+                output_artifact.annotation.resolved_annotation,
+                ...,
+            )
+            for output_artifact in pipeline._output_artifacts
+        },
+    )
+    return output_model_class.model_json_schema(mode="serialization")
```
schustmi: @stefannica @safoinme And a second question here: How should we treat this if the output's contain a type that is not JSON serializable? Don't store any output schema at all I assume?

src/zenml/config/compiler.py
safoinme: i'm more leaning more towards `<step-name>_<output_name>`

src/zenml/config/compiler.py
schustmi: As discussed, the schema will be set to `None` if the schema generation failed.

src/zenml/config/compiler.py
schustmi: Did this with some added logic to guarantee it's unique. But now that I think about it, for you it's probably easier if it actually uses the real step name and output name, so you can easier match those to the actual step outputs, correct? I can also remove my `.replace("-", "_")` calls if you feel that's better.

src/zenml/config/compiler.py
```diff
@@ -643,12 +647,59 @@ def _compute_pipeline_spec(
...
+            output_schema=output_schema,
+            source=pipeline.resolve(),
+            parameters=pipeline._parameters,
+        )
+
 
-        return PipelineSpec(steps=step_specs, **additional_spec_args)
+def compute_pipeline_output_schema(
+    pipeline: "Pipeline",
+) -> Dict[str, Any]:
```
stefannica: Isn't this better placed as a Pipeline method ?

src/zenml/config/compiler.py
```diff
@@ -643,12 +647,59 @@ def _compute_pipeline_spec(
...
+
+    def _get_schema_output_name(output_artifact: "StepArtifact") -> str:
+        return (
+            output_artifact.invocation_id.replace("-", "_")
+            + "-"
+            + output_artifact.output_name.replace("-", "_")
+        )
+
+    output_model_class: Type[BaseModel] = create_model(
+        f"{pipeline.name}_output",
```
stefannica:
```suggestion
        "PipelineOutput",
```

Please don't use a generated name here, it's useless. We don't use one for the input model either.

src/zenml/config/compiler.py
```diff
@@ -643,12 +647,59 @@ def _compute_pipeline_spec(
...
+        __config__=ConfigDict(arbitrary_types_allowed=True),
+        **{
+            _get_schema_output_name(output_artifact): (
+                output_artifact.annotation.resolved_annotation,
```
stefannica: What's the reason for using `.resolved_annotation` here? I'm asking because we're not using that for input parameters in `Pipeline.get_parameters_model` (and maybe we should?).

stefannica: Nevermind, this question is a result of my lack of information.

schustmi: `output_artifact.annotation` is not actually the real type annotation, instead it's an object of type `OutputSignature` (one of our pydantic models), which contains the actual user type annotation in the `resolved_annotation` attribute

src/zenml/config/compiler.py
schustmi: I'm undecided, I've been trying to use the pipeline class as more of a user-facing object with only relevant methods that users should call, but I'm happy to also move it there if you think it makes more sense.

stefannica (approved):

PR #3977, opened 2025-09-18
New quickstart

htahir1:
## Describe changes

## Pre-requisites

## Types of changes


examples/quickstart/configs/agent.yaml
```diff
@@ -0,0 +1,7 @@
+settings:
+  docker:
+    requirements: requirements.txt
+    parent_image: zenmldocker/zenml:0.85.0-py3.12
+    platform: linux/amd64
```
strickvl, line 5: why are you specifying the platform here? probably not needed?

strickvl (changes requested):
First pass review:

- that global `_router` object + that `ClassifierManager` I don't think we want to highlight that as a good practice. Isn't there some nicer way of handling that? Like that hook where we `_load_production_classifier_if_any` seems like not a best practice. Ideally I think I'd want that classifier loaded at the start of the inference step instead of using the hook and adding it to global state.
- adding a simple eval (i.e. comparison between the two modes) step would be really nice + bring the whole story home. include a confusion matrix visualisation + some viz of accuracy / F1 scores.
- maybe you could visualise the latency of calls as well? you'd have all the data I think to be able to do that.
- For sure add more examples to the dataset.
- do we want to bind all this together into the Model Control Plane?
- Do we want to add basic experiment tracking? Could we showcase the comparison between the two pipelines' evals in the dashboard instead of experiment tracking (assuming we've written a basic eval)
- probably too much try/excepts in here. I'm not sure that's a good pattern for ZenML pipeline. Maybe in the quickstart we want it to work for people, but if you catch all of the exceptions then people maybe lose track of what's going on and that pipelines can fail?

PR #3982, opened 2025-09-22
Migrate pyproject.toml to project structure

schustmi:
## Describe changes
This PR updates our pyproject.toml to use the `project` section instead of using `tool.poetry`. This enables two things that are useful for us:
- In the previous structure, it was impossible to install a dependency (e.g. `rich`) without extras in the base package, while having e.g. `rich[jupyer]` in our own `jupyer` extra. Because of that, we would have to duplicate/maintain the dependencies of that rich extra ourselves
- In the previous structure, an extra couldn't depend on another extra of the same package. That way, we would have to duplicate requirements in multiple extras (e.g. DB dependencies in the `local` and `server` extra. In the future, our `server` extra could simply require the `local` extra.

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ for reference.

## Pre-requisites

## Types of changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,6 @@ dependencies = [
     "packaging>=24.1",
     "psutil>=5.0.0",
     "pydantic>=2.0,<=2.11.9",
-    "pydantic-settings",
     "python-dateutil>=2.8.1,<3.0.0",
     "pyyaml>=6.0.1",
     "rich>=12.0.0", # Make sure to keep the version aligned with our `jupyter` extra
--- a/docs/book/getting-started/hello-world.md
+++ b/docs/book/getting-started/hello-world.md
@@ -82,7 +82,7 @@ The fastest way to create a cloud stack is through the **Infrastructure-as-Code*
 
 You'll need:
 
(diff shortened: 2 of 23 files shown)


pyproject.toml
```diff
@@ -1,15 +1,15 @@
-[tool.poetry]
+[build-system]
+requires = ["uv_build >= 0.8.17, <0.9.0"]
+build-backend = "uv_build"
```
bcdurak, line 3: I see that we now use the uv_build as a build backend. I was just wondering how the packaging is done. Previously, we had the `exclude` section for poetry and manually point it to `src/`. Do we need something like that here as well?

bcdurak: Also that reminds me, we had this little trick for the CLI output rerouting. With the new format is this still possible?

schustmi: I checked and none of the exclude/include is necessary, it detects the needed files by itself.

As for the CLI output: I haven't tested, but I'm sure it'll be possible somehow. If not, we can also switch back to the old build backend with this new structure

PR #3990, opened 2025-09-23
Feature:3973 Extend list pipeline filters

Json-Andriopoulos:
## Describe changes
- Filter pipelines by the name or ID of the latest executor user

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/models/v2/core/pipeline.py
+++ b/src/zenml/models/v2/core/pipeline.py
@@ -21,6 +21,7 @@ from typing import (
     Optional,
     Type,
     TypeVar,
+    Union,
 )
 from uuid import UUID
 
@@ -268,6 +269,7 @@ class PipelineFilter(ProjectScopedFilter, TaggableFilter):
         *ProjectScopedFilter.FILTER_EXCLUDE_FIELDS,
         *TaggableFilter.FILTER_EXCLUDE_FIELDS,
         "latest_run_status",
+        "latest_run_user",
     ]
     CLI_EXCLUDE_FIELDS: ClassVar[List[str]] = [
         *ProjectScopedFilter.CLI_EXCLUDE_FIELDS,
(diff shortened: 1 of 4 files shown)


src/zenml/models/v2/core/pipeline.py
```diff
@@ -313,28 +334,39 @@ def apply_filter(
...
+                    PipelineRunSchema.pipeline_id
+                    == latest_pipeline_run_subquery.c.pipeline_id,
+                    PipelineRunSchema.created
+                    == latest_pipeline_run_subquery.c.created,
+                ),
+            )
+
+            if self.last_executor_name_or_id:
+                query = query.join(
+                    UserSchema, UserSchema.id == PipelineSchema.user_id
```
schustmi: I think this should probably be `PipelineRunSchema.user_id` to filter for the user that created the run? Currently, this would filter pipelines by the user that created the pipeline

src/zenml/models/v2/core/pipeline.py
```diff
@@ -284,6 +285,22 @@ class PipelineFilter(ProjectScopedFilter, TaggableFilter):
         "This will always be applied as an `AND` filter for now.",
     )
 
+    last_executor_name_or_id: Optional[str] = Field(
```
schustmi: In the response model above (`PipelineResponseResources`), we call this `latest_run_user`. I think it's probably best if we also reuse that name here for consistency?

Json-Andriopoulos: Good point. `latest_run_user_name_or_id` or `latest_run_user`? I think we have the `name_or_id` naming convention in other places we also do user filtering.

schustmi: I prefer `latest_run_user`, I think in most of our filter objects (which are also used to define the parameters of the endpoints) use this without the `name_or_id` suffix.

PR #4002, opened 2025-09-25
Fixing race conditions for tags

bcdurak:
## Describe changes
A quick version to the tag race condition.

## Pre-requisites

## Types of changes


src/zenml/client.py
```diff
@@ -8141,22 +8141,37 @@ def list_tags(
...
-            tag_name_or_id: name or id of the tag to be attached.
+            tag: name or id of the tag to be attached.
             resources: the resources to attach the tag to.
         """
-        if isinstance(tag_name_or_id, str):
-            try:
-                tag_model = self.create_tag(name=tag_name_or_id)
-            except EntityExistsError:
-                tag_model = self.get_tag(tag_name_or_id)
+        if isinstance(tag, str):
```
schustmi, line 8636: This should IMO use `uuid_utils.is_valid_uuid` to catch UUIDs passed as string. We do this in most client methods already I think

bcdurak: The docstring was misleading but this was just handling strings and `Tag` objects for now. In our codebase, the only place we are using this is the `add_tags` function which works with strings and `Tag`s as well.

bcdurak: I can add a UUID to the Union though if you think that we need to handle that case as well.

schustmi: Fine with me if its just names/tag objects

bcdurak: Fixed the docstring.

schustmi (approved):

src/zenml/client.py
```diff
@@ -8547,13 +8548,33 @@ def get_tag(
...
+            KeyError: If the tag is not found.
+
         Returns:
             The tag of interest.
         """
-        return self.zen_store.get_tag(
-            tag_name_or_id=tag_name_or_id,
-            hydrate=hydrate,
-        )
+        if is_valid_uuid(tag_name_or_id):
```
schustmi: Any reason we can't use the helper methods that most other get methods use to solve this fetching by name/ID?

bcdurak: Nope, I have changed it. Had to adjust some of the methods to use the correct `allow_name_prefix_match` value but it should be done now.

PR #4011, opened 2025-09-29
Feature:4005 Docker settings usage warnings

Json-Andriopoulos:
## Describe changes

## Pre-requisites

## Types of changes


strickvl:
It's a somewhat common thing to switch between a local orchestrator and a cloud-based one, esp during development. I'm wondering if this warning might get a bit annoying?

htahir1:
100% i wouldnt raise usage warnings for this due to what alex said

bcdurak:
@strickvl @htahir1, I think I disagree with this a little bit. If you are working with a pipeline where you define some `DockerSettings` and the orchestrator is currently ignoring it, I would want to see the warning.

bcdurak:
It is an indicator of how to do it properly. If I am in the development setting, I would not be bothered by this if I switch to a local orchestrator.

htahir1:
I think we have talked a few times how overloading warnings for use-cases that dont matter causes people to just ignore them @bcdurak ... I will guarantee this will cauase one of two things a) Most people will ignore this warning b) People will start having if conditions in their code to not apply dockersettings on certain stacks, which is ugly AF

I am quite strongly opposed to this PR tbh

schustmi:
@htahir1 @Json-Andriopoulos "a) Most people will ignore this warning": This is not an argument against having a warning at all. I think ZenML should warn users about stuff that is worthy of a warning. If they decide to ignore them and run into issues because of it, that's on them and warning them is the best we could do.

Regarding what we should do in this case:
- When running a pipeline with an "invalid settings key" (e.g. `settings={"not_docker": DockerSettings(...)}` we currently log an info message: https://github.com/zenml-io/zenml/blob/e0252935dc99a29a08f7ef77f3e3fccf4cfa5f4d/src/zenml/config/compiler.py#L400-L403 This case includes when you specify settings for a component or flavor that is not in your active stack when running.
- When providing some keys for a settings class which don't exist (e.g. (e.g. `settings={"docker": DockerSettings(invalid_key=...)}` we currently log a warning message: https://github.com/zenml-io/zenml/blob/e0252935dc99a29a08f7ef77f3e3fccf4cfa5f4d/src/zenml/config/compiler.py#L413-L417

The case we're trying to catch as part of this PR is probably more similar to the first bullet point, so maybe changing it to an info log would be the most inline with how we treat it currently? But in my opinion, both the first bullet point here and the message as part of this PR logically are warning messages.

htahir1:
@schustmi @Json-Andriopoulos 

Overall: the intent (alerting when `DockerSettings` are defined but won’t be honored) is good, but the UX needs to avoid becoming noisy—especially when devs bounce between local and remote orchestrators. Here are some suggestions:


1. **Warn only when it’s actionable**

   * Trigger **only if** `DockerSettings` are present **and** the current orchestrator is non-containerized (or otherwise ignores them).
   * If `DockerSettings` are empty/default → no warning.

2. **Make the message precise + helpful**

   * Example: *“`DockerSettings` detected, but `<LocalOrchestrator>` does not build/use containers. Switch to `<ContainerizedOrchestrator>` or remove `DockerSettings`. See: <docs path>.”*
   * Include the exact pipeline/run and the settings keys detected.

3. **Right severity in dev vs. prod**

   * Default to **INFO** for known “local/dev” orchestrators.
   * **WARNING** when running on remote/cloud where users are more likely to expect containerization.

4. **De-dupe & throttle**

   * One warning per process/run (or per unique pipeline+orchestrator combo), with “suppressed N similar messages” thereafter.

5. **Give users control**

   * Add a project/user config toggle or an env var to silence.
   * Consider `--strict` mode to escalate to error/exit for CI.

6. **Surface it where it matters**

   * CLI log + dashboard run details (a small banner with “settings ignored”).
   * Optionally emit a distinct log code (e.g., `ZNL-DS001`) so teams can grep/alert.

7. **Test the matrix**

   * Cases: (a) `DockerSettings` + local orchestrator, (b) `DockerSettings` + containerized orchestrator, (c) no `DockerSettings`, (d) partial support (some fields ignored), (e) multi-stack projects switching orchestrators.

This balances the concerns in the thread: folks worried about annoyance when switching locally (valid) and those wanting a clear heads-up when a real misconfiguration happens (also valid).

I'm happy if at least first three of the above are taken into account

schustmi:
@htahir1 @Json-Andriopoulos IMO we could simply add a warning log message in the compiler class (in the place I linked above) if there are non-empty docker settings defined when running with a local orchestrator?

src/zenml/utils/warnings/registry.py
```diff
@@ -0,0 +1,59 @@
...
+
+from zenml.utils.warnings.base import (
+    WarningCategory,
+    WarningConfig,
+    WarningSeverity,
+    WarningVerbosity,
+)
+
+
+class WarningCodes(str, Enum):
```
schustmi: We have a `StrEnum` base class that we usually use in those cases, has some additional properties that you might/might not need :)

src/zenml/utils/warnings/controller.py
```diff
@@ -0,0 +1,162 @@
...
+        """Method to execute warning handling logic with warning log level.
+
+        Args:
+            warning_code: The code of the warning (see WarningCodes enum)
+            message: The message to display.
+            **kwargs: Keyword arguments (for formatted messages).
+        """
+        self._log(warning_code, message, LoggingLevels.WARNING, **kwargs)
+
+    def info(self, *, warning_code: str, message: str, **kwargs: Any) -> None:
```
schustmi:
```suggestion
    def info(self, *, warning_code: WarningCodes, message: str, **kwargs: Any) -> None:
```
We can use the enum here, no? Same in other places

src/zenml/utils/warnings/controller.py
```diff
@@ -0,0 +1,162 @@
...
+logger = logging.getLogger(__name__)
+
+
+class WarningController(metaclass=SingletonMetaClass):
+    """Class responsible for centralized handling of warning messages."""
+
+    def __init__(self) -> None:
+        """WarningController constructor."""
+        self._warning_configs: dict[str, WarningConfig] = {}
+        self._warning_statistics: dict[str, int] = defaultdict(int)
```
schustmi: There's also `from collections import Counter` which we could use here, I'll leave it up to you.

Has some helper methods to e.g. count the total amount of warnings, in case we ever want to log some statistics.

src/zenml/utils/warnings/base.py
```diff
@@ -0,0 +1,57 @@
...
+    LOW = "low"
+    MEDIUM = "medium"
+    HIGH = "high"
+
+
+class WarningConfig(BaseModel, use_enum_values=True):
+    """Warning config class describing how warning messages should be displayed."""
+
+    description: str = Field(description="Description of the warning message")
+    severity: WarningSeverity = WarningSeverity.MEDIUM
```
schustmi, line 47: Seems like the severity is unused at the moment?

src/zenml/config/compiler.py
```diff
@@ -369,6 +371,46 @@ def _verify_upstream_steps(
...
+            ContainerizedOrchestrator,
+            LocalOrchestrator,
+        )
+
+        if not docker_settings or isinstance(
+            stack.orchestrator, ContainerizedOrchestrator
+        ):
+            return
+
+        warning_message = (
```
schustmi, line 392: We do actually have a few orchestrators that are not containerized but still use some DockerSettings attributes to configure their behaviour (`WheeledOrchestrator`, `LightningOrchestrator`). So we either need to get rid of this warning here or also include those two

src/zenml/utils/warnings/base.py
Json-Andriopoulos: Yes, it is placeholder at the moment. On follow-up implementation it can be used in combination with env vars or CLI parameters to filter out warnings. E.g. strict mode could display all warnings but other modes may want to keep MEDIUM+.

src/zenml/config/compiler.py
Json-Andriopoulos: Good point. Do we have any production orchestrators that don't make use of the docker settings? Seems like we just need to catch the LocalOrchestrator here and info the message 🤔

src/zenml/utils/warnings/controller.py
Json-Andriopoulos: Yeap, good idea. I mainly want to avoid constructing values if the key is missing, Counter supports that as well.

src/zenml/utils/warnings/controller.py
Json-Andriopoulos: Good point!

src/zenml/utils/warnings/base.py
schustmi: I like that, good idea!

src/zenml/config/compiler.py
schustmi: Yeah I think just the local one ignores them entirely

schustmi (approved):

PR #4013, opened 2025-09-29
Follow up fixes for deployments

schustmi:
## Describe changes
- Cleans up some of the in-memory materializer code
- Uses correct artifact types when using the in-memory materializer (from the original materializer that would have been used to store the artifact to disk)
- Cleans up the `deployment_snapshot_request_from_source_snapshot` function and fixes the source snapshot ID
- Fixes in memory storage to allow `None` values
- Fix parameter values not included in response
- Some cosmetic changes for the deployment CLI
- Fix for available port check on MacOS

## Pre-requisites

## Types of changes


src/zenml/zen_stores/schemas/deployment_schemas.py
```diff
@@ -157,8 +157,8 @@ def get_query_options(
         if include_resources:
             options.extend(
                 [
-                    joinedload(jl_arg(DeploymentSchema.user)),
-                    joinedload(jl_arg(DeploymentSchema.deployer)),
+                    selectinload(jl_arg(DeploymentSchema.user)),
+                    selectinload(jl_arg(DeploymentSchema.deployer)),
```
stefannica, line 161: Ok, but why ?

schustmi: When listing deployments, the user or deployer can often be the same for multiple items in the response. When using a joined load, the same row will be joined and included in the DB result multiple times. Selectinload instead first fetches all the deployments, and then does a second DB query to fetch all the unique users/deployers/...

Issue #4007, opened 2025-09-26
Deployment details: Show custom visualizations


safoinme:
### Feature Description

Depends a bit on how this is supposed to work with the backend.

Backend assumptions:
* included: support to associate visualizations with artifacts.
* not Included: support for “naming visualizations”
* this works via the existing support to load visualizations from the artifact store in the dashboard

### Priority

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #4016, opened 2025-09-29
Deployment custom visualizations

safoinme:
## Describe changes
- extend the deployment domain models and API surface to represent curated deployment visualizations, including request/update/response models and filter support.
- persist curated visualizations by introducing the corresponding SQLModel schema, Alembic migration, and REST/SQL store logic, and wiring the new routes into the FastAPI server.
- expose client helpers so SDK users can add, list, update, and delete the curated visualizations associated with a deployment while keeping responses hydrated with attached artifacts.

## Why
- product teams need a first-class way to curate which artifact visualizations show up alongside a deployment so dashboards can surface the right insights without manual filtering.
- storing these curation decisions in the backend ensures they survive across sessions, are shareable across users, and can participate in RBAC/project scoping.
- wiring the API through the client and stores keeps the SDK in sync with the server capabilities, enabling immediate consumption by CLI/UI flows.

This is related to https://github.com/zenml-io/zenml/issues/4007 and the Custom visualisation ticket in Notion 

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/models/v2/core/curated_visualization.py
+++ b/src/zenml/models/v2/core/curated_visualization.py
@@ -129,7 +129,6 @@ class CuratedVisualizationResponseBody(ProjectScopedResponseBody):
         ),
     )
     artifact_version_id: UUID = Field(
-        default=None,
         title="The artifact version ID.",
         description=(
             "Identifier of the artifact version that owns the curated visualization. "


schustmi:
@safoinme Can you write a description for this PR please, and include why we need all of this?

src/zenml/zen_stores/schemas/deployment_schemas.py
```diff
@@ -220,6 +237,14 @@ def to_model(
                 pipeline=self.snapshot.pipeline.to_model()
                 if self.snapshot and self.snapshot.pipeline
                 else None,
+                visualizations=[
+                    visualization.to_model(
+                        include_metadata=include_metadata,
+                        include_resources=include_resources,
```
schustmi: These should both be false to avoid huge responses. This is a convention that is used in all our current responses: If something is in the `resources` part of a response, it will only include a body, not metadata or resources.

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,288 @@
...
+        return CuratedVisualizationResponse(
+            id=self.id,
+            body=body,
+            metadata=metadata,
+            resources=response_resources,
+        )
+
+
+def curated_visualization_relationship_kwargs(
+    parent_column_factory: Callable[[], Any],
```
stefannica: why a callable here instead of an actual column ?

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,369 @@
...
+    resources: List[CuratedVisualizationResource] = Field(
+        title="Resources associated with this visualization.",
+        description=(
+            "List of resources (deployments, models, pipelines, pipeline runs, "
+            "pipeline snapshots, projects) that should surface this visualization. "
+            "Must include at least one resource. Multiple resources of "
+            "different types can be specified to reuse visualizations "
+            "across the ML workflow."
+        ),
+    )
```
stefannica: This model doesn't seem the right way to represent curated visualization from a UX perspective. There's "display name" and "display order" in here, but at the same time multiple target resources.

The way this works is each resource has its own display order and display name, so grouping them this way feels wrong.

I know this is inspired from the run metadata, but it's different there: all attributes in this model apply uniformly to all the linked resources.

stefannica: From my POV as a user, I find it that the more natural way to group these to implement a better UX is to have them reflect how I would visualize them in the dashboard: have a single resource and a multiple of `[artifact_version_id+visualization_index+display_name+display_order]` mappings. But that's probably overkill and you might want to use the KISS principle here and simply restrict that there is a single `resource` instead of a list of them.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5408,330 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+
+            resource_schema_map: Dict[
+                VisualizationResourceTypes, Type[BaseSchema]
+            ] = {
+                VisualizationResourceTypes.DEPLOYMENT: DeploymentSchema,
+                VisualizationResourceTypes.MODEL: ModelSchema,
+                VisualizationResourceTypes.PIPELINE: PipelineSchema,
+                VisualizationResourceTypes.PIPELINE_RUN: PipelineRunSchema,
+                VisualizationResourceTypes.PIPELINE_SNAPSHOT: PipelineSnapshotSchema,
+                VisualizationResourceTypes.PROJECT: ProjectSchema,
```
stefannica, line 5550: Be aware that you're still missing schema changed for most of these. All you have right now is models and deployments.

src/zenml/models/v2/core/curated_visualization.py
stefannica: The good news is: if you make `resource` a single entry, your SQL zen store implementation will be drastically simplified.

src/zenml/client.py
```diff
@@ -3739,6 +3745,264 @@ def get_deployment(
...
 
+    def create_curated_visualization(
+        self,
+        artifact_version_id: UUID,
+        visualization_index: int,
+        *,
+        resources: List[CuratedVisualizationResource],
+        project_id: Optional[UUID] = None,
+        display_name: Optional[str] = None,
+        display_order: Optional[int] = None,
```
stefannica: Same thing here: you would expect different resources to have different display name and display order values. Grouping them like this doesn't make sense.

However, I do see the need to group a bunch of visualizations at once with *a single resource*  as a batch operation (similar to tags). In this case, you could implement a `batch_create_` method in the zen store that even the dashboard might find useful to call, but this is totally optional. Right now, operations for a single resource would be more than enough.

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,373 @@
...
+        if sort_by in {"created", "updated", "visualization_index"}:
+            column = getattr(table, sort_by)
+            if operand == SorterOps.DESCENDING:
+                return cast(
+                    AnyQuery,
+                    query.order_by(desc(column), asc(table.id)),
+                )
+            return cast(AnyQuery, query.order_by(asc(column), asc(table.id)))
+
+        return super().apply_sorting(query=query, table=table)
```
stefannica: Why did you need to write custom sorting code for these fields? Shouldn't the base class cover them? I don't see anything special about their data types or their queries.

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,373 @@
...
+
+    sort_by: str = Field(
+        default="display_order",
+        description="Which column to sort by.",
+    )
+
+    artifact_version: Optional[UUID] = Field(
+        default=None,
+        description="ID of the artifact version associated with the visualization.",
+    )
```
stefannica:
```suggestion
    artifact_version: Optional[Union[UUID, str]]= Field(
        default=None,
        description="ID of the artifact version associated with the visualization.",
        union_mode="left_to_right",        
    )
```

you should allow for string here too (look at other filter models as examples).

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,373 @@
...
+            column = getattr(table, "display_order")
+            if operand == SorterOps.DESCENDING:
+                return cast(
+                    AnyQuery,
+                    query.order_by(desc(column).nulls_last(), asc(table.id)),
+                )
+            return cast(
+                AnyQuery,
+                query.order_by(asc(column).nulls_first(), asc(table.id)),
+            )
```
stefannica:
```suggestion
        if sort_by == "display_order":
            column = getattr(table, "display_order")
            if operand == SorterOps.DESCENDING:
                return cast(
                    AnyQuery,
                    query.order_by(desc(column).nulls_last()),
                )
            return cast(
                AnyQuery,
                query.order_by(asc(column).nulls_first()),
            )
```

you don't need these as tie-breakers. They are added by the base class.

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,373 @@
...
+                == self.visualization_index
+            )
+        if self.display_order is not None:
+            custom_filters.append(
+                getattr(table, "display_order") == self.display_order
+            )
+        if self.layout_size is not None:
+            custom_filters.append(
+                getattr(table, "layout_size") == self.layout_size
+            )
```
stefannica: Same here: why not let the base class deal with these fields ?

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,373 @@
...
+
+        Returns:
+            A list of custom filters.
+        """
+        custom_filters = super().get_custom_filters(table)
+
+        if self.artifact_version:
+            custom_filters.append(
+                getattr(table, "artifact_version_id") == self.artifact_version
+            )
```
stefannica: If you rename `artifact_version` to `artifact_version_id`, you also don't need this block either and can remove `get_custom_filter` completely.

src/zenml/zen_server/rbac/models.py
```diff
@@ -80,6 +83,30 @@ class ResourceType(StrEnum):
...
+        mapping = {
+            VisualizationResourceTypes.DEPLOYMENT: cls.DEPLOYMENT,
+            VisualizationResourceTypes.MODEL: cls.MODEL,
+            VisualizationResourceTypes.PIPELINE: cls.PIPELINE,
+            VisualizationResourceTypes.PIPELINE_RUN: cls.PIPELINE_RUN,
+            VisualizationResourceTypes.PIPELINE_SNAPSHOT: cls.PIPELINE_SNAPSHOT,
+            VisualizationResourceTypes.PROJECT: cls.PROJECT,
+        }
+        return mapping.get(visualization_type)
+
```
stefannica: If you look at the rest of this module and its code, you'll see that this doesn't belong here. There's nothing else this specific in the module and there are other models that resemble the curated visualizations (e.g. the run metadata, the tags) that would've warranted similar changes but didn't. This is a good indication that you're modeling this incorrectly.

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+    Args:
+        visualization: The curated visualization to create.
+
+    Returns:
+        The created curated visualization.
+    """
+    return verify_permissions_and_create_entity(
+        request_model=visualization,
+        create_method=zen_store().create_curated_visualization,
+    )
```
stefannica: Calling `verify_permissions_and_create_entity` only works if your "entity" is covered by the RBAC system. Right now the RBAC implementation has no idea what curated visualizations are (i.e. there is no `ResourceType.CURATED_VISUALIZATION` enum value) and has never seen your `CuratedVisualizationRequest` model before.

This is where you should ask yourself: is it worth it to add `ResourceType.CURATED_VISUALIZATION` as a new RBAC resource type? Adding an RBAC resource type is quite a lot of work and is high-impact, because:
* you need to update the Pro RBAC implementation to also reflect this new type of resource, including the dashboard.
* you need to update the default roles to include permissions for this type of resource
* you need to implement a database migration for it
* you also need to document it
* in addition to all that, you now expose the users to yet another resource type, which makes the RBAC feature even more complex

In your case, I would say you don't need to model another resource type. Even though we have resource types for tags and run metadata, this resource is too small to really warrant a change this big.

What you should do instead here is to check the UPDATE permission for the referenced resource (see the `create_run_metadata` endpoint for an example of that) and the READ permission for the referenced artifact version.

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Returns:
+        The updated curated visualization.
+    """
+    return verify_permissions_and_update_entity(
+        id=visualization_id,
+        update_model=visualization_update,
+        get_method=zen_store().get_curated_visualization,
+        update_method=zen_store().update_curated_visualization,
+    )
```
stefannica: Same here: check the UPDATE permission for the referenced resource (see the `create_run_metadata` endpoint for an example of that)  but don't call `verify_permissions_and_update_entity` here.

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+    """Delete a curated visualization.
+
+    Args:
+        visualization_id: The ID of the curated visualization to delete.
+    """
+    verify_permissions_and_delete_entity(
+        id=visualization_id,
+        get_method=zen_store().get_curated_visualization,
+        delete_method=zen_store().delete_curated_visualization,
+    )
```
stefannica: Same here: check the UPDATE permission for the referenced resource (see the create_run_metadata endpoint for an example of that).

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+            by including metadata fields in the response.
+
+    Returns:
+        The curated visualization with the given ID.
+    """
+    return verify_permissions_and_get_entity(
+        id=visualization_id,
+        get_method=zen_store().get_curated_visualization,
+        hydrate=hydrate,
+    )
```
stefannica: I would only check the READ permission for the referenced resource here (see the create_run_metadata endpoint for an example of that). I would not check the READ permission for the referenced artifact for the same reason discussed in the `list` endpoint.

Don't call `verify_permissions_and_get_entity` on the visualization model, because I think the RBAC response hydration will also enforce the READ permission for the referenced artifact version and remove it from the model if not present (@schustmi maybe can confirm).

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,189 @@
...
+        batch_verify_permissions_for_models(page.items, action=Action.READ)
+        return dehydrate_page(page)
+    else:
+        # Concrete resource type available - use standard RBAC flow
+        return verify_permissions_and_list_entities(
+            filter_model=visualization_filter_model,
+            resource_type=resource_type,
+            list_method=zen_store().list_curated_visualizations,
+            hydrate=hydrate,
+        )
```
stefannica: TL;DR: this entire endpoint is too complex to implement correctly when RBAC is involved and users really don't need this level of flexibility, so I'd strongly suggest that you get rid of the endpoint, the `CuratedVisualizationFilter` and all `list_curated_visualizations` methods in the zen store and Client.

---

Here's why this gets tricky. You have to ask yourself: "if I'm looking at a snapshot in the dashboard and I don't have permissions to see one of the artifacts listed in the curated visualizations, should I still be able to see it in the dashboard ?

This behavior also needs to be consistent with the other endpoints (`get` and `update`) and with the `visualizations` resource fields that you added to the various resource models that can reference visualizations. Right now, there is no filtering done for those `visualizations` fields, so you can effectively get all the visualizations attached to a resource if you have READ permission for that resource without checking any artifact version permissions.

One way to keep the implementation consistent is to keep it simple and follow the following convention throughout:

* you check artifact version READ permissions ONLY when someone tries to attach a visualization to a resource. You can think of this as someone "granting" artifact version READ access to everyone else who can already access the referenced resource (deployment/pipeline/etc) from that point on.
* for the rest of the lifecycle and endpoints, you no longer have to check the artifact version permissions for this curated visualization.

In effect, this answers "yes" to the question at the top: everyone who can access e.g. a Deployment will also be able to see its "linked" artifact visualizations without necessarily having access to the artifacts.

Given that you can already list all the curated visualizations attached to each resource type via their `resources` model fields, you don't really need this endpoint. We will lose things like pagination and sorting, but we can bring back the list endpoint and the filter later, if it turns out those are really needed (e.g. we don't know how many visualizations users will actually want to link to a resource).

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,243 @@
...
+        ondelete="CASCADE",
+        nullable=False,
+    )
+
+    visualization_index: int = Field(nullable=False)
+    display_name: Optional[str] = Field(default=None)
+    display_order: Optional[int] = Field(default=None)
+    layout_size: CuratedVisualizationSize = Field(
+        default=CuratedVisualizationSize.FULL_WIDTH, nullable=False
+    )
```
stefannica: please don't use enums in the schema. Use a string and make the conversion in the to_model/from_request/update methods. This is a pattern used throughout the code for good reasons that @bcdurak can explain better than me.

src/zenml/client.py
```diff
@@ -3739,6 +3746,285 @@ def get_deployment(
...
+    def add_visualization_to_deployment(
+        self,
+        deployment_id: UUID,
+        artifact_version_id: UUID,
+        visualization_index: int,
+        *,
+        display_name: Optional[str] = None,
+        display_order: Optional[int] = None,
+        layout_size: CuratedVisualizationSize = CuratedVisualizationSize.FULL_WIDTH,
+    ) -> CuratedVisualizationResponse:
```
stefannica: This is redundant if you have `create_curated_visualization`.

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
stefannica: when you fix this, don't forget to fix the migration too

src/zenml/models/v2/core/curated_visualization.py
safoinme: The main reason for this is display_order which can be null, so we still need to sort properly, even though some visualisations are null i assume base class can't handle ordering with null. 

But maybe the question is actually if display order can be null? especially that we are adding this visualisations in the dashboard in an actual order. @stefannica

src/zenml/zen_server/routers/curated_visualization_endpoints.py
safoinme: That's a great observation and point, i tried to apply this + the re-work on RBAC would love to get another review for this

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,278 @@
...
+    display_name: Optional[str] = Field(default=None)
+    display_order: Optional[int] = Field(default=None)
+    layout_size: str = Field(
+        default=CuratedVisualizationSize.FULL_WIDTH.value,
+        nullable=False,
+    )
+
+    artifact_version: Optional["ArtifactVersionSchema"] = Relationship(
+        sa_relationship_kwargs={"lazy": "selectin"}
+    )
```
stefannica: This can never be null, right ? Because the `artifact_version_id` field is itself non-nullable and auto-delete is configured, so if the artifact is deleted, the `CuratedVisualizationSchema` referencing it is also deleted automatically. 

```suggestion
    artifact_version: "ArtifactVersionSchema" = Relationship(
        sa_relationship_kwargs={"lazy": "selectin"}
    )
```

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,278 @@
...
+    resource: Optional[CuratedVisualizationResourceSchema] = Relationship(
+        back_populates="visualization",
+        sa_relationship_kwargs={
+            "lazy": "selectin",
+            "cascade": "all, delete",
+            "uselist": False,
+        },
+    )
```
stefannica:
```suggestion
    resource: CuratedVisualizationResourceSchema = Relationship(
        back_populates="visualization",
        sa_relationship_kwargs={
            "lazy": "selectin",
            "cascade": "all, delete",
            "uselist": False,
        },
    )
```

This is also never null, right ? Because you're always creating them together and there's a cascade delete constraint involved.

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,278 @@
...
+        response_resources = None
+        if include_resources:
+            artifact_version_model = (
+                self.artifact_version.to_model(
+                    include_metadata=include_metadata,
+                    include_resources=include_resources,
+                )
+                if self.artifact_version
+                else None
+            )
```
stefannica:
```suggestion
            artifact_version_model = (
                self.artifact_version.to_model(
                    include_metadata=include_metadata,
                    include_resources=include_resources,
                )
            )
```

This field can never be None

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class CuratedVisualizationResponseResources(ProjectScopedResponseResources):
+    """Response resources for curated visualizations."""
+
+    artifact_version: Optional["ArtifactVersionResponse"] = Field(
+        default=None,
+        title="The artifact version.",
+        description="Artifact version from which the visualization originates.",
+    )
```
stefannica: The artifact can never be None, because you modeled it as a non-nullable relationship with an auto-delete trigger: if the referenced artifact is deleted, then the curated visualization is also deleted.

```suggestion
    artifact_version: "ArtifactVersionResponse" = Field(
        title="The artifact version.",
        description="Artifact version from which the visualization originates.",
    )
```

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,242 @@
...
+    def layout_size(self) -> CuratedVisualizationSize:
+        """The layout size of the visualization.
+
+        Returns:
+            The layout size of the visualization.
+        """
+        return self.get_body().layout_size
+
+    @property
+    def artifact_version(self) -> Optional["ArtifactVersionResponse"]:
```
stefannica:
```suggestion
    def artifact_version(self) -> "ArtifactVersionResponse":
```

can't be None

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,278 @@
...
+    build_foreign_key_field,
+    build_index,
+)
+from zenml.zen_stores.schemas.utils import jl_arg
+
+if TYPE_CHECKING:
+    from zenml.zen_stores.schemas.artifact_schemas import ArtifactVersionSchema
+
+
+class CuratedVisualizationResourceSchema(SQLModel, table=True):
```
stefannica: If you now have a 1-to-1 relationship between the resource and the visualization, this entire separate CuratedVisualizationResourceSchema doesn't make sense anymore. If anything, it makes things more difficult for you.

You should merge the resource_id and resource_type fields directly in the CuratedVisualizationSchema.

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,278 @@
...
+                self.artifact_version.to_model(
+                    include_metadata=include_metadata,
+                    include_resources=include_resources,
+                )
+                if self.artifact_version
+                else None
+            )
+
+            resource_model = None
+            if self.resource:
```
stefannica: You should always have a referenced resource.

src/zenml/models/v2/misc/curated_visualization.py
```diff
@@ -0,0 +1,33 @@
...
+"""Models describing curated visualization resources."""
+
+from uuid import UUID
+
+from pydantic import BaseModel, Field
+
+from zenml.enums import VisualizationResourceTypes
+
+
+class CuratedVisualizationResource(BaseModel):
```
stefannica: This entire separate model doesn't make sense anymore, because you'll always have one and only one resource associated with the curated visualization. You can just move these two fields in the CuratedVisualizationResponse model.

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    store = zen_store()
+    hydrated_visualization = store.get_curated_visualization(
+        visualization_id, hydrate=True
+    )
+    resource = hydrated_visualization.resource
+    if resource is None:
+        raise RuntimeError(
+            f"Curated visualization '{visualization_id}' is missing its resource reference."
+        )
```
stefannica: The resource should never be None. In fact, the resource shouldn't even be a separate model, because there's always a 1-to-1 relationship between a visualization and its resource.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -241,6 +245,14 @@ class PipelineRunSchema(NamedSchema, RunMetadataInterface, table=True):
...
         ),
     )
+    visualizations: List["CuratedVisualizationSchema"] = Relationship(
+        sa_relationship_kwargs=dict(
+            secondary="curated_visualization_resource",
+            primaryjoin=f"and_(foreign(CuratedVisualizationResourceSchema.resource_type)=='{VisualizationResourceTypes.PIPELINE_RUN.value}', foreign(CuratedVisualizationResourceSchema.resource_id)==PipelineRunSchema.id)",
+            secondaryjoin="CuratedVisualizationSchema.id == CuratedVisualizationResourceSchema.visualization_id",
+            overlaps="visualizations",
+        ),
+    )
```
stefannica, line 259:
```suggestion
    visualizations: List["CuratedVisualizationSchema"] = Relationship(
        sa_relationship_kwargs=dict(
            secondary="curated_visualization_resource",
            primaryjoin=f"and_(foreign(CuratedVisualizationResourceSchema.resource_type)=='{VisualizationResourceTypes.PIPELINE_RUN.value}', foreign(CuratedVisualizationResourceSchema.resource_id)==PipelineRunSchema.id)",
            secondaryjoin="CuratedVisualizationSchema.id == CuratedVisualizationResourceSchema.visualization_id",
            overlaps="visualizations",
            cascade="all, delete-orphan",       
        ),
    )
```

There's a data integrity issue in your newly established relationships: when the resource is deleted, the curated visualizations are orphaned. There's not much you can do to automate this as an auto-delete at the database level without creating too much complexity, but you can at least enforce the auto-delete at the ORM level.

NOTE: I don't know if this really works with a one-to-many relationship, you'll have to test it. As an alternative, you can explicitly delete the associated visualizations in the SQL Zen store *for every type of resource* in their respective `delete_...` method implementation.

This comment applies to all the other supported resource types.

Also: given that you don't need the `CuratedVisualizationResourceSchema` anymore, this  relationship also needs to change and is simplified.

stefannica (changes requested):
The RBAC permission checks look good now, but I discovered another problem with how the SQL schemas are implemented: they are over-engineered and the foreign relationships still need some work.

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,214 @@
...
+            f"Curated visualization '{visualization_id}' is missing its resource identifier or type."
+        )
+
+    resource_model = _get_resource_model(resource_type, resource_id)
+    verify_permission_for_model(resource_model, action=Action.READ)
+
+    if hydrate:
+        return hydrated_visualization
+
+    return store.get_curated_visualization(visualization_id, hydrate=False)
```
stefannica:
```suggestion
    hydrated_visualization = store.get_curated_visualization(
        visualization_id, hydrate=hydrate
    )
    resource_type = hydrated_visualization.resource_type
    resource_id = hydrated_visualization.resource_id

    resource_model = _get_resource_model(resource_type, resource_id)
    verify_permission_for_model(resource_model, action=Action.READ)

    return hydrated_visualization
```

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,214 @@
...
+    store = zen_store()
+    existing_visualization = store.get_curated_visualization(
+        visualization_id, hydrate=True
+    )
+    resource_type = existing_visualization.resource_type
+    resource_id = existing_visualization.resource_id
+    if resource_type is None or resource_id is None:
+        raise RuntimeError(
+            f"Curated visualization '{visualization_id}' is missing its resource identifier or type."
+        )
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/curated_visualization_endpoints.py
```diff
@@ -0,0 +1,214 @@
...
```
stefannica:
```suggestion
```

src/zenml/zen_stores/migrations/versions/0ede689b266a_add_vizualisations_and_link_them_to_.py
```diff
@@ -0,0 +1,94 @@
+"""add vizualisations and link them to resources [0ede689b266a].
```
stefannica: Can you fix the name of this file please?

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,254 @@
...
+
+    artifact_version_id: UUID = Field(
+        title="The artifact version ID.",
+        description="Identifier of the artifact version providing the visualization.",
+    )
+    visualization_index: int = Field(
+        ge=0,
+        title="The visualization index.",
+        description="Index of the visualization within the artifact version payload.",
+    )
```
schustmi: Is there any reason why we're using the artifact version ID and visualization index to reference the visualization, instead of the ID of the visualization itself (`ArtifactVisualizationSchema`)?

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,235 @@
...
+
+        Args:
+            include_metadata: Whether to include metadata in the response.
+            include_resources: Whether to include resources in the response.
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            The created response model.
+        """
+        layout_size_value = (
```
schustmi: Why does this need a fallback, it will always have a value, no?

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,235 @@
...
+            self.layout_size or CuratedVisualizationSize.FULL_WIDTH.value
+        )
+        try:
+            layout_size_enum = CuratedVisualizationSize(layout_size_value)
+        except ValueError:
+            layout_size_enum = CuratedVisualizationSize.FULL_WIDTH
+
+        resource_type_str = self.resource_type
+        if resource_type_str:
```
schustmi: This will always be True

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,235 @@
...
+        )
+
+        metadata = None
+        if include_metadata:
+            metadata = CuratedVisualizationResponseMetadata()
+
+        response_resources = None
+        if include_resources:
+            artifact_version_model = self.artifact_version.to_model(
+                include_metadata=include_metadata,
```
schustmi: This is not how we usually handle this. If some other model is included in the resources of a response, that sub-model will only have a body. Which means both `include_metadata` as well as `include_resources` should be `False` here.

Is there any reason we need to make this exception here?

src/zenml/zen_stores/schemas/deployment_schemas.py
```diff
@@ -155,13 +174,22 @@ def get_query_options(
...
                 [
                     selectinload(jl_arg(DeploymentSchema.user)),
                     selectinload(jl_arg(DeploymentSchema.deployer)),
                     selectinload(jl_arg(DeploymentSchema.snapshot)).joinedload(
                         jl_arg(PipelineSnapshotSchema.pipeline)
                     ),
+                    selectinload(
+                        jl_arg(DeploymentSchema.visualizations)
+                    ).selectinload(
+                        jl_arg(CuratedVisualizationSchema.artifact_version)
```
schustmi: This second `selectinload` can be removed (here and in all other schemas that include curated visualizations) if you apply my changes on the `CuratedVisualizationSchema.to_model(...)`

schustmi (changes requested):

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,253 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+            }
+
+            if visualization.resource_type not in resource_schema_map:
+                raise IllegalOperationError(
```
schustmi: I think `ValueError` might be better, it's what we usually raise and results in a `422` response which I think is appropriate here with an invalid request that we can't process?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,253 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+
+            self._assert_curated_visualization_duplicate(
+                session=session,
+                artifact_version_id=visualization.artifact_version_id,
+                visualization_index=visualization.visualization_index,
+                resource_id=visualization.resource_id,
+                resource_type=visualization.resource_type,
+            )
+
+            schema: CuratedVisualizationSchema = (
```
schustmi:
```suggestion
            schema = (
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,253 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+        Returns:
+            The updated curated visualization.
+        """
+        with Session(self.engine) as session:
+            schema = self._get_schema_by_id(
+                resource_id=visualization_id,
+                schema_class=CuratedVisualizationSchema,
+                session=session,
+            )
+            schema.update(visualization_update)
```
schustmi, line 5664: What happens if I update a curated visualization to have the same display index as another one? Does that cause issues?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,253 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+        Raises:
+            IllegalOperationError: If the resource does not belong to the same project as the curated visualization.
+            KeyError: If the resource does not exist.
+        """
+        with Session(self.engine) as session:
+            self._set_request_user_id(
+                request_model=visualization, session=session
+            )
+
+            artifact_version: ArtifactVersionSchema = self._get_schema_by_id(
```
schustmi: Does the `_get_reference_schema_by_id` method that we usually use to check the project consistency not work here?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,253 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+                resource_id=visualization.artifact_version_id,
+                schema_class=ArtifactVersionSchema,
+                session=session,
+            )
+
+            # Validate explicitly provided project before defaulting
+            if visualization.project:
```
schustmi: This is always True

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,253 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+                artifact_version_id=visualization.artifact_version_id,
+                visualization_index=visualization.visualization_index,
+                resource_id=visualization.resource_id,
+                resource_type=visualization.resource_type,
+            )
+
+            schema: CuratedVisualizationSchema = (
+                CuratedVisualizationSchema.from_request(visualization)
+            )
+            schema.project_id = project_id
```
schustmi: This already happens as part of `CuratedVisualizationSchema.from_request`

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,235 @@
...
+
+class CuratedVisualizationSchema(BaseSchema, table=True):
+    """SQL Model for curated visualizations."""
+
+    __tablename__ = "curated_visualization"
+    __table_args__ = (
+        build_index(
+            __tablename__, ["artifact_version_id", "visualization_index"]
+        ),
+        build_index(__tablename__, ["display_order"]),
```
schustmi: Any reason we need this specific index?

src/zenml/zen_stores/schemas/deployment_schemas.py
```diff
@@ -133,6 +140,18 @@ class DeploymentSchema(NamedSchema, table=True):
         ),
     )
 
+    visualizations: List["CuratedVisualizationSchema"] = Relationship(
```
schustmi, line 143: You might want an `order_by` relationship kwarg here as well, otherwise there is no guarantee this shows up in the right oder. I thought the `display_order` exists for this specific case, no?

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
safoinme: yes it's required

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
safoinme: You’re right—there’s no special case here

src/zenml/models/v2/core/curated_visualization.py
safoinme: No valid technical reason, rather than missing it, will update this to use ID of the visualization itself

src/zenml/zen_stores/sql_zen_store.py
safoinme: maybe question for @Cahllagerfeld? how are we expecting to send this from front-end

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
safoinme: not really

src/zenml/zen_stores/sql_zen_store.py
safoinme: Yes, it's updated now

src/zenml/models/v2/core/artifact_visualization.py
```diff
@@ -105,6 +117,23 @@ def artifact_version_id(self) -> UUID:
...
+    def artifact_version(self) -> "ArtifactVersionResponse":
+        """The artifact version resource, if the response was hydrated with it.
+
+        Returns:
+            The artifact version resource associated with this visualization.
+
+        Raises:
+            RuntimeError: If the artifact version resource is not available.
+        """
+        resources = self.get_resources()
```
schustmi: `self.get_resources()` will always hydrate the response, so you can remove all the code below (See any property of any other model that returns something from `resources`)

src/zenml/zen_stores/schemas/artifact_visualization_schemas.py
```diff
@@ -107,8 +122,18 @@ def to_model(
                 artifact_version_id=self.artifact_version_id,
             )
 
+        resources = None
+        if include_resources and self.artifact_version is not None:
```
schustmi: If `include_resources` is True, you should always include it in the response. It might be that the artifact version is `None`, but the resources should still be included.

src/zenml/models/v2/core/artifact_visualization.py
```diff
@@ -105,6 +117,23 @@ def artifact_version_id(self) -> UUID:
         """
         return self.get_metadata().artifact_version_id
 
+    @property
+    def artifact_version(self) -> "ArtifactVersionResponse":
```
schustmi: The return of this is `Optional`, as the visualization might not have a linked artifact version in the DB.

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,288 @@
...
+    def artifact_version(self) -> "ArtifactVersionResponse":
+        """The artifact version resource, if available.
+
+        Returns:
+            The artifact version resource associated with the curated visualization.
+
+        Raises:
+            RuntimeError: If the artifact version is not included in the hydrated resources.
+        """
+        artifact_visualization = self.artifact_visualization
```
schustmi: Same here, please follow what all other models do. There is no need to raise a custom error here, because the resources should always be hydrated.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -632,6 +652,13 @@ def to_model(
                 tags=[tag.to_model() for tag in self.tags],
                 logs=client_logs[0].to_model() if client_logs else None,
                 log_collection=[log.to_model() for log in self.logs],
+                visualizations=[
+                    visualization.to_model(
+                        include_metadata=False,
+                        include_resources=False,
+                    )
+                    for visualization in (self.visualizations or [])
```
```suggestion
                    for visualization in self.visualizations
```
The type annotation for this is a list, so we don't need a fallback. Same in all other schemas.

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,241 @@
...
+            resource_id=self.resource_id,
+            resource_type=resource_type_enum,
+        )
+
+        metadata = None
+        if include_metadata:
+            metadata = CuratedVisualizationResponseMetadata()
+
+        response_resources = None
+        if include_resources and self.artifact_visualization is not None:
```
schustmi: Again, resources should always be included if `include_resources` is `True`.

src/zenml/zen_stores/schemas/curated_visualization_schemas.py
```diff
@@ -0,0 +1,241 @@
...
+        options: List[ExecutableOption] = []
+
+        if include_resources:
+            from zenml.zen_stores.schemas.artifact_visualization_schemas import (
+                ArtifactVisualizationSchema,
+            )
+
+            options.append(
+                selectinload(jl_arg(cls.artifact_visualization)).options(
+                    selectinload(
```
schustmi: The second one isn't necessary, as it won't be included in the visualization response.

src/zenml/zen_stores/sql_zen_store.py
schustmi: Well I assume that backend should just guarantee that they have unique values? The frontend gets a list I assume, but if you want to allow users to define the order for this, it should work as expected. (Which currently, when we allow duplicate values, it doesn't. I can set something to display in position 2, but if 10 others already have the same value, it might only show up in position 11.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,275 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+            schema = self._get_schema_by_id(
+                resource_id=visualization_id,
+                schema_class=CuratedVisualizationSchema,
+                session=session,
+            )
+            update_fields = visualization_update.model_dump(exclude_unset=True)
+            if "display_order" in update_fields:
+                new_display_order = update_fields["display_order"]
+                if new_display_order is not None:
+                    self._assert_curated_visualization_display_order_unique(
```
schustmi, line 5653: This is prone to race conditions if multiple requests to update display order happen at the same time. I'm okay if we merge this now because we're short on time, but the real solution would be a unique index on resource type, resource ID and display order.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5394,6 +5417,275 @@ def delete_deployment(self, deployment_id: UUID) -> None:
...
+                        session=session,
+                        resource_id=schema.resource_id,
+                        resource_type=VisualizationResourceTypes(
+                            schema.resource_type
+                        ),
+                        display_order=new_display_order,
+                        exclude_visualization_id=visualization_id,
+                    )
+                # Explicit None clears the display order, so uniqueness validation is skipped.
```
schustmi, line 5662: Have you verified that this works? I know we've had problems before when relying on `None` values in update models to clear values.

Can you test the following to make sure:
- Run with a server
- Set the display order of a visualization
- Send an empty update for the visualization from the client
- Check display order is still set
- Send an update with `display_order=None`
- Check display order is removed

docs/book/how-to/artifacts/visualizations.md
```diff
@@ -65,6 +65,185 @@ There are three ways how you can add custom visualizations to the dashboard:
...
+    artifact_visualization_id=artifact_visualizations[2].id,
+    resource_type=VisualizationResourceTypes.PIPELINE,
+    resource_id=pipeline_id,
+    display_name="Raw output preview",
+    display_order=15,  # Now appears between A and B
+    layout_size=CuratedVisualizationSize.FULL_WIDTH,
+)
+```
+
+> **Note:** Reusing the same `display_order` for multiple curated visualizations targeting the same resource now raises an error. Pick distinct values—leaving gaps still helps when you later insert additional tiles.
```
schustmi: `...now raises an error.`: This is the first time users ever see these visualizations, so whatever the behaviour "now" is is the default. I don't think we need to mention this is in the docs, it's very clear that an ordered thing should have unique values.

src/zenml/models/v2/core/curated_visualization.py
```diff
@@ -0,0 +1,271 @@
...
+        description=(
+            "Identifier of the artifact visualization that should be surfaced "
+            "for the target resource."
+        ),
+    )
+    display_name: Optional[str] = Field(
+        default=None,
+        title="The display name of the visualization.",
+    )
+    display_order: Optional[int] = Field(
```
schustmi: Should we make this an `Optional[NonNegativeInt]` instead? (`from pydantic import NonNegativeInt`)

Negative values seem a little weird to me here, but if you think they make sense feel free to ignore as well.

schustmi (changes requested):

PR #4027, opened 2025-10-01
feat(integrations): Add MLX array materializer

nicholasjng:
This enables saving and loading MLX arrays when using ZenML on Apple platforms.

Structurally, this is very similar to the recently added JAX integration, it's just the array type that is different, plus we're using MLX's own array IO, so there is no hard dependency on NumPy.

## Types of changes

### DIFF
--- a/tests/integration/integrations/mlx/test_mlx_array_materializer.py
+++ b/tests/integration/integrations/mlx/test_mlx_array_materializer.py
@@ -12,12 +12,20 @@
 #  or implied. See the License for the specific language governing
 #  permissions and limitations under the License.
 
+import sys
+
 import mlx.core as mx
+import pytest
 
 from tests.unit.test_general import _test_materializer
+from zenml.integrations.mlx import SUPPORTED_PLATFORMS
 from zenml.integrations.mlx.materializer import MLXArrayMaterializer
 
 
+@pytest.mark.skipif(
+    sys.platform not in SUPPORTED_PLATFORMS,
+    reason="MLX only runs on Apple and Linux",
+)
 def test_mlx_array_materializer():
     """Test the MLX array materializer."""
     arr = mx.ones(5)
--- a/src/zenml/integrations/mlx/__init__.py
+++ b/src/zenml/integrations/mlx/__init__.py
@@ -13,15 +13,43 @@
 #  permissions and limitations under the License.
 """Initialization of the MLX integration."""
 
+import sys
+from typing import List, Optional
+
 from zenml.integrations.constants import MLX
 from zenml.integrations.integration import Integration
+
+# MLX can run on Linux, but only with CPU or Cuda devices,
+# see https://ml-explore.github.io/mlx/build/html/install.html#cuda ff.
+SUPPORTED_PLATFORMS = ("darwin", "linux")
 
 
 class MLXIntegration(Integration):
     """Definition of MLX array integration for ZenML."""
 
     NAME = MLX
-    REQUIREMENTS = ["mlx"]
+
+    @classmethod
+    def check_installation(cls) -> bool:
+        if sys.platform not in SUPPORTED_PLATFORMS:
+            return False
+        return super().check_installation()
+
+    @classmethod
+    def get_requirements(
+        cls,
+        target_os: Optional[str] = None,
+        python_version: Optional[str] = None,
+    ) -> List[str]:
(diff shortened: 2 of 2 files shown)


src/zenml/integrations/mlx/materializer.py
```diff
@@ -0,0 +1,63 @@
...
+    def load(self, data_type: Type[Any]) -> mx.array:
+        """Reads data from a `.npy` file, and returns an MLX array.
+
+        Args:
+            data_type: The type of the data to read.
+
+        Returns:
+            The MLX array.
+        """
+        numpy_file = os.path.join(self.uri, NUMPY_FILENAME)
```
schustmi, line 47: If using the `mx.save(...)` function, does that still output a file in the actual numpy format? Or does it make sense to rename this?

nicholasjng: Yes, it does. MLX rolls its own implementation, source here (it's C++ though): https://github.com/ml-explore/mlx/blob/e88f2d4a8e35656e8b4918fca4858a70480678ae/mlx/io/load.cpp

But it is numpy format, and the extension `.npy` is specifically associated with single arrays, not dicts of arrays. (In fact, it requires the output file be suffixed with `.npy`, or it will fail - for dicts of arrays, use `savez()`.)

schustmi: Makes sense, thanks!

schustmi:
@nicholasjng Can you add a quick test that just saves/loads a simple array? We have a `tests/unit/test_general.py::_test_materializer` util function for that, should be relatively quick I hope

nicholasjng:
Done, please take another look when you find the time.

tests/integration/integrations/mlx/test_mlx_array_materializer.py
```diff
@@ -0,0 +1,30 @@
+#  Copyright (c) ZenML GmbH 2025. All Rights Reserved.
```
schustmi, line 1: One more thing, can you move this to `tests/integration/integrations/mlx` instead, as I think the environment in which our unit tests run don't usually have all the integrations installed.

schustmi (changes requested):

tests/integration/integrations/mlx/test_mlx_array_materializer.py
nicholasjng: Done.

schustmi:
Still failing the linting, otherwise lgtm

nicholasjng:
Fixed (I hope) in https://github.com/zenml-io/zenml/pull/4027/commits/057ef78b76cfd62c4c10d3a694ea78e30b75dbc1.

schustmi:
@nicholasjng Seems like it's still failing because the CLI is missing the `libmlx.so` library. I'll take a look on monday, not sure where to best install this, and how to solve it on windows or whether we need to skip that test there.

nicholasjng:
I think this is just because mlx only works on Apple Silicon, and CI is run on Ubuntu.

schustmi:
@nicholasjng Ah I see, in that case I have a few ideas to make this better:
- Use the `@pytest.mark.skipif` decorator to skip the test if the OS isn't MacOS
- Overwrite the `get_requirements()` method in your integration class. This is more flexible than defining the requirements in the list, and in this case we can return an empty list when the OS is not MacOS
- Overwrite the `check_installation()` method in your integration class, and return `False` if the OS is not MacOS. Otherwise we can fallback to the superclass implementation. This is required in combination with the second bullet point, as otherwise the integration would always be installed on Linux because it has no requirements. This would cause us to call the `activate()` method which would try to import the materializer and fail.

nicholasjng:
thanks! That seems to have worked. One last question - is there a way to inject dependencies based on environment markers?

Reason I'm asking is that `mlx` also has a `cuda` extra, which, similarly to `cpu`, is only applicable on Linux, but only works on machines with an nvidia GPU. Is there support for this fine-grained integration setup in ZenML? (I guess we could solve it with an envvar in the end, just being curious)

schustmi:
No we don't have anything like that atm. I'd suggest we merge this as-is for now, worst case people will have to install the GPU version manually for now I guess?

nicholasjng:
sounds good!

schustmi (approved):

PR #4036, opened 2025-10-03
Update DockerHub flavor Container Registry documentation for using open source Quay remote registry.

JustGitting:
## Describe changes
Updated the DockerHub documentation to show how to use open source docker image registry, Quay. I've also left space for other FOSS remote image registries, such as Harbor.

Question for discussion: The "dockerhub" flavour name is misleading as it implies that it only works with dockerhub and not other remote container registries. A better flavour name could be "remote_docker", which would be consistent with the existing "local_docker" flavour.

Also minor update to Kaniko documentation.

## Pre-requisites

## Types of changes


schustmi (changes requested):
> Question for discussion: The "dockerhub" flavour name is misleading as it implies that it only works with dockerhub and not other remote container registries. A better flavour name could be "remote_docker", which would be consistent with the existing "local_docker" flavour.

There is a `default` flavor for container registries, which is meant to be used for any container registry which doesn't fit any of the other flavors. In reality however, the `default` and `dockerhub` container registries are currently functionally equivalent, as both of them only use the python docker client to push images.

**However**: In the future, the `dockerhub` flavor implementation might actually implement logic that only works for dockerhub container registries. So while I think all your documentation makes sense, I think it should instead be for the `default` flavor.

schustmi:
Also, regarding the Kaniko changes: I wasn't aware of this, but that's great news. Adding it to the docs however is a little misleading I think, as our current implementation still uses the unmaintained container images. Would you mind removing that from the docs here, and instead creating a GH issue for us to update the integration to this new fork?

schustmi:
@JustGitting Are you still planning on addressing those changes?

JustGitting:
@schustmi yes, just getting back into zenml after working on other projects.

PR #4038, opened 2025-10-03
Upgrade Modal step operator to support `modal>=1`

strickvl:
## What this does

Upgrades the ZenML Modal step operator to work with the **Modal Python SDK `>=1.0`**.

Modal's 1.0 release removed the private/internal APIs the operator previously used to build container images from a registry. This PR swaps those internals for Modal's public 1.x APIs, bumps the dependency pin, and tightens the step-operator runtime wiring around Modal auth, environments, resources, and sandbox submission.

## Changes

- **Dependency pin:** `modal>=0.64.49,<1` → `modal>=1,<2`, keeping the integration explicitly on Modal SDK v1 until v2 is tested.
- **Image building:** replaced the private pre-1.0 path —
  `modal.Image._from_args(...)` + `modal.image._ImageRegistryConfig(api_pb2.REGISTRY_AUTH_TYPE_STATIC_CREDS, ...)` + `modal.secret._Secret.from_dict(...)` —
  with the public 1.x API:
  `modal.Image.from_registry(image, secret=modal.Secret.from_dict({...}))`,
  and dropped the now-unneeded `from modal_proto import api_pb2` import.
- **Runtime environment handling:** ZenML step runtime environment variables are now passed to `modal.Sandbox.create(env=...)` instead of being attached to the Modal image definition with `Image.env(...)`, avoiding sensitive runtime values appearing in Modal image build logs. Obvious sensitive runtime credentials such as `ZENML_STORE_API_TOKEN` are split out of plain `env` and passed through `modal.Secret.from_dict(...)` / `Sandbox.create(secrets=[...])` instead.
- **Modal auth/environment wiring:** `token_id` and `token_secret` now create a lazily cached explicit `modal.Client.from_credentials(...)` client that is passed to `App.lookup`, `Sandbox.create`, and `Sandbox.from_id`. If no token pair is configured, ZenML passes `client=None` so Modal uses its normal ambient auth. `modal_environment` is passed explicitly to `App.lookup(..., environment_name=...)`; ZenML no longer temporarily mutates `MODAL_TOKEN_ID`, `MODAL_TOKEN_SECRET`, `MODAL_ENVIRONMENT`, or unsupported `MODAL_WORKSPACE` values. The unsupported `workspace` config field was removed/deferred.
- **Sandbox submission correctness:** `submit()` now preserves the entrypoint command as argv items with `modal.Sandbox.create(*entrypoint_command, ...)` instead of flattening it through `bash -c " ".join(...)`. If metadata publication or local metadata update fails after sandbox creation, ZenML now terminates the sandbox before re-raising so it does not leave an untracked remote process running.
- **Timeout handling:** the configured Modal step-operator `timeout` is now passed to `Sandbox.create()` instead of always using 24h.
- **Registry auth behavior:** registry credentials are passed to Modal when available; if none are configured, Modal attempts an anonymous image pull.
- **Resource handling:** GPU type/count validation is stricter, explicit `gpu_count=0` now means CPU-only, and memory requests are rounded up to Modal's MB integer value.
- **Cleanup/tests/docs:** removed the now-unused Modal utility path, centralized the Click `CliRunner` compatibility helper, tightened the root CLI smoke test, removed unsupported Modal workspace wording, and updated Modal docs to match the runtime behavior.

## Testing

Targeted checks run locally:

```shell
uv run pytest tests/integration/integrations/modal/flavors/test_flavor_config.py tests/integration/integrations/modal/step_operators/test_modal_step_operator.py
# 16 passed

uv run pytest tests/integration/integrations/modal/step_operators/test_modal_step_operator.py
# 10 passed

uv run pytest tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py
# 16 passed

uv run pytest tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py tests/integration/functional/cli/test_cli.py -q
# 25 passed

uv run ruff check --ignore D100,D102,D103,D107 src/zenml/integrations/modal/flavors/modal_step_operator_flavor.py src/zenml/integrations/modal/step_operators/modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/functional/cli/test_model.py tests/integration/functional/cli/test_tag.py

uv run ruff format --check src/zenml/integrations/modal/flavors/modal_step_operator_flavor.py src/zenml/integrations/modal/step_operators/modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/functional/cli/test_model.py tests/integration/functional/cli/test_tag.py

bash scripts/format.sh --no-yamlfix src/zenml/integrations/modal/step_operators/modal_step_operator.py tests/integration/integrations/modal/step_operators/test_modal_step_operator.py
```

Modal SDK 1.x signatures were verified against installed Modal `1.4.3` for the public APIs used here (`Image.from_registry`, `App.lookup`, `Sandbox.create`, `Sandbox.from_id`, `Sandbox.create(env=...)`, `Client.from_credentials`, and explicit `client=` / `environment_name=` wiring). The latest refactor removes temporary Modal SDK env-var mutation; non-default Modal Environment behavior should be revalidated end-to-end through `App.lookup(..., environment_name=...)` in full CI or manual stack testing.

Latest targeted checks after the explicit-client refactor:

```shell
uv run pytest tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py
# 22 passed

uv run ruff check src/zenml/integrations/modal/step_operators/modal_step_operator.py src/zenml/integrations/modal/flavors/modal_step_operator_flavor.py
# All checks passed

uv run mypy src/zenml/integrations/modal/step_operators/modal_step_operator.py
# Success: no issues found in 1 source file
```

Latest targeted checks after sandbox hardening:

```shell
uv run pytest tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py
# 25 passed, 1 warning

uv run bash scripts/lint.sh
# Ruff, pydoclint, and yamlfix passed; zizmor failed while trying to reach GitHub for actions/checkout tag lookup (TLS handshake EOF), unrelated to the three Modal files changed in the latest commit.
```

Latest follow-up after review comments:

```shell
uv run ruff format src/zenml/integrations/modal/step_operators/modal_step_operator.py tests/integration/integrations/modal/step_operators/test_modal_step_operator.py
# 2 files left unchanged

uv run ruff check src/zenml/integrations/modal/step_operators/modal_step_operator.py
# All checks passed

uv run ruff check tests/integration/integrations/modal/step_operators/test_modal_step_operator.py --extend-ignore D
# All checks passed

uv run mypy src/zenml/integrations/modal/step_operators/modal_step_operator.py
# Success: no issues found in 1 source file

uv run pydoclint src/zenml/integrations/modal/step_operators/modal_step_operator.py
# No violations

uv run pytest tests/integration/integrations/modal/step_operators/test_modal_step_operator.py tests/integration/integrations/modal/flavors/test_flavor_config.py
# 24 passed, 1 warning
```

This follow-up also fixes the CI mypy invariance error for `modal.Secret.from_dict(...)` by typing the sensitive runtime secret environment as `dict[str, str | None]`.

End-to-end remote sandbox execution should still be validated against a real Modal account in full CI or manual stack testing.

Additional manual validation:

- Ran a local smoke pipeline on the `modal` stack against the Modal `dev` environment.
- Validated one CPU-only Modal step and one GPU Modal step (`L4`, `gpu_count=1`) end-to-end.
- The GPU step ran `nvidia-smi` and checked `torch.cuda.is_available()` inside the Modal sandbox.
- Confirmed the pipeline completed successfully with both Modal step outputs flowing into the final step.
- Removed the deprecated `environment_name` argument from `Sandbox.create`; Modal environment selection now happens through `App.lookup(..., environment_name=...)`. ZenML no longer uses a temporary local `MODAL_ENVIRONMENT` override.

## Reviewer notes

The core compatibility fix is the public Modal image API replacement. The follow-up runtime changes avoid shell command flattening, keep ZenML runtime env vars out of Modal image build logs, use explicit Modal clients for configured credentials, preserve ambient auth with `client=None`, and align docs/tests with the actual Modal SDK >=1 behavior.


src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -153,7 +106,7 @@ def launch(
         self,
         info: "StepRunInfo",
         entrypoint_command: List[str],
-        environment: Dict[str, str],
+        environment: Optional[Dict[str, str]],
```
schustmi: Why was this change made? The base class still has a non-optional dict there

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -163,80 +116,95 @@ def launch(
...
-
-        zenml_image = modal.Image._from_args(
-            dockerfile_function=lambda *_, **__: spec,
-            force_build=False,
-            image_registry_config=modal.image._ImageRegistryConfig(
-                api_pb2.REGISTRY_AUTH_TYPE_STATIC_CREDS, my_secret
-            ),
-        ).env(environment)
+        try:
+            zenml_image = build_modal_image(image_name, stack, environment)
```
schustmi: This is not a "zenml_image", as the function name indicates

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -163,80 +116,95 @@ def launch(
...
+        except Exception as e:
+            raise RuntimeError(
+                "Failed to build the Modal image from your Docker registry. "
+                "Action required: ensure your ZenML stack's container registry is configured with valid credentials "
+                "and that the base image exists and is accessible. "
```
schustmi: I doubt any of those two cases are actually the cause for failure here? We've built that image just before that pipeline run, and apparently also had permissions to push it there?

schustmi: Also, what is a "base image" for the user?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -163,80 +116,95 @@ def launch(
...
+                        )
+
+                        try:
+                            if (
+                                not entrypoint_command
+                                or not entrypoint_command[0]
+                            ):
+                                raise ValueError(
+                                    "Empty step entrypoint command is not allowed."
+                                )
```
schustmi: This is not happening
```suggestion
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -163,80 +116,95 @@ def launch(
...
-                        timeout=86400,  # 24h, the max Modal allows
-                    )
-
-                    await sb.wait.aio()
-
-            future.set_result(None)
-            return future
+                try:
+                    async with app.run():
+                        memory_int = (
```
schustmi: This we can maybe move outside the `with app.run` context?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -163,80 +116,95 @@ def launch(
...
+                                memory=memory_int,
+                                cloud=settings.cloud,
+                                region=settings.region,
+                                app=app,
+                                timeout=settings.timeout,
+                            )
+                        except Exception as e:
+                            raise RuntimeError(
+                                "Failed to create a Modal sandbox. "
+                                "Action required: verify that the referenced Docker image exists and is accessible, "
```
schustmi: The docker image exits for sure, otherwise it would have failed earlier

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -163,80 +116,95 @@ def launch(
...
+                                "permissions allow sandbox creation. "
+                                f"Context: image='{image_name}', gpu='{gpu_values}', region='{settings.region}', cloud='{settings.cloud}'."
+                            ) from e
+
+                        try:
+                            await sb.wait.aio()
+                        except Exception as e:
+                            raise RuntimeError(
+                                "Modal sandbox execution failed. "
+                                "Action required: inspect the step logs in Modal, validate your entrypoint command, "
```
schustmi: The entrypoint command is coming from ZenML and the user shouldn't care

src/zenml/integrations/modal/utils.py
```diff
@@ -0,0 +1,331 @@
...
+        {
+            "REGISTRY_USERNAME": docker_username,
+            "REGISTRY_PASSWORD": docker_password,
+        }
+    )
+
+    try:
+        modal_image = modal.Image.from_registry(
+            image_name, secret=registry_secret
+        ).pip_install("modal")
```
schustmi: Modal should already be installed in that image as it must include the stack requirements

schustmi (changes requested):

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Oh. Right hmmm I think I got confused between environment here (which is about env variables) and the Modal 'environment' which is a separate thing specific to them. I'll revert.

tests/integration/functional/cli/utils.py
```diff
@@ -35,6 +38,16 @@
 ]
 
 
+def cli_runner(**kwargs) -> CliRunner:
+    """Return a Click runner that stays compatible across Click releases."""
+    runner_kwargs = dict(kwargs)
```
schustmi: kwargs is already a dict

src/zenml/integrations/modal/flavors/modal_step_operator_flavor.py
```diff
@@ -36,20 +41,92 @@ class ModalStepOperatorSettings(BaseSettings):
...
+        "When token_id and token_secret are both provided, ZenML creates an explicit Modal SDK "
+        "client from those credentials. If not provided, Modal falls back to its default "
+        "authentication from environment variables or ~/.modal.toml.",
+    )
+
+    @model_validator(mode="after")
+    def validate_modal_token_pair(self) -> "ModalStepOperatorConfig":
+        """Validate that Modal token fields are configured together."""
+        token_id = self.token_id.strip() if self.token_id else None
+        token_secret = self.token_secret.strip() if self.token_secret else None
```
schustmi, line 122: These could still be referencing a secret key, for which the value will be empty. Just FYI, not sure if this means we'd need to do another check of this at runtime.

schustmi: nvm, happens at runtime already.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -67,6 +128,37 @@ class ModalStepOperator(BaseStepOperator):
...
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the Modal step operator."""
+        super().__init__(*args, **kwargs)
+        self._modal_client: Optional["modal.Client"] = None
+        self._modal_client_lock = threading.Lock()
+
+    def _get_modal_client(self) -> Optional["modal.Client"]:
+        """Get an explicit Modal client when credentials are configured."""
+        token_id = _normalize_optional_config_value(self.config.token_id)
```
schustmi: This should only happen if we're actually creating the client, not in case we're just returning the cached one, right?

Background: In case the value is a secret reference, this results in an API call.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -67,6 +128,37 @@ class ModalStepOperator(BaseStepOperator):
...
+            self.config.token_secret
+        )
+
+        if bool(token_id) != bool(token_secret):
+            raise StackComponentInterfaceError(
+                "Modal token_id and token_secret must be configured together."
+            )
+
+        if not token_id or not token_secret:
+            return None
```
schustmi: Is there no `modal.Client.from_env` or similar that we could use to make this method return a non-optional client?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -40,24 +42,83 @@
...
+
+    Raises:
+        StackComponentInterfaceError: If the configuration is inconsistent or invalid.
     """
-    if not settings.gpu:
-        return None
+    gpu_type_raw = settings.gpu
+    gpu_type = gpu_type_raw.strip() if gpu_type_raw is not None else None
+    if gpu_type == "":
+        gpu_type = None
```
schustmi: `_normalize_optional_config_value` does exactly this, no?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -40,24 +42,83 @@
...
+
     gpu_count = resource_settings.gpu_count
-    return f"{settings.gpu}:{gpu_count}" if gpu_count else settings.gpu
+    if gpu_count is not None:
+        try:
+            gpu_count = int(gpu_count)
```
schustmi: gpu_count is guaranteed to be a non-negative int by pydantic validators on resource settings.

schustmi (changes requested):

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Ok yeah I can check the cached client before reading token fields.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Modal does have `Client.from_env`, but it is async/hidden-ish (not even listed in their docs). The way I'm doing it currently is simpler than what we'd have to do to use `from_env`. WDYT?

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Yeah correct. I'll replace it with the helper. thanks

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: Yeah probably over-defensive. We'd never reach that state. I'll remove it.

src/zenml/integrations/modal/step_operators/modal_step_operator.py
schustmi: Let's keep it as is then

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -67,6 +134,39 @@ class ModalStepOperator(BaseStepOperator):
...
+        super().__init__(*args, **kwargs)
+        self._modal_client: Optional["modal.Client"] = None
+        self._modal_client_lock = threading.Lock()
+
+    def _get_modal_client(self) -> Optional["modal.Client"]:
+        """Get an explicit Modal client when credentials are configured."""
+        with self._modal_client_lock:
+            modal_client = self._modal_client
+            if modal_client is not None and not modal_client.is_closed():
+                return modal_client
```
schustmi, line 152: One tiny improvement: I think this check could happen outside the lock?

schustmi (approved):

src/zenml/integrations/modal/step_operators/modal_step_operator.py
strickvl: https://github.com/zenml-io/zenml/pull/4038/changes/f9b59e26f2d91da4618a283a8709cfa6b80da71e

PR #4039, opened 2025-10-06
Prevent MLX integration install on Intel Macs

strickvl:
MLX wheels are only available for Apple Silicon (ARM64) Macs, not Intel (x86_64). Added architecture detection using `platform.machine()` to `check_installation()` and `get_requirements()` to prevent installation attempts on unsupported Intel Macs.

The integration now properly detects:
- Apple Silicon Macs (arm64/aarch64): Supported
- Intel Macs (x86_64): Not supported
- Linux (any arch): Supported

### DIFF
--- a/src/zenml/integrations/mlx/__init__.py
+++ b/src/zenml/integrations/mlx/__init__.py
@@ -20,12 +20,6 @@ from typing import List, Optional
 from zenml.integrations.constants import MLX
 from zenml.integrations.integration import Integration
 
-# MLX can run on Linux, but only with CPU or Cuda devices,
-# see https://ml-explore.github.io/mlx/build/html/install.html#cuda ff.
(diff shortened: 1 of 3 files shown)


src/zenml/integrations/mlx/__init__.py
```diff
@@ -21,19 +22,37 @@
 
 # MLX can run on Linux, but only with CPU or Cuda devices,
 # see https://ml-explore.github.io/mlx/build/html/install.html#cuda ff.
+# On macOS, MLX only supports Apple Silicon (ARM64), not Intel (x86_64).
 SUPPORTED_PLATFORMS = ("darwin", "linux")
```
schustmi: I think this can be removed now? It's probably still used in the test, which should now also skip the test depending on the `_is_supported_platform` check instead I assume?

nicholasjng:
ups, that was my bad. thanks for cleaning this up :)

PR #4040, opened 2025-10-06
Additional caching configurations

schustmi:
## Describe changes
This PR adds some additional options to configure caching:
- Files for which content changes should invalidate the caching
- Python objects (modules, classes, functions) for which code changes should invalidate the caching
- Custom cache function, for which a different result invalidates caching
- Cache expiration: The amount of seconds for which a step run should stay a valid cache candidate

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/user-guide/starter-guide/cache-previous-executions.md
+++ b/docs/book/user-guide/starter-guide/cache-previous-executions.md
@@ -190,8 +190,7 @@ cache_policy = CachePolicy(source_dependencies=["run.my_helper_function"])
 #### Cache expiration
 
 By default, any step that executes successfully is a caching candidate for future step runs. Any step with the same [cache key](#understanding-cache-keys) running afterwards
-can reuse the output artifacts produced by the caching candidate instead of actually executing the step code. In some cases however you might want to limit the time for how long
-a step run is a valid cache candidate for future steps. You can do that by configuring an expiration time for your step runs:
+can reuse the output artifacts produced by the caching candidate instead of actually executing the step code. In some cases however you might want to limit for how long a step run is a valid cache candidate for future steps. You can do that by configuring an expiration time for your step runs:
 
 ```python
 from zenml.config import CachePolicy
--- a/src/zenml/models/v2/core/step_run.py
+++ b/src/zenml/models/v2/core/step_run.py
@@ -782,8 +782,9 @@ class StepRunFilter(ProjectScopedFilter, RunMetadataFilterMixin):
             )
 
         if self.cache_expired is True:
-            cache_expiration_filter = (
-                col(StepRunSchema.cache_expires_at) < utc_now()
+            cache_expiration_filter = and_(
+                col(StepRunSchema.cache_expires_at).is_not(None),
+                col(StepRunSchema.cache_expires_at) < utc_now(),
             )
             custom_filters.append(cache_expiration_filter)
         elif self.cache_expired is False:
--- a/src/zenml/orchestrators/cache_utils.py
+++ b/src/zenml/orchestrators/cache_utils.py
@@ -23,7 +23,7 @@ from zenml.constants import CODE_HASH_PARAMETER_NAME
 from zenml.enums import ExecutionStatus, SorterOps
 from zenml.logger import get_logger
 from zenml.orchestrators import step_run_utils
-from zenml.utils import io_utils, source_code_utils, source_utils
+from zenml.utils import source_code_utils, source_utils
 
 if TYPE_CHECKING:
     from zenml.artifact_stores import BaseArtifactStore
@@ -74,6 +74,7 @@ def generate_cache_key(
         ValueError: If some file dependencies are outside the source root or
             missing.
         ValueError: If the cache function is invalid.
+        RuntimeError: If executing the cache function failed.
 
     Returns:
         A cache key.
@@ -173,12 +174,11 @@ def generate_cache_key(
         for file_path in sorted(absolute_paths):
             # Use relative path to source root as key so we don't generate
             # different hashes on different machines
-            relative_path = os.path.relpath(
-                file_path, source_utils.get_source_root()
-            )
+            relative_path = os.path.relpath(file_path, source_root)
             hash_.update(relative_path.encode())
-            file_content = io_utils.read_file_contents_as_string(file_path)
-            hash_.update(file_content.encode())
+            with open(file_path, "rb") as f:
+                while chunk := f.read(1024 * 1024):
+                    hash_.update(chunk)
 
     if cache_policy.source_dependencies:
         source_dependencies = [
@@ -199,7 +199,10 @@ def generate_cache_key(
                 "not callable."
             )
 
-        result = cache_func()
+        try:
+            result = cache_func()
+        except Exception as e:
+            raise RuntimeError("Failed to run cache function.") from e
 
         if not isinstance(result, str):
             raise ValueError(


docs/book/user-guide/best-practices/set-up-your-repository.md
```diff
@@ -170,7 +170,7 @@ Collect all your notebooks in one place.
 
 #### .zen
 
-By running `zenml init` at the root of your project, you define the project scope for ZenML. In ZenML terms, this will be called your "source root". This will be used to resolve import paths and store configurations.
+By running `zenml init` at the root of your project, you define the [source root](https://docs.zenml.io/concepts/steps_and_pipelines/sources#source-root) for you project.
```
Json-Andriopoulos: your*

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -130,6 +131,12 @@ def populate_request(
         )
         request.cache_key = cache_key
 
+        if step.config.cache_policy.expires_after:
+            assert request.start_time
```
Json-Andriopoulos: I would avoid using assert here. If this is a scenario that can happen we should introduce some error handling logic as well.

docs/book/user-guide/best-practices/set-up-your-repository.md
schustmi: Fixed

src/zenml/orchestrators/step_run_utils.py
schustmi: This was not actually a scenario that could happen (only one place where we create these requests, and the start time was always set). I don't even know why the attribute was optional in the first place to be honest, but I made it required now which means I could remove the assert.

Json-Andriopoulos (approved):

src/zenml/client.py
```diff
@@ -4882,6 +4890,26 @@ def list_run_steps(
...
+
+        Args:
+            step_run_id: The ID of the step run to update.
+            cache_expires_at: The time at which this step run should not be
+                used for cached results anymore.
+
+        Returns:
+            The updated step run.
+        """
+        update = StepRunUpdate(cache_expires_at=cache_expires_at)
```
Json-Andriopoulos, line 4908: What concerns me a bit here is that this function may be extended to do other step run updates. But if I use it to set another value won't I set the cache expiration to `None` by default?

Maybe we need a specific function to set the cache expiration to None and only allow actual values here.

For instance have

```
def remove_cache_expiration(step_run_uuid: UUID)
```

schustmi: If you look at `StepRunSchema.update(...)`, we do actually ignore all unset and `None` values in updates.
This means using this class to update other values will not modify the expiration date, however it's a missing feature in general that we do not allow updating any value to `None` 😅

docs/book/how-to/steps-pipelines/sources.md
```diff
@@ -0,0 +1,69 @@
+---
+description: Understanding source roots and source paths
+icon: folders
+---
+
+# Source Code and Imports
+
+When ZenML interacts with your pipeline code, it needs to understand how to locate and import your code. This page explains how ZenML determines the source root directory and how to construct source paths for referencing your Python objects.
+
+## Source Root
```
stefannica, line 10: Great that you decided to document this more thoroughly, this topic is easily misunderstood.

docs/book/user-guide/starter-guide/cache-previous-executions.md
```diff
@@ -159,6 +161,60 @@ Setting `include_step_code=False` can lead to unexpected behavior if you modify
...
+cache_policy = CachePolicy(source_dependencies=[my_helper_function])
+# ..or pass the function source. This also works when
+# configuring the cache policy with a config file
+cache_policy = CachePolicy(source_dependencies=["run.my_helper_function"]) 
+```
+
+#### Cache expiration
+
+By default, any step that executes successfully is a caching candidate for future step runs. Any step with the same [cache key](#understanding-cache-keys) running afterwards
+can reuse the output artifacts produced by the caching candidate instead of actually executing the step code. In some cases however you might want to limit the time for how long
```
stefannica:
```suggestion
can reuse the output artifacts produced by the caching candidate instead of actually executing the step code. In some cases however you might want to limit for how long
```

docs/book/user-guide/starter-guide/cache-previous-executions.md
```diff
@@ -159,6 +161,60 @@ Setting `include_step_code=False` can lead to unexpected behavior if you modify
...
+```
+
+{% hint style="info" %}
+If you want to expire one of your step runs as a cache candidate manually, you can do so by setting it's cache expiration date (in UTC timezone):
+```python
+from zenml import Client
+from datetime import datetime, timezone
+
+now = datetime.now(timezone.utc)
+Client().update_step_run(<STEP-RUN-ID>, cache_expires_at=now)
```
stefannica, line 214: Just a nit: this example is pretty useless unless you can show people how to get to the step run ID. You could include a series of calls to get from pipeline run to step run ID first.

src/zenml/models/v2/core/step_run.py
```diff
@@ -746,4 +781,16 @@ def get_custom_filters(
                 col(StepRunSchema.status) != ExecutionStatus.RETRIED.value
             )
 
+        if self.cache_expired is True:
+            cache_expiration_filter = (
+                col(StepRunSchema.cache_expires_at) < utc_now()
```
stefannica: how does this behave with null `cache_expires_at` values?

src/zenml/orchestrators/cache_utils.py
```diff
@@ -135,6 +143,72 @@ def generate_cache_key(
...
+                    f"your source root ({source_root})."
+                )
+
+            absolute_paths.append(file_path)
+
+        for file_path in sorted(absolute_paths):
+            # Use relative path to source root as key so we don't generate
+            # different hashes on different machines
+            relative_path = os.path.relpath(
+                file_path, source_utils.get_source_root()
```
stefannica:
```suggestion
                file_path, source_root
```

src/zenml/orchestrators/cache_utils.py
```diff
@@ -135,6 +143,72 @@ def generate_cache_key(
...
+            )
+            hash_.update(relative_path.encode())
+            file_content = io_utils.read_file_contents_as_string(file_path)
```
stefannica: This is a horrible way to hash external files for caching purposes. I imagine some users will want to use this feature with bigger files that hold data. Loading the entire file in memory to hash it may lead to OOM crashes. May I suggest:

* hash the file size and "last modification" timestamp if the file exceeds a certain size. In most cases, this is enough.
* implement a function that streams the file content to be hashed in batches instead of loading the entire file in memory.

The least you could do is show a warning if the file exceeds a certain size.

src/zenml/orchestrators/cache_utils.py
```diff
@@ -135,6 +143,72 @@ def generate_cache_key(
...
+    if cache_policy.cache_func:
+        cache_func = source_utils.load(cache_policy.cache_func)
+
+        if not callable(cache_func):
+            raise ValueError(
+                f"The cache function {cache_policy.cache_func.import_path} is "
+                "not callable."
+            )
+
+        result = cache_func()
```
stefannica: I would try/except this, just in case

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -164,6 +170,11 @@ def populate_request(
                 request.status = ExecutionStatus.CACHED
                 request.end_time = request.start_time
 
+                # Cached step runs themselves are not valid candidates for
+                # caching, but just in case we set the cache expiration time to
+                # the time when the cached step run expires.
+                request.cache_expires_at = cached_step_run.cache_expires_at
```
stefannica, line 176: Interesting that you caught this.

docs/book/user-guide/starter-guide/cache-previous-executions.md
schustmi: You could argue that for pretty much any place where users need any ID, how do they get the ID of the run? I don't think all docs pages can always start from zero knowledge, and if this is something that a user cannot figure out, I'm not sure they should be concerned with cache expirations.

src/zenml/models/v2/core/step_run.py
schustmi: Nice catch, I forgot to exclude columns that are `None` in this. Added it!

src/zenml/orchestrators/cache_utils.py
schustmi: Switched to hashing the file in chunks.

stefannica (approved):

PR #4049, opened 2025-10-07
Feature:4015 Log metadata changes

Json-Andriopoulos:
- Allow to specify multiple entities in same log_metadata() call
- Batch submit RunMetadataResource objects
- Allow to attach metadata for multiple artifacts with infer
- Split RunMetadataResource resolution logic for testability

## Describe changes

## Pre-requisites

## Types of changes


src/zenml/models/v2/misc/run_metadata.py
```diff
@@ -28,6 +29,19 @@ class RunMetadataResource(BaseModel):
...
+        """Equality operator."""
+        if not isinstance(other, RunMetadataResource):
+            raise ValueError(
+                f"Can not compare {other.__class__.__name__} to {self.__class__.__name__}"
+            )
+        return hash(self) == hash(other)
+
+    def __hash__(self) -> int:
+        """Hash operator."""
+        # Only safe if _key() is made from immutable values
```
bcdurak: What does `_key()` here refer to?

docs/book/how-to/metadata/metadata.md
```diff
@@ -387,6 +387,51 @@ def downstream(x: int) -> None:
...
+    artifact_name="output",
+    artifact_version="1",
+    model_name="model",
+    model_version="1"
+)
+```
+
+The `log_metadata` function does not support logging metadata for
+multiple entities of the same type. To do it, you can use the ZenML Client
+functionality directly:
```
bcdurak:
```suggestion
The `log_metadata` function does not support logging the same metadata for
multiple entities of the same type at once. To do it, you can use the ZenML Client
functionality directly:
```

src/zenml/utils/metadata_utils.py
```diff
@@ -26,87 +26,236 @@
...
-    *,
-    metadata: Dict[str, MetadataType],
-    step_id: UUID,
-) -> None: ...
+# Manual logging to a step
+log_metadata(metadata={}, step_name=..., run_id_name_or_prefix=...)
+log_metadata(metadata={}, step_id=...)
 
+# Manual logging to a run
+log_metadata(metadata={}, run_id_name_or_prefix=...)
```
bcdurak: Not sure how to fix this without breaking the functionality but I think there is something off here. Previously, if I would provide:

1. Step id -> I log it to the step
2. Run id name or prefix -> I log it to the run
3. Run id name or prefix and step name -> I log it to the step

While the first two stay the same, the third one now logs it not only to the step but also the run. In a setting, where there are a lot of steps, this might end up overloading the pipeline run. Any ideas on how to fix it?

bcdurak: Also, if I provide all three of them together, I think `step_id` will be processed but `step_run_name` and `run_id_name_or_prefix` will be silently ignored. I haven't checked it yet but the same might apply to other entities as well. I am not sure if ignoring it is the best idea here. I was more leaning towards raising a ValueError or in the worst case, throwing a warning.

Json-Andriopoulos: I can add a validation mechanism, sure. So you either pass values by id or name/version. I feel like the best possible solution here is to introduce a new function with a new signature that groups those options to valid parameter groups:

```
def log_metadata(
    step_identifier: StepIdentifier | None,
    model_identifier: VersionedIdentifier | None,
    artifact_identifier: VersionedIdentifier | None,
    pipeline_run_id
):
```

Where for instance identifiers are:

```
class VersionedIdentifier:

    uuid: UUID | None
    name: str | None
    version: str | None

    def validate_options():
        # validate either id or name/version are set
        ...
```

For step we could have something like:

```
class StepRunIdentifier:
     id: UUID | None
     name: str | None
     pipeline_identifier: str | UUID | None
     
    def validate_options():
        # validate either id or name/pipeline are used
        ...
```

With this implementation when you specify the StepRunIdentifier it doesn't get mixed up with the pipeline_run_id, so you can control and separate when it is logged for the pipeline or not.

src/zenml/models/v2/misc/param_groups.py
```diff
@@ -0,0 +1,113 @@
...
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Parameter group classes."""
+
+from uuid import UUID
+
+from pydantic import BaseModel, model_validator
+
+
+class VersionedIdentifier(BaseModel):
```
bcdurak, line 21: Functionally, this would probably be redundant but what do you think about implementing an `ArtifactVersionIdentifier` and a `ModelVersionIdentifier` that simply subclasses this class? I think this would be slightly better for our UX.

src/zenml/models/v2/misc/param_groups.py
```diff
@@ -0,0 +1,113 @@
+#  Copyright (c) ZenML GmbH 2024. All Rights Reserved.
```
bcdurak:
```suggestion
#  Copyright (c) ZenML GmbH 2025. All Rights Reserved.
```

😄

bcdurak: I'll leave this up to you but the name `param_groups` sounds a bit off to me. Maybe something like `identifiers`?

tests/unit/utils/test_metadata_utils.py
```diff
@@ -0,0 +1,119 @@
...
+        len(mock_client.create_run_metadata.call_args.kwargs["resources"]) == 6
+    )
+
+    assert all(
+        isinstance(r, RunMetadataResource)
+        for r in mock_client.create_run_metadata.call_args.kwargs["resources"]
+    )
+
+
+def test_bulk_log_metadata_infer_artifacts():
```
bcdurak: Are these upcoming tests?

src/zenml/utils/metadata_utils.py
```diff
@@ -366,3 +371,158 @@ def log_metadata(
...
+    """Logs metadata for multiple entities in a single invocation.
+
+    Args:
+        metadata: The metadata to log.
+        pipeline_runs: A list of pipeline runs to log metadata for.
+        step_runs: A list of step runs to log metadata for.
+        artifact_versions: A list of artifact versions to log metadata for.
+        model_versions: A list of model versions to log metadata for.
+        infer_models: Flag - when enabled infer model to log metadata for from step context.
+        infer_artifacts: Flag - when enabled infer artifact to log metadata for from step context.
```
bcdurak, line 395: I am wondering whether these could be a part of the respective `VersionedIdentifiers`. In the current version, it is impossible to log metadata for an output artifact of the step and an external artifact at the same time.

bcdurak (commented):
Leaving a few small notes but overall the changes look good. We might need some documentation to reflect the new changes though.

src/zenml/models/v2/misc/param_groups.py
Json-Andriopoulos: I was thinking maybe we can add other similar objects here, some may not be identifiers 🤔

src/zenml/models/v2/misc/param_groups.py
Json-Andriopoulos: Ouch, this will generate many changes :( .

You are right 🤔 . I wouldn't focus on ux that much - I am thinking if we need to extend separately in the future a common signature will be a blocker.

src/zenml/utils/metadata_utils.py
Json-Andriopoulos: I think the distinction we do leads to better usage patterns though. Use `infer` within a function or explicit values out of it makes sense. Why would you log metadata for another step's artifact? Sounds a bit bug-prone.

IMO we should try to guide users to safer usage and if they are restricted by the signatures or the functionality they can always use the client directly (as per the updated docs).

src/zenml/models/v2/misc/param_groups.py
```diff
@@ -0,0 +1,113 @@
...
+
+        return self
+
+
+class StepRunIdentifier(BaseModel):
+    """Class grouping different step run identifiers."""
+
+    id: UUID | None = None
+    name: str | None = None
+    pipeline: PipelineRunIdentifier | None = None
```
schustmi: Should probably be `run`?

src/zenml/models/v2/misc/param_groups.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if not (self.id or self.name):
+            raise ValueError(
+                "You have to use at least one identification option."
+                "Use either id or name."
+            )
+
+        if bool(self.name) ^ bool(self.pipeline):
+            raise ValueError(
+                "To identify a run by name you need to specify a pipeline run identifier."
```
schustmi:
```suggestion
                "To identify a step run by name you need to specify a pipeline run identifier."
```

src/zenml/models/v2/misc/run_metadata.py
```diff
@@ -28,6 +29,31 @@ class RunMetadataResource(BaseModel):
...
+            other: The object to compare.
+
+        Returns:
+            True if the object is equal to the given object.
+
+        Raises:
+            TypeError: If the object is not an instance of RunMetadataResource.
+        """
+        if not isinstance(other, RunMetadataResource):
+            raise TypeError(f"Expected RunMetadataResource, got {type(other)}")
```
schustmi: Any reason this should fail? Usually, python comparisons simply return False in this case, e.g. `"a" == 1`.

As far as I remember, the best practice for overwriting `__eq__` is to implement if in case `other` is actually the correct type, and `return NotImplemented` otherwise (the default handling kicks in in that case).

src/zenml/utils/metadata_utils.py
```diff
@@ -366,3 +371,161 @@ def log_metadata(
...
+        step_context = None
+
+    if (infer_models or infer_artifacts) and step_context is None:
+        raise ValueError(
+            "Infer options can be used only within a step function code."
+        )
+
+    # resolve pipeline runs and add metadata resources
+
+    for pipeline in pipeline_runs or []:
```
schustmi:
```suggestion
    for run in pipeline_runs or []:
```

src/zenml/utils/metadata_utils.py
schustmi: > Why would you log metadata for another step's artifact

No strong opinion on whether we allow it in the bulk operations, but I've seen some usage of this before on Slack and thought I'll mention it here:
Imagine the following very simple pipeline:
```python
@pipeline
def training_pipeline():
  train_data, test_data = load_data()
  model = train(train_data)
  eval(test_data, model)
```
In the eval step, you might want to attach some metadata to the artifact of the model, e.g. some metrics how it performed on the test data.

docs/book/how-to/metadata/metadata.md
```diff
@@ -195,6 +195,102 @@ log_metadata(
 )
 ```
 
+## Bulk Metadata Logging
+
+The log_metadata function does not support logging the same metadata for multiple entities of the same type simultaneously. To achieve this, you can use the bulk_log_metadata function:
```
bcdurak:
```suggestion
The `log_metadata` function does not support logging the same metadata for multiple entities simultaneously. To achieve this, you can use the `bulk_log_metadata` function:
```

src/zenml/models/v2/misc/param_groups.py
bcdurak: Thank you for changing this 👍

docs/book/how-to/metadata/metadata.md
```diff
@@ -195,6 +195,102 @@ log_metadata(
...
+    ],
+    model_versions=[
+        ModelVersionIdentifier(id="<model_version_id>"),
+        ModelVersionIdentifier(name="model_name", version="model_version")
+    ]
+)
+
+~~~
+
+Note that the bulk_log_metadata function has a slightly different signature compared to log_metadata.
```
bcdurak:
```suggestion
Note that the `bulk_log_metadata` function has a slightly different signature compared to `log_metadata`.
```

docs/book/how-to/metadata/metadata.md
```diff
@@ -195,6 +195,102 @@ log_metadata(
...
+
+* VersionedIdentifiers
+  * ArtifactVersionIdentifier & ModelVersionIdentifier
+  * Specify either an id or a combination of name and version.
+* PipelineRunIdentifier
+  * Specify an id, name, or prefix.
+* StepRunIdentifier
+  * Specify an id or a combination of name and a pipeline run identifier.
+
+Similar to the log_metadata function, if you are calling bulk_log_metadata from within a step, you can use the infer options to automatically log metadata for the step’s model version or artifacts:
```
bcdurak:
```suggestion
Similar to the `log_metadata` function, if you are calling `bulk_log_metadata` from within a step, you can use the infer options to automatically log metadata for the step’s model version or artifacts:
```

docs/book/how-to/metadata/metadata.md
```diff
@@ -195,6 +195,102 @@ log_metadata(
...
+    bulk_log_metadata(
+        metadata={"python_version": "3.11", "environment": "macosx"},
+        infer_models=True,
+        infer_artifacts=True
+    )
+    
+    return train_dataset, test_dataset
+```
+
+Keep in mind that when using the infer_artifacts option, the bulk_log_metadata function logs metadata to all output artifacts of the step.
```
bcdurak:
```suggestion
Keep in mind that when using the `infer_artifacts` option, the `bulk_log_metadata` function logs metadata to all output artifacts of the step.
```

src/zenml/utils/metadata_utils.py
bcdurak: +1 to @schustmi. That was the point that I was trying to make.

bcdurak: Depending on whether we change this or not, the docs might need some small updates as well.

Json-Andriopoulos: Ok, got the point. Cool, will change the implementation to enable usage of both modes. This should enable users to log metadata both by infering entities from step context and using provided entities from the arguments. Just keep into account that since these 2 options are separate blocks of execution and manipulate data differently I can not easily and cleanly merge the options by moving infer into VersionedIdentifier definition.

schustmi (approved):

PR #4060, opened 2025-10-10
Fix for inconsistencies between numba and the wrapped print

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/base-package-functionality.yml
+++ b/.github/workflows/base-package-functionality.yml
@@ -27,7 +27,7 @@ on:
       python-version:
         description: Python version
         type: choice
-        options: ['3.9', '3.10', '3.11', '3.12']
+        options: ['3.10', '3.11', '3.12', '3.13']
         required: false
         default: '3.11'
       git-ref:
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -43,10 +43,10 @@ jobs:
         uses: actions/checkout@v4.2.2
         with:
           fetch-depth: 0
-      - name: Set up Python 3.9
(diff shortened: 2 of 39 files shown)


schustmi:
@bcdurak Actually I thought more about this on the weekend, and I think it would be even better to have the wrapper function defined globally in the module, instead of what we're doing now.

Reason is the following:
- Let's say we're in python process A, where ZenML has been imported and wrapped the print function. Because the wrapping happened, `zenml.logger.wrapped_print` now exists and points to the correct wrapper function.
- `numba` (or any other library) now tries to access that same function in the same process, and everything works as expected.
- Now the (potential) issue: What if we're now in a different process, and ZenML has not wrapped the print function (yet). Now, the import would fail, because the `setup_global_print_wrapping()` function hasn't been called, and therefore `zenml.logger.wrapped_print` doesn't exist.

This could be fixed by simply moving the wrapper to the module level. I initially thought this would be complicated because it would use some local variables from the `setup_global_print_wrapping` function, but as far as I can see, it doesn't. So moving it outside is both the clearest and safest option IMO, and doesn't even require the `setattr(...)` workaround.
what do you think?

bcdurak:
@schustmi that makes sense. I have already moved it outside of the function and removed the `setattr(...)` workaround.

The only thing is, we were saving the original print function both locally and in the builtins with the key `_zenml_original_print`, and we used the locally saved one in the wrapped version. Now, I am fetching the one we saved in the `builtins` instead to use it in the globally defined feature. 

I have tested this locally, the solution still seems solid. WDYT?

schustmi (approved):

PR #4061, opened 2025-10-10
Fix for Click 8.2 breaking change

strickvl:
Fixes Click 8.2+ compatibility issues in the ZenML CLI. Click 8.2 introduced breaking changes: removed the `mix_stderr` parameter from `CliRunner`, tightened validation requiring definition list entries to be pairs (not triples), changed `no_args_is_help` behavior, and can set `width` to `None` in non-interactive contexts.

**Changes:** 

- Refactored `ZenFormatter.write_dl()` to detect 2-column vs 3-column rows and handle each appropriately, delegating standard pairs to Click while rendering tagged triples manually.
- Added `_safe_width()` helper for robust width handling when Click returns `None`. 
- Updated CLI root to manually display help when invoked without subcommands, ensuring consistent behavior across Click versions. 
- Created `cli_runner()` helper in test utils that handles Click version differences. 
- Updated all CLI tests to use the helper and verify exit codes. 
- Renamed `tests/integration/functional/utils/` to `tests/integration/functional/utils_functional/` to avoid name shadowing with the `zenml.utils` module. 
- Added CLI testing best practices to `CLAUDE.md`.
- Fixed `test_reusing_private_secret_name_succeeds` to properly handle REST store authentication by creating a fresh client/store after exiting UserContext, preventing credential state pollution.


src/zenml/config/global_config.py
```diff
@@ -701,6 +712,30 @@ def set_store(
...
+        if callable(close_fn):
+            try:
+                close_fn()
+            except Exception as e:
+                logger.debug(
+                    "Error while closing zen store during cleanup: %s",
+                    e,
+                    exc_info=True,
+                )
+        # Clear the cached store reference irrespective of cleanup result.
```
schustmi: There's no real reason to make this code so ugly:
If we want a close (or maybe `disconnect`?) method, it should be defined on the `ZenStoreInterface`, which means you can simply call `store.close()` here. Additionally, you'll also be able to implement that functionality for the Rest store, which can potentially also drain it's connections. Or it simply does nothing.

src/zenml/cli/cli.py
```diff
@@ -158,6 +170,15 @@ def cli() -> None:
         # packages directory
         source_utils.set_custom_source_root(source_root=os.getcwd())
 
+    # Manually show help and exit with code 0 when invoked without a subcommand.
+    # Click 8.2+ raises NoArgsIsHelpError before the callback runs when
+    # no_args_is_help=True. By relying solely on invoke_without_command=True
+    # and handling help here, we ensure consistent behavior across Click
+    # versions while leveraging our custom help formatter in ZenMLCLI.get_help().
+    if ctx.invoked_subcommand is None and not ctx.resilient_parsing:
```
schustmi, line 173: Not sure I really get why this is necessary. Why wouldn't click display the help anymore in newer versions with our previous setup?

src/zenml/cli/cli.py
```diff
@@ -137,17 +137,29 @@ def format_commands(
...
-                with formatter.section(colored_section_title):
-                    formatter.write_dl(rows)  # type: ignore[arg-type]
-
-
-@click.group(cls=ZenMLCLI)
+                if isinstance(formatter, ZenFormatter):
+                    section_title = "[dim cyan]Available ZenML Commands (grouped)[/dim cyan]"
+                    with formatter.section(section_title):
+                        formatter.write_dl(rows)
+                else:
```
schustmi: Is there any case where the formatter isn't ours? I thought that was defined on the context of this class, which means it will always be?

schustmi (changes requested):

src/zenml/cli/cli.py
strickvl: The issue isn't that Click won't display help in newer versions (it will). The problem is that Click 8.2+ would display help using its default formatter instead of our custom `ZenMLCLI.get_help()` formatter. 

(As far as I understand it, without `invoke_without_command=True`, Click defaults to `no_args_is_help=True`. In Click 8.2+, this causes Click to raise `NoArgsIsHelpError` before our `cli()` callback runs, meaning it handles the help display internally with its default formatter, completely bypassing our custom formatting (categories, rich colors, custom layout). By using `invoke_without_command=True` and manually checking for no subcommand in the callback, we ensure our custom formatter is always used across all Click versions.)

I will improve the comment to make it clearer maybe.

src/zenml/cli/cli.py
strickvl: Yes good point. I'll remove the fallback.

src/zenml/config/global_config.py
strickvl: Done.

src/zenml/cli/cli.py
schustmi: It's just that I can't reproduce it. I have click 8.2+, I run `zenml` or `zenml --help`, and everything shows up correct (on develop). So it uses the correct formatter I assume?

schustmi: Also if the code that defined `invoke_without_command=True` gets executed, at that point anyway I should also know about our custom class and use that context/formatter I assume?

strickvl:
Ok the only failing test now is this windows DLL failure on integration tests, but I've seen this failing on other PRs that were merged in, so I guess it can be ignored?

PR #4062, opened 2025-10-14
Upgrading mypy

bcdurak:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -158,8 +158,10 @@ def _reconstruct_nodes(
     )
     for job in job_list.items:
         annotations = job.metadata.annotations or {}
-        if step_name := annotations.get(STEP_NAME_ANNOTATION_KEY, None):
-            node = nodes[step_name]
+        if step_name_annotation_key := annotations.get(
```
schustmi: The result of this is not a key, but a value. It's the value `step_name`, which is what the variable was (and should be) IMO

src/zenml/utils/json_utils.py
```diff
@@ -114,7 +114,7 @@ def pydantic_encoder(obj: Any) -> Any:
 
     if isinstance(obj, BaseModel):
         return obj.model_dump(mode="json")
-    elif is_dataclass(obj):
+    elif is_dataclass(obj) and not isinstance(obj, type):
```
schustmi, line 117: Why was this necessary? This seems like not just a linting adjustment

src/zenml/utils/string_utils.py
```diff
@@ -232,7 +232,7 @@ def substitute_string(value: V, substitution_func: Callable[[str], str]) -> V:
 
             model_values[k] = new_value
 
-        return cast(V, type(value).model_validate(model_values))
+        return cast(V, type(value).model_validate(model_values))  # type: ignore[redundant-cast]
```
schustmi, line 235: I'm assuming we can't get rid of the cast entirely because this doesn't work on all python versions?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
bcdurak: The variable `step_name` is already used at this point, which leads to:

> src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py:161: error: Incompatible types in assignment (expression has type "Any | None", variable has type "str")

You are right though, it is not a key. I have changed it now to `step_name_annotation`, would it be ok for you?

src/zenml/utils/json_utils.py
bcdurak: `is_dataclass` returns True if `obj` is a dataclass itself or if it's an instance of a dataclass. However, the `asdict` in the next line only accepts instances and mypy raises:

> src/zenml/utils/json_utils.py:118: error: Argument 1 to "asdict" has incompatible type "DataclassInstance | type[DataclassInstance]"; expected "DataclassInstance"  [arg-type]

Even in python 3.11, providing a type would lead to an error, so I doubt this changes the functionality. WDYT?

src/zenml/utils/string_utils.py
bcdurak: Not really. If that would be the case, it would lead to an `unused-ignore` error on the other versions. 

As for your question, I was a bit hesitant to remove the redundant casts. We used to do a few tricks with the way we handled pydantic objects and I thought removing the cast might mess up

schustmi: The cast does nothing at runtime, so it certainly wouldn't lead to any issues. But I guess mypy is the issue then.

src/zenml/utils/json_utils.py
schustmi: Ok 👍

schustmi (approved):

PR #4064, opened 2025-10-15
Customizable Deployment server settings

stefannica:
# Framework-Agnostic Deployment App Factory

Implements: https://github.com/zenml-io/zenml/issues/4078

## Overview

This PR implements a framework-agnostic deployment ASGI app factory system that allows customizing all aspects of the ASGI web application that powers up the pipeline deployments:
- app factory abstraction compatible with any ASGI framework (e.g. FastAPI, Django, Flask, Falcon, Quart, BlackSheep, etc.). Currently implemented is only the FastAPI version.
- unified endpoint/middleware specifications make it possible to add endpoints and middleware via configuration independently of ASGI framework
- app extensions make it possible to contribute with framework-specific code to the process of building the ASGI application (e.g. add middleware, endpoints, routes, advanced security mechanisms etc.)

If necessary, the core components of the deployment server - the app factory (aka app runner) and the deployment service classes can be extended and custom implementations can be used instead of the built-in ones via configuration options. 

### `zenml.config.source.SourceOrObject`
A hybrid type that can hold either a source string OR a loaded object:
- Accepts strings, Source objects, or actual importable objects (types, functions, global variables).
- Lazy loading via `load()` method
- Lazy serialization to source strings

### `zenml.config.deployment_config.DeploymentSettings`
Similar to `DockerSettings`, can be used to customize the configuration and behavior of the ASGI application that is run by the pipeline deployment server:
- Contains a rich set of settings, from descriptive to URL paths, to custom endpoints, middleware and extensions
- Relies on `SourceOrObject` to encode the location of classes and functions used as values
- Can only be used at pipeline level

### `zenml.deployers.server.app.BaseDeploymentAppRunner`

This is the abstract factory used to build and run the pipeline deployment ASGI application according to the specifications in the `DeploymentSettings`. This ASGI application is just a wrapper around the core service that implements the pipeline deployment operations (see next point). The responsibilities of this component are:
- Construct an ASGI application using one of the supported frameworks (e.g. FastAPI, Django, Flask, Falcon, Quart, BlackSheep, etc.) and according to the configuration values in the `DeploymentSettings`.
- Provide a unified, framework agnostic API that can be used to extend the ASGI application with custom endpoints and middleware.
- Implement the REST API specific logic around the core deployment service.

The only built-in implementation provided for this class is the one for FastAPI: `zenml.deployers.server.fastapi.app.FastAPIDeploymentAppRunner`.

Users can implement this abstract class to support any other ASGI capable framework.

### `zenml.deployers.server.app.BaseDeploymentAppRunnerFlavor`

This implements a simple flavor system on top of `BaseDeploymentAppRunner` classes, to be able to collect software requirements for app runner implementations independently of the actual implementation.

### `zenml.deployers.server.service.BasePipelineDeploymentService`

This is the base class for the service that runs the core pipeline deployment logic. The responsibilities of this component are:
- Initialize and cleanup the global state shared by all pipeline deployment invocations (e.g. run the init hooks, configure the orchestrator).
- Provide information about the input and output deployment schemas.
- Run one or more actual pipelines according to an invocation request (i.e. input parameters) and return its output.
- Implement some rudimentary health check, status and metrics.
- This code needs to be ASGI / HTTP agnostic.

The only built-in implementation provided for this class is the one that uses the local orchestrator to run pipelines: `zenml.deployers.server.service.PipelineDeploymentService`.

Users can implement this abstract class to provide their own custom implementations of running one or more pipelines.

## Other Changes

* upgrades the `secure` package to latest (1.0.1) and removes all code handling the `X-XSS-Protection` security header that is no longer supported
* adds a `install_deployment_requirements` flag to the `DockerSettings` and includes deployment settings in the set of software requirements used for building container images
* updates existing Deployer stack components (Docker, GCP, AWS) to remove fields from their configurations that are now customizable via the DeploymentSettings (ports, health-check API url paths)
* changes the entrypoint for the deployment container to take in only one argument: the deployment UUID
* moves software requirements for deployment containers from `ContainerizedDeployer.CONTAINER_REQUIREMENTS` to the `BaseDeploymentAppRunnerFlavor` class

# Examples

### Configure URL Paths

```python
from zenml import pipeline
from zenml.config.deployment_settings import DeploymentSettings

deployment_settings = DeploymentSettings(
    # Customize endpoint paths
    invoke_url_path="/predict",
    health_url_path="/healthz",
    metrics_url_path="/api/metrics",
    docs_url_path="/api-docs",
    
    # App metadata
    app_title="Customer Churn Prediction Service",
    app_description="ML service for predicting customer churn",
    app_version="1.2.0",
)

@pipeline(settings={"deployment": deployment_settings})
def my_pipeline():
    ...
```

### Configure CORS

```python
from zenml.config.deployment_settings import (
    DeploymentSettings,
    CORSConfig,
)

deployment_settings = DeploymentSettings(
    cors=CORSConfig(
        allow_origins=[
            "https://myapp.example.com",
            "https://admin.example.com",
        ],
        allow_methods=["GET", "POST", "OPTIONS"],
        allow_headers=["Content-Type", "Authorization"],
        allow_credentials=True,
    ),
)
```

### Configure Security Headers

```python
from zenml.config.deployment_settings import (
    DeploymentSettings,
    SecureHeadersConfig,
)

deployment_settings = DeploymentSettings(
    secure_headers=SecureHeadersConfig(
        hsts="max-age=31536000; includeSubDomains",
        xfo="DENY",
        csp=(
            "default-src 'self'; "
            "script-src 'self' 'unsafe-inline'; "
            "style-src 'self' 'unsafe-inline'; "
            "img-src 'self' data: https:;"
        ),
        referrer="strict-origin-when-cross-origin",
        cache="no-store",
    ),
)
```

### Framework-Agnostic: Simple Endpoint Function

```python
from zenml.config.deployment_settings import (
    DeploymentSettings,
    EndpointSpec,
    EndpointMethod,
)

async def health_detailed() -> Dict[str, Any]:
    """Detailed health check with system metrics."""
    import psutil
    
    return {
        "status": "healthy",
        "cpu_percent": psutil.cpu_percent(),
        "memory_percent": psutil.virtual_memory().percent,
        "disk_percent": psutil.disk_usage('/').percent,
    }

deployment_settings = DeploymentSettings(
    custom_endpoints=[
        EndpointSpec(
            path="/health/detailed",
            method=EndpointMethod.GET,
            handler=health_detailed,
            auth_required=False,
        ),
    ],
)
```

### Framework-Agnostic: Endpoint Builder Function

```python
from typing import Callable
from zenml.deployers.server import BaseDeploymentAppRunner

class PredictionRequest(BaseModel):
    features: list[float]

class PredictionResponse(BaseModel):
    prediction: float
    confidence: float
    model_version: str

def create_custom_predict_endpoint(
    app_runner: BaseDeploymentAppRunner,
    model_path: str,
) -> Callable:
    """Builder function that creates a custom prediction endpoint.
    
    This pattern allows you to access the app_runner and inject
    configuration at build time.
    """
    # Load model during app build time (once)
    import joblib
    model = joblib.load(model_path)
    
    async def predict_endpoint(
        request: PredictionRequest,
    ) -> PredictionResponse:
        """The actual endpoint implementation."""
        prediction = model.predict([request.features])[0]
        confidence = model.predict_proba([request.features]).max()
        
        return PredictionResponse(
            prediction=float(prediction),
            confidence=float(confidence),
            model_version=app_runner.deployment.name,
        )
    
    return predict_endpoint

deployment_settings = DeploymentSettings(
    custom_endpoints=[
        EndpointSpec(
            path="/predict/custom",
            method=EndpointMethod.POST,
            handler=create_custom_predict_endpoint,
            init_kwargs={"model_path": "/models/custom_model.pkl"},
            auth_required=True,
        ),
    ],
)
```

### FastAPI-Specific: Native FastAPI Router

```python
from fastapi import APIRouter, HTTPException, Depends
from zenml.config.deployment_settings import (
    DeploymentSettings,
    EndpointSpec,
    EndpointMethod,
)

# Create a native FastAPI router with all FastAPI features
admin_router = APIRouter(prefix="/admin", tags=["admin"])

@admin_router.get("/stats")
async def get_stats():
    """Get deployment statistics."""
    return {"total_requests": 1000, "uptime_hours": 72}

@admin_router.post("/reload")
async def reload_model():
    """Reload the model."""
    # Implementation here
    return {"status": "reloaded"}

@admin_router.delete("/cache")
async def clear_cache():
    """Clear the cache."""
    # Implementation here
    return {"status": "cleared"}

# Register as native endpoint
deployment_settings = DeploymentSettings(
    custom_endpoints=[
        EndpointSpec(
            path="",  # Router has its own prefix
            method=EndpointMethod.GET,
            handler=admin_router,
            native=True,  # Treat as native FastAPI object
            auth_required=True,
        ),
    ],
)
```

### Framework-Agnostic: Simple Middleware Function

```python
from zenml.config.deployment_settings import (
    DeploymentSettings,
    MiddlewareSpec,
)
import time

class RequestTimingMiddleware:
    """ASGI middleware to measure request processing time.
    
    Uses the standard ASGI interface (scope, receive, send) which works
    across all ASGI frameworks: FastAPI, Django, Starlette, Quart, etc.
    """
    
    def __init__(self, app):
        self.app = app
    
    async def __call__(self, scope, receive, send):
        """Process ASGI request with timing measurement.
        
        Args:
            scope: ASGI connection scope (contains request info).
            receive: Async callable to receive ASGI events.
            send: Async callable to send ASGI events.
        """
        if scope["type"] != "http":
            return await self.app(scope, receive, send)
        
        start_time = time.time()
        
        async def send_wrapper(message):
            """Intercept response to add timing header."""
            if message["type"] == "http.response.start":
                process_time = (time.time() - start_time) * 1000
                headers = list(message.get("headers", []))
                headers.append((
                    b"x-process-time-ms",
                    str(process_time).encode(),
                ))
                message = {**message, "headers": headers}
            
            await send(message)
        
        await self.app(scope, receive, send_wrapper)

deployment_settings = DeploymentSettings(
    custom_middlewares=[
        MiddlewareSpec(
            middleware=RequestTimingMiddleware,
            order=10,
        ),
    ],
)
```

### FastAPI-Specific: Native FastAPI Middleware

```python
from fastapi.middleware.gzip import GZipMiddleware
from zenml.config.deployment_settings import (
    DeploymentSettings,
    MiddlewareSpec,
)

deployment_settings = DeploymentSettings(
    custom_middlewares=[
        # Use native FastAPI GZip middleware
        MiddlewareSpec(
            middleware=GZipMiddleware,
            native=True,
            init_kwargs={"minimum_size": 1000},
            order=100,  # Run late in the chain
        ),
    ],
)
```

### Simple Extension Function

```python
from zenml.deployers.server import BaseDeploymentAppRunner

def add_monitoring_extension(
    app_runner: BaseDeploymentAppRunner,
    prometheus_path: str = "/prometheus",
):
    """Simple extension that adds Prometheus metrics endpoint."""
    from prometheus_client import (
        make_asgi_app,
        Counter,
        Histogram,
    )
    
    # Create Prometheus metrics
    request_counter = Counter(
        'deployment_requests_total',
        'Total requests',
    )
    
    # Mount Prometheus metrics app
    metrics_app = make_asgi_app()
    app_runner.asgi_app.mount(prometheus_path, metrics_app)
    
    print(f"✅ Prometheus metrics available at {prometheus_path}")

deployment_settings = DeploymentSettings(
    app_extensions=[
        AppExtensionSpec(
            extension=add_monitoring_extension,
            extension_kwargs={"prometheus_path": "/metrics/prometheus"},
        ),
    ],
)
```

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/how-to/deployment/deployment_settings.md
+++ b/docs/book/how-to/deployment/deployment_settings.md
@@ -927,10 +927,9 @@ from __future__ import annotations
 
 from typing import Literal, Sequence, Set
 
-from fastapi import Depends, FastAPI, HTTPException, status
+from fastapi import FastAPI, HTTPException, Request, status
 from fastapi.openapi.utils import get_openapi
 from fastapi.responses import JSONResponse
-from fastapi.routing import APIRoute
 from fastapi.security import APIKeyHeader
 
 from zenml.config import AppExtensionSpec, DeploymentSettings
@@ -946,12 +945,10 @@ class FastAPIAuthExtension(BaseAppExtension):
         scheme: Literal["api_key"] = "api_key",
         header_name: str = "x-api-key",
         valid_keys: Sequence[str] | None = None,
-        protect_paths_prefix: str = "/api",
     ) -> None:
         self.scheme = scheme
         self.header_name = header_name
         self.valid_keys: Set[str] = set(valid_keys or [])
-        self.protect_paths_prefix = protect_paths_prefix
 
     def install(self, app_runner: BaseDeploymentAppRunner) -> None:
         app = app_runner.asgi_app
@@ -959,24 +956,26 @@ class FastAPIAuthExtension(BaseAppExtension):
             raise RuntimeError("FastAPIAuthExtension requires FastAPI")
 
         api_key_header = APIKeyHeader(
-            name=self.header_name, auto_error=False
+            name=self.header_name, auto_error=True
         )
 
-        async def verify_api_key(
-            api_key: str | None = Depends(api_key_header),
-        ) -> None:
-            if not api_key or api_key not in self.valid_keys:
-                raise HTTPException(
-                    status_code=status.HTTP_401_UNAUTHORIZED,
-                    detail="Invalid or missing API key",
-                )
+        # Find endpoints that have auth_required=True
+        protected_endpoints = [
+            endpoint.path
+            for endpoint in app_runner.endpoints
+            if endpoint.auth_required
+        ]
 
-        # Attach dependency to protected routes
-        for route in app.routes:
-            if isinstance(route, APIRoute) and route.path.startswith(
-                self.protect_paths_prefix
-            ):
-                route.dependencies.append(Depends(verify_api_key))
+        @app.middleware("http")
+        async def api_key_guard(request: Request, call_next):
+            if request.url.path in protected_endpoints:
+                api_key = await api_key_header(request)
+                if api_key not in self.valid_keys:
+                    raise HTTPException(
+                        status_code=status.HTTP_401_UNAUTHORIZED,
+                        detail="Invalid or missing API key",
+                    )
+            return await call_next(request)
 
         # Auth error handler
         @app.exception_handler(HTTPException)
@@ -1027,7 +1026,6 @@ settings = DeploymentSettings(
                 "scheme": "api_key",
                 "header_name": "x-api-key",
                 "valid_keys": ["secret-1", "secret-2"],
-                "protect_paths_prefix": "/api",
             },
         )
     ]
--- a/src/zenml/deployers/server/app.py
+++ b/src/zenml/deployers/server/app.py
@@ -146,6 +146,10 @@ class BaseDeploymentAppRunner(ABC):
         self.middleware_adapter = self._create_middleware_adapter()
         self._asgi_app: Optional[ASGIApplication] = None
 
+        self.endpoints: List[EndpointSpec] = []
+        self.middlewares: List[MiddlewareSpec] = []
+        self.extensions: List[AppExtensionSpec] = []
+
     @property
     def asgi_app(self) -> ASGIApplication:
         """Get the ASGI application.
@@ -635,6 +639,7 @@ class BaseDeploymentAppRunner(ABC):
                     app_runner=self,
                     **ext_spec.extension_kwargs,
                 )
+            self.extensions.append(ext_spec)
 
     def register_endpoints(self, *endpoint_specs: EndpointSpec) -> None:
         """Register the given endpoints.
@@ -645,6 +650,7 @@ class BaseDeploymentAppRunner(ABC):
         for endpoint_spec in endpoint_specs:
             endpoint_spec.load_sources()
             self.endpoint_adapter.register_endpoint(self, endpoint_spec)
+            self.endpoints.append(endpoint_spec)
 
     def register_middlewares(self, *middleware_specs: MiddlewareSpec) -> None:
         """Register the given middleware.
@@ -655,6 +661,7 @@ class BaseDeploymentAppRunner(ABC):
         for middleware_spec in middleware_specs:
             middleware_spec.load_sources()
             self.middleware_adapter.register_middleware(self, middleware_spec)
+            self.middlewares.append(middleware_spec)
 
     def _run_startup_hook(self) -> None:
         """Run the startup hook.


src/zenml/config/deployment_settings.py
```diff
@@ -0,0 +1,498 @@
...
+    ) -> Callable[..., Any]:
+        """Resolve the extension handler from the spec."""
+        extension = self.extension.load()
+        if not callable(extension):
+            raise ValueError(
+                f"The extension object {extension} must be callable"
+            )
+        return extension
+
+    model_config = ConfigDict(
```
safoinme, line 377: Any reason why `model_config` is defined twice before and after the 2 functions, line 361 and here

src/zenml/deployers/containerized_deployer.py
```diff
@@ -38,7 +38,9 @@
 class ContainerizedDeployer(BaseDeployer, ABC):
     """Base class for all containerized deployers."""
 
-    CONTAINER_REQUIREMENTS: List[str] = []
+    # TODO: this needs to come from the deployment settings or from the
```
schustmi: TODO

src/zenml/integrations/gcp/deployers/gcp_deployer.py
```diff
@@ -1073,13 +1067,16 @@ def do_provision_deployment(
...
+            snapshot.pipeline_configuration.deployment_settings.uvicorn_port
+        )
         container = run_v2.Container(
             image=image,
             command=entrypoint,
             args=arguments,
             env=env_vars,
             resources=resources,
-            ports=[run_v2.ContainerPort(container_port=settings.port)],
+            ports=[run_v2.ContainerPort(container_port=container_port)],
```
schustmi, line 1079: Can we deprecate `settings.port` if it's been replaced by the generic settings?

src/zenml/integrations/aws/deployers/aws_deployer.py
```diff
@@ -1312,8 +1306,11 @@ def do_provision_deployment(
                 f"deploying to App Runner."
             )
 
+        container_port = (
+            snapshot.pipeline_configuration.deployment_settings.uvicorn_port
+        )
         image_config: Dict[str, Any] = {
-            "Port": str(settings.port),
+            "Port": str(container_port),
```
schustmi, line 1313: Can we deprecate settings.port if it's been replaced by the generic settings?

src/zenml/steps/__init__.py
```diff
@@ -26,14 +26,12 @@
...
 """
 
-from zenml.config import ResourceSettings
 from zenml.steps.base_step import BaseStep
 from zenml.steps.step_context import StepContext, get_step_context
 from zenml.steps.step_decorator import step
 
 __all__ = [
     "BaseStep",
-    "ResourceSettings",
```
schustmi, line 36: This can break user code and needs to be mentioned in the release notes + makes this PR breaking, is that intentional?

src/zenml/utils/source_utils.py
```diff
@@ -820,3 +824,198 @@ def _should_load_from_main_module(source: Source) -> bool:
...
+ObjectType = Union[
+    Type[Any],
+    Callable[..., Any],
+    ModuleType,
+    FunctionType,
+    BuiltinFunctionType,
+]
+
+
+class SourceOrObject:
```
schustmi: I love this class and the way you made it work with pydantic!

I think that class might make more sense in `zenml.config.source`, wdyt?

src/zenml/utils/source_utils.py
```diff
@@ -820,3 +824,198 @@ def _should_load_from_main_module(source: Source) -> bool:
...
+        """
+        if self._source is None:
+            if self._object is None:
+                raise RuntimeError(
+                    "SourceOrObject is not initialized with a object"
+                )
+            self._source = resolve(self._object)
+        return self._source
+
+    def to_source_string(self) -> str:
```
schustmi: Any reason you're serializing this to a string? The `Source` object actually contains more information and is used in all other places where we serialize sources

src/zenml/integrations/gcp/deployers/gcp_deployer.py
stefannica: Yes, it's on my list along with some other deployer settings that are now captured in the deployment settings (e.g. API routes for healthchecks)

src/zenml/steps/__init__.py
stefannica: I think it was a mistake that this was listed here and I fixed it. Nobody should import ResourceSettings from this module, given that all our docs and examples use the proper import module for this, and if they do, they will see their mistake and correct it. I can mention it in the release notes, but I won't mark this PR as breaking over such a small thing.

src/zenml/utils/source_utils.py
stefannica: I don't think I understand this question. I need to serialize this to a JSON-able data type (dict, string, list, etc.). This is only used in serialization, so I can't return a `Source` object here instead. It needs to be something that `json.dump(...)` can work with.

src/zenml/config/deployment_settings.py
stefannica: No reason other than my stupid brain didn't see it :smile: Thanks

src/zenml/deployers/containerized_deployer.py
stefannica: I implemented a simple flavor class system to collect the software dependencies at runtime

src/zenml/integrations/gcp/deployers/gcp_deployer.py
stefannica: Done

src/zenml/utils/source_utils.py
stefannica: Done

src/zenml/utils/source_utils.py
schustmi: Maybe I'm missing something, but you're not JSON serializing this by itself, but only as an attribute of other pydantic models, correct? And `Source` itself is a pydantic model, which works just fine if you use `DeploymentSettings.model_dump(...)`?

src/zenml/steps/__init__.py
schustmi: - I don't think it's a subjective discussion "if something is breaking or not". This import was previously part of our public API (which the steps module certainly is), and this PR removes it. So anyone that imported resource settings from here has a broken codebase after the upgrade. That makes this PR and the next release a breaking release, no matter what you or I think.
- This import was definitely not in here by mistake, because I remember when Hamza told me he wanted it there. The reason for this is because resource settings are usually defined on steps, and we wanted both to be importable from the same module for convenience (`from zenml.steps import step, ResourceSettings`).

Just FYI, I actually agree with you that the import shouldn't be here, and I'm all for removing it. Whether we do that as part of this PR or in a huge bulk PR where we do breaking changes whenever we do the next breaking release, I'll leave that up to you.

schustmi (changes requested):

src/zenml/steps/__init__.py
stefannica: > This import was definitely not in here by mistake

I wasn't aware of this. I'll bring it back.

src/zenml/utils/source_utils.py
stefannica: I see what you mean. I'll try it.

stefannica: Yep, this isn't possible because this function is already used in the `PlainSerializer` serialization config and must return a JSON-friendly type.

stefannica: I reworked the entire `SourceOrObject` class so that it actually extends the `Source` class, which feels like a better design.

src/zenml/config/source.py
```diff
@@ -162,6 +164,197 @@ def model_dump_json(self, **kwargs: Any) -> str:
...
+    def is_loaded(self) -> bool:
+        """Whether the object has been loaded.
+
+        Returns:
+            True if the object has been loaded, False otherwise.
+        """
+        return self._is_loaded
+
+    @classmethod
+    def convert_source_or_object(cls, source: Any) -> Any:
```
schustmi: Doesn't this return an item of the class, not `Any`? If so, `-> Self` I think would be better here

src/zenml/config/source.py
```diff
@@ -162,6 +164,197 @@ def model_dump_json(self, **kwargs: Any) -> str:
...
+
+        if isinstance(source, Source):
+            return cls.from_source(source)
+
+        if isinstance(source, dict):
+            return source
+
+        return cls.from_object(source)
+
+    def model_dump(self, **kwargs: Any) -> Dict[str, Any]:
```
schustmi, line 324: There are many more methods that can potentially be called on pydantic objects which won't work. Probably not an issue at the moment, not sure it's going to bite as at some point in the future though. The only real way to work around that is to resolve immediately, which from what I understand you didn't want to do. So we can just keep it as is I think, just FYI.

src/zenml/config/source.py
```diff
@@ -162,6 +164,197 @@ def model_dump_json(self, **kwargs: Any) -> str:
...
+ObjectType = Union[
+    Type[Any],
+    Callable[..., Any],
+    ModuleType,
+    FunctionType,
+    BuiltinFunctionType,
+]
+
+
+class SourceOrObject(Source):
```
schustmi, line 192: Love this!

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -579,6 +588,31 @@ def gather_requirements_files(
                     ", ".join(f"`{r}`" for r in integration_requirements_list),
                 )
 
+        if (
```
schustmi: Something about this feels very wrong to me.

Previously, we've managed to keep all the container image building configuration in a single class (`DockerSettings`), and now we're passing this additional huge `DeploymentSettings` object here, which contains exactly one relevant line (the app runner source). We then try to import the runner class, which most likely contains some requirements that I wouldn't even have in my local environment (because why would I have fastapi/... locally).

I like that this is now as configurable as it is, but I'm wondering whether all this complexity really is necessary. I know you want to make this as simple as possible, but can't we just 
- tell users to include the requirements for their app runner in the `DockerSettings`?
- maybe in `BaseDeployer.requirements(...)`, detect whether it's the default app runner and include the requirements there as part of the stack requirements, which probably covers 99% of use cases where people use the default (or a custom FastAPI) app runner?

src/zenml/config/source.py
stefannica: It is absolutely irrelevant how this `convert_source_or_object` function is annotated, because it's only used internally by `PlainSerializer`. I'll annotate it better nonetheless.

src/zenml/config/source.py
stefannica: The reason why I'm not resolving immediately is that I also want to support lazy-serialization - i.e. serialize the objects to source paths only if serialization is really necessary. I really need this to be able to use these data types to store temporary objects that are not resolvable simply because they cannot be imported. These two methods and the `SourceOrObjectField` custom serialization should cover most cases.

src/zenml/utils/pipeline_docker_image_builder.py
stefannica: > We then try to import the runner class, which most likely contains some requirements that I wouldn't even have in my local environment (because why would I have fastapi/... locally).

There's one misconception here that I'd like to clarify: it's the runner **flavor** class that's being imported here, which acts similarly to the integration class - it contains software dependencies and doesn't need any other packages to be imported.

That said, I do understand that passing around this big object feels wrong. I'll look into passing around a list of extra requirements *alongside the docker settings* extracted and populated early on instead.

stefannica: Ah, splendid, I see there's an `extra_files` setting in the `BuildConfiguration` class, which just happens to be populated by individual stack components based on the snapshot contents. I'll use this instead.

schustmi: Or maybe add `extra_requirements` or something if the files don't work?

PR #4073, opened 2025-10-20
Feature:3963 Step HeartBeat components

Json-Andriopoulos:
- Backend heartbeat support (DB, API)
- Heartbeat monitoring worker

## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/orchestrators/step_launcher.py
+++ b/src/zenml/orchestrators/step_launcher.py
@@ -435,7 +435,7 @@ class StepLauncher:
 
         Raises:
             StepHeartBeatTerminationException: if step heartbeat is enabled and the step is remotely stopped.
-
+            KeyboardInterrupt: Will capture, evaluate and reraise keyboard interrupts.
         """
         from zenml.deployers.server import runtime
 
--- a/src/zenml/zen_stores/migrations/versions/a5a17015b681_add_heartbeat_column_for_step_runs.py
+++ b/src/zenml/zen_stores/migrations/versions/a5a17015b681_add_heartbeat_column_for_step_runs.py
@@ -1,7 +1,7 @@
 """Add heartbeat column for step runs [a5a17015b681].
 
 Revision ID: a5a17015b681
-Revises: 0.90.0
+Revises: af27025fe19c
 Create Date: 2025-10-13 12:24:12.470803
 
 """
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "a5a17015b681"
-down_revision = "0.91.0"
+down_revision = "af27025fe19c"
 branch_labels = None
 depends_on = None
 


src/zenml/steps/heartbeat.py
```diff
@@ -0,0 +1,178 @@
...
+                        self._terminated = True
+                        logger.info(
+                            "%s received HeartBeatTerminationException; "
+                            "interrupting main thread",
+                            self.name,
+                        )
+                        _thread.interrupt_main()  # raises KeyboardInterrupt in main thread
+                    # Ensure we stop our own loop as well.
+                    self._running = False
+                except Exception:
```
Json-Andriopoulos: TODO: Improve this. For sure try to capture HTTP errors in more verbose logs to avoid excessive log generation if the error is for instance server raising 500 status code.

Json-Andriopoulos:
Questions/Comments for reviewers @schustmi @bcdurak :

**Log info records**

I see that in general core components (StepLauncher, StepRunner, etc.) we display a very small number of log records. For better visibility during development I have some log records in the heartbeat worker, should these be removed? I am assuming we display few systemic logs to avoid polluting the user experience as they would be interested in their step function logs only? Some follow-up recommendations would be - a) use structured logs with context variables (https://www.structlog.org/en/stable/) to easily filter records by metadata values b) introduce a systemic logger that is configurable. Suppressed by default, when activated it would present all systemic logs.

**Handling of constants** 

Currently heartbeat interval is hard set as a class variable for the StepHeartBeatWorker cls. For sure I don't want to expose this to user-provided settings as this should be a system setting (too frequent heartbeats from multiple steps may end-up overloading the rest server). I believe a good value would be somewhere in the range of 30-60 seconds. Where would you organize this value? Under `config/constants.py`? In a config object? 

**Interrupt implementation**

I went over our signals/daemonize implementations. While that would be the proper implementation for any unix-based system it is not compatible with Windows. I opted to use _thead.interrupt_main() instead which raises a KeyboardInterrupt exception by default, capture it with a context manager that reraises it with a custom exception. Let me know your thoughts.

src/zenml/constants.py
```diff
@@ -439,6 +439,7 @@ def handle_int_env_var(var: str, default: int = 0) -> int:
 STATUS = "/status"
 STEP_CONFIGURATION = "/step-configuration"
 STEPS = "/steps"
+HEARTBEAT = "heartbeat"
```
schustmi:
```suggestion
HEARTBEAT = "/heartbeat"
```
Probably, because all the other ones have it?

src/zenml/utils/exception_utils.py
```diff
@@ -91,3 +93,41 @@ def collect_exception_information(
...
+    def __init__(
+        self,
+        source_exceptions: list[Type[BaseException]],
+        target_exception: Type[BaseException],
+        message: str,
+        propagate_traceback: bool = True,
+    ) -> None:
+        self._source_exceptions = source_exceptions
+        self._target_exception = target_exception
+        self._message = message
```
schustmi: I guess to simplify this, we could just pass an instance of the exception here instead of the class and message? That would additionally also allow some exceptions which can/need to be instantiated with multiple arguments.

src/zenml/models/v2/core/step_run.py
```diff
@@ -795,3 +808,14 @@ def get_custom_filters(
...
+
+
+# ------------------ Heartbeat Model ---------------
+
+
+class StepHeartbeatResponse(BaseModel):
+    """Light-weight model for Step Heartbeat responses."""
+
+    id: UUID
+    status: str
```
schustmi: This should probably be of type `ExecutionStatus`?

src/zenml/steps/heartbeat.py
```diff
@@ -0,0 +1,158 @@
...
+                except StepHeartBeatTerminationException:
+                    # One-shot: signal the main thread and stop the loop.
+                    if not self._terminated:
+                        self._terminated = True
+                        logger.info(
+                            "%s received HeartBeatTerminationException; "
+                            "interrupting main thread",
+                            self.name,
+                        )
+                        _thread.interrupt_main()  # raises KeyboardInterrupt in main thread
```
schustmi, line 143: My dynamic pipelines PR introduces running multiple steps in different threads, which doesn't work with this I think.

Can we somehow store the thread from which the heartbeat worker was started, and then interrupt that thread instead of the main one?

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -200,6 +202,30 @@ def update_step(
...
+
+    Args:
+        step_run_id: ID of the step.
+
+    Returns:
+        The step heartbeat response (id, status, last_heartbeat).
+    """
+    step = zen_store().get_run_step(step_run_id, hydrate=True)
+    pipeline_run = zen_store().get_run(step.pipeline_run_id)
+    verify_permission_for_model(pipeline_run, action=Action.UPDATE)
```
schustmi: I'm wondering whether this RBAC check is even necessary, as running all of this will take quite some time (two calls to the DB, then a request to the RBAC service).

Is there any real harm in leaving this unprotected? I guess it would allow users potential access to the status of the step, which I'm not sure really is a concern.

src/zenml/steps/heartbeat.py
Json-Andriopoulos: Yeah that is an important change, good point. interrupt_main will not work here, we will need to change the pattern a bit. Should I work my changes from your branch?

src/zenml/zen_server/routers/steps_endpoints.py
Json-Andriopoulos: True, we can probably do both authenication & authorization with pipeline tokens. Will discuss with @stefannica for directions.

stefannica: Let me suggest an alternative: we could limit this endpoint to only be accessed by running pipelines.

Running pipelines (the containerized environment where the steps are running actually) use something called "a workload API token" which is only valid as long as the pipeline run itself is not yet finalized. These workload API tokens are tied to a particular pipeline run (or schedule, in case of scheduled pipelines). So we can also use their scope to limit the range of targets that they can update.

Some references:

* this is the code that verifies the pipeline scoped tokens (you can see some leeway is involved): https://github.com/zenml-io/zenml/blob/main/src/zenml/zen_server/auth.py#L406-L475
* same thing for the schedule-scoped tokens: https://github.com/zenml-io/zenml/blob/main/src/zenml/zen_server/auth.py#L363-L404

A sketch of how you can use this in your endpoint:

```
def update_heartbeat(
    step_run_id: UUID,
    auth_context: AuthContext = Security(authorize),
) -> StepHeartbeatResponse:

    ...
    if not auth_context.access_token or not auth_context.access_token.schedule_id and not auth_context.access_token.pipeline_run_id:
        raise AuthorizationException("Not authorized")

    if auth_context.access_token.pipeline_run_id:
        # optionally, check that the step ID is part of this run ID
    else: # if auth_context.access_token.schedule_id
        # optionally, check that the step ID is part of a run ID that was scheduled with this schedule
```

This will no longer rely on RBAC calls, but it might still flood the database with a lot of requests, so maybe you could also implement a mini-caching system like the ones used in the previous code references, to reduce its impact.

schustmi: @stefannica That was my initial idea as well, but do we use those tokens also when running pipelines with service accounts? I thought at some point we used the API key directly when running scheduled pipelines, but I might be misremembering.

I know for sure though that there is a way to generate a generic unscoped token instead of a workload token when running a pipeline (by setting some token expiration env variable), so we'll have to think about how we handle this case.

stefannica: @schustmi yes, even when running pipelines with service accounts, we generate workload API tokens scoped to the pipeline run or schedule. The only case where we use a generic unscoped token is if you set the `ZENML_PIPELINE_API_TOKEN_EXPIRATION` env variable. But this is a very obscure case, which I don't think we need to handle separately. In that case, we can just not run any RBAC like checks on this endpoint.

schustmi: The problem is that this is a client-side env variable and I'm not sure we can recognize this case in the server endpoint that receives the heartbeat requests. It will just be a generic token that is not scoped to the run, and if we allow those to call the endpoint without any checks then everyone can do so, no?

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -200,6 +203,59 @@ def update_step(
...
+                )
+        elif ctx.access_token.schedule_id:
+            pipeline_run = zen_store().get_run(
+                step_.pipeline_run_id, hydrate=False
+            )
+            schedule = zen_store().get_schedule(
+                ctx.access_token.schedule_id, hydrate=True
+            )
+
+            if pipeline_run.pipeline.id != schedule.pipeline_id:
```
schustmi: Maybe instead of this we can check `pipeline_run.schedule.id == ctx.access_token.schedule_id`, to make sure the run is actually triggered by the schedule that the token is scoped to?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -443,35 +444,59 @@ def _run_step(
...
-                self._run_step_with_step_operator(
-                    step_operator_name=step_operator_name,
-                    step_run_info=step_run_info,
+
+        # To have a cross-platform compatible handling of main thread termination
+        # we use Python's interrupt_main instead of termination signals (not Windows supported).
+        # Since interrupt_main raises KeyboardInterrupt we want in this context to capture it
+        # and handle it as a custom exception.
+
+        with ContextReraise(
```
schustmi: Just a question for clarification: If someone actually interrupts the python process with `CTRL-C`, that raises a `KeyboardInterrupt` in the main thread I assume?

src/zenml/orchestrators/step_launcher.py
```diff
@@ -443,35 +444,59 @@ def _run_step(
...
+            source_exceptions=[KeyboardInterrupt],
+            target_exception=StepHeartBeatTerminationException,
+            message=f"Step {self._invocation_id} has been remotely stopped - terminating",
+            propagate_traceback=False,
+        ) as ctx_reraise:
+            logger.info(
+                f"Initiating heartbeat for step: {self._invocation_id}"
+            )
+
+            StepHeartbeatWorker(step_id=step_run.id).start()
```
schustmi: I'm guessing you don't stop the worker here and instead rely on the server response for the thread to shut down automatically?

Wouldn't it be better to add an explicit stop once the user code finished executing?

src/zenml/orchestrators/step_launcher.py
Json-Andriopoulos: It would be captured the same way and a misleading exception would appear I would think. But is that a scenario we need to worry about?

src/zenml/zen_server/routers/steps_endpoints.py
Json-Andriopoulos: Nice, missed that. Yes, that's better 1 DB call shorter.

src/zenml/orchestrators/step_launcher.py
Json-Andriopoulos: Yeap makes sense to add it. Won't make much of a difference now but it is safer and more future-proof.

src/zenml/orchestrators/step_launcher.py
schustmi: I guess I personally do it quite a lot which is why I thought about it, not sure how common/important it is though. I guess it's quite easy to solve though, so maybe we can account for it?

It could just be a boolean flag on the heartbeat worker that signals that it interrupted the main thread, and we can use that flag to decide whether to re-raise with the HeartbeatInterrupt exception or keep the original KeyboardInterrupt?

Json-Andriopoulos: Yeap, that flag already exists in the heartbeat it is `.is_terminated`. Easy to do, will cover that as well.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -443,35 +444,66 @@ def _run_step(
...
+                    )
+                else:
+                    self._run_step_without_step_operator(
+                        pipeline_run=pipeline_run,
+                        step_run=step_run,
+                        step_run_info=step_run_info,
+                        input_artifacts=step_run.regular_inputs,
+                        output_artifact_uris=output_artifact_uris,
+                    )
+            except StepHeartBeatTerminationException as exc:
```
schustmi: Maybe I'm getting the logic wrong here, but this somehow seems off to me:
- We're in the main thread, and in the outermost layer we have the `ContextReraise` context manager, that converts `KeyboardInterrupt` to `StepHeartBeatTerminationException`.
- Here, we catch the `StepHeartBeatTerminationException` and potentially reraise it as a `KeyboardInterrupt` if the heatbeat worker isn't terminated. This will anyway be converted back to a `StepHeartBeatTerminationException` by the `ContextReraise`

Shouldn't the logic rather be the following:
- No `ContextReraise` needed at all

```python
try:
  heartbeat_worker.start()
  run_step()
except KeyboardInterrupt:
  if heartbeat_worker.is_terminated:
    raise StepHeartBeatTerminationException(...)
  else:
    raise
except:
  ...
finally:
  heartbeat_worker.stop()
```

schustmi (changes requested):

src/zenml/orchestrators/step_launcher.py
Json-Andriopoulos: yy you are right this is much simpler

PR #4074, opened 2025-10-21
Dynamic pipelines v0

schustmi:
## Describe changes

This PR implements dynamic pipelines for the local/kubernetes orchestrators.

### Example
```python

@step
def generate_int() -> int:
  return 3

@step
def do_something(index: int) -> None:
  ...

@pipeline(dynamic=True)
def dynamic_pipeline() -> None:
  count = generate_int()
  # `count` is an artifact, we now load the data
  count_data = count.load()

  for idx in range(count_data):
    # This will run sequentially, like regular python code would.
    # See the features below for an example on how to run steps
    # in parallel.
    do_something(idx)

if __name__ == "__main__":
  dynamic_pipeline()
```

### Features
- Dynamic configuration for steps:
```python
@pipeline(dynamic=True)
def dynamic_pipeline():
  some_step.with_options(enable_cache=False)()
```
- Specify a step runtime: The runtime configures whether the step will be run in the orchestration environment (`runtime=inline`) or if the orchestrator will spin up a separate step execution environment (`runtime=isolated`).
```python
@step(runtime="isolated")
def some_step() -> None:
  ...
```
- Execute multiple steps in parallel by using `step.submit(...)`. This will either execute the step or launch a new container in a new thread.
```python
@step
def some_step(arg: int) -> int:
  ...

@pipeline(dynamic=True)
def dynamic_pipeline():
  future = some_step.submit(arg=1)
  artifact = future.result()  # wait and get artifact response(s)
  data = future.load()  # wait and load artifact data
  downstream_step(future)  # pass the output to another step

  # Run multiple steps in parallel
  for idx in range(3):
    some_step.submit(arg=idx)
```
- Specify config templates for steps using `depends_on`:
```yaml
# config.yaml
steps:
  some_step:
    parameters:
      arg: 3
```

```python
# run.py
@step
def some_step(arg: int) -> None:
  ...

@pipeline(dynamic=True, depends_on=[some_step])
def dynamic_pipeline():
  some_step()

if __name__ == "__main__":
  dynamic_pipeline.with_options(config_path="config.yaml")()
```

### Limitations/Known issues
- Our logging storage isn't threadsafe yet, which means logs of parallel steps are mixed up.
- When running multiple steps concurrently, failure in one step does not stop the other steps. Instead, they continue executing until finished.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/pipelines/dynamic/pipeline_definition.py
+++ b/src/zenml/pipelines/dynamic/pipeline_definition.py
@@ -174,8 +174,8 @@ class DynamicPipeline(Pipeline):
             )
 
         logger.warning(
-            "Dynamic pipelines are currently an experimental stage. There "
-            "might be missing features, bugs and the interface is subject to "
+            "Dynamic pipelines are currently an experimental feature. There "
+            "are known issues and limitations and the interface is subject to "
             "change. If you encounter any issues or have feedback, please "
             "let us know at https://github.com/zenml-io/zenml/issues."
         )


src/zenml/execution/pipeline/dynamic/outputs.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def artifacts(self) -> StepRunOutputs:
+        """Get the step run output artifacts.
+
+        Returns:
+            The step run output artifacts.
+        """
+        return self._wrapped.result()
+
+    def load(self) -> Any:
```
Json-Andriopoulos: Perhaps we should keep a cache dictionary for materialized artifacts, so if (for some weird reason) users do multiple load calls (instead of assigning and reusing) we can return the cached results.

src/zenml/execution/pipeline/dynamic/outputs.py
```diff
@@ -0,0 +1,221 @@
...
+        if result is None:
+            return None
+        elif isinstance(result, ArtifactVersionResponse):
+            return result.load()
+        elif isinstance(result, tuple):
+            return tuple(item.load() for item in result)
+        else:
+            raise ValueError(f"Invalid step run output: {result}")
+
+    def __getitem__(self, key: Union[str, int]) -> ArtifactFuture:
```
Json-Andriopoulos: Maybe we should provide a custom implementation for __setitem__ as a better UX/guidance to users than getting an error like this: `object does not support item assignment`.

src/zenml/execution/pipeline/dynamic/outputs.py
```diff
@@ -0,0 +1,221 @@
...
+        Raises:
+            ValueError: If the key or index is of an invalid type.
+            IndexError: If the index is out of range.
+
+        Returns:
+            The artifact future.
+        """
+        if isinstance(key, str):
+            index = self._output_keys.index(key)
+        elif isinstance(key, int):
```
Json-Andriopoulos: I would recommend adopting one kind of wrapper behavior and stick to it. If we enable users to access futures both dict-style and list/tuple style I think the more we extend the class magic functions the more the behavior may be more confusing regarding what gets executed and what the users expects to be executed.

Plus in general accessing by key is a safer operation and would be a good pattern to enforce. The results may change length, order and accessing by position is a bit un-safe.

src/zenml/execution/pipeline/dynamic/outputs.py
```diff
@@ -0,0 +1,221 @@
...
+        if index > len(self._output_keys):
+            raise IndexError(f"Index out of range: {index}")
+
+        return ArtifactFuture(
+            wrapped=self._wrapped,
+            invocation_id=self._invocation_id,
+            index=index,
+        )
+
+    def __iter__(self) -> Any:
```
Json-Andriopoulos, line 223: Would `__contains__` make sense to also implement here? 🤔

src/zenml/execution/pipeline/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@contextmanager
+def prevent_pipeline_execution() -> Generator[None, None, None]:
+    """Context manager to prevent pipeline execution.
+
+    Yields:
+        None.
+    """
+    with env_utils.temporary_environment(
```
Json-Andriopoulos: Not in the context of this PR but we can do a better job here, thread-safety wise. In general os.environ should be treated as an immutable value, changing its values may also affect un-intentionally other execution paths (for instance code running in other threads).

I think masking the environment under a custom object/class and making that accessible with context vars resolves the issue (context vars are thread-local, wrapper object loads and freezes state and exposes all the operations we want with relative safety).

src/zenml/execution/pipeline/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def should_prevent_pipeline_execution() -> bool:
+    """Whether to prevent pipeline execution.
+
+    Returns:
+        Whether to prevent pipeline execution.
+    """
+    return handle_bool_env_var(
+        ENV_ZENML_PREVENT_PIPELINE_EXECUTION, default=False
```
Json-Andriopoulos: It is hard to track those environment variables references. Maybe a BaseSettings object to organize those in a single place would be a good idea. Also we wouldn't need this function, pydantic would validate the boolean value for us in one go.

src/zenml/execution/step/utils.py
```diff
@@ -0,0 +1,103 @@
...
+            snapshot=snapshot,
+            step=step,
+            orchestrator_run_id=orchestrator_run_id,
+        )
+        return launcher.launch()
+
+    if not retry:
+        step_run = _launch_without_retry()
+    else:
+        retries = 0
```
Json-Andriopoulos, line 70: Would it make sense to make this a utility function? Implementation-wise it is the same with `base_orchstrator.py` lines 425-430. I think these are good sensible defaults and could be shared.

🤔 if it would make sense to instantiate StepRetryConfig by default with those options (instead of optional, have a default object) and then use them directly.

src/zenml/execution/step/utils.py
```diff
@@ -0,0 +1,103 @@
...
+        retry_config = step.config.retry
+        max_retries = retry_config.max_retries if retry_config else 0
+        delay = retry_config.delay if retry_config else 0
+        backoff = retry_config.backoff if retry_config else 1
+
+        while retries <= max_retries:
```
Json-Andriopoulos, line 76: This could also make sense to abstract and re-use 🤔 .

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -178,6 +178,10 @@ class PipelineRunUpdate(BaseUpdate):
         max_length=STR_FIELD_MAX_LENGTH,
     )
     end_time: Optional[datetime] = None
+    finished: Optional[bool] = Field(
```
Json-Andriopoulos: Maybe `is_finished` would be more consistent with the naming patterns we follow.

src/zenml/execution/pipeline/dynamic/outputs.py
Json-Andriopoulos: For instance, I can get dict-style (by specifying key) but the `__iter__` function I may expect it to return the keys not the values. For better consistency I would say we have 2 options:

1. No magic functions - We provide public methods with clean documentation. Uses can manipulate futures based on the available functions and their signatures. No confusion, users stick to following docstrings.
2. We implement magic functions but we assume implementation wraps one kind of hidden data structure. Users can manipulate it as dict (preferable imo) or as a list/tuple.

src/zenml/execution/step/utils.py
schustmi: Doesn't the `base_orchestrator.py` now use this util function instead? I actually just copied it into a separate file because I needed it in another place, but yes the code is identical.

src/zenml/execution/pipeline/utils.py
schustmi: Yep agreed in general that would be a nice thing. We even have some classes that do this for a subset of env variables (`ServerConfiguration` and `GlobalConfiguration`), but not for all of them. This function also treats values like `"yes"` as True, not sure how pydantic would handle that natively, but we could implement it for sure.

src/zenml/execution/pipeline/utils.py
schustmi: Yes you're right! This is not a problem in this case, but when running multiple steps in parallel we also set different environment variables which is problematic. At least in this case I can switch to using a context var instead

src/zenml/execution/pipeline/dynamic/outputs.py
schustmi: ~~Oh yes I thought I already did, that definitely makes sense!~~
Actually if we do implement it as a tuple-like data structure the contains will be with values and might not make much sense, what do you think?

src/zenml/execution/pipeline/dynamic/outputs.py
schustmi: I like that idea. I'll also add a boolean flag to the `load` method to prevent this, so users can at least manually disable it for huge artifacts.

src/zenml/execution/pipeline/dynamic/outputs.py
schustmi: Makes sense, I agree we should only support one way. My vote goes for tuple like behaviour though, as the main purpose of my implementations was to actually support the following: When calling any python function with multiple outputs, there are two cases:
```python
def f():
  return 1, "str"

# Return value is a tuple that can be accessed with `int` keys
tuple_result =  f()
int_result = tuple_result[0]

# Automatic unpacking
int_result, str_result = f()
```
I think the latter is most common, and is the use-case I think we should support to make it feel as pythonic as possible (and also mirrors how you would call sync steps, in which case the return value will not be a future but instead of tuple of artifacts).

Tuples are also immutable which in our case holds true as well as you can't add outputs to the future result of a step run.

We can then add some helper methods like `get_output(key: str)` like you suggested to get allow fetching specific outputs by key.

Json-Andriopoulos: I see, yes, tuple should do it as well. As long as we are consistent I think we are ok.

src/zenml/execution/pipeline/dynamic/outputs.py
Json-Andriopoulos: yy it would make more sense in a dict-like scenario!

src/zenml/execution/pipeline/utils.py
Json-Andriopoulos: We can work something here, but different story of course. ContextVars should work perfectly, maybe in combo with a centralized BaseSettings object. Will create the story and we can discuss implementation.

src/zenml/execution/pipeline/utils.py
Json-Andriopoulos: I think Pydantic captures multiple values as well, I can cross compare. Will open a new story for this just wanted to get your opinion :)

src/zenml/execution/step/utils.py
Json-Andriopoulos: Ah I see what you mean my bad. Resolving.

PR #4097, opened 2025-10-24
Various deployment fixes and improvements

stefannica:
* fixed the GCP deployer service name max size
* fixed a bug that prevented the previous port to be reused when a local deployment was re-provisioned
* adde the uvicorn reload setting for local deployments - this helps speed up local development, because the deployment doesn't have to be reprovisioned with every change that is made to the pipeline/step code
* reworked the app runner to correctly use sub-processes for multiple workers and for the reload function - uvicorn needs the ASGI app to be provided as a source path for these cases and therefore it has to be built inside the context of the uvicorn process (i.e. the so called factory mode) instead of outside
* fixed the concurrency issue that was creating extra pipeline runs when multiple invocation requests were made in parallel. Instantiating a deployment orchestrator for each request rather than sharing one solved the problem.
* added a connection pool size configuration setting to the REST zen store and configured it to match the ASGI app thread count - this speeds up running multiple parallel deployment invocation requests and avoids polluting the logs with warnings
* implemented client-side healthcheck for all deployments

Fixes:

* https://github.com/zenml-io/zenml/issues/3999
* 

## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/deployers/server/service.py
+++ b/src/zenml/deployers/server/service.py
@@ -317,6 +317,7 @@ class PipelineDeploymentService(BasePipelineDeploymentService):
         self.service_start_time = time.time()
         self.last_execution_time: Optional[datetime] = None
         self.total_executions = 0
+        self.orchestrator_class = SharedLocalOrchestrator
 
         try:
             # Execute init hook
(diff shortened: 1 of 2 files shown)


safoinme (approved):
only minor issue with the failing test because we are calling `service._orchestrator = _DummyOrchestrator()`  and we don't have _orchestrator anymore

PR #4101, opened 2025-10-27
Add stack to deployment response resource

safoinme:
## Describe changes

## Pre-requisites

## Types of changes


src/zenml/models/v2/core/deployment.py
```diff
@@ -212,6 +213,11 @@ class DeploymentResponseResources(ProjectScopedResponseResources):
         default_factory=list,
         title="Curated deployment visualizations.",
     )
+    stack: Optional["StackResponse"] = Field(
```
schustmi, line 216: Do we need this for the frontend? Otherwise, in code it would also be quite easy to just fetch it from the snapshot that is already passed as part of the resources.

safoinme: Yes, this is mainly for front-end

safoinme:
@schustmi does it make sense to do `include_metadata=True` for the stack so we get full components so it shows up in front-end? or is there an alternative to fetch that in front via ID of stack?

schustmi:
> @schustmi does it make sense to do `include_metadata=True` for the stack so we get full components so it shows up in front-end? or is there an alternative to fetch that in front via ID of stack?

@safoinme No I don't think that makes sense. Anything in `resources` is useful to show in a list of items (e.g. show the stack as a column in the deployment list), but if the frontend needs more information (like the components) on a detail page, it can easily send another request to fetch that.

PR #4104, opened 2025-10-27
Adjust release process after quickstart changes

schustmi:
## Describe changes
The new quickstart doesn't include running on a cloud stack anymore. This PR removes any files/code that was still building Docker images for that purpose.

Note: We still run the quickstart on cloud stacks as part of the release process.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/deployers/local.md
+++ b/docs/book/component-guide/deployers/local.md
@@ -52,6 +52,7 @@ For additional configuration of the Local deployer, you can pass the following `
   * `port_range`: The range of ports to search for a free port. Defaults to `(8000, 65535)`.
   * `address`: The address that the deployment server will listen on. Defaults to `127.0.0.1`.
   * `blocking`: Whether to run the deployment in the current process instead of running it as a daemon process. Defaults to False. Use this if you want to debug issues with the deployment ASGI application itself.
+  * `auto_reload`: Whether to enable auto-reload for the uvicorn server. This is useful to speed up local development by automatically restarting the server when code changes are detected without requiring a re-provisioning of the entire deployment. Defaults to False. NOTE: the `auto_reload` setting has no effect on changes in the pipeline configuration, step configuration or stack configuration.
 
 Check out [this docs page](https://docs.zenml.io/concepts/steps_and_pipelines/configuration) for more information on how to specify settings.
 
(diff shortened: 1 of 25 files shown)


release-cloudbuild-preparation.yaml
```diff
@@ -77,7 +77,6 @@ steps:
           --platform linux/amd64 \
           --build-arg BASE_IMAGE=$$USERNAME/prepare-release:base-${_ZENML_NEW_VERSION} \
           --build-arg CLOUD_PROVIDER=gcp \
-          --build-arg ZENML_BRANCH=${_ZENML_BRANCH} \
```
bcdurak, line 80: I am confused about all this. Maybe you can clear it for me. I see now that we don't even build any quickstart images for the release (which I like). But if that's the case, why do we even need to build them now for testing the release? Isn't ok for us to use the base image to test?

Previously, different quickstarts required a different "pre-specified" list of requirements to run the quickstart. As far as I can see, this is no longer a concern of the new one.

.github/workflows/release_prepare.yml
```diff
@@ -100,10 +100,6 @@ jobs:
             fi
           done
           git add examples/quickstart
-      - name: Regenerate quickstart cloud requirements
```
bcdurak, line 103: As far as I can see this used to generate `examples/quickstart/requirements_{aws,azure,gcp}.txt` files for the quickstart testing script. Since this is now disabled (which was the right call I think), wouldn't the `qs_run_release_flow` fail?

It has a check:

```bash
...
REQS="requirements_${CLOUD}.txt"
...
if [[ ! -f "${REQS}" ]]; then
  echo "Error: Requirements file not found: ${REQS}" >&2
  exit 1
fi
...
``

bcdurak: Same script also looks for config files that do not exist.

bcdurak: The installation seems a bit weird as well. The runner starts, checks out to the branch, already install ZenML once, passes the branch name to the test script, the test script manually replaces the hard-coded zenml requirement (which is the only requirement), pip installs zenml throught the `requirements.txt` file again and only then runs the example. This seems like a very complicated and redundant process.

bcdurak (changes requested):
I am quite confused by the current state of this workflow right now. Maybe I am missing something, but this definitely needs to be tested with a dummy release branch before being merged in.

.github/workflows/release_prepare.yml
schustmi: Yep, this whole thing has been a mess. I removed as much of it as I could now.

release-cloudbuild-preparation.yaml
schustmi: These image still installed the stack requirements in there, but I don't even understand why that was necessary back then. I removed all of that now though.

Issue #4092, opened 2025-10-24
Add enterprise authentication methods to the Hashicorp Vault secrets store


stefannica:
### Feature Description

Currently, direct tokens is the only authentication method supported by the ZenML Hashicorp vault secrets store implementation. Furthermore, the token has to be long lived, because auto-renewal is also not supported. So users would need to generate a token with a TTL that is something like 3 months or 6 months, which is not recommended as it comes with security risks, and then also rotate the token regularly, which is not maintainable.

There are other authentication methods available that are safer for production use. The client also needs to support auto-refreshing temporary tokens before they expire.

### Priority

None

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #4110, opened 2025-10-29
Add more Hashicorp Vault secrets store auth methods

stefannica:
## Describe changes
We currently only support the token authentication method for the Hashicorp Vault secrets store. This authentication method is unusable in production because tokens are usually set to expire after a short time and managing their rotation is painful.

This PR adds a few more authentication method options and also implements automatic token refresh / re-login upon expiration:

* [App Role authentication](https://python-hvac.org/en/stable/usage/auth_methods/approle.html) - authentication using a Vault App Role (app role ID and secret ID)
* [AWS authentication](https://python-hvac.org/en/stable/usage/auth_methods/aws.html) - implicit authentication using an AWS IAM role (to be used for ZenML Pro workspaces)

Implements: https://github.com/zenml-io/zenml/issues/4092

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/secrets_stores/hashicorp_secrets_store.py
+++ b/src/zenml/zen_stores/secrets_stores/hashicorp_secrets_store.py
@@ -14,7 +14,6 @@
 """HashiCorp Vault Secrets Store implementation."""
 
 from datetime import datetime, timedelta
-from enum import StrEnum
 from typing import (
     Any,
     ClassVar,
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/secrets_stores/hashicorp_secrets_store.py
```diff
@@ -111,21 +147,101 @@ def client(self) -> hvac.Client:
...
+            """
+            expires_in = response.get("auth", {}).get("lease_duration")
+            renewable = response.get("auth", {}).get("renewable", False)
+
+            if expires_in:
+                self._expires_at = datetime.now() + timedelta(
+                    seconds=expires_in * 0.8  # 80% of the lease duration
+                )
+
+            if renewable and expires_in:
```
schustmi, line 169: I think in this case we also want to set `self._renewable`?

stefannica: Correct. I even realized this was sort of redundant and dropped it completely in favor of using just `_renew_at`.

schustmi (approved):

PR #4111, opened 2025-10-30
Log store abstraction

bcdurak:
This is a large PR, so I would advise you to read the whole description before starting the review.

## What do we capture now?

Previously, ZenML used to add a new handler to the root logger. This root logger was used to capture all the logs that goes through the root logger and store them in the artifact store. Additionally, we used to wrap the built-in `print` function in a way, that we stored the `print`ed messages as well. However, in this case, we missed on a couple of sources such as messages from loggers that do not propagate to the root logger, anything on the stdout/stderr aside from log messages and `print` statements.

Now, we do the following: 

- `stdout` and `stderr` are now wrapped. we keep the original `stdout` and `stderr`.
- everything that goes through the new wrapped `stdout/stderr` still go through the original `stdout/stderr`.
- however, additionally, every message is also directed a `classmethod` called `LoggingContext.emit(...)`. (will be explained in the following section).
- we still add a handler to the root logger. With this the root logger ends up having two handler: the `console_handler` and the `zenml_handler`. 
- While the `console_handler` formats and writes things to the console, the `zenml_handler` is responsible for routing all the incoming log messages to the `LoggingContext.emit(...)` as well.

Screenshot 2025-12-04 at 10 25 04

## The new `LoggingContext` class

We have a new `LoggingContext` class now that replaces the old `PipelineLogsContext`. It's still a context manager, but operates a bit differently. 

When you `__init__` this class, it stores the reference to the log store within your active stack. Every the `__enter__` method gets called, it checks a context variable called `active_logging_context`, if there is one, it stores it and replaces the context variable with itself. Similarly, when `__exit__` gets called, it removes itself from the context variable and puts the old value back. 

One of the most critical parts is the fact that you require a `LogsResponse` to initiate a `LoggingContext` now. So, when we ultimately call the `emit(...)` classmethod, it passes the message and active logging context (along with the corresponding`LogsResponse`) to the `emit(...)` method of the log store.

# The new `LogStore` component

We have a new type of stack component called a `LogStore`. It handles log collection and retrieval . Different implementations can plug into this interface to provide different storage backends without changing how logs are captured or accessed.

This PR also introduces three layers of implementation:

### 1. Layer: `BaseLogStore`

This layer introduces the main abstraction for the new stack component. Main abstract methods include:

- `emit(...)`: receives log records and sends them to a specific backend
- `fetch(...)`: retrieves stored logs for the dashboard and API based on time filters and limits
- `finalize(...)`: finalizes the stream of logs associated with a specific log response

### 2. Layer: `OtelLogStore`

This is yet another abstraction built on the base log store that implements the core OTEL infrastructure:

- `emit(...)`: Activates the log store if not yet activated and translates log recored objects into OTEL format with zenml-specific attributes (e.g., zenml.log_id, zenml.log_uri, zenml.log_store_id) and emits them through the OTEL logger
- `activate(...)`: Sets up the OpenTelemetry pipeline including the LoggerProvider, BatchLogRecordProcessor, and LoggingHandler
- `deactivate(...)`: Flushes pending logs and shuts down the processor and its background thread

Moreover, It introduces configuration options for the OTEL-standardized logs including: `service_name`, `service_version`, `max_queue_size`, `schedule_delay_millis`, `max_export_batch_size`

The following abstract methods are exposed and must be implemented by subclasses:

- `get_exporter(...)`: Returns the specific `LogExporter` instance for the backend
- `fetch(...)`: Backend-specific log retrieval (since each backend has different query mechanisms)

#### ArtifactLogStore

The artifact log store writes logs directly to the artifact store, providing a zero-configuration logging solution that works out of the box:
- Uses a custom `ArtifactLogExporter` that writes `LogEntry` objects to the artifact store (compatible with our previous approach)
- Handles both mutable filesystems (single file with append) and immutable filesystems like GCS (directory with timestamped files that get merged on finalization)
- Automatically chunks large messages (>5KB) to prevent storage issues with UTF-8 boundary handling
- Implements fetch(...) by streaming log files line-by-line from the artifact store
- Can be created automatically from an existing artifact store via `from_artifact_store(...)` class method
- Supports log finalization via an `END_OF_STREAM_MESSAGE` marker that triggers file merging on immutable filesystems and version removal on others

#### DatadogLogStore

- The Datadog log store exports logs to Datadog's HTTP intake API using OTLP:
- Uses an `OTLPLogExporter` configured with Datadog's OTLP endpoint
- Requires an `api_key` for log ingestion and an `application_key` for log retrieval
- Implements fetch(...) using Datadog's Logs Search API (/api/v2/logs/events/search) with:
- Query filtering by `service` and `zenml.log_id`

# Interaction of the stack with log stores

Similar to the image builders, if you don't have a log store within your active stack, an `ArtifactLogStore` flavor will be used instead. Since our default approach requires the `opentelemetry-sdk`, it is added to the pyproject.toml as an additional dependency to the base package.

## Various other changes

- There was a context variable called `redirected` which was defaulted to `False` and never used afterwards. That is now removed.
- We changed the way prepare the logs uri for the artifact log store.

## Notes

- Recheck the flush/deactivate solution.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/log_stores/datadog/datadog_flavor.py
+++ b/src/zenml/log_stores/datadog/datadog_flavor.py
@@ -15,13 +15,16 @@
 
 from typing import Type
 
-from pydantic import Field
+from pydantic import Field, field_validator
 
 from zenml.enums import StackComponentType
 from zenml.log_stores import BaseLogStore, BaseLogStoreConfig
 from zenml.log_stores.otel.otel_flavor import OtelLogStoreConfig
 from zenml.stack.flavor import Flavor
 from zenml.utils.secret_utils import PlainSerializedSecretStr
+
+# Datadog API limits: https://docs.datadoghq.com/api/latest/logs/
+DATADOG_MAX_BATCH_SIZE = 1000
 
 
 class DatadogLogStoreConfig(OtelLogStoreConfig):
@@ -31,6 +34,7 @@ class DatadogLogStoreConfig(OtelLogStoreConfig):
         api_key: Datadog API key for log ingestion.
         application_key: Datadog application key for log extraction.
         site: Datadog site (e.g., "datadoghq.com", "datadoghq.eu").
+        max_export_batch_size: Maximum batch size for exports (Datadog limit: 1000).
     """
 
     api_key: PlainSerializedSecretStr = Field(
@@ -43,6 +47,31 @@ class DatadogLogStoreConfig(OtelLogStoreConfig):
         default="datadoghq.com",
         description="Datadog site (e.g., datadoghq.com, datadoghq.eu)",
     )
+    max_export_batch_size: int = Field(
+        default=500,
+        description="Maximum batch size for exports (Datadog limit: 1000)",
+    )
+
+    @field_validator("max_export_batch_size")
+    @classmethod
+    def validate_max_export_batch_size(cls, v: int) -> int:
+        """Validate that max_export_batch_size doesn't exceed Datadog's limit.
+
+        Args:
+            v: The value to validate.
+
+        Returns:
+            The validated value.
+
+        Raises:
+            ValueError: If the value exceeds Datadog's limit.
+        """
+        if v > DATADOG_MAX_BATCH_SIZE:
+            raise ValueError(
+                f"max_export_batch_size cannot exceed {DATADOG_MAX_BATCH_SIZE} "
+                f"(Datadog API limit). Got: {v}"
+            )
+        return v
 
 
 class DatadogLogStoreFlavor(Flavor):


src/zenml/log_stores/artifact/artifact_log_store_flavor.py
```diff
@@ -0,0 +1,85 @@
...
+        return "https://docs.zenml.io/stack-components/log-stores/artifact"
+
+    @property
+    def logo_url(self) -> str:
+        """URL to the flavor logo.
+
+        Returns:
+            The URL to the flavor logo.
+        """
+        # TODO: Add a logo for the artifact log store
```
bcdurak: Reminder to myself.

src/zenml/log_stores/datadog/datadog_flavor.py
```diff
@@ -0,0 +1,111 @@
...
+    @property
+    def sdk_docs_url(self) -> str:
+        """URL to the SDK docs for this flavor.
+
+        Returns:
+            The URL to the SDK docs for this flavor.
+        """
+        return self.docs_url
+
+    # TODO: Add logo for the Datadog log store
```
bcdurak: Reminder again.

tests/integration/functional/zen_stores/test_zen_store.py
```diff
@@ -84,7 +84,6 @@
     EntityExistsError,
     IllegalOperationError,
 )
-from zenml.logging.step_logging import fetch_log_records, prepare_logs_uri
```
bcdurak, line 87: Reminder to fix

src/zenml/orchestrators/step_launcher.py
```diff
@@ -271,107 +273,78 @@ def launch(self) -> StepRunResponse:
...
-                    ):
-                        force_write_logs = (
-                            logs_context.storage.send_merge_event
+                logs_context = nullcontext()
+                if step_run.logs:
+                    logs_context = LoggingContext(log_model=step_run.logs)
+
+                with logs_context:
+                    try:
+                        # TODO: We still need to apply the fix for step operators here
```
bcdurak: This is an open question as well. Currently, this is not a problem as we do not have a merge mechanism for the artifact log store. But we need to think about this.

stefannica:
Idea: have a default log store of type `artifact` and add it to the default stack.

stefannica:
Another idea: at some point, we might want to use this OTEL logging feature to log messages *inside the ZenML server itself*. Just like we use secrets stores. The implementation should allow us to easily do that - i.e. instantiate one log store instance inside the ZenML server and use a logging context to capture all the server's logs.

stefannica:
Bug report: logging.exception calls do not capture the trace information.

With this step code:

```
     try:
        raise Exception("This is an exception from step one")
    except Exception as e:
        logging.exception("step one: Exception caught")
```

* on my client, I see this:

```
step one: Exception caught
  File "/home/silent-assassin/aspyre/src/zenml/examples/as_simple_as_it_gets/run.py", line 52, in step_one
    raise Exception("This is an exception from step one")
Exception: This is an exception from step one
```

* in the step logs, all I see is this

image

stefannica:
@bcdurak I could not figure this out on my own: how do you prevent log entries emitted by the log store and artifact store and adjiacent libraries like s3fs from being processed themselves and eventually cause a cascade reaction that floods your log storage with entries in an infinite loop ?

Update: it turns out this wasn't covered yet, but now it is.

src/zenml/zen_stores/schemas/logs_schemas.py
```diff
@@ -65,19 +65,24 @@ class LogsSchema(BaseSchema, table=True):
...
+        nullable=True,
+    )
+    log_store_id: Optional[UUID] = build_foreign_key_field(
+        source=__tablename__,
+        target=StackComponentSchema.__tablename__,
+        source_column="log_store_id",
+        target_column="id",
+        ondelete="CASCADE",
+        nullable=True,
     )
```
stefannica, line 83: would it make sense to make this a non-nullable field and populate it during the DB migration with the UUID of a default log store also created during the DB migration and added to the default stack ?

src/zenml/stack/stack.py
```diff
@@ -517,6 +528,44 @@ def deployer(self) -> Optional["BaseDeployer"]:
...
+                name="default",
+                flavor=flavor.name,
+                type=flavor.type,
+                config=ArtifactLogStoreConfig(),
+                environment={},
+                user=Client().active_user.id,
+                created=now,
+                updated=now,
+                secrets=[],
+            )
```
stefannica: Given that every ArtifactLogStore instance has its own workload resources (e.g. background thread) that it initializes and uses to process logs, creating a new instance here every time the propery is called is very dangerous.

Instead of creating a new default log store every time this property is called, you should instead add a default log store to the stack during stack validation, similar to what happens with the local image builder.

src/zenml/log_stores/artifact/artifact_log_exporter.py
```diff
@@ -0,0 +1,272 @@
...
+        if not log_model.uri or not log_model.artifact_store_id:
+            logger.warning(
+                f"Skipping log write: missing uri or artifact_store_id for log {log_model.id}"
+            )
+            return
+
+        client = Client()
+        artifact_store = _load_artifact_store(
+            log_model.artifact_store_id, client.zen_store
+        )
```
stefannica: This tells me that there isn't an exclusive 1-to-1 relationship between ArtifactLogStore and ArtifactStore, but rather that one ArtifactLogStore can be used as a transparent proxy with any and all ArtifactStore instances that the emitted messages might be targetting. Doesn't that mean that the same ArtifactLogStore instance should be shared by all Stacks in the current process that need a default log store, regardless of ArtifactStore that they might use ?

src/zenml/log_stores/artifact/artifact_log_exporter.py
```diff
@@ -0,0 +1,272 @@
...
+            log_id: The log ID for tracking file counters.
+        """
+        if not log_model.uri or not log_model.artifact_store_id:
+            logger.warning(
+                f"Skipping log write: missing uri or artifact_store_id for log {log_model.id}"
+            )
+            return
+
+        client = Client()
+        artifact_store = _load_artifact_store(
```
stefannica: Calling this every time you need to export a batch of logs has a few downsides:

1. makes expensive calls to the ZenML API to get the artifact store
2. creates a new artifact store instance (and initializes it and any associated resources such as service connector client, connection pool etc) for every batch of logs being exported

All this makes it extremely inefficient and slow, not to mention that it puts a lot of unnecessary load on the ZenML API.

You should keep a cache of loaded artifact store instances. Either that, or create a 1-to-1 relationship between the log store instance and the artifact store used as the export target.

stefannica (changes requested):

src/zenml/zen_stores/schemas/logs_schemas.py
bcdurak: I wanted to build an experience that is somewhat similar to image builders.

We can potentially add it to the default stack but this comes with a different requirement as well. In that case, they will have to register a log store everytime they register a new stack (which I don't like as a part of the UX).

src/zenml/stack/stack.py
bcdurak: If I am not missing anything, we don't do that though? 

We get the log store through the active stack and the client. The client is a singleton, and within this client, the `active_stack` gets created only once, and within this stack, we only create a `log_store` when someone calls the property for the first time. Both the `_active_stack` and the corresponding `_log_store` get stored within their respective models afterwards.

src/zenml/log_stores/artifact/artifact_log_exporter.py
bcdurak: Will fix this.

stefannica:
@claude give me a stefan-review please

src/zenml/orchestrators/step_launcher.py
bcdurak: We will handle merge functionality.

src/zenml/stack/stack.py
stefannica: It is in fact a cache mechanism that is implemented in the Stack.from_model method that prevents this.

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,304 @@
...
+        artifact_store.makedirs(logs_uri)
+    else:
+        logs_uri = os.path.join(logs_base_uri, f"{log_id}{LOGS_EXTENSION}")
+        if artifact_store.exists(logs_uri):
+            logger.warning(
+                f"Logs file {logs_uri} already exists! Removing old log file..."
+            )
+            artifact_store.remove(logs_uri)
+
+    return sanitize_remote_path(logs_uri)
```
stefannica: This entire function is copied almost unchanged from the old step_logging module. In the context of the new artifact log store, it isn't accurate anymore. There should be no interacting with the artifact store from this side of the log store, because that may come into conflict with the exporter. All parent directories should be created by the exporter instead, if necessary.


```suggestion
    if artifact_store.config.IS_IMMUTABLE_FILESYSTEM:
        logs_uri = os.path.join(logs_base_uri, log_id)
    else:
        logs_uri = os.path.join(logs_base_uri, f"{log_id}{LOGS_EXTENSION}")

    return sanitize_remote_path(logs_uri)
```

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -611,7 +611,7 @@ def to_model(
             client_logs = [
                 log_entry
                 for log_entry in self.logs
-                if log_entry.source == "client"
+                if log_entry.source in ["client", "deployment"]
```
stefannica, line 614: TODO: check with @Cahllagerfeld why the front end needs this instead of using the `log_collection` field

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: I moved some of this code over to the exporter side. The code that removes logs I deleted completely, because we should never delete logs.

src/zenml/log_stores/artifact/artifact_log_exporter.py
bcdurak: This is now addressed properly.

src/zenml/log_stores/artifact/artifact_log_exporter.py
bcdurak: This is now addressed properly.

stefannica:
A new situation emerged: circular imports are now logged weirdly:

```
2025-12-01 17:29:47,416 - zenml.materializers.materializer_registry - DEBUG - Registered materializer <class 'zenml.materializers.built_in_materializer.BuiltInMaterializer'> for <class 'bool'> (materializer_registry.py:45)
2025-12-01 17:29:48,181 - zenml.logger - DEBUG - Logging set to level: DEBUG (logger.py:239)
2025-12-01 17:29:48,182 - zenml.logger - DEBUG - Logging set to level: DEBUG (logger.py:239)
Error in sys.excepthook:
2025-12-01 17:29:48,183 - zenml.logger - DEBUG - Logging set to level: DEBUG (logger.py:239)
object address  : 0x7616b27ee2c0
object refcount : 1
object type     : 0xa35360
object type name: ImportError
object repr     : ImportError("cannot import name 'LoggingContext' from partially initialized module 'zenml.utils.logging_utils' (most likely due to a circular import) (/home/silent-assassin/aspyre/src/zenml/src/zenml/utils/logging_utils.py)")
lost sys.stderr
2025-12-01 17:29:48,184 - zenml.logger - DEBUG - Logging set to level: DEBUG (logger.py:239)

Original exception was:
2025-12-01 17:29:48,184 - zenml.logger - DEBUG - Logging set to level: DEBUG (logger.py:239)
object address  : 0x7616b29d72c0
object refcount : 4
object type     : 0xa35360
object type name: ImportError
object repr     : ImportError("cannot import name 'get_step_context' from partially initialized module 'zenml' (most likely due to a circular import) (/home/silent-assassin/aspyre/src/zenml/src/zenml/__init__.py)")
lost sys.stderr
```

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -153,26 +156,22 @@ def pipeline(self) -> "DynamicPipeline":
...
+                orchestrator_run_id=self._orchestrator_run_id,
+            )
+
+            logging_context = nullcontext()
+            if is_pipeline_logging_enabled(
+                self._snapshot.pipeline_configuration
+            ):
+                logging_context = setup_run_logging(
+                    pipeline_run=run,
+                    source="orchestrator",
```
stefannica: Just a note that this adds one more API call to every pipeline run. We should revisit this later.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -257,20 +259,11 @@ def main() -> None:
...
-                break
     else:
         orchestrator_run_id = orchestrator_pod_name
+
         if args.run_id:
-            pipeline_run = client.zen_store.update_run(
-                run_id=args.run_id,
-                run_update=PipelineRunUpdate(
-                    orchestrator_run_id=orchestrator_run_id
-                ),
```
stefannica, line 272: Did we lose the `orchestrator_run_id` here ?

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+            raise ValueError(
+                "logs_model.artifact_store_id does not match the artifact store "
+                "id of the log store."
+            )
+
+        if start_time or end_time:
+            logger.warning(
+                "start_time and end_time are not supported for "
+                "ArtifactLogStore.fetch(). Both parameters will be ignored."
+            )
```
stefannica, line 331: Is this a functionality we had and now lost because of this refactoring?

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+
+        client = Client()
+        log_entries = fetch_log_records(
+            zen_store=client.zen_store,
+            artifact_store_id=logs_model.artifact_store_id,
```
stefannica: can you update this utility function to receive an artifact store instance instead of an ID, so it won't need to load it again?


```suggestion
            artifact_store=self._artifact_store,
```

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+    if old_format:
+        timestamp = old_format.group(1) + "Z"
+        line = line.replace(old_format.group(0), "").strip()
+
+    return LogEntry(
+        message=line,
+        name=None,
+        level=LoggingLevels.INFO,
+        timestamp=timestamp,
+    )
```
stefannica, line 195: you should test this on a bunch of old logs to see if it still works. You also need to document this limitation: old logs will all be shown as INFO logs instead of their true severity (is this some functionality we had and lost?)

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+        For plain text logs, only message is populated with INFO level default.
+        Returns None only for empty lines.
+    """
+    line = log_line.strip()
+    if not line:
+        return None
+
+    if line.startswith("{") and line.endswith("}"):
+        try:
+            return LogEntry.model_validate_json(line)
```
stefannica, line 177: just for future compatibility, did you configure `LogEntry` to allow/ignore extra fields, or will it blow up in the future if we change its structure to e.g. remove an attribute?

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+        artifact_store_id: The ID of the artifact store.
+        logs_uri: The URI of the log file or directory.
+
+    Yields:
+        Individual log lines as strings.
+
+    Raises:
+        DoesNotExistException: If the artifact does not exist in the artifact store.
+    """
+    artifact_store = load_artifact_store(artifact_store_id, zen_store)
```
stefannica: this should receive an artifact store instance, not create it again

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+            # Sort files to read them in order
+            files.sort()
+
+            for file in files:
+                file_path = os.path.join(logs_uri, str(file))
+                with artifact_store.open(file_path, "r") as f:
+                    for line in f:
+                        yield line.rstrip("\n\r")
+    finally:
+        artifact_store.cleanup()
```
stefannica: (see my other comment) please don't cleanup the artifact store here if you're receiving it as an external object. You should clean it up as close to the code location where you instantiate it, for good visibility. In this case, it probably makes sense to call cleanup for both the log store and the artifact store in the `fetch_logs` utility

src/zenml/log_stores/artifact/artifact_log_store.py
```diff
@@ -0,0 +1,329 @@
...
+                for line in file:
+                    yield line.rstrip("\n\r")
+        else:
+            # Directory case - may contain multiple log files
+            files = artifact_store.listdir(logs_uri)
+            if not files:
+                raise DoesNotExistException(
+                    f"Folder '{logs_uri}' is empty in artifact store "
+                    f"'{artifact_store.name}'."
+                )
```
stefannica: does this really support live-watching the logs while the pipeline is still running, in the context in which multiple log files are created and merged only at the end of the step/pipeline run ? I think I've seen some inconsistencies like older client logs not showing up until the pipeline run is complete but recent logs are available. we have to retest this.

src/zenml/log_stores/artifact/artifact_log_store_flavor.py
stefannica: what should this logo be ? if you give me one, I'll upload it for you. Also see my comment on whether we should publish this as a builtin flavor at all.

src/zenml/stack/flavor_registry.py
```diff
@@ -85,6 +89,8 @@ def builtin_flavors(self) -> List[Type[Flavor]]:
             GitHubContainerRegistryFlavor,
             LocalImageBuilderFlavor,
             DockerDeployerFlavor,
+            ArtifactLogStoreFlavor,
```
stefannica: I suspect this flavor will be shown in the UI but users won't / shouldn't actually be able to use it. Should we even register it as a built-in flavor at all ?

src/zenml/log_stores/datadog/datadog_flavor.py
stefannica: let me take care of this for you

src/zenml/log_stores/otel/otel_flavor.py
```diff
@@ -0,0 +1,125 @@
...
+        default=15000,
+        description="Export interval in milliseconds",
+    )
+    max_export_batch_size: int = Field(
+        default=512,
+        description="Maximum batch size for exports",
+    )
+
+
+class OtelLogStoreFlavor(Flavor):
```
stefannica: do we really need an OTEL log store flavor? we only really use this as a base class, which makes little sense to keep around. Better to have flavor classes implemented from scratch (i.e. inherit directly from `BaseLogStoreFlavor`)

src/zenml/log_stores/otel/otel_flavor.py
```diff
@@ -0,0 +1,125 @@
...
+        return self.docs_url
+
+    @property
+    def logo_url(self) -> str:
+        """URL to the flavor logo.
+
+        Returns:
+            The URL to the flavor logo.
+        """
+        # TODO: Add a logo for the OpenTelemetry log store
```
stefannica: we would never really need this logo, because this flavor is never actually directly used.

src/zenml/log_stores/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    "BaseLogStore",
+    "BaseLogStoreConfig",
+    "BaseLogStoreFlavor",
+    "DatadogLogStore",
+    "DatadogLogStoreConfig",
+    "DatadogLogStoreFlavor",
+    "OtelLogStore",
+    "OtelLogStoreConfig",
+    "OtelLogStoreFlavor",
+]
```
stefannica, line 53: It's generally not a good idea to have flavors and implementations forcefully mashed together in a module init file like this. You don't want to import the OTEL or datadog dependencies if all you need is to access the flavor class/configuration.

src/zenml/log_stores/datadog/datadog_flavor.py
stefannica: DONE

src/zenml/log_stores/base_log_store.py
```diff
@@ -0,0 +1,165 @@
...
+        """Register an emitter for the log store."""
+        with self._lock:
+            self._emitter_counter += 1
+
+    def deregister_emitter(self) -> None:
+        """Deregister an emitter for the log store."""
+        with self._lock:
+            self._emitter_counter -= 1
+            if self._emitter_counter == 0:
+                self.flush()
```
stefannica, line 105: Note to me to perhaps rework this if there's time

src/zenml/models/v2/core/step_run.py
```diff
@@ -314,6 +313,15 @@ class StepRunResponseMetadata(ProjectScopedResponseMetadata):
 class StepRunResponseResources(ProjectScopedResponseResources):
     """Class for all resource models associated with the step run entity."""
 
+    logs: Optional["LogsResponse"] = Field(
+        title="Logs associated with this step run.",
+        default=None,
+    )
```
stefannica, line 319: I somehow expected this to break the UI, but it didn't. Why move them from the metadata in the first place ?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1587,7 +1587,7 @@ def migrate_database(self) -> None:
         else:
             alembic_logger.setLevel(logging.WARNING)
 
-        alembic_logger.addHandler(get_console_handler())
+        alembic_logger.addHandler(get_zenml_handler())
```
stefannica: Why are we redirecting alembic logs to the zenml log store ? These logs are very rarely used in a pipeline run context.

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -485,9 +484,18 @@ def to_model(
...
+            # TODO: This will be safe to remove in future releases (>0.84.0).
+            step_logs = [
+                log_entry
+                for log_entry in self.logs
+                if log_entry.source == "step"
+            ]
             resources = StepRunResponseResources(
                 user=self.user.to_model() if self.user else None,
                 model_version=model_version,
+                logs=step_logs[0].to_model() if step_logs else None,
```
stefannica, line 497: you moved the logs from the metadata to resources, but you didn't update the `get_query_options` implementation.

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -326,31 +324,34 @@ def get_step_status(
...
-        artifact_store_id=step.logs.artifact_store_id,
-        logs_uri=step.logs.uri,
-    )
+    if step.log_collection:
+        for logs_response in step.log_collection:
+            if logs_response.source == source:
+                return fetch_logs(
+                    logs=logs_response,
+                    zen_store=store,
+                    limit=MAX_ENTRIES_PER_REQUEST,
```
stefannica, line 354: is this limitation recently introduced, or was there a similar hardcoded limit enforced before ?

src/zenml/zen_server/routers/steps_endpoints.py
```diff
@@ -326,31 +324,34 @@ def get_step_status(
...
+    verify_permission_for_model(pipeline_run, action=Action.READ)
 
-    return fetch_log_records(
-        zen_store=store,
-        artifact_store_id=step.logs.artifact_store_id,
-        logs_uri=step.logs.uri,
-    )
+    if step.log_collection:
+        for logs_response in step.log_collection:
+            if logs_response.source == source:
```
stefannica, line 350: nit: it feels wrong to have this modeled as a list. Given that there is only one log per source per step, it should be a dictionary. similar comment about the log collection on the pipeline run response model.

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: Previously, we didn't have this functionality with the Artifact store logs.

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: done

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: the code and the limitation were there before.

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: done

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: done

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: done

src/zenml/stack/flavor_registry.py
stefannica: we removed the flavor completely

src/zenml/log_stores/artifact/artifact_log_store_flavor.py
stefannica: we removed this flavor completely

src/zenml/log_stores/otel/otel_flavor.py
stefannica: removed the flavor

src/zenml/log_stores/__init__.py
stefannica: shut the f**k up stefan, you hypocrite, you did the same thing for deployers

src/zenml/models/v2/core/step_run.py
stefannica: The UI doesn't look at the logs in the metadata at all. It calls the `/api/v1/steps/uuid/logs` endpoint directly, without passing a source. @Cahllagerfeld will have to change the UI code to look at the log_collection field and show a dropdown with all the available sources.

src/zenml/zen_stores/sql_zen_store.py
stefannica: reverted

src/zenml/zen_stores/schemas/step_run_schemas.py
stefannica: done

src/zenml/zen_server/routers/steps_endpoints.py
stefannica: this limitation was there before. the size of each entry is also limited (chunking happens every time if it exceeds a certain size (in the exporter)

src/zenml/zen_server/routers/steps_endpoints.py
stefannica: we decided earlier that there might be more than one log entry associated with the same source

src/zenml/log_stores/artifact/artifact_log_exporter.py
```diff
@@ -0,0 +1,377 @@
...
+                        key=ZENML_OTEL_LOG_STORE_CONTEXT_KEY,
+                        context=log_data.log_record.context,
+                    ),
+                )
+                flush = otel_context.get_value(
+                    key=ZENML_OTEL_LOG_STORE_FLUSH_KEY,
+                    context=log_data.log_record.context,
+                )
+                if not log_model:
+                    continue
```
stefannica: move the check up
```suggestion
                if not log_model:
                    continue
                flush = otel_context.get_value(
                    key=ZENML_OTEL_LOG_STORE_FLUSH_KEY,
                    context=log_data.log_record.context,
                )
```

src/zenml/log_stores/artifact/artifact_log_store.py
stefannica: re-tested with datadog and GCS. no inconsistencies were found

PR #4119, opened 2025-11-02
Huggingface deployer

htahir1:
## Describe changes
I implemented a new Hugging Face deployer component that allows deploying containerized applications to Hugging Face Spaces. This deployer complements the existing Hugging Face model deployer by providing a more general-purpose deployment option for any containerized application.

The implementation includes:
- A new `HuggingFaceDeployer` class that extends `ContainerizedDeployer`
- Configuration options for Space hardware, storage, and visibility
- Support for authentication via token, secret, or environment variable
- Methods to provision, deprovision, and check deployment status

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/deployers/README.md
+++ b/docs/book/component-guide/deployers/README.md
@@ -33,7 +33,7 @@ Out of the box, ZenML comes with a `local` deployer already part of the default 
 | [Docker](docker.md)                | `docker`   | Built-in      | Deploys pipelines as locally running Docker containers                                |
 | [GCP Cloud Run](gcp-cloud-run.md)            | `gcp`     | `gcp`         | Deploys pipelines to Google Cloud Run for serverless execution             |
 | [AWS App Runner](aws-app-runner.md)           | `aws`     | `aws`         | Deploys pipelines to AWS App Runner for serverless execution                       |
-| [Hugging Face](huggingface.md)                | `huggingface` | `huggingface` | Deploys pipelines to Hugging Face Spaces as Docker applications            |
+| [Hugging Face](huggingface.md)                | `huggingface` | `huggingface` | Deploys pipelines to Hugging Face Spaces as Docker Spaces            |
 
 If you would like to see the available flavors of deployers, you can use the command:
 
--- a/docs/book/component-guide/deployers/README.md
+++ b/docs/book/component-guide/deployers/README.md
@@ -33,6 +33,7 @@ Out of the box, ZenML comes with a `local` deployer already part of the default 
 | [Docker](docker.md)                | `docker`   | Built-in      | Deploys pipelines as locally running Docker containers                                |
 | [GCP Cloud Run](gcp-cloud-run.md)            | `gcp`     | `gcp`         | Deploys pipelines to Google Cloud Run for serverless execution             |
 | [AWS App Runner](aws-app-runner.md)           | `aws`     | `aws`         | Deploys pipelines to AWS App Runner for serverless execution                       |
+| [Hugging Face](huggingface.md)                | `huggingface` | `huggingface` | Deploys pipelines to Hugging Face Spaces as Docker applications            |
 
 If you would like to see the available flavors of deployers, you can use the command:
 
--- a/docs/book/component-guide/deployers/huggingface.md
+++ b/docs/book/component-guide/deployers/huggingface.md
@@ -103,7 +103,7 @@ For additional configuration of the Hugging Face deployer, you can pass the foll
 
   * `space_hardware` (default: `None`): Hardware tier for the Space (e.g., `'cpu-basic'`, `'cpu-upgrade'`, `'t4-small'`, `'t4-medium'`, `'a10g-small'`, `'a10g-large'`). If not specified, uses free CPU tier. See [Hugging Face Spaces GPU documentation](https://huggingface.co/docs/hub/spaces-gpus) for available options and pricing.
   * `space_storage` (default: `None`): Persistent storage tier for the Space (e.g., `'small'`, `'medium'`, `'large'`). If not specified, no persistent storage is allocated.
-  * `private` (default: `False`): Whether to create the Space as private. Public Spaces are visible to everyone.
+  * `private` (default: `True`): Whether to create the Space as private. Set to `False` to make the Space publicly visible to everyone.
   * `app_port` (default: `8000`): Port number where your deployment server listens. Defaults to 8000 (ZenML server default). Hugging Face Spaces will route traffic to this port.
 
 Check out [this docs page](https://docs.zenml.io/concepts/steps_and_pipelines/configuration) for more information on how to specify settings.
@@ -116,7 +116,7 @@ from zenml.integrations.huggingface.deployers import HuggingFaceDeployerSettings
 huggingface_settings = HuggingFaceDeployerSettings(
     space_hardware="t4-small",
     space_storage="small",
-    private=True,
+    # private=True is the default for security
 )
 
 @pipeline(
@@ -167,12 +167,12 @@ The Hugging Face deployer handles secrets and environment variables **securely**
 - **Nothing is baked into the Dockerfile** - no risk of leaked credentials even in public Spaces
 
 **What this means:**
-- ✅ Safe to use with public Spaces (the default)
-- ✅ Secrets remain encrypted and hidden from public view
+- ✅ Safe to use with both private and public Spaces
+- ✅ Secrets remain encrypted and hidden from view
 - ✅ Environment variables are managed through HF's secure API
 - ✅ No credentials exposed in Dockerfile or repository files
 
-This is especially important since Hugging Face Spaces are **public by default** (`private: bool = False`). Without this secure approach, any secrets would be visible to anyone viewing your Space's repository.
+This secure approach ensures that if you choose to make your Space public (`private=False`), credentials remain protected and are never visible to anyone viewing your Space's repository.
 
 ### Container Registry Requirement
 
--- a/docs/book/component-guide/deployers/huggingface.md
+++ b/docs/book/component-guide/deployers/huggingface.md
@@ -15,7 +15,7 @@ This component is only meant to be used within the context of a [remote ZenML in
 You should use the Hugging Face deployer if:
 
 * you're already using Hugging Face for model hosting or datasets.
-* you want to share your ML pipelines as publicly accessible or private Spaces.
+* you want to share your AI pipelines as publicly accessible or private Spaces.
 * you're looking for a simple, managed platform for deploying Docker-based applications.
 * you want to leverage Hugging Face's infrastructure for hosting your pipeline deployments.
 * you need an easy way to showcase ML workflows to the community.
--- a/docs/book/component-guide/deployers/huggingface.md
+++ b/docs/book/component-guide/deployers/huggingface.md
@@ -104,7 +104,7 @@ For additional configuration of the Hugging Face deployer, you can pass the foll
   * `space_hardware` (default: `None`): Hardware tier for the Space (e.g., `'cpu-basic'`, `'cpu-upgrade'`, `'t4-small'`, `'t4-medium'`, `'a10g-small'`, `'a10g-large'`). If not specified, uses free CPU tier.
   * `space_storage` (default: `None`): Persistent storage tier for the Space (e.g., `'small'`, `'medium'`, `'large'`). If not specified, no persistent storage is allocated.
   * `private` (default: `False`): Whether to create the Space as private. Public Spaces are visible to everyone.
-  * `app_port` (default: `7860`): Port number where your deployment server listens. Hugging Face Spaces will route traffic to this port.
+  * `app_port` (default: `8000`): Port number where your deployment server listens. Defaults to 8000 (ZenML server default). Hugging Face Spaces will route traffic to this port.
 
 Check out [this docs page](https://docs.zenml.io/concepts/steps_and_pipelines/configuration) for more information on how to specify settings.
 
@@ -153,66 +153,75 @@ https://huggingface.co/spaces/<YOUR_USERNAME>/<SPACE_PREFIX>-<DEPLOYMENT_NAME>
 
 By default, the space prefix is `zenml` but this can be configured using the `space_prefix` parameter when registering the deployer.
 
-## Deployment Modes
+## Important Requirements
 
-The Hugging Face deployer supports two deployment modes depending on your stack configuration:
+### Container Registry Requirement
 
-### Mode 1: Image Reference (Stack with Container Registry)
+{% hint style="warning" %}
+The Hugging Face deployer **requires** a container registry to be part of your ZenML stack. The Docker image must be pre-built and pushed to a **publicly accessible** container registry.
+{% endhint %}
 
-When your ZenML stack includes a container registry, the deployer references a pre-built Docker image:
+**Why public access is required:**
+Hugging Face Spaces cannot authenticate with private Docker registries when building Docker Spaces. The platform pulls your Docker image during the build process, which means it needs public access.
 
-**How it works:**
-- ZenML builds and pushes an image to your container registry
-- The Dockerfile in Hugging Face Spaces references this image with `FROM <your-image>`
-- The deployment server starts automatically with the correct entrypoint
+**Recommended registries:**
+- [Docker Hub](https://hub.docker.com/) public repositories
+- [GitHub Container Registry (GHCR)](https://ghcr.io) with public images
+- Any other public container registry
 
-**Requirements:**
-- Container registry must be part of your stack
-- ⚠️ **The Docker image must be publicly accessible** - Hugging Face Spaces cannot authenticate with private registries
-
-**Use when:**
-- You have a public container registry or public repository
-- You want to pre-build images in your CI/CD pipeline
-- You're using services like Docker Hub with public repositories
-
-### Mode 2: Full Build (Stack without Container Registry)
-
-When your stack does NOT have a container registry, the deployer builds the complete image from scratch in Hugging Face Spaces:
-
-**How it works:**
-- Deployer uploads your source code to the Space
-- Generates a complete Dockerfile that installs all dependencies
-- Hugging Face Spaces builds the entire image from your code
-- ✅ **No private registry authentication needed!**
-
-**Requirements:**
-- No container registry in your stack
-- Source code and dependencies uploadable to Hugging Face
-
-**Use when:**
-- You don't want to manage a container registry
-- You want to avoid private registry authentication issues
-- Your code and dependencies aren't too large for uploading
-- You're okay with longer build times in Hugging Face Spaces
-
-### Private Registry Workaround
-
-If you're currently using a private registry and want to deploy to Hugging Face, you have two options:
-
-1. **Remove the container registry from your stack** to automatically use Full Build Mode:
+**Example setup with GitHub Container Registry:**
 ```shell
-# Create a new stack without container registry
-zenml stack copy my-stack hf-stack
-zenml stack update hf-stack --container-registry=null
-zenml stack set hf-stack
-```
-
-2. **Use a public container registry** for Hugging Face deployments:
-```shell
-zenml container-registry register public_registry \
+# Register a public container registry
+zenml container-registry register ghcr_public \
     --flavor=default \
-    --uri=docker.io/<your-public-username>
+    --uri=ghcr.io/<your-github-username>
+
+# Add it to your stack
+zenml stack update <STACK_NAME> --container-registry=ghcr_public
 ```
+
+### Configuring iframe Embedding (X-Frame-Options)
+
+By default, ZenML's deployment server sends an `X-Frame-Options` header that prevents the deployment UI from being embedded in iframes. This causes issues with Hugging Face Spaces, which displays deployments in an iframe.
+
+**To fix this**, you must configure your pipeline's `DeploymentSettings` to disable the `X-Frame-Options` header:
+
+```python
+from zenml import pipeline
+from zenml.config import DeploymentSettings, SecureHeadersConfig
+
+# Configure deployment settings
+deployment_settings = DeploymentSettings(
+    app_title="My ZenML Pipeline",
+    app_description="ML pipeline deployed to Hugging Face Spaces",
+    app_version="1.0.0",
+    secure_headers=SecureHeadersConfig(
+        xfo=False,  # Disable X-Frame-Options to allow iframe embedding
+        server=True,
+        hsts=False,
+        content=True,
+        referrer=True,
+        cache=True,
+        permissions=True,
+    ),
+    cors={
+        "allow_origins": ["*"],
+        "allow_methods": ["GET", "POST", "OPTIONS"],
+        "allow_headers": ["*"],
+        "allow_credentials": False,
+    },
+)
+
+@pipeline(
+    name="my_hf_pipeline",
+    settings={"deployment": deployment_settings}
+)
+def my_pipeline():
+    # Your pipeline steps here
+    pass
+```
+
+Without this configuration, the Hugging Face Spaces UI will show a blank page or errors when trying to display your deployment.
 
 ## Additional Resources
 
--- a/docs/book/component-guide/deployers/huggingface.md
+++ b/docs/book/component-guide/deployers/huggingface.md
@@ -1,0 +1,195 @@
+---
+description: Deploying your pipelines to Hugging Face Spaces.
+---
+
+# Hugging Face Deployer
+
+[Hugging Face Spaces](https://huggingface.co/spaces) is a platform for hosting and sharing machine learning applications. The Hugging Face deployer is a [deployer](./) flavor included in the ZenML Hugging Face integration that deploys your pipelines to Hugging Face Spaces as Docker-based applications.
+
+{% hint style="warning" %}
+This component is only meant to be used within the context of a [remote ZenML installation](https://docs.zenml.io/getting-started/deploying-zenml). Usage with a local ZenML setup may lead to unexpected behavior!
+{% endhint %}
+
+## When to use it
+
+You should use the Hugging Face deployer if:
+
+* you're already using Hugging Face for model hosting or datasets.
+* you want to share your ML pipelines as publicly accessible or private Spaces.
+* you're looking for a simple, managed platform for deploying Docker-based applications.
+* you want to leverage Hugging Face's infrastructure for hosting your pipeline deployments.
+* you need an easy way to showcase ML workflows to the community.
+
+## How to deploy it
+
+{% hint style="info" %}
+The Hugging Face deployer requires a remote ZenML installation. You must ensure that you are connected to the remote ZenML server before using this stack component.
+{% endhint %}
+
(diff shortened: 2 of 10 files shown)


src/zenml/integrations/huggingface/__init__.py
```diff
@@ -20,6 +20,7 @@
 from zenml.stack import Flavor
 
 HUGGINGFACE_MODEL_DEPLOYER_FLAVOR = "huggingface"
+HUGGINGFACE_DEPLOYER_FLAVOR = "huggingface"
```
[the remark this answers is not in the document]
htahir1: No these dont clash

docs/book/component-guide/deployers/huggingface.md
```diff
@@ -0,0 +1,222 @@
...
+
+* Basic settings common to all Deployers:
+
+  * `auth_key`: A user-defined authentication key to use to authenticate with deployment API calls.
+  * `generate_auth_key`: Whether to generate and use a random authentication key instead of the user-defined one.
+  * `lcm_timeout`: The maximum time in seconds to wait for the deployment lifecycle management to complete.
+
+* Hugging Face Spaces-specific settings:
+
+  * `space_hardware` (default: `None`): Hardware tier for the Space (e.g., `'cpu-basic'`, `'cpu-upgrade'`, `'t4-small'`, `'t4-medium'`, `'a10g-small'`, `'a10g-large'`). If not specified, uses free CPU tier.
```
strickvl: Maybe reference this page? https://huggingface.co/docs/hub/spaces-gpus

docs/book/component-guide/deployers/huggingface.md
```diff
@@ -0,0 +1,222 @@
...
+  * `space_storage` (default: `None`): Persistent storage tier for the Space (e.g., `'small'`, `'medium'`, `'large'`). If not specified, no persistent storage is allocated.
+  * `private` (default: `False`): Whether to create the Space as private. Public Spaces are visible to everyone.
```
strickvl: Shouldn't we default to private?

docs/book/component-guide/deployers/huggingface.md
```diff
@@ -0,0 +1,222 @@
...
+- Generates a complete Dockerfile that installs all dependencies
+- Hugging Face Spaces builds the entire image from your code
+- ✅ **No private registry authentication needed!**
+
+**Requirements:**
+- No container registry in your stack
+- Source code and dependencies uploadable to Hugging Face
+
+**Use when:**
+- You don't want to manage a container registry
```
strickvl: earlier in the docs you said that a remote container registry was required

docs/book/component-guide/deployers/README.md
```diff
@@ -33,6 +33,7 @@ Out of the box, ZenML comes with a `local` deployer already part of the default
 | [Docker](docker.md)                | `docker`   | Built-in      | Deploys pipelines as locally running Docker containers                                |
 | [GCP Cloud Run](gcp-cloud-run.md)            | `gcp`     | `gcp`         | Deploys pipelines to Google Cloud Run for serverless execution             |
 | [AWS App Runner](aws-app-runner.md)           | `aws`     | `aws`         | Deploys pipelines to AWS App Runner for serverless execution                       |
+| [Hugging Face](huggingface.md)                | `huggingface` | `huggingface` | Deploys pipelines to Hugging Face Spaces as Docker applications            |
```
strickvl:
```suggestion
| [Hugging Face](huggingface.md)                | `huggingface` | `huggingface` | Deploys pipelines to Hugging Face Spaces as Docker Spaces            |
```

src/zenml/integrations/huggingface/flavors/huggingface_deployer_flavor.py
```diff
@@ -0,0 +1,118 @@
...
+        return self.generate_default_sdk_docs_url()
+
+    @property
+    def logo_url(self) -> str:
+        """A URL to represent the flavor in the dashboard.
+
+        Returns:
+            The flavor logo.
+        """
+        return "https://public-flavor-logos.s3.eu-central-1.amazonaws.com/deployer/huggingface.png"
```
strickvl, line 132: I guess this needs adding / uploading by someone?

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+        Returns:
+            The token from config, secret, or environment.
+        """
+        if self.config.token:
+            return self.config.token
+
+        if self.config.secret_name:
+            client = Client()
+            secret = client.get_secret(self.config.secret_name)
+            return secret.secret_values.get("token")
```
strickvl: given that you've hard-coded this, you should state in the docs that the secret key must be 'token'

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+        """
+        lines = [f"FROM {image}"]
+
+        # Add environment variables
+        env_vars = {**environment, **secrets}
+        for k, v in env_vars.items():
+            # Escape backslashes and quotes
+            escaped_value = v.replace(chr(92), chr(92) * 2).replace(
+                chr(34), chr(92) + chr(34)
+            )
```
strickvl:
```suggestion
            escaped_value = v.replace("\\", "\\\\").replace('"', '\\"')
```

Maybe as alternative. The current version is maybe a bit unclear.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+            raise DeploymentNotFoundError("Space ID not found in metadata")
+
+        api = self._get_hf_api()
+
+        try:
+            api.delete_repo(repo_id=space_id, repo_type="space")
+            logger.info(f"Deleted Space: {space_id}")
+            return None
+        except Exception as e:
+            if "404" in str(e):
```
strickvl: Can't we get the 404 error type directly from the object? just grabbing it from the error string feels not ideal.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+                    from huggingface_hub import SpaceHardware
+
+                    api.request_space_hardware(
+                        repo_id=space_id,
+                        hardware=getattr(
+                            SpaceHardware, hardware.upper().replace("-", "_")
+                        ),
+                    )
+                except Exception as e:
+                    logger.warning(f"Failed to set hardware {hardware}: {e}")
```
strickvl: Should we actually fail here instead of just logging a warning?

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+
+        return "\n".join(lines)
+
+    def do_provision_deployment(
+        self,
+        deployment: DeploymentResponse,
+        stack: "Stack",
+        environment: Dict[str, str],
+        secrets: Dict[str, str],
+        timeout: int,
```
strickvl, line 258: We never actually use this... let's use it or get rid of it.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+        Returns:
+            The Space ID in format 'username/space-name'.
+        """
+        api = self._get_hf_api()
+        username = api.whoami()["name"]
+
+        # Simple sanitization: alphanumeric, hyphens, underscores only
+        sanitized = re.sub(r"[^a-zA-Z0-9\-_]", "-", deployment.name).lower()
+        sanitized = sanitized.strip("-") or "deployment"
+        space_name = f"{self.config.space_prefix}-{sanitized}"
```
strickvl: do we need to validate the space name as well? HF must have a max # of chars allowed for this.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
strickvl: Something like catch `HfHubHTTPError` and check `e.response.status_code == 404`

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+        """Get the Space ID for a deployment.
+
+        Args:
+            deployment: The deployment.
+
+        Returns:
+            The Space ID in format 'username/space-name'.
+        """
+        api = self._get_hf_api()
+        username = api.whoami()["name"]
```
strickvl: does this work for an organization? def works for a normal user, but does it work if you want to deploy to the ZenML org on HF Spaces? Not sure it would.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,480 @@
...
+        logger.info(
+            f"Deploying image {image} to Hugging Face Space. "
+            "Ensure the image is publicly accessible."
+        )
+
+        try:
+            # Create Space if it doesn't exist
+            try:
+                api.space_info(space_id)
+                logger.info(f"Updating existing Space: {space_id}")
```
strickvl: I *think* that if someone changes the visibility of the space (from private to public or vice-versa) then it wouldn't actually make those changes with the way the code is currently written, so we should make sure to handle that.

strickvl (changes requested):
One thing that's probably not ideal is that (esp in the context of setting these as public by default) we write the secrets to the Dockerfile which then means your HF tokens (and whatever else) is basically leaked / in public the moment you use our deployer. I think `HfApi.add_space_secret(repo_id, key, value)` and `HfApi.add_space_variable(repo_id, key, value)` are the preferred ways of handling secrets / env vars with spaces, which we should use instead of injecting them into the Dockerfile.

Otherwise a bunch of smaller comments below.

docs/book/component-guide/deployers/huggingface.md
htahir1: @strickvl i dont think so actually... i think most of the time its a demo or something?

src/zenml/integrations/huggingface/flavors/huggingface_deployer_flavor.py
htahir1: done

stefannica: I did

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: ive actually gotten rid of this entire logic

docs/book/component-guide/deployers/huggingface.md
strickvl: Perhaps. Personally I'd always rather something defaulted to private and then allowed me to turn it on / set it to public whenever I want. That way at least you avoid having leaked something you didn't want to.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: done

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: We need to keep it i think for the base deployer

strickvl: Maybe just log a warning message that it isn't used, in case someone sets it?

htahir1: im following the lead here, perhaps i can just add add it in the docstring https://github.com/zenml-io/zenml/blob/claude/huggingface-deployer-011CUj51UperMPAytgCheWG1/src/zenml/deployers/local/local_deployer.py#L205

Adding a warning would pollute the logs

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+        timeout: int,
+    ) -> DeploymentOperationalState:
+        """Provision a Huggingface Space deployment.
+
+        Args:
+            deployment: The deployment to run.
+            stack: The active stack.
+            environment: Environment variables for the app.
+            secrets: Secret environment variables for the app.
+            timeout: Maximum time to wait for deployment.
```
htahir1:
```suggestion
            timeout: Maximum time to wait for deployment (unused).
```

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,552 @@
...
+        """
+        # Get owner (organization or username)
+        if self.config.organization:
+            owner = self.config.organization
+        else:
+            api = self._get_hf_api()
+            owner = api.whoami()["name"]
+
+        # Sanitize deployment name: alphanumeric, hyphens, underscores only
+        sanitized = re.sub(r"[^a-zA-Z0-9\-_]", "-", deployment.name).lower()
```
stefannica, line 177: I recommend you use at least some characters from the deployment UUID as a suffix, to make these more unique. See the existing deployer implementations for a similar pattern.

Reason: the names are unique at the level of the project

docs/book/component-guide/deployers/huggingface.md
```diff
@@ -48,10 +48,10 @@ To use the Hugging Face deployer, you need:
 
 You need a Hugging Face access token with write permissions to deploy pipelines. You can create one at [https://huggingface.co/settings/tokens](https://huggingface.co/settings/tokens).
 
-You have two different options to provide credentials to the Hugging Face deployer:
+You have two options to provide credentials to the Hugging Face deployer:
 
 * Pass the token directly when registering the deployer using the `--token` parameter
-* (recommended) Store the token in a ZenML secret and reference it using the `--secret_name` parameter
+* (recommended) Store the token in a ZenML secret and reference it using secret syntax
```
strickvl: Maybe link to the part of the docs where we describe what this is + how to use?

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -295,10 +300,22 @@ def do_provision_deployment(
...
+                space_info = api.space_info(space_id)
                 logger.info(f"Updating existing Space: {space_id}")
+
+                # Update visibility if changed
+                if space_info.private != settings.private:
+                    logger.info(
+                        f"Updating Space visibility: "
+                        f"{'private' if settings.private else 'public'}"
+                    )
+                    api.update_repo_visibility(
```
strickvl: From the HF docs: "update_repo_visibility method has been removed. Please use update_repo_settings instead." (https://huggingface.co/docs/huggingface_hub/v1.0.0.rc7/en/concepts/migration#other-deprecated-features)

strickvl (changes requested):

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+                        repo_id=space_id,
+                        private=settings.private,
+                        repo_type="space",
+                    )
+            except Exception:
```
stefannica: This exception is too broad. Isn't there a more concrete exception that the HF API throws to tell you that the space you are fetching with `api.space_info(space_id)` is not found ?

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+                try:
+                    api.add_space_variable(
+                        repo_id=space_id,
+                        key=key,
+                        value=value,
+                    )
+                except Exception as e:
+                    logger.warning(
+                        f"Failed to set environment variable {key}: {e}"
+                    )
```
stefannica: This should be an error. If you can't set the env vars, then the deployment will fail anyway. Better to fail early.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+            logger.info(f"Setting {len(secrets)} secrets...")
+            for key, value in secrets.items():
+                try:
+                    api.add_space_secret(
+                        repo_id=space_id,
+                        key=key,
+                        value=value,
+                    )
+                except Exception as e:
+                    logger.warning(f"Failed to set secret {key}: {e}")
```
stefannica: Similarly here: this should be an error. If you can't set the secrets (required to connect to the ZenML server), then the deployment will fail anyway. Better to fail early.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+        if not space_id:
+            raise DeploymentNotFoundError("Space ID not found in metadata")
+
+        api = self._get_hf_api()
+
+        try:
+            runtime = api.get_space_runtime(repo_id=space_id)
+
+            # Map Space stage to deployment status
+            if runtime.stage in ["RUNNING", "RUNNING_BUILDING"]:
```
stefannica: What is this `RUNNING_BUILDING` stage ? Feels more like a PENDING deployment state than a RUNNING one. RUNNING should only be returned when the deployment is fully provisioned and ready to be health-checked (i.e. when the health endpoint is available).

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+            elif runtime.stage in ["BUILDING", "NO_APP_FILE"]:
+                status = DeploymentStatus.PENDING
+            else:
+                # BUILD_ERROR, RUNTIME_ERROR, STOPPED, etc.
+                status = DeploymentStatus.ERROR
+
+            return DeploymentOperationalState(
+                status=status,
+                url=f"https://huggingface.co/spaces/{space_id}",
+                metadata={"space_id": space_id},
```
stefannica: You should also store the external state in the metadata, to help with debugging.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+            from huggingface_hub.utils import HfHubHTTPError
+
+            api.delete_repo(repo_id=space_id, repo_type="space")
+            logger.info(f"Deleted Space: {space_id}")
+            return None
+        except HfHubHTTPError as e:
+            if e.response.status_code == 404:
+                logger.info(f"Space {space_id} already deleted")
+                return None
+            raise DeploymentNotFoundError(
```
stefannica:
```suggestion
            raise DeploymentDeprovisionError(
```

DeploymentNotFoundError is the same as saying "the resources have been deleted in the back-end", which is not the case here. Raising DeploymentDeprovisionError signals to the core functionality that something is wrong in the back-end and that the user should either check the back-end for errors or force-delete the deployment and risk leaving orphaned resources behind.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+                except Exception as e:
+                    logger.warning(f"Failed to set secret {key}: {e}")
+
+            # Set hardware if specified (fail if this doesn't work)
+            hardware = settings.space_hardware or self.config.space_hardware
+            if hardware:
+                from huggingface_hub import SpaceHardware
+
+                try:
+                    api.request_space_hardware(
```
stefannica, line 426: Have you tried changing the hardware requirements for an existing deployment and then updating ? Doesn't calling this multiple times for the same HF space keep adding more hardware on the same space ? I.e. is this a "replace hardware resources" call or an "add hardware resources" call ?

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+                        f"Failed to set hardware {hardware}: {e}"
+                    ) from e
+
+            # Set storage if specified (fail if this doesn't work)
+            storage = settings.space_storage or self.config.space_storage
+            if storage:
+                from huggingface_hub import SpaceStorage
+
+                try:
+                    api.request_space_storage(
```
stefannica, line 450: Same here: is this a "replace storage resources" call or an "add storage resources" call ?

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+                    logger.warning(
+                        f"Failed to set environment variable {key}: {e}"
+                    )
+
+            # Set secrets using Space secrets API
+            # This is secure - secrets are encrypted and not exposed
+            logger.info(f"Setting {len(secrets)} secrets...")
+            for key, value in secrets.items():
+                try:
+                    api.add_space_secret(
```
stefannica, line 409: Will calling this a second, third etc. time overwrite the previous values ? Maybe it's the function name that just sends the wrong message - adding something to the space feels like an additive operation, not a replacement one.

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,545 @@
...
+        """
+        assert deployment.snapshot, "Pipeline snapshot not found"
+
+        settings = cast(
+            HuggingFaceDeployerSettings,
+            self.get_settings(deployment.snapshot),
+        )
+
+        api = self._get_hf_api()
+        space_id = self._get_space_id(deployment)
```
stefannica, line 283: There's a corner-case here that you haven't addressed: if there's an existing space ID stored in `deployment.deployment_metadata["space_id"]` that doesn't match the new space ID, for example because the deployment was renamed or the prefix settings changed. When this happens, you could simply call `do_deprovision_deployment` to clean up the previous space.

stefannica (changes requested):

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: It's a "replace hardware resources" call, not additive.

Evidence:

# From API signature I checked:
(self, repo_id: 'str', hardware: 'SpaceHardware', *, token: ..., sleep_time: ...) -> 'SpaceRuntime'
The API doc says: "Request new hardware for a Space" (emphasis on "new"). Each call replaces the current hardware tier. You can't have multiple hardware tiers on the same Space.

What I added to code:

# Note: request_space_hardware replaces the current hardware tier

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: Answer: It's a "replace storage resources" call, not additive.

Evidence:

# From API signature:
(self, repo_id: 'str', storage: 'SpaceStorage', *, token: ...) -> 'SpaceRuntime'
The API doc says: "Request persistent storage for a Space". Each call sets the storage tier. You can't stack multiple storage tiers.

What I added to code:

# Note: request_space_storage replaces the current storage tier

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: Answer: Yes, calling it multiple times overwrites previous values. Despite the name "add", it's actually an upsert operation.

Evidence:

# From API signature:
(self, repo_id: 'str', key: 'str', value: 'str', *, description: ..., token: ...) -> 'None'
The API doc explicitly says: "Adds or updates a secret in a Space" (emphasis added). The function name is misleading - it should probably be upsert_space_secret, but HuggingFace chose add_space_secret.

What I added to code:

# Note: add_space_secret is an upsert operation (adds or updates)
# Note: add_space_variable is an upsert operation (adds or updates)

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: This is now handled in the code at lines 301-313. When old_space_id != space_id (e.g., deployment renamed or prefix changed), it calls do_deprovision_deployment() to clean up the old Space before creating the new one. This prevents orphaned Spaces from accumulating.

strickvl (approved):

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,612 @@
...
+
+        # Add user
+        lines.append("USER 1000")
+
+        # Add entrypoint and command
+        entrypoint_line, cmd_line = self._get_entrypoint_and_command(
+            deployment
+        )
+        lines.append(entrypoint_line)
+        lines.append(cmd_line)
```
stefannica, line 248: When you're building a container image FROM a base image, you don't need to re-supply the CMD and ENTRYPOINT, they are inherited from the base class.

So you can get rid of this part, and the `_get_entrypoint_and_command` method

stefannica (changes requested):

docs/book/component-guide/deployers/huggingface.md
```diff
@@ -0,0 +1,250 @@
...
+{% hint style="warning" %}
+This component is only meant to be used within the context of a [remote ZenML installation](https://docs.zenml.io/getting-started/deploying-zenml). Usage with a local ZenML setup may lead to unexpected behavior!
+{% endhint %}
+
+## When to use it
+
+You should use the Hugging Face deployer if:
+
+* you're already using Hugging Face for model hosting or datasets.
+* you want to share your ML pipelines as publicly accessible or private Spaces.
```
htahir1:
```suggestion
* you want to share your AI pipelines as publicly accessible or private Spaces.
```

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
```diff
@@ -0,0 +1,620 @@
...
+        space_hardware: Hardware tier for the Space (e.g., 'cpu-basic', 't4-small')
+        space_storage: Persistent storage tier (e.g., 'small', 'medium', 'large')
+        private: Whether to create a private Space (default: True for security)
+        app_port: Port the container exposes (default 8000 for ZenML server)
+    """
+
+    space_hardware: Optional[str] = None
+    space_storage: Optional[str] = None
+    private: bool = True
+    app_port: int = 8000
```
stefannica: The settings class should be in the flavor module and the `HuggingFaceDeployerConfig` class should inherit from it.

The `app_port` setting should also be removed, because this is something that is customized through the DeploymentSettings instead:

```
        port = (
            snapshot.pipeline_configuration.deployment_settings.uvicorn_port
        )
```

stefannica (changes requested):

src/zenml/integrations/huggingface/deployers/huggingface_deployer.py
htahir1: @stefannica fixed

stefannica (approved):

PR #4127, opened 2025-11-04
Add Kubernetes Deployer for Pipeline Deployments

safoinme:
## Describe changes
This PR introduces a new Kubernetes Deployer that enables deploying ZenML pipelines as long-running services on any Kubernetes cluster.

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -1,0 +1,48 @@
+---
+name: Close stale PRs
+on:
+  schedule:
+    # Run daily at 1:30 AM UTC
+    - cron: 30 1 * * *
+  workflow_dispatch:
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v10
+        with:
+          # Only process PRs, not issues
+          days-before-stale: -1
+          days-before-issue-stale: -1
+          days-before-close: -1
+          days-before-issue-close: -1
+
+          # PR-specific configuration
+          days-before-pr-stale: 14  # 2 weeks - first reminder
+          days-before-pr-close: 14  # Closes 14 days after stale (28 days total = 4 weeks)
+          # Messages
+          stale-pr-message: |
+            ⚠️ This PR has been inactive for 2 weeks and has been marked as stale.
+            **Timeline:**
+            - Week 2 (now): First reminder - PR marked as stale
+            - Week 4: PR will be automatically closed if no activity
+            Please update this PR or leave a comment to keep it active. Any activity will reset the timer and remove the stale label.
+          close-pr-message: |
+            🔒 This PR has been automatically closed due to 4 weeks of inactivity.
+            If you'd like to continue working on this, please feel free to reopen it. We're happy to help you get it merged!
+
+          # Labels
+          stale-pr-label: stale
+
+          # Reset behavior - remove stale label when PR is updated or commented
+          remove-stale-when-updated: true
+          remove-pr-stale-when-updated: true
+
+          # Remove stale label when activity occurs
+          labels-to-remove-when-unstale: stale
+
+          # Operations limit
+          operations-per-run: 5000
--- a/src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
+++ b/src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
@@ -39,6 +39,18 @@ logger = get_logger(__name__)
 WANDB_API_KEY = "WANDB_API_KEY"
 
 
+def sanitize_tag(tag: str) -> str:
+    """Sanitize a tag to be a valid Wandb tag.
+
+    Args:
+        tag: The tag to sanitize.
+
+    Returns:
+        The sanitized tag.
+    """
+    return tag[:64]
+
+
 class WandbExperimentTracker(BaseExperimentTracker):
     """Track experiment using Wandb."""
 
@@ -70,7 +82,11 @@ class WandbExperimentTracker(BaseExperimentTracker):
         settings = cast(
             WandbExperimentTrackerSettings, self.get_settings(info)
         )
-        tags = settings.tags + [info.run_name, info.pipeline.name]
+        tags = settings.tags + [
+            sanitize_tag(info.run_name),
+            sanitize_tag(info.pipeline.name),
+        ]
+
         wandb_run_name = (
             settings.run_name or f"{info.run_name}_{info.pipeline_step_name}"
         )


src/zenml/cli/pipeline.py
```diff
@@ -387,6 +387,23 @@ def run_pipeline(
...
+    "Manifests will be saved to .zenml-deployments/ for inspection.",
+)
+@click.option(
+    "--print-manifests",
+    "print_manifests",
+    is_flag=True,
+    default=False,
+    required=False,
+    help="Print generated Kubernetes manifests to console (deployment, service, and any additional resources).",
+)
```
stefannica: This is an argument that is specific to the Kubernetes deployer. This is a generic CLI command that is used with any deployer flavor. This does not belong here in this form.

src/zenml/cli/pipeline.py
```diff
@@ -448,9 +471,122 @@ def deploy_pipeline(
...
+            if hasattr(deployer_settings, "model_dump"):
+                deployer_settings = deployer_settings.model_dump()
+            elif not isinstance(deployer_settings, dict):
+                deployer_settings = {}
+
+            deployer_settings["print_manifests"] = True
+            existing_settings["deployer"] = deployer_settings
+            pipeline_instance = pipeline_instance.with_options(
+                settings=existing_settings
+            )
```
stefannica: You need to think of another way to model these deployer specific args. This code is just too hacky and ignores all design considerations and separation of concerns.

src/zenml/cli/pipeline.py
```diff
@@ -448,9 +471,122 @@ def deploy_pipeline(
...
+            stack.prepare_pipeline_submission(snapshot=snapshot)
+
+            if not stack.deployer:
+                cli_utils.error(
+                    "The stack does not have a deployer. Please add a "
+                    "deployer to the stack in order to perform dry-run."
+                )
+
+            # Check if the deployer supports dry-run mode
+            if not hasattr(stack.deployer, "do_dry_run_deployment"):
```
stefannica: I'll leave a general comment here about all calls made to `hasattr` and `getattr`: you should very very rarely need these two functions in your Python code, especially if you're following typed conventions, like we do in ZenML (i.e. if you use types such as classes and pydantic objects).

`hasattr` and `getattr` is what LLMs use when they don't understand your code. Please review your own code and identify these occurrences before opening it up for review.

src/zenml/cli/pipeline.py
```diff
@@ -448,9 +471,122 @@ def deploy_pipeline(
...
+                metadata=DeploymentResponseMetadata(
+                    deployment_metadata={},
+                    auth_key=None,
+                ),
+                resources=DeploymentResponseResources(
+                    snapshot=snapshot,
+                    deployer=deployer_response,
+                    tags=[],
+                ),
+            )
```
stefannica: this is another thing that just feels wrong: if you have to "fake" things like deployment responses in order to get the rest of your code to behave correctly, that basically looks like you're pushing a square peg in a round hole: it's not good design.

this tells me that you need to make the base deployer class and everything else around it aware of this new "dry-run" mode

safoinme:
Hey @stefannica i have taking out all dry-run related code out of this PR, it should be now just k8s deployer related.

All the above comments are related to the code that was taking out and are handled part of the dry-run PR

docs/book/component-guide/deployers/kubernetes.md
```diff
@@ -0,0 +1,575 @@
...
+
+## How to use it
+
+To use the Kubernetes deployer, you need:
+
+*   The ZenML `kubernetes` integration installed. If you haven't done so, run
+
+    ```shell
+    zenml integration install kubernetes
+    ```
```
strickvl:
```suggestion
```shell
zenml integration install kubernetes
```
```

File: src/zenml/integrations/kubernetes/flavors/kubernetes_deployer_flavor.py [hunk 6]
```diff
@@ -0,0 +1,456 @@
...
+        Returns:
+            The validated namespace value.
+
+        Raises:
+            ValueError: If the namespace is invalid.
+        """
+        if v is None:
+            return v
+
+        import re
```
strickvl: to the top?

src/zenml/integrations/kubernetes/flavors/kubernetes_deployer_flavor.py
```diff
@@ -0,0 +1,456 @@
...
+                f"Invalid namespace: '{v}'. Kubernetes namespaces must be lowercase "
+                "alphanumeric characters or hyphens, and must start and end with "
+                "an alphanumeric character."
+            )
+        if len(v) > 63:
+            raise ValueError(
+                f"Namespace too long: {len(v)} > 63 characters. "
+                f"Kubernetes resource names must be at most 63 characters."
+            )
+        return v
```
strickvl: smells like something that probably should be a helper function with its own test to validate it does what you / we think it should?

src/zenml/integrations/kubernetes/flavors/kubernetes_deployer_flavor.py
```diff
@@ -0,0 +1,456 @@
...
+            The validated settings.
+        """
+        # Warn if wait_for_load_balancer_timeout is set for non-LoadBalancer services
+        if (
+            self.service_type != "LoadBalancer"
+            and self.wait_for_load_balancer_timeout > 0
+        ):
+            from zenml.logger import get_logger
+
+            logger = get_logger(__name__)
```
strickvl: Probably shove this up at the top of the file

strickvl: both the import and the logger var.

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1188,193 @@ def apply_default_resource_requests(
...
+    Returns:
+        Tuple of (requests, limits, replicas) in Kubernetes format.
+        - requests: Dict with 'cpu', 'memory', and optionally 'nvidia.com/gpu' keys
+        - limits: Dict with 'cpu', 'memory', and optionally 'nvidia.com/gpu' keys
+        - replicas: Number of replicas
+
+    Raises:
+        ValueError: If replica configuration is invalid.
+    """
+    from zenml.config.resource_settings import ByteUnit
```
strickvl: up top?

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1188,193 @@ def apply_default_resource_requests(
...
+        memory_value = resource_settings.get_memory(unit=ByteUnit.MIB)
+        if memory_value is not None:
+            # Use Gi only for clean conversions to avoid precision loss
+            if memory_value >= MIB_TO_GIB and memory_value % MIB_TO_GIB == 0:
+                memory_str = f"{int(memory_value / MIB_TO_GIB)}Gi"
+            else:
+                memory_str = f"{int(memory_value)}Mi"
+
+            requests["memory"] = memory_str
+            limits["memory"] = memory_str
```
strickvl: some of this stuff has the feel that they need to be helper functions with their own tests. perhaps not. not sure if that adds more overhead, but there's at least some logic here which feels like it might want to be tested?

src/zenml/integrations/kubernetes/kube_utils.py
strickvl, line 1: some of these new util functions feel like they want tests adding for them. also they are easy wins for unit tests, it seems.

src/zenml/integrations/kubernetes/template_engine.py
strickvl, line 1: this jinja / template stuff feels like scope creep to me. Possibly worthy of removing from this PR and reviewing / implementing it separately?

tests/integration/integrations/kubernetes/deployers/test_kubernetes_deployer.py
strickvl, line 1: not sure these new tests do much that's super useful? wdyt? also there's an orphan comment at the bottom.

src/zenml/integrations/kubernetes/template_engine.py
schustmi: Yes please

src/zenml/integrations/kubernetes/template_engine.py
```diff
@@ -0,0 +1,572 @@
...
+
+import json
+import os
+import re
+import tempfile
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+import yaml
+from jinja2 import (
```
schustmi: I don't think `jinja` gets installed as part of the Kubernetes integration?

docs/book/component-guide/deployers/kubernetes.md
```diff
@@ -0,0 +1,574 @@
...
+* you want production-grade features like autoscaling, health probes, and high availability.
+* you need to deploy to on-premises infrastructure or air-gapped environments.
+* you want to leverage existing Kubernetes expertise and tooling in your organization.
+* you need to integrate with existing Kubernetes resources (Ingress, NetworkPolicies, ServiceMonitors, etc.).
+
+## How to deploy it
+
+{% hint style="info" %}
+Would you like to skip ahead and deploy a full ZenML cloud stack already, including a Kubernetes deployer? Check out the ZenML stack deployment guides for [AWS](https://docs.zenml.io/how-to/infrastructure-deployment/stack-deployment/deploy-a-cloud-stack-with-terraform), [GCP](https://docs.zenml.io/how-to/infrastructure-deployment/stack-deployment/deploy-a-cloud-stack-with-terraform), and [Azure](https://docs.zenml.io/how-to/infrastructure-deployment/stack-deployment/deploy-a-cloud-stack-with-terraform).
+{% endhint %}
```
stefannica:
```suggestion
```

This is incorrect. We don't deploy kubernetes with Terraform.

docs/book/component-guide/deployers/kubernetes.md
```diff
@@ -0,0 +1,574 @@
...
+* Cloud provider authentication (EKS, GKE, AKS)
+
+#### Configuration use-case: In-cluster deployment
+
+If your ZenML server runs inside the same Kubernetes cluster where you want to deploy pipelines, you can use in-cluster authentication:
+
+```shell
+zenml deployer register <DEPLOYER_NAME> \
+    --flavor=kubernetes \
+    --kubernetes_namespace=zenml-deployments
```
stefannica, line 106:
```suggestion
zenml deployer register <DEPLOYER_NAME> \
    --flavor=kubernetes \
    --incluster=True \
    --kubernetes_namespace=zenml-deployments
```

docs/book/component-guide/deployers/kubernetes.md
```diff
@@ -0,0 +1,574 @@
...
+
+You can now [deploy any ZenML pipeline](https://docs.zenml.io/concepts/deployment) using the Kubernetes deployer:
+
+```shell
+zenml pipeline deploy --name my_deployment my_module.my_pipeline
+```
+
+## Advanced configuration
+
+The Kubernetes deployer follows a progressive complexity model, allowing you to start simple and add configuration as needed:
```
stefannica, line 132: Loving this !

docs/book/component-guide/deployers/kubernetes.md
```diff
@@ -0,0 +1,574 @@
...
+* `{{service_port}}`: Service port number
+* `{{service_type}}`: Service type (LoadBalancer, NodePort, ClusterIP)
+* `{{labels}}`: Dict of ZenML-managed labels
+* `{{labels['managed-by']}}`: Always 'zenml'
+* `{{labels['zenml-deployment-id']}}`: Deployment UUID
+* `{{labels['zenml-deployment-name']}}`: Human-readable deployment name
+* `{{replicas}}`: Configured replica count
+* `{{image}}`: Container image
+* `{{deployment_id}}`: Deployment UUID
+* `{{annotations}}`: Pod annotations dict
```
stefannica: This feels limited and difficult maintain. I think users would rather have access to objects like the DeploymentResponse and DeploymentSettings here instead of individual fields.

docs/book/component-guide/deployers/kubernetes.md
```diff
@@ -0,0 +1,574 @@
...
+        livenessProbe:
+          httpGet:
+            path: {{ liveness_probe_path }}
+            port: {{ service_port }}
+          initialDelaySeconds: {{ liveness_probe_initial_delay }}
+          periodSeconds: {{ liveness_probe_period }}
+```
+
+{% hint style="info" %}
+**Tip**: Start with ZenML's built-in templates as a reference. You can find them in the ZenML repository at `src/zenml/integrations/kubernetes/templates/`. Copy and modify them for your needs.
```
stefannica: Perhaps a URL works better here

src/zenml/integrations/kubernetes/template_engine.py
```diff
@@ -0,0 +1,572 @@
...
+
+        try:
+            source_root = source_utils.get_source_root()
+            if source_root:
+                manifest_path = (
+                    Path(source_root) / ".zenml-deployments" / deployment_name
+                )
+                return str(manifest_path)
+        except Exception as e:
+            logger.debug(f"Could not use source root for manifests: {e}")
```
stefannica: You should keep these in `~/.config/zenml/deployments/kubernetes` instead of the repo root or source root.

src/zenml/integrations/kubernetes/template_engine.py
```diff
@@ -0,0 +1,572 @@
...
+
+        return obj_dict
+
+    # ========================================================================
+    # Template and Resource Rendering
+    # ========================================================================
+
+    def render_template(
+        self,
+        template_or_file: str,
```
stefannica: To avoid confusion and possible mistakes, can you please make this function only take `template_file` as input.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+        Returns:
+            Template context dictionary.
+        """
+        resources = {}
+        if resource_requests:
+            resources["requests"] = resource_requests
+        if resource_limits:
+            resources["limits"] = resource_limits
+
+        env_dict = dict(env_vars)
```
stefannica: dict() is unneeded nere

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+                            "name": secret_name,
+                            "key": key,
+                        }
+                    },
+                }
+            )
+
+        pod_settings_for_template = settings.pod_settings
+        if secret_env_list and settings.pod_settings:
+            existing_env = list(settings.pod_settings.env or [])
```
stefannica: list(..) is not needed here and .env is never None

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+            "deployment_name": resource_name,
+            "service_name": resource_name,
+            "namespace": namespace,
+            "deployment_id": deployment_id,
+            "labels": labels,
+            "annotations": settings.annotations,
+            "replicas": replicas,
+            "image": image,
+            "image_pull_policy": settings.image_pull_policy,
+            "image_pull_secrets": settings.image_pull_secrets or [],
```
stefannica: image_pull_secrets is never None

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+
+        Returns:
+            The deployment context.
+
+        Raises:
+            DeployerError: If the deployment context is not initialized.
+        """
+        if self._ctx is None:
+            raise DeployerError("Deployment context not initialized.")
+        return self._ctx
```
stefannica: Caching this context may lead to errors if the same deployer instance is used to manage different deployments without it being re-created. Please don't cache this as a deployer variable. Call `_initialize_deployment_context` directly from wherever it's needed.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+            incluster: Whether to use in-cluster config.
+
+        Returns:
+            Authenticated Kubernetes API client.
+
+        Raises:
+            RuntimeError: If connector returns invalid client type.
+        """
+        if incluster is None:
+            incluster = self.config.incluster
```
stefannica, line 138: This looks like it's based on the kubernetes orchestrator, where we recently changed the `incluster` behavior to allow service connector or kubecontext as a fallback. Can you please update this to reflect that same approach ?

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+        return k8s_client.CoreV1Api(self.get_kube_client())
+
+    @property
+    def k8s_applier(self) -> KubernetesApplier:
+        """Get or create Kubernetes Applier instance.
+
+        Returns:
+            Kubernetes Applier instance.
+        """
+        if not self._applier:
```
stefannica, line 209: FYI this will ignore service connector expiration. It should probably include this at the top:

```
if self.connector_has_expired():
    self._applier = None
```

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+            if not container_registry:
+                return False, "Container registry is required"
+
+            if not self.config.is_local and container_registry.config.is_local:
+                return False, (
+                    "Cannot use local container registry with remote Kubernetes cluster"
+                )
+
+            return True, ""
+
```
stefannica, line 381: The Kubernetes orchestrator has some validation here that checks the kubernetes contexts, if set. Can you also add them here ?

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+        snapshot = deployment.snapshot
+        assert snapshot, "Pipeline snapshot not found"
+
+        self._initialize_deployment_context(deployment)
+        ctx = self.require_ctx()
+
+        self._engine = KubernetesTemplateEngine(
+            custom_templates_dir=ctx.settings.custom_templates_dir
+            or self.config.custom_templates_dir
+        )
```
stefannica: You don't really need to save the engine as a deployer variable.

src/zenml/integrations/kubernetes/flavors/kubernetes_deployer_flavor.py
```diff
@@ -0,0 +1,440 @@
...
+        "If the stack component is linked to a Kubernetes service connector, "
+        "this field is ignored. Otherwise, it is mandatory.",
+    )
+
+    kubernetes_namespace: str = Field(
+        "zenml-deployments",
+        description="Default Kubernetes namespace for deployments. "
+        "Can be overridden per-deployment using the `namespace` setting. "
+        "Defaults to 'zenml-deployments'.",
+    )
```
stefannica: you should remove this and use the `namespace` attr in the settings.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+        snapshot = deployment.snapshot
+        if not snapshot:
+            raise DeployerError(
+                f"Deployment '{deployment.name}' has no snapshot"
+            )
+
+        settings = cast(
+            KubernetesDeployerSettings, self.get_settings(snapshot)
+        )
+        namespace = settings.namespace or self.config.kubernetes_namespace
```
stefannica: use `namespace` in settings and remove `kubernetes_namespace` in config.
```suggestion
        namespace = settings.namespace
```

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1009 @@
...
+        """
+        snapshot = deployment.snapshot
+        assert snapshot, "Pipeline snapshot not found"
+
+        self._initialize_deployment_context(deployment)
+        ctx = self.require_ctx()
+
+        self._engine = KubernetesTemplateEngine(
+            custom_templates_dir=ctx.settings.custom_templates_dir
+            or self.config.custom_templates_dir
```
stefannica: You don't need to check for both. The ZenML core will merge config into settings for you.


```suggestion
            custom_templates_dir=ctx.settings.custom_templates_dir
```

src/zenml/integrations/kubernetes/template_engine.py
stefannica: Code was only used by the dry-run feature and was removed

src/zenml/integrations/kubernetes/template_engine.py
stefannica: Done

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+                f"Deployment '{deployment.name}' has no snapshot"
+            )
+
+        settings = cast(
+            KubernetesDeployerSettings, self.get_settings(snapshot)
+        )
+        namespace = settings.namespace or self.config.kubernetes_namespace
+        resource_name = kube_utils.sanitize_label(f"zenml-{deployment.id}")[
+            :MAX_K8S_NAME_LENGTH
+        ]
```
stefannica:
```suggestion
        resource_name = kube_utils.sanitize_label(f"zenml-{deployment.id}")
```

The trimming is done by `sanitize_label` for you already

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+    DeploymentResponse,
+)
+from zenml.stack import StackValidator
+
+if TYPE_CHECKING:
+    from zenml.stack import Stack
+
+logger = get_logger(__name__)
+
+MAX_K8S_NAME_LENGTH = 63
```
stefannica: you don't need this

src/zenml/integrations/kubernetes/flavors/kubernetes_deployer_flavor.py
```diff
@@ -0,0 +1,449 @@
...
+            v: The namespace value to validate.
+
+        Returns:
+            The validated namespace value.
+        """
+        if v is None:
+            return v
+
+        kube_utils.validate_namespace_name(v)
+        return v
```
stefannica: You should never call integration specific code from the flavor. This will fail now if you don't have Kubernetes installed on your machine, even if you list stack components.

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -162,6 +178,33 @@ def sanitize_label(label: str) -> str:
...
+    - Must be lowercase alphanumeric characters or hyphens
+    - Must start and end with an alphanumeric character
+    - Must be at most 63 characters long
+
+    Args:
+        namespace: The namespace name to validate.
+
+    Raises:
+        ValueError: If the namespace name is invalid.
+    """
```
stefannica: This validation function applies to most Kubernetes resource names, not just namespaces. Can you name it differently.

Hint: you can also remove it, given that it's not really used anywhere (aside from the flavor validator, which shouldn't call it anyway).

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1216,191 @@ def apply_default_resource_requests(
...
+        Tuple of (requests, limits, replicas) in Kubernetes format.
+        - requests: Dict with 'cpu', 'memory', and optionally 'nvidia.com/gpu' keys
+        - limits: Dict with 'cpu', 'memory', and optionally 'nvidia.com/gpu' keys
+        - replicas: Number of replicas
+
+    Raises:
+        ValueError: If replica configuration is invalid.
+    """
+    requests: Dict[str, str] = {}
+    limits: Dict[str, str] = {}
```
stefannica, line 1211: why does this function also set CPU and memory limits to the same value used for requests? I think most users think of resource settings as requests, not limits and given that there's no way to configure them separately, you should not assume limits need to be set too.

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1216,191 @@ def apply_default_resource_requests(
...
+        limits["cpu"] = cpu_str
+
+    if resource_settings.memory is not None:
+        memory_value = resource_settings.get_memory(unit=ByteUnit.MIB)
+        if memory_value is not None:
+            # Use Gi only for clean conversions to avoid precision loss
+            if memory_value >= MIB_TO_GIB and memory_value % MIB_TO_GIB == 0:
+                memory_str = f"{int(memory_value / MIB_TO_GIB)}Gi"
+            else:
+                memory_str = f"{int(memory_value)}Mi"
```
stefannica: can't you use the `resource_settings.get_memory` again here instead of re-implementing that logic ?

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+            deployment, environment, secrets
+        )
+        ctx = self.require_ctx()
+
+        try:
+            created_objects, inventory = self.k8s_applier.provision(
+                rendered_resources,
+                default_namespace=ctx.namespace,
+                timeout=timeout,
+            )
```
stefannica: This implementation of `do_provision` completely ignores existing resources that may have been created in previous deployment calls and that aren't needed anymore, like what happens if some custom Jinja templates are removed.

You have to look at the existing inventory associated with the deployment and delete everything that is not in the new inventory.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+        """Wait for deployment and service to become ready.
+
+        Args:
+            deployment: The deployment response.
+            settings: Kubernetes deployer settings.
+            timeout: Timeout in seconds.
+
+        Raises:
+            DeploymentProvisionError: If deployment doesn't become ready in time.
+        """
```
stefannica, line 616: The implementation of this method relies on a few assumptions that aren't necessarily true if your user can completely override all the resources being applied:

* there's one and only one deployment with the name set as `zenml-<deployment-UUID>`
* there's a load balancer if `settings.service_type == "LoadBalancer"`

I think you don't have the luxury of relying on the settings or even the naming conventions anymore. What this function should do is attempt a best-effort generic "wait for deployment with any number of k8s resources of any type":

* list all deployments in the inventory, wait for each one
* list all services in the inventory, wait for each one that has a load-balancer

You should probably also save the new state (after wait) of the resource in the metadata inventory instead of the ones you got from running `apply`.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+            existing_env_vars: Optional dictionary of existing environment variables
+                to check for collisions.
+
+        Returns:
+            Sanitized secret key name.
+
+        Raises:
+            DeploymentProvisionError: If sanitization causes a collision with an existing key.
+        """
+        sanitized = re.sub(r"[^A-Za-z0-9_]", "_", key)
```
stefannica: You're silently converting the secret key here, but you should raise an error instead, otherwise the user will have to debug this at runtime, when it gets more difficult.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+                Any,
+                DeploymentEntrypointConfiguration.get_entrypoint_command(),
+            )
+        if not context.get("args"):
+            context["args"] = cast(
+                Any,
+                DeploymentEntrypointConfiguration.get_entrypoint_arguments(
+                    **{DEPLOYMENT_ID_OPTION: deployment.id}
+                ),
+            )
```
stefannica: These feel like they belong in the `_build_template_context` method, not here.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+
+        try:
+            stored_metadata = deployment.deployment_metadata
+            inventory_data = (
+                stored_metadata.get("resource_inventory")
+                if stored_metadata
+                else None
+            )
+
+            if not inventory_data:
```
stefannica: you should always rely on the inventory data stored in the metadata as your known last operational state. If this is not set, then simply assume there is nothing to deprovision instead of trying to clean up resources by selector. Think of `metadata` as playing the same role as the Terraform state file.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -0,0 +1,1037 @@
...
+                )
+
+            return None
+
+        except ApiException as e:
+            if e.status == 404:
+                logger.info(
+                    f"Deployment '{deployment.name}' not found (already deleted)"
+                )
+                return None
```
stefannica: you should do some research and be careful how you interpret these errors:

* what happens if only some resources are not found by the `self.k8s_applier.delete_from_inventory` call ?
* what happens if some resources cannot be deleted by the `self.k8s_applier.delete_from_inventory` call ?

whatever you do, you should not return `None` or raise `DeploymentNotFoundError` here unless you're absolutely sure that all resources have been deprovisioned or will for sure be deprovisioned, otherwise you'll end up with orphan resources.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
safoinme: yes, fixed

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
safoinme: re-worked this to call ctx everytime instead and also added few more needed fields

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1187,180 @@ def apply_default_resource_requests(
...
+        # Neither specified - default to 1
+        replicas = 1
+
+    if (
+        resource_settings.gpu_count is not None
+        and resource_settings.gpu_count > 0
+    ):
+        # GPU requests must be integers; Kubernetes auto-sets requests=limits for GPUs
+        gpu_str = str(resource_settings.gpu_count)
+        requests["nvidia.com/gpu"] = gpu_str
```
stefannica, line 1238: my previous comment only applied to CPU/ram. I think you _have_ to set both requests and limits for GPU requirements.

src/zenml/integrations/kubernetes/k8s_applier.py
```diff
@@ -0,0 +1,570 @@
...
+        super().__init__(message)
+        self.inventory = inventory
+        self.errors = errors
+
+
+def _flatten_items(objs: Iterable[Dict[str, Any]]) -> Iterable[Dict[str, Any]]:
+    """Yield single resources from possibly 'List' wrappers.
+
+    Args:
+        objs: Iterable of resource dicts (already normalized). Supports 'kind: List'.
```
schustmi: `Supports 'kind: List'.` This is not clear at all what it does

src/zenml/integrations/kubernetes/k8s_applier.py
```diff
@@ -0,0 +1,570 @@
...
+            "Service",
+        )
+        host = self._svc_lb_host(svc)
+        if not host:
+            raise RuntimeError(
+                "Service is ready but no external IP/hostname found"
+            )
+        return host
+
+    def _svc_lb_host(self, service_obj: Any) -> Optional[str]:
```
schustmi, line 620: Why `Any`

src/zenml/integrations/kubernetes/k8s_applier.py
```diff
@@ -0,0 +1,570 @@
...
+    """
+    if isinstance(resource, dict):
+        return resource
+    if hasattr(resource, "to_dict"):
+        d = api_client.sanitize_for_serialization(resource)
+        if not isinstance(d, dict):
+            raise ValueError(
+                f"Expected dict after serialization, got {type(d)}"
+            )
+        if "api_version" in d and "apiVersion" not in d:
```
schustmi: Is that a bug in the Kubernetes package that they don't convert this one exact key?

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1187,191 @@ def apply_default_resource_requests(
...
+    Args:
+        core_api: Kubernetes Core V1 API client.
+        service: Kubernetes Service resource.
+        namespace: Kubernetes namespace.
+        ingress: Optional Kubernetes Ingress resource.
+
+    Returns:
+        Service URL or None if not yet available.
+    """
+    # Handle ingress
```
schustmi: Can you please remove all these useless comments.

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1187,191 @@ def apply_default_resource_requests(
...
+        cpu_value = resource_settings.cpu_count
+        # Kubernetes accepts CPU as whole numbers (e.g., "2") or millicores (e.g., "500m")
+        if cpu_value < 1:
+            # Fractional CPUs: 0.5 → "500m"
+            cpu_str = f"{int(cpu_value * 1000)}m"
+        else:
+            if cpu_value == int(cpu_value):
+                cpu_str = str(int(cpu_value))  # 2.0 → "2"
+            else:
+                cpu_str = f"{int(cpu_value * 1000)}m"  # 1.5 → "1500m"
```
schustmi: isn't this much simpler and the same?
```python
if cpu_value == int(cpu_value):
    cpu_str = str(int(cpu_value))  # 2.0 → "2"
else:
    cpu_str = f"{int(cpu_value * 1000)}m"
```

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1187,191 @@ def apply_default_resource_requests(
...
+
+        requests["cpu"] = cpu_str
+
+    if resource_settings.memory is not None:
+        memory_mib = resource_settings.get_memory(unit=ByteUnit.MIB)
+        if memory_mib is not None:
+            requests["memory"] = f"{int(memory_mib)}Mi"
```
schustmi:
```python
if memory_mib := resource_settings.get_memory(unit=ByteUnit.MIB)
  requests["memory"] = f"{int(memory_mib)}Mi"
```

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1187,191 @@ def apply_default_resource_requests(
...
+    # Determine replica count from min/max settings
+    # For standard K8s Deployments, we use min_replicas as the baseline
+    # (autoscaling requires a separate HPA resource)
+    min_r = resource_settings.min_replicas
+    max_r = resource_settings.max_replicas
+
+    # Validate and determine replica count
+    if min_r is not None and max_r is not None:
+        if min_r > max_r:
+            raise ValueError(
```
schustmi: This should probably be a validator on the resource settings?

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1171,3 +1187,191 @@ def apply_default_resource_requests(
...
+        # Only max_replicas specified - use it as baseline
+        replicas = max(1, max_r)  # At least 1
+    else:
+        # Neither specified - default to 1
+        replicas = 1
+
+    if (
+        resource_settings.gpu_count is not None
+        and resource_settings.gpu_count > 0
+    ):
```
schustmi: `if resource_settings.gpu_count`

src/zenml/integrations/kubernetes/template_engine.py
```diff
@@ -0,0 +1,157 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Kubernetes template engine."""
+
+import json
+from typing import Any, Dict, List
+
+import yaml
+from jinja2 import Environment, StrictUndefined, TemplateError, Undefined
```
schustmi, line 20: My question still stands: Where is jinja getting installed from?

schustmi (changes requested):

src/zenml/integrations/kubernetes/k8s_applier.py
safoinme: Updated docstring to explain this. 

Kubernetes allows manifests that wrap [multiple resources in a single object](https://kubernetes.io/docs/reference/using-api/api-concepts/)
```
kind: List
items:
  - kind: Deployment
    ...
  - kind: Service
    ...
```

src/zenml/integrations/kubernetes/k8s_applier.py
safoinme: Because we are using kubernetes client which also get resource type as any and basically it could return:
	•	a raw dict coming from to_dict() or some other JSON-ish source, or
	•	a Kubernetes client / dynamic client object (e.g. ResourceInstance) that has a .to_dict() method but doesn’t have a nice static type in the Python client.

src/zenml/integrations/kubernetes/k8s_applier.py
safoinme: technically kubernetes only accepted `apiVersion` but the `api_version` is usually the variable name in python client so this is just extra step to not fail on that naming otherwise this will happen because user provided wrong key and we can also just fail

src/zenml/integrations/kubernetes/template_engine.py
safoinme: I have just added it to k8s integration requirements

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
safoinme: updated based on this

PR #4121, opened 2025-11-03
Print ZenML UI Deployment Link to Terminal 

safoinme:
## Describe changes
Linked to #4118 

## Pre-requisites
Please ensure you have done the following:
- [ ] I have read the **CONTRIBUTING.md** document.
- [ ] I have added tests to cover my changes.
- [ ] I have based my new branch on `develop` and the open PR is targeting `develop`. If your branch wasn't based on develop read [Contribution guide on rebasing branch to develop](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md#-pull-requests-rebase-your-branch-on-develop).
- [ ] **IMPORTANT**: I made sure that my changes are reflected properly in the following resources:
  - [ ] [ZenML Docs](https://docs.zenml.io)
  - [ ] Dashboard: Needs to be communicated to the frontend team.
  - [ ] Templates: Might need adjustments (that are not reflected in the template tests) in case of non-breaking changes and deprecations.
  - [ ] [Projects](https://github.com/zenml-io/zenml-projects): Depending on the version dependencies, different projects might get affected.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other (add details above)

PR #4130, opened 2025-11-05
Improve deployment CLI output for easier command copying

strickvl:
This PR improves the UX of the deployment CLI output by making management commands easier to copy and paste into the terminal.

## Changes

- Replaced the bordered Rich Table format with a simple list layout in the deployment management commands section
- Commands now appear on their own lines in bold green text
- Descriptions appear indented and dimmed below each command
- Removed table box-drawing characters that interfered with text selection

## Motivation

The previous table format with rounded borders made it difficult to copy commands cleanly. When users tried to select commands to paste into their terminal, they would often end up selecting the box-drawing characters along with the command text, requiring manual cleanup.

## Example Output

Before: Commands in a bordered table (hard to select)
After:
```
⚙️  Management Commands

zenml deployment logs weather_agent -f
  Follow deployment logs in real-time

zenml deployment describe weather_agent
  Show detailed deployment information
```

## Testing

- Code formatting passed with `bash scripts/format.sh`
- Visual output can be tested by running any deployment command (e.g., `zenml pipeline deploy`)

### DIFF
--- a/src/zenml/cli/utils.py
+++ b/src/zenml/cli/utils.py
@@ -2640,7 +2640,7 @@ def pretty_print_deployment(
         else:
             declare("  [dim]No metadata available[/dim]")
 
-    # Management section: bold white header and commands, no decorative emojis or green color
+    # Management section
     declare("\n[bold]Management commands[/bold]\n")
 
     console.print(f"[bold]zenml deployment logs {deployment.name} -f[/bold]")


htahir1:
@strickvl note this might clash with https://github.com/zenml-io/zenml/pull/4121

htahir1 (approved):
It *looks* fine but im not sure how the output looks so a screenshot would be nice

strickvl:
> It _looks_ fine but im not sure how the output looks so a screenshot would be nice

Here's what it looks like now:

CleanShot 2025-11-07 at 09 25 31@2x

I'd love it if @znegrin had thoughts. It feels like a lot of green now, which idk if we ever had in our CLI before?

znegrin:
@strickvl I think it should be more consistent with the rest of the CLI, where we use green only for the success messages. I'd rather show them in the default white. I'm also not a big fan of the emojis, and the checkmark for the URL doesn't make sense. I think it would be nice to see how other commands work and are displayed in our CLI, and try to keep the consistency.

strickvl:
> @strickvl I think it should be more consistent with the rest of the CLI, where we use green only for the success messages. I'd rather show them in the default white. I'm also not a big fan of the emojis, and the checkmark for the URL doesn't make sense. I think it would be nice to see how other commands work and are displayed in our CLI, and try to keep the consistency.

Ok I'll work on this. Note that most of what you said is stuff that was already implemented and shipped, so doesn't really relate to my specific update, but I'll take a look at the bigger picture too.

src/zenml/cli/utils.py
```diff
@@ -2493,11 +2493,11 @@ def pretty_print_deployment(
         show_schema: Whether to show the schema.
         no_truncate: Whether to truncate the metadata.
     """
-    # Header section
-    status = format_deployment_status(deployment.status)
+    # Header section: show deployment name in bold cyan and plain uppercase status with status emoji.
```
strickvl:
```suggestion
    # Header section
```

src/zenml/cli/utils.py
```diff
@@ -2578,32 +2577,34 @@ def pretty_print_deployment(
...
     if show_metadata:
-        declare("\n📋 [bold]Deployment Metadata[/bold]")
+        declare("\n[bold]Deployment metadata:[/bold]")
 
         # Get the metadata - it could be from deployment_metadata property or metadata
         metadata = deployment.deployment_metadata
 
         if metadata:
-            # Recursively format nested dictionaries and lists
+            # Recursively format nested dictionaries and lists with minimal styling for readability
```
strickvl:
```suggestion
            # Recursively format nested dictionaries and lists
```

src/zenml/cli/utils.py
```diff
@@ -2639,35 +2640,24 @@ def format_value(value: Any, indent_level: int = 0) -> str:
         else:
             declare("  [dim]No metadata available[/dim]")
 
-    # Management section
-    declare("\n⚙️  [bold]Management Commands[/bold]")
+    # Management section: bold white header and commands, no decorative emojis or green color
```
strickvl:
```suggestion
    # Management section
```

strickvl:
> @strickvl I think it should be more consistent with the rest of the CLI, where we use green only for the success messages. I'd rather show them in the default white. I'm also not a big fan of the emojis, and the checkmark for the URL doesn't make sense. I think it would be nice to see how other commands work and are displayed in our CLI, and try to keep the consistency.

https://app.warp.dev/block/rPI6jHX9Nxsu0b8FMCrRh8

CleanShot 2025-11-10 at 11 35 16@2x

This is what it looks like in my CLI now @znegrin. I did the following:

- removed emojis
- removed green styling
- applied consistency fixes as suggested
- (kept status emojis)

PR #4138, opened 2025-11-06
Misc small fixes

schustmi:
## Describe changes
- Correctly handle keyboard interrupt during run monitoring
- Improve warning message for unused docker settings
- Fail when passing invalid stack component config key

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/warnings/registry.py
+++ b/src/zenml/utils/warnings/registry.py
@@ -49,7 +49,7 @@ WARNING_CONFIG_REGISTRY = {
         ## Unused docker settings warning
         
         The user provides a valid docker settings object but the stack
-        does not containerized components and will not make use of it.
+        does not have any components that make use of it.
         """,
         verbosity=WarningVerbosity.LOW,
         severity=WarningSeverity.MEDIUM,


AlexejPenner (approved):
🦭

src/zenml/stack/utils.py
```diff
@@ -80,14 +80,15 @@ def validate_stack_component_config(
...
-    # Make sure extras are forbidden for the config class. Due to inheritance
-    # order, some config classes allow extras by accident which we patch here.
-    validation_config_class: Type[StackComponentConfig] = type(
-        config_class.__name__,
-        (config_class,),
-        {"model_config": {"extra": "ignore"}},
-    )
-    configuration = validation_config_class(**configuration_dict)
+
+    extra_keys = set(configuration_dict) - set(config_class.model_fields)
```
schustmi, line 86: @stefannica In some PR, you made all of our pydantic classes have `extra=ignore`, and this included the validation of stack component configurations here. That lead to the following: Whenever some invalid attribute was passed e.g. using the CLI, it would simply be ignored silently.
I think was mainly to have less breaking changes when using different client/server versions, but this is something that is not happening as part of that and only with manual user interaction. Which is why I think it's okay to revert this?

stefannica: Yes, it should be okay. This is only used during create/update stack/stack component.

PR #4140, opened 2025-11-06
[Bug:2258] Allow case sensitive updates

Json-Andriopoulos:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -12260,7 +12260,6 @@ class SqlZenStore(BaseZenStore):
             raise RuntimeError(f"Schema {schema_class.__name__} has no name.")
 
         operation: Literal["create", "update"] = "create"
-        project_id: Optional[UUID] = None
         if isinstance(resource, BaseRequest):
             # Create operation
             if isinstance(resource, ProjectScopedRequest):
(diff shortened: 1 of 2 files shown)


Json-Andriopoulos:
Behavior is caused by the following:

1. When creating or updating entities we do a check to verify that no existing names are used.
2. MySQL default collation (varchar value handling) handles names in a case-insensitive way (for optimization of join/sort operations). Therefore even case variants of the name are found in the DB and raise the EntityExistsExeption.
3. We introduce a simple chang for updates that filters out the update target entity itself. So you can update/rename a component with a case variant.
4. You still can't create name case variants for other entities (and shouldn't).
5. The change affects the core name verification component so similar issues with other types (Stack is the reported one on the issue) will also be fixed.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -12245,6 +12244,10 @@ def _verify_name_uniqueness(
 
         query = select(schema_class).where(schema_class.name == name)
 
+        # Exclude from existing name checks the update target itself.
+        if isinstance(resource, BaseUpdate):
```
schustmi: We should make sure that `schema` in this case is actually an instance right, otherwise `schema` and `schema_class` are the same, and we'll have an `WHERE schema_class.id=schema_class.id` in the query?

Json-Andriopoulos: Yeap, good point. I think this is always the case with updates but it doesn't hurt to do the actual check.

schustmi (approved):

PR #4144, opened 2025-11-06
Support Newer Version for Databricks Orchestrator

safoinme:
## Describe changes
Updating the way files are sent to Databricks Workspace to support newer version of DRB.

## Pre-requisites

## Types of changes


src/zenml/integrations/databricks/orchestrators/databricks_orchestrator.py
```diff
@@ -67,9 +68,9 @@
 logger = get_logger(__name__)
 
 ZENML_STEP_DEFAULT_ENTRYPOINT_COMMAND = "entrypoint.main"
-DATABRICKS_WHEELS_DIRECTORY_PREFIX = "dbfs:/FileStore/zenml"
+DATABRICKS_WHEELS_DIRECTORY_PREFIX = "/Workspace/Shared/.zenml"
```
htahir1, line 71: this breaks current pipelines maybe? add a warning?

src/zenml/integrations/databricks/orchestrators/databricks_orchestrator.py
```diff
@@ -313,18 +314,23 @@ def _construct_databricks_pipeline(
...
+        databricks_client.workspace.mkdirs(path=databricks_directory)
+
+        # Upload the wheel file using Workspace API (works with DBR 15+)
+        with open(wheel_path, "rb") as f:
+            databricks_client.workspace.upload(
+                path=databricks_wheel_path,
+                content=f.read(),
+                format=ImportFormat.AUTO,
+                overwrite=True,
+            )
```
htahir1: a try catch wouldnt hurt?

htahir1:
```try:
    # Create parent directory structure in Workspace
    databricks_client.workspace.mkdirs(path=databricks_directory)
    
    # Upload the wheel file using Workspace API (works with DBR 15+)
    with open(wheel_path, "rb") as f:
        databricks_client.workspace.upload(
            path=databricks_wheel_path,
            content=f.read(),
            format=ImportFormat.AUTO,
            overwrite=True,
        )
    logger.info(f"Successfully uploaded wheel to {databricks_wheel_path}")
except Exception as e:
    logger.error(f"Failed to upload wheel to Databricks: {e}")
    raise```

htahir1: e.g.

htahir1 (changes requested):
Two questions:

1. Has this been tested manually with both DBR 14.x and 16.x?
2. What happens if a user is still on DBR 13 or 14?

src/zenml/integrations/databricks/orchestrators/databricks_orchestrator.py
safoinme: No it doesn't this is something that supports for older version to install from `/Workspace/` but the opposite is not possible because of security implications

safoinme:
> We might need doc changes to tell people that it only works on 16.x now?

We don't need too what we have right now [was supported since 11.x,](https://docs.databricks.com/aws/en/libraries/#recommendations-for-uploading-libraries).
It was possible to install either using workspace libraries(this is what we used to have and got deprecated after 15.x) or workspace files (what we are doing right now and it's supported in all version that Databricks have support for)

htahir1:
@safoinme please fix docstring things

PR #4146, opened 2025-11-06
Add `skip_owner_references` setting and make `incluster` setting dynamic for the Kubernetes orchestrator; Add `runtime_environment` to DockerSettings

stefannica:
## Describe changes
Change the behavior of the `incluster` setting to fall back to the configured service connector or kubernetes context if the in-cluster authentication fails. This is needed to run pipelines from the UI in cases where a service connector is not feasible for security reasons and the client uses the locally available credentials.

This change allows users to set `incluster` to True at the same time as configuring a service connector or local kubernetes context. The orchestrator will try to use in-cluster authentication, if possible, and if not, fall back on the other configured source of credentials.

This PR also adds a new setting - `skip_owner_references` - which can be used to skip transferring the ownership of the step jobs and pods over to the main orchestrator job. This is necessary for use-cases where this operation is forbidden by cluster security rules.

A `runtime_environment` setting has been added to the DockerSettings. This makes it possible to set environment variables _after_ all the RUN entries are done. This is very useful for configuring environment variables such as HTTP_PROXY and HTTPS_PROXY, which are meant for runtime and not for build time. 

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -2123,11 +2123,12 @@ class Client(metaclass=ClientMetaClass):
 
         if configuration is not None:
             existing_configuration = component.configuration
-            existing_configuration.update(configuration)
-            existing_configuration = {
-                k: v
-                for k, v in existing_configuration.items()
-                if v is not None
+            merged_configuration = {
+                **existing_configuration,
+                **configuration,
+            }
+            merged_configuration = {
+                k: v for k, v in merged_configuration.items() if v is not None
             }
 
             from zenml.stack.utils import (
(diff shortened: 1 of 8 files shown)


src/zenml/config/docker_settings.py
```diff
@@ -239,6 +241,7 @@ class DockerSettings(BaseSettings):
     local_project_install_command: Optional[str] = None
     apt_packages: List[str] = []
     environment: Dict[str, Any] = {}
+    post_environment: Dict[str, Any] = {}
```
schustmi: Maybe something like `runtime_environment`? `post_environment` is so unspecific to me, post what?

stefannica: done

AlexejPenner: agreed on this

PR #4159, opened 2025-11-10
Map/Reduce implementation for dynamic pipelines

schustmi:
## Describe changes

This PR adds an initial map/reduce implementation for dynamic pipelines.

### Example code
```python

@step
def producer() -> list[int]:
    return [1, 2, 3]

@step
def worker(value: int) -> int:
    return value * 2

@step
def reducer(values: list[int]) -> int:
    return sum(values)


@pipeline(dynamic=True, enable_cache=False)
def map_reduce():
    values = producer()
    results = worker.map(values)
    reducer(results)


if __name__ == "__main__":
    map_reduce()
```

### Features
- Add new `get_item_count(...)` and `load_item(...)` methods to base materializer. The item count is stored for each artifact version so that we don't have to load the data in the orchestration environment to determine how many steps to launch.
- Enables passing lists of artifacts as step inputs. This enables the above syntax where we simply pass the output of the `map(...)` operation to the reducer step.
- A step has both a `.map(...)` as well as a `.product(...)` method. When multiple sequence-like inputs are passed to these, they behave in the following way:
  - `step.map(...)` ensures that all inputs are the same length `n`, and then creates `n` steps. The first step gets as input the first element of all inputs.
  - `step.product(...)` creates a step for each combination of the elements in the input sequences.
- If you want to pass a sequence-like artifact to `.map(...)` without splitting it up across the steps (-> full artifact is passed to all steps), you can use the `unmapped(...)` function.
```python
from zenml import pipeline, step, unmapped

@step
def producer(length: int) -> list[int]:
  return [1] * length

@step
def consumer(a: int, b: list[int]) -> None:
    pass

@pipeline(dynamic=True)
def unmapped_example():
  a = producer(length=3)
  b = producer(length=4)
  consumer.map(a=a, b=unmapped(b))
```
- Because the backend representation is changed to allow passing artifact lists instead of just single artifacts, this also makes the reduce part for static pipelines much nicer:
```python

@step
def producer() -> int:
  return 1

@step
def reducer(values: list[int]) -> int:
    return sum(values)

@pipeline
def static_reduce():
  # The step amount is still static..
  values = [producer_step(...) for _ in range(10)]
  # ..but we can pass the list of artifacts directly to the downstream step
  reducer(values)
```

### Limitations
- Right now, only mapping over step output artifacts from the same pipeline run is allowed. Mapping over raw data or external artifacts is not possible
- The chunk size is currently not configurable and defaults to `1`.

### Breaking changes
- `ArtifactVersionResponse.regular_inputs` and `StepContext.inputs` is now of type `Dict[str, List[ArtifactVersionResponse]]` instead of `Dict[str, ArtifactVersionResponse]`
- ~~Client-Server incompatibility because of changes to the `StepSpec.inputs` type~~ This has been solved in some other way to keep this version compatible with older clients.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/pandas/materializers/pandas_materializer.py
+++ b/src/zenml/integrations/pandas/materializers/pandas_materializer.py
@@ -341,26 +341,32 @@ class PandasMaterializer(BaseMaterializer):
             return int(data.shape[1])
         return None
 
-    def load_chunk(
-        self, data_type: Type[Any], chunk_index: int, chunk_size: int
-    ) -> Any:
-        """Load a specific chunk of the data.
+    def load_item(self, data_type: Type[Any], index: int) -> Any:
+        """Load a specific item of the data.
 
         Args:
             data_type: The type of the data to load.
-            chunk_index: The index of the chunk to load.
-            chunk_size: The size of the chunk to load.
+            index: The index of the item to load.
 
         Returns:
-            The loaded chunk of the data.
+            The loaded item.
         """
         data = self.load(data_type)
-        if chunk_size == 1:
-            chunk = data_type(data[chunk_index])
-        else:
-            chunk = data[chunk_index : chunk_index + chunk_size]
+        item = data[index]
 
-        if not isinstance(chunk, data_type):
-            chunk = data_type(chunk)
+        if not isinstance(item, data_type):
+            try:
+                item = data_type(item)
+            except Exception as e:
+                # We only log an error here, potentially pydantic can handle the
+                # conversion when validating step function inputs.
+                logger.error(
+                    "Failed to convert item `%s` to expected type `%s`. This "
+                    "is most likely due to a mismatching type annotation on "
+                    "your step function input. Error: %s",
+                    item,
+                    data_type.__name__,
+                    e,
+                )
 
-        return chunk
+        return item
--- a/src/zenml/materializers/built_in_materializer.py
+++ b/src/zenml/materializers/built_in_materializer.py
@@ -569,46 +569,52 @@ class BuiltInContainerMaterializer(BaseMaterializer):
             return len(data)
         return None
 
-    def load_chunk(
-        self, data_type: Type[Any], chunk_index: int, chunk_size: int
-    ) -> Any:
-        """Load a specific chunk of the data.
+    def load_item(self, data_type: Type[Any], index: int) -> Any:
+        """Load a specific item of the data.
 
         Args:
             data_type: The type of the data to load.
-            chunk_index: The index of the chunk to load.
-            chunk_size: The size of the chunk to load.
+            index: The index of the item to load.
 
         Raises:
             RuntimeError: If the metadata format is not supported.
 
         Returns:
-            The loaded chunk of the data.
+            The loaded item.
         """
         if self.artifact_store.exists(self.data_path):
-            outputs = yaml_utils.read_json(self.data_path)
-            outputs = outputs[chunk_index : chunk_index + chunk_size]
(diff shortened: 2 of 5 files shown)


src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -325,3 +325,42 @@ def extract_metadata(
...
+            data: The data to get the number of items for.
+
+        Returns:
+            The number of items for the given data.
+        """
+        if isinstance(data, pd.Series):
+            return len(data)
+        elif isinstance(data, pd.DataFrame):
+            return int(data.shape[1])
+        return None
```
stefannica, line 342: My immediate thought to this is that we probably need to be more formal about sequence-typed artifacts and also store the artifact element data type, in addition to the count ('get_data_type'). We would need this to validate and/or return the correct data type when unpacking elements in 'load_chunk'

src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -325,3 +325,42 @@ def extract_metadata(
...
+            The loaded chunk of the data.
+        """
+        data = self.load(data_type)
+        if chunk_size == 1:
+            chunk = data_type(data[chunk_index])
+        else:
+            chunk = data[chunk_index : chunk_index + chunk_size]
+
+        if not isinstance(chunk, data_type):
+            chunk = data_type(chunk)
```
stefannica: This sort of forced type conversion can lead to some unexpected results. I think we need to store the data type of individual elements inside the collection in the artifact DB entry and validate the requested data type here. We do validate the loaded data against the 'data_type' supplied value in some materializers, but not in all of them. The same should apply here.

src/zenml/integrations/pandas/materializers/pandas_materializer.py
```diff
@@ -325,3 +325,42 @@ def extract_metadata(
...
+            data_type: The type of the data to load.
+            chunk_index: The index of the chunk to load.
+            chunk_size: The size of the chunk to load.
+
+        Returns:
+            The loaded chunk of the data.
+        """
+        data = self.load(data_type)
+        if chunk_size == 1:
+            chunk = data_type(data[chunk_index])
```
stefannica: This behavior is not consistent. What if the user code really needs to load just a slice of length = 1? I think you need a separate function for this sub-case, or an argument that makes it clear that the user wants to load just one element.

src/zenml/materializers/built_in_materializer.py
```diff
@@ -555,3 +555,60 @@ def compute_content_hash(self, data: Any) -> Optional[str]:
...
+                    element = materializer.load(type_)
+                    outputs.append(element)
+
+            else:
+                raise RuntimeError(f"Unknown metadata format: {metadata}.")
+
+        if len(outputs) == 1:
+            return data_type(outputs[0])
+
+        return data_type(outputs)
```
stefannica: same thing here: blindly converting the data type without any validation will lead to all sorts of hard to debug errors. We need to use explicit data type validation here.

src/zenml/integrations/pandas/materializers/pandas_materializer.py
schustmi: I thought about this as well, but what would the element type of a DataFrame even be? Somehow a tuple of all the column types?

stefannica:
@claude give me a michael-review please

stefannica (approved):

PR #4178, opened 2025-11-13
Cache key index for faster cache lookup

schustmi:
## Describe changes
This PR adds an additional index on the `step_run.cache_key` column for faster and more efficient cache lookups.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -81,6 +84,12 @@ class StepRunSchema(NamedSchema, RunMetadataInterface, table=True):
             "version",
             name="unique_step_name_for_pipeline_run",
         ),
+        build_index(
```
Json-Andriopoulos, line 87: General question: Are we including indices on the table definition as a reference to what the active indices are for each table?

I guess all of them are b-trees for now but can we extend the utility to include that information as well?

src/zenml/zen_stores/migrations/versions/4dd9d3afd2c0_cache_key_index.py
```diff
@@ -0,0 +1,35 @@
...
+# revision identifiers, used by Alembic.
+revision = "4dd9d3afd2c0"
+down_revision = "0.91.1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
```
Json-Andriopoulos, line 20: My OCD would get rid of this comment 🥶

Json-Andriopoulos (approved):
LGTM

src/zenml/zen_stores/schemas/step_run_schemas.py
schustmi: Not just as a reference, but it's also used for the following: When we connect to a database, one of two things happens:
- If the DB is empty (no `alembic_version` table exists), we have never connected to this DB before. Instead of running all DB migrations, we initialize the DB from the current SQLModel classes. For that reason, all the indexes need to be defined on the model classes, otherwise they would be missing.
- If we've connected to the DB before, we run alembic migrations.

Yep all the indices right now are b-trees, we never explicitly specify anything else. I guess we could add that to the `build_index` helper function to make it more explicit?

src/zenml/zen_stores/migrations/versions/4dd9d3afd2c0_cache_key_index.py
schustmi: I usually leave it as is, unless I modify the code in some way, at that point I remove it 😅

src/zenml/zen_stores/schemas/step_run_schemas.py
Json-Andriopoulos: Gotcha thanks. No need to update build index now we can do it on the first index use case that needs something different (I am guessing this won't be the case any time soon).

PR #4187, opened 2025-11-13
Fix documentation errors in containerization and Vertex orchestrator

strickvl:
Fixes two documentation issues:

1. **Containerization docs**: Missing comma and unclosed string in `pyproject_export_command` example
2. **Vertex orchestrator docs**: Service account key filename incorrectly suggested "workload" when it should be "client"


htahir1 (approved):
beautiful!

Issue #4185, opened 2025-11-13
UnboundLocalError: cannot access local variable 'cert_content' where it is not associated with a value


bamaas:
### System Information

~/git/project/zenml main*
.venv ❯ zenml info -a -s                    
ZENML_LOCAL_VERSION: 0.91.1
ZENML_SERVER_VERSION: 0.91.1
ZENML_SERVER_DATABASE: mysql
ZENML_SERVER_DEPLOYMENT_TYPE: kubernetes
ZENML_CONFIG_DIR: /home/execfalcon/.config/zenml
ZENML_LOCAL_STORE_DIR: /home/execfalcon/.config/zenml/local_stores
ZENML_SERVER_URL: https://zenml.project-d.platform.company.com
ZENML_ACTIVE_REPOSITORY_ROOT: None
PYTHON_VERSION: 3.12.11
ENVIRONMENT: wsl
SYSTEM_INFO: {'os': 'linux', 'linux_distro': 'ubuntu', 'linux_distro_like': 'debian', 'linux_distro_version': '24.04'}
ACTIVE_PROJECT: default
ACTIVE_STACK: newstack
ACTIVE_USER: project
TELEMETRY_STATUS: disabled
ANALYTICS_CLIENT_ID: 5aedad4b-56c0-4531-9cff-5d8454ec3289
ANALYTICS_USER_ID: 4f5702a9-9be8-4de2-8dd1-bbfd145ca2cd
ANALYTICS_SERVER_ID: dd283418-66ec-4794-9057-69382388741c
INTEGRATIONS: ['bitbucket', 'pigeon', 'kaniko', 'airflow', 's3', 'kubernetes']
PACKAGES: {'annotated-types': '0.7.0', 'alembic': '1.15.2', 'jsonref': '1.1.0', 'pydantic': '2.11.1', 'configparser': '7.2.0', 'SQLAlchemy-Utils': '0.42.0', 'jmespath': '1.0.1', 'zenml': '0.91.1', 'pyasn1': '0.6.1', 'multidict': '6.6.4',
'google-auth': '2.41.1', 'oauthlib': '3.3.1', 'wrapt': '1.17.3', 'pexpect': '4.9.0', 'stack-data': '0.6.3', 'aiohttp': '3.12.15', 'jedi': '0.19.2', 'gitdb': '4.0.12', 'boto3': '1.40.18', 'wcwidth': '0.2.14', 'aiohappyeyeballs': '2.6.1', 
[...]

PR #4188, opened 2025-11-13
Fix UnboundLocalError in SSL certificate validation

htahir1:
Tries to fix: https://github.com/zenml-io/zenml/issues/4185

The validate_verify_ssl method in RestZenStoreConfiguration had a bug where cert_content was only defined when verify_ssl pointed to an existing file, but was used unconditionally when writing to ca_bundle.pem. This caused an UnboundLocalError when verify_ssl was a string containing certificate content rather than a file path.

Now the method handles both cases: if verify_ssl is a file path, it reads the content from the file; otherwise it treats verify_ssl as the certificate content itself. This allows users to pass either file paths or inline certificate content.

Fixes the SSL certificate validation error that occurred during pipeline execution when using custom SSL certificates with artifact stores.

## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/rest_zen_store.py
+++ b/src/zenml/zen_stores/rest_zen_store.py
@@ -386,6 +386,8 @@ class RestZenStoreConfiguration(StoreConfiguration):
         if os.path.isfile(verify_ssl):
             with open(verify_ssl, "r") as f:
                 cert_content = f.read()
+        else:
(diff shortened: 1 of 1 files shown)


htahir1:
@schustmi I'm not sure if this fixes it, but it was a try. You can close it if it doesnt make sense

PR #4205, opened 2025-11-18
Add option to build docker images using subprocess calls

schustmi:
## Describe changes
The Docker python library is not compatible with Docker BuildKit, which means some features are not available when using it. This PR introduces an option on the local image builder that allows users to build images using a `docker build` subprocess call.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/image-builders/local.md
+++ b/docs/book/component-guide/image-builders/local.md
@@ -43,6 +43,26 @@ zenml image-builder register <NAME> --flavor=local
 zenml stack register <STACK_NAME> -i <NAME> ... --set
 ```
 
+#### Use the Docker CLI for builds (subprocess mode)
+
+By default, the local image builder uses the Docker Python SDK to run builds. You can alternatively configure it to call the Docker CLI (`docker build`) via a subprocess. This can help if you require some options that are only available for Docker BuildKit, which is not supported in the Python SDK yet.
+
+Enable subprocess mode when registering or updating the component:
+
+```shell
+# Register with subprocess mode enabled
+zenml image-builder register <NAME> --flavor=local --use_subprocess_call=true
+
+# Or update an existing local image builder
+zenml image-builder update <NAME> --use_subprocess_call=true
+```
+
+In subprocess mode, the build options passed via `DockerSettings` are handled in the following way:
+- The `buildargs`, `labels` dictionaries are passed as `--build-arg KEY=VALUE` and `--label KEY=VALUE` respectively
+- Lists will pass multiple command line arguments with the same name for all values
+- `True` boolean values means the argument will be passed without any value (`--KEY`)
+- All other values are converted to a string and passed as `--KEY VALUE`
+
 For more information and a full list of configurable attributes of the local image builder, check out the [SDK Docs](https://sdkdocs.zenml.io/latest/core_code_docs/core-image_builders.html#zenml.image_builders.local_image_builder) .
 
 <figure><img src="https://static.scarf.sh/a.png?x-pxid=f0b4f458-0a54-4fcd-aa95-d5ee424815bc" alt="ZenML Scarf"><figcaption></figcaption></figure>


src/zenml/image_builders/local_image_builder.py
```diff
@@ -119,14 +119,25 @@ def build(
...
+            if docker_build_options.get("pull", False):
+                command.append("--pull")
+
+            if "platform" in docker_build_options:
+                command.append("--platform")
+                command.append(docker_build_options["platform"])
+
+            build_args = docker_build_options.get("build_args", {})
+            for key, value in build_args.items():
+                command.append(f"--build-arg {key}={value}")
```
avishniakov:
```suggestion
            build_args = docker_build_options.get("buildargs", {})
            for key, value in build_args.items():
                command.append("--build-arg")
                command.append(f"{key}={value}")
```

src/zenml/image_builders/local_image_builder.py
```diff
@@ -128,10 +171,63 @@ def build(
...
+                if isinstance(value, list):
+                    for val in value:
+                        command.extend([f"--{key}", str(val)])
+                elif isinstance(value, bool):
+                    if value:
+                        command.append(f"--{key}")
+                else:
+                    command.extend([f"--{key}", str(value)])
+
+            process = subprocess.Popen(command, stdin=f)
```
stefannica, line 223: just a thought, but passing the file object here might not work if you don't seek it back to the first position

stefannica (changes requested):

src/zenml/image_builders/local_image_builder.py
schustmi: The `write_archive(...)` method already seeks to the beginning after writing the archive

stefannica (approved):

src/zenml/image_builders/local_image_builder.py
```diff
@@ -128,10 +171,63 @@ def build(
...
+
+            labels = docker_build_options.pop("labels", None) or {}
+            for key, value in labels.items():
+                command.extend(["--label", f"{key}={value}"])
+
+            cache_from = docker_build_options.pop("cache_from", None) or []
+            for value in cache_from:
+                command.extend(["--cache-from", value])
+
+            for key, value in docker_build_options.items():
```
avishniakov, line 213: I feel it a bit counter-intuitive to break the "standard" notation in such a way: maybe the `docker_build_options` shall be a `list[dict[str,Any]|str]` instead? So users can pass flags as string and some things like `label` as a `[{'label':'a'},{'label':'b'}]`?
Otherwise it is a lot of hidden/insider knowledge and the learning curve will be steeper.

schustmi: @avishniakov The `docker_build_options` also need to work for other image builders though, and the suggested notation would be very weird syntax when using the python Docker SDK, don't you think?

I'm not sure what the best cause of action here would be, ideally those build options should be image-builder independent all all image builder implementations convert them to their internal syntax.

avishniakov: Totally with you - better it be something unified for all builders, but now it is already quite specific to the type of builder, isn't it?
Please don't take my comment as a blocker to merge, I'm just suggesting in this case to push those "special" options up to a proper params of a builder configuration, so it is easy to follow, rather keeping in memory that one shall pass 'labels', even if it doesn't exist in the naive CLI list of arguments.
Generally, I don't get how these custom options help in unification right now even, since options from local builder would make a few or no sense for GCP one, or vice versa, how I perceive this.

schustmi: Agreed that not all builders can be completely unified, but at least for the local one we can keep them consistent right? Additionally, you could also use the "label" key and pass a list of strings (which to me would be the most intuitive way at to pass them?), which would be covered by the remaining code below those special handlings. The list of dicts somehow feels unintuitive to me, but if we find some other better solution I'm happy to adjust this.

avishniakov: Exactly, I'm advocating for the same - kick all "special" parameters out of the build options to dedicated parameters or introduce new layer of complexity, e.g. special LocalDockerSettngs built on top of generic DockerSetting, but it might be not that intuitive and beneficial for the learning curve, IMO.

schustmi: What do you mean by dedicated parameters? Maybe a pydantic class like this, which has some attributes but also allows extras, which will then be handled by the image builders?

```python
class DockerBuildOptions(BaseModel):
  labels: Dict[str, str] = {}
  build_args: Dict[str, str] = {}

  model_config = ConfigDict(extra="allow")
```

avishniakov: Yeah, that would be awesome! Just include all other "special", like:
```python
class DockerBuildOptions(BaseModel):
  pull: bool = False
  labels: Dict[str, str] = {}
  build_args: Dict[str, str] = {}
  cache_from: List[str] = []

  model_config = ConfigDict(extra="allow")
```
WDYT?

schustmi: I like it :)

@stefannica Any opinions?

schustmi: I will implement this in a follow up PR as it affects other image builders as well

PR #4216, opened 2025-11-19
Feature:4151 Heartbeat activation logic

Json-Andriopoulos:
## Describe changes

- Moved heartbeat worker control to step runner (from step launcher).
- Compilation layer changes to calculate heartbeat enabling flag

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/compiler.py
+++ b/src/zenml/config/compiler.py
@@ -38,6 +38,7 @@ from zenml.config.step_configurations import (
     StepConfigurationUpdate,
     StepSpec,
 )
+from zenml.enums import StepRuntime
 from zenml.environment import get_run_environment_dict
 from zenml.exceptions import StackValidationError
 from zenml.models import PipelineSnapshotBase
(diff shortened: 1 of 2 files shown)


src/zenml/config/compiler.py
```diff
@@ -197,7 +197,7 @@ def compile_spec(self, pipeline: "Pipeline") -> PipelineSpec:
         pipeline = copy.deepcopy(pipeline)
 
         invocations = [
-            self._get_step_spec(invocation=invocation)
```
schustmi, line 200: Why do we always disable the heartbeat in this case?

Json-Andriopoulos: Good question. I am unsure where that method is used, I only found usages of it in the tests. But, to properly determine the heartbeat flag we need a bit of context the method lacks (e.g. the stack).

schustmi: Ah in that case, we might just throw the whole method away as well. I think I probably recently removed the last usage of it which is why it's now unused.

schustmi (approved):

Issue #4210, opened 2025-11-18
During Login allow user to also set project not just workspace


AlexejPenner:
### Feature Description

When I login from CLI i get taken to the browser where I select org and workspace - but it feels strange to then have to select project with a separate command

### Priority

Low - Nice to have

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

adam6878:
Hi @AlexejPenner , I would like to work on this issue, I saw that zenml does set a project by default after you log in, so is the change needed in the UI side of things? were when you select the workspace you also want to select the project in the zeml pro page?

htahir1:
@adam6878 We would be more than happy for support! Go ahead and start a PR! Make sure to read [CONTRIBUTING ](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md) guidelines before starting :-)

adam6878:
Thank you @htahir1 , I was thinking the solution should be that right after the login attempt, the cli will continue by asking the user to provide the name for the active project they want and if the user does not provide any or provides a project that doesn't exists, zenml just sets a default project as the active one.

htahir1:
@adam6878 I think the `zenml login` standalone behavior is quite fine right now. But maybe the way to do it right now is `zenml login <WORKSPACE_NAME> --project <PROJECT_ID>` and then we allow to set it afterwards. WDYT?

htahir1:
Btw check this out https://github.com/zenml-io/zenml/issues/4221#issue-3642117148 also related?

adam6878:
[> zenml login <WORKSPACE_NAME> --project <PROJECT_ID>](https://github.com/zenml-io/zenml/issues/4210#issuecomment-3553048677) Sounds like a good idea! I will start design and implement a solution according to this, and yes https://github.com/zenml-io/zenml/issues/4221#issue-3642117148  is related but i thought first that the feature is about changing the UI in the zenml PRO web page to include the option for selecting project.

htahir1:
@adam6878 youre right having a project dropdown on the UI will be a good idea too but I’m afraid this will be hard honestly.. the problem is that the pro ui repo is closed source and projects are a closed source feature really but the oss cli gives ability to interact with closed source elements like projects

adam6878:
@htahir1 OK got it, I will Open a PR soon for this Issue!

PR #4224, opened 2025-11-19
Added the option for setting project when logging in

adam6878:
## Describe changes
I implemented a --project flag to zenml login so users can immediately switch to a specific project when connecting to any server (local, REST, SQL, or Pro). The command now routes the requested project through the various login paths and sets it only after the connection succeeds, ensuring the active project matches the context the user intends right after authentication.
This PR will close [#4210](https://github.com/zenml-io/zenml/issues/4210)

## Pre-requisites

## Types of changes

## Summary by CodeRabbit

* **New Features**
  * Added `--project` flag to the login command, allowing users to specify and automatically set an active project during the authentication process.
  * Project is now set as active immediately upon successful connection, with user-friendly error handling for invalid or missing projects.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -75,6 +75,7 @@ jobs:
           remove-android: 'true'
           remove-haskell: 'true'
           build-mount-path: /var/lib/docker/
+        if: inputs.os == 'ubuntu-latest'
       - name: Checkout code
         uses: actions/checkout@v4.2.2
         with:
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -86,6 +86,7 @@ jobs:
           remove-android: 'true'
           remove-haskell: 'true'
           build-mount-path: /var/lib/docker/
+        if: inputs.os == 'ubuntu-latest'
       - name: Checkout code
         uses: actions/checkout@v4.2.2
         with:
--- a/docs/book/how-to/steps-pipelines/dynamic_pipelines.md
+++ b/docs/book/how-to/steps-pipelines/dynamic_pipelines.md
@@ -96,6 +96,90 @@ Use `runtime="inline"` when you need:
 - Shared resources with the orchestrator
 - Sequential execution
 
+### Map/Reduce over collections
+
+Dynamic pipelines support a high-level map/reduce pattern over sequence-like step outputs. This lets you fan out a step across items of a collection and then reduce the results without manually writing loops or loading data in the orchestration environment.
+
+```python
+from zenml import pipeline, step
+
+@step
+def producer() -> list[int]:
+    return [1, 2, 3]
+
+@step
+def worker(value: int) -> int:
+    return value * 2
+
+@step
+def reducer(values: list[int]) -> int:
+    return sum(values)
+
+@pipeline(dynamic=True, enable_cache=False)
+def map_reduce():
+    values = producer()
+    results = worker.map(values)   # fan out over collection
+    reducer(results)               # pass list of artifacts directly
+```
+
+Key points:
+- `step.map(...)` fans out a step over sequence-like inputs.
+- Steps can accept lists of artifacts directly as inputs (useful for reducers).
+- You can pass the mapped output directly to a downstream step without loading in the orchestration environment.
+
+#### Mapping semantics: map vs product
+
+- `step.map(...)`: If multiple sequence-like inputs are provided, all must have the same length `n`. ZenML creates `n` mapped steps where the i-th step receives the i-th element from each input.
+- `step.product(...)`: Creates a mapped step for each combination of elements across all input sequences (cartesian product).
+
+Example (cartesian product):
+
+```python
+from zenml import pipeline, step
+
+@step
+def int_values() -> list[int]:
+    return [1, 2]
+
+@step
+def str_values() -> list[str]:
+    return ["a", "b", "c"]
+
+@step
+def do_something(a: int, b: str) -> int:
+    ...
+
+@pipeline(dynamic=True)
+def cartesian_example():
+    a = int_values()
+    b = str_values()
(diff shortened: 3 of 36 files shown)


htahir1:
@claude /full-review

htahir1:
@coderabbitai review

htahir1:
@codex full-review

src/zenml/cli/login.py
```diff
@@ -581,6 +597,29 @@ def _fail_if_authentication_environment_variables_set() -> None:
         )
 
 
+def _set_active_project(project_name_or_id: str) -> None:
```
htahir1, line 595: **1. Exception Handling**
Does `set_active_project` raise a `KeyError` when the project isn't found? I recall ZenML usually raising `zenml.exceptions.EntityNotFoundError` for missing resources. We might want to catch that (or both) to ensure we get the nice error message instead of a traceback.

**2. Null Safety**
Just a minor nit, but if `project` comes back as `None` (depending on the API version), `project.name` will crash. Might be worth adding a quick `if project:` check before logging the success message.

Here is a suggestion that covers these cases:

```python
def _set_active_project(project_name_or_id: str) -> None:
    from zenml.exceptions import EntityNotFoundError
    
    # Ensure we are using the fresh store config
    client = Client()

    try:
        project = client.set_active_project(project_name_or_id)
    except (KeyError, EntityNotFoundError):
        cli_utils.error(
            f"No project named or with ID '{project_name_or_id}' exists on "
            "the connected ZenML server."
        )
    except IllegalOperationError as exc:
        cli_utils.error(
            f"Failed to set the active project to '{project_name_or_id}': {exc}"
        )
    else:
        if project:
            cli_utils.declare(f"✔ The active project has been set to {project.name}")
```

htahir1 (changes requested):
Nice work\! This is a super handy QoL improvement—saving that extra step to switch projects after logging in is going to be very popular.

The logic for passing the argument down looks correct, but I have a few concerns about the `_set_active_project` helper function regarding how the `Client` singleton behaves

htahir1:
Also Not strictly caused by your PR, but while we’re here:

In the SQL branch of connect_to_server, you still unconditionally print:

`cli_utils.declare(f"Connected to SQL database '{url}'")`

even if set_store fails and you hit IllegalOperationError or CredentialsNotValid. This predates your changes, but if you want to be precise, you might want to move that line into the else: block so it only prints on success.

`--local --blocking --project X` is a bit odd: in blocking mode, `start_local_server` is supposed to block the CLI, so the `_set_active_project(project)` call will only run after the server stops. be aware that setting the project in blocking mode is not very meaningful.

src/zenml/cli/login.py
adam6878: Great suggestions will be changed!

adam6878: should we log something if the project is in fact None?

src/zenml/cli/login.py
```diff
@@ -581,6 +597,28 @@ def _fail_if_authentication_environment_variables_set() -> None:
...
+    from zenml.exceptions import EntityNotFoundError
```
htahir1: no inline imports please!

src/zenml/cli/login.py
htahir1: no id just raise an exception

adam6878: So if project = None raise an exception?

adam6878:
@htahir1 any comments?

htahir1:
@claude full-review

htahir1:
I think at this point id love to hear @stefannica's thoughts on this or @schustmi 's

src/zenml/cli/login.py
```diff
@@ -37,6 +38,7 @@
     AuthorizationException,
     CredentialsNotValid,
     IllegalOperationError,
+    EntityNotFoundError,
```
schustmi: This class doesn't exist

src/zenml/cli/login.py
```diff
@@ -581,6 +598,28 @@ def _fail_if_authentication_environment_variables_set() -> None:
...
+    client = Client()
+
+    try:
+        project = client.set_active_project(project_name_or_id)
+    except (KeyError, EntityNotFoundError):
+        cli_utils.error(
+            f"No project named or with ID '{project_name_or_id}' exists on "
+            "the connected ZenML server."
+        )
+    except IllegalOperationError as exc:
```
schustmi: This should catch all other exceptions here, not just `IllegalOperationError`

src/zenml/cli/login.py
```diff
@@ -296,8 +303,11 @@ def connect_to_server(
             )
         except CredentialsNotValid as e:
             cli_utils.warning(f"Authorization error: {e}")
+        else:
+            if project:
+                _set_active_project(project)
 
-        cli_utils.declare(f"Connected to SQL database '{url}'")
+            cli_utils.declare(f"Connected to SQL database '{url}'")
```
schustmi: I think it makes more sense for this log to happen before setting the active project? We first connect to the DB/server, and then try to set the active project. Even if setting the active project fails, we still successfully connected to the DB/server and should let the user know about that.

src/zenml/cli/login.py
adam6878: so ``expect Exception as e`` here?

schustmi: Yes!

adam6878:
Fixed @schustmi

schustmi:
@adam6878 Seems like your test is still failing, otherwise looks good to me! Thanks for the PR!

htahir1 (approved):
Thanks @adam6878 ! Ill let the linting run through and if it works then merge!

PR #4231, opened 2025-11-20
Add pipeline deployment to onboarding

htahir1:
## Describe changes
Add pipeline deployment to onboarding events. Fixes https://github.com/zenml-io/zenml-dashboard/issues/913

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -5299,8 +5299,13 @@ class SqlZenStore(BaseZenStore):
 
             session.refresh(deployment_schema)
 
+            # Track deployment and mark onboarding as completed
             self._update_onboarding_state(
-                completed_steps={OnboardingStep.SNAPSHOT_DEPLOYED},
+                completed_steps={
+                    OnboardingStep.PIPELINE_DEPLOYED,
+                    OnboardingStep.OSS_ONBOARDING_COMPLETED,
+                    OnboardingStep.PRO_ONBOARDING_COMPLETED,
+                },
                 session=session,
             )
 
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5061,6 +5061,14 @@ def create_snapshot(
             )
             session.refresh(new_snapshot)
 
+            # Only track onboarding for user-created named snapshots, not internal
+            # system-generated snapshots created during pipeline runs
```
htahir1:
```suggestion
```

htahir1:
```suggestion
```

Cahllagerfeld (commented):
This is something I cannot answer myself, but just as a thing we need to consider. When finishing the flow, we're adding the `oss_onboarding_completed` or `pro_onboarding_completed` for the oss and pro onboarding. If we have the new step now, we may need to also adjust when these `completed` items are set

Edit:
If we only have one new collapsible for `Deploy your pipeline`, we probably dont need to track the `snapshot_created` one, as its happening inside the same collapsible 🤔 We would only need both, if we explicitly want the user to create the snapshot.

htahir1:
@Cahllagerfeld  Given that we might remove snapshots and subsume into deployments, i would agree and remove the snapshot event". Then we can ask the user to directly deploy the pipeline:

`zenml pipeline deploy run.my_pipeline`

WDYT?

schustmi:
I don't think snapshots will gone from ZenML as a concept in the next few years. I think that's a confusion from some discussion that we had on "using deployments to run a snapshot" instead of some internal ZenML pro concept. There will still be a need to have "something" (= a snapshot) that you want to run. Or that you want to deploy.

Not saying I disagree with removing them from the onboarding, but they won't be gone anytime soon, even if we change the way they can be executed from the server.

htahir1:
@schustmi I think I mean something different = I think just we dont need to expose to the user right now in onboarding. that way also pro and oss onboarding are unified

htahir1:
@claude full-review

PR #4241, opened 2025-11-21
Improve CLI Tables and IDs handling

safoinme:
## Describe changes
This PR refactors the ZenML CLI table rendering system to:
- Enable pipeable output (JSON/YAML/CSV/TSV) by separating stdout/stderr
- Consolidate table rendering logic into centralized functions
- Improve table aesthetics with intelligent column sizing
- Simplify CLI command implementations across artifact, model, tag, and deployment commands

Key technical changes:
- New `zenml_cli` module for stdout/stderr routing
- New rendering functions: `render_list_output()`, `handle_output()`, `prepare_output()`
- Enhanced `list_options` decorator with auto-rendering support
- Column width management with protected widths for IDs (36), URLs (40), status (18)

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
+++ b/src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
@@ -180,12 +180,3 @@ class WandbExperimentTracker(BaseExperimentTracker):
                     "Weave enabled but no project_name specified. "
                     "Skipping weave initialization."
                 )
-        else:
-            import weave
-
-            if self.config.project_name:
-                logger.info("Disabling weave")
-                weave.init(
-                    project_name=self.config.project_name,
-                    settings={"disabled": True},
-                )


safoinme:
@claude full-review

htahir1:
@safoinme can you please paste some screenshots or a video clip of new usage?

safoinme:
> @safoinme can you please paste some screenshots or a video clip of new usage?

image


Still trying to figure out what's the best way to handle this, what do you think of this?

htahir1:
@safoinme fix docstrings please!

src/zenml/cli/deployment.py
```diff
@@ -85,13 +100,18 @@ def list_deployments(**kwargs: Any) -> None:
...
-            return
+        return
+
+    if not deployments.items:
+        cli_utils.declare("No deployments found for this filter.")
+        return
 
-        cli_utils.print_deployment_table(deployments=deployments.items)
-        cli_utils.print_page_info(deployments)
+    items = cli_utils.format_page_items(
```
Json-Andriopoulos: Seems like this is pretty standardized now right? Maybe we can replace those 4-5 lines with a single call and compress our code a bit>

src/zenml_cli/__init__.py
```diff
@@ -0,0 +1,77 @@
+#  Copyright (c) ZenML GmbH 2025. All Rights Reserved.
```
schustmi, line 1: This is currently not called when a user runs the `zenml` command in the shell. For that, you need to update the scripts section of the `pyproject.toml`.

src/zenml/cli/utils.py
```diff
@@ -3070,3 +2866,671 @@ def requires_mac_env_var_warning() -> bool:
...
+    """
+    if isinstance(value, list):
+        if not value:
+            return value
+        if all(isinstance(item, dict) and "name" in item for item in value):
+            return [item["name"] for item in value]
+        return [_simplify_value(item) for item in value]
+
+    if isinstance(value, dict):
+        if "name" in value and "id" in value:
```
schustmi: Why do we need the check for `id` here if we don't use it?

docs/book/reference/global-settings.md
```diff
@@ -47,6 +47,10 @@ Using the default local database.
 ┗━━━━━━━━┷━━━━━━━━━━━━┷━━━━━━━━┷━━━━━━━━━┷━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━┛
 ```
```
bcdurak, line 48: As we are changing the way the tables look, it might make sense to go over the docs and clean up some outputs like the one above. It's from 0.13.2.

src/zenml/cli/utils.py
safoinme: The intention is to check zenml response model and block hydration

docs/book/reference/global-settings.md
safoinme: this needs to be a different PR we literally have like 100s of tables

docs/book/user-guide/best-practices/quick-wins.md
```diff
@@ -854,3 +855,61 @@ Using the zenmldocs MCP server, show me how to register an MLflow experiment tra
...
+## 15 Export CLI data in multiple formats
+
+**Why** -- pipe ZenML CLI output into scripts, CI/CD, or analysis tools with machine-readable formats. All `zenml list` commands support flexible output formatting, making it easy to automate workflows, analyze data in spreadsheets, or integrate with external tools.
+
+```bash
+# Get stack data as JSON for processing with jq
+zenml stack list --output=json | jq '.items[] | select(.name=="production")'
+
+# Export pipeline runs to CSV for analysis
+zenml pipeline runs list --output=csv > pipeline_runs.csv
```
bcdurak, line 868: The piping does not work properly right now. When I set the `ZENML_LOGGING_VERBOSITY` to `DEBUG`, some of the debug messages still leak into the corresponding `csv`, `json`, etc.

src/zenml/cli/stack.py
```diff
@@ -1042,27 +1042,67 @@ def rename_stack(
 
 
 @stack.command("list")
-@list_options(StackFilter)
+@list_options(
+    StackFilter,
+    default_columns=[
```
Json-Andriopoulos, line 1048: I think here we are adding too many columns. The *default* behaviour the user is initially seeing will be *worse* than the previous version:

Screenshot 2025-11-25 at 11 25 36 AM

Json-Andriopoulos: Minor: Also not sure how easy it is to fix but if you see in the screenshot - when I select very few columns (2 or 3) rendering is a bit weird.

safoinme: hmm it suppose to be adding that much columns for stack only when terminal window width allows that, but surely we can limit just to the the most essential 6

znegrin:
@safoinme When I run `zenml pipeline list`, I get this warning, which I don't fully understand. 

`Unknown column(s) ignored: num_runs. Available: created, id, latest_run_id, latest_run_status, latest_run_user, name, project_id, tags, updated, user, user_id, visualizations`

It seems like I did something wrong, but I just ran a regular command. I'm unsure if this is an error or if it's intentional to display it, but if you run the default command, it shouldn't be shown as a warning. (If we want to inform users that they can add columns, I would style it more subtly.)

src/zenml/cli/utils.py
```diff
@@ -112,21 +116,39 @@
...
+
+
+ENV_ZENML_DEFAULT_OUTPUT = "ZENML_DEFAULT_OUTPUT"
+ENV_ZENML_CLI_COLUMN_WIDTH = "ZENML_CLI_COLUMN_WIDTH"
 MAX_ARGUMENT_VALUE_SIZE = 10240
+ALLOWED_OUTPUT_FORMATS: Tuple[str, ...] = (
+    "table",
+    "json",
+    "yaml",
+    "csv",
```
Json-Andriopoulos: Look I am not sure why we need those. `tsv` and `csv` are just like the `table` format but will always be much worse at rendering as they have no column boundaries & alignment.

As alternative rendering option json & yaml make  total sense but I would get rid of other table alternatives. See how bad it looks:

Screenshot 2025-11-25 at 11 55 25 AM
Screenshot 2025-11-25 at 11 55 39 AM

src/zenml/cli/stack.py
```diff
@@ -1042,27 +1042,67 @@ def rename_stack(
...
+        "annotator",
+        "image_builder",
+        "feature_store",
+        "model_deployer",
+        "model_registry",
+    ],
+)
 @click.pass_context
-def list_stacks(ctx: click.Context, /, **kwargs: Any) -> None:
+def list_stacks(
```
bcdurak, line 1059: Image

Some columns are missing in the help message and when I give a wrong column, the error message shows the columns in the wrong format.

znegrin:
@safoinme similar to the other comment, if I reduce the size of the window, I receive another warning that I feel should be styled a bit more neutral. I think we can also remove the highlighting of `json` as it probably doesn't make sense in this context.

Current: 
image

Proposal:
image

The same command is currently showing a big space below the warnings, I guess that just one backspace is enough:
image

docs/book/user-guide/best-practices/quick-wins.md
safoinme: with @schustmi comment, now it's fixed and should solve this

src/zenml/cli/stack.py
```diff
@@ -1042,27 +1042,67 @@ def rename_stack(
...
+    ctx: click.Context,
```
bcdurak, line 1060: Image

Additionally, every time I do `zenml stack list`. I see this warning message. Since this is one of the most widely used CLI commands, I would propose finding a way to remove this.

Perhaps, as a solution, we can change the default columns here and make them act a bit dynamic. Things like alerters and annotators are really not that common. We can maybe show them only when we have one. WDYT?

znegrin:
If I do `zenml artifact version list` it shows weird spacing between rows:
image

I think it is probably related to this column:
image

Same with model versions:
image

src/zenml/cli/stack.py
```diff
@@ -1042,27 +1042,67 @@ def rename_stack(
...
+    columns: str,
+    output_format: OutputFormat,
+    **kwargs: Any,
+) -> None:
     """List all stacks that fulfill the filter requirements.
 
     Args:
-        ctx: the Click context
-        kwargs: Keyword arguments to filter the stacks.
+        ctx: The Click context.
```
bcdurak, line 1069: Image

I think this was broken even before but I just realized that the first paragraph in the help message is misformatted. Would be nice to fix that as well.

src/zenml/cli/stack_components.py
```diff
@@ -152,32 +152,70 @@ def generate_stack_component_list_command(
         A function that can be used as a `click` command.
     """
 
-    @list_options(ComponentFilter)
+    @list_options(
+        ComponentFilter,
+        default_columns=["active", "name", "component_id", "flavor", "owner"],
```
bcdurak: Can we use id instead of component id here?

src/zenml/cli/pipeline.py
```diff
@@ -574,25 +574,29 @@ def create_run_template(
 
 
 @pipeline.command("list", help="List all registered pipelines.")
-@list_options(PipelineFilter)
-def list_pipelines(**kwargs: Any) -> None:
+@list_options(PipelineFilter, default_columns=["id", "name", "num_runs"])
```
bcdurak: Image

`num_runs` does not exist for pipelines yet.

src/zenml/cli/stack.py
znegrin: I agree, I think we can show by default the most relevant (orchestrator, artifact store, and maybe one more -deployer?-). They can now edit the columns anyway, so I think it is fine with fewer columns, especially if they are all empty like in the screenshot @Json-Andriopoulos shared.

src/zenml/cli/pipeline.py
bcdurak: I would also propose adding "latest_run_status", "latest_run_id" and "created" columns to the default ones as well.

src/zenml/cli/pipeline.py
```diff
@@ -639,24 +643,26 @@ def schedule() -> None:
 
 
 @schedule.command("list", help="List all pipeline schedules.")
-@list_options(ScheduleFilter)
-def list_schedules(**kwargs: Any) -> None:
+@list_options(ScheduleFilter, default_columns=["id", "name"])
```
bcdurak: Same here. Maybe we can add `cron_expression` here?

src/zenml/cli/stack.py
bcdurak: I am seeing the same error on `zenml artifact list` as well.

docs/book/reference/global-settings.md
```diff
@@ -47,6 +47,10 @@ Using the default local database.
...
 
+{% hint style="info" %}
+The output format can be customized with `--output` (json, yaml, csv, tsv) and `--columns` flags. See [environment variables](environment-variables.md#cli-output-formatting) for details.
```
Json-Andriopoulos: Having an `all` option would be great:

`zenml stack list --columns all`

src/zenml/cli/artifact.py
```diff
@@ -115,25 +121,31 @@ def version() -> None:
     """Commands for interacting with artifact versions."""
 
 
-@cli_utils.list_options(ArtifactVersionFilter)
 @version.command("list", help="List all artifact versions.")
-def list_artifact_versions(**kwargs: Any) -> None:
+@list_options(ArtifactVersionFilter)
```
bcdurak: Since we don't have any defaults here, artifact version listing is a bit weird.

Image

in the first one its even impossible to see what the error is.

Image

The second one shows that the shaping gets messed up due to columns like `materializer`. In most cases, I think we have to deal with these columns in a specific way. For instance for data types and materializers, we can just show their name.

safoinme:
@znegrin all the issues you pointed have been adressed, the issue behind this is that we needed to specify which default column to pass. the extra spacing is also adressed

znegrin:
After checking a few outputs, I feel we could add some spacing, maybe always having one space before the list header, and showing the pagination between two spaces to make it cleaner. Also having a bit of padding to the left of the list so it is not so tight. Example:

Existing:
image

Proposal:
image

src/zenml/cli/authorized_device.py
```diff
@@ -59,24 +59,39 @@ def describe_authorized_device(id_or_prefix: str) -> None:
...
-def list_authorized_devices(**kwargs: Any) -> None:
+@list_options(
+    OAuthDeviceFilter,
+    default_columns=[
+        "id",
+        "status",
+        "expires",
+        "hostname",
+        "os",
+        "zenml_version",
```
bcdurak: `zenml_version` and `python_version` do not exist as columns here.

docs/book/reference/environment-variables.md
```diff
@@ -129,6 +129,28 @@ To set the path to the global config file, used by ZenML to manage and store the
...
+
+### Default output format
+
+Set the default output format for all CLI list commands:
+
+```bash
+export ZENML_DEFAULT_OUTPUT=json
+```
+
+Choose from `table` (default), `json`, `yaml`, `csv`, or `tsv`. This applies to commands like `zenml stack list`, `zenml pipeline list`, etc.
```
Json-Andriopoulos, line 142: Not sure how much others agree but I like colors in my terminal. Makes reading large objects pretty easy. How about we do some colorful rendering. See a quick example for json:

Screenshot 2025-11-25 at 12 22 30 PM

The script I quickly patched is:

```
import json

obj = {
  "items": [
    {
        "first_name": "John",
        "last_name": "Smith",
        "email": "john.smith@random.com",
        "country": "US",
        "city": "Springfield",
    },
    {
        "first_name": "Becky",
        "last_name": "Smith",
        "email": "becky.smith@random.com",
        "country": "US",
        "city": "Alabama",
    },
  ],
  "pagination": {
    "index": 1,
    "max_size": 20,
    "total_pages": 1,
    "total": 2
  }
}

formatted_json = json.dumps(obj, sort_keys=True, indent=4)

from pygments import highlight, lexers, formatters
colorful_json = highlight(formatted_json, lexers.JsonLexer(), formatters.TerminalFormatter())
print(colorful_json)

```

src/zenml/cli/utils.py
Json-Andriopoulos: I am also concerned regarding the handling of values that need escaping (e.g. nested objects, quoted text, etc.). I know we added the option for easy loading to data management tools like DBs or spreadsheets but JSON can be used for the same purpose without the inherent issues of csv.

src/zenml/cli/model.py
```diff
@@ -390,25 +408,31 @@ def version() -> None:
     """Interact with model versions in the Model Control Plane."""
 
 
-@cli_utils.list_options(ModelVersionFilter)
 @version.command("list", help="List model versions with filter.")
-def list_model_versions(**kwargs: Any) -> None:
+@list_options(ModelVersionFilter)
+def list_model_versions(
```
bcdurak, line 408: This one suffers from the same problem artifact versions are suffering from:

Image

bcdurak: btw, since the length is also limited by default, even if I have a larger window, the table gets truncated and the warning is shown.

src/zenml/cli/artifact.py
```diff
@@ -115,25 +118,29 @@ def version() -> None:
...
 
 
-@cli_utils.list_options(ArtifactVersionFilter)
 @version.command("list", help="List all artifact versions.")
-def list_artifact_versions(**kwargs: Any) -> None:
+@list_options(
+    ArtifactVersionFilter,
+    default_columns=["id", "artifact", "version", "type"],
+)
+def list_artifact_versions(
```
Json-Andriopoulos, line 127: Available columns are not present when using the `--help` flag in the CLI:

Screenshot 2025-11-25 at 12 44 06 PM

Same for artifacts. But it appears when I pass irrelevant options:

Screenshot 2025-11-25 at 12 47 19 PM

Works as expected in other entities I tried like stacks.

src/zenml/cli/project.py
```diff
@@ -36,32 +38,51 @@ def project() -> None:
 
 
 @project.command("list")
-@list_options(ProjectFilter)
+@list_options(ProjectFilter, default_columns=["active", "id", "name"])
```
bcdurak: Maybe we can add `description` to the list here?

src/zenml/cli/secret.py
```diff
@@ -165,11 +166,15 @@ def create_secret(
 @secret.command(
     "list", help="List all registered secrets that match the filter criteria."
 )
-@list_options(SecretFilter)
-def list_secrets(**kwargs: Any) -> None:
+@list_options(SecretFilter, default_columns=["id", "name"])
```
bcdurak, line 168: We can add `scope` as a default column here.

bcdurak: btw tables look a bit off when the console is relatively wide but there are only 1-2 columns to show: 

Image

src/zenml/cli/server.py
```diff
@@ -736,14 +736,6 @@ def server_list(
     cli_utils.print_pydantic_models(  # type: ignore[type-var]
         all_servers,
         columns=columns,
-        rename_columns={
```
bcdurak, line 739: Why did we remove those? 

Now I am getting: 

Image

src/zenml/cli/service_connectors.py
```diff
@@ -992,16 +1010,48 @@ def list_service_connectors(
...
-    if not connectors:
+    if not connectors.items:
         cli_utils.declare("No service connectors found for the given filters.")
         return
 
-    cli_utils.print_service_connectors_table(
-        client=client,
-        connectors=connectors.items,
-        show_active=not is_sorted_or_filtered(ctx),
+    show_active = not is_sorted_or_filtered(ctx)
```
bcdurak, line 1015: Just noticed something. When you list items that can be `active`, the column that shows the state with either `●` or `-` does not have a header name. Same goes for other element like stacks or users.

src/zenml/cli/utils.py
safoinme: I think csv is mainly there to be saved into a file instead of terminal view itself

src/zenml/cli/user_management.py
```diff
@@ -77,34 +81,51 @@ def describe_user(user_name_or_id: Optional[str] = None) -> None:
 
 
 @user.command("list")
-@list_options(UserFilter)
+@list_options(UserFilter, default_columns=["active", "id", "name"])
```
bcdurak: The name refers to the email here. I would suggest adding the actual name and the role as well.

src/zenml/models/v2/base/page.py
```diff
@@ -43,6 +43,20 @@ def size(self) -> int:
         """
         return len(self.items)
 
+    @property
```
bcdurak: Why do we need to add this here? 

As far as I can see, all these attributes are already a part of the Page instance.

src/zenml/cli/artifact.py
safoinme: which command exactly?

src/zenml/cli/stack.py
safoinme: sorry what columns are missing exactly? and what's wrong with the format? do you mean the uppercased names?

src/zenml/cli/stack.py
safoinme: the message changed from warning to decalre which is less visible and annoying, also now we only show most important 6 columns rest can be added via column option

src/zenml/cli/stack.py
safoinme: fixed

src/zenml/cli/model.py
safoinme: this is fixed

src/zenml/cli/service_connectors.py
safoinme: should we add word active in top? i mean i thought it wasn't needed and it only added extra space but i can add it back

src/zenml/cli/user_management.py
safoinme: same here role isn't in UserResponse

src/zenml/cli/server.py
safoinme: we decided that column renaming is out-of-scope and would make this more complex PR

docs/book/reference/global-settings.md
safoinme: added

src/zenml/models/v2/base/page.py
safoinme: that was added by you in the previous commits of zuri's PR but that's a good idea we don't need it, it's removed now

docs/book/reference/environment-variables.md
safoinme: Implemnted

safoinme:
@znegrin @Json-Andriopoulos @schustmi @bcdurak all clear comments are adressed, i have left some unresolved because i didn't understand or we have decided in the spec to follow different approach for

src/zenml/cli/stack.py
bcdurak: 1. for instance, the `--help` text doesnt include columns like `created`, `updated`, `user` or `user_id` (which are mentioned in the warning)

2. yes, `artifact_store` would be the correct column input here instead of `ARTIFACT STORE`

src/zenml/cli/service_connectors.py
bcdurak: up to you. when I was testing service connectors, I saw that this column was unnamed and filed with `-`s. only at this point, it might be a bit unclear imo.

src/zenml/cli/server.py
bcdurak: I'll leave this up to you but i think "allowing the users to rename" and "us renaming some columns" are two different things. I do agree that the first one can be left out of scope of the this PR, however, the second one should be relatively easy to implement to solve issues like this.

docs/book/user-guide/best-practices/quick-wins.md
bcdurak: @safoinme not sure if you have tested this but on my end, this is still not resolved.

src/zenml/cli/stack_components.py
```diff
@@ -152,32 +153,64 @@ def generate_stack_component_list_command(
...
+            if not components.items:
                 cli_utils.declare("No components found for the given filters.")
                 return
 
-            cli_utils.print_components_table(
-                client=client,
-                component_type=component_type,
-                components=components.items,
-                show_active=not is_sorted_or_filtered(ctx),
+            show_active = not is_sorted_or_filtered(ctx)
```
bcdurak, line 180: Image

The active should be the first column like the other tables.

src/zenml/cli/artifact.py
Json-Andriopoulos: @safoinme `zenml artifact list` and `zenml artifact version list`

src/zenml/cli/deployment.py
Json-Andriopoulos: @safoinme Why is this resolved? I don't see a change 🤔 . The proposal is introducing a common function like:
```

def render_list_entities(paginated_result, generate_row_fn: Callable, pagination_info):

    if not paginated_result.items:
        cli_utils.declare("No records found for this filter.")
        return
    items = cli_utils.format_page_items(deployments, generate_row_fn, output_format)
    cli_utils.handle_output(
        items, pagination_info, columns, output_format
    )
```

And using that across all duplicate invocations.

safoinme: sorry thought it was already handled, should be done now, by extending print_page itself

Json-Andriopoulos (approved):

pyproject.toml
```diff
@@ -2,6 +2,9 @@
 requires = ["uv_build >= 0.8.17, <0.9.0"]
 build-backend = "uv_build"
 
+[tool.uv.build-backend]
```
schustmi, line 5: @safoinme Can you check if this behaves correctly in the following cases:
- Fresh install in a virtualenv
- Upgrading an existing ZenML installation

safoinme: yes both works

schustmi: Thanks!

safoinme:
@schustmi issue was with added syntax_highlight function, it should be fixed now

docs/book/reference/environment-variables.md
```diff
@@ -129,6 +129,28 @@ To set the path to the global config file, used by ZenML to manage and store the
...
 ```
 
+## CLI output formatting
+
+### Default output format
+
+Set the default output format for all CLI list commands:
+
+```bash
+export ZENML_DEFAULT_OUTPUT=json
```
bcdurak, line 139: Just a very small inconsistency that I found while testing this: 

Image

The columns names in the regular list are all upper case, while the json outputs have half-upper half-lower case. Should be easy to fix. If it is not, feel free to ignore for now.

docs/book/reference/global-settings.md
bcdurak:
```suggestion
The output can be customized with an `--output` (json, yaml, csv, tsv, table) option and a `--columns` selection. See [environment variables](environment-variables.md#cli-output-formatting) for more details.
```

docs/book/user-guide/best-practices/quick-wins.md
```diff
@@ -854,3 +855,61 @@ Using the zenmldocs MCP server, show me how to register an MLflow experiment tra
 - For bleeding-edge features on develop, consult the repo or develop docs directly
 
 Learn more: [Access ZenML documentation via llms.txt and MCP](https://docs.zenml.io/reference/llms-txt)
+
+## 15 Export CLI data in multiple formats
```
bcdurak, line 859: [Our CLI docs](https://sdkdocs.zenml.io/0.91.1/cli.html) is based on the `src/zenml/cli/__init__.py` file. Would be nice to have the same section over there as well.

docs/book/user-guide/best-practices/quick-wins.md
```diff
@@ -854,3 +855,61 @@ Using the zenmldocs MCP server, show me how to register an MLflow experiment tra
...
+
+**Why** -- pipe ZenML CLI output into scripts, CI/CD, or analysis tools with machine-readable formats. All `zenml list` commands support flexible output formatting, making it easy to automate workflows, analyze data in spreadsheets, or integrate with external tools.
```
bcdurak: I don't understand the first sentence here.

src/zenml/cli/utils.py
```diff
@@ -2415,112 +2047,313 @@ def get_deployment_status_emoji(
...
+        stack: The stack response.
+        output_format: The output format.
+        active_id: ID of the active stack for highlighting.
+
+    Returns:
+        Dict with stack data for display.
+    """
+    is_active = active_id is not None and stack.id == active_id
+
+    row: Dict[str, Any] = {
```
bcdurak, line 2168: When I use something like the json output format, I would expect the `active` field to be `true` or `false`. Right now, it is a string which says `"Yes"` or an empty string which is harder to interpret IMO.

bcdurak: This might be true for the other entities that can be "active" as well.

src/zenml/cli/model.py
```diff
@@ -77,23 +79,34 @@ def model() -> None:
...
-@cli_utils.list_options(ModelFilter)
 @model.command("list", help="List models with filter.")
-def list_models(**kwargs: Any) -> None:
+@list_options(
+    ModelFilter,
+    default_columns=[
+        "id",
+        "name",
+        "latest_version_name",
+        "description",
```
bcdurak: `description` and `use_cases` do not exist.

src/zenml/cli/pipeline.py
```diff
@@ -882,11 +898,15 @@ def builds() -> None:
 
 
 @builds.command("list", help="List all pipeline builds.")
-@list_options(PipelineBuildFilter)
-def list_pipeline_builds(**kwargs: Any) -> None:
+@list_options(PipelineBuildFilter, default_columns=["id", "name"])
```
bcdurak: Pipeline build do not have `name`s.

bcdurak: I would show other things than the `id` here as well.

src/zenml/cli/project.py
```diff
@@ -36,32 +37,50 @@ def project() -> None:
 
 
 @project.command("list")
-@list_options(ProjectFilter)
+@list_options(
+    ProjectFilter, default_columns=["active", "id", "name", "description"]
```
bcdurak, line 41: `Unknown column(s) ignored: description. Available: active, created, display_name, id, name, updated`. Maybe you are not listing them in an hydrated manner?

safoinme: i mean this is a list command, to add hydrated command we need to create custom generate row function

docs/book/user-guide/best-practices/quick-wins.md
safoinme: fixed

src/zenml/cli/project.py
bcdurak: not really. If you want to add the description to the list, you need to add `hydrate=True` to line 62 on this file as I initially guessed.

this is a pattern that you were already using btw. for builds and snapshots, the `hydrate=True` is there (which I would even suggest to remove because the values that you get through this flag are ignored by default).

bcdurak: @safoinme

src/zenml/cli/pipeline.py
bcdurak: again, listing builds just lists some ID's now.

bcdurak: Screenshot 2025-12-01 at 16 10 20

This is how it used to look like before for the sake of comparison.

safoinme: updated, thank you

bcdurak (approved):

PR #4246, opened 2025-11-25
Add support for dynamic pipelines to the Vertex orchestrator

schustmi:
## Describe changes
This PR adds support for dynamic pipelines to the Vertex orchestrator.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
+++ b/src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
@@ -144,6 +144,7 @@ class VertexOrchestrator(ContainerizedOrchestrator, GoogleCredentialsMixin):
     """Orchestrator responsible for running pipelines on Vertex AI."""
 
     _pipeline_root: str
+    _job_service_client: Optional[aiplatform.gapic.JobServiceClient] = None
 
     @property
     def config(self) -> VertexOrchestratorConfig:
@@ -260,6 +261,25 @@ class VertexOrchestrator(ContainerizedOrchestrator, GoogleCredentialsMixin):
             Path to the pipeline directory.
         """
         return os.path.join(self.root_directory, "pipelines")
+
+    def get_job_service_client(self) -> aiplatform.gapic.JobServiceClient:
+        """Get the job service client.
+
+        Returns:
+            The job service client.
+        """
+        if self.connector_has_expired():
+            self._job_service_client = None
+
+        if self._job_service_client is None:
+            credentials, _ = self._get_authentication()
+            client_options = {
+                "api_endpoint": self.config.location + VERTEX_ENDPOINT_SUFFIX
+            }
+            self._job_service_client = aiplatform.gapic.JobServiceClient(
+                credentials=credentials, client_options=client_options
+            )
+        return self._job_service_client
 
     def _create_container_component(
         self,
@@ -696,34 +716,38 @@ class VertexOrchestrator(ContainerizedOrchestrator, GoogleCredentialsMixin):
             network=self.config.network,
         )
 
-        credentials, project_id = self._get_authentication()
-        client_options = {
-            "api_endpoint": self.config.location + VERTEX_ENDPOINT_SUFFIX
-        }
-        client = aiplatform.gapic.JobServiceClient(
-            credentials=credentials, client_options=client_options
+        client = self.get_job_service_client()
+        parent = (
+            f"projects/{self.gcp_project_id}/locations/{self.config.location}"
         )
-        parent = f"projects/{project_id}/locations/{self.config.location}"
         job_model = client.create_custom_job(
             parent=parent, custom_job=job_request
         )
 
-        wait_for_completion = None
+        _wait_for_completion = None
         if settings.synchronous:
-            wait_for_completion = lambda: monitor_job(
-                job_id=job_model.name,
-                credentials_source=self,
-                client_options=client_options,
-            )
 
-        self._initialize_vertex_client()
-        job = aiplatform.CustomJob.get(job_model.name)
+            def _wait_for_completion() -> None:
+                logger.info("Waiting for the VertexAI job to finish...")
+                monitor_job(
+                    job_id=job_model.name,
+                    get_client=self.get_job_service_client,
+                )
+                logger.info("VertexAI job completed successfully.")
+
+        credentials, project_id = self._get_authentication()
+        job = aiplatform.CustomJob.get(
+            job_model.name,
+            project=project_id,
+            location=self.config.location,
+            credentials=credentials,
+        )
         metadata = self.compute_metadata(job)
 
         logger.info("View the Vertex job at %s", job._dashboard_uri())
 
         return SubmissionResult(
-            wait_for_completion=wait_for_completion,
+            wait_for_completion=_wait_for_completion,
             metadata=metadata,
         )
 
@@ -765,14 +789,10 @@ class VertexOrchestrator(ContainerizedOrchestrator, GoogleCredentialsMixin):
             network=self.config.network,
         )
 
-        credentials, project_id = self._get_authentication()
-        client_options = {
-            "api_endpoint": self.config.location + VERTEX_ENDPOINT_SUFFIX
-        }
-        client = aiplatform.gapic.JobServiceClient(
-            credentials=credentials, client_options=client_options
+        client = self.get_job_service_client()
+        parent = (
+            f"projects/{self.gcp_project_id}/locations/{self.config.location}"
         )
-        parent = f"projects/{project_id}/locations/{self.config.location}"
         logger.info(
             "Submitting custom job='%s', path='%s' to Vertex AI Training.",
             job_request["display_name"],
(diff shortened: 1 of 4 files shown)


htahir1:
@claude full-review

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
```diff
@@ -200,150 +176,45 @@ def launch(
...
-        #   The following logic will keep polling the state of the job until
-        #   the job enters a final state.
-        #
-        # During the polling, if a connection error was encountered, the GET
-        # request will be retried by recreating the Python API client to
-        # refresh the lifecycle of the connection being used. See
-        # https://github.com/googleapis/google-api-python-client/issues/218
-        # for a detailed description of the problem. If the error persists for
-        # _CONNECTION_ERROR_RETRY_LIMIT consecutive attempts, the function
-        # will raise ConnectionError.
```
stefannica, line 303: Can you please keep comments like these? I know you're not a fan of code comments that are redundant, but this is the opposite of that.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -811,14 +975,26 @@ def get_pipeline_run_metadata(
...
-        )
+        if ENV_ZENML_VERTEX_RUN_ID in os.environ:
+            # Static pipeline -> Pipeline job
+            run_url = (
+                f"https://console.cloud.google.com/vertex-ai/locations/"
+                f"{self.config.location}/pipelines/runs/"
+                f"{self.get_orchestrator_run_id()}"
+            )
+        else:
+            # Dynamic pipeline -> Custom job
```
stefannica, line 1005: This felt forced when I was enabling dynamic pipeline support for Sagemaker and I can see this feels equally out of place. In the future, I think we should figure out a way to pass the snapshot to `get_pipeline_metadata` and use it as the source of truth instead of relying on these third party artifacts.

src/zenml/integrations/gcp/utils.py
```diff
@@ -0,0 +1,215 @@
...
+            access the job.
+        client_options: The client options to use for the job service client.
+
+    Raises:
+        RuntimeError: If the job fails.
+    """
+    retry_count = 0
+    client = get_job_service_client(
+        credentials_source=credentials_source, client_options=client_options
+    )
```
stefannica: Here and everywhere else you need to fetch credentials and configure clients, you have to make a few changes to account for the fact that now you need more clients and credentials may need to be valid for far longer:

1. keep a cache of these clients and reuse them across steps instead of creating new ones every time you need one
2. if they are based on service connector, make sure to check the service connector expiry every time you use the client and re-create the client if the credentials expired

If you think it helps, you can check the sagemaker orchestrator for a similar pattern.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
schustmi: Yes agreed, I was just too lazy to do it as part of this PR 😄

src/zenml/integrations/gcp/utils.py
schustmi: I think all of them are one-time use apart from the monitoring, where I implemented the retry functionality if the connector expired. You're right though, I think we can avoid one recreation of the client, I will update the implementation

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
schustmi: I would say that 95% of the words are boilerplate and not useful at all. This is the most standard and basic way to monitor something, which is to poll a state and sleep. And it has retries if a request fails. There is no magic to it. I'll keep the two links though in a comment if you think they're useful.

schustmi: The first link doesn't even work anymore, and you can literally see the values by clicking on the Enum in your code editor where you're reading the code.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -621,6 +656,143 @@ def dynamic_pipeline() -> None:
...
+            resource_settings=snapshot.pipeline_configuration.resource_settings,
+            environment=environment,
+            labels=labels,
+            encryption_spec_key_name=self.config.encryption_spec_key_name,
+            service_account=self.config.workload_service_account,
+            network=self.config.network,
+        )
+
+        _, project_id = self._get_authentication()
+        client = self.get_job_service_client()
```
stefannica: this ends up calling `_get_authentication` twice, which has the unfortunate effect that it uses the service connector twice to generate credentials. If you need the project ID, you should perhaps store it as a local variable and set it when you call `get_job_service_client`

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -621,6 +656,143 @@ def dynamic_pipeline() -> None:
...
+            resource_settings=step_run_info.config.resource_settings,
+            environment=environment,
+            labels=labels,
+            encryption_spec_key_name=self.config.encryption_spec_key_name,
+            service_account=self.config.workload_service_account,
+            network=self.config.network,
+        )
+
+        _, project_id = self._get_authentication()
+        client = self.get_job_service_client()
```
stefannica: same here: you create a new set of credentials that you don't use just so that you fetch the project ID

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -935,129 +1122,161 @@ def fetch_status(
...
-            pipeline_status = ExecutionStatus.COMPLETED
-        elif status == PipelineState.PIPELINE_STATE_CANCELLING:
-            pipeline_status = ExecutionStatus.STOPPING
-        elif status == PipelineState.PIPELINE_STATE_CANCELLED:
-            pipeline_status = ExecutionStatus.STOPPED
-        elif status == PipelineState.PIPELINE_STATE_FAILED:
-            pipeline_status = ExecutionStatus.FAILED
+
+        if run.snapshot and run.snapshot.is_dynamic:
+            status = aiplatform.CustomJob.get(run_id).state
```
stefannica: isn't there a way to use the client object returned by `get_job_service_client` at least here, for dynamic pipelines, instead of relying on this global aiplatform state? I really don't trust the `aiplatform` global state to do the right thing here, especially in situations where multiple ochestrator instances are created in the server.

src/zenml/integrations/gcp/step_operators/vertex_step_operator.py
```diff
@@ -200,150 +197,39 @@ def launch(
...
-                "kmsKeyName": self.config.encryption_spec_key_name
-            }
-            if self.config.encryption_spec_key_name
-            else {},
-        }
-        logger.debug("Vertex AI Job=%s", custom_job)
+        logger.debug("Vertex AI Job=%s", job_request)
+
+        _, project_id = self._get_authentication()
+        client = self.get_job_service_client()
```
stefannica: same thing here about the useless call to generate credentials

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
schustmi: I actually found a way to use the aiplatform API without doing the global initialization, as both the `PipelineJob` as well as the `CustomJob` allow passing region, project and credentials in their initialization. I'm not sure why we even used the global initializer before. That should be fine as well, right?

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
schustmi: It's now caching the credentials and the client in all the places. I think it's quite ugly, and IMO should be responsibility of the service connector. But I don't think I'll be touching that as part of this PR.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
stefannica: I am not sure. Caching the credentials sounds okay, but it should be the client that is actually cached instead of passing the credentials to `aiplatform.CustomJob.get`, because the client also keeps a list of open connections or a connection pool that you want to reuse across client calls. I think calls like `aiplatform.CustomJob.get` end up creating new clients or connections under the hood, which makes them inefficient if you make multiple calls to it.

stefannica: yup, if you follow the bread crumbs, you'll find that `aiplatform.CustomJob.get` eventually creates a temporary one-shot client (see below).
image

this makes calls like `aiplatform.CustomJob.get` very inefficient if called multiple times in a short period of time, but I don't think that's your case. It looks like you're only calling it once per process, which should be okay.

the only case in which this stops being okay is if the Vertex orchestrator is used in a Deployment to run multiple pipelines in different threads, but that is something of a sci-fi scenario for now at least :)

PR #4247, opened 2025-11-26
Feature:4135 Orchestrator generic graceful stopping

Json-Andriopoulos:
## Describe changes

This story extends the existing step-heartbeat logic by:

- Letting users set a maximum “heartbeat_healthy_threshold” for heartbeat-enabled steps before they’re deemed unhealthy (system-gated).
- Adding evaluation of step health from heartbeat signals.
- Using that health check to preemptively stop unhealthy step execution in the Kubernetes orchestrator.
- Using that health check to invalidate pipeline tokens when a pipeline enters an unhealthy state.
- Ensuring all orchestrators can gracefully stop running heartbeat-enabled steps when a pipeline is stopped.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/migrations/versions/2837a7ab3d77_add_cached_heartbeat_threshold_column.py
+++ b/src/zenml/zen_stores/migrations/versions/2837a7ab3d77_add_cached_heartbeat_threshold_column.py
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "2837a7ab3d77"
-down_revision = "5c0a1c787128"
+down_revision = "6e4eb89f632d"
 branch_labels = None
 depends_on = None
 


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -720,13 +710,84 @@ def _stop_run(
...
+        Args:
+            pipeline_run: A pipeline run response to stop (already updated to STOPPING status).
 
         Raises:
-            NotImplementedError: If any orchestrator inheriting from the base
-                class does not implement this logic.
+            RuntimeError: If steps fail to be set to STOPPING or steps don't have heartbeat enabled.
         """
+
+        from zenml.client import Client
```
schustmi: As discussed, I think to avoid any race conditions and make this much simpler, we should set the pipeline run status instead and adjust the server-side logic to look at this one when deciding on the heartbeat response.

src/zenml/models/v2/core/step_run.py
```diff
@@ -239,6 +239,7 @@ class StepRunResponseBody(ProjectScopedResponseBody):
         title="The substitutions of the step run.",
         default={},
     )
+    cached_heartbeat_threshold: Optional[int] = Field(title="", default=None)
```
schustmi: Missing title here I think?

Also, I'm not sure what the `cached` in the name here indicates? We can also just call this `heartbeat_threshold` right? I'm wondering whether we even need this attribute on the response, or if it's enough just on the DB schema. When we have the model already, we might as well just do `.config.heartbeat_threshold` I assume?

src/zenml/steps/heartbeat.py
```diff
@@ -161,8 +167,71 @@ def _heartbeat(self) -> None:
...
+        True if the step heartbeat is unhealthy, False otherwise.
+    """
+    if not heartbeat_threshold:
+        return False
+
+    if status.is_finished:
+        return False
+
+    if latest_heartbeat:
+        heartbeat_diff = datetime.now(tz=timezone.utc) - to_local_tz(
```
schustmi: Why do we need to convert to the local timezone here, and then subtract that from a utc timezone datetime?

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -117,7 +117,7 @@ class StepRunSchema(NamedSchema, RunMetadataInterface, table=True):
             nullable=True,
         )
     )
-
+    cached_heartbeat_threshold: Optional[int] = Field(nullable=True)
```
schustmi: I would call this just `heartbeat_threshold`, the `cached_` IMO just confuses things. This is really just the configured threshold right, whether this is a duplicate from the config doesn't really play a role when using it?

src/zenml/utils/pagination_utils.py
```diff
@@ -40,3 +40,28 @@ def depaginate(
         items += list(page.items)
 
     return items
+
+
+def depaginate_stream(
```
schustmi, line 45: I think this is not used anymore? Feel free to leave it in though, I've always thought we should have an option to get them as an iterator, potentially even on our client methods

src/zenml/steps/base_step.py
```diff
@@ -1000,6 +1009,7 @@ def _convert_to_tuple(value: Any) -> Tuple[Source, ...]:
                 "substitutions": substitutions,
                 "cache_policy": cache_policy,
                 "runtime": runtime,
+                "heartbeat_healthy_threshold": heartbeat_healthy_threshold,
```
schustmi, line 1012: There is also `BaseStep.with_options(...)` that needs to have this new argument

src/zenml/steps/heartbeat.py
```diff
@@ -161,8 +167,71 @@ def _heartbeat(self) -> None:
...
 
-        if response.status in {
+        if response.pipeline_run_status in {
             ExecutionStatus.STOPPED,
             ExecutionStatus.STOPPING,
         }:
             self._terminated = True
+
+
+def cached_is_heartbeat_unhealthy(
```
schustmi: Not sure I understand the `cached` in the name here either?

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator_entrypoint.py
```diff
@@ -652,6 +664,12 @@ def check_job_status(node: Node) -> NodeStatus:
                     error_message,
                 )
                 return NodeStatus.FAILED
+            elif is_node_heartbeat_unhealthy(node):
+                logger.error(
+                    "Heartbeat for step `%s` indicates unhealthy status.",
+                    step_name,
+                )
+                return NodeStatus.FAILED
```
schustmi, line 679: If we get to this point, it means the Kubernetes job is still actively running but for some reason the heartbeat stopped, correct? In that case, it would probably be a good idea to also stop the job from continuing?

src/zenml/orchestrators/step_runner.py
```diff
@@ -251,46 +251,55 @@ def run(
...
-                        # depend on the step launcher to publish the exception info
+                    if (
+                        isinstance(step_exception, KeyboardInterrupt)
+                        and heartbeat_worker.is_terminated
+                    ):
                         Client().zen_store.update_run_step(
                             step_run_id=step_run_info.step_run_id,
                             step_run_update=StepRunUpdate(
-                                exception_info=exception_info,
+                                status=ExecutionStatus.STOPPING,
```
schustmi, line 246: Any reason you update to `STOPPING` here instead of `STOPPED`?

src/zenml/steps/heartbeat.py
Json-Andriopoulos: We need to have both objects timezone-aware. Basically this part of the utiltiy:

```
if dt.tzinfo is None:
    dt = dt.replace(tzinfo=timezone.utc)
```

src/zenml/utils/pagination_utils.py
Json-Andriopoulos: yy decided to leave it a more memory-safe option.

src/zenml/orchestrators/step_runner.py
Json-Andriopoulos: The logic behind this is that step runner sets to stopping and the step launcher which could be in a different process does any post execution steps and sets to stopped. It could be set to stopped directly I guess but to me it makes more sense as status transitions go (it is truly stopped after the step-launcher handles the signal/interruption).

src/zenml/models/v2/core/step_run.py
```diff
@@ -609,6 +609,17 @@ def latest_heartbeat(self) -> Optional[datetime]:
...
+    @property
+    def heartbeat_threshold(self) -> Optional[int]:
+        """The `heartbeat_threshold` property.
+
+        Returns:
+            the value of the property.
+        """
+        if self.get_metadata().spec.enable_heartbeat:
+            return self.get_metadata().config.heartbeat_healthy_threshold
+        return None
```
schustmi, line 629:
```suggestion
        if self.spec.enable_heartbeat:
            return self.config.heartbeat_healthy_threshold
        return None
```

src/zenml/orchestrators/step_runner.py
schustmi: Makes sense! I'm wondering if we will run into some cases where the step will stay in `STOPPING` because the step launcher never gets a chance to report back that state, but let's hope that's not the case.

src/zenml/zen_stores/migrations/versions/2837a7ab3d77_add_cached_heartbeat_threshold_column.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = "0.92.0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    with op.batch_alter_table("step_run", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("heartbeat_threshold", sa.Integer(), nullable=True)
```
bcdurak, line 23: Just out of curiosity, are we adding this as a column, because we don't want to load the step configuration each time we utilize it?

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -117,7 +117,7 @@ class StepRunSchema(NamedSchema, RunMetadataInterface, table=True):
             nullable=True,
         )
     )
-
+    heartbeat_threshold: Optional[int] = Field(nullable=True)
```
bcdurak, line 120: I am a bit confused about this. There is migration script which adds this attribute here. If i am not mistaken we only set this during the `create_run_step` method of the `sql_zen_store`. Then, when we want to read it in the corresponding response model, I believe we infer it from the config of the step, which ends up loading the step configuration, which defeats the purpose of putting it as a separate column I think. This might be completely ignored atm.

src/zenml/zen_stores/migrations/versions/2837a7ab3d77_add_cached_heartbeat_threshold_column.py
```diff
@@ -0,0 +1,30 @@
+"""Add cached heartbeat threshold column [2837a7ab3d77].
+
+Revision ID: 2837a7ab3d77
+Revises: d203788f82b9
```
bcdurak, line 4:
```suggestion
Revises: 0.92.0
```

bcdurak: Also, if I end up merging the log PR, this might have to change.

src/zenml/steps/heartbeat.py
```diff
@@ -161,8 +167,71 @@ def _heartbeat(self) -> None:
...
+
+    logger.debug("Step %s heartbeat diff=%s", step_run_id, heartbeat_diff)
+
+    if heartbeat_diff.total_seconds() > heartbeat_threshold * 60:
+        return True
+
+    return False
+
+
+def is_step_run_heartbeat_unhealthy(step_run: "StepRunResponse") -> bool:
```
bcdurak: What do you think of removing this function for cleaning up purposes? As far as I can see, it just maps the attributes of a step to the right inputs of the main heartbeat check function and is only used once.

src/zenml/models/v2/core/step_run.py
bcdurak: If we are adding this as a separate column to the DB, I would even say that it should be fetched directly like the other columns.

src/zenml/config/step_configurations.py
```diff
@@ -221,6 +221,14 @@ class StepConfigurationUpdate(FrozenBaseModel):
         "run inline unless a step operator or docker/resource settings "
         "are configured. This is only applicable for dynamic pipelines.",
     )
+    heartbeat_healthy_threshold: int = Field(
```
bcdurak, line 224: Should we throw a warning if someone tries to set this for a step that can not have a heartbeat based on its spec?

src/zenml/models/v2/core/step_run.py
bcdurak, line 851: Just as a small future reference, you can also use the `misc` module within the `models` for light weight models such as this in the future. Almost every file within the `core` module features the corresponding request, response and filter models. Feel free to leave it here though.

bcdurak (commented):
Exciting changes coming, left a few comments for now.

src/zenml/zen_stores/migrations/versions/2837a7ab3d77_add_cached_heartbeat_threshold_column.py
Json-Andriopoulos: Yeap, as a fast shortcut for the step_in_progress check.

src/zenml/config/step_configurations.py
Json-Andriopoulos: Yeap good idea.

src/zenml/steps/heartbeat.py
Json-Andriopoulos: Good point

src/zenml/zen_stores/schemas/step_run_schemas.py
Json-Andriopoulos: The value of the field is a bit confusing for end-users so we don't want it to flow end-to-end to the API payload (which reminds I need to test that it doesn't). We want the DB column as a fast "cacheing" of the value to access it from low-level models (like the Schema classes so we can access it fast with simple select statements and not complex joins and json payload validations).

bcdurak: @Json-Andriopoulos I understand the motivation, which I also like 👍 The execution works for the schemas that you mentioned. The problem is the fact that this "cacheing" is not effective when it comes to the response models. For instance, there is a section in the Kubernetes entrypoint that looks like this:

```python
            sr_ = client.list_run_steps(
                name=node.id, pipeline_run_id=pipeline_run.id
            )

            if sr_.items:
                return is_heartbeat_unhealthy(
                    step_run_id=sr_.items[0].id,
                    status=sr_.items[0].status,
                    start_time=sr_.items[0].start_time,
                    heartbeat_threshold=sr_.items[0].heartbeat_threshold,
                    latest_heartbeat=sr_.items[0].latest_heartbeat,
                )
```

Currently, in examples like this, `sr_.items[0]` is a `StepRunResponse` and when you try to get the `heartbeat_threshold` of a `StepRunResponse`, you will have to call the `get_metadata(...)` function (with all the extra complex joins and json payload validations). This can get out of hand for repetitive operations like checking whether the heartbeat is healthy or not. 

What I am proposing is this: Instead of fetching it like we do right now, let's adjust the `to_model(...)` method of the  step schemas to add this column to the body and this way you can change the property to this:
```python
   @property
    def heartbeat_threshold(self) -> Optional[int]:
        """The `heartbeat_threshold` property.

        Returns:
            the value of the property.
        """
        if self.get_body().heartbeat_threshold
 ```
        
What do you think?

PR #4249, opened 2025-11-26
[WIP] Fix set serialization in metadata system

### LINKED ISSUE #4248: Allow sets on Metadata

_No response_

### System Information

Though set is listed as allowed in MetadataType it can't be used, cause there is a json.dumps call under the hood, which fails for obvious reasons on sets.

### What happened?

Sets dont work in MetadataType

### Reproduction steps

just log some set as run metadata.

### Relevant log output

```shell

```

### Code of Conduct

htahir1:
Just running this to run tests

Sets were documented as a supported MetadataType but failed during serialization because json.dumps() cannot serialize Python sets. This change adds a helper function to convert sets (and tuples) to lists before JSON serialization, making them compatible with JSON while preserving type information via MetadataTypeEnum.

Fixes #4248

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -75,6 +75,7 @@ jobs:
           remove-android: 'true'
           remove-haskell: 'true'
           build-mount-path: /var/lib/docker/
+        if: inputs.os == 'ubuntu-latest'
       - name: Checkout code
         uses: actions/checkout@v4.2.2
         with:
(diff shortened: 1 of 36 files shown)


src/zenml/metadata/metadata_types.py
```diff
@@ -210,6 +210,24 @@ def cast_to_metadata_type(
     return typed_value  # type: ignore[no-any-return]
 
 
+def serialize_metadata_value(value: MetadataType) -> str:
```
schustmi: This is not a good implementation. It does cover top-level sets, but for example fails if it's nested in some other element like a dictionary or list. There is a way to do this natively with json encoders, and there even is a pydantic encoder that we use in our code to solve this in other places.

schustmi (changes requested):

PR #4252, opened 2025-11-26
Fix ImportError when using zenml login without SQL dependencies

### LINKED ISSUE #4251: `zenml login` fails without SQL dependencies when connecting to remote server

_No response_

### System Information

OS: Linux 6.8.0-1030-azure
Python: 3.10
ZenML Version: 0.91.2
Installation: pip install zenml (base package, no extras)

### What happened?

When trying to connect to a **self-hosted remote ZenML server (deployed on Kubernetes)** using `zenml login https://your-zenml-server.com`, the command fails with an ImportError about missing `sqlalchemy_utils`, even though:

1. The base `zenml` package should be sufficient for connecting to remote servers (as documented at https://docs.zenml.io/getting-started/installation#base-package)
2. SQL dependencies should only be needed for local SQL store usage, not remote connections

### Workaround
Installing `zenml[local]` instead of the base `zenml` package resolves the issue. However, this contradicts the [documentation](https://docs.zenml.io/getting-started/installation#base-package) which states:
> "Installing the base package only allows you to connect to a deployed ZenML server."

This suggests the base package *should* work for remote connections without needing SQL extras.

## Probable Root Cause

The issue occurs in the following call chain:

1. **`zenml/cli/login.py:898`**
   - Calls `gc.store_configuration` to check if user is already connected to a server

2. **`zenml/config/global_config.py:634`**
   - `store_configuration` property calls `_get_store_configuration()`

3. **`zenml/config/global_config.py:567`**
   - When no store config exists (first run), calls `self.get_default_store()`
   - This unconditionally creates a default SQL store

4. **`zenml/zen_stores/base_zen_store.py:260-263`**
   - `get_default_store()` imports SQL store modules:
   ```python
   from zenml.zen_stores.secrets_stores.sql_secrets_store import SqlSecretsStoreConfiguration
   from zenml.zen_stores.sql_zen_store import SqlZenStoreConfiguration
   ```

5. **`zenml/zen_stores/secrets_stores/sql_secrets_store.py:16-25`**
   - Module imports fail if `sqlalchemy`, `sqlmodel`, or `sqlalchemy_utils` aren't installed

### Key Issue

The `zenml login` command triggers default SQL store creation even though the user is connecting to a remote server and doesn't need a local SQL store.

## Ideas
- Add a parameter to `_get_store_configuration()` to skip default store creation
- Use lazy imports in the SQL store modules
- Check for SQL dependencies before attempting to create default store in `zenml login`

### Reproduction steps

1. Create a fresh Python 3.10 conda environment
2. Install base ZenML: `pip install zenml` (no extras)
3. Try to login to a self-hosted remote ZenML server: `zenml login https://your-zenml-server.com`

**Note:** The remote server is a self-hosted ZenML deployment on Kubernetes (not ZenML Pro).

### Relevant log output

```shell
$ zenml login https://your-zenml-server.com
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /anaconda/envs/zenml/lib/python3.10/site-packages/zenml/cli/login.py:898 in login                │
│    895 │   # Get the server that the client is currently connected to, if any                    │
│    896 │   current_non_local_server: Optional[str] = None                                        │
│    897 │   gc = GlobalConfiguration()                                                            │
│ ❱  898 │   store_cfg = gc.store_configuration                                                    │
│ /anaconda/envs/zenml/lib/python3.10/site-packages/zenml/config/global_config.py:634 in           │
│ store_configuration                                                                              │
│ ❱ 634 │   │   return self._get_store_configuration()                                             │
│ /anaconda/envs/zenml/lib/python3.10/site-packages/zenml/config/global_config.py:567 in           │
│ _get_store_configuration                                                                         │
│ ❱ 567 │   │   │   store = self.get_default_store()                                               │
│ /anaconda/envs/zenml/lib/python3.10/site-packages/zenml/zen_stores/base_zen_store.py:260 in      │
│ get_default_store_config                                                                         │
│ ❱ 260 │   │   from zenml.zen_stores.secrets_stores.sql_secrets_store import (                    │
│   261 │   │   │   SqlSecretsStoreConfiguration,                                                  │
│   262 │   │   )                                                                                  │
│ /anaconda/envs/zenml/lib/python3.10/site-packages/zenml/zen_stores/secrets_stores/sql_secrets_st │
│ ore.py:19 in <module>                                                                            │
│ ❱  19 │   raise ImportError(...)                                                                 │
ImportError: It seems like you've installed the `zenml` package without the `local` 
extra, but are trying to use ZenML with a local database.
* If you want to use ZenML in a local setup, please install `zenml[local]` instead
* If you want to connect to a server, run `zenml login`
```

### Code of Conduct

strickvl:
## Summary

When a user installs the base ZenML package (`pip install zenml`) without the `[local]` extra and attempts to connect to a remote ZenML server using `zenml login https://your-server.com`, the command fails with an ImportError about missing `sqlalchemy_utils`.

This happens because the login command checks if the user is currently connected to a server by accessing `GlobalConfiguration().store_configuration`. When no config file exists yet, this triggers `get_default_store()` which unconditionally imports SQL store modules - even though the user only wants to connect to a REST-based remote server that doesn't need SQL dependencies.

**The fix:**

- Added an `allow_default` parameter to `_get_store_configuration()` that allows callers to skip default SQL store creation when no configuration exists
- Updated the login command to use `allow_default=False` when merely checking for existing connections
- This prevents SQL module imports when the user is connecting to REST-based remote servers

The base ZenML package should support connecting to remote ZenML servers without requiring local SQL dependencies, since REST connections only need HTTP client libraries.

Fixes #4251

## Test plan

- [ ] Install base zenml package only: `pip install zenml` (without `[local]` extra)
- [ ] Run `zenml login https://your-server.com` - should no longer fail with ImportError
- [ ] Verify existing functionality still works when SQL dependencies are installed
- [ ] Verify `zenml login --local` still correctly fails with a helpful message about missing dependencies

---

## Additional fix: Python 3.10 mock patch compatibility

This PR also fixes a Python 3.10-specific test failure introduced by PR #4224, where CLI tests using `zenml.cli.login.*` patch paths failed with `ModuleNotFoundError` or `AttributeError`.

**Root cause:** `zenml/cli/__init__.py` executes `from zenml.cli.login import *`, which binds the package attribute `zenml.cli.login` to the Click `Command` object named `login`. On Python 3.10, `unittest.mock.patch` uses a legacy `_importer` that resolves dotted names by walking existing package attributes—since `zenml.cli.login` refers to the Command, patching targets that object and fails. On Python 3.12+, `mock.patch` uses `pkgutil.resolve_name`, which imports submodules despite same-named attributes, so the patch works.

**Why it wasn't caught:** CI for PR #4224 only ran on Python 3.11, not Python 3.10, so the incompatibility wasn't detected.

**Fix:** 
- Import the login module BEFORE setting up mocks
- Use `mocker.patch.object(login_module, ...)` for all functions/classes bound at module load time
- This directly replaces attributes on the module object, bypassing both the attribute-shadowing issue and the import-binding timing issue

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,92 @@ Use filesystem navigation tools to explore the codebase structure as needed.
 - Use descriptive variable names and documentation
 - Keep function size manageable (aim for < 50 lines) though there are exceptions
 
+### Util Function Placement
+
+When deciding whether to place a helper function in a utils file or on a class, follow these guidelines:
+
+1. **If a method only makes sense within the context of a class** → Put it on the class
+2. **If a static/util method is heavily used by subclasses** → Put it on the parent class
+
+**Rationale for placing methods on classes:**
+- Saves imports for users and subclasses
+- Subclasses can simply call `self.something()` instead of finding and importing from a util file
+- Keeps related functionality co-located
+
+**Example:** `requires_resources_in_orchestration_environment` in `base_orchestrator.py:495-514`
+
+```python
+# This is a @staticmethod on BaseOrchestrator, not a standalone util
+@staticmethod
+def requires_resources_in_orchestration_environment(step: "Step") -> bool:
+    """Checks if the orchestrator should run this step on special resources."""
+    if step.config.step_operator:
+        return False
+    return not step.config.resource_settings.empty
+```
+
+This method could be a global util, but it's placed on the class because:
+- All orchestrator subclasses frequently need it
+- Subclasses can call `self.requires_resources_in_orchestration_environment(step)` without imports
+- It's conceptually tied to orchestrator behavior
+
+**When to use utils files:**
+- Truly generic functions used across unrelated modules
+- Functions that don't logically belong to any class
+- Pure utility functions (string manipulation, date formatting, etc.)
+
+**Key utils locations:**
+- `src/zenml/utils/` — General utilities
+- `src/zenml/orchestrators/utils.py` — Orchestrator-specific utilities
+- `src/zenml/orchestrators/step_run_utils.py` — Step execution utilities
+- `src/zenml/orchestrators/publish_utils.py` — Status/metadata publishing
+
+### Private Methods and API Stability
+
+Methods and functions starting with `_` (underscore) are **private** and should NOT be called from outside their class or module.
+
+**The rule:**
+- `_method()` on a class → only call from within that class
+- `_function()` in a utils module → only call from within that module
+- This isn't always consistently applied in the codebase, but it's the intended convention
+
+**Backwards compatibility — case-by-case judgment:**
+
+There are no strict written rules; evaluate each change individually:
+
+| Symbol type | Part of public API? | Breaking change if modified? |
+|-------------|---------------------|------------------------------|
+| Classes/functions exported in `zenml.__init__` | ✅ Definitely public | ⚠️ Yes — requires deprecation |
+| Public methods on those classes | ✅ Public | ⚠️ Yes — requires deprecation |
+| Internal methods deep in the codebase (no underscore) | ❌ Not intended for users | ✅ No — update all internal usages |
+| `_private_method()` | ❌ No | ✅ No — can change freely |
+
+**When changing any non-underscore method:**
+1. Check if the class/function is exported in `zenml.__init__` — if so, it's public API
+2. Search for usages **within the ZenML codebase** (grep/find references)
+3. Update all internal usages
+4. For truly internal code not exported at the root, no deprecation needed
+
+**Best practice for integrations (future-proofing):**
+
+> ⚠️ **Integrations should avoid using ZenML private methods**
+
+This is primarily a future concern: when integrations eventually move out of the main ZenML repo (external packages), mypy won't detect if a private method they depend on was changed, leading to silent breakage. Even while integrations live in-repo, using only public APIs is good practice and prepares for this transition.
+
+```python
+# Bad - integration code using private method
+from zenml.orchestrators.base_orchestrator import BaseOrchestrator
+
+class MyOrchestrator(BaseOrchestrator):
+    def submit_pipeline(self, ...):
+        self._some_private_helper()  # ❌ Don't do this
+
+# Good - use only public methods or reimplement logic
+class MyOrchestrator(BaseOrchestrator):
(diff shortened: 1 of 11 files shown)


strickvl:
@claude /full-review

src/zenml/config/global_config.py
```diff
@@ -489,8 +489,10 @@ def get_config_environment_vars(self) -> Dict[str, str]:
         return environment_vars
 
     def _get_store_configuration(
-        self, baseline: Optional[StoreConfiguration] = None
-    ) -> StoreConfiguration:
+        self,
```
schustmi, line 506: This can be much nicer with typing overloads, where it returns a non-optional store config in case `allow_default=True` is passed.

schustmi (changes requested):

src/zenml/config/global_config.py
strickvl: TIL about `@overload`! will make that fix.

src/zenml/cli/login.py
```diff
@@ -952,8 +952,8 @@ def login(
     # Get the server that the client is currently connected to, if any
     current_non_local_server: Optional[str] = None
     gc = GlobalConfiguration()
-    store_cfg = gc.store_configuration
-    if store_cfg.type == StoreType.REST:
+    store_cfg = gc._get_store_configuration(allow_default=False)
```
stefannica: could you please create a second public property and call it `store_configuration_no_default` or similar instead of calling this `_get_store_configuration` private method from here ?

strickvl: haha @schustmi just what we were talking about :)

schustmi: Just make the method non-private then? I don't think we need a separate one @stefannica ?

stefannica: that works too

strickvl: > Just make the method non-private then? I don't think we need a separate one @stefannica ?

Ok doing this.

schustmi: What do you think @strickvl ? With the current implementation, we now have three different properties/methods that do essentially the same thing 😅

strickvl: Done

PR #4265, opened 2025-11-28
Fix weave import side-effects when `enable_weave=False`

strickvl:
## Summary

- Fixes weave import triggering `wandb_init_hook` and remote project creation even when `enable_weave=False`
- Removes unnecessary `else` branch that imported weave and called `weave.init(..., settings={"disabled": True})`
- Prevents `PERMISSION_ERROR` for users not logged into W&B who have Weave disabled

## Problem

With wandb 0.23.0 + weave 0.52.17+, runs fail even when Weave is explicitly disabled. The root cause:

1. The old code imported `weave` in both branches (enabled and disabled)
2. With newer weave versions, `import weave` alone loads ~548 modules including `weave.trace.autopatch`
3. This triggers `wandb_init_hook` which attempts remote project creation
4. Users without credentials get `PERMISSION_ERROR` despite having `enable_weave=False`

## Solution

Simply remove the `else` branch. When `enable_weave=False`, we don't import or interact with weave at all. Not importing a library is the ultimate way to disable it.

```diff
         if settings.enable_weave:
             import weave
             if self.config.project_name:
                 weave.init(project_name=self.config.project_name)
-        else:
-            import weave
-            if self.config.project_name:
-                weave.init(project_name=..., settings={"disabled": True})
```

## Test plan

- [x] Verified fix with wandb 0.23.0 + weave 0.52.20 in isolated venv
- [x] Confirmed `import weave` loads 548 modules with potential side-effects
- [x] Confirmed the fix prevents any weave-related code when disabled
- [ ] Manual testing with ZenML pipeline using wandb tracker with `enable_weave=False`

### DIFF
--- a/src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
+++ b/src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
@@ -180,12 +180,3 @@ class WandbExperimentTracker(BaseExperimentTracker):
                     "Weave enabled but no project_name specified. "
                     "Skipping weave initialization."
                 )
-        else:
-            import weave
-
-            if self.config.project_name:
-                logger.info("Disabling weave")
-                weave.init(
-                    project_name=self.config.project_name,
-                    settings={"disabled": True},
-                )


strickvl:
> is this tested? i remember i added this else statement for a reason but i forgot why

Not yet, but I'd be interested in the reason if you remember it. I couldn't understand why that was there.

htahir1:
@strickvl sadly i cant remember... maybe because at some point weave was enabled by default or soemthing? or i wanted to make sure if false was passed to disable it for good? i just cant remember it rn

strickvl:
@AlexejPenner @htahir1 I tested this just now:

Manually tested with:
  - wandb 0.23.0 + weave 0.52.20
  - Fresh venv with ZenML from this branch
  - Pipeline with wandb experiment tracker and `enable_weave=False`

  Verified:
  1. Main `weave` package is NOT imported when disabled (only harmless `wandb.integration.weave.*` stubs load)
  2. Pipeline runs successfully with no permission errors
  3. Metrics logged correctly to W&B

htahir1:
@strickvl Then lets merge!

src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
```diff
@@ -180,12 +180,3 @@ def _initialize_wandb(
                     "Weave enabled but no project_name specified. "
                     "Skipping weave initialization."
                 )
-        else:
-            import weave
-
-            if self.config.project_name:
-                logger.info("Disabling weave")
-                weave.init(
```
bcdurak, line 188: I am not sure what this implies, if the issue is to import it, perhaps we can cover it in a `try-except`. Basically, you can still try to disable it if weave is installed?

strickvl: yeah Hamza also couldn't remember what it implied @bcdurak but I tested it 🤷

bcdurak: Yeah I saw the discussion above. He mentioned that he implemented this for a reason. That's exactly why I still think that it would be safer to just wrap this in a `try-except`, but I will leave it up to you.

htahir1: its fine lets merge it

bcdurak: @htahir1 it is already merged.

bcdurak: it's just a late comment here.

PR #4270, opened 2025-12-01
Enable orchestration environment restarts for dynamic pipelines

schustmi:
## Describe changes
This PR adds support for restarting the orchestrator environment when running dynamic pipelines. For orchestrators to support this, they must implement restarts of the orchestration container, and additionally also make sure that the `get_orchestrator_run_id()` method returns the same value even after a restart. This is currently only implemented for the Kubernetes orchestrator.

**Technical implementation details:**
- When restarting the orchestration environment, we re-execute the pipeline function
- When a step function is executed, we first check if a step run for the given invocation ID already exists
  - If a step run exists, we either return its results (in case it finished) or restart monitoring (in case it's still running). If the step is running in `inline` mode, we instead mark it as failed and potentially retry it.
  - If no step run exists, we run it as usual

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/execution/pipeline/dynamic/runner.py
+++ b/src/zenml/execution/pipeline/dynamic/runner.py
@@ -185,6 +185,9 @@ class DynamicPipelineRunner:
     def run_pipeline(self) -> None:
         """Run the pipeline."""
         if self._run:
+            if self._run.status.is_finished:
+                logger.info("Run `%s` is already finished.", str(self._run.id))
+                return
             if self._run.orchestrator_run_id:
                 logger.info("Continuing existing run `%s`.", str(self._run.id))
                 run = self._run
@@ -360,9 +363,9 @@ class DynamicPipelineRunner:
         )
 
         should_retry = _should_retry_locally(
-            compiled_step,
-            self._snapshot.pipeline_configuration.docker_settings,
-            self._orchestrator,
+            step=compiled_step,
+            pipeline_docker_settings=self._snapshot.pipeline_configuration.docker_settings,
+            orchestrator=self._orchestrator,
         )
 
         def _launch_step_and_load_outputs(
@@ -399,6 +402,7 @@ class DynamicPipelineRunner:
             runtime = get_step_runtime(
                 step_config=compiled_step.config,
                 pipeline_docker_settings=self._snapshot.pipeline_configuration.docker_settings,
+                orchestrator=self._orchestrator,
             )
             if (
                 runtime == StepRuntime.INLINE
@@ -425,7 +429,7 @@ class DynamicPipelineRunner:
             if step_run.status == ExecutionStatus.RUNNING:
                 logger.info(
                     "Restarting the monitoring of existing step `%s` "
-                    "(step run ID: %s). Remaining retries: %d",
+                    "(ID: %s). Remaining retries: %d",
                     step_run.name,
                     step_run.id,
                     remaining_retries,


src/zenml/enums.py
```diff
@@ -103,6 +105,7 @@ def is_finished(self) -> bool:
             ExecutionStatus.FAILED,
             ExecutionStatus.COMPLETED,
             ExecutionStatus.CACHED,
+            ExecutionStatus.RETRYING,
```
stefannica: Are you sure? this doesn't sound like a final state that should be interpreted as an execution completion.

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -972,9 +967,27 @@ def get_orchestrator_run_id(self) -> str:
...
+                    namespace=self.config.kubernetes_namespace,
+                )
+            except Exception as e:
+                logger.warning(
+                    "Failed to fetch orchestrator job name. Using pod name as "
+                    "fallback, which may cause issues if the orchestrator pod "
+                    "is restarted. Error: %s",
+                    e,
+                )
+            return orchestrator_run_id or pod_name
```
stefannica, line 1010: This is fine, but personally I am getting tired of this cat-mouse game we have to play to magically arrange for all environments related to the same pipeline run to get the same orchestrator run id.

Correct me if I'm wrong, but the only reason we need to have this orchestrator run ID in addition to the actual pipeline run ID to uniquely identify a run is schedules. I really hope with the upcoming native schedules support, we can get rid of this awkward orchestrator run ID concept.

src/zenml/execution/pipeline/dynamic/outputs.py
```diff
@@ -34,7 +35,27 @@ class OutputArtifact(ArtifactVersionResponse):
 StepRunOutputs = Union[None, OutputArtifact, Tuple[OutputArtifact, ...]]
 
 
-class _BaseStepRunFuture:
+class BaseFuture:
```
stefannica:
```suggestion
class BaseFuture(ABC):
```

IIRC, without this, you won't catch child classes that don't implement all abstract methods.

src/zenml/execution/pipeline/dynamic/utils.py
```diff
@@ -46,3 +55,31 @@ def unmapped(value: T) -> _Unmapped[T]:
...
+            return step_run
+
+        logger.debug(
+            "Waiting for step run with ID %s to finish (current status: %s)",
+            step_run_id,
+            step_run.status,
+        )
+        time.sleep(sleep_interval)
+        if sleep_interval < max_sleep_interval:
+            sleep_interval *= 2
```
stefannica, line 86: thanks for using an exponential backoff here !

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -158,17 +176,29 @@ def pipeline(self) -> "DynamicPipeline":
...
         else:
-            run = create_placeholder_run(
-                snapshot=self._snapshot,
+            existing_runs = Client().list_pipeline_runs(
+                snapshot_id=self._snapshot.id,
                 orchestrator_run_id=self._orchestrator_run_id,
             )
+            if existing_runs.total == 1:
+                run = existing_runs.items[0]
+                logger.info("Continuing existing run `%s`.", str(run.id))
```
stefannica: What if the run is already completed ?

stefannica: Have you thought about how this feature might interact with the step heartbeat feature ?

src/zenml/enums.py
schustmi: When a step run is published as `failed` but there are retries left, we set it to `retrying`. So it is a final step for one particular step run, but there will be other step runs published later which are retries of this one.

schustmi: It doesn't feel right to me either, maybe I catch it explicitly in the code where I need it

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
schustmi: Yes schedules and re-runs on the orchestration backend, which we semi-support I guess.

I'm with you, I don't really like it either but I still can't think of an alternative. Even when we introduce native schedules, as long as we keep around the orchestrator schedules I think this logic has to stay.

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: Good point, I handle that case now.

schustmi: Yes I think it should be fine from what I checked

src/zenml/execution/pipeline/dynamic/outputs.py
```diff
@@ -68,7 +69,27 @@ def chunk(self, index: int) -> "OutputArtifact":
 StepRunOutputs = Union[None, OutputArtifact, Tuple[OutputArtifact, ...]]
 
 
-class _BaseStepRunFuture:
+class BaseFuture(ABC):
+    """Base future."""
+
+    @abstractmethod
+    def running(self) -> bool:
```
Json-Andriopoulos, line 76: Maybe `is_running` would be a better name?

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -122,10 +131,19 @@ def __init__(
...
-        )
         self._futures: List[StepRunOutputsFuture] = []
+        self._invocation_ids: Set[str] = set()
+
+        self._existing_step_runs: Dict[str, "StepRunResponse"] = {}
+        if run and run.orchestrator_run_id:
+            self._orchestrator_run_id = run.orchestrator_run_id
+
+            if run.status == ExecutionStatus.RUNNING:
+                self._existing_step_runs = run.steps.copy()
```
Json-Andriopoulos, line 150: Should we do deepcopy() maybe?

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -197,6 +231,9 @@ def run_pipeline(self) -> None:
                         # them and raise any exceptions that occurred.
                         self.await_all_step_run_futures()
                     except:
+                        # TODO: this call already invalidates the token, so
+                        # the steps will keep running but won't be able to
```
Json-Andriopoulos, line 256: Maybe we should expand the status checks on the heartbeat. It now checks for stopped/stopping but we should tirgger interruption with failed as well?

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: Actually we don't even need a copy here, as `run.steps` is a property that fetches new steps each time we call it

src/zenml/execution/pipeline/dynamic/outputs.py
schustmi: I'd agree, but the current future interface is just trying to mirror the `concurrent.Futures` methods which users might already be used to? WDYT?

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: You mean if the pipeline run is failed, the heartbeat should interrupt the step? If so, we should only do that on the `fail_fast` execution mode I think, in other cases we want the steps to continue

PR #4288, opened 2025-12-03
Add pipeline run index attribute

schustmi:
## Describe changes

This PR adds an `index` attribute for each pipeline run, which specifies the index of the run within its pipeline.

Additionally, the `pipeline` attribute of the `PipelineRunRequest` was removed as the information is already included in the `snapshot` that is part of the request.

Limitations:
- If a run fails to be created (could be because a run with the same name exists or other reasons), the run number will still be bumped and that index will simply be unused
- When running a pipeline client-side, the following happens: We first create a pipeline run in the database, and then ask the orchestrator to submit the pipeline. If this submission fails, we currently delete the pipeline run. This will also cause a gap in the indexes of pipeline runs.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -6102,6 +6102,7 @@ class SqlZenStore(BaseZenStore):
                         jl_arg(PipelineRunSchema.start_time),
                         jl_arg(PipelineRunSchema.end_time),
                         jl_arg(PipelineRunSchema.status),
+                        jl_arg(PipelineRunSchema.index),
                     ),
                 ],
             )
@@ -6336,6 +6337,7 @@ class SqlZenStore(BaseZenStore):
                     for triggered_run in step_run.triggered_runs:
                         triggered_run_metadata: Dict[str, Any] = {
                             "status": triggered_run.status,
+                            "index": triggered_run.index,
                         }
 
                         if triggered_run.start_time:


src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -379,6 +381,7 @@ def from_request(
             orchestrator_environment=orchestrator_environment,
             start_time=request.start_time,
             status=request.status.value,
+            number=number,
```
htahir1: do we really want to call this "number"? Can we just say "index"? @schustmi

schustmi: Haven't thought about names yet, I'm open for any suggestion. Just have this PR for testing at the moment

schustmi: But yeah index sounds good to me, maybe @znegrin also has something in mind as this is his suggestion?

htahir1: verified with @znegrin , and `index` sounds good to us

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
```diff
@@ -0,0 +1,108 @@
...
+        sa.update(run_table)
+        .where(run_table.c.pipeline_id.is_(None))
+        .values(index=0)
+    )
+
+    result = connection.execute(
+        sa.select(
+            run_table.c.id,
+            run_table.c.pipeline_id,
+            run_table.c.created,
```
Json-Andriopoulos: Do we need to fetch created? We already fetch the results in order no?

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        .where(run_table.c.pipeline_id.is_not(None))
+        .order_by(run_table.c.pipeline_id, run_table.c.created, run_table.c.id)
+    ).fetchall()
```
Json-Andriopoulos, line 53: fetchall() fetches all results in memory (I am not sure if MySQL provides server-side cursors to stream results in batches - I think not). Let's think about the worst case scenario in terms of memory consumption.

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
```diff
@@ -0,0 +1,108 @@
...
+
+    current_pipeline_id = None
+    index_within_pipeline = 0
+    run_updates: list[dict[str, Any]] = []
+    run_counts: dict[str, int] = {}
+    for row in result:
```
Json-Andriopoulos, line 59: Normally I would advocate to execute data population for optional values in a post-deployment or post-migration step. Let's think about the worst case scenario in terms of execution time.

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
```diff
@@ -0,0 +1,108 @@
...
+            .values(index=sa.bindparam("index")),
+            run_updates,
+        )
+
+    if run_counts:
+        pipeline_updates = [
+            {"id_": pipeline_id, "run_count": run_count}
+            for pipeline_id, run_count in run_counts.items()
+        ]
+        connection.execute(
```
Json-Andriopoulos: I suggest we do that in batches. I am not sure what the supported limits are in terms of payload (if you try to update 10k records at once for instance). Best to have something more manageable (e.g. 1000 record batches) with a retry policy?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6504,6 +6504,31 @@ def _get_duplicate_run_name_error_message(
...
+
+        Returns:
+            The next run index for the pipeline.
+        """
+        # Commit before acquiring the exclusive lock on the pipeline
+        session.commit()
+        session.execute(
+            update(PipelineSchema)
+            .where(col(PipelineSchema.id) == pipeline_id)
+            .values(run_count=col(PipelineSchema.run_count) + 1)
```
Json-Andriopoulos, line 6524: Good one! I *think* this should be ok with the default isolation level in MySQL.

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
schustmi: Good idea!

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
schustmi: Nope you're right, I'll remove it

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
schustmi: Yeah I thought this should not be a problem here:
After I removed the created_at, this fetches 2 UUIDs (-> `32 bytes`), so even if someone had 10 million pipeline runs, this would consume 320 megabytes of memory. Should be okay, wdyt?

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
schustmi: Actually the index should be non-optional and the column type is changed at the end of the DB migration to reflect that, which is why we need some values for each existing row. I guess we could populate it with a placeholder and then later on replace by real values, not sure it's worth the effort here though?

src/zenml/zen_stores/sql_zen_store.py
schustmi: Yep I think so! Also in my tests it seemed to work just fine. I also didn't add a unique constraint on `pipeline_id, index`, so even if this for some reason fails to generate unique values, we will not fail to create a run but instead only have a duplicate index.

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
Json-Andriopoulos: I think as python objects go you will have a bit of additional memory but not significant, we should be good.

src/zenml/zen_stores/migrations/versions/6e4eb89f632d_unique_run_index.py
Json-Andriopoulos: Nah let's no change the migration logic, let's just estimate the execution time for the backfill (worst case scenario). The general rule is migrations should complete quickly.

stefannica: > Normally I would advocate to execute data population for optional values in a post-deployment or post-migration step

@Json-Andriopoulos I'm curious, why is it not acceptable to do these sort of changes in the alembic migration script ? what are the pros/cons ?

PR #4292, opened 2025-12-03
Add Changelog link to README

strickvl:


htahir1 (changes requested):
Two things:

There is a line "🎉 For the latest release, see the [release notes](https://github.com/zenml-io/zenml/releases)." -> we can point to changelog instead?

Also there is a file in the rpeo called RELEASE_NOTES.md i think we should delete now  to avoid confusion (or maybe add the changelog link at the top)

PR #4293, opened 2025-12-03
Removed outdated recommendation

AlexejPenner:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/how-to/steps-pipelines/advanced_features.md
+++ b/docs/book/how-to/steps-pipelines/advanced_features.md
@@ -281,78 +281,21 @@ Pipeline composition allows you to build complex workflows from simpler, well-te
 
 ### Fan-out and Fan-in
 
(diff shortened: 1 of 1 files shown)


docs/book/how-to/steps-pipelines/advanced_features.md
```diff
@@ -281,78 +281,21 @@ Pipeline composition allows you to build complex workflows from simpler, well-te
...
-
-- Parallel data processing
-- Distributed model training
-- Ensemble methods
-- Batch processing
-- Data validation across multiple sources
-- Hyperparameter tuning
-
-Note that when implementing the fan-in step, you'll need to use the ZenML Client to query the results from previous parallel steps, as shown in the example above, and you can't pass in the result directly.
+{% hint style="info" %}
```
schustmi, line 286: Dynamic pipelines is still experimental, I wouldn't recommend them to anyone yet. But even once they're ready, don't you think it makes sense to keep the code sample on how to do it with static pipelines as well, just for the users that want to/must use static pipelines for some reason or another?

We can make the fan-in much nicer now even, as in static pipelines we also allow list of input artifacts, which gets rid of this whole code in the combine step. The only real limitation of static pipelines is the fan-out amount is fixed, but in some cases this might not even be a problem for users.

schustmi (changes requested):

PR #4300, opened 2025-12-04
Enable dynamic pipelines to be deployed

stefannica:
## Describe changes
Enable dynamic pipelines to be deployed.

This PR makes some minor changes to the dynamic pipeline runner to allow it to be used with deployed pipelines:

* disable init and cleanup hooks if the pipeline run is a deployment invocation (because init/cleanup is run globally by the deployment service)
* disable setting up the logging context if the pipeline run is a deployment invocation (logging is set up by the deployment service)_
* use the orchestrator passed as constructor argument instead of that from the active stack. This is required because deployment invocations use a locally instantiated local orchestrator and ignore the one in the active stack

IMPORTANT: the fact that the deployment uses a locally instantiated local orchestrator to run the pipelines isn't respected throughout the code. There are still a lot of occurrences where the orchestrator in the active stack is used instead.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
```diff
@@ -486,12 +486,7 @@ def to_model(
             runnable = True
 
         deployable = False
-        if (
-            not self.is_dynamic
-            and self.build
-            and self.stack
-            and self.stack.has_deployer
-        ):
+        if self.stack and self.stack.has_deployer:
```
schustmi: I'm wondering whether we should keep the `self.build` in here and make this a boolean which mean "can be deployed from the server"?

stefannica: done

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -256,13 +278,24 @@ def launch_step(
             The step run outputs or a future for the step run outputs.
         """
         step = step.copy()
+        step_config = None
+        if self._run and self._run.triggered_by_deployment:
+            # Deployment-specific step overrides
+            step_config = StepConfigurationUpdate(
```
schustmi, line 284: One thing I forgot: We should also set `runtime="inline"` here for deployments, otherwise the steps might still launch in a remote environment using the orchestrator.

PR #4301, opened 2025-12-04
Bug/4295 improve schedule state consistency

Json-Andriopoulos:
## Describe changes

Improves the behavior of Schedule management and the data consistency between database & orchestrator:

1. Deletes dangling DB schedule records in case of failed submission (e.g. invalid cron expression etc.)
3. Enables users to delete records of inconsistent state (users could not delete DB records of schedules if the equivalent records are not available in the orchestrator backend).

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
+++ b/src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
@@ -30,6 +30,7 @@
 # inspired by the Kubernetes dag runner implementation of tfx
 """Kubernetes-native orchestrator."""
 
+import json
 import os
 import random
 import socket
@@ -82,7 +83,11 @@ from zenml.integrations.kubernetes.pod_settings import KubernetesPodSettings
 from zenml.logger import get_logger
 from zenml.metadata.metadata_types import MetadataType
 from zenml.models.v2.core.schedule import ScheduleUpdate
-from zenml.orchestrators import ContainerizedOrchestrator, SubmissionResult
+from zenml.orchestrators import (
+    ContainerizedOrchestrator,
+    PipelineSubmissionError,
+    SubmissionResult,
+)
 from zenml.stack import StackValidator
 
 if TYPE_CHECKING:
@@ -469,13 +474,15 @@ class KubernetesOrchestrator(ContainerizedOrchestrator):
             run_id=placeholder_run.id if placeholder_run else None,
         )
 
-        return self._submit_orchestrator_job(
+        result = self._submit_orchestrator_job(
             snapshot=snapshot,
             command=command,
             args=args,
             environment=base_environment,
             placeholder_run=placeholder_run,
         )
+
+        return result
 
     def submit_dynamic_pipeline(
         self,
@@ -721,6 +728,8 @@ class KubernetesOrchestrator(ContainerizedOrchestrator):
 
         Raises:
             RuntimeError: If a schedule without cron expression is given.
+            PipelineSubmissionError: If an APIException happens during job submission,
+                it is propagated as custom error.
 
         Returns:
             Optional submission result.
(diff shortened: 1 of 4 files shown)


htahir1:
@claude full-review

src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py
```diff
@@ -1216,10 +1217,25 @@ def delete_schedule(self, schedule: "ScheduleResponse") -> None:
         cron_job_name = schedule.run_metadata.get(
             KUBERNETES_CRON_JOB_METADATA_KEY
         )
+
         if not cron_job_name:
-            raise RuntimeError("Unable to find cron job name for schedule.")
+            logger.warning(
+                "Unable to find cron job name %s for schedule %.",
```
schustmi:
```suggestion
                "Unable to find cron job %s for schedule %.",
```

src/zenml/utils/time_utils.py
```diff
@@ -136,3 +138,39 @@ def expires_in(
     if expires_at < now:
         return expired_str
     return seconds_to_human_readable(int((expires_at - now).total_seconds()))
+
+
+def validate_cron_expression(
```
schustmi: I remember that some of the orchestrators we use support a custom cron syntax (or at least supported). This is probably something that we have to take into account here?

pyproject.toml
```diff
@@ -65,6 +65,7 @@ local = [
     "sqlalchemy>=2.0.0,<3.0.0",
     "sqlalchemy_utils",
     "sqlmodel==0.0.18",
+    "croniter>=6.0.0",
```
schustmi: Also this adds two more dependencies (`cronitor` and `humanize`) to ZenML, I'm not sure it's worth it for this check?

Wouldn't the following implementation achieve the same thing:
- Create the schedule in our DB (we already do this)
- Pass it to the orchestrator to create the schedule (we already do this)
- If the orchestrator fails to create the schedule (they already do some validation here), delete it from our DB (we don't do this yet)

Json-Andriopoulos: *Most probably* I will need croniter either way for the following (if implemented):

1. To validate frequency (if we apply a validation on cron frequency).
2. To calculate next occurrence (native scheduling).

Json-Andriopoulos: I like the idea of the alternative algorithm it aligns also with your comment regarding custom cron systax.

schustmi: For native scheduling, we can at least keep the dependency server-side only which I think is much nicer as it doesn't affect client envs and docker images built for pipeline execution.

src/zenml/utils/time_utils.py
stefannica: Correct, Sagemaker.

stefannica: I believe in Sagemaker's case though, we should make a conversion from what should be considered a unified format instead of forcing the user to respect Sagemaker's format.

schustmi (changes requested):
One general question on this: Instead of raising the `PipelineSubmissionError` manually (and only in the Kubernetes orchestrator right now), can we also handle this in the `BaseOrchestrator.run(...)` method more elegantly?

For example: If anything happens in the `Orchestrator.submit_pipeline(...)` method call, that essentially is a submission error and we can delete the schedule in that case? Or do you think that will run into problems with orchestrators that do actually submit the schedule, but then potentially raise an exception before `submit_pipeline(...)` returns?

Json-Andriopoulos:
> One general question on this: Instead of raising the `PipelineSubmissionError` manually (and only in the Kubernetes orchestrator right now), can we also handle this in the `BaseOrchestrator.run(...)` method more elegantly?
> 
> For example: If anything happens in the `Orchestrator.submit_pipeline(...)` method call, that essentially is a submission error and we can delete the schedule in that case? Or do you think that will run into problems with orchestrators that do actually submit the schedule, but then potentially raise an exception before `submit_pipeline(...)` returns?

Yeap, I would want to avoid the scenario of deleting the schedule but actually having the cron submitted. In general in such cases I prefer custom exceptions and explicit handling as it can allow me to be more future-proof and avoid false positives.

PR #4309, opened 2025-12-05
Log store improvements

stefannica:
## Describe changes
Adds some improvements to the existing log store implementations:

1. introduces a generic OTEL log store flavor that can be connected to any OTEL/HTTP/JSON compatible log intake endpoint to store logs, but has not log fetching capabilities (picture shows logs emitted by zenml and stored in dash0).
image
image

2. introduces the Emitter as the primary point of interfacing with the log stores to emit logs.
3. log metadata can be attached to the emitter instead of attaching it to every log.
4. implements an asynchronous log flushing system. This makes **deployed pipelines runs complete up to 15 times faster** from the POV of the clients, as they don't have to wait for the logs to be synchronized before the deployment server responds to an invocation request:
  * before (1 run, GCS log store):
    ```
    $ time curl -X POST http://127.0.0.1:8000/invoke -H "Content-Type: application/json" -d '{"parameters": {}}' 
    real	0m5.390s
    user	0m0.002s
    sys	0m0.005s
    ```
  * after (1 run, GCS log store):
  ```
    $ time curl -X POST http://127.0.0.1:8000/invoke -H "Content-Type: application/json" -d '{"parameters": {}}' 
    real	0m1.970s
    user	0m0.004s
    sys	0m0.004s
  ```
  * before (20 parallel runs, GCS log store):
  ```
    for i in {1..20}; do curl -X POST http://127.0.0.1:8000/invoke -H "Content-Type: application/json" -d '{"parameters": {}}' & done; time wait
    real	2m21.530s
    user	0m0.079s
    sys	0m0.120s
  ```

  * after (20 parallel runs, GCS log store ):
  ```
    for i in {1..20}; do curl -X POST http://127.0.0.1:8000/invoke -H "Content-Type: application/json" -d '{"parameters": {}}' & done; time wait
    real	0m29.128s
    user	0m0.068s
    sys	0m0.106s
  ```
  * before (20 parallel runs, Datadog log store):
  ```
    for i in {1..20}; do curl -X POST http://127.0.0.1:8000/invoke -H "Content-Type: application/json" -d '{"parameters": {}}' & done; time wait
    real	0m31.448s
    user	0m0.073s
    sys	0m0.119s
  ```

  * after (20 parallel runs, Datadog log store ):
  ```
    for i in {1..20}; do curl -X POST http://127.0.0.1:8000/invoke -H "Content-Type: application/json" -d '{"parameters": {}}' & done; time wait
    real	0m28.801s
    user	0m0.078s
    sys	0m0.107s
  ```
5. keeps track of emiters in the log store to automatically activate the OTEL log store, finalize (merge) log files in the artifact log store and flush logs when all emitters are done.
6. implements a Datadog specific exporter that solves the problem of mapping the timestamp and severity fields of OTEL specific log entries to be compatible with Datadog
7. re-structures some of the existing log metadata (attributes) into logical hierarchies, for better readability (see screenshot below)
image
8. add missing log store CLI commands

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/log_stores/otel/otel_log_store.py
+++ b/src/zenml/log_stores/otel/otel_log_store.py
@@ -230,7 +230,7 @@ class OtelLogStore(BaseLogStore):
         self,
         origin: BaseLogStoreOrigin,
         record: logging.LogRecord,
-        metadata: Dict[str, Any],
+        metadata: Optional[Dict[str, Any]] = None,
     ) -> None:
         """Process a log record by sending to OpenTelemetry.
 
@@ -251,7 +251,9 @@ class OtelLogStore(BaseLogStore):
                 raise RuntimeError("OpenTelemetry provider is not initialized")
 
             emit_kwargs = self._handler._translate(record)
-            emit_kwargs["attributes"].update(metadata)
+            emit_kwargs["attributes"].update(origin.metadata)
+            if metadata:
+                emit_kwargs["attributes"].update(metadata)
 
             origin.logger.emit(**emit_kwargs)
 
--- a/src/zenml/utils/logging_utils.py
+++ b/src/zenml/utils/logging_utils.py
@@ -162,7 +162,8 @@ class LoggingContext:
                 if message and message.strip():
                     if context._origin:
                         context._log_store.emit(
-                            context._origin, record, context._metadata
+                            context._origin,
+                            record,
                         )
             except Exception:
                 logger.debug("Failed to emit log record", exc_info=True)
--- a/src/zenml/log_stores/base_log_store.py
+++ b/src/zenml/log_stores/base_log_store.py
@@ -85,6 +85,15 @@ class BaseLogStoreOrigin:
         """
         return self._log_model
 
+    @property
+    def metadata(self) -> Dict[str, Any]:
+        """The metadata associated with the origin.
+
+        Returns:
+            The metadata.
+        """
+        return self._metadata
+
     def deregister(self) -> None:
         """Deregister the origin from the log store."""
         self._log_store.deregister_origin(self)
@@ -164,7 +173,7 @@ class BaseLogStore(StackComponent, ABC):
         self,
         origin: BaseLogStoreOrigin,
         record: logging.LogRecord,
-        metadata: Dict[str, Any],
+        metadata: Optional[Dict[str, Any]] = None,
     ) -> None:
         """Process a log record from the logging system.
 


src/zenml/log_stores/datadog/datadog_log_store.py
```diff
@@ -125,7 +136,7 @@ def fetch(
                     "page": {
                         "limit": page_limit,
                     },
-                    "sort": "@otel.timestamp",
+                    "sort": "timestamp",
```
bcdurak, line 139: We have changed this on purpose because of the sorting. I know the exporter has changed. Is this handled properly now?

src/zenml/log_stores/otel/otel_log_store.py
```diff
@@ -217,6 +327,10 @@ def fetch(
...
             limit: Maximum number of log entries to return.
 
-        Returns:
-            List of log entries from the backend.
+        Raises:
+            NotImplementedError: Log fetching is not supported by the OTEL log
+                store.
         """
+        raise NotImplementedError(
+            "Log fetching is not supported by the OTEL log store."
```
bcdurak, line 337: Would be nice to have a clearer description here. Maybe we can add a docs link about how they can extend the otellogstore to add the fetching functionality in the future.

src/zenml/log_stores/otel/otel_flavor.py
```diff
@@ -60,3 +80,95 @@ class OtelLogStoreConfig(BaseLogStoreConfig):
...
+        return self.generate_default_sdk_docs_url()
+
+    @property
+    def logo_url(self) -> str:
+        """URL to the flavor logo.
+
+        Returns:
+            The URL to the flavor logo.
+        """
+        return "https://public-flavor-logos.s3.eu-central-1.amazonaws.com/log_store/otel.png"
```
bcdurak, line 145: Just leaving a small comment here, as I am not sure if you have uploaded this already. But, I was able to open up the datadog icon but not the otel one.

src/zenml/log_stores/datadog/datadog_log_exporter.py
```diff
@@ -0,0 +1,87 @@
...
+
+class DatadogLogExporter(OTLPLogExporter):
+    """Datadog log exporter.
+
+    This exporter writes OpenTelemetry logs to Datadog's HTTP intake API
+    with slightly modified log records to adapt to Datadog's format.
+    """
+
+    @classmethod
+    def _encode_log(cls, log_data: "LogData") -> Dict[str, Any]:
```
bcdurak, line 37: Just to confirm it here, this solves the problem with the timestamps and severity info, right? Unfortunately, I can not test this part myself.

src/zenml/log_stores/datadog/datadog_log_store.py
```diff
@@ -48,18 +48,29 @@ def config(self) -> DatadogLogStoreConfig:
         """
         return cast(DatadogLogStoreConfig, self._config)
 
-    def get_exporter(self) -> OTLPLogExporter:
+    def get_exporter(self) -> DatadogLogExporter:
         """Get the Datadog log exporter.
 
         Returns:
             OTLPLogExporter configured with API key and site.
```
bcdurak:
```suggestion
            DatadogExporter with the proper configuration.
```

src/zenml/log_stores/datadog/datadog_log_exporter.py
```diff
@@ -0,0 +1,87 @@
...
+
+DEFAULT_TIMEOUT = 10
+
+if TYPE_CHECKING:
+    from opentelemetry.sdk._logs import LogData
+
+logger = get_logger(__name__)
+
+
+class DatadogLogExporter(OTLPLogExporter):
```
bcdurak, line 29: The hierarchical relationship sounds a bit off to me. As far as I know, when you are dealing with an OTLPExporter, you also need to conform to the standard payload structure of OTLP, right? If so, technically, the datadog log exporter isn't really an OTLP log exporter anymore. Do you think it would make sense to separate the HTTP exporter to a base class and implement a general OTLP exporter and a separate Datadog exporter that both inherit from this base class?

src/zenml/utils/logging_utils.py
```diff
@@ -360,12 +377,14 @@ def setup_run_logging(
     """
     log_metadata = get_run_log_metadata(pipeline_run)
     log_metadata.update(dict(source=source))
+    name = f"zenml.pipeline_run.{pipeline_run.id}.{source}"
```
bcdurak: We briefly talked about this before, but currently, there is a unique check with an entity id and its `source`. That means, that I can not have two log responses within our DB that belong to the same pipeline run with the same source such as `client`. But if we ever change this behaviour, this name might become a problem. As far as I can see, you are using this as a key while registering emitters. How about we switch the source with the `log_model.id`?

src/zenml/log_stores/datadog/datadog_log_store.py
stefannica: yes.  the exported logs have the Datadog timestamp correctly set now, so we can use it when we fetch the entries

src/zenml/log_stores/otel/otel_log_store.py
stefannica: This would show up in the UI. I don't think users care about implementing their own log store at that point. We can at most tell them to use another log store that does support fetching.

src/zenml/log_stores/otel/otel_flavor.py
stefannica: done

src/zenml/log_stores/datadog/datadog_log_exporter.py
stefannica: yes, it does

src/zenml/utils/logging_utils.py
stefannica: the log model is a concept that is internal to ZenML. I thought this name should be unique but at the same time representative enough from the user's POV to be used as a grouping mechanism. 

In the end it doesn't really matter what name we use here, so I wouldn't worry too much about key changes in the future.

stefannica: I switched the pipeline run ID for the pipeline run name

PR #4316, opened 2025-12-08
Fix `google_adk` and `crewai` examples' OTEL version conflict

strickvl:
## Summary

Three changes to fix agent examples failing due to OpenTelemetry version conflicts:

1. **Pin google-adk to ≤1.14.1** - Versions 1.15.0+ cap `opentelemetry-sdk` to ≤1.37.0, conflicting with ZenML's ==1.38.0 requirement
2. **Pin crewai to ≤1.6.1** - Version 1.7.0 caps `opentelemetry-sdk` to ~=1.34.0 (meaning <1.35.0), conflicting with ZenML's ==1.38.0 requirement
3. **Combine dependency installs in CI script** - Ensures uv sees all constraints together and fails fast on conflicts

## Problem

ZenML's new log store uses `Logger.emit(body=..., attributes=...)` which requires OpenTelemetry SDK 1.38.0+. However:

| Package | OTEL Constraint | 
|---------|-----------------|
| zenml | `==1.38.0` |
| google-adk 1.15.0+ | `<=1.37.0,>=1.37.0` (exactly 1.37) |
| google-adk ≤1.14.1 | `<2.0.0,>=1.31.0` (flexible) ✅ |
| crewai 1.7.0 | `~=1.34.0` (meaning >=1.34.0, <1.35.0) |
| crewai ≤1.6.1 | `>=1.30.0` (flexible) ✅ |

The previous CI script did two separate installs, allowing the second to silently downgrade OTEL without error.

## Solution

1. **google_adk/requirements.txt**: Pin `google-adk>=1.6.1,<=1.14.1` until Google relaxes their OTEL constraint
2. **crewai/requirements.txt**: Pin `crewai>=1.2.0,<=1.6.1` until crewai relaxes their OTEL constraint
3. **run-agent-examples.sh**: Combine into single `uv pip install -e . -r requirements.txt` so conflicts are caught at install time

## Test plan

- [ ] Verify google_adk example installs successfully with combined deps
- [ ] Verify google_adk example runs without OTEL-related errors
- [ ] Verify crewai example installs successfully with combined deps
- [ ] Verify crewai example runs without OTEL-related errors
- [ ] Verify other agent examples still pass


examples/agent_framework_integrations/google_adk/requirements.txt
```diff
@@ -2,3 +2,6 @@ google-adk>=1.6.1
 zenml[local]>=0.84.3
 jinja2>=3.1.0
 secure>=0.3.0
+# Pin OTEL SDK to ensure Logger.emit(body=..., attributes=...) API is available.
+# ZenML's log store finalize() requires this API introduced in 1.38.0.
+opentelemetry-sdk>=1.38.0
```
schustmi: Shouldn't `pip/uv` figure this out if we install this requirements file, which also installs ZenML?

Or do we somehow not pin that in our zenml dependencies?

stefannica: zenml already pins opentelemetry-sdk to 1.38.0, so this should't be needed here

PR #4339, opened 2025-12-12
Schedule archiving support

Json-Andriopoulos:
## Describe changes

Introduces archiving functionality for schedules - archiving is a soft deletion operation that permanently deactivates a schedule while preserving schedule references for historical data. Therefore, when users delete schedules, pipeline runs triggered by the schedule do not lose this associated information.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/client.py
+++ b/src/zenml/client.py
@@ -4528,7 +4528,7 @@ class Client(metaclass=ClientMetaClass):
         catchup: Optional[Union[str, bool]] = None,
         hydrate: bool = False,
         run_once_start_time: Optional[Union[datetime, str]] = None,
-        is_archived: bool | None = None,
+        is_archived: bool = False,
     ) -> Page[ScheduleResponse]:
         """List schedules.
 
@@ -4666,6 +4666,7 @@ class Client(metaclass=ClientMetaClass):
         self,
         name_id_or_prefix: Union[str, UUID],
         project: Optional[Union[str, UUID]] = None,
+        soft: bool = True,
     ) -> None:
         """Delete a schedule.
 
@@ -4673,6 +4674,7 @@ class Client(metaclass=ClientMetaClass):
             name_id_or_prefix: The name, id or prefix id of the schedule
                 to delete.
             project: The project name/ID to filter by.
+            soft: If set to true, archives the schedule. If false, hard deletes it.
         """
         schedule = self.get_schedule(
             name_id_or_prefix=name_id_or_prefix,
@@ -4694,7 +4696,13 @@ class Client(metaclass=ClientMetaClass):
         else:
             orchestrator.delete_schedule(schedule)
 
-        self.zen_store.delete_schedule(schedule_id=schedule.id, soft=True)
+        if not soft:
+            logger.warning(
+                "You are deleting a schedule with option soft set to False. "
+                "All historical data and references to this schedule will be deleted."
+            )
+
+        self.zen_store.delete_schedule(schedule_id=schedule.id, soft=soft)
 
     # ----------------------------- Pipeline runs ------------------------------
 
--- a/src/zenml/models/v2/core/schedule.py
+++ b/src/zenml/models/v2/core/schedule.py
@@ -312,14 +312,6 @@ class ScheduleResponse(
         """
         return self.get_body().is_archived
 
-    def __bool__(self) -> bool:
-        """Boolean validation of the object.
-
-        Returns:
-            `True` if the schedule is not archived, `False` otherwise.
-        """
-        return not self.is_archived
-
 
 # ------------------ Filter Model ------------------
 
@@ -370,7 +362,7 @@ class ScheduleFilter(ProjectScopedFilter):
         union_mode="left_to_right",
     )
 
-    is_archived: bool | None = Field(
-        default=None,
+    is_archived: bool = Field(
+        default=False,
         description="Whether or not the schedule is archived",
     )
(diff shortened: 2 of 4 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -7376,6 +7376,11 @@ def update_schedule(
                 session=session,
             )
 
+            if existing_schedule.is_archived:
+                raise IllegalOperationError(
+                    "Archived schedules can not be updated."
+                )
+
```
stefannica, line 7386: Very clever ! I would have missed this.

stefannica (approved):

src/zenml/models/v2/core/schedule.py
```diff
@@ -302,6 +303,23 @@ def run_metadata(self) -> Dict[str, MetadataType]:
...
+    @property
+    def is_archived(self) -> bool:
+        """The `is_archived` property.
+
+        Returns:
+            the value of the property.
+        """
+        return self.get_body().is_archived
+
+    def __bool__(self) -> bool:
```
schustmi: Can you explain the reasoning for this? Why would a response return false on a boolean check if it exists?

src/zenml/models/v2/core/schedule.py
```diff
@@ -351,3 +369,8 @@ class ScheduleFilter(ProjectScopedFilter):
         description="The time at which the schedule should run once",
         union_mode="left_to_right",
     )
+
+    is_archived: bool | None = Field(
```
schustmi: Shouldn't this default to `False` somewhere? Currently, archiving does nothing really right? They still show up when listing schedules, which at least to me seems unexpected?

src/zenml/client.py
```diff
@@ -4691,7 +4694,7 @@ def delete_schedule(
         else:
             orchestrator.delete_schedule(schedule)
 
-        self.zen_store.delete_schedule(schedule_id=schedule.id)
+        self.zen_store.delete_schedule(schedule_id=schedule.id, soft=True)
```
schustmi: Shouldn't this at least be an option for users to have a hard delete here, but we default to a soft one?

src/zenml/zen_server/routers/schedule_endpoints.py
```diff
@@ -193,15 +193,18 @@ def update_schedule(
...
     schedule_id: UUID,
+    soft: bool,
     _: AuthContext = Security(authorize),
 ) -> None:
     """Deletes a specific schedule using its unique id.
 
     Args:
         schedule_id: ID of the schedule to delete.
+        soft: Soft deletion will archive the schedule.
     """
```
schustmi, line 204: Do you think we need different RBAC permissions for archiving and hard deletion?

Json-Andriopoulos: Nope, no need to complicate it that much.

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: It exists but is archived. So when you do `if snapshot.schedule` it covers both the cases of being null or archived. Otherwise I would need to replace all schedule checks with `if schedule is not None and not schedule.is_archived` and I might miss some cases if they are not str8forward.

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: It is set to False by default on the DB Schema level. The reasoning is you can not really create something that is archived so you can only update the value.

Json-Andriopoulos: Not sure what you mean with `Currently, archiving does nothing really right` . Once archived you can still see the schedule but the actual cron job is deleted and you can not use or update the archived schedule.

src/zenml/client.py
Json-Andriopoulos: Done, also exposed on the cli

src/zenml/models/v2/core/schedule.py
schustmi: Agreed you shouldn't be able to create archived objects. But this question here is around the default value for the filter model, which is related to listing schedules.

When I think of any tool that supports archiving, I'd imagine the following: I archive something, and then it doesn't show up in any lists anymore, unless I explicitly want to see archived objects. Which means both in the list received by the client, in the CLI, in the frontend (once we add it). Which seems to me that this should be a server-side default, if someone fetches a list of schedules that archived schedules are not included in that list?

src/zenml/models/v2/core/schedule.py
schustmi: This feels very hacky to me, we should simply check `if schedule is None` instead in places where we care. Or if that is to error-prone, add a boolean property `has_active_schedule` or something along those lines.

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: Gotcha, yes that makes sense will add it as the default value for Schedule filters.

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: Cool, removed. Checks are done in the orchestrators which are running right after submission - I think we can skip the check on this use case (does it make sense to check whether the schedule is archived at this point?) . Any other meaningful check uses the .active property so we can remove the __bool__ method safely.

schustmi: Yes those orchestrator methods should only be called right after the schedule was created (at least for the orchestrator schedules that we currently have), so no need to check for archived there I think.

PR #4340, opened 2025-12-12
Add parallel deletion support to artifact prune CLI

strickvl:
## Summary

This PR adds parallel deletion support to the `zenml artifact prune` command, significantly improving performance when pruning large numbers of unused artifacts.

**CLI Changes:**
- Adds `--threads` / `-t` option to control parallelism (default: 1 for backwards compatibility)
- Uses `ThreadPoolExecutor` with bounded in-flight submissions to avoid memory issues
- Thread-local `Client` instances ensure thread safety
- Post-pass artifact cleanup checks server for remaining versions (more correct than relying on cached data)
- Adds mutual exclusivity validation for `--only-artifact` and `--only-metadata` flags
- Tracks failed deletions and provides accurate completion message

**Client Changes:**
- Adds `_skip_unused_check` parameter to `delete_artifact_version()` and `_delete_artifact_version()`
- This bypasses the redundant `depaginate(list_artifact_versions, only_unused=True)` check that was being called for every single deletion
- Fixes pagination race condition when deleting in parallel (the total item count changes while pages are being fetched, causing "Invalid page X" errors)
- Also improves single-threaded performance by avoiding N full pagination queries for N deletions

## Test plan

- [x] Manual testing with `zenml artifact prune --threads 15` on a server with 170 unused artifacts
- [ ] Verify `--only-artifact` and `--only-metadata` together produces an error
- [ ] Verify `--ignore-errors` mode reports failure count accurately
- [ ] Verify single-threaded mode (default) still works as before

### DIFF
--- a/src/zenml/config/compiler.py
+++ b/src/zenml/config/compiler.py
@@ -505,6 +505,7 @@ class Compiler:
             step_runtime = get_step_runtime(
                 step_config=step_config,
                 pipeline_docker_settings=pipeline.configuration.docker_settings,
+                orchestrator=stack.orchestrator,
             )
             if step_runtime == StepRuntime.ISOLATED:
                 # dynamic pipelines & isolated execution
--- a/src/zenml/deployers/utils.py
+++ b/src/zenml/deployers/utils.py
@@ -31,7 +31,7 @@ from zenml.deployers.exceptions import (
     DeploymentNotFoundError,
     DeploymentProvisionError,
 )
-from zenml.enums import DeploymentStatus
+from zenml.enums import DeploymentStatus, StepRuntime
 from zenml.models import (
     CodeReferenceRequest,
     DeploymentResponse,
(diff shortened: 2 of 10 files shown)


src/zenml/client.py
```diff
@@ -5404,38 +5405,87 @@ def delete_artifact_version(
...
+                match the project used when fetching the artifact version.
 
         Raises:
             ValueError: If the artifact version is still used in any runs.
         """
-        if artifact_version not in depaginate(
-            self.list_artifact_versions, only_unused=True
-        ):
+        # Query for this specific version with only_unused=True filter
+        # If the version is used, it won't appear in the results
```
strickvl:
```suggestion
```

src/zenml/cli/artifact.py
```diff
@@ -328,27 +441,113 @@ def prune_artifacts(
...
+                    max_in_flight=threads,
+                    delete_metadata=delete_metadata,
+                    delete_from_artifact_store=delete_from_artifact_store,
                 )
-            else:
-                cli_utils.error(
-                    f"Failed to delete artifact version {unused_artifact_version.id}: {str(e)}"
+
+    # If fail-fast was triggered, abort before artifact cleanup to prevent
+    # any further destructive operations after an unexpected failure
```
strickvl:
```suggestion
```

strickvl (commented):
Self-review

strickvl:
I'll reopen this later...

PR #4358, opened 2025-12-16
Add mydumper/myloader DB backup and restore support

stefannica:
## Describe changes
Implements a mydumper/myloader DB backup strategy.

The mydumper/myloader DB backup strategy is far superior in terms of speed and used storage space:

* **30x** faster database backup times
* **2.5x** faster database restore times
* **10x** lower storage space required

Example:

* the file dump strategy:

```
$ time zenml backup-database -s dump-file
Database was backed up to /home/silent-assassin/.config/zenml/database_backup/zenml-backup.json.

real	3m38.208s
user	0m22.391s
sys	0m1.246s

$ du -sh /home/silent-assassin/.config/zenml/database_backup/zenml-backup.json
794M	/home/silent-assassin/.config/zenml/database_backup/zenml-backup.json

$ time zenml restore-database -s dump-file
Dropping database 'zenml'
Creating database 'zenml'
Database was restored from /home/silent-assassin/.config/zenml/database_backup/zenml-backup.json.

real	2m0.481s
user	1m22.547s
sys	0m0.997s
```

* vs. the new mydumper strategy (with compression enabled):

```
$ time zenml backup-database -s mydumper
Starting mydumper backup of database zenml to /home/silent-assassin/.config/zenml/database_backup/zenml-backup
Database zenml successfully backed up to /home/silent-assassin/.config/zenml/database_backup/zenml-backup
Database was backed up to /home/silent-assassin/.config/zenml/database_backup/zenml-backup.

real	0m7.041s
user	0m13.646s
sys	0m0.395s

$ du -sh /home/silent-assassin/.config/zenml/database_backup/zenml-backup
82M	/home/silent-assassin/.config/zenml/database_backup/zenml-backup

$ time zenml restore-database -s mydumper
Dropping database 'zenml'
Creating database 'zenml'
Starting myloader restore of database zenml from /home/silent-assassin/.config/zenml/database_backup/zenml-backup
Database zenml successfully restored from /home/silent-assassin/.config/zenml/database_backup/zenml-backup
Database was restored from /home/silent-assassin/.config/zenml/database_backup/zenml-backup.

real	0m48.295s
user	0m4.312s
sys	0m0.475s
```

Other related changes:
* if a previous backup exists, it will NOT be deleted by the backup mechanism anymore. It will instead be used as the new baseline to which to restore in case of a migration failure. This is necessary to avoid overwriting a healthy backup made during a previous migration attempt with a database state that was potentially already broken by the previous failed migration.
* refactored the existing database backup code to be cleaner, extensible and maintainable
* allows implementing and configuring a custom database backup/restore strategy
* extends the helm chart to support the mydumper/myloader and custom backup strategies

## Implementation details

The previous `MigrationUtils` class that held all the code required for all the available database backup strategies has been reorganized into a hierarchy of classes:

* the `BaseDatabaseBackupEngine` class defines the database backup/restore engine abstraction. It contains several common utilities such as database creation/deletion and the `backup_database` , `restore_database` and `cleanup_database_backup` abstract methods that every backup engine has to implement.
* the `SQLAlchemyDatabaseBackupEngine` is a base class that uses SQLAlchemy to dump/load data from/to the database. It is used as a base class to implement the existing DB strategies: in-memory, file-dump and database cloning. These are implemented in the `InMemoryDatabaseBackupEngine`, `FileDatabaseBackupEngine` and `DBCloneDatabaseBackupEngine` classes respectively.
* the new mydumper/myloader backup strategy is implemented in its own `MyDumperDatabaseBackupEngine` class that extends directly from `BaseDatabaseBackupEngine`

A lot of code that was spread across the SQLZenStore implementation has been moved to the base backup engine class and its implementations, so that the SQLZenStore class is no longer concerned with the backup/restore functionality.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -1464,7 +1464,12 @@ class SqlZenStore(BaseZenStore):
             )
 
             backup_engine_class = MyDumperDatabaseBackupEngine
+        elif strategy == DatabaseBackupStrategy.DISABLED:
+            from zenml.zen_stores.migrations.backup.base import (
+                DisabledDatabaseBackupEngine,
+            )
 
+            backup_engine_class = DisabledDatabaseBackupEngine
         elif strategy == DatabaseBackupStrategy.CUSTOM:
             custom_backup_engine = self.config.custom_backup_engine
             if custom_backup_engine is None:
--- a/src/zenml/zen_stores/migrations/backup/base.py
+++ b/src/zenml/zen_stores/migrations/backup/base.py
@@ -293,3 +293,49 @@ class BaseDatabaseBackupEngine(ABC):
             self.cleanup_database_backup()
         except Exception:
             logger.exception("Failed to cleanup the database backup.")
+
+
+class DisabledDatabaseBackupEngine(BaseDatabaseBackupEngine):
+    """Database backup engine that is disabled.
+
+    This is used when the backup strategy is set to disabled.
+    """
+
+    def backup_database(
+        self,
+        overwrite: bool = False,
+    ) -> None:
+        """Backup the database.
+
+        Args:
+            overwrite: Whether to overwrite an existing backup if it exists.
+                If set to False, the existing backup will be reused.
+
+        Raises:
+            NotImplementedError: Database backup is not implemented.
+        """
+        raise NotImplementedError("Database backup is disabled.")
+
+    def restore_database(
+        self,
+        cleanup: bool = False,
+    ) -> None:
+        """Restore the database.
+
+        Args:
+            cleanup: Whether to cleanup the backup after restoring the database.
+
+        Raises:
+            NotImplementedError: Database backup is not implemented.
+        """
+        raise NotImplementedError("Database backup is disabled.")
+
+    def cleanup_database_backup(
+        self,
+    ) -> None:
+        """Delete the database backup.
+
+        Raises:
+            NotImplementedError: Database backup is not implemented.
+        """
+        raise NotImplementedError("Database backup is disabled.")


src/zenml/zen_stores/migrations/backup/mydumper.py
```diff
@@ -0,0 +1,311 @@
...
+            if self.config.ssl_cert:
+                args.extend(
+                    ["--cert", self.config.ssl_cert.get_secret_value()]
+                )
+            if self.config.ssl_key:
+                args.extend(["--key", self.config.ssl_key.get_secret_value()])
+
+        return args
+
+    def _get_mysql_env(self) -> Optional[Dict[str, str]]:
```
Json-Andriopoulos: General comment: I think we can start adopting 3.10+ type hints. In this case `dict[str, str] | None`

helm/values.yaml
```diff
@@ -430,11 +430,34 @@ zenml:
...
     # backupDatabase: "zenml_backup"
-
+    #
+    # mydumperThreads: 4
+    # mydumperCompress: true
+    # mydumperExtraArgs: []
+    # myloaderThreads: 4
+    # myloaderExtraArgs: []
+    #
+    # customBackupEngine: my.custom.backup.engine.MyCustomBackupEngine
```
schustmi, line 457: Are these already propagated as environment variables for the sql zen store configuration?

src/zenml/zen_stores/migrations/backup/mydumper.py
```diff
@@ -0,0 +1,311 @@
...
+            RuntimeError: If the backup directory does not exist, if the
+                database name is not set, or if the restore process fails.
+        """
+        if not os.path.isdir(self.backup_location):
+            raise RuntimeError(
+                f"Backup directory `{self.backup_location}` does not exist. "
+                "Please backup the database first."
+            )
+
+        if not self.url.database:
```
Json-Andriopoulos: We do those validations repeatedly (both on dump and restore). Some validations we do on the property level (e.g. on `def backup_location`. It would be good for code clarity to group those validations and execute them once at instantiation? Maybe we could pass them as a pydantic params object and handle any validations there?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -689,8 +701,43 @@ def _validate_backup_strategy(self) -> "SqlZenStoreConfiguration":
...
+            "custom_backup_engine_config",
+            "mydumper_extra_args",
+            "myloader_extra_args",
+        ]:
+            value = data.get(attr)
+            if isinstance(value, str):
+                try:
+                    data[attr] = json.loads(value)
+                except json.JSONDecodeError:
+                    pass
```
schustmi: Is it possible that a string is a valid value here as well? Or should we log a warning at least telling them that the value they passed is invalid?

src/zenml/zen_stores/migrations/backup/mydumper.py
```diff
@@ -0,0 +1,311 @@
...
+            raise RuntimeError(
+                "Database name is required for myloader restore."
+            )
+
+        # Drop and re-create the primary database
+        self.create_database(drop=True)
+
+        cmd = ["myloader"]
```
Json-Andriopoulos: I would move those lines to a helper method. It would make them testable with simple unit test cases and make your core method a bit shorter.

helm/values.yaml
stefannica: No, they were missing. I added them now.

src/zenml/zen_stores/sql_zen_store.py
stefannica: A string will not make sense here. I'll raise an error if the JSON conversion fails.

src/zenml/zen_stores/migrations/backup/mydumper.py
stefannica: Done

src/zenml/zen_stores/sql_zen_store.py
stefannica: Done

src/zenml/zen_stores/migrations/backup/mydumper.py
stefannica: Done

src/zenml/zen_stores/migrations/backup/mydumper.py
stefannica: This is in fact guaranteed to be not null, but had to do something like this to please mypy.

PR #4363, opened 2025-12-16
Support for dynamic pipelines on AzureML

bcdurak:
## Describe changes
Implemented support for dynamic pipelines on AzureML.

## Pre-requisites

## Types of changes


bcdurak:
@schustmi I have tested both static and dynamic pipelines on AzureML with step-specific compute settings. Everything seems to be working 👍

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
```diff
@@ -470,6 +490,154 @@ def _wait_for_completion() -> None:
...
+
+        _wait_for_completion = None
+        if settings.synchronous:
+
+            def _wait_for_completion() -> None:
+                logger.info("Waiting for AzureML job to finish...")
+                ml_client.jobs.stream(job.name)
+
+        return SubmissionResult(
+            metadata=self.compute_metadata(job),
```
schustmi, line 585: This will be a different job (CommandJob) than the one created by a static pipeline (PipelineJob), but I assume the metadata computation works for both?

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
```diff
@@ -205,6 +218,24 @@ def _create_command_component(
             resources=resources,
         )
 
+    def get_azureml_client(self) -> MLClient:
```
schustmi, line 223: Especially for dynamic pipelines with many isolated steps, it's nice if the client is cached instead of being recreated each time a new step is started. I think we can refactor this (similar to `get_job_service_client()` in the vertex orchestrator for example) to only create a new client if the connector expired. WDYT?

schustmi (changes requested):

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
bcdurak: Yes, I have checked both the name and the url. In the static case, the name is the name of the pipeline and the URL opens up the pipeline view, on the other hand, in the dynamic case, the name is the name of the command job and the URL opens up the overview of the job that ran the orchestrator.

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
bcdurak: Good idea, implemented the required changes.

PR #4369, opened 2025-12-18
Improve dynamic pipeline step monitoring

schustmi:
## Describe changes
This PR updates the way isolated steps are executed and monitored in dynamic pipelines.

Previously, it worked as follows: When the user decided to run an isolated step, we would call an orchestrator method that would launch the step and then wait for the step to finish before returning. This meant that the thread was blocked for the entire duration.

With this PR, the submitting of an isolated step is now separated from the monitoring. When an isolated step needs to be run, the orchestrator submits that steps to its backend. After that, the thread is freed up and can be used to launch other steps. The monitoring of all isolated steps happens in a single thread that is just responsible for the monitoring.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/execution/pipeline/dynamic/runner.py
+++ b/src/zenml/execution/pipeline/dynamic/runner.py
@@ -99,7 +99,6 @@ from zenml.utils import (
     string_utils,
 )
 from zenml.utils.logging_utils import (
-    LoggingContext,
     is_pipeline_logging_enabled,
     setup_logging_context,
 )
@@ -435,7 +434,7 @@ class DynamicPipelineRunner:
         logs_context: ContextManager[Any] = nullcontext()
         if is_pipeline_logging_enabled(self._snapshot.pipeline_configuration):
             logs_context = setup_logging_context(source="orchestrator")
-        
+
         with logs_context:
             if self._run.status.is_finished:
                 logger.info("Run `%s` is already finished.", str(self._run.id))
@@ -482,9 +481,11 @@ class DynamicPipelineRunner:
                     # them and raise any exceptions that occurred.
                     self.await_all_step_futures()
                 except Exception as e:
-                    exception_info = exception_utils.collect_exception_information(
-                        exception=e,
-                        user_func=self.pipeline.entrypoint,
+                    exception_info = (
+                        exception_utils.collect_exception_information(
+                            exception=e,
+                            user_func=self.pipeline.entrypoint,
+                        )
                     )
                     # TODO: this call already invalidates the token, so
                     # the steps will keep running but won't be able to
--- a/src/zenml/orchestrators/step_runner.py
+++ b/src/zenml/orchestrators/step_runner.py
@@ -100,22 +100,15 @@ class StepRunner:
         self,
         step: "Step",
         stack: "Stack",
-        publish_exception_info: bool = True,
     ):
         """Initializes the step runner.
 
         Args:
             step: The step to run.
             stack: The stack on which the step should run.
-            publish_exception_info: Whether to publish the exception info for
-                the step run. If set to False, the exception info will be
-                stored in the `publish_utils.step_exception_info` context
-                variable instead, and the caller is responsible for publishing
-                the exception info.
         """
         self._step = step
         self._stack = stack
-        self._publish_exception_info = publish_exception_info
 
     @property
     def configuration(self) -> StepConfiguration:
--- a/src/zenml/orchestrators/step_launcher.py
+++ b/src/zenml/orchestrators/step_launcher.py
@@ -616,9 +616,7 @@ class StepLauncher:
             input_artifacts: The input artifact versions of the current step.
             output_artifact_uris: The output artifact URIs of the current step.
         """
-        runner = StepRunner(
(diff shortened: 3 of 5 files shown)


src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -182,62 +236,235 @@ def pipeline(self) -> "DynamicPipeline":
...
+                if orchestrator_status in [
+                    ExecutionStatus.INITIALIZING,
+                    ExecutionStatus.PROVISIONING,
+                    ExecutionStatus.RUNNING,
+                ]:
+                    # Step still running on the orchestrator side, no need to
+                    # do anything here.
+                    continue
+
+                step_run = Client().get_run_step(step_run.id)
```
bcdurak, line 287: This reminds me of the time when we implemented a special method to understand whether a run was `in_progress` or not, which only fetched the corresponding schema and returned basic fields from it. Considering that this method might be called in succession (if the `get_isolated_step_status` wouldn't work), would it make sense to implement a something similar?

bcdurak: I think the same logic can help the `wait_for_isolated_step` method as a fallback method as well.

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -182,62 +236,235 @@ def pipeline(self) -> "DynamicPipeline":
...
+
+                    if remaining_retries > 0:
+                        step = Step(
+                            spec=step_run.spec,
+                            config=step_run.config,
+                            step_config_overrides=step_run.config,
+                        )
+                        self._queue_concurrent_isolated_step(step=step)
+
+                time.sleep(monitoring_delay)
```
bcdurak, line 391: Should this be within the `for` loop? If I have 10 steps to monitor, by default, we will have to wait 10 seconds here, which is equal to the interval already.

src/zenml/execution/pipeline/dynamic/runner.py
bcdurak: I just realized that my first comment might not be very accurate as you use the resulting `step_run` in different ways as well, I still think a specialized method like that could prove useful.

src/zenml/integrations/azure/orchestrators/azureml_orchestrator.py
```diff
@@ -642,9 +650,43 @@ def run_isolated_step(
         )
 
         job = ml_client.jobs.create_or_update(command_job)
-        logger.info("Waiting for AzureML job `%s` to finish...", job.name)
-        ml_client.jobs.stream(job.name)
-        logger.info("AzureML job `%s` completed.", job.name)
+        publish_step_run_metadata(
```
bcdurak, line 653: I am just going to leave a small comment here. In our case, I don't think it is a big problem, but wanted to mention just in case. There are scenarios where this call might skip the creation of certain metadata without failing (data too large, or unsupported type), in which case, there might be a drift between the `step_run` in the info and in the DB.

src/zenml/orchestrators/step_runner.py
```diff
@@ -259,16 +260,10 @@ def run(
                                 step_exception, step_instance.entrypoint
                             )
                         )
-
-                        if self._publish_exception_info:
```
bcdurak, line 271: I think the `_publish_exception_info` parameter is completely unused right now and can be potentially removed.

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: There are two configurable delays:
- The monitoring delay, which runs in between the monitoring of steps. This is so we don't send too many requests to the orchestrator API, and also our API. This defaults to `0.1`s, so you'll need 100 steps to equal the interval. And even if reaches the interval length, there's nothing wrong with that either.
- The monitoring interval, which roughly dictates how often every step should be monitored. This is the time we wait after monitoring all steps, before starting another round. The wait for this happens a little lower, outside of the for loop.

src/zenml/orchestrators/step_runner.py
schustmi: You're right, I'll remove it. Thanks!

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: Yeah I just checked again and I at least need the duration as well, which makes it so that I can't just use a `GET status` endpoint here. In that case, I don't think it makes sense to implement that as part of this PR, even if it might be a good idea to have this in other places in the future.

bcdurak: Yeah, that's what I figured as well. Resolving this for now, but let's keep the idea maybe we can implement it later.

src/zenml/execution/pipeline/dynamic/runner.py
bcdurak: oh, I must have misread the value up there. I thought the monitoring delay was set to 1 sec, thus my comment. That's why I was a bit surprised to see it within the for loop. Makes sense now.

PR #4372, opened 2025-12-18
Heartbeat - performance improvements

Json-Andriopoulos:
## Describe changes

Improvements to enhance performance of heartbeat operations:

1. Increase polling period of heartbeat workers (reduce API calls)
2. Improve heartbeat update performance (reduce DB calls)
3. Improve in-progress checks for token validity (reduce DB calls)
4. Improve heartbeat_threshold resolution (dehydrated API calls)
5. Disable request cacheing (reduces DB stress by a lot)

Pending:

1. Update and extend tests
2. Update documentation

## Pre-requisites

## Types of changes


src/zenml/pipelines/pipeline_definition.py
```diff
@@ -473,12 +487,16 @@ def configure(
         if merge and secrets and self._configuration.secrets:
             secrets = self._configuration.secrets + list(secrets)
 
+        if not enable_heartbeat:
```
schustmi: When called, this will always set `enable_heartbeat=True`, even if it was set to `False` before, no? We often call this method even internally, so it shouldn't overwrite any value in the configuration unless explicitly set. See the `remove_none_values(...)` call below.

src/zenml/steps/heartbeat.py
```diff
@@ -139,6 +149,7 @@ def _run(self) -> None:
             while self._running:
                 try:
                     self._heartbeat()
+                    self._consecutive_failures = 0
```
schustmi, line 152: Should this also reset the heartbeat interval down to the normal value, or do we keep it at the maximum here?

src/zenml/config/compiler.py
```diff
@@ -491,6 +491,9 @@ def _get_step_spec(
     def _get_heartbeat_flag(
         pipeline: "Pipeline", stack: "Stack", step_config: "StepConfiguration"
     ) -> bool:
+        if not pipeline.enable_heartbeat:
```
schustmi: When running a snapshot through the server, we allow users to overwrite values in the pipeline configuration and then create a new snapshot (=compiled pipeline), but don't use the `Compiler` class to do so. This is because the `Compiler` class expects the pipeline instance for which the code is not available in that environment.

We need to also make sure setting that value in the pipeline configuration gets propagated to the steps correctly. That code would be in `src/zenml/zen_server/pipeline_execution/utils.py::snapshot_request_from_source_snapshot`

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -7179,6 +7179,44 @@ def count_runs(self, filter_model: PipelineRunFilter) -> int:
...
+    def disable_run_heartbeat(self, run_id: UUID) -> None:
+        with Session(self.engine) as session:
+            existing_run = self._get_schema_by_id(
+                resource_id=run_id,
+                schema_class=PipelineRunSchema,
+                session=session,
+            )
+
+            if existing_run.get_pipeline_configuration().enable_heartbeat:
+                raise IllegalOperationError(
```
schustmi: Do we really need to fail here?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -7179,6 +7179,44 @@ def count_runs(self, filter_model: PipelineRunFilter) -> int:
...
+                    f"The heartbeat flag for run {run_id} is already disabled."
+                )
+
+            # update run configuration - new step runs will set heartbeat flags to False
+
+            if existing_run.snapshot:
+                config = json.loads(
```
schustmi: I think we can't really do it like this: This configuration is supposed to be the static configuration at compilation time, and in no place do we update it so far. Even more, you're changing the configuration of the snapshot, which means this also changes behaviour for potentially scheduled pipelines in the future.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -7179,6 +7179,44 @@ def count_runs(self, filter_model: PipelineRunFilter) -> int:
...
+                config = json.loads(existing_run.pipeline_configuration)
+
+            config["enable_heartbeat"] = False
+            existing_run.snapshot.pipeline_configuration = json.dumps(config)
+
+            session.add(existing_run)
+
+            # set heartbeat threshold to null for all created steps
+
+            stmt = (
```
schustmi, line 7065: Is there a potential race condition here if a step is being created just after this statement is happening?

src/zenml/steps/heartbeat.py
Json-Andriopoulos: I would say best not to as it may cause the behaviour again.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Not really. We can silently skip as well.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Yeah, I wasn't really sure about it either. Going to make it a pipeline run attribute I guess.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Could be, good catch. New steps use the equivalent pipeline run to resolve the threshold, we can make sure the pipeline run heartbeat flag is set to false in which case we get rid of the race condition.

src/zenml/pipelines/pipeline_definition.py
Json-Andriopoulos: 🤦

src/zenml/config/compiler.py
Json-Andriopoulos: Should be covered now.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Should be covered now.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Should be covered now.

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -473,12 +487,16 @@ def configure(
         if merge and secrets and self._configuration.secrets:
             secrets = self._configuration.secrets + list(secrets)
 
+        if enable_heartbeat is None:
+            enable_heartbeat = True
```
schustmi: Isn't this the same thing: Anytime the `.configure(...)` method is called without a value, it would reset the heartbeat to `True` no matter the previous value?

What I suggest we do instead: Just allow having this as `None`, and if it's `None` in the configuration the server counts that as heartbeat enabled. wdyt?

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -145,6 +145,9 @@ class PipelineRunRequest(ProjectScopedRequest):
         default=None,
         title="Logs of the pipeline run.",
     )
+    enable_heartbeat: bool = Field(
```
schustmi: Why do we need this in the request separately? Shouldn't this just always pick the value from the configuration that is stored in the snapshot as the default?

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -118,6 +118,7 @@ class PipelineRunSchema(NamedSchema, RunMetadataInterface, table=True):
         sa_column=Column(TEXT, nullable=True)
     )
     index: int = Field(nullable=False)
+    enable_heartbeat: bool | None = Field(nullable=False)
```
```suggestion
    enable_heartbeat: bool = Field(nullable=False)
```

schustmi (changes requested):

Aaditya-ops-cpu:
following up I noticed PipelineRunSchema did not define a enable_heartbeat field. That I think explains the migration issues, since heartbeat behavior is being currently added at the run level without the schema fully supporting it. Since heartbeat is already controlled by config and step settings, it may be safer to handle this at runtime unless the schema is updated to match the changes.

src/zenml/pipelines/pipeline_definition.py
Json-Andriopoulos: Yeap, that makes sense

src/zenml/models/v2/core/pipeline_run.py
Json-Andriopoulos: Good point, resolved.

schustmi (approved):

PR #4379, opened 2026-01-07
Misc fixes

schustmi:
## Describe changes
- Use correct working dir in (dynamic) pipeline container: The working dir was `/zenml` instead of `/app`, which caused issues in cases where users relied on fixed absolute paths where ZenML will download the user code to
- Convert component setting shortcut keys when running snapshots
- Trigger a pipeline run status update even if the status is the same. This fixes the following bug: If a pipeline run is in `CONTINUE_ON_FAILURE` execution mode, the run status will be updated to Failed. If other steps then finish (successfully or not, doesn't matter), they will never trigger a run status update, which means the `PipelineRunSchema.in_progress` value never gets updated and therefore the token doesn't get invalidated.
- We now only set the end time of a run once the run is no longer in progress. Otherwise, the run end time was set to the time when the first step failed, even though other steps continued running.
- The step exception info was not correctly published for isolated steps in dynamic pipelines because the step runner assumed the step launcher would be running in the same environment.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/entrypoints/base_entrypoint_configuration.py
+++ b/src/zenml/entrypoints/base_entrypoint_configuration.py
@@ -241,8 +241,8 @@ class BaseEntrypointConfiguration(ABC):
         snapshot_id = UUID(self.entrypoint_args[SNAPSHOT_ID_OPTION])
         return Client().zen_store.get_snapshot(snapshot_id=snapshot_id)
 
-    def download_code_if_necessary(self) -> None:
-        """Downloads user code if necessary.
+    def prepare_code_environment(self) -> None:
+        """Prepares the code environment.
 
         Raises:
             CustomFlavorImportError: If the artifact store flavor can't be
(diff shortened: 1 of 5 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -10743,9 +10743,7 @@ def _update_pipeline_run_status(
             is_dynamic_pipeline=is_dynamic_pipeline,
         )
 
-        if new_status == pipeline_run.status or (
-            pipeline_run.is_placeholder_run() and not new_status.is_finished
-        ):
+        if pipeline_run.is_placeholder_run() and not new_status.is_finished:
```
bcdurak, line 10746: Great catch over here. Would be nice to remove the next line though to reflect the changes though.

src/zenml/orchestrators/step_runner.py
```diff
@@ -257,17 +260,14 @@ def run(
                             )
                         )
 
-                        if ENV_ZENML_STEP_OPERATOR in os.environ:
```
bcdurak, line 260: With this line being removed, I think we can even remove this env variable, right? We set it in the launcher but it must be unused as far as I can see.

src/zenml/zen_stores/sql_zen_store.py
schustmi: Good catch, I removed it!

src/zenml/orchestrators/step_runner.py
schustmi: I decided to keep it in there as it might still be interesting for users to sometimes detect whether they're running in a step operator environment.

src/zenml/entrypoints/base_entrypoint_configuration.py
```diff
@@ -241,52 +241,63 @@ def _load_snapshot(self) -> "PipelineSnapshotResponse":
...
-        else:
-            raise RuntimeError(
-                "Code download required but no code reference or path provided."
-            )
+        # Change the working directory to make sure we're in the correct
+        # directory where the files in the Docker image should be included.
+        # This is necessary as some services overwrite the working directory
+        # configured in the Docker image itself.
+        os.makedirs("/app", exist_ok=True)
+        os.chdir("/app")
```
stefannica, line 258: This sticks out like a sore thumb because you're making two assumptions here:

1. this code is always running in a container
2. the `/app` folder is configured in the container image as the working directory

What about the rest of the cases where this is not true ?

Also, does this folder need to be created even when downloading the code isn't necessary ?

src/zenml/entrypoints/base_entrypoint_configuration.py
```diff
@@ -241,52 +241,63 @@ def _load_snapshot(self) -> "PipelineSnapshotResponse":
...
+                )
+
+            logger.info("Code download finished.")
 
-        logger.info("Code download finished.")
+        # If the working directory is not in the sys.path, we include it to make
+        # sure user code gets correctly imported.
+        cwd = os.getcwd()
+        if cwd not in sys.path:
+            sys.path.insert(0, cwd)
```
stefannica, line 300: Why is this change needed now ? I.e. how did this work before ? Wasn't the .zen folder used somehow to augment the sys.path ?

src/zenml/entrypoints/base_entrypoint_configuration.py
stefannica: Update: I see now that you merely moved this code here from the step entrypoint, therefore extending the assumptions from the step container (where they were already met) to every other container. So my previous comment can be safely dismissed.

src/zenml/entrypoints/base_entrypoint_configuration.py
stefannica: Disregard this too...

PR #4381, opened 2026-01-08
Improve bot detection for Segment analytics

strickvl:
## Summary
- Adds detection for spoofed browsers with empty `userAgentData` fields (empty `brands` array and empty `platform` string)
- Improves the "Not A Brand" regex pattern to catch more headless browser variations like `Not;A=Brand`

## Problem
We're seeing continued bot traffic from crawlers that:
- Spoof Chrome user-agents (`Chrome/99.0.4844.51` from March 2022)
- Don't properly implement the Client Hints API (return empty `brands: []` and `platform: ""`)
- Systematically crawl old SDK documentation versions
- All originate from `zh-CN` locale / `Asia/Shanghai` timezone

Real browsers **always** populate `userAgentData.brands` and `userAgentData.platform`, so empty values are a strong bot indicator.

## Test plan
- [ ] Deploy to docs.zenml.io and monitor Segment traffic
- [ ] Verify legitimate user traffic is not affected
- [ ] Check for reduction in bot-originated page events

## Note
A separate PR against `gh-pages` will be needed to apply similar fixes to the versioned SDK docs (`sdkdocs.zenml.io`). -- made that one https://github.com/zenml-io/zenml/pull/4382

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -22,6 +22,7 @@ Louvre = "Louvre"
 [default.extend-words]
 # Don't correct the surname "Teh"
 aks = "aks"
+cros = "cros"  # Chrome OS identifier in user agent strings
 hashi = "hashi"
 womens = "womens"
 prepend = "prepend"


docs/mkdocs/overrides/main.html
```diff
@@ -5,61 +5,140 @@
...
+    if (getChromeMajor(userAgent) !== null && hasChromiumBrand && sawEmptyExpectedBrandVersion) {
+      return true;
+    }
+
     return false;
   }
   
   // Only send analytics data if this is NOT bot traffic
   if (!isBotTraffic()) {
+    analytics.load("CaE5VaNjhyaUt48qXK4rLwwKJ0PdAOFM");
```
htahir1, line 141: whats this for?

htahir1 (changes requested):

docs/mkdocs/overrides/main.html
strickvl: This is what sends the data to Segment. It's not an addition. It just moves that line from where it was previously a bit further up in the script.

htahir1 (approved):

PR #4382, opened 2026-01-08
Improve bot detection for Segment analytics on SDK docs

strickvl:
## Summary

Updates the Segment analytics gating logic across all versioned SDK documentation (apidocs.zenml.io / sdkdocs.zenml.io) to catch sophisticated bot traffic that was consuming Segment MTU quota.

**Key improvements:**
- Detects bots with empty `userAgentData.brands` + empty `platform` (catches the specific spoofed Chrome traffic we observed)
- Detects platform mismatch between `navigator.userAgent` and `userAgentData.platform`
- Detects GREASE-only brand arrays (only "Not A Brand" variants present without real Chromium brands)
- Detects Chrome UA with empty expected brand versions
- Expanded bot signature list (added phantomjs, slimerjs)

**Scope:**
- 114 versioned SDK doc pages updated (0.11.0 through 0.93.0)
- Versions 0.11.0 - 0.74.0: Upgraded from basic to advanced bot detection
- Versions 0.75.0 - 0.93.0: Had NO bot detection before, now have full protection

## Test plan

- [ ] Verify a sample of updated files have the new `isBotTraffic()` function
- [ ] After merge, spot-check that legitimate browser traffic still triggers Segment (check DevTools → Network for `cdn.segment.com` requests)
- [ ] Monitor Segment dashboard for reduced bot traffic volume

## Related

- PR #3858 - Original bot detection for main docs
- PR #3859 - Original bot detection for SDK docs (this PR improves that)
- PR #4381 - Improved detection for main docs (this PR ports those improvements to SDK docs)


strickvl:
> Aren’t these auto generated ? And if so shouldn’t we fix the script first?

genau yes we fixed it here https://github.com/zenml-io/zenml/pull/4381.

htahir1:
so we fix them so that the older versins get fixed?

strickvl:
> so we fix them so that the older versins get fixed?

Yeah if we don't fix the old versions, then the bots will continue to rack up requests + users which are created on Segment. We also did that previously https://github.com/zenml-io/zenml/pull/3859 but something about a reo.dev key addition removed the fixes from 0.75.0 going forward which is why we have to add the script to those versions, and also we're amending the old script because it was still letting users in.

0.11.0/overrides/main.html
```diff
@@ -5,61 +5,140 @@
...
-  analytics.load("CaE5VaNjhyaUt48qXK4rLwwKJ0PdAOFM");
-  
+  // (Removed line) analytics.load("CaE5VaNjhyaUt48qXK4rLwwKJ0PdAOFM");
```
htahir1, line 8: why do we do this

strickvl: We move it all the way down to the end (see the comment you made on the other PR as well that I answered).

PR #4405, opened 2026-01-15
Fixes around the logging context and more

bcdurak:
## Describe changes
Logging improvements.

- New `create` and `update` endpoints for logs
- New `LogsUpdate` model
- Added support for UUIDs for `StepRunRequest` and `PipelineRunRequest` for existing logs
- Adjusted the CLI message when you try to activate a stack that you don't have permission to
- Added an env variable to manage the max log entries that can be fetched with a single request
- Updated the docs to reflect the limitations of the dashboard
- Fixed the workflows.

for more detailed explanations check the comments below.

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/integration-test-fast.yml
+++ b/.github/workflows/integration-test-fast.yml
@@ -113,22 +113,19 @@ jobs:
       run:
         shell: bash
     steps:
-      - name: Maximize space (/opt)
-        uses: easimon/maximize-build-space@fc881a613ad2a34aca9c9624518214ebc21dfc0c  # v10
-        with:
-          # Python from actions/setup-python lives under /opt/hostedtoolcache on Linux runners,
-          # and this workflow installs ZenML via `--system` into that interpreter.
-          # Mounting the expanded build volume at /opt prevents toolcache + site-packages
-          # installs from exhausting the root filesystem.
-          root-reserve-mb: 30000
-          swap-size-mb: 1024
-          remove-dotnet: 'true'
-          remove-android: 'true'
-          remove-haskell: 'true'
-          remove-codeql: 'true'
-          remove-docker-images: 'true'
-          build-mount-path: /opt
+      - name: Free disk space (Ubuntu)
         if: inputs.os == 'ubuntu-latest'
+        run: |
+          set -euxo pipefail
+          echo "Before cleanup:"
+          df -h
+          sudo rm -rf /usr/share/dotnet || true
+          sudo rm -rf /usr/local/lib/android || true
+          sudo rm -rf /opt/ghc || true
+          sudo rm -rf /opt/hostedtoolcache/CodeQL || true
+          sudo docker image prune --all --force || true
+          echo "After cleanup:"
+          df -h
       - name: Set uv cache + temp dirs (/opt)
         if: inputs.os == 'ubuntu-latest'
         run: |
--- a/.github/workflows/integration-test-slow.yml
+++ b/.github/workflows/integration-test-slow.yml
@@ -111,20 +111,19 @@ jobs:
       run:
         shell: bash
     steps:
-      - name: Maximize space (/opt)
-        uses: easimon/maximize-build-space@fc881a613ad2a34aca9c9624518214ebc21dfc0c  # v10
-        with:
-          # Python from actions/setup-python lives under /opt/hostedtoolcache on Linux runners,
-          # and this workflow installs ZenML via `--system` into that interpreter.
-          root-reserve-mb: 20000
-          swap-size-mb: 1024
-          remove-dotnet: 'true'
-          remove-android: 'true'
-          remove-haskell: 'true'
-          remove-codeql: 'true'
-          remove-docker-images: 'true'
-          build-mount-path: /opt
+      - name: Free disk space (Ubuntu)
         if: inputs.os == 'ubuntu-latest'
+        run: |
+          set -euxo pipefail
+          echo "Before cleanup:"
+          df -h
+          sudo rm -rf /usr/share/dotnet || true
+          sudo rm -rf /usr/local/lib/android || true
+          sudo rm -rf /opt/ghc || true
+          sudo rm -rf /opt/hostedtoolcache/CodeQL || true
+          sudo docker image prune --all --force || true
+          echo "After cleanup:"
+          df -h
       - name: Set uv cache + temp dirs (/opt)
         if: inputs.os == 'ubuntu-latest'
         run: |
--- a/.github/workflows/release_prepare.yml
+++ b/.github/workflows/release_prepare.yml
@@ -232,15 +232,18 @@ jobs:
           - cloud: azure
           - cloud: gcp
     steps:
-      - name: Maximize space for Docker
-        uses: easimon/maximize-build-space@fc881a613ad2a34aca9c9624518214ebc21dfc0c # v10
-        with:
-          root-reserve-mb: 20000
-          swap-size-mb: 1024
-          remove-dotnet: 'true'
-          remove-android: 'true'
-          remove-haskell: 'true'
-          build-mount-path: /var/lib/docker/
+      - name: Free disk space
+        run: |
+          set -euxo pipefail
+          echo "Before cleanup:"
+          df -h
+          sudo rm -rf /usr/share/dotnet || true
+          sudo rm -rf /usr/local/lib/android || true
+          sudo rm -rf /opt/ghc || true
+          sudo rm -rf /opt/hostedtoolcache/CodeQL || true
+          sudo docker image prune --all --force || true
+          echo "After cleanup:"
+          df -h
       - name: Reload Docker
         run: sudo systemctl restart docker
       - name: Checkout code
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -77,20 +77,19 @@ jobs:
       run:
         shell: bash
     steps:
-      - name: Maximize space
-        uses: easimon/maximize-build-space@fc881a613ad2a34aca9c9624518214ebc21dfc0c  # v10
-        with:
-          # Python from actions/setup-python lives under /opt/hostedtoolcache on Linux runners,
-          # and this workflow installs ZenML via `--system` into that interpreter.
-          root-reserve-mb: 8000
-          swap-size-mb: 1024
-          remove-dotnet: 'true'
-          remove-android: 'true'
-          remove-haskell: 'true'
-          remove-codeql: 'true'
-          remove-docker-images: 'true'
-          build-mount-path: /opt
+      - name: Free disk space (Ubuntu)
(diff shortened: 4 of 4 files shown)


src/zenml/models/v2/core/logs.py
```diff
@@ -51,6 +52,14 @@ class LogsRequest(BaseRequest):
         default=None,
         title="The log store ID that collected these logs",
     )
+    pipeline_run_id: Optional[UUID] = Field(
```
schustmi, line 55: So I can now either update a `Log` entry to point it to a pipeline/step run, but I can also do the reverse and pass the ID in a pipeline/step run request. Do we need both?

src/zenml/zen_stores/schemas/logs_schemas.py
```diff
@@ -35,14 +36,6 @@ class LogsSchema(BaseSchema, table=True):
     """SQL Model for logs."""
 
     __tablename__ = "logs"
-    __table_args__ = (
-        UniqueConstraint(
```
schustmi, line 39: Why do we not need this anymore? Is it intended that we create multiple logs for the same source/step for example?

src/zenml/zen_server/routers/logs_endpoints.py
```diff
@@ -40,6 +42,29 @@
...
+) -> LogsResponse:
+    """Create a new log model.
+
+    Args:
+        logs: The log model to create.
+
+    Returns:
+        The created log model.
+    """
+    return verify_permissions_and_create_entity(
```
schustmi, line 103: We currently do not have RBAC permissions for logs, which means anyone will be able to do these calls. Is that something we should add, or how should we handle those?

I guess we could at least add the following checks:
- If the request contains a pipeline run ID, check for `UPDATE` permissions on the run
- If the request contains a pipeline run ID, check for `UPDATE` permissions on the parent run

Same for the update endpoint below.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -6454,45 +6574,69 @@ def _create_run(
...
-            if pipeline_run.logs.artifact_store_id:
-                self._get_reference_schema_by_id(
-                    resource=pipeline_run,
-                    reference_schema=StackComponentSchema,
-                    reference_id=pipeline_run.logs.artifact_store_id,
-                    session=session,
-                    reference_type="logs artifact store",
-                )
+            if isinstance(pipeline_run.logs, UUID):
+                existing_log_entry = session.get(LogsSchema, pipeline_run.logs)
```
schustmi: I think we should also use `_get_reference_schema_by_id` here, to make sure the logs entry is in the same project.

schustmi: Same for the step creation method below

bcdurak: Done.

src/zenml/zen_server/routers/logs_endpoints.py
bcdurak: I guess, you mean the step ID in the second one.

The good thing is if it is a log operation regarding a step, the pipeline id is always present as well, I now use it to do the `UPDATE` check.

Additionally I am adding an additional `READ` check to see whether we can fetch the log store or the artifact store mentioned in the request.

src/zenml/zen_stores/schemas/logs_schemas.py
bcdurak: As discussed, I will keep it like this for the time being.

src/zenml/models/v2/core/logs.py
bcdurak: You can even create it with the ids. I think, they all have uses. For instance:

- with the step runner, when we initiate the context we already have a step run and a pipeline run. In this case, it makes sense to create the log entry with the proper ids already. 
- as for the launcher, we might have a run but the step gets created within the launch method so I add the IDs to the step request.
- finally, for the k8s orchestrator entrypoint, we mught end up getting, updating or creating a run. this is a moment where I use the update functionality.

If it is alright, I would keep it like this.

src/zenml/zen_server/routers/logs_endpoints.py
```diff
@@ -40,6 +47,50 @@
...
+) -> LogsResponse:
+    """Create a new log model.
+
+    Args:
+        logs: The log model to create.
+
+    Returns:
+        The created log model.
+    """
+    if LogsRequest.pipeline_run_id:
```
schustmi:
```suggestion
    if logs.pipeline_run_id:
```

src/zenml/zen_server/routers/logs_endpoints.py
schustmi: While all your current calls of the endpoint include the run ID when a step ID is provided, that doesn't matter at all. Users can simply call this endpoint without a run ID and just a step ID, and then we're missing RBAC checks.

schustmi (changes requested):

src/zenml/zen_server/routers/logs_endpoints.py
bcdurak: Fixed.

src/zenml/zen_stores/schemas/logs_schemas.py
bcdurak: This became an itch that I had to scratch 😅 I implemented this string that we talked about yesterday. I will provide a more detailed explanation down below.

src/zenml/zen_server/routers/logs_endpoints.py
bcdurak: Absolutely right. I have made some adjustments to both the request and the update model. I will provide a more detailed explanation down below.

bcdurak:
@schustmi I have implemented some changes. I thought I would summarize them here.

1. In order to throw a better error than the one that we have seen yesterday, I am now fetching the run and the step if necessary when I am creating a log entry.
2. I have also added two new methods to the sql zen store, namely `_create_logs` and `_update_logs`. These are both used in the original create and update methods for logs but also when we create/update runs and steps. This way I control the create/update behaviour through one method. It also accepts `session` as an input for performance.
3. I have updated the logs models. For requests, if you provide a step run id, you need to provide a run id as well. Similarly, the update model needs to either have just a pipeline id or a combination of a pipeline id or step id. Either way, the pipeline id needs to be present in the update model. I just do not want to make an extra call if someone just provides a step id, then I would have to fetch it to look at the pipeline id to make an RBAC call.
4. The check for the artifact store and the log store has been moved to a validator as well.
5. I have adjusted the RBAC checks as well.
6. I have implemented a `log_key` to serve as a unique constraint for the `logs` table. The format is `pipelineid-stepid-source` if it is associated with an entity. Otherwise, it is null.
7. I adjusted the migration script accordingly as well.

I will still have to test everything before merging.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4404,6 +4405,136 @@ def get_logs(self, logs_id: UUID, hydrate: bool = True) -> LogsResponse:
...
+                    "The logs entry is already associated with a different entity."
+                )
+
+        self._get_schema_by_id(
+            resource_id=logs_update.pipeline_run_id,
+            schema_class=PipelineRunSchema,
+            session=session,
+        )
+        if logs_update.step_run_id:
+            step_schema = self._get_schema_by_id(
```
schustmi: We need to use the reference schema method here in order for the project consistency checks. Same in most other places where you changed it.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -4404,6 +4405,136 @@ def get_logs(self, logs_id: UUID, hydrate: bool = True) -> LogsResponse:
...
+                session=session,
+            )
+        if logs.log_store_id:
+            self._get_schema_by_id(
+                resource_id=logs.log_store_id,
+                schema_class=StackComponentSchema,
+                session=session,
+            )
+        if logs.pipeline_run_id:
+            self._get_schema_by_id(
```
schustmi: When you call this from `_create_run`, you've already created/fetched this run. I think that means this is just an extra unnecessary call?

schustmi: Same for step creation/update

bcdurak:
hey @schustmi, I have included some more updates here:

- The logs entries are now project scoped, meaning they have a user id and a project id field in the DB.
- I have adjusted the DB migration to add these columns and backfill them if possible.
- I have fixed the fetch calls and replaced them back with `_get_reference_...` calls.
- I implemented additional RBAC checks to the endpoints. Most importantly, you can not fetch any logs that are not attached to anything.
- Added another section to the docs that signify the thread/async limitations of ZenML log capturing.
- Added extra validation on the request models. Previously if both a step id and a pipeline id would be given, it would end up being associated to both entities.

I have already run some pipelines (both static and dynamic) remotely, and things seem fine. LMK what you think.

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Addressed this down below.

src/zenml/zen_server/routers/logs_endpoints.py
```diff
@@ -40,6 +48,56 @@
...
+
+    Args:
+        logs: The log model to create.
+
+    Returns:
+        The created log model.
+    """
+    if logs.pipeline_run_id:
+        verify_permission_for_model(
+            model=zen_store().get_run(logs.pipeline_run_id),
```
schustmi: Can be non-hydrated

schustmi: Same for all `get` calls below

src/zenml/zen_stores/migrations/versions/8e36b8227160_remove_logs_unique_constraint.py
```diff
@@ -0,0 +1,130 @@
...
+    connection = op.get_bind()
+
+    # Prefer inferring scope from step runs when available, since a step run
+    # always belongs to a single pipeline run.
+    connection.execute(
+        sa.text("""
+            UPDATE logs
+            SET
+                project_id = (
+                    SELECT project_id
```
schustmi: This has the same sub-query twice. Not sure how MySQL/SQLite handle this, but I assume fetching the steps runs first should be more efficient?

schustmi (changes requested):

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Fixed with a `verify` flag which determines whether we should do the check on the pipeline run / step run or not.

src/zenml/zen_server/routers/logs_endpoints.py
bcdurak: Done.

PR #4416, opened 2026-01-19
Add data model changes for native schedules

Json-Andriopoulos:
## Describe changes

- Extends the data model and operations on ZenML OSS to facilitate native schedules.

Pending:

- [ ] Extend tests

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/migrations/versions/97109a4a8d26_remove_triggers.py
+++ b/src/zenml/zen_stores/migrations/versions/97109a4a8d26_remove_triggers.py
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "97109a4a8d26"
-down_revision = "d1e8c2c4a9ef"
+down_revision = "73c397e3c3ba"
 branch_labels = None
 depends_on = None
 


src/zenml/enums.py
```diff
@@ -576,3 +576,10 @@ class GroupType(StrEnum):
 
     MANUAL = "manual"
     MAP = "map"
+
+
+class ScheduleEngine(StrEnum):
+    """All possible schedule engines."""
+
+    native = "native"
+    orchestrator = "orchestrator"
```
stefannica: we used uppercase so far as the convention for naming enum values

src/zenml/models/v2/core/schedule.py
```diff
@@ -44,13 +53,15 @@ class ScheduleRequest(ProjectScopedRequest):
 
     name: str
     active: bool
+    engine: ScheduleEngine
```
stefannica: would it be possible to set this to ScheduleEngine.ORCHESTRATOR by default for now, to preserve backwards compatibility and make it easier to transition to the new scheduling system ?

src/zenml/models/v2/core/schedule.py
```diff
@@ -44,13 +53,15 @@ class ScheduleRequest(ProjectScopedRequest):
...
 
     cron_expression: Optional[str] = None
     start_time: Optional[datetime] = None
     end_time: Optional[datetime] = None
     interval_second: Optional[timedelta] = None
     catchup: bool = False
     run_once_start_time: Optional[datetime] = None
+    next_occurrence: datetime | None = None
```
stefannica: why would this be set on the schedule request by the user ?

stefannica: update: if I understand it correctly, this field is not actually set by the user, it's computed from the other fields. In this case, it should be either a property or a private field that is not published by FastAPI as a user-facing attribute in the OpenAPI spec.

still, the question remains: if this is not set by the user, why store it as a field here, instead of computing it in the SQL store or wherever it's needed ?

src/zenml/models/v2/core/schedule.py
```diff
@@ -162,6 +223,8 @@ class ScheduleResponseMetadata(ProjectScopedResponseMetadata):
 class ScheduleResponseResources(ProjectScopedResponseResources):
     """Class for all resource models associated with the schedule entity."""
 
+    snapshot: "PipelineSnapshotResponse | None"
```
stefannica: I forgot what we decided, so please forgive my ignorance here. Shouldn't a schedule be associated with multiple snapshots ? In the context of the larger trigger discussion I think it was mentioned that the same schedule/trigger might be configured for several snapshots, because the user might want all of them to be triggered at the same time and to avoid the operational burden of having to configure multiple schedules/triggers for multiple snapshots.

src/zenml/models/v2/core/schedule.py
```diff
@@ -312,13 +375,36 @@ def is_archived(self) -> bool:
         """
         return self.get_body().is_archived
 
+    @property
+    def is_engine_native(self) -> bool:
```
stefannica: if you're exposing this as a bool here, it will probably end up being used as a bool everywhere in the code with the exception of the DB-to-model conversion. Why go to all the trouble of modeling it as an enum, if you yourself expect it to behave as a bool ? Why not keep it simple and also store it in the domain model classes and the DB as a bool ?

src/zenml/models/v2/core/schedule.py
```diff
@@ -361,8 +447,43 @@ class ScheduleFilter(ProjectScopedFilter):
...
         default=False,
         description="Whether or not the schedule is archived",
     )
+    engine: str | None = Field(
+        default=None, description="The engine controlling the schedule"
+    )
+    next_occurrence_upper_bound: datetime | None = Field(
+        default=None,
+        description="Upper bound filter on next occurrence (occurrence must be lesser/older than upper bound).",
+    )
```
stefannica: this doesn't really fit well with how filters work in general. The attributes themselves already support included comparison operators. Look at this for example:

```
$ zenml stack list --created 'gte:2025-12-01 00:00:00'
The current repo active project is no longer available.
  active   ID                                     NAME            USER              ARTIFACT STORE   ORCHESTRATOR    DEPLOYER                                                                 
  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────                                                            
    -      425a12db-dae8-4261-9b93-985c0050de1f   s3-datadog      stefan@zenml.io   s3               default         default                                                                  
    -      56c09dba-6ee9-41bc-aa27-1c5250f28147   frankenstein    stefan@zenml.io   aws-s3           aws-sagemaker   gcp-cloud-run                                                            
    -      dd299464-a36f-4ee9-a22b-98ff32311591   local-datadog   stefan@zenml.io   default          default         default                                                                  
    -      91c97ba7-9eb1-40c9-9d29-d4365b07b352   local-dash0     stefan@zenml.io   s3               default         default                                                                  
    -      25ae3d14-4c18-4394-bbe0-4aec8892ae3f   aws-eks-gcs     stefan@zenml.io   gcp-gcs          eks-staging     -                                                                        
                                                                                                                                                                                              
Page `(1/1)`, `5` items found for the applied filters.

```

This means you can simply have a `next_occurrence` attribute modeled just like `created` is:

```
    created: Optional[Union[datetime, str]] = Field(
        default=None, description="Created", union_mode="left_to_right"
    )
```

This also means you don't need to treat this as a custom filter, because the inherited functionality does it all for you.

src/zenml/zen_stores/migrations/versions/50d5ae331122_add_new_scheduling_columns.py
```diff
@@ -0,0 +1,64 @@
...
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    with op.batch_alter_table("schedule", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("next_occurrence", sa.DateTime(), nullable=True)
+        )
+        batch_op.add_column(
+            sa.Column("engine", sa.VARCHAR(length=16), nullable=False)
+        )
```
stefannica:
```suggestion
        batch_op.add_column(
            sa.Column("engine", sa.VARCHAR(length=16), nullable=True)
        )
```

when you add a new column, you have to declare it as nullable, then populate it, then update it to non-nullable again.

src/zenml/zen_stores/migrations/versions/50d5ae331122_add_new_scheduling_columns.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    with op.batch_alter_table("schedule", schema=None) as batch_op:
+        batch_op.create_index(
+            get_index_name(
+                table_name="schedule", column_names=["next_occurrence"]
+            ),
+            ["next_occurrence"],
+            unique=False,
+        )
```
stefannica: this index doesn't seem to be mirrored by the ScheduleSchema changes. the alembic migration script has to be 100% aligned to the schema definitions, otherwise running alembic again after the migration is applied will detect differences that you didn't expect. 

it is generally a good idea not to change the generated alembic migration files to introduce new indices, keys, columns etc.

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: Makes sense, we can compute internally on the store.

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: Wow, super helpful, thanks. Will apply

src/zenml/models/v2/core/schedule.py
Json-Andriopoulos: Updated schema covers the many to many relationship :D .

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+        " it is preserved in the DB but otherwise unusable.",
+    )
+    trigger_type: str = Field(
+        sa_column=Column(VARCHAR(20), nullable=False),
+        description="The Trigger type e.g. webhook, schedule, etc.",
+    )
+    category: str = Field(
+        sa_column=Column(VARCHAR(50), nullable=False),
+        description="The Trigger category e.g. native schedule, github webhook, etc.",
+    )
```
stefannica: historically, I think we called this sort of thing `flavor`, because it might come in different flavors from different vendors, right ? github, gitlab, jira, who knows what else...

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+    data: str = Field(
```
stefannica: name nit: `configuration` fits better

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+
+    # ------------------ RELATIONSHIPS --------------------------------------------
+
+    snapshot_links: list["TriggerSnapshotSchema"] = Relationship(
+        back_populates="trigger",
+        sa_relationship_kwargs={
+            "cascade": "all, delete-orphan",
+            "passive_deletes": True,
+        },
+    )
```
stefannica: I don't think you need this tbh.

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+        Args:
+            trigger_update: A TriggerUpdate object.
+
+        Returns:
+            The updated TriggerSchema.
+
+        Raises:
+            ValueError: If trigger_update is invalid.
+        """
+        from zenml.models.v2.core.triggers import apply_schedule_update
```
stefannica: This feels wrong because:
* you are calling particular code (schedule) in a generic context (triggers)
* you're processing data in the DB layer, instead of processing it in the domain layer and then just passing it here when it's ready

You should apply any updates you need and prepare the data _before_ it reaches the trigger schema. Preferably somewhere where there's schedule specific code (i.e. not the SQL zen store either).

src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
```diff
@@ -243,6 +245,20 @@ class PipelineSnapshotSchema(BaseSchema, table=True):
         ),
     )
 
+    trigger_links: list["TriggerSnapshotSchema"] = Relationship(
+        back_populates="snapshot",
+        sa_relationship_kwargs={
+            "cascade": "all, delete-orphan",
+            "passive_deletes": True,
+        },
+    )
```
stefannica: I don't think you'll need this either.

src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
```diff
@@ -608,6 +624,10 @@ def to_model(
                     )
                     for visualization in self.visualizations
                 ],
+                triggers=[
+                    trigger.to_model(include_resources=False)
+                    for trigger in self.triggers
+                ],
```
stefannica: Resources are always pulled. This is where you would list information that needs to be included in the entries being listed via e.g. list_snapshot even when hydration is off.

You **only** need to put stuff here that you need to represent in the UI for every item in the list screen. So you only need this here if the "list snapshot" UI screen needs to include all the triggers for all the listed items. If you need to include the triggers in the hydrated model, add it to the metadata instead.

src/zenml/zen_stores/schemas/trigger_assoc.py
```diff
@@ -0,0 +1,97 @@
...
+        ),
+    )
+    pipeline_run_id: int = Field(
+        sa_column=Column(
+            ForeignKey("pipeline_run.id", ondelete="CASCADE"),
+            nullable=False,
+            primary_key=True,
+        ),
+    )
+
```
stefannica, line 89: The trigger execution would in the future have a metadata section where you store the metadata associated with the particular event that activates the trigger. Or some other schema that we call something else (e.g. event ?).

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+    TriggerSnapshotSchema,
+)
+from zenml.zen_stores.schemas.user_schemas import UserSchema
+from zenml.zen_stores.schemas.utils import (
+    RunMetadataInterface,
+    jl_arg,
+)
+
+
+class TriggerSchema(NamedSchema, RunMetadataInterface, table=True):
```
stefannica:
```suggestion
class TriggerSchema(NamedSchema, table=True):
```

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+        build_index(
+            table_name=__tablename__,
+            column_names=[
+                "category",
+            ],
+        ),
+        build_index(
+            table_name=__tablename__,
+            column_names=["category", "next_occurrence"],
+        ),
```
stefannica, line 76: doesn't this cover `category` as a separate index too ?

stefannica: update: it covers it, but only because it's the first entry in the list

stefannica: then you don't need the separate one

src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -1650,6 +1651,130 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+
+        Comment: Useful to associate triggers & runs.
+
+        Args:
+            trigger_id: The ID of the trigger.
+            pipeline_run_id: The ID of the pipeline run.
+
+        Raises:
+            KeyError: if the entities don't exist.
+        """
```
stefannica: If this will only be called at the database level, don't have it in the interface. You may have it only in the SQL zen store.

src/zenml/zen_stores/rest_zen_store.py
```diff
@@ -2086,6 +2088,159 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+            trigger_id: The ID of the trigger.
+            pipeline_run_id: The ID of the pipeline run.
+
+        Raises:
+            KeyError: if the entities don't exist.
+        """
+        self.put(
+            path=f"{TRIGGERS}/{trigger_id}/{RUNS}/{pipeline_run_id}",
+            timeout=5,
+        )
```
stefannica: probably not needed if it's internal

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+        snapshot_id=snapshot_id,
+    )
+
+
+@router.put(
+    "/{trigger_id}}" + RUNS + "/{pipeline_run_id}",
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+@async_fastapi_endpoint_wrapper
+def create_trigger_execution(
```
stefannica: probably not needed as an API endpoint

src/zenml/zen_stores/migrations/versions/3855b5d051cd_add_v2_of_trigger_schemas.py
```diff
@@ -0,0 +1,122 @@
...
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "name", "project_id", name="unique_trigger_name_in_project"
+        ),
+    )
+    with op.batch_alter_table("trigger", schema=None) as batch_op:
+        batch_op.create_index(
+            "ix_trigger_category", ["category"], unique=False
+        )
```
stefannica: not needed (covered by the compound below)

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+        sa_column=Column(VARCHAR(20), nullable=False),
+        description="The Trigger type e.g. webhook, schedule, etc.",
+    )
+    category: str = Field(
+        sa_column=Column(VARCHAR(50), nullable=False),
+        description="The Trigger category e.g. native schedule, github webhook, etc.",
+    )
+    data: str = Field(
+        sa_column=Column(
+            String(length=MEDIUMTEXT_MAX_LENGTH).with_variant(
```
stefannica: this is too big (16 MB), try text (64k)

src/zenml/zen_stores/migrations/versions/3855b5d051cd_add_v2_of_trigger_schemas.py
```diff
@@ -0,0 +1,122 @@
...
+        sa.Column("updated", sa.DateTime(), nullable=False),
+        sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("active", sa.Boolean(), nullable=False),
+        sa.Column("is_archived", sa.Boolean(), nullable=False),
+        sa.Column("trigger_type", sa.VARCHAR(length=20), nullable=False),
+        sa.Column("category", sa.VARCHAR(length=50), nullable=False),
+        sa.Column(
+            "data",
+            sa.String(length=16777215).with_variant(
+                mysql.MEDIUMTEXT(), "mysql"
```
stefannica: this is too big (16 MB), try text (64k)

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+    responses={401: error_response, 409: error_response, 422: error_response},
+)
+# TODO: the workspace scoped endpoint is only kept for dashboard compatibility
+# and can be removed after the migration
+@workspace_router.post(
+    "/{project_name_or_id}" + TRIGGERS,
+    responses={401: error_response, 409: error_response, 422: error_response},
+    deprecated=True,
+    tags=["triggers"],
+)
```
stefannica:
```suggestion
```

not needed anymore (legacy crap)

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+@async_fastapi_endpoint_wrapper
+def create_trigger(
+    trigger: TriggerRequest,
+    project_name_or_id: Optional[Union[str, UUID]] = None,
```
stefannica: this you won't need either. You have the project scope now in the model itself.
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+    _: AuthContext = Security(authorize),
+) -> TriggerResponse:
+    """Creates a trigger.
+
+    Args:
+        trigger: The trigger to create.
+        project_name_or_id: Optional name or ID of the project.
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Returns:
+        The created trigger.
+    """
+    if project_name_or_id:
+        project = zen_store().get_project(project_name_or_id)
+        trigger.project = project.id
+
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+    responses={401: error_response, 404: error_response, 422: error_response},
+)
+# TODO: the workspace scoped endpoint is only kept for dashboard compatibility
+# and can be removed after the migration
+@workspace_router.get(
+    "/{project_name_or_id}" + TRIGGERS,
+    responses={401: error_response, 404: error_response, 422: error_response},
+    deprecated=True,
+    tags=["triggers"],
+)
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+@async_fastapi_endpoint_wrapper
+def list_triggers(
+    trigger_filter_model: TriggerFilter = Depends(
+        make_dependable(TriggerFilter)
+    ),
+    project_name_or_id: Optional[Union[str, UUID]] = None,
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+    hydrate: bool = False,
+    _: AuthContext = Security(authorize),
+) -> Page[TriggerResponse]:
+    """Gets a list of triggers.
+
+    Args:
+        trigger_filter_model: Filter model used for pagination, sorting,
+            filtering
+        project_name_or_id: Optional name or ID of the project.
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+        hydrate: Flag deciding whether to hydrate the output model(s)
+            by including metadata fields in the response.
+
+    Returns:
+        List of trigger objects.
+    """
+    if project_name_or_id:
+        trigger_filter_model.project = project_name_or_id
+
```
stefannica:
```suggestion
```

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+    """Attaches a trigger to a snapshot.
+
+    Args:
+        trigger_id: The ID of the trigger.
+        snapshot_id: The ID of the snapshot.
+    """
+    verify_permission_for_model(
+        model=zen_store().get_trigger(trigger_id=trigger_id, hydrate=True),
+        action=Action.UPDATE,
+    )
```
stefannica, line 205: and READ permissions on the snapshot.

stefannica: btw, there's an RBAC utility that you can use to verify both in one call: `batch_verify_permissions_for_models`. otherwise you make two calls to the RBAC API.

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -0,0 +1,291 @@
...
+        The created trigger.
+    """
+    if project_name_or_id:
+        project = zen_store().get_project(project_name_or_id)
+        trigger.project = project.id
+
+    return verify_permissions_and_create_entity(
+        request_model=trigger,
+        create_method=zen_store().create_trigger,
+    )
```
stefannica, line 74: You need some changes in the RBAC layer for this to work:

1. resource type (covered)
2. update `get_resource_type_for_model` with your trigger models
3. update `get_schema_for_resource_type` with your schema
4. talk to @schustmi to add this new RBAC resource type to the Pro RBAC component (requires a complicated DB migration there that only he can write)
5. talk to @Cahllagerfeld to add this new resource type in the UI (I don't think they are synced automatically to the backend)

src/zenml/zen_server/routers/trigger_endpoints.py
stefannica: but you can't use it with different actions :-(

src/zenml/models/v2/core/triggers.py
```diff
@@ -0,0 +1,401 @@
...
+
+    name: str
+    active: bool = Field(
+        default=True,
+        description="Whether to activate this trigger upon creation.",
+    )
+    description: str | None = None
+    trigger_type: TriggerType
+    category: TriggerCategory
+    data: SchedulePayload
```
stefannica: just leave a note here that this will be `ScheduleConfiguration| WebhookConfiguration` in the future.

src/zenml/models/v2/core/triggers.py
```diff
@@ -0,0 +1,401 @@
...
+
+        return self
+
+
+class TriggerUpdate(BaseUpdate):
+    """Update model for triggers."""
+
+    name: str | None = None
+    active: bool | None = None
+    data: ScheduleUpdatePayload | None = None
```
stefannica: having this specialization in the update model (and the create one) allows the REST API to be more specific and also validate that it has the right fields set (as opposed to having an opaque Dict). in other words, it's better to have a typed configuration as part of the API model.

the consequence is that there has to be some code on the other side of the REST API that validates and applies the create/update operation before it reaches the database.

this validation will only target the basic trigger types, but not expand into category specificity.

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -0,0 +1,340 @@
...
+    def from_request(cls, trigger_request: TriggerRequest) -> "TriggerSchema":
+        """Creates a TriggerSchema object from a TriggerRequest.
+
+        Args:
+            trigger_request: A TriggerRequest object.
+
+        Returns:
+            A TriggerSchema object.
+        """
+        from zenml.utils.native_schedules import calculate_first_occurrence
```
stefannica: this too belongs in the SQL zen store

src/zenml/client.py
```diff
@@ -3856,6 +3863,156 @@ def delete_run_template(
 
         self.zen_store.delete_run_template(template_id=template_id)
 
+    # ------------------------------- Triggers ---------------------------------
+
+    def create_native_schedule_trigger(
```
stefannica:
```suggestion
    def create_native_schedule(
```

src/zenml/client.py
```diff
@@ -3856,6 +3863,156 @@ def delete_run_template(
...
+                    interval=interval,
+                    run_once_start_time=run_once_start_time,
+                    end_time=end_time,
+                    start_time=start_time,
+                    engine=ScheduleEngine.native,
+                ),
+            )
+        )
+
+    def update_native_schedule_trigger(
```
stefannica:
```suggestion
    def update_native_schedule(
```

src/zenml/client.py
```diff
@@ -3856,6 +3863,156 @@ def delete_run_template(
...
+                active=active,
+                data=ScheduleUpdatePayload(
+                    cron_expression=cron_expression,
+                    interval=interval,
+                    start_time=start_time,
+                ),
+            ),
+        )
+
+    def get_trigger(self, trigger_id: UUID) -> TriggerResponse:
```
stefannica: this should be get_native_scheduler

src/zenml/client.py
```diff
@@ -3856,6 +3863,156 @@ def delete_run_template(
...
+        """
+        if not soft:
+            logger.warning(
+                "You are deleting a trigger with option soft set to False. "
+                "All historical data and references to this trigger will be deleted."
+            )
+
+        self.zen_store.delete_trigger(trigger_id=trigger_id, soft=soft)
+
+    def attach_trigger_to_snapshot(
```
stefannica, line 3983:
```suggestion
    def attach_schedule_to_snapshot(
```

src/zenml/models/v2/core/triggers.py
```diff
@@ -0,0 +1,401 @@
...
+        if not (self.end_time and self.start_time):
+            return self
+
+        if self.end_time <= self.start_time:
+            raise ValueError("The end time must be after the start time.")
+
+        return self
+
+
+class TriggerRequest(ProjectScopedRequest):
```
stefannica: Is it at all possible that you model ScheduleRequest as a subclass of TriggerRequest? and then have the POST /trigger endpoint expect a union of ScheduleRequest | WebhookRequest.

This would be great to have, because then you can also have ScheduleResponse subclass that you can return from `get_schedule` instead of `TriggerResponse` in the Client.

src/zenml/triggers/models.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def get_extra_fields(self) -> dict[str, Any]:
+        """Calculates extra, type-specific fields needed for the trigger.
+
+        Returns:
+            A dictionary of extra fields (e.g. {"next_occurrence': "..."}
+        """
```
stefannica:
```suggestion
        Returns:
            A dictionary of extra fields (e.g. {"next_occurrence': "..."}) to be stored in the database.
            IMPORTANT: These fields must match the attributes defined in the `TriggerSchema` ORM class.
        """
```

src/zenml/triggers/models.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class TriggerFilter(ProjectScopedFilter):
+    """Base class for filtering triggers."""
+
+    name: str
+    active: bool
+    is_archived: bool
+    flavor: TriggerFlavor
+    type: TriggerType
```
stefannica: You should check with @Cahllagerfeld too, but I think it might be useful to filter triggers by their associations too, such as triggers linked to a snapshot UUID or triggers used to run a pipeline.

src/zenml/triggers/models.py
```diff
@@ -0,0 +1,129 @@
+# Copyright (c) ZenML GmbH 2026. All Rights Reserved.
```
stefannica: these are still API models, so you should keep them in the models/v2/core folder.

src/zenml/models/v2/core/triggers.py
```diff
@@ -0,0 +1,346 @@
+# Copyright (c) ZenML GmbH 2026. All Rights Reserved.
```
stefannica, line 1: These are also API models, so you should keep them in the models/v2/core folder.

This is probably my fault. I hinted at the idea of having a `triggers` or `schedules` root module, not for models, but for core classes that you might want to add to the SDK (e.g. to have a `Trigger` and a `Schedule` class and use them in the user-facing part of the SDK library instead of TriggerRequest/Response and ScheduleRequest/Response, just like `Stack` is used instead of `StackRequest`). But you haven't done that, so my suggestion doesn't apply.

src/zenml/zen_server/rbac/utils.py
```diff
@@ -523,6 +526,9 @@ def get_resource_type_for_model(
         ProjectResponse: ResourceType.PROJECT,
         ProjectRequest: ResourceType.PROJECT,
         # UserResponse: ResourceType.USER,
+        ScheduleTriggerRequest: ResourceType.TRIGGER,
+        ScheduleTriggerResponse: ResourceType.TRIGGER,
+        TriggerRequest: ResourceType.TRIGGER,
```
stefannica, line 531: Seeing this makes me wonder whether we should actually model a resource type for each type of trigger instead of a global `TRIGGER` one. From a UX perspective, it totally makes sense to separate permissions for schedules and webhooks for example. This is a design decision though, so it should involve @znegrin .

src/zenml/triggers/registry.py
```diff
@@ -0,0 +1,38 @@
...
+
+TYPE_TO_RESPONSE_MAPPING = {
+    TriggerType.SCHEDULE.value: ScheduleTriggerResponse,
+}
+
+# Union type objects - should be extended with the future type classes (e.g. Webhook)
+
+TRIGGER_UPDATE_TYPE_UNION = ScheduleTriggerUpdate
+TRIGGER_CREATE_TYPE_UNION = ScheduleTriggerRequest
+TRIGGER_RETURN_TYPE_UNION = ScheduleTriggerResponse
```
stefannica, line 38: This works for now, but I'm worried that when we will actually extend these with the WebhookTrigger model, so that they are actual Unions, we'll find out that they don't work with FastAPI because there's ambiguity between the models.

Json-Andriopoulos: We can't have ambiguity in the models now. Updates are full, not partial. Requests and responses are not ambiguous by default. Even if they become we can use Fastapi's discriminator to resolve this.

src/zenml/triggers/models.py
Json-Andriopoulos: We'll do it on follow-up PRs when it becomes clear which associations/filters we need exactly.

src/zenml/zen_server/rbac/utils.py
znegrin: @stefannica This is indeed one of the proposals I have in my research document: https://www.notion.so/zenml/Schedules-Triggers-Research-Summary-2e9f8dff2538807eabd1d0d04b5f7a2d?source=copy_link#07f1f678280c44139b8096542c1cf7c0

image

Json-Andriopoulos: Hey @znegrin that is the modeling we follow as well, though Stefan is referring to the permission/auth layer. In this PR we cover all trigger types under a single umbrella, it may make sense in the future to separate them. We can revisit this when webhooks are introduced though.

stefannica (approved):

src/zenml/zen_server/rbac/utils.py
znegrin: sounds good to me!

PR #4431, opened 2026-01-23
Add missing index to the snapshot table

stefannica:
## Describe changes
When fetching a snapshot, the last run associated with it is also fetched. However, this uses a different column (`source_snapshot_id`) that was previously not configured as an index. On databases with very large number of snapshots, this leads to very high DB query times that impacts ZenML's scalabilty.

This PR adds the missing index.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/migrations/versions/d1e8c2c4a9ef_add_snapshot_index.py
+++ b/src/zenml/zen_stores/migrations/versions/d1e8c2c4a9ef_add_snapshot_index.py
@@ -1,0 +1,31 @@
+"""add snapshot index [d1e8c2c4a9ef].
+
+Revision ID: d1e8c2c4a9ef
+Revises: 0e5c99d02aa7
+Create Date: 2026-01-23 16:25:05.328420
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "d1e8c2c4a9ef"
+down_revision = "0e5c99d02aa7"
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/migrations/versions/d1e8c2c4a9ef_add_snapshot_index.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = "0e5c99d02aa7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    with op.batch_alter_table("pipeline_snapshot", schema=None) as batch_op:
+        batch_op.create_index(
+            "ix_pipeline_snapshot_source_snapshot_id",
```
Json-Andriopoulos, line 22: Minor: we have a utility that creates index names on a common format. Maybe we could use that here.

stefannica: Correct, we do, but we decided not to use that in migration scripts for two reasons:

1. these are generated by alembic to match the changes in the DB schema generated from code
2. including ZenML code in alembic migration scripts is risky, because zenml code can change, whereas migration scripts must be immutable and have the same net effect even 6-12 months from now.

PR #4438, opened 2026-01-28
Rewrite the filtering queries to be more efficient

stefannica:
## Describe changes
The way that the ZenML filtering system generates the SQL queries used to fetch items from the database is still inefficient at scale. Particularly around these areas:

* counting items (i.e. to determine what the total number of fetched pages is) doesn't need sorting
* the DISTINCT statement applied to multiple columns is awful in terms of performance and not necessary when there's an ID primary key column involved
* the OR subqueries are particularly strenuous on the database because they are implemented using cartesian products, so each statement has 1000 entries, the rows generated by combining then with an OR statement will be 1 million entries. Union statements are preferred.

Examples - artifact version queries on a table with ~20k entries:

1. AND query: 2 sec

```
$ time curl 'https://.../api/v1/artifact_versions?hydrate=false&sort_by=created&logical_operator=and&page=1&size=20&project=stress&artifact=output14&type=DataArtifact'
real    0m2.137s
user    0m0.019s
sys    0m0.007s
```

2. same query, with OR instead of AND: 3 minutes !!!

```
time curl 'https://.../api/v1/artifact_versions?hydrate=false&sort_by=created&logical_operator=or&page=1&size=20&project=stress&artifact=output14&type=DataArtifact'
real    3m0.207s
user    0m0.039s
sys    0m0.017s
```

3. same query, with OR and the patch in this PR: 2 seconds

```
$ time curl 'https://0fbda6c6-zenml.staging.cloudinfra.zenml.io/api/v1/artifact_versions?hydrate=false&sort_by=created&logical_operator=or&page=1&size=20&project=stress&artifact=output14&type=DataArtifact'
real    0m2.404s
user    0m0.024s
sys    0m0.005s
```

4. to make it more complicated, with an added pipeline ID filter: 2 seconds

```
$ time curl 'https://0fbda6c6-zenml.staging.cloudinfra.zenml.io/api/v1/artifact_versions?hydrate=false&sort_by=created&logical_operator=or&page=1&size=20&project=stress&artifact=output14&type=DataArtifact&pipeline_run=bab4021f-7b36-40bb-a809-e4cdbdb057c8'
real	0m2.405s
user	0m0.025s
sys	0m0.008s
```

Other changes included in this PR:

* fix the queries in the stack component filter
* get rid of unneeded code in the service filter

### Details

This PR was brought on by the following query which has a dramatic effect on a database with 20000 artifact versions:

```
SELECT count(*) AS count_1 FROM (
SELECT DISTINCT artifact_version.id AS id, artifact_version.created AS created, artifact_version.updated AS updated, artifact_version.version AS version, artifact_version.version_number AS version_number, artifact_version.type AS type, artifact_version.uri AS uri, artifact_version.materializer AS materializer, artifact_version.data_type AS data_type, artifact_version.save_type AS save_type, artifact_version.content_hash AS content_hash, artifact_version.artifact_id AS artifact_id, artifact_version.artifact_store_id AS artifact_store_id, artifact_version.user_id AS user_id, artifact_version.project_id AS project_id
FROM artifact_version, step_run_output_artifact, step_run, step_run_input_artifact, pipeline_run
WHERE (
artifact_version.id = step_run_output_artifact.artifact_id AND step_run_output_artifact.step_id = step_run.id
OR artifact_version.id = step_run_input_artifact.artifact_id AND step_run_input_artifact.step_id = step_run.id
)
AND step_run.pipeline_run_id = pipeline_run.id AND pipeline_run.id = 'bab4021f7b3640bba809e4cdbdb057c8'
ORDER BY artifact_version.created ASC, artifact_version.id ASC
)
```

This query is generated when trying to list all artifact versions for a particular pipeline run.

This query is simply **catastrophic** in terms of performance, for the following reasons:

* Unconstrained table on each OR branch: When the LEFT side of the OR matches (output artifacts), step_run_input_artifact has no join condition — it cross-joins with all rows. Vice versa for the RIGHT side.
* Exponential intermediate rows: With N output artifacts and M input artifacts, the query produces O(N × M) intermediate rows before filtering, instead of O(N + M).
* Indexes bypassed: The OR across different join paths prevents the optimizer from using indexes efficiently on both branches simultaneously.

This PR simplifies the calculation of COUNT and replaces the OR in the artifact version filter with a UNION, so the final query looks like this:

```
SELECT count(*) AS count_1 FROM (
SELECT DISTINCT artifact_version.id AS id
FROM artifact_version
INNER JOIN (
    SELECT step_run_output_artifact.artifact_id AS artifact_id
    FROM step_run_output_artifact 
    INNER JOIN step_run ON step_run_output_artifact.step_id = step_run.id
    INNER JOIN pipeline_run ON step_run.pipeline_run_id = pipeline_run.id
    WHERE pipeline_run.id = 'bab4021f7b3640bba809e4cdbdb057c8' 
       OR pipeline_run.name = 'bab4021f-7b36-40bb-a809-e4cdbdb057c8'
    UNION
    SELECT step_run_input_artifact.artifact_id
    FROM step_run_input_artifact 
    INNER JOIN step_run ON step_run_input_artifact.step_id = step_run.id
    INNER JOIN pipeline_run ON step_run.pipeline_run_id = pipeline_run.id
    WHERE pipeline_run.id = 'bab4021f7b3640bba809e4cdbdb057c8' 
       OR pipeline_run.name = 'bab4021f-7b36-40bb-a809-e4cdbdb057c8'
) AS matched_artifacts ON artifact_version.id = matched_artifacts.artifact_id;
```

The actual query used to fetch the rows is also greatly simplified as a result.

When OR is used as the logical operators in filters, UNION is used instead of OR to construct the final query.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/models/v2/core/component.py
+++ b/src/zenml/models/v2/core/component.py
@@ -31,6 +31,7 @@ from pydantic import BaseModel, Field, field_validator
 from zenml.constants import STR_FIELD_MAX_LENGTH
 from zenml.enums import StackComponentType
 from zenml.models.v2.base.base import BaseUpdate
+from zenml.models.v2.base.filter import AnyQuery
 from zenml.models.v2.base.scoped import (
     UserScopedFilter,
     UserScopedRequest,
@@ -435,6 +436,33 @@ class ComponentFilter(UserScopedFilter):
         """
         self.scope_type = component_type
 
+    def apply_filter(
+        self,
+        query: AnyQuery,
+        table: Type["AnySchema"],
+    ) -> AnyQuery:
+        """Applies the filter to a query.
+
+        Args:
+            query: The query to which to apply the filter.
+            table: The query table.
+
+        Returns:
+            The query with filter applied.
+        """
+        from sqlmodel import col
+
+        from zenml.zen_stores.schemas import StackComponentSchema
+
+        query = super().apply_filter(query=query, table=table)
+
+        if self.scope_type:
+            query = query.where(
+                col(StackComponentSchema.type) == self.scope_type
+            )
+
+        return query
+
     def get_custom_filters(
         self, table: Type["AnySchema"]
     ) -> List["ColumnElement[bool]"]:
@@ -451,16 +479,12 @@ class ComponentFilter(UserScopedFilter):
         """
         custom_filters = super().get_custom_filters(table)
 
-        from sqlmodel import and_, col
+        from sqlmodel import and_
 
         from zenml.zen_stores.schemas import (
             StackComponentSchema,
             StackCompositionSchema,
         )
-
-        if self.scope_type:
-            type_filter = col(StackComponentSchema.type) == self.scope_type
-            custom_filters.append(type_filter)
 
         if self.stack_id:
             stack_filter = and_(


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -1262,6 +1261,17 @@ def filter_and_paginate(
                 filter_model.offset : filter_model.offset + filter_model.size
             ]
         else:
+            # Apply sorting and distinct for the data fetch
+            query = filter_model.apply_sorting(
+                query=query, table=table
+            ).distinct()
```
schustmi, line 1269: Your changes makes it so that if a custom fetching method is provided, the `distinct` part is now not applied anymore. Do you think this is something that we have to care about, or does it not matter for those cases?

stefannica: The only table that uses this custom query is the service connectors. I applied it, just in case.

PR #4439, opened 2026-01-28
Run:AI Integration - Step Operator

safoinme:
## Describe changes
Adds a new Run:AI step operator integration that enables running individual pipeline steps on Run:AI clusters with fractional GPU allocation.                                                                                    
                                                                                                                                                                                                                                   
#### Features                                                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  - Fractional GPU support (e.g., 0.25, 0.5, 0.75 GPU per step)                                                                                                                                                                    
  - Comprehensive compute settings (GPU memory, CPU cores, node pools, tolerations)                                                                                                                                                
  - Run:AI API client wrapper with typed responses                                                                                                                                                                                 
  - Support for both Run:AI SaaS and self-hosted deployments                                                                                                                                                                       
                                                                                                                                                                                                                                   
#### Usage                                                                                                                                                                                                                            
```                                                                                                                                                                                                     
  from zenml import step                                                                                                                                                                                                           
  from zenml.integrations.runai.flavors import RunAIStepOperatorSettings                                                                                                                                                           
                                                                                                                                                                                                                                   
  @step(                                                                                                                                                                                                                           
      step_operator="runai",                                                                                                                                                                                                       
      settings={"step_operator": RunAIStepOperatorSettings(gpu_portion_request=0.5)}                                                                                                                                               
  )                                                                                                                                                                                                                                
  def train_model():                                                                                                                                                                                                               
      # Runs on Run:AI with half a GPU                                                                                                                                                                                             
```                                                                                                                                                                                                                       
                                                                                                                                                                                                                                   
#### Changes                                                                                                                                                                                                                          
                                                                                                                                                                                                                                   
  - New integration: src/zenml/integrations/runai/                                                                                                                                                                                 
  - Documentation: docs/book/component-guide/step-operators/runai.md                                                                                                                                                               
  - Example pipeline: examples/runai/                                                                                                                                                                                              
  - Unit tests: 84 tests covering constants and settings validation 

## Pre-requisites
  - [X] [ZenML Docs](https://docs.zenml.io)

## Types of changes
<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->


htahir1:
@claude stefan-review

safoinme:
@claude full-review ignore tests and example for now

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,210 @@
...
+
+### How to deploy it
+
+The Run:AI step operator requires access to a Run:AI cluster. You will need:
+
+* Access to a Run:AI cluster (SaaS or self-hosted)
+* A Run:AI project with sufficient resource quota
+* Run:AI API credentials (client ID and secret):
+  * Navigate to Run:AI control plane → Settings → Applications
+  * Create a new application to obtain client credentials
```
strickvl: Maybe this needs a link to some place in the runai docs?

strickvl: Also not sure if you need to clarify something about the credentials... i.e. does everyone in the org have the same client ID + secret, or it's user-specific etc? i.e. think about it from the platform engineer's perspective. As in what would they care about reading here. I imagine yuo learned some things while setting it up i.e. around tradeoffs or something?

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,210 @@
...
+@step(
+    step_operator="runai",
+    settings={"step_operator": runai_settings}
+)
+def train_model():
+    # This step runs with half a GPU
+    ...
+```
+
+#### Fractional GPU Allocation
```
strickvl: didn't we just see this example above?

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,210 @@
...
+| Option | Type | Required | Description |
+|--------|------|----------|-------------|
+| `client_id` | str | Yes | Run:AI client ID for API authentication |
+| `client_secret` | str | Yes | Run:AI client secret for API authentication |
+| `runai_base_url` | str | Yes | Run:AI control plane URL (e.g., `https://org.run.ai`) |
+| `project_name` | str | Yes | Run:AI project name for workload submission |
+| `cluster_name` | str | No | Run:AI cluster name (uses project's cluster if not specified) |
+| `image_pull_secret_name` | str | No | Name of Run:AI image pull secret for private registries |
+| `monitoring_interval` | float | No | Interval in seconds to poll workload status (default: 30) |
+| `workload_timeout` | int | No | Maximum time in seconds for workload completion |
```
strickvl, line 135: Check the formatting here. Something seems weird in the markdown rendering from github:

Image

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,210 @@
...
+- Install the Run:AI integration: `zenml integration install runai`
+
+#### Viewing Logs
+
+Run:AI workload logs can be viewed in the Run:AI control plane UI:
+1. Navigate to your Run:AI control plane
+2. Go to Workloads → Training
+3. Find your step workload and click to view logs
+
+<figure><img src="https://static.scarf.sh/a.png?x-pxid=f0b4f458-0a54-4fcd-aa95-d5ee424815bc" alt="ZenML Scarf"><figcaption></figcaption></figure>
```
strickvl:
```suggestion
```

We don't use this any more

docs/book/component-guide/step-operators/runai.md
strickvl, line 1: probably a screenshot of the runai interface would be good somewhere in this page.

src/zenml/integrations/runai/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+RUNAI_STEP_OPERATOR_FLAVOR = "runai"
+
+
+class RunAIIntegration(Integration):
+    """Definition of Run:AI integration for ZenML."""
+
+    NAME = RUNAI
+    REQUIREMENTS = [
+        "runapy>=0.1.0",
```
strickvl: maybe include `<1` just for safety?

src/zenml/integrations/runai/constants.py
```diff
@@ -0,0 +1,143 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Run:AI integration constants and status mappings."""
+
+from enum import Enum
+
+from zenml.enums import ExecutionStatus
+
+# Workload name constraints
```
strickvl:
```suggestion
```

src/zenml/integrations/runai/constants.py
```diff
@@ -0,0 +1,143 @@
...
+    SUCCEEDED = "succeeded"
+    COMPLETED = "completed"
+    SUCCESS = "success"
+    FAILED = "failed"
+    ERROR = "error"
+    STOPPED = "stopped"
+    STOPPING = "stopping"
+
+
+# Mapping from Run:AI status to ZenML ExecutionStatus
```
strickvl:
```suggestion
```

src/zenml/integrations/runai/constants.py
```diff
@@ -0,0 +1,143 @@
...
+    RunAIWorkloadStatus.SUCCEEDED: ExecutionStatus.COMPLETED,
+    RunAIWorkloadStatus.COMPLETED: ExecutionStatus.COMPLETED,
+    RunAIWorkloadStatus.SUCCESS: ExecutionStatus.COMPLETED,
+    RunAIWorkloadStatus.FAILED: ExecutionStatus.FAILED,
+    RunAIWorkloadStatus.ERROR: ExecutionStatus.FAILED,
+    RunAIWorkloadStatus.STOPPED: ExecutionStatus.FAILED,
+    RunAIWorkloadStatus.STOPPING: ExecutionStatus.STOPPING,
+}
+
+# Sets for quick status classification
```
strickvl:
```suggestion
```

etc... my `/deslopify` CC command would handle all these FYI

src/zenml/integrations/runai/constants.py
```diff
@@ -0,0 +1,143 @@
...
+_TERMINAL_STATUSES = frozenset(
```
strickvl, line 50: TIL about `frozenset`!

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        Raises:
+            RunAIClientError: If the query fails.
+        """
+        try:
+            response = self._raw_client.workloads.trainings.get_training(
+                workload_id
+            )
+            if response.data and isinstance(response.data, dict):
+                return cast(Dict[str, Any], response.data)
+            return {}
```
strickvl: Why would we return an empty dict here? Feels like maybe a silent failure here? i.e. how can the caller of this function distinguish between 'workload exists but has no data' and 'something is wrong'? Maybe we should raise an error here instead of passing back a dict?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+                or response.data.get("id")
+                or response.data.get("workload_id")
+            )
+        return getattr(response.data, "workloadId", None) or getattr(
+            response.data, "id", None
+        )
+
+    def get_training_workload_status(self, workload_id: str) -> Optional[str]:
+        """Get the status of a training workload.
+
```
strickvl, line 459: In this function we return `None` in some fairly different scenarios:

1. when the response data is empty
2. then the workload isn't found (we only log a warning which feels the wrong approach?)
3. when the status field is missing

Thinking this should be made a bit more robust at the very least for the user to bubble up what's going on with their training workload? Think about what it'd be like to debug these different scenarios.

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+            workload_id = self._extract_workload_id(response)
+            return WorkloadSubmissionResult(
+                workload_id=workload_id or request.name,
+                workload_name=request.name,
+            )
+        except RunAIClientError:
+            raise
+        except Exception as exc:
+            raise RunAIClientError(
+                f"Failed to submit Run:AI workload: {exc}"
```
strickvl:
```suggestion
                f"Failed to submit Run:AI workload ({type(exc).__name__}): {exc}"
```

strickvl: at the very least... this exception handling is a bit too broad I think. will make for unpleasant debugging experience when something fails

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        except RunAIClientError:
+            raise
+        except Exception as exc:
+            error_msg = str(exc).lower()
+            if "not found" in error_msg or "404" in error_msg:
+                logger.warning(f"Workload {workload_id} not found")
+                return None
+            else:
+                raise RunAIClientError(
+                    f"Failed to query workload status: {exc}"
```
strickvl:
```suggestion
                    f"Failed to query workload status: ({type(exc).__name__}): {exc}"
```

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+                workload_id
+            )
+            if response.data and isinstance(response.data, dict):
+                return cast(Dict[str, Any], response.data)
+            return {}
+        except RunAIClientError:
+            raise
+        except Exception as exc:
+            raise RunAIClientError(
+                f"Failed to query Run:AI workload {workload_id}: {exc}"
```
strickvl:
```suggestion
                f"Failed to query Run:AI workload {workload_id}: ({type(exc).__name__}): {exc}"
```

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        Raises:
+            RunAIClientError: If deletion fails.
+        """
+        try:
+            self._raw_client.workloads.trainings.delete_training(workload_id)
+        except RunAIClientError:
+            raise
+        except Exception as exc:
+            raise RunAIClientError(
+                f"Failed to delete Run:AI workload {workload_id}: {exc}"
```
strickvl:
```suggestion
                f"Failed to delete Run:AI workload {workload_id}: ({type(exc).__name__}): {exc}"
```

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+
+        Run:AI step operator never runs locally.
+
+        Returns:
+            False, as Run:AI step operator never runs locally.
+        """
+        return False
+
+
+class RunAIStepOperatorFlavor(BaseStepOperatorFlavor):
```
strickvl, line 393: missing `display_name` in here?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+            if timeout and (time.time() - start_time) > timeout:
+                logger.warning(
+                    f"Attempting to stop timed-out workload {workload_id}"
+                )
+                try:
+                    client.delete_training_workload(workload_id)
+                except Exception as cleanup_exc:
+                    logger.error(
+                        f"Failed to cleanup workload {workload_id}: {cleanup_exc}"
+                    )
```
strickvl: Wondering a bit about this... we only log a failed cleanup. So users might have orphaned workloads if it fails.

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+        ...,
+        description="Run:AI client secret for API authentication. Obtain from Run:AI "
+        "control plane under Settings > Application > Applications. Stored securely",
+    )
+    runai_base_url: str = Field(
+        ...,
+        description="Run:AI control plane base URL. For SaaS deployments, use format "
+        "'https://<organization>.run.ai'. For self-hosted, use your control plane URL. "
+        "Example: 'https://my-org.run.ai'",
+    )
```
strickvl, line 301: Should we include Pydantic validation here that this is actually a URL?

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+            "resources within quota",
+        )
+    )
+    priority_class: Optional[str] = Field(
+        default=None,
+        description="Kubernetes PriorityClass name for scheduling priority within project queue. "
+        "Run:AI presets: 'train' (50), 'build' (50), 'interactive-preemptible' (100), 'inference' (125)",
+    )
+
+    tolerations: Optional[List[Dict[str, Any]]] = Field(
```
strickvl, line 145: Probably low prio, but wondering if we should have a structured model for the tolerations instead of this relatively loose `Dict[str, Any`?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        Raises:
+            RunAIClientError: If no clusters are available.
+        """
+        clusters = self.get_clusters()
+        if not clusters:
+            raise RunAIClientError("No Run:AI clusters available")
+        return clusters[0]
+
+    def create_training_workload(
+        self, request: Any
```
strickvl: Maybe a better type than `Any` for this?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+                    sleep_time = base_interval
+                elif status and is_success_status(status):
+                    missing_status_retries = 0
+                    return
+
+                elif status and is_failure_status(status):
+                    missing_status_retries = 0
+                    try:
+                        client.delete_training_workload(workload_id)
```
strickvl: Hmmm so we delete the workload when we fail. Doesn't this then delete / remove all the evidence that someone would need to debug this failed workload etc? (You could maybe have this as a `config` option that someone could set or configure about whether they want these workloads to be deleted on failure or not? `cleanup_on_failure: bool` or something?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        try:
+            result = client.create_training_workload(training_request)
+        except RunAIClientError as exc:
+            raise RuntimeError(
+                f"Failed to submit step '{info.pipeline_step_name}' to Run:AI: {exc}. "
+                "Verify credentials, project name, cluster access, and quota."
+            ) from exc
+
+        self._wait_for_completion(client, result.workload_id)
+        logger.info("Run:AI step operator job completed.")
```
strickvl: This one from RP, but other step operators (Sagemaker / Vertex) call `info.force_write_logs()` before creating the workload and wondering if we should do that as well here?

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,210 @@
...
+
+* An [image builder](https://docs.zenml.io/stacks/image-builders/) to build the Docker images.
+
+We can then register the step operator:
+
+```shell
+zenml step-operator register <NAME> \
+    --flavor=runai \
+    --client_id=<YOUR_CLIENT_ID> \
+    --client_secret=<YOUR_CLIENT_SECRET> \
```
strickvl: probably we don't want to recommend people to have their secrets in open text, so maybe a hint in the docs to remind people to use the zenml secret store + have it referenced here instead?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        Returns:
+            A unique workload name string conforming to Run:AI naming requirements.
+        """
+        step_name = info.pipeline_step_name.lower().replace("_", "-")
+        pipeline_name = info.pipeline.name.lower().replace("_", "-")
+
+        base_name = f"zenml-{pipeline_name}-{step_name}"
+
+        if base_name and not base_name[0].isalpha():
+            base_name = f"z{base_name}"
```
strickvl: Not sure if we're already enforcing name sanitization somewhere else in the codebase, but here we're only handling this small substitution, but should we check for spaces or dots in the name as well? (maybe not important this one... I feel like we check this earlier in the pipeline validation on the zenml side?

strickvl (changes requested):
I like it! Mainly pointed out a bunch of small things. And def include a screenshot either in the docs and the PR description as well that we can use for marketing as well..

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+        description="Kubernetes annotations to add to the workload pod. "
+        "Example: {'prometheus.io/scrape': 'true'}",
+    )
+
+    environment_variables: Dict[str, str] = Field(
+        default_factory=dict,
+        description="Additional environment variables to set in the Run:AI workload container. "
+        "These are merged with ZenML's environment variables. "
+        "Example: {'LOG_LEVEL': 'DEBUG', 'BATCH_SIZE': '32'}",
+    )
```
stefannica: Aren't these redundant with [the general environment variables support ](https://docs.zenml.io/concepts/environment-variables#configuring-environment-variables-on-pipelines)? If yes, then I would remove these from the settings.

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+
+    client_id: str = Field(
+        ...,
+        description="Run:AI client ID for API authentication. Obtain from Run:AI "
+        "control plane under Settings > Application > Applications. Required for API access",
+    )
+    client_secret: SecretStr = Field(
+        ...,
+        description="Run:AI client secret for API authentication. Obtain from Run:AI "
+        "control plane under Settings > Application > Applications. Stored securely",
```
stefannica: how can it be stored securely if you're keeping it in the component config ? 🥲

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+    runai_base_url: str = Field(
+        ...,
+        description="Run:AI control plane base URL. For SaaS deployments, use format "
+        "'https://<organization>.run.ai'. For self-hosted, use your control plane URL. "
+        "Example: 'https://my-org.run.ai'",
```
stefannica:
```suggestion
        description="Run:AI control plane base URL. For Run:AI SaaS, use format "
        "'https://<organization>.run.ai'. For self-hosted Run:AI deployments, use your control plane URL. "
        "Example: 'https://my-org.run.ai'",
```

best to be specific, given that zenml also uses those same terms

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+        return self.generate_default_sdk_docs_url()
+
+    @property
+    def logo_url(self) -> str:
+        """A url to represent the flavor in the dashboard.
+
+        Returns:
+            The flavor logo.
+        """
+        return "https://www.run.ai/wp-content/uploads/2023/01/runai-logo.svg"
```
stefannica: The convention is that we point this to one of our own logos in the public flavor logos S3 bucket. This is important, because that bucket is downloaded and added to the Pro UI docker image in order for the logos to still work in air-gapped situations.

You have to talk to @znegrin to get a properly formatted logo first.

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+            client_id: Run:AI client ID for authentication.
+            client_secret: Run:AI client secret for authentication.
+            runai_base_url: Run:AI control plane base URL.
+        """
+        self._raw_client = self._create_client(
+            client_id, client_secret, runai_base_url
+        )
+
+    def _load_module(self, module_path: str) -> Any:
+        """Load a Run:AI SDK module.
```
stefannica: I don't understand the point of this method. If you need to import a module, why not just import them at the top of this file ? why do you need to have an allow list for this ?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        return self._raw_client
+
+    @property
+    def models(self) -> Any:
+        """Load the runai.models module.
+
+        Returns:
+            The runai.models module.
+        """
+        return self._load_module("runai.models")
```
stefannica: Same here: why wrap a property around an external module ?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        )
+
+        return [
+            RunAIProject(
+                id=str(p.get("id")),
+                name=p.get("name", ""),
+                cluster_id=p.get("clusterId"),
+            )
+            for p in projects_data
+        ]
```
stefannica: I'll only leave this question here, but it applies equally to all other places that make calls to the runapy library.

You should catch all expected and unexpected exceptions that might result from making calls to the Run:AI client or from unpacking the response values instead of letting them surface to the user in their original form. You should catch them and raise your own errors (i.e. inherited from `RunAIClientError`) instead, just as you did with `RunAIProjectNotFoundError` for example.

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+
+    workload_id: str
+    workload_name: str
+
+
+class RunAIClient:
+    """Wrapper around the runapy SDK providing typed responses.
+
+    This client encapsulates all Run:AI API interactions and provides
+    typed dataclasses instead of raw dictionaries.
```
stefannica, line 128: I love that you decided to used typed classes vs raw dictionaries.

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+            request: TrainingCreationRequest from runai.models.
+
+        Returns:
+            WorkloadSubmissionResult with the workload ID.
+
+        Raises:
+            RunAIClientError: If submission fails.
+        """
+        try:
+            response = self._raw_client.workloads.trainings.create_training1(
```
stefannica, line 420: just checking: `create_training1` is not a typo here ?

src/zenml/integrations/runai/client/runai_client.py
stefannica: if you decided to use typed classes for responses, why not do the same for requests ?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+
+        Raises:
+            RunAIClientError: If the query fails.
+        """
+        try:
+            response = self._raw_client.workloads.trainings.get_training(
+                workload_id
+            )
+            if response.data and isinstance(response.data, dict):
+                return cast(Dict[str, Any], response.data)
```
stefannica: what happened to "using dataclassess instead of raw dictionaries" here ?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+                )
+                return cast(Optional[str], status)
+            return None
+        except RunAIClientError:
+            raise
+        except Exception as exc:
+            error_msg = str(exc).lower()
+            if "not found" in error_msg or "404" in error_msg:
+                logger.warning(f"Workload {workload_id} not found")
+                return None
```
stefannica: Isn't it better to raise `RunAIWorkloadNotFoundError` here, to be consistent with the other client methods ?

src/zenml/integrations/runai/client/runai_client.py
stefannica: Same here: Isn't it better to raise `RunAIWorkloadNotFoundError` ?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,430 @@
...
+        """Delete a training workload.
+
+        Args:
+            workload_id: The workload ID to delete.
+
+        Raises:
+            RunAIClientError: If deletion fails.
+        """
+        try:
+            self._raw_client.workloads.trainings.delete_training(workload_id)
```
stefannica, line 554: what if the workload doesn't exist anymore ?

stefannica, line 554: does calling this actually wait until the workload is deleted before returning ?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+            info: Information about the step run.
+            entrypoint_command: Command that executes the step.
+            environment: Environment variables to set in the step operator
+                environment.
+
+        Raises:
+            RuntimeError: If workload submission or execution fails.
+        """
+        settings = cast(RunAIStepOperatorSettings, self.get_settings(info))
+        client = self._get_client()
```
stefannica: You should cache this using a private attribute masked by a `client` property or something. You can reuse the client across subsequent calls made to the same step operator.

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+            client: The RunAIClient instance.
+            base_environment: Base environment variables from ZenML.
+            additional_vars: Additional environment variables from settings.
+
+        Returns:
+            List of Run:AI EnvironmentVariable objects.
+        """
+        EnvironmentVariable = client.models.EnvironmentVariable
+
+        merged_env = {**base_environment, **additional_vars}
```
stefannica: This is what I meant earlier when I said you should remove the environment from the step operator settings. They are redundant.

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+            client, environment, settings.environment_variables
+        )
+
+        image_pull_secrets = self._build_image_pull_secrets(client)
+
+        command, args = self._build_command_and_args(entrypoint_command)
+
+        models = client.models
+        TrainingCreationRequest = models.TrainingCreationRequest
+        TrainingSpecSpec = models.TrainingSpecSpec
```
stefannica: Why this weird looking code instead of importing them at the top and even using them as annotations in the client class you wrote ?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        tolerations_list = self._build_tolerations(client, settings)
+        labels_list = self._build_labels(client, settings)
+        annotations_list = self._build_annotations(client, settings)
+
+        spec_dict: Dict[str, Any] = {
+            "image": image,
+            "command": command,
+            "compute": compute,
+            "environmentVariables": env_vars,
+        }
```
stefannica: why use a dict here when you have perfectly good typed classes that you can use instead: `TrainingCreationRequest` and `TrainingSpecSpec`?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+
+        Args:
+            client: The RunAIClient instance.
+            settings: The step operator settings.
+
+        Returns:
+            A SupersetSpecAllOfCompute object.
+        """
+        SupersetSpecAllOfCompute = client.models.SupersetSpecAllOfCompute
+        ExtendedResource = client.models.ExtendedResource
```
stefannica: why here instead of the top ?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        ImagePullSecret = client.models.ImagePullSecret
+        return [
+            ImagePullSecret(
+                name=self.config.image_pull_secret_name, user_credential=False
+            )
+        ]
+
+    def _build_tolerations(
+        self, client: RunAIClient, settings: RunAIStepOperatorSettings
+    ) -> Optional[List[Any]]:
```
stefannica:
```suggestion
    ) -> Optional[List[Toleration]]:
```

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+            client: The RunAIClient instance.
+            settings: The step operator settings.
+
+        Returns:
+            List of Toleration objects or None.
+        """
+        if not settings.tolerations:
+            return None
+
+        Toleration = client.models.Toleration
```
stefannica: why here ?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+                toleration_dict["value"] = t["value"]
+            if "effect" in t:
+                toleration_dict["effect"] = t["effect"]
+            tolerations_list.append(Toleration(**toleration_dict))
+
+        return tolerations_list
+
+    def _build_labels(
+        self, client: RunAIClient, settings: RunAIStepOperatorSettings
+    ) -> Optional[List[Any]]:
```
stefannica:
```suggestion
    ) -> Optional[List[Label]]:
```

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        """
+        if not settings.labels:
+            return None
+
+        Label = client.models.Label
+        return [Label(key=k, value=v) for k, v in settings.labels.items()]
+
+    def _build_annotations(
+        self, client: RunAIClient, settings: RunAIStepOperatorSettings
+    ) -> Optional[List[Any]]:
```
stefannica:
```suggestion
    ) -> Optional[List[Annotation]]:
```

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        max_missing_status_checks = 3
+
+        while True:
+            sleep_time = base_interval
+            if timeout and (time.time() - start_time) > timeout:
+                logger.warning(
+                    f"Attempting to stop timed-out workload {workload_id}"
+                )
+                try:
+                    client.delete_training_workload(workload_id)
```
stefannica: Really nice of you to stop timed out workloads instead of just bailing out here.

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+                    missing_status_retries = 0
+                    try:
+                        client.delete_training_workload(workload_id)
+                    except Exception as cleanup_exc:
+                        logger.error(
+                            f"Failed to cleanup workload {workload_id}: {cleanup_exc}"
+                        )
+                    raise RuntimeError(
+                        f"Run:AI workload {workload_id} failed with status: {status}"
+                    )
```
stefannica, line 638: the AI caught this one too: deleting failed workloads is not a good idea, because you lose traceability. Leaving them there in a failed state doesn't cost anything.

generally speaking, you should try to have a consistent behavior: if a successful workload leaves some trace in the run:ai system (i.e. like a "completed workload" entry that users can go back to) you should try to have the same for all workloads, even when they failed or have been running for too long and need to be stopped.

src/zenml/integrations/runai/step_operators/runai_step_operator.py
stefannica: Well, I said stop, but I was wrong. You are not stopping the workload, you are deleting it. Does that still keep a trace in the Run:AI system ? if not, wouldn't it be a better idea to stop it instead of deleting it ?

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,624 @@
...
+        """
+        if not settings.annotations:
+            return None
+
+        Annotation = client.models.Annotation
+        return [
+            Annotation(key=k, value=v) for k, v in settings.annotations.items()
+        ]
+
+    def _wait_for_completion(
```
stefannica, line 570: some use-cases to figure out maybe with @schustmi 's help:

1. what happens if the user decides to stop the pipeline early ? this is a feature we now support with some orchestrators. Is there a way to detect that and stop the step operator early too @schustmi ?
2. when running a dynamic pipeline with multiple step operator steps, they might run in different threads. You'll probably need to make your step operator implementation thread-safe. Are there any shared resources in particular that you need to protect against race conditions ?

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,409 @@
...
+        description="Interval in seconds to poll Run:AI API for workload status. "
+        "Lower values provide faster status updates but increase API load. "
+        "Recommended: 30-60 seconds for production",
+    )
+    workload_timeout: Optional[PositiveInt] = Field(
+        default=None,
+        description="Maximum time in seconds to wait for workload completion. "
+        "If the workload doesn't finish within this time, it will be marked as failed. "
+        "If not set, workloads can run indefinitely. Example: 3600 for 1 hour",
+    )
```
stefannica: I think this might belong in `RunAIStepOperatorSettings`. You should allow different values for different steps.

I also suggest you add a separate `pending_timeout` setting - use to configure how much time a workload is allowed to be in a "pending" state - i.e. how much time a step should wait to be scheduled by Run:AI before bailing out. Some users might set this to None and `workload_timeout` to a non-zero value or vice-versa, to get the desired effect.

docs/book/component-guide/step-operators/runai.md
safoinme: Somehow looks fine for me, what's issue exactly?

strickvl: Yeah just look at the colours that github renders them in my screenshot. In any case, it seems to be ok... I guess we'll just have to see how it renders on gitbook. I guess it's fine.

src/zenml/integrations/runai/step_operators/runai_step_operator.py
schustmi: 1. Early stopping of steps with step operators is not yet supported as their interface is all sync (submit and wait for it to finish). That will change in the future for sure, but for now that isn't a concern.
2. Yes absolutely right, this is something that can happen and something we should test to make sure it doesn't cause any problems

src/zenml/integrations/runai/client/runai_client.py
safoinme: it's not typo unfortunatly

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
safoinme: We can't unless we eithger create a toleration object that match this pydantic toleration class from run_ai models or force user to use that, not sure if it makes sense

src/zenml/integrations/runai/client/runai_client.py
safoinme: that was issue related to install/detection failure of runapy (it's named runapy but models are runai) it should be removed now, thank you for pointing it

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,476 @@
...
+            return [
+                RunAIProject(
+                    id=str(p.get("id")),
+                    name=p.get("name", ""),
+                    cluster_id=p.get("clusterId"),
+                )
+                for p in projects_data
+            ]
+        except RunAIClientError:
+            raise
```
stefannica: there's no way this code could raise the exception you defined, right?
```suggestion
```

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,476 @@
...
+
+            return [
+                RunAICluster(
+                    id=c.get("uuid", c.get("id", "")),
+                    name=c.get("name", ""),
+                )
+                for c in clusters_data
+            ]
+        except RunAIClientError:
+            raise
```
stefannica: same here
```suggestion
```

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,476 @@
...
+            cluster_id: The cluster ID to find.
+
+        Returns:
+            The matching RunAICluster or None if not found.
+        """
+        clusters = self.get_clusters()
+        for cluster in clusters:
+            if cluster.id == cluster_id:
+                return cluster
+        return None
```
stefannica: shouldn't you raise RunAIClusterNotFoundError here, just as you do in get_cluster_by_name, for consistency ?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,476 @@
...
+            response = self._raw_client.workloads.trainings.create_training1(
+                training_creation_request=request
+            )
+            workload_id = self._extract_workload_id(response)
+            return WorkloadSubmissionResult(
+                workload_id=workload_id or request.name,
+                workload_name=request.name,
+            )
+        except RunAIClientError:
+            raise
```
stefannica:
```suggestion
```

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,476 @@
...
+        Args:
+            workload_id: The workload ID to delete.
+
+        Raises:
+            RunAIClientError: If deletion fails.
+        """
+        try:
+            self._raw_client.workloads.trainings.delete_training(workload_id)
+        except RunAIClientError:
+            raise
```
stefannica:
```suggestion
```

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,476 @@
...
+        Args:
+            workload_id: The workload ID to suspend.
+
+        Raises:
+            RunAIClientError: If suspending fails.
+        """
+        try:
+            self._raw_client.workloads.trainings.suspend_training(workload_id)
+        except RunAIClientError:
+            raise
```
stefannica:
```suggestion
```

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
strickvl: @safoinme reminder on this one

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,176 @@
...
+|--------|------|----------|-------------|
+| `client_id` | str | Yes | Run:AI client ID for API authentication |
+| `client_secret` | str | Yes | Run:AI client secret for API authentication |
+| `runai_base_url` | str | Yes | Run:AI control plane URL (e.g., `https://org.run.ai`) |
+| `project_name` | str | Yes | Run:AI project name for workload submission |
+| `cluster_name` | str | No | Run:AI cluster name (uses project's cluster if not specified) |
+| `image_pull_secret_name` | str | No | Name of Run:AI image pull secret for private registries |
+| `monitoring_interval` | float | No | Interval in seconds to poll workload status (default: 30) |
+| `workload_timeout` | int | No | Maximum time in seconds for workload completion |
+| `cleanup_on_failure` | bool | No | Delete failed workloads (default: True). Set to False for debugging |
```
strickvl: Docs say `cleanup_on_failure` (default True), but code has `delete_on_failure` (default False)?

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,176 @@
...
+| `gpu_request_type` | str | "portion" | GPU allocation method: "portion" or "memory" |
+| `gpu_memory_request` | str | None | GPU memory to request (e.g., "20Gi") when using memory type |
+| `cpu_core_request` | float | 1.0 | Number of CPU cores to request |
+| `cpu_memory_request` | str | "4G" | Memory to request (e.g., "4G", "8Gi") |
+| `node_pools` | list | None | Ordered list of node pool names for scheduling |
+| `node_type` | str | None | Node type label for GPU selection |
+| `preemptibility` | str | None | "preemptible" or "non-preemptible" |
+| `priority_class` | str | None | Kubernetes PriorityClass name |
+| `large_shm_request` | bool | False | Request large /dev/shm for PyTorch DataLoader |
+| `environment_variables` | dict | {} | Additional environment variables |
```
strickvl: Are env variables available per step? I think `RunAIStepOperatorSettings` doesn't have this and `launch()` doesn't merge in the env variables.

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,466 @@
...
+                    "Available keys: %s",
+                    list(response.data.keys()),
+                )
+            return cast(Optional[str], status)
+        except RunAIClientError:
+            raise
+        except Exception as exc:
+            error_msg = str(exc).lower()
+            if "not found" in error_msg or "404" in error_msg:
+                raise RunAIWorkloadNotFoundError(workload_id) from exc
```
strickvl, line 504: This isn't super nice. Is there an `ApiException` or something that runapy exposes via a `status` property or something?

src/zenml/integrations/constants.py
strickvl: Not your code, but I just notice that `vault` is listed twice here...

strickvl (changes requested):

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,482 @@
...
+        """Create the underlying runapy client.
+
+        The runai SDK does not ship type hints, so keep the untyped call
+        isolated to this helper.
+        """
+        return RunapyClient(ApiClient(config))  # type: ignore[no-untyped-call]
+
+    @staticmethod
+    def _get_status_code(exc: Exception) -> Optional[int]:
+        """Extract an HTTP status code from an untyped Run:AI SDK exception."""
```
strickvl: no docstring? not sure how this is passing CI

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,482 @@
...
+            )
+            self._raw_client = self._create_raw_client(config)
+        except Exception as exc:
+            raise RunAIAuthenticationError(
+                f"Failed to initialize Run:AI client ({type(exc).__name__}): {exc}. "
+                "Verify your client_id, client_secret, and runai_base_url are correct."
+            ) from exc
+
+    def _create_raw_client(self, config: Configuration) -> RunapyClient:
+        """Create the underlying runapy client.
```
strickvl, line 165: same no args or returns mentioned in docstring. not sure how this passes darglint

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
```diff
@@ -0,0 +1,445 @@
...
+            The validated and normalized URL.
+
+        Raises:
+            ValueError: If the URL is invalid.
+        """
+        if not value.startswith(("http://", "https://")):
+            raise ValueError(
+                f"Invalid URL '{value}'. Must start with http:// or https://"
+            )
+        return value.rstrip("/")
```
strickvl, line 334: validates the hostname, but not the URL, so `https://` on its own would pass.

src/zenml/integrations/runai/step_operators/runai_step_operator.py
```diff
@@ -0,0 +1,690 @@
...
+        """
+        logger.info(f"Stopping workload {workload_id}: {reason}")
+        try:
+            client.suspend_training_workload(workload_id)
+        except Exception as suspend_exc:
+            logger.error(
+                f"Failed to stop workload {workload_id}: {suspend_exc}. "
+                "You may need to stop it manually in Run:AI UI."
+            )
+            return
```
strickvl: When suspend fails in _cleanup_workload, the method logs an error but returns early without attempting deletion. Perhaps we could include the runAI url for manual cleanup?

docs/book/component-guide/step-operators/runai.md
```diff
@@ -0,0 +1,177 @@
...
+* you want to leverage Run:AI's scheduling and resource management features.
+
+### Key Features
+
+* **Fractional GPU Support**: Allocate portions of a GPU (e.g., 0.25, 0.5, 0.75) to maximize resource utilization
+* **Training Workloads**: Submit individual ZenML steps as Run:AI training workloads
+* **Project-based Resource Management**: Workloads are organized by Run:AI projects with quota policies
+* **SaaS and Self-hosted**: Works with both Run:AI SaaS and self-hosted deployments
+
+![Run:AI interface](../../assets/runai-interface.png)
```
strickvl: Check it, but I think the image is at `../../.gitbook/assets/runai-interface.png` instead?

src/zenml/integrations/runai/client/runai_client.py
```diff
@@ -0,0 +1,482 @@
...
+class RunAIClientError(Exception):
+    """Base exception for Run:AI client errors."""
+
+
+class RunAIAuthenticationError(RunAIClientError):
+    """Raised when authentication with Run:AI fails."""
+
+
+class RunAIConnectionError(RunAIClientError):
+    """Raised when connection to Run:AI API fails."""
```
strickvl, line 38: I might be wrong, but I think this is never actually used?

src/zenml/integrations/runai/flavors/runai_step_operator_flavor.py
strickvl: Also maybe a unit test for this?

strickvl (approved):

PR #4453, opened 2026-01-30
Improve API transaction management for better scalability

stefannica:
## Describe changes
The current API transaction management implementation becomes inefficient when larger payloads are involved, such as those of snapshot responses that result from pipelines with a large number of steps and step configurations.

This change introduces some improvements to address these issues:

* moves the cleanup of expired transaction to its own independent thread that runs periodically. Previously, cleanup was called as part of each and every API request, which slowed down individual requests unnecessarily and also created a blocking situation when multiple API requests waited to get write access to the API transaction table
* moves the result blob from the API transaction table to its own table. Keeping the result into a separate table helps because regular operations on the main API transaction table are no longer affected by those longer-running operations on the result table, such as periodic cleanup calls.
* use gzip compression to store the result. This reduces the result size stored in the DB by 10x or more, which speeds up transfers and DB operations on that table.
* exceptions that may be thrown by the API transaction database operations are now properly caught and handled

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/server_config.py
+++ b/src/zenml/config/server_config.py
@@ -251,17 +251,8 @@ class ServerConfiguration(BaseModel):
         server_request_timeout: The timeout for server requests in seconds. If
             not specified, the default value of 20 seconds will be used. This
             value should be lower than the client's request timeout.
-        api_transaction_cleanup_batch_size: The initial batch size for cleaning
-            up expired API transactions.
         api_transaction_cleanup_interval: The interval in seconds between
             cleanup batches.
-        api_transaction_cleanup_min_batch_size: The minimum batch size for the
-            adaptive cleanup algorithm.
-        api_transaction_cleanup_max_batch_size: The maximum batch size for the
-            adaptive cleanup algorithm.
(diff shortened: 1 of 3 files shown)


src/zenml/log_stores/otel/otel_log_exporter.py
```diff
@@ -155,20 +156,13 @@ def _export(
...
-                )
-            else:
-                gzip_data = BytesIO()
-                with gzip.GzipFile(fileobj=gzip_data, mode="w") as gzip_stream:
-                    gzip_stream.write(serialized_data)
-                data = gzip_data.getvalue()
-                self._session.headers.update(
-                    {"Content-Encoding": self._compression.value}
-                )
+            gzip_data = BytesIO()
```
schustmi, line 159: The exception handling here was unnecessary because we have `gzip` as a dependency I assume? Or did something else change?

stefannica: gzip has been part of the python distribution for some time

image

PR #4456, opened 2026-02-02
Step/Pipeline replays

schustmi:
## Describe changes
This PR adds methods to replay an existing step or pipeline run. What this means concretely:
- When replaying a step, ZenML will run the step with the same inputs and configuration of the original step run.
- When replaying a run, ZenML will run the pipeline and allows users to specify a set of steps that should be skipped and re-used from the original pipeline run.

There is also a debug mode, which means the pipeline/step will run on the active stack but switch to a local orchestrator to enable easier debugging using breakpoints and other local debugging tools.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/execution/pipeline/utils.py
+++ b/src/zenml/execution/pipeline/utils.py
@@ -189,7 +189,7 @@ def skip_steps_and_prune_snapshot(
     )
 
     explicitly_skipped_steps = set(pipeline_run.config.steps_to_skip)
-    skipped_invocations = set()
+    skipped_invocations: Set[str] = set()
 
     for invocation_id, step in snapshot.step_configurations.items():
         explicitly_skipped = invocation_id in explicitly_skipped_steps


src/zenml/zen_stores/migrations/versions/bf4203afb657_pipeline_run_replays.py
```diff
@@ -0,0 +1,49 @@
...
+
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "bf4203afb657"
+down_revision = "0.93.3"
```
bcdurak: Not sure which PR would make it in first, but schedules do, we will have to change the version.

src/zenml/pipelines/pipeline_definition.py
```diff
@@ -1816,3 +1846,91 @@ def _clear_state(self) -> None:
...
+            )
+
+        original_parameters = original_run.config.parameters or {}
+        # Only include pipeline parameters of the original run for inputs that
+        # the user didn't provide a value for.
+        inputs = {
+            k: v
+            for k, v in original_parameters.items()
+            if k in self.missing_parameters
+        }
```
bcdurak, line 1922: Can you explain this part really quickly? Why are we only allowing our users to override inputs that the user didnt provide a value for during the original run?

docs/book/user-guide/tutorial/replaying-runs-steps.md
```diff
@@ -0,0 +1,132 @@
...
+# Replay a specific run
+training_pipeline.replay(pipeline_run="run_name_or_id")
+```
+
+### Skipping steps
+
+You can skip steps that don't need to be re-executed. Skipped steps reuse
+their output artifacts from the original run.
+
+```python
```
bcdurak, line 51: How does this interact with the caching? Would I still need to skip steps if I set `enable_cache` to `True`?

src/zenml/execution/pipeline/utils.py
```diff
@@ -146,3 +147,112 @@ def compute_invocation_id(
...
+                if not explicitly_skipped:
+                    logger.debug(
+                        "Not skipping successful step `%s` because upstream "
+                        "step `%s` is not skipped.",
+                        invocation_id,
+                        upstream_step,
+                    )
+                    should_skip = False
+                    break
+                raise RuntimeError(
```
bcdurak: I ran into this as I was playing around. I had step 1 and 2 feed into 3. I immediately tried to skip 3 and it told me that it can't be skipped because 1 is not skipped, I added it to the list and it told me it still can not be skipped because 2 is not skipped. Showing all the "need-to-skip" steps in one go could be a bit cleaner.

docs/book/user-guide/tutorial/replaying-runs-steps.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Automatically skip all steps that succeeded in the original run
+training_pipeline.replay(skip_successful_steps=True)
+```
+
+{% hint style="info" %}
+A step can only be skipped if all of its upstream dependencies are also skipped.
+{% endhint %}
+
+### Overriding pipeline parameters
```
bcdurak, line 63: Not sure where to put this or if this is already handled but the skipped steps in the DAG visualizer need an icon. 

Image

bcdurak (commented):
All I could find so far. Wanted to share it to speed up the process.

docs/book/user-guide/tutorial/replaying-runs-steps.md
bcdurak: Since we have discussed this already yesterday, resolving this for now.

docs/book/user-guide/tutorial/replaying-runs-steps.md
schustmi: DOne on a feature branch

src/zenml/pipelines/pipeline_definition.py
schustmi: The user might have run `pipeline.configure(...)` or `with_options(...)` before to configure some parameters. We don't want to overwrite those (=new/current) parameters with parameters from the original run.

src/zenml/execution/pipeline/utils.py
schustmi: Done

src/zenml/pipelines/pipeline_definition.py
bcdurak: Oh, I see. Makes sense. Thanks for the explanation.

PR #4459, opened 2026-02-02
Fix default sorting for CLI commands to return newest first

### LINKED ISSUE #4455: CLI list commands return oldest items first instead of newest
## Summary

`zenml pipeline runs list --size=1` returns the **oldest** run instead of the **newest**, which is counter-intuitive for most users. This affects the CLI and API defaults for several time-sensitive entities.

## Steps to Reproduce

```bash
# Create some pipeline runs over time, then:
zenml pipeline runs list --size=1
```

**Expected:** Returns the most recent pipeline run  
**Actual:** Returns the oldest pipeline run

## Root Cause Analysis

There are **competing default values** for `sort_by` in the codebase:

| Location | Default | Effective Order |
|----------|---------|-----------------|
| `BaseFilter.sort_by` | `"created"` | Ascending (oldest first) |
| `Client.list_pipeline_runs()` | `"desc:created"` | Descending (newest first) |

The issue is in how the CLI injects defaults:

1. The `@list_options(PipelineRunFilter)` decorator in `src/zenml/cli/pipeline.py` auto-generates click options from filter model field defaults
2. This means the CLI **always passes** `sort_by="created"` to the client, even when the user doesn't specify it
3. This overrides the client's sensible `"desc:created"` default

### Code Flow
```
CLI command (--size=1)
    ↓ @list_options injects sort_by="created" from PipelineRunFilter default
Client.list_pipeline_runs(sort_by="created", size=1)  
    ↓ Client's default "desc:created" is overridden
PipelineRunFilter(sort_by="created")
    ↓ No prefix means ascending (BaseFilter.sorting_params behavior)
SQL: ORDER BY created ASC, id ASC
    ↓ With LIMIT 1
Returns: oldest run ❌
```

The `BaseFilter.sorting_params` property interprets `"created"` (no `asc:`/`desc:` prefix) as ascending order.

## Workaround

Users can explicitly specify descending order:

```bash
zenml pipeline runs list --size=1 --sort_by=desc:created
```

## Proposed Fix

Override the `sort_by` default in filter models for time-sensitive entities. This is the minimal-scope fix that doesn't change global behavior.

### Primary Fix: `PipelineRunFilter`

In `src/zenml/models/v2/core/pipeline_run.py`:

```python
class PipelineRunFilter(ProjectScopedFilter, TaggableFilter, RunMetadataFilterMixin):
    """Model to enable advanced filtering of all pipeline runs."""
    
    # ... existing class vars ...
    
    # Add this field to override BaseFilter default
    sort_by: str = Field(
        default="desc:created",
        description="Which column to sort by. Default is newest runs first.",
    )
```

### Other Candidates to Review

These entities are also time-sensitive and users likely expect newest-first ordering:

| Filter Model | File | Likely Should Default to `desc:created`? |
|--------------|------|------------------------------------------|
| `StepRunFilter` | `src/zenml/models/v2/core/step_run.py` | ✅ Yes |
| `ArtifactVersionFilter` | `src/zenml/models/v2/core/artifact_version.py` | ✅ Yes |
| `ModelVersionFilter` | `src/zenml/models/v2/core/model_version.py` | ✅ Yes |
| `LogsFilter` | `src/zenml/models/v2/core/logs.py` | ✅ Yes |
| `RunMetadataFilter` | `src/zenml/models/v2/core/run_metadata.py` | ⚠️ Maybe |
| `PipelineBuildFilter` | `src/zenml/models/v2/core/pipeline_build.py` | ⚠️ Maybe |
| `ScheduleFilter` | `src/zenml/models/v2/core/schedule.py` | ⚠️ Maybe |

### Alternative: Change `BaseFilter` Default Globally

A broader fix would be to change `BaseFilter.sort_by` default from `"created"` to `"desc:created"`. However, this would affect **all** list operations globally, which may not be desired for entities like Users, Stacks, Components, etc. where alphabetical or oldest-first might be acceptable.

## Additional Context

The fact that `Client.list_pipeline_runs()` explicitly defaults to `sort_by="desc:created"` is strong evidence that **newest-first was the intended behavior** for runs. The CLI's auto-injection of filter model defaults accidentally defeats this intent.

## Related Files

- `src/zenml/models/v2/base/filter.py` — `BaseFilter.sort_by` default and `sorting_params` property
- `src/zenml/cli/utils.py` — `list_options` decorator that injects filter defaults
- `src/zenml/client.py` — Client methods with their own `sort_by` defaults
- `src/zenml/models/v2/core/pipeline_run.py` — `PipelineRunFilter`

MukeshK17:
##  Bug Fix: Default pipeline run sorting

This PR fixes a counter-intuitive behavior in the ZenML CLI where list commands returned the **oldest** item first instead of the **most recent** one.

**Breaking changes**:
- The default output order of CLI `list` commands has been changed to return the latest items first. Previously, the default was to return the oldest items first.

---
Fixes #4455


src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -717,6 +717,11 @@ class PipelineRunFilter(
         *RunMetadataFilterMixin.API_MULTI_INPUT_PARAMS,
     ]
 
+    sort_by: str = Field(
```
schustmi: This field already exists on the parent class, and while it does solve it by changing the default value of the filter, it only applies to pipeline runs and not all other CLI commands.

schustmi (changes requested):

src/zenml/models/v2/core/pipeline_run.py
schustmi: The actual issue related to the `@list_options` decorator is explained in some more detail here https://github.com/zenml-io/zenml/issues/4455, probably the best fix for this would be to make sure that decorator uses `desc:created` as a default value for the `sort_by` field.

schustmi: I also think we unfortunately can't change the default value of `BaseFilter.sort_by` because that would be a breaking change.

MukeshK17: I have made the suggested changes. Please have a look.

schustmi: @MukeshK17 Thanks, looks good to me! Even this change however is a breaking change as some users might use some CLI commands in scripts and they might rely on the existing ordering for their code to work. I have marked the PR as a breaking change, and we'll keep it open until we do another breaking release.

It might take a month or so until this PR get's merged, but we'll take care of that and no more work should be required on your end. Thanks!

MukeshK17: Thank you for the review and clarification

schustmi (approved):

MukeshK17:
Thanks for the reminder! Keeping this PR active until the next breaking release as discussed.

schustmi:
Haven't forgotten about this @MukeshK17, hopefully will be merged soon! :)

schustmi:
@MukeshK17 Thanks for the PR again, finally it made it in :)

PR #4465, opened 2026-02-04
Resource pools

schustmi:
## Describe changes
This PR adds schemas, models, endpoints, store interfaces and SDK methods and CLI commands to manage resource pools and associated objects.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/resource_settings.py
+++ b/src/zenml/config/resource_settings.py
@@ -185,7 +185,13 @@ class ResourceSettings(BaseSettings):
 
     @property
     def empty(self) -> bool:
-        """Returns if this object is "empty" (=no values configured) or not.
+        """Returns if this object is "empty" or not.
+
+        A ResourceSettings instance is considered empty if none of the
+        generic resource-related values are configured. This excludes the
+        preemptible flag and the pool_resources map, which are only relevant
+        for resource pool scheduling and currently ignored by workload
+        scheduling stack components like orchestrators and step operators.
 
         Returns:
             `True` if no values were configured, `False` otherwise.
--- a/tests/unit/config/test_resource_settings.py
+++ b/tests/unit/config/test_resource_settings.py
@@ -100,6 +100,7 @@ def test_resource_config_memory_conversion():
 def test_pool_resources_empty_dict_merges_to_empty() -> None:
     """Empty pool_resources is accepted; merged map has no keys from the map."""
     rs = ResourceSettings(pool_resources={})
+    assert rs.empty is True
     assert rs.merged_requested_resources() == {}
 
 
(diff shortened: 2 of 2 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3818,148 +3844,3578 @@ def update_stack_component(
...
+            return self.filter_and_paginate(
+                session=session,
+                query=query,
+                table=ResourcePoolSchema,
+                filter_model=filter_model,
+                hydrate=hydrate,
+                apply_query_options_from_schema=True,
+            )
+
+    def list_resource_pool_queue_items(
```
stefannica: Unused

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -3818,148 +3844,3578 @@ def update_stack_component(
...
+                ResourcePoolQueueItem(
+                    request=item.request.to_model(
+                        include_metadata=True, include_resources=True
+                    ),
+                    priority=item.priority,
+                )
+                for item in queue_items
+            ]
+
+    def list_resource_pool_allocations(
```
stefannica: Also unused

src/zenml/orchestrators/step_launcher.py
```diff
@@ -716,3 +734,93 @@ def _run_step_in_current_thread(
...
+            component: The component that is requesting the resources.
+
+        Raises:
+            RuntimeError: If the resources could not be acquired.
+        """
+        # TODO: Bool on component that indicates if we should request resources?
+        # Probably don't need this if we request resources on the step run
+        # request instead.
+        requested_resources = component.get_requested_resources(step_run_info)
+        requested_resources["step_run"] = 1
```
stefannica: What happens with the existing pipeline code that includes resource settings when they upgrade their server without defining resource pools ?

What happens if resource pools are defined, but do not include any step_run resources ?

This code also needs to be extended to allow for custom resources to be specified in ResourceSettings.

stefannica: > What happens with the existing pipeline code that includes resource settings when they upgrade their server without defining resource pools ?

> What happens if resource pools are defined, but do not include any step_run resources ?

The answer to both these questions is this: if there are no resource pools attached to any components in the stack that regulate one or more of the resources in the resource settings, they will be ignored and the step execution continues as if nothing happens. Correct @schustmi ?

stefannica: I missed the exceptions coded in `RESOURCE_POOL_UNBOUNDED_KEYS`: these resource keys are skipped from checks if none of the pool policies include them.

schustmi: You got everything correct @stefannica :)
- If no pool is attached to a component (the default for everything), it will simply run.
- If pool(s) are attached, the behaviour depends on the resource keys (for some keys, no explicit resources in the pool means unlimited. For all others, no explicit resources specified for the pool means 0 available)

src/zenml/zen_stores/sql_zen_store.py
schustmi: The and the other I think you mentioned in our chat earlier. They were left in some intermediate state where I wasn't sure how to best expose them to the frontend.

tests/unit/zen_stores/test_resource_request_pool_lifecycle.py
```diff
@@ -0,0 +1,2016 @@
+from copy import deepcopy
```
schustmi, line 1: I guess this file we also need to move to the plugins repo?

PR #15, opened 2021-01-18
Zenml cli

njunge94:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the **CONTRIBUTING.md** document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

This PR contains zenml cli functionality, included:

* zenml pipeline-commands, which let you list, get and search pipelines
* zenml datasource commands, which let you list, get and create datasources
* zenml steps commands, which let you list steps by their versions and types
* zenml config list, a command that lists the whole ZenML config for the current repository

A quick review is much appreciated, as it's my first try on CLI building code.

PR #4471, opened 2026-02-06
Fix systemic CI flakiness across multiple test clusters

### LINKED ISSUE #12: Add Google's Model card toolkit support to collect the model metrics and build the model documentation automatically
**Is your feature request related to a problem? Please describe.**

Using this [Toolkit](https://ai.googleblog.com/2020/07/introducing-model-card-toolkit-for.html) will help users of zenml align with best practices in the industry for model reporting. This will also help with increased transparency in the workflow and being more accountable towards the stakeholders in the project and the users.

**Describe the solution you'd like**
A suggested pseudo code from @hamzamaiot  could look like this:
```
# pseudo-code
model_card = mct.scaffold_assets()
# fill in ur model card details
pipeline.add_model_card(model_card)
```

**Describe alternatives you've considered**
Alternatively one could use [Dalex](https://github.com/ModelOriented/DALEX) as a model agnostic tool to build the model card self. [Here](https://github.com/ModelOriented/DALEX-docs/blob/master/jupyter-notebooks/dalex-tensorflow.ipynb) is an example in a jupyter notebook. Otherwise, can you view it [here](https://nbviewer.jupyter.org/github/ModelOriented/DALEX-docs/blob/master/jupyter-notebooks/dalex-tensorflow.ipynb)

strickvl:
## Summary

Fixes systemic CI flakiness across multiple test clusters. Each fix targets a specific, reproducible failure.

**Review status:** Reviewed with Michael (2026-02-17). Two high-risk changes were removed from this PR and spun out:
- **StepLauncher signal handler refactor** — removed, tracked for separate investigation
- **Integration activation exception handling** — spun out to #4490

### Known remaining CI failures (addressed in follow-up PRs)

| Failure | Cause | Follow-up |
|---------|-------|-----------|
| Windows integration activation crash | Torch DLL `OSError` poisons all integrations | #4490 |
| Noisy signal handler traces on CI timeout | StepLauncher handler chain accumulation | Separate investigation planned |

---

#### 1. Neural prophet dtype crash (Linux docker/mysql)
Bundled a local CSV dataset and added explicit dtype enforcement to eliminate the `TypeError` caused by pandas/numpy version drift and network-fetched data.

#### 2. Windows: skip pytorch-family integrations in CI install
Skip `pytorch`, `neural_prophet`, `pytorch_lightning` during Windows CI install since they have known DLL issues.

#### 3. Pin pandas<3 in neural_prophet integration
`pandas` 3.0 removed `Series.view()`, which `neuralprophet` 0.4.2 still calls. Pin `pandas<3` in the neural_prophet integration requirements.

#### 4. Pin setuptools<82 in CI setup + migration tests
`setuptools` 82.0.0 removed `pkg_resources`. Old ZenML versions and `deepchecks` still import it, crashing macOS/Windows integration tests. Pin `setuptools<82` in both CI setup action and migration test script.

#### 5. Disable whylogs analytics & Rich traceback in CI
The whylogs integration spawns background HTTP threads for analytics. During interpreter shutdown these threads raise exceptions that Rich's `sys.excepthook` tries to render to piped stderr, causing exit code 1 on Python 3.10. Fix: set `WHYLOGS_NO_ANALYTICS=True` and `ZENML_ENABLE_RICH_TRACEBACK=false` in CI workflows and test subprocess environments.

#### 6. Lazy import torch in PyTorch materializer
Move `import torch` from module level to inside `load`/`save` methods to avoid eager import side effects.

#### 7. Parallel pipeline test: fix deadlock + OS-conditional parallelism
The parallel tests spawned subprocesses with `stdout=subprocess.PIPE` + `.wait()`, which deadlocks when output exceeds ~64 KB. Replaced with temp-file redirection and added 15-minute timeout. Run count is now OS-conditional: 10 on Linux (full stress test), 5 on macOS (resource-constrained runners).

#### 8. Heartbeat test: mock instead of real KeyboardInterrupt
The heartbeat worker test relied on catching a real `KeyboardInterrupt` from `_thread.interrupt_main()`, which leaked into the test runner. Replaced with a mock that sets a `threading.Event`.

#### 9. PyTorch Docker test crash (getpass.getuser / uid 1001)
PyTorch 2.6+ eagerly imports TorchInductor cache, calling `getpass.getuser()` which fails in Docker containers with numeric UIDs. Set `TORCHINDUCTOR_CACHE_DIR` via `__ZENML__` prefix propagation.

#### 10. macOS: install integrations + reruns in ci-slow
Add `install_integrations: 'yes'` and `reruns: 1` to macOS unit test invocation.

#### 11. CLI: restore strict error for unknown integration install
Changed `install()` to use `error()` instead of `warning()` when a nonexistent integration name is passed. Updated test to expect non-zero exit code. (Previously this was a permissive `warning()` — Michael's review flagged this as incorrect behavior.)

#### 12. Materializer tests: replace importorskip with explicit Windows skip
Removed `pytest.importorskip` guards from 3 materializer tests (pytorch dataloader, pytorch module, neural prophet) and replaced with direct imports + `pytest.mark.skipif(platform.system() == "Windows")`. This ensures missing dependencies surface as real failures on Linux/macOS instead of silent test skips. (**Note:** the `skipif` marker was later fixed in #15 below.)

#### 13. Link checker + zizmor SHA fix
Exempt `modal.com` 406 and `ghcr.io` 502 from link checker. Fix zizmor SHA-tag mismatch in stale-prs.yml.

#### 14. US spelling convention docs
Fix spelling examples in CLAUDE.md/AGENTS.md. Add .gitignore exception for bundled test CSV.

#### 15. Fix Windows collection crash: module-level pytest.skip
Fix #12 used `pytestmark = pytest.mark.skipif(...)` which is evaluated *after* module import — so the top-level `from neuralprophet import NeuralProphet` crashed during pytest collection before the skip marker was ever checked. Replaced with `pytest.skip(allow_module_level=True)` placed *before* the optional imports, which fires during collection and gracefully skips the module on Windows. This was the cause of the Windows `integration-tests-slow` CI failure (`ModuleNotFoundError: No module named 'neuralprophet'`).

#### 16. Fix zizmor impostor-commit audit + broken relative link
Exempt `require-release-label.yml` from zizmor's `impostor-commit` audit since `mheap/github-action-required-labels` returns 401 on `git-upload-pack`, preventing tag verification. Fix broken relative link to Kubernetes orchestrator docs in scheduling.md (`../` → `../../`).

## Changes

| File | What |
|------|------|
| `tests/integration/functional/pipelines/test_pipeline_parallel.py` | Replace PIPE+wait with temp-file redirection; timeout; OS-conditional parallelism (10 Linux / 5 macOS) |
| `tests/unit/steps/test_heartbeat_worker.py` | Mock `_thread.interrupt_main` instead of real KeyboardInterrupt |
| `src/zenml/cli/integration.py` | `warning` → `error` for nonexistent integration install |
| `tests/integration/functional/cli/test_integration.py` | Expect non-zero exit for nonexistent integration |
| `src/zenml/integrations/pytorch/materializers/base_pytorch_materializer.py` | Lazy `import torch` in methods |
| `src/zenml/integrations/neural_prophet/__init__.py` | Add `pandas<3` to requirements |
| `tests/integration/examples/neural_prophet/steps/data_loader/data_loader_step.py` | Local CSV + dtype casting |
| `tests/integration/examples/neural_prophet/data/SF_PV.csv` | Bundled test dataset (100 rows) |
| `.gitignore` | Exception for bundled test CSV |
| `scripts/install-zenml-dev.sh` | Skip pytorch-family on Windows |
| `scripts/test-migrations.sh` | Pin `setuptools<82` for old ZenML versions |
| `.github/actions/setup_environment/action.yml` | Pin `setuptools<82` in shared CI setup |
| `.github/workflows/ci-slow.yml` | macOS: install_integrations + reruns |
| `.github/workflows/integration-test-{fast,slow}{,-services}.yml` | ZENML_ENABLE_RICH_TRACEBACK=false, WHYLOGS_NO_ANALYTICS=True |
| `.github/workflows/stale-prs.yml` | Fix zizmor SHA-tag mismatch |
| `.github/zizmor.yml` | Exempt require-release-label.yml from impostor-commit audit |
| `docs/book/how-to/steps-pipelines/scheduling.md` | Fix broken relative link to K8s orchestrator docs |
| `tests/integration/examples/test_pytorch.py` | TORCHINDUCTOR_CACHE_DIR env |
| `tests/integration/examples/test_neural_prophet.py` | TORCHINDUCTOR_CACHE_DIR env |
| `tests/integration/examples/utils.py` | Propagate Rich/whylogs env to subprocesses |
| `tests/integration/integrations/*/materializers/test_*.py` | Module-level `pytest.skip(allow_module_level=True)` + direct imports (3 files) |
| `docs/link_checker.py` | Exempt modal.com 406, ghcr.io 502 |
| `CLAUDE.md`, `AGENTS.md` | Fix US spelling examples |

### Removed from this PR (per review)
| File | Why | Follow-up |
|------|-----|-----------|
| `src/zenml/orchestrators/step_launcher.py` | Too much change in high-consequence code | Separate investigation |
| `src/zenml/integrations/registry.py` | Risk of hiding real failures | #4490 |

## Test plan

- [ ] ci-fast integration shards pass (parallel test deadlock + neural prophet fixes)
- [ ] ci-slow macOS integration tests pass (setuptools pin + parallelism)
- [ ] ci-slow macOS unit tests complete (heartbeat mock)
- [ ] ci-slow pytorch Docker tests pass (TORCHINDUCTOR_CACHE_DIR)
- [ ] ci-slow ubuntu whylogs tests pass (Rich traceback + analytics disabled)
- [ ] sqlite-db-migration-testing-random passes (setuptools pin)
- [ ] typos spell checker passes
- [ ] CLI test for nonexistent integration expects failure (exit != 0)
- [ ] Windows integration-tests-slow passes (module-level pytest.skip fix)
- [ ] zizmor security scan passes (impostor-commit exemption)
- [ ] check-relative-links passes (scheduling.md link fix)
- [ ] **Windows integration activation may still have issues at runtime** (separate from collection fix; tracked in #4490)

### DIFF
--- a/tests/unit/utils/test_networking_utils.py
+++ b/tests/unit/utils/test_networking_utils.py
@@ -15,7 +15,7 @@
 import socket
 
 import hypothesis.strategies as st
-from hypothesis import given
+from hypothesis import assume, given
 
 from zenml.utils import networking_utils
 
@@ -79,6 +79,7 @@ def test_replace_internal_hostname_works(
     sample_text,
 ):
     """Test method returns url when not internal hostname."""
+    assume(sample_text not in ("host.docker.internal", "host.k3d.internal"))
     response1 = networking_utils.replace_internal_hostname_with_localhost(
         sample_text
     )


tests/integration/examples/utils.py
```diff
@@ -69,10 +69,19 @@ def __call__(self, *args: str) -> None:
...
         """
+        env = os.environ.copy()
+        # Disable Rich traceback in subprocesses so that if an exception
+        # occurs during interpreter shutdown, the default excepthook
+        # writes the traceback before the pipe closes (Rich's handler
+        # can fail with "Error in sys.excepthook:" on Python 3.10).
+        env.setdefault("ZENML_ENABLE_RICH_TRACEBACK", "false")
+        # Disable whylogs anonymous usage stats to avoid background HTTP
+        # threads that can raise exceptions during interpreter shutdown.
+        env.setdefault("WHYLOGS_NO_ANALYTICS", "True")
```
schustmi, line 80: These should be set by the GH workflows now already I think? So this is for the local case I assume?

schustmi (approved):

tests/integration/examples/utils.py
strickvl: Correct. Both `ZENML_ENABLE_RICH_TRACEBACK` and `WHYLOGS_NO_ANALYTICS` are already set in `integration-test-fast.yml`, `integration-test-slow.yml`, and both -services variants. So in CI these are already covered. The `setdefault()` here is for the local case (when someone runs the integration example tests outside of GH Actions, the subprocess wouldn't inherit these unless they're explicitly set).

PR #4481, opened 2026-02-10
Feature/improve filter and assoc models

Json-Andriopoulos:
## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/schemas/trigger_schemas.py
+++ b/src/zenml/zen_stores/schemas/trigger_schemas.py
@@ -227,6 +227,7 @@ class TriggerSchema(NamedSchema, table=True):
             configuration=trigger_request.get_config(),
             flavor=trigger_request.flavor,
             type=trigger_request.type,
+            concurrency=trigger_request.concurrency,
         )
 
         for field_name, value in extra_fields.items():
@@ -248,6 +249,9 @@ class TriggerSchema(NamedSchema, table=True):
 
         if trigger_update.name is not None:
             self.name = trigger_update.name
+
+        if trigger_update.concurrency is not None:
+            self.concurrency = trigger_update.concurrency
 
         self.configuration = trigger_update.get_config()
 
@@ -285,6 +289,7 @@ class TriggerSchema(NamedSchema, table=True):
             type=TriggerType(self.type),
             flavor=TriggerFlavor(self.flavor),
             name=self.name,
+            concurrency=self.concurrency,
             **json.loads(self.configuration),
         )
 
--- a/docker/zenml-server-dev.Dockerfile
+++ b/docker/zenml-server-dev.Dockerfile
@@ -88,7 +88,7 @@ COPY --chown=$USERNAME:$USER_GID src/zenml_cli/__init__.py ./src/zenml_cli/
 # NOTE: we uninstall zenml at the end because we install it separately in the
 # final stage
 RUN pip install --upgrade pip uv setuptools \
-    && uv pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex,native-schedules] "alembic==1.15.2" \
+    && uv pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex] "alembic==1.15.2" \
     && uv pip uninstall zenml \
     && uv pip freeze > requirements.txt
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ local = [
     "sqlmodel==0.0.18",
 ]
 server = [
-    "zenml[local]",  # Includes all the DB dependencies
+    "zenml[local,native-schedules]",  # Includes all the DB dependencies
     "fastapi>=0.100,<=0.115.8",
     "uvicorn[standard]>=0.17.5",
     "python-multipart~=0.0.9",
@@ -80,6 +80,7 @@ server = [
     "secure~=1.0.1",
     "tldextract~=5.1.0",
     "itsdangerous~=2.2.0",
+    "croniter>=6.0.0",
 ]
 jupyter = ["rich[jupyter]>=12.0.0"]
 templates = ["copier>=8.1.0", "jinja2-time>=0.2.0,<0.3.0", "ruff>=0.1.7", "pyyaml-include<2.0"]
@@ -169,9 +170,6 @@ dev = [
     "types-termcolor>=1.1.2",
     "types-psutil>=5.8.13",
     "types-passlib>=1.7.7",
-]
-native-schedules = [
-    "croniter>=6.0.0",
 ]
 
 [tool.pytest.ini_options]


src/zenml/enums.py
```diff
@@ -572,3 +572,10 @@ class TriggerFlavor(StrEnum):
     """Enum representing trigger flavors."""
 
     NATIVE_SCHEDULE = "native schedule"  # TODO: extend with new flavors
+
+
+class TriggerRunConcurrency(StrEnum):
+    """Enum representing trigger run concurrency."""
+
+    SKIP = "skip"
+    SUBMIT = "submit"
```
schustmi, line 581: I guess there are some more options here, like "stop the current one and submit" or "submit up to a certain amount". Probably not something we want initially though right?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -7044,6 +7043,30 @@ def get_trigger(
                 include_metadata=hydrate, include_resources=True
             )
 
+    def list_triggers_unscoped(
+        self,
+        triggers_filter_model: "NonScopedTriggerFilter",
```
schustmi: Maybe also call this `Unscoped`?

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -142,6 +148,30 @@ class TriggerSchema(NamedSchema, table=True):
...
+                select(TriggerExecutionSchema)
+                .where(TriggerExecutionSchema.trigger_id == self.id)
+                .order_by(desc(TriggerExecutionSchema.created_at))
+                .limit(1)
+            )
+
+            return session.execute(stmt).scalars().one_or_none()
+        else:
+            raise RuntimeError(
+                "Missing DB session to fetch latest run for pipeline."
```
schustmi:
```suggestion
                "Missing DB session to fetch latest execution for trigger."
```

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -142,6 +148,30 @@ class TriggerSchema(NamedSchema, table=True):
         description="The next occurrence. Applicable for schedules.",
     )
 
+    @property
+    def latest_execution(self) -> TriggerExecutionSchema | None:
```
schustmi: Seems like we're only using this to get the actual latest run of that trigger execution. If so, that currently runs two queries:
- One to fetch the trigger execution
- One to fetch the run

We could optimize that by simply doing that in one query here I assume?

src/zenml/zen_stores/schemas/trigger_assoc.py
```diff
@@ -38,14 +40,14 @@ class TriggerSnapshotSchema(SQLModel, table=True):
         ),
     )
 
-    trigger_id: int = Field(
+    trigger_id: UUID = Field(
```
schustmi: We have a util function to create foreign keys. We introduced that at some point to have fixed FK names, as otherwise deleting that FK in some future migration is much more painful.

src/zenml/models/v2/core/triggers.py
```diff
@@ -120,16 +131,109 @@ class TriggerResponseResources(ProjectScopedResponseResources):
...
+    )
+    active: bool | None = Field(
+        default=None,
+        description="Whether the trigger should be active.",
+    )
+    is_archived: bool = Field(
+        default=False,
+        description="Whether the trigger should be archived.",
+    )
+    flavor: TriggerFlavor | None = Field(
```
schustmi: I think having `str` as an additional type here enables additional filter options like `contains`, `oneof`.

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -154,6 +158,8 @@ def run_snapshot(
         sync: Whether to run the snapshot synchronously.
         template_id: The ID of the template from which to create the snapshot
             request.
+        create_new_snapshot: Whether to create a new, copy snapshot.
+        implicit_auth_context: Whether to create a new, implicit auth context.
```
schustmi: Either I'm misunderstanding, or the docstring here is wrong: If True, this simply uses the implicit auth context. Otherwise, it explicitly sets it.

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -214,7 +224,10 @@ def run_snapshot(
 
     ensure_async_orchestrator(snapshot=snapshot_request, stack=stack)
 
-    new_snapshot = zen_store().create_snapshot(snapshot_request)
+    if create_new_snapshot:
+        target_snapshot = zen_store().create_snapshot(snapshot_request)
+    else:
```
schustmi, line 229: I think here we have to check that the run configuration doesn't contain any values that would require creating a new snapshot. Probably not an issue right now as it's never provided, but just to be sure.

Also: Currently, when running a snapshot from the server, we always create a new snapshot in this method, and set the original one as the `source_snapshot`. This source snapshot is later used in two places:
- It's included in `PipelineRunResponeResources.source_snapshot`, to show the snapshot that a run was triggered from in the dashboard (see e.g. the snapshot column here: https://staging.cloud.zenml.io/workspaces/dev/projects/default/pipelines/2a1d0f35-d473-44c2-8fd4-c3f9b60b1436/runs)
- It's used to compute the `latest_run` of a snapshot (see https://staging.cloud.zenml.io/workspaces/dev/projects/default/pipelines/2a1d0f35-d473-44c2-8fd4-c3f9b60b1436/snapshots, implemented in the `PipelineSnapshotSchema.latest_run` property)

Currently both of these would not work with triggers:
- A run that was triggered by any trigger would not have a `source_snapshot`.
- Those runs would not show up as the `latest_run` of a snapshot

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -164,6 +170,10 @@ def run_snapshot(
     Returns:
         ID of the new pipeline run.
     """
+    if not implicit_auth_context:
```
schustmi, line 173: Just for my understanding, this is because we're calling this from the executor which is not receiving any user requests and therefore doesn't have the auth context set.

Why do we even need the auth context set in that case? Can't we simply use the one that is already passed to this function, or is it used somewhere deeply hidden that makes this complicated to implement?

src/zenml/client.py
```diff
@@ -3959,6 +3978,77 @@ def get_schedule_trigger(
...
+        project: UUID | str | None = None,
+        id: UUID | str | None = None,
+        created: datetime | None = None,
+        updated: datetime | None = None,
+        name: str | None = None,
+        active: bool | None = None,
+        concurrency: str | None = None,
+        is_archived: bool = False,
+        flavor: TriggerFlavor = TriggerFlavor.NATIVE_SCHEDULE,
+        type: Literal[TriggerType.SCHEDULE] = TriggerType.SCHEDULE,
```
schustmi: This line can be removed and simply hardcoded in the filter.

Not sure about the above one, what's the difference between type and flavor exactly? It seems like they're both just schedules now.

src/zenml/cli/trigger.py
```diff
@@ -0,0 +1,300 @@
...
+logger = get_logger(__name__)
+
+
+@cli.group(cls=TagGroup, tag=CliCategories.MANAGEMENT_TOOLS)
+def trigger() -> None:
+    """Interact with triggers."""
+
+
+@trigger.group()
+def schedule() -> None:
```
schustmi, line 39: This will be a little confusing I guess with our currently existing schedule CLI as well 😅

src/zenml/enums.py
Json-Andriopoulos: Correct, this will be extended with more options in next iterations.

src/zenml/cli/trigger.py
Json-Andriopoulos: Yy but unavoidable unfortunately. Will try to cover it in the documentation. My plan is to effectively mark as deprecated the current schedules feature when triggers will have support for orchestrator-based schedules in which case users can use the triggers feature for both types.

src/zenml/zen_server/pipeline_execution/utils.py
Json-Andriopoulos: `or is it used somewhere deeply hidden that makes this complicated to implement` that was the main point for this, yes.

src/zenml/zen_server/pipeline_execution/utils.py
Json-Andriopoulos: I think it is more important to avoid creating copy snapshots at this point since this can generate a large volume of snapshot objects.

If I am not mistaken, current scheduled runs have the same limitation no? In which case I would try and improve it in the next iteration (probably by extending PipelineSnapshotSchema.latest_run to include this type of runs).

PR #4483, opened 2026-02-10
Add `concurrency_policy` and `starting_deadline_seconds` to K8s `CronJob` config

### LINKED ISSUE #4480: Kubernetes orchestrator: allow configuring additional CronJob spec fields
## Description

When using the Kubernetes orchestrator with scheduled pipelines, ZenML creates CronJobs via `build_cron_job_manifest()`. Currently, only a subset of CronJob spec fields are configurable through `KubernetesOrchestratorSettings`:

- `successful_jobs_history_limit`
- `failed_jobs_history_limit`

However, the `V1CronJobSpec` API supports several other useful fields that users may need to set for compliance or operational reasons:

- **`concurrencyPolicy`** — controls whether concurrent job executions are allowed (`Allow`, `Forbid`, `Replace`)
- **`startingDeadlineSeconds`** — the deadline in seconds for starting a job if it misses its scheduled time

These fields are not currently exposed in `KubernetesOrchestratorSettings` and there is no generic pass-through mechanism for CronJob-level spec fields (the existing `additional_pod_spec_args` only applies to `V1PodSpec` attributes).

## Proposed Solution

Add optional fields to `KubernetesOrchestratorSettings`:

```python
concurrency_policy: Optional[str] = Field(
    default=None,
    description="Concurrency policy for scheduled pipeline CronJobs. "
    "Valid values: 'Allow' (default), 'Forbid', 'Replace'",
)
starting_deadline_seconds: Optional[int] = Field(
    default=None,
    description="Deadline in seconds for starting a job if it misses "
    "its scheduled time",
)
```

Then wire these through to `build_cron_job_manifest()` in `manifest_utils.py`.

This is a small, backwards-compatible change (new optional fields with `None` defaults).

## Affected Files

- `src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py`
- `src/zenml/integrations/kubernetes/manifest_utils.py`
- `src/zenml/integrations/kubernetes/orchestrators/kubernetes_orchestrator.py`

strickvl:
## Summary

- Expose `concurrency_policy` and `starting_deadline_seconds` as optional fields in `KubernetesOrchestratorSettings` for scheduled pipelines backed by Kubernetes CronJobs
- Add a Pydantic validator that normalizes `concurrency_policy` casing (e.g. `"forbid"` → `"Forbid"`) and rejects invalid values early
- Wire both fields through `build_cron_job_manifest()` into the `V1CronJobSpec` constructor
- Update Kubernetes orchestrator and scheduling docs with new settings and usage examples
- Add manifest builder and validator tests

Closes #4480

## Test plan

- [ ] CI passes (formatting, mypy, tests)
- [ ] New tests in `test_manifest_utils.py` verify CronJob manifest fields are set correctly and default to `None`
- [ ] New tests in `test_kubernetes_orchestrator.py` verify the `concurrency_policy` validator accepts valid values (case-insensitive) and rejects invalid ones
- [ ] Manual: deploy a scheduled pipeline with `concurrency_policy="Forbid"` and `starting_deadline_seconds=20`, verify the CronJob spec via `kubectl get cronjob <name> -o yaml`

### DIFF
--- a/src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
+++ b/src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
@@ -87,6 +87,20 @@ class KubernetesOrchestratorSettings(BaseSettings):
         default=None,
         description="Number of failed scheduled jobs to retain in history.",
     )
+    concurrency_policy: Optional[str] = Field(
+        default=None,
+        description="CronJob concurrency policy for scheduled pipelines. "
+        "Controls whether concurrent job executions are allowed. "
+        "Valid values: 'Allow', 'Forbid', 'Replace'. "
+        "Only applies when a pipeline has a cron schedule",
+    )
+    starting_deadline_seconds: Optional[NonNegativeInt] = Field(
+        default=None,
+        description="CronJob starting deadline in seconds for scheduled "
+        "pipelines. If a scheduled run misses its trigger time, it can "
+        "still start within this window. Only applies when a pipeline "
+        "has a cron schedule",
+    )
     ttl_seconds_after_finished: Optional[NonNegativeInt] = Field(
         default=None,
         description="Seconds to keep finished jobs before automatic cleanup.",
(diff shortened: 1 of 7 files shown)


htahir1:
Do you want @schustmi to review this? I looked through it and it looks good

strickvl:
> Do you want @schustmi to review this? I looked through it and it looks good

Eventually yes but I'm aware this week is a crunch week for everyone so trying not to add to everyone's plates until next week.

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -212,6 +226,33 @@ class KubernetesOrchestratorSettings(BaseSettings):
         ("pod_name_prefix", "job_name_prefix"),
     )
 
+    @field_validator("concurrency_policy", mode="before")
```
schustmi, line 229: I'm wondering if we need this. Does kubernetes fail when an invalid value is provided?

Just in case they ever decide to add another concurrency policy, we wouldn't need to adjust our code to allow it.

strickvl: So I think maybe I agree it's simpler for us not to validate. But it's probably nicer for the user if we fail fast rather than them hitting an error after it's run for a while? Also the concurrency policy set has been stable in the K8S API for a long while.

Elsewhere in our API we somehow validate these small things but we do it with Enums. (e.g. `KubernetesServiceType`). (The bigger looser objects we don't validate).

Fine with going either way. Your call.

schustmi: Yeah I'm fine with both as well, this way it catches issues before building the Docker images.

schustmi (approved):

PR #4496, opened 2026-02-17
Include git submodule files in code archive

### LINKED ISSUE #4408: Git submodules not included when downloading code from artifact store

_No response_

### System Information

ZenML Version: 0.93.0
Python Version: 3.11.14
Operating System: Linux (Ubuntu 22.04)
Git Version: 2.x

### What happened?

When using ZenML with `allow_download_from_artifact_store: true` and git submodules, the code archive uploaded to the artifact store is incomplete. Submodule files are missing, causing `FileNotFoundError` when the pipeline runs.

**Root Cause:**
ZenML's `CodeArchive` class (in `zenml/utils/code_utils.py`) uses `git ls-files` to discover files for archiving. However, it doesn't use the `--recurse-submodules` flag, so it only captures the submodule directory reference, not the actual files inside submodules.

**Result:** Archive is ~60 KB instead of ~4 MB, missing all submodule content.

**Expected:** ZenML should include submodule files when archiving code for artifact store upload.

**Why Git Submodules Matter:**
For rapidly-evolving external dependencies, git submodules allow:
- Pinning to specific commits while tracking upstream changes
- Testing different versions without waiting for package releases

Alternative approaches don't work well:
- Local packages (`pip install -e ../project`) require Docker image rebuilds on every change
- Vendoring loses git history and makes syncing with upstream difficult

### Workaround

Manually patch `zenml/utils/code_utils.py` (around line 110) - see proposed fix below for the code change.

In `zenml/utils/code_utils.py`, modify the `CodeArchive.get_files()` method (around line 110) to support git submodules:

```python
def get_files(self) -> Dict[str, str]:
    if repo := self.git_repo:
        try:
            # Split commands to support --recurse-submodules
            tracked = repo.git.ls_files("--cached", "--recurse-submodules", self._root)
            untracked = repo.git.ls_files("--others", "--modified", "--exclude-standard", self._root)
            result = tracked + "\n" + untracked
        except Exception as e:
            logger.warning("Failed to get non-ignored files from git: %s", str(e))
            all_files = self._get_all_files(archive_root=self._root)
        else:
            for file in result.split():
                if not file.strip():
                    continue
                file_path = os.path.join(repo.working_dir, file)
                path_in_archive = os.path.relpath(file_path, self._root)
                if os.path.exists(file_path):
                    all_files[path_in_archive] = file_path
```

**Note:** `--recurse-submodules` cannot be combined with `--others` or `--modified`, so commands must be split.

**Backward Compatibility:** This change is backward compatible - repositories without submodules should work exactly as before.

### Reproduction steps

1. Add a git submodule to your ZenML project: `git submodule add <url> external/mylib`
2. Configure Docker settings with `allow_download_from_artifact_store: true`
3. Create a pipeline that imports from the submodule
4. Run pipeline → fails with `ModuleNotFoundError` or `FileNotFoundError`
5. Check archive size → only ~60 KB (missing submodule files)
6. Apply workaround patch → archive becomes ~4 MB, pipeline succeeds

### Relevant log output

```shell
FileNotFoundError: Submodule not found at /app/code/external/mylib
Run: git submodule update --init

or

ModuleNotFoundError: No module named 'mylib'
```

### Code of Conduct

apoorvdarshan:
## Summary
- Split the `git ls-files` call in `CodeArchive.get_files()` into two separate calls: one with `--recurse-submodules` for tracked files (including submodule contents) and one for untracked/modified files
- This is necessary because `--recurse-submodules` is only compatible with `--cached` mode per git documentation
- Backwards compatible: repos without submodules work exactly as before
- Only affects the artifact store upload path (`allow_download_from_artifact_store`), not the code repository download path

Fixes #4408

## Test plan
- [x] Unit tests verifying `--recurse-submodules` is passed in the tracked-files call
- [x] Unit tests verifying untracked/modified files are still included via the second call
- [x] Unit test verifying fallback behavior when git commands fail
- [ ] Manual verification with a repo containing git submodules and `allow_download_from_artifact_store: true`

### DIFF
--- a/src/zenml/utils/code_utils.py
+++ b/src/zenml/utils/code_utils.py
@@ -109,13 +109,21 @@ class CodeArchive(Archivable):
 
         if repo := self.git_repo:
             try:
-                result = repo.git.ls_files(
+                # Split into two calls because --recurse-submodules is only
+                # compatible with --cached, not with --others or --modified.
+                tracked = repo.git.ls_files(
                     "--cached",
+                    "--recurse-submodules",
+                    "--exclude-standard",
+                    self._root,
+                )
+                untracked_and_modified = repo.git.ls_files(
                     "--others",
                     "--modified",
                     "--exclude-standard",
                     self._root,
                 )
+                result = tracked + "\n" + untracked_and_modified
             except Exception as e:
                 logger.warning(
                     "Failed to get non-ignored files from git: %s", str(e)
@@ -123,6 +131,8 @@ class CodeArchive(Archivable):
                 all_files = self._get_all_files(archive_root=self._root)
             else:
                 for file in result.split():
+                    if not file:
+                        continue
                     file_path = os.path.join(repo.working_dir, file)
                     path_in_archive = os.path.relpath(file_path, self._root)
 
--- a/tests/unit/utils/test_code_utils.py
+++ b/tests/unit/utils/test_code_utils.py
@@ -1,0 +1,115 @@
+#  Copyright (c) ZenML GmbH 2024. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at:
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


schustmi:
@apoorvdarshan Your PR currently contains 65 changed files, can you fix this please.

schustmi (approved):

schustmi:
@apoorvdarshan Can you please merge the latest develop changes into this PR to fix the failing CI checks.

apoorvdarshan:
Done! Rebased onto the latest develop.

PR #4500, opened 2026-02-18
Remove Discord notification from release prep and reduce CI runner waste

strickvl:
## Summary

**Change 1: Remove Discord release notification entirely**
- Removes the Discord webhook step from the release preparation workflow (`release_prepare.yml`)
- This step posted to Discord when a `misc/prepare-release-*` branch was created, alerting growth and product teams not to merge to develop
- The Discord channel is no longer actively monitored at ZenML, so this is dead code
- The step was standalone (no downstream jobs depended on it), so removal is safe
- Note: the `DISCORD_WEBHOOK_RELEASE` secret in GitHub can be cleaned up separately

**Change 2: Reduce runner waste from label-triggered workflows**
- Opening a PR with 4 labels was spawning ~18 workflow runs, with 8 being pure waste from label-triggered workflows
- `snack-it.yml`: Added job-level `if` filter so it only allocates a runner when the `snack-it` label is actually added (previously spun up a runner for every label just to check and skip)
- `require-release-label.yml`: Added concurrency group (by PR number) so rapid label additions cancel previous runs, plus job-level filter to only run when `release-notes` or `no-release-notes` labels are added/removed
- Added missing concurrency groups to `check-markdown-links`, `codeql`, and `pr_labeler` workflows to prevent pile-ups from rapid pushes

## Test plan

- [x] Verify the workflow YAML is valid
- [ ] Observe reduced runner usage on next PR with multiple labels


.github/workflows/release_prepare.yml
```diff
@@ -153,17 +153,21 @@ jobs:
       - name: Send message to Discord
         # Only run this part once when the release branch is created
         if: ${{ github.event.created }}
+        continue-on-error: true
```
schustmi: IMO, throw out this whole discord message thing.

bcdurak: At first, it was supposed to alert everyone not to merge anything but by the time I have implemented it, it wasnt possible to use tags over webhooks. Not sure if it has changed, but I think you can throw it out as well.

strickvl: Removed it completely.

PR #4502, opened 2026-02-19
fix: default CLI list sort order to newest-first for time-sensitive entities

### LINKED ISSUE #4455: CLI list commands return oldest items first instead of newest

cluster2600:
## What

Overrides the `sort_by` default in four time-sensitive filter models so
that CLI `list` commands return the **newest** items first instead of
the oldest.

Affected models:
- `PipelineRunFilter`
- `StepRunFilter`
- `ArtifactVersionFilter`
- `ModelVersionFilter`

## Why

Closes #4455.

`BaseFilter.sort_by` defaults to `"created"` (ascending). The
`@list_options` decorator auto-generates Click options from filter model
field defaults and **always** passes `sort_by="created"` to the client,
even when the user did not explicitly set it. This silently overrides the
`Client.list_pipeline_runs()` sensible default of `"desc:created"`.

As a result, `zenml pipeline runs list --size=1` returns the *oldest*
run instead of the *newest*, which is the opposite of what most users
expect.

The fact that `Client.list_pipeline_runs()` already hard-codes
`sort_by="desc:created"` is strong evidence that newest-first was the
intended behaviour.

## How

Override `sort_by` in each of the time-sensitive filter models with:

```python
sort_by: str = Field(
    default="desc:created",
    description="Which column to sort by. Default is newest first.",
)
```

This is the **minimal-scope fix** — it does not change the global
`BaseFilter` default, so non-time-sensitive entities (users, stacks,
components, …) are completely unaffected.

### Code flow (after fix)
```
CLI command (--size=1)
    ↓ @list_options injects sort_by="desc:created" from PipelineRunFilter
Client.list_pipeline_runs(sort_by="desc:created", size=1)
    ↓ PipelineRunFilter(sort_by="desc:created")
SQL: ORDER BY created DESC, id DESC
    ↓ With LIMIT 1
Returns: newest run ✅
```

## Testing

```bash
# Before: returns oldest run
zenml pipeline runs list --size=1

# After: returns newest run (no workaround needed)
zenml pipeline runs list --size=1

# Explicit override still works
zenml pipeline runs list --size=1 --sort_by=asc:created
```

Unit test: `BaseFilter.sort_by` validator accepts both `"desc:created"`
and bare `"created"` — no validator changes required.

## Checklist

- [x] Fixes reported bug (oldest item returned by default)
- [x] Minimal scope — does not affect non-time-sensitive list commands
- [x] Consistent with existing `Client.list_pipeline_runs()` default
- [x] No database migrations required (sort order is query-level only)
- [x] Covers `PipelineRunFilter`, `StepRunFilter`, `ArtifactVersionFilter`, `ModelVersionFilter`


schustmi:
Thanks for the PR, but this has already been addressed in #4459. Closing as a duplicate.

PR #4515, opened 2026-02-23
Split step operator submission and monitoring

schustmi:
## Describe changes

This PR reworks the way step operators work. Instead of submitting the step and waiting for it synchronously in one method, this is now split into two separate methods. This was necessary to work with the improved monitoring loop when running dynamic pipelines. Additionally, these new methods are implemented for all step operators except the Spark one.

Breaking changes:
- The spark step operator can not be used in dynamic pipelines anymore.
- Any custom step operator flavors will not work in dynamic pipelines anymore until they implement the new methods. In static pipelines, it will fallback to the legacy method.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/step-operators/README.md
+++ b/docs/book/component-guide/step-operators/README.md
@@ -19,16 +19,16 @@ A step operator should be used if one or more steps of a pipeline require resour
 
 Step operators to execute steps on one of the big cloud providers are provided by the following ZenML integrations:
 
-| Step Operator                      | Flavor       | Integration  | Notes                                                                    |
-| ---------------------------------- | ------------ | ------------ | ------------------------------------------------------------------------ |
-| [AzureML](azureml.md)              | `azureml`    | `azure`      | Uses AzureML to execute steps                                            |
-| [Kubernetes](kubernetes.md)        | `kubernetes` | `kubernetes` | Uses Kubernetes Pods to execute steps                                    |
-| [Modal](modal.md)                  | `modal`      | `modal`      | Uses Modal to execute steps                                              |
-| [SageMaker](sagemaker.md)          | `sagemaker`  | `aws`        | Uses SageMaker to execute steps                                          |
-| [Run:AI](runai.md)                 | `runai`      | `runai`      | Uses Run:AI to execute steps                                          |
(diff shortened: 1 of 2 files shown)


src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -409,21 +411,53 @@ def _maybe_stop_isolated_steps(self) -> None:
...
+        Returns:
+            The status of the step run.
+        """
+        if step_run.config.step_operator:
+            assert self._step_operator
+            return self._step_operator.get_status(step_run)
+        else:
+            return self._orchestrator.get_isolated_step_status(step_run)
+
+    def _try_to_stop_isolated_step(self, step_run: "StepRunResponse") -> bool:
```
bcdurak: This is just a small nitpick, so feel free to ignore it if you wish. But there are two new methods here, both are used in a single place, and both try to execute an operation which might eventually fail. For this, one handles the try-except within the function and the other one outside (when it eventually get called). Keeping it consistent might make it easier to go through the code.

bcdurak: Also the output is never used I believe but we can keep it.

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -714,26 +748,16 @@ def _queue_concurrent_isolated_step(self, step: "Step") -> StepFuture:
...
-                # - When running isolated steps using the orchestrator, the
-                #   monitoring loop is responsible for retrying failed steps.
-                retry=uses_step_operator,
-                # Step operators only support sync execution right now -> We
-                # wait for the step to finish synchronously.
-                wait=uses_step_operator,
+                # The monitoring loop is responsible for monitoring and retrying
+                # steps.
+                retry=False,
+                wait=True,
```
bcdurak: The `wait` is `True` under any circumstance right now, I think. Just wanted to point it out in case you would want to clean it.

schustmi:
> Would still be nice to reflect the changes in the docs at some point but otherwise LGTM.

Good point, I'll adjust the "Implement your own step operator" page.

PR #4522, opened 2026-02-24
fix(deployer): merge pod_settings.resources into Kubernetes deployment template

eliottiti:
## Description

The Kubernetes deployer's `_build_template_context` method only populated the template's `resources` variable from pipeline-level `ResourceSettings` (via `convert_resource_settings_to_k8s_format`). The `pod_settings.resources` attribute — which uses native Kubernetes format with both `requests` and `limits` — was accepted by `KubernetesPodSettings` but **silently ignored** when rendering the deployment template.

This made it impossible to set resource **limits** on deployed pods via `KubernetesDeployerSettings`, since `ResourceSettings` only supports requests (except for GPU).

### Impact

On clusters with OPA Gatekeeper policies like `container-must-have-limits`, all ZenML pipeline deployments fail with:

```
admission webhook "validation.gatekeeper.sh" denied the request:
[container-must-have-limits] container <main> has no resource limits
[container-must-have-limits] container <main> has no resource requests
```

There is no workaround other than adding a namespace-level `LimitRange` or providing a custom deployment template.

### Root Cause

In `kubernetes_deployer.py`, the `_build_template_context` method builds the `resources` template variable exclusively from the `resource_requests` and `resource_limits` parameters (sourced from `ResourceSettings`). It never reads `settings.pod_settings.resources`, even though:

1. `KubernetesPodSettings.resources` is a documented attribute (line 84 of `pod_settings.py`)
2. The Kubernetes orchestrator and step operator **do** use `pod_settings.resources` (via `apply_default_resource_requests` and `build_pod_manifest`)
3. The deployment template uses `pod_settings.*` for every other field (labels, annotations, env, tolerations, volumes, etc.) — but not for resources

### Fix

After building the `resources` dict from `ResourceSettings`, merge in any `pod_settings.resources` values. `pod_settings.resources` takes precedence when both are specified, since it's more specific (deployer-level Kubernetes-native format vs pipeline-level abstract format).

```python
if settings.pod_settings and settings.pod_settings.resources:
    ps_resources = settings.pod_settings.resources
    for key in ("requests", "limits"):
        if key in ps_resources:
            if key in resources:
                resources[key].update(ps_resources[key])
            else:
                resources[key] = dict(ps_resources[key])
```

### Example

This configuration now works as expected:

```yaml
settings:
  deployer:
    pod_settings:
      resources:
        requests:
          cpu: "100m"
          memory: "400Mi"
        limits:
          cpu: "200m"
          memory: "800Mi"
```

## Tests

Added 5 unit tests covering:
- `pod_settings.resources` applied when `ResourceSettings` are empty
- `pod_settings.resources` overrides `ResourceSettings` values
- No resources when neither source provides values (backward compatible)
- `ResourceSettings` alone still works (backward compatible)
- `pod_settings` with only limits (no requests)

All existing Kubernetes integration tests continue to pass.

## Checklist

- [x] Tests added/updated
- [x] Backward compatible — no changes to existing behavior when `pod_settings.resources` is empty (the default)

### DIFF
--- a/src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
+++ b/src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
@@ -428,6 +428,20 @@ class KubernetesDeployer(ContainerizedDeployer):
         if resource_limits:
             resources["limits"] = resource_limits
 
+        # Merge pod_settings.resources into the resources dict.
+        # pod_settings.resources uses the native Kubernetes format
+        # (e.g. {"requests": {"cpu": "100m"}, "limits": {"cpu": "200m"}})
+        # and takes precedence over pipeline-level ResourceSettings, which
+        # only supports requests (not limits).
+        if settings.pod_settings and settings.pod_settings.resources:
+            ps_resources = settings.pod_settings.resources
+            for key in ("requests", "limits"):
+                if key in ps_resources:
+                    if key in resources:
+                        resources[key].update(ps_resources[key])
+                    else:
+                        resources[key] = dict(ps_resources[key])
+
         secret_env_list = []
         for key in secret_env_vars.keys():
             secret_env_list.append(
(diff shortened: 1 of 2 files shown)


eliottiti:
Closing to retarget to `develop` branch per CONTRIBUTING.md.

PR #4523, opened 2026-02-24
Merge `pod_settings.resources` into deployment template context

eliottiti:
## Describe changes

The Kubernetes deployer's `_build_template_context` ignores `pod_settings.resources` when building the template context. Only pipeline-level `ResourceSettings` (which cannot express limits) is used. This makes it impossible to set resource limits on deployed pods, causing failures on clusters with OPA Gatekeeper policies like `container-must-have-limits`.

The fix merges `pod_settings.resources` into the `resources` template variable, with `pod_settings.resources` taking precedence. This is consistent with how the Kubernetes orchestrator and step operator already handle `pod_settings.resources`.

**Why this is a bug, not a feature:**
- `KubernetesPodSettings.resources` is a documented attribute with a validator (`_convert_resources`)
- The orchestrator and step operator both use it — only the deployer ignores it
- The attribute is accepted silently, producing manifests with no resource spec
- There is currently no way to set resource limits on deployer pods (`ResourceSettings` only generates requests)

## Pre-requisites

  - [x] [ZenML Docs](https://docs.zenml.io) — No doc changes needed; the existing docs for `KubernetesPodSettings.resources` already describe this attribute correctly.
  - [x] Dashboard: No impact.
  - [x] Templates: No impact.
  - [x] Projects: No impact.

## Types of changes


tests/integration/integrations/kubernetes/deployers/test_kubernetes_deployer.py
```diff
@@ -0,0 +1,206 @@
+#  Copyright (c) ZenML GmbH 2025. All Rights Reserved.
```
safoinme, line 1:
```suggestion
#  Copyright (c) ZenML GmbH 2026. All Rights Reserved.
```

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -428,6 +428,20 @@ def _build_template_context(
         if resource_limits:
             resources["limits"] = resource_limits
 
+        # Merge pod_settings.resources into the resources dict.
+        # pod_settings.resources uses the native Kubernetes format
+        # (e.g. {"requests": {"cpu": "100m"}, "limits": {"cpu": "200m"}})
+        # and takes precedence over pipeline-level ResourceSettings, which
+        # only supports requests (not limits).
```
strickvl, line 435:
```suggestion
        # pod_settings.resources takes precedence over pipeline-level ResourceSettings,
	      # which can only express requests (not limits).
```

Trimmed the comments slightly. Keeps it focused on the 'why' rather than on the 'what'.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
strickvl, line 427:
```suggestion
            resources["requests"] = dict(resource_requests)
```

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
```diff
@@ -428,6 +428,20 @@ def _build_template_context(
         if resource_limits:
             resources["limits"] = resource_limits
```
strickvl, line 429:
```suggestion
            resources["limits"] = dict(resource_limits)
```

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
strickvl: Two small  suggestions. Somehow your version could mutate the caller's dict in-place. So if `_build_template_context` was ever called twice for the same deployment, then the second call would get a corrupted `resource_requests`.

tests/integration/integrations/kubernetes/deployers/test_kubernetes_deployer.py
strickvl, line 1: Possible test to consider adding: The test `test_pod_settings_resources_overrides_resource_settings` covers overlap in requests, but there's no test for the case where `resource_requests` has `{"cpu": "100m", "memory": "400Mi"}` and `pod_settings.resources` also has `{"requests": {"cpu": "500m"}, "limits": {"memory": "2Gi"}}` i.e. verifying that memory from `resource_requests` isn't lost while `cpu` is overridden. Might be worth adding.

src/zenml/integrations/kubernetes/deployers/kubernetes_deployer.py
strickvl, line 470:
```python
					# Sync pod_settings_for_template.resources with the final merged
          # resources so custom templates see a consistent view from either
          # context variable.
          if resources:
              if pod_settings_for_template is not None:
                  pod_settings_for_template = pod_settings_for_template.model_copy(
                      update={"resources": resources}
                  )
              else:
                  pod_settings_for_template = KubernetesPodSettings(
                      resources=resources
                  )
```

The formatting is off here, but I think you also need this inserted here (given the changes made above). But you should check this!

The idea is that after this, `context["resources"]` and `context["pod_settings"].resources` will always point to the same merged dict.

strickvl:
This LGTM but we need @safoinme to test it.

strickvl (approved):
Approved from my side. Needs testing, though.

bcdurak:
@safoinme

safoinme (approved):

bcdurak:
hey @eliottiti, thank you for your contribution.

PR #4525, opened 2026-02-24
Improve k8s client retries

avishniakov:
## Describe changes
The PR adds config options for kubernetes orchestrator/step operator to specify timeouts for the `retry_on_api_exception` calls to ensure that the Kubernetes API calls are properly retried and not hanging for unnecessary long, failing at the end.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/step_operators/kubernetes_step_operator.py
+++ b/src/zenml/integrations/kubernetes/step_operators/kubernetes_step_operator.py
@@ -294,6 +294,7 @@ class KubernetesStepOperator(BaseStepOperator):
                 batch_api=self._k8s_batch_api,
                 namespace=self.config.kubernetes_namespace,
                 label_selector=label_selector,
+                api_request_timeout=self.config.api_request_timeout,
             )
         except Exception as e:
             logger.warning(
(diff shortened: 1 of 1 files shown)


src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -648,9 +649,14 @@ def wrapper(*args: Any, **kwargs: Any) -> R:
         retries = 0
         while retries <= max_retries:
             try:
+                if "_request_timeout" not in kwargs:
+                    kwargs["_request_timeout"] = 30
```
schustmi: I know that I've had requests that took more than 30 sec to execute (e.g. deleting a pod in foreground mode, which executes some shutdown logic). So I don't think we can simply apply this here and hope. If anything, this should be configurable somewhere, and default to no timeout to not break stuff for any user currently relying on this for whatever reason (slow, overloaded cluster, ...)

avishniakov: Yeah, this makes a lot of sense, that the behavior for existing users shall remain the same.
I incorporated that into the setting of the orchestrator and step operator and ensured passing it down to the methods relying on the `retry_on_api_exception`.

schustmi:
@avishniakov Still some merge conflicts, otherwise LGTM

PR #4527, opened 2026-02-25
Better gcsfs credentials refresh

schustmi:
## Describe changes
Implements a native GCP credentials refresh handler that uses the service connector logic to refresh GCP credentials when they expire. This replaces the need for periodic expiration checks in the clients that use the temporary OAuth2 GCP credentials issued by the GCP service connectors, which is not 100% reliable.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
+++ b/src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
@@ -35,6 +35,7 @@ import google.auth
 import google.auth.exceptions
 import requests
 from google.auth import aws as gcp_aws
+from google.auth import credentials as gcp_credentials
 from google.auth import external_account as gcp_external_account
 from google.auth import (
     impersonated_credentials as gcp_impersonated_credentials,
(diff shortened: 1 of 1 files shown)


schustmi:
@stefannica One thing that we could still do as part of this is maybe fix the type annotations in the GCP service connector?
Both the `_authenticate` and `get_session` methods mark their return type as a `google.oauth2.credentials.Credentials` object. But e.g. when using service account credentials, those do not actually match that annotation. More importantly, they do not have the refresh handler in their interface, which made me very confused when I initially implemented. WDYT?

stefannica:
> @stefannica One thing that we could still do as part of this is maybe fix the type annotations in the GCP service connector? Both the `_authenticate` and `get_session` methods mark their return type as a `google.oauth2.credentials.Credentials` object. But e.g. when using service account credentials, those do not actually match that annotation. More importantly, they do not have the refresh handler in their interface, which made me very confused when I initially implemented. WDYT?

oh yeah, absolutely ! thanks for the reminder, this got lost in my bitbucket

stefannica:
> @stefannica One thing that we could still do as part of this is maybe fix the type annotations in the GCP service connector?

done

PR #4551, opened 2026-03-03
Misc fixes

schustmi:
## Describe changes
- Don't fail for external artifacts with `None` value
- Reduce chunk size when fetching many step runs, as we ran into some 414 URI too large errors. This is now also configurable via an env variable.

## Pre-requisites

## Types of changes


src/zenml/orchestrators/step_run_utils.py
```diff
@@ -471,8 +476,9 @@ def fetch_step_runs_by_names(
     chunks = []
     current_chunk = []
     current_length = 0
-    # stay under 6KB for good measure.
-    max_chunk_length = 6000
+    max_chunk_length = handle_int_env_var(
```
bcdurak, line 479: Very small nitpick, feel free to ignore. I usually do the `handle_x_env_var` call in the constants and only import the value in the corresponding file, thinking it's a slightly easier to read. But leaving it completely up to you.

schustmi: I do not, because I want to evaluate the env var at runtime and not what it was set to potentially hours ago when starting the python process :)

schustmi: I would even say it's bad practice. In some orchestrators, the env var set via the step decorator would not be set yet when this code is running, so it wouldn't work.

bcdurak: Makes sense 👍

PR #4566, opened 2026-03-05
Advanced filters: Phase 1 - new ops, multiple filters

### LINKED ISSUE #4568: Advanced Filters: Phase 1 - Repeating filters, new ops

_No response_

### Feature Description

Original spec document [here](https://www.notion.so/zenml/Advanced-Filters-backend-and-frontend-implications-313f8dff253880909d41d44f9ce58a23). [Spec Issue](https://github.com/zenml-io/zenml-dashboard/issues/936) here

This ticket includes: 
- Ability to use multiple filters on the same field.
- New operators to add to GenericFilterOps: NOT_ONEOF, IS_EMPTY, IS_NOT_EMPTY, NOT_CONTAINS
- Date "between" / range query support:
- Fixes for broken filters:
  - Fix tags filter on Pipelines list
  - Fix author filter on Pipelines list
  - Fix author filter on Pipeline Runs list
  - Audit all filter params across all list endpoints

For more info, check [this document](https://www.notion.so/zenml/Advanced-Filters-backend-and-frontend-implications-313f8dff253880909d41d44f9ce58a23) as well.

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

### LINKED ISSUE #4761: Allow filtering by null values

_No response_

### Feature Description

Currently, the API does not allow filtering for NULL in certain columns. For example, it is not possible to list all pipeline runs for which `source_snapshot_id` is NULL, because these filters only apply if a non-None value is configured.

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

bcdurak:
## Phase 1 for Advanced Filters

- Adds support for new filter ops: `notoneof`, `notcontains`, `isnull`, `isnotnull`
- Adds support for applying multiple filters on the same field

## Breaking Changes

- The option to provide a singular `tag` in the `TaggableFilter`s have been deprecated for a long time. Now, they are removed.

## Pre-requisites

## Types of changes


src/zenml/models/v2/base/filter.py
```diff
@@ -808,15 +923,23 @@ def _resolve_operator(value: Any) -> Tuple[Any, GenericFilterOps]:
...
+
             split_value = value.split(":", 1)
             if (
                 len(split_value) == 2
                 and split_value[0] in GenericFilterOps.values()
             ):
                 value = split_value[1]
                 operator = GenericFilterOps(split_value[0])
+                if operator in VALUELESS_FILTER_OPS:
+                    value = None
```
stefannica: Logging a warning or even throwing a ValueError here is also acceptable.

src/zenml/models/v2/base/filter.py
```diff
@@ -1221,11 +1349,18 @@ def _define_datetime_filter(
...
+                column=column,
+                value=None,
+            )
+
         try:
             filter_value: Union[datetime, Tuple[datetime, datetime]]
             if isinstance(value, datetime):
                 filter_value = value
-            elif "," in value:
+            elif isinstance(value, str) and "," in value:
```
stefannica, line 1389: the following `else` will also run if the type is not string, in which case the `datetime.strptime` conversion might fail. But this check here is only needed to please mypy, correct ?

src/zenml/models/v2/base/scoped.py
```diff
@@ -184,11 +189,11 @@ class UserScopedFilter(BaseFilter):
         "user",
     ]
 
-    scope_user: Optional[UUID] = Field(
+    scope_user: UUIDOrList = Field(
```
stefannica: This is an internal field and should only be set to one user. It is used to implement the private secrets feature. In fact, if you check the secret filter, you'll see that the code there also needs to be updated to match the new type (maybe mypy is also complaining).

src/zenml/models/v2/base/scoped.py
```diff
@@ -372,7 +386,7 @@ class ProjectScopedFilter(UserScopedFilter):
         *UserScopedFilter.FILTER_EXCLUDE_FIELDS,
         "project",
     ]
-    project: Optional[Union[UUID, str]] = Field(
+    project: UUIDOrList = Field(
```
stefannica: how can code like this still work in the router/RBAC logic if you changed this to potentially be a list:

Image

src/zenml/models/v2/core/triggers.py
```diff
@@ -194,11 +200,11 @@ class TriggerFilter(UnScopedTriggerFilter, ProjectScopedFilter):
         "type",
     ]
 
-    pipeline_id: str | None = Field(
+    pipeline_id: StrOrList = Field(
```
stefannica: I think this was a mistake and it was always meant to be `UUIDOrList`. Can you check please ?

src/zenml/models/v2/core/triggers.py
```diff
@@ -194,11 +200,11 @@ class TriggerFilter(UnScopedTriggerFilter, ProjectScopedFilter):
...
         default=None,
         description="Filter triggers by pipeline ID (triggers that are attached to this pipeline's snapshots)",
     )
-    snapshot_id: str | None = Field(
+    snapshot_id: StrOrList = Field(
```
stefannica: Same here: I think this was a mistake and it was always meant to be `UUIDOrList`.

src/zenml/cli/utils.py
```diff
@@ -2745,6 +2745,33 @@ def _is_list_field(field_info: Any) -> bool:
...
+    Returns:
+        A normalized default value compatible with Click.
+    """
+    if not multiple:
+        return default
+
+    if default is None:
+        return ()
+
+    if isinstance(default, Sequence) and not isinstance(default, str):
```
stefannica: something about this condition doesn't feel right. The second condition always evaluates true if the first one is also true. Can you double-check that you didn't forget something, or maybe used and instead of or ?

src/zenml/client.py
```diff
@@ -3983,17 +3989,17 @@ def list_schedule_triggers(
...
         logical_operator: LogicalOperators = LogicalOperators.AND,
-        user: str | UUID | None = None,
-        project: UUID | str | None = None,
-        id: UUID | str | None = None,
-        created: datetime | None = None,
-        updated: datetime | None = None,
-        name: str | None = None,
-        active: bool | None = None,
+        user: UUIDOrList = None,
+        project: UUIDOrList = None,
```
stefannica: As I mentioned before: listing project scoped resources must be scoped to a single project for RBAC to work properly. This should remain a scalar.

stefannica (changes requested):
Great feature update, looking forward to using all these cool new filter operators in the UI.

src/zenml/models/v2/base/filter.py
```diff
@@ -64,10 +64,23 @@
...
 )
 
+VALUELESS_FILTER_OPS = {
+    GenericFilterOps.IS_EMPTY,
+    GenericFilterOps.IS_NOT_EMPTY,
+}
+
+
+StrOrList = Optional[Union[str, List[str]]]
+IntOrList = Optional[Union[int, str, List[Union[int, str]]]]
```
schustmi: Maybe `IntFilterOption` or something like that? IntOrList is somehow incomplete, and I would never guess that i can use a string as well. At least `IntFilterOption` is unclear enough for me to click it, and then see that it's also a `str`?

src/zenml/models/v2/base/filter.py
bcdurak: Changed it to a ValueError here.

src/zenml/models/v2/core/triggers.py
bcdurak: I think this was a mistake even before. I specifically didn't want to change any of these in case there was a logic somewhere that depended on this value being a string instead of a UUID.

src/zenml/models/v2/core/triggers.py
bcdurak: Same as above.

src/zenml/client.py
bcdurak: Now that I think about it, I agree about the project. I would still keep the user as a list though if that's alright.

src/zenml/models/v2/base/filter.py
bcdurak: Good recommendation. I have renamed them.

src/zenml/models/v2/base/scoped.py
bcdurak: I will switch the project back to a scalar value.

src/zenml/models/v2/base/filter.py
bcdurak: Yes.

src/zenml/cli/utils.py
bcdurak: Not really. Even with a simple string, that is not the case:

```python
from typing import Sequence

test = "test"
print(isinstance(test, Sequence)) # -> True
print(not isinstance(test, str)) # -> False
```

src/zenml/models/v2/base/scoped.py
bcdurak: I switched this back as you suggested.

src/zenml/cli/utils.py
```diff
@@ -2963,15 +2992,21 @@ def list_stacks(columns: str, output_format: str, **kwargs: Any) -> None:
     def inner_decorator(func: F) -> F:
         options = []
         data_type_descriptors = set()
+        multi_value_fields: Set[str] = set()
```
Json-Andriopoulos: Minor: Moving forward we should follow 3.10+ type hints (`set[str]` in this case)

src/zenml/cli/utils.py
```diff
@@ -3013,9 +3048,24 @@ def inner_decorator(func: F) -> F:
         )
 
         def wrapper(function: F) -> F:
+            @wraps(function)
+            def normalized_function(*args: Any, **kwargs: Any) -> Any:
```
Json-Andriopoulos, line 3088: I know this doesn't trigger our Ci checks but non straightforward functions like these should have docstrings for better code readability

src/zenml/models/v2/base/filter.py
```diff
@@ -62,12 +63,33 @@
...
-    "provided value is a json formatted list."
+    "When you are using the 'oneof:'/'notoneof:' filtering make sure that "
+    "the provided value is a json formatted list."
+)
+NO_VALUE_ERROR = (
+    "When using 'isempty:'/'isnotempty:', use the explicit 'operator:' form "
+    "without a value (e.g. 'isempty:')."
 )
 
+VALUELESS_FILTER_OPS = {
```
Json-Andriopoulos, line 80: Maybe `EMPTINESS_FILTER_OPS` is a better name?

src/zenml/models/v2/base/filter.py
```diff
@@ -254,6 +297,12 @@ def generate_query_conditions_from_column(self, column: Any) -> Any:
         if self.operation == GenericFilterOps.NOT_EQUALS:
             return self._handle_not_equals(column, is_json_encoded)
 
+        if self.operation == GenericFilterOps.IS_EMPTY:
```
Json-Andriopoulos: Maybe for strings IS_EMPTY should also cover empty string?

src/zenml/models/v2/base/filter.py
```diff
@@ -323,6 +372,32 @@ def _handle_oneof(self, column: Any, is_json_encoded: bool) -> Any:
...
+        Returns:
+            The query condition.
+        """
+        from sqlalchemy import and_
+
+        conditions = []
+
+        assert isinstance(self.value, list)
+
+        for value in self.value:
```
Json-Andriopoulos, line 425: Maybe this instead is a cleaner SQL expression? : 

```
  values = self.value
  if is_json_encoded:
      values = [v for value in self.value for v in (value, f'"{value}"')]

  return or_(
      column.is_(None),
      column.not_in(values),
  )
```

src/zenml/models/v2/base/filter.py
```diff
@@ -1292,6 +1407,41 @@ def _define_uuid_filter(
         )
         return uuid_filter
 
+    @staticmethod
+    def _define_int_filter(
+        column: str, value: Any, operator: GenericFilterOps
+    ) -> NumericFilter:
+        """Define a int filter for a given column.
```
Json-Andriopoulos: Define an* 🐸

src/zenml/models/v2/base/filter.py
```diff
@@ -1292,6 +1407,41 @@ def _define_uuid_filter(
...
+        if operator in VALUELESS_FILTER_OPS:
+            return NumericFilter(
+                operation=GenericFilterOps(operator),
+                column=column,
+                value=None,
+            )
+
+        try:
+            value = int(value)
+        except ValueError:
```
Json-Andriopoulos: Maybe also add TypeError here just in case?

src/zenml/models/v2/base/filter.py
```diff
@@ -1337,15 +1490,24 @@ def _define_bool_filter(
...
+        if operator in VALUELESS_FILTER_OPS:
+            return BoolFilter(
+                operation=GenericFilterOps(operator),
+                column=column,
+                value=None,
             )
+
+        try:
+            value = bool(value)
+        except ValueError:
```
Json-Andriopoulos: With standard python types this will never raise. I am a bit concerned though what this operator does, because for instance `bool("False")` is True. I think this method here should be a stricter validator (Expect value True/False or some other formats that are clearly refering to boolean values? E.g. "false" in any casing, 0 is False, "true" in any casing and 1 is True 🤔

src/zenml/models/v2/base/scoped.py
```diff
@@ -491,7 +481,16 @@ def apply_filter(
 
         query = super().apply_filter(query=query, table=table)
 
-        if self.tags:
+        tag_values = [self.tags] if isinstance(self.tags, str) else self.tags
+
+        if tag_values and any(
+            self._resolve_operator(tag)[1]
+            not in {
```
Json-Andriopoulos: We have the VALUELESS set for this group right? (or EMPTINESS if change is applied)

src/zenml/models/v2/base/filter.py
```diff
@@ -240,11 +277,17 @@ def generate_query_conditions_from_column(self, column: Any) -> Any:
...
+        if self.operation == GenericFilterOps.NOT_ONEOF:
+            assert isinstance(self.value, list)
+            return self._handle_not_oneof(column, is_json_encoded)
 
         # Handle pattern matching operations
         if self.operation == GenericFilterOps.CONTAINS:
             return column.like(f"%{self.value}%")
 
+        if self.operation == GenericFilterOps.NOT_CONTAINS:
+            return ~column.like(f"%{self.value}%")
```
Json-Andriopoulos: Personal opinion: The use of `~` is easy to miss and a bit cryptic. I would use `not_like` as more explicit and instantly readable operator.

src/zenml/cli/utils.py
bcdurak: Changed it now.

src/zenml/cli/utils.py
bcdurak: Added a docstring to both functions here.

src/zenml/models/v2/base/filter.py
bcdurak: I would rather keep the name here if that's ok. You are right, it only includes the emptiness checks right now but this set is used in a more general sense. We might want to add additional filters to this set in the future.

src/zenml/models/v2/base/filter.py
bcdurak: Perhaps, my intention here is misrepresented by the naming here. What I want to implement here is a `IS_NULL` and `IS_NOT_NULL` check. While this can be quite useful, grouping `None` values and empty strings can be a bit confusing IMO. I will rename the operation to reflect it better.

bcdurak: Renaming done.

src/zenml/models/v2/base/filter.py
bcdurak: Good catch, fixed it now.

src/zenml/models/v2/base/filter.py
bcdurak: Done.

src/zenml/models/v2/base/filter.py
bcdurak: I have extended the `BoolFilter` to validate the `value` and do a proper mapping. Feel free to check it out.

src/zenml/models/v2/base/scoped.py
bcdurak: You are right. Changed the check here.

src/zenml/models/v2/base/filter.py
bcdurak: Changed this as well.

Json-Andriopoulos (changes requested):
Grouping some feedback:

1) As a user if I provide a wrong filter option (e.g. oneoff instead of oneof) I get an empty result set. A warning would be of value (e.g. I send a ":" separated expression, get empty results then I should get a warning like (Are you trying to use an operator? oneoff is not a valid operator see here the full list)
2) We should really document the existing and new options. SDK users should be able to benefit from those rich options.
3) Not in the scope of the PR, but would be good to address: Operators like (gt, gte, lt, lte) behave in a weird way when used with string filters. As a user, I use name="gt:hello" It would expect some alphabetical filtering. What current happens in the value is cast as a float and I get the following error: `Failed to compare the column ... to the value 'hello' (must be numeric)`. I recommend we either disallow those operators for string filters or we properly support them. I guess it could be 2 step (First drop, the follow-up with proper support in another branch).

bcdurak:
> Grouping some feedback:
> 1. As a user if I provide a wrong filter option (e.g. oneoff instead of oneof) I get an empty result set. A warning would be of value (e.g. I send a ":" separated expression, get empty results then I should get a warning like (Are you trying to use an operator? oneoff is not a valid operator see here the full list)
> 2. We should really document the existing and new options. SDK users should be able to benefit from those rich options.
> 3. Not in the scope of the PR, but would be good to address: Operators like (gt, gte, lt, lte) behave in a weird way when used with string filters. As a user, I use name="gt:hello" It would expect some alphabetical filtering. What current happens in the value is cast as a float and I get the following error: Failed to compare the column ... to the value 'hello' (must be numeric). I recommend we either disallow those operators for string filters or we properly support them. I guess it could be 2 step (First drop, the follow-up with proper support in another branch).

1. This is due to the way that our filters are built. If you don't provide a valid operator like `parameter='cantains:value'`, the code tries to find entries with `paramater` is equal to `cantains:value`. Do you suggest we add a warning each time there is a `:` with an invalid operator?
2. Good idea. TBH, I do not know where to place this fully at the moment. Would it be alright that we handle the documentation in a separate PR?
3. Perhaps as a first step, this is something that we can also explain with the documentation update as well? I like supporting the functionality and if I recall correctly this was used exactly as you phrased, but I would have to dig a bit further to find where it was being used.

@Json-Andriopoulos

bcdurak:
@Json-Andriopoulos I had another look now:

1. I do not want to throw a warning everything people use a filter with a `:` in the value where there is no operator. These could be common cases such as `url="https://example.com", image="ghcr.io/zenml/image:latest"`, `created="2026-06-10 15:43:00"`, `name="team:production"`. Instead I throw a debug message now.
2. I added some docs to the SDK docs and CLI docs.

docs/book/sdk-docs/zenml/example-usages.md
```diff
@@ -83,6 +83,32 @@ These methods always return a [Page](https://sdkdocs.zenml.io/latest/core_code_d
 
 You can further restrict your search by passing additional arguments that will be used to filter the results. E.g., most resources have a `user_id` associated with them that can be set to only list resources created by that specific user. The available filter argument options are different for each list method; check out the method declaration in the [Client SDK documentation](https://sdkdocs.zenml.io/latest/core_code_docs/core-client.html) to find out which exact arguments are supported or have a look at the fields of the corresponding filter model class.
 
+**Advanced Filters**
+
+By default, filter arguments check for equality. For fields that support advanced filters, you can use the `operator:value` syntax:
```
Json-Andriopoulos, line 88: Which fields support advanced filters? :P 

I like the section, I feel like we can afford to make it more verbose. I'd structure it as follows:

1) A table breakdown of operators: a) Name b) Structure c) Description (what the operator does) d) Field types where it is applicable
2) A list of examples, one for each operator

Json-Andriopoulos (approved):

bcdurak:
@Json-Andriopoulos I have applied the fix. I will let the CI run for the tests 🤞

stefannica (approved):

PR #4581, opened 2026-03-06
Heartbeat logs and neptune update

bcdurak:
## Describe changes

- The logs from the step heartbeat thread is now being tracked properly.
- Neptune updates

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/getting-started/zenml-pro/triggers.md
+++ b/docs/book/getting-started/zenml-pro/triggers.md
@@ -5,6 +5,10 @@ icon: server
 
 # Triggers
 
+{% hint style="info" %}
+Triggers are part of ZenML's paid features. For details on availability and supported plans, visit the [pricing page](https://www.zenml.io/pricing)
+{% endhint %}
+
 In the [snapshots](./snapshots.md) section, you learned how to prepare snapshots and execute them on demand via 
 the dashboard, CLI, or SDK. In many cases, however, pipelines need to run automatically - either on a schedule or in 
 response to an event.


src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -92,6 +93,72 @@
...
+        ExecutionStatus.CACHED,
+        ExecutionStatus.SKIPPED,
+        ExecutionStatus.RETRYING,
+        ExecutionStatus.RETRIED,
+    }:
+        raise IllegalOperationError(
+            f"Execution status `{requested_status}` is not valid for pipeline runs."
+        )
+
+    if requested_status in {ExecutionStatus.INITIALIZING}:
```
schustmi: I'm assuming we don't call this method when creating a run? because in that case the status will be initializing.

But if we don't call this method for validation when creating a run, a user could create a run with an invalid state (cached, skipped).

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -92,6 +93,72 @@
...
+        if requested_status in {
+            ExecutionStatus.STOPPED,
+            ExecutionStatus.FAILED,
+        }:
+            return ExecutionStatus.STOPPED
+        elif requested_status == ExecutionStatus.STOPPING:
+            return ExecutionStatus.STOPPING
+        else:
+            raise IllegalOperationError(
+                "Stopping steps can only be transitioned to `stopped` or `failed`."
```
```suggestion
                "Stopping runs can only be transitioned to `stopped` or `failed`."
```

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
```diff
@@ -736,33 +803,19 @@ def update(self, run_update: "PipelineRunUpdate") -> "PipelineRunSchema":
...
-                self.in_progress = self._check_if_run_in_progress()
+        if run_update.is_finished:
+            self.in_progress = False
+        elif self.snapshot and self.snapshot.is_dynamic:
+            # In dynamic pipelines, we can't actually check if the run is
+            # in progress by inspecting the DAG. Only once the orchestration
+            # container finishes we know for sure.
+            pass
+        else:
+            self.in_progress = self._check_if_run_in_progress()
```
schustmi: You've removed updating the status reason from what I can see, this needs to be added back in.

Also, I think the `in_progress` checks should not happen on every update, only on ones that update the status. Maybe we need an additional `RunSchema.update_status(status, status_reason, finished)` method that gets called from the sql zen store and takes care of all of this?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -10969,17 +10954,25 @@ def _update_pipeline_run_status(
 
         # Snapshots always exists for pipeline runs of newer versions
         assert pipeline_run.snapshot
-        num_steps = pipeline_run.snapshot.step_count
-        is_dynamic_pipeline = pipeline_run.snapshot.is_dynamic
-        new_status = get_pipeline_run_status(
+
+        # Determine the status based on the run and step statuses.
+        current_status = get_pipeline_run_status(
```
schustmi: - The current status should be the one in the DB (which is the current status)
- The requested status should be the result of this method call

Or am I misunderstanding something?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -10991,19 +10984,17 @@ def _update_pipeline_run_status(
             session.commit()
             return
 
-        run_update = PipelineRunUpdate(status=new_status)
+        pipeline_run.status = new_status.value
         if new_status.is_finished:
-            run_update.end_time = utc_now()
+            pipeline_run.end_time = utc_now()
```
schustmi: Here we should call `RunSchema.update_status(...)` IMO

src/zenml/zen_stores/schemas/step_run_schemas.py
```diff
@@ -73,6 +74,59 @@
...
+        IllegalOperationError: If the transition is invalid.
+    """
+    if requested_status in {
+        ExecutionStatus.RETRYING,
+        ExecutionStatus.RETRIED,
+    }:
+        raise IllegalOperationError(
+            f"Execution status `{requested_status}` cannot be set manually."
+        )
+
```
schustmi: Initializing/Provision not legal atm for steps.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: Your assumption is correct. We only use this when we call `_update_pipeline_run_status` and this only happens under three different circumstances: updating a run, creating a step, updating a step.

I would propose validating the initial state with a pydantic validator.

src/zenml/zen_stores/schemas/pipeline_run_schemas.py
bcdurak: Completely right about the status reason. This was the next todo in my head to handle this properly.

bcdurak:
@schustmi I have removed all the status-update related code and resolved the comments. This is just a small PR for the heartbeat logs and neptune updates.

PR #4588, opened 2026-03-09
Pause and resume pipeline runs

schustmi:
### Features
-`zenml.wait(...)` to wait for external inputs in a pipeline
  - Only works for dynamic pipelines
  - Once a user-configured timeout is reached, pauses the pipeline and frees all resources
  - When the external input is provided, the run can be resumed
    - When using a remote orchestrator and a server that allows running snapshots, the resume happens automatically
    - CLI command to resume manually: `zenml pipeline runs resume <ID>`
    - See retry/resume limitations below
- CLI command to resolve pending wait conditions: `zenml pipeline runs wait-conditions resolve`
- `zenml pipeline runs retry <ID>` CLI command that retries a failed run. This does not create a new run, it simply spins up the orchestration environment again and executes the pipeline function. This will then skip all steps that already executed before.
  - Only works for dynamic pipelines
  - See retry/resume limitations below

### Retry/Resume limitations:
- When using local orchestrators, the code needs to be available locally. It will run with the local code, and not do any checks if the code changed from when the initial run started.
- No logic to ensure the resume/retry tries to follow the same branches as the original run. For example:
  ```python
  @pipeline(dynamic=True)
  def my_pipeline():
    if get_current_hour() < 15:
      step_a()
    else:
      step_b()
    # This is where the run crashed unexpectedly or got paused.
    # In the initial execution of the run (at 14:00), it executed step_a.
    # When resuming/retrying the run at 16:00, it will execute step_b.
  ```

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/execution/pipeline/dynamic/runner.py
+++ b/src/zenml/execution/pipeline/dynamic/runner.py
@@ -1240,7 +1240,7 @@ class DynamicPipelineRunner:
 
     def await_all_step_futures(self) -> None:
         """Await all step futures."""
-        for future in self._futures.values():
+        for future in list(self._futures.values()):
             future.wait()
         self._futures = {}
 
@@ -1250,7 +1250,7 @@ class DynamicPipelineRunner:
         Returns:
             True if there are any in-progress steps, False otherwise.
         """
-        return any(future.running() for future in self._futures.values())
+        return any(future.running() for future in list(self._futures.values()))
 
 
 def compile_dynamic_step_invocation(


docs/book/how-to/steps-pipelines/wait_resume.md
```diff
@@ -0,0 +1,103 @@
+---
+description: Pause a dynamic pipeline for external input and resume it after the input is resolved.
+icon: hourglass-half
+---
+
+# Wait for external input and resume a run
```
bcdurak, line 6: Just thought of a scenario. What happens if I retry a run that requires an external input and I provided it on the first try. Does it `wait(...)` again or does it use the cached input by the user? Should this be configurable?

src/zenml/models/v2/core/pipeline_run.py
```diff
@@ -163,6 +166,18 @@ def is_placeholder_request(self) -> bool:
             ExecutionStatus.PROVISIONING,
         }
 
+    @model_validator(mode="after")
+    def _validate_status(self) -> "PipelineRunRequest":
```
bcdurak, line 170: I like this. It's missing the docstring though. Should we add a similar validator to the step as well?

src/zenml/models/__init__.py
```diff
@@ -43,6 +43,12 @@
     ProjectScopedResponseResources,
     ProjectScopedFilter,
 )
+from zenml.enums import (
```
bcdurak: Why are we importing these here? Should they just be enums?

src/zenml/orchestrators/publish_utils.py
```diff
@@ -212,66 +208,10 @@ def publish_pipeline_run_status_update(
         run_update=PipelineRunUpdate(
             status=status,
             status_reason=status_reason,
-            end_time=end_time,
         ),
     )
 
 
-def get_pipeline_run_status(
```
bcdurak, line 220: hahaha, finally this function is out of here ❤️ 😅

src/zenml/cli/pipeline.py
```diff
@@ -840,6 +845,287 @@ def list_pipeline_runs(
...
+            cli_utils.error(
+                f"Wait condition `{wait_condition_id}` is currently "
+                f"`{condition.status.value}`, not `pending`."
+            )
+        conditions = [condition]
+    else:
+        page = 1
+        size = 100
+        while True:
+            page_result = client.list_run_wait_conditions(
```
bcdurak: Very small upgrade and I find this to be an unlikely case but if someone ends up with a run with a large number of wait conditions, it might be better to get the first page, process it and then fetch the next page. This time they would not have to wait at the start and dont have to do any redundant calls if they end up cancelling the interactive process.

src/zenml/cli/pipeline.py
```diff
@@ -893,6 +1179,137 @@ def stop_pipeline_run(
...
+            run_update=PipelineRunUpdate(
+                status=ExecutionStatus.RESUMING,
+                status_reason="Manual resume requested by user.",
+            ),
+        )
+        try:
+            with cli_utils.temporary_active_stack(
+                stack_name_or_id=snapshot.stack.id
+            ) as stack:
+                stack.orchestrator.resume_run(
```
bcdurak, line 1271: This means if I run a pipeline on a remote stack, I will have to have the dependencies of the stack locally on my machine, before I can use the CLI. Can we somehow avoid that?

src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -466,6 +466,52 @@ def run(
...
+            submission_result = self.submit_dynamic_pipeline(
+                snapshot=snapshot,
+                stack=stack,
+                environment=base_environment,
+                placeholder_run=run,
+            )
+
+            if submission_result:
+                if submission_result.metadata:
+                    # TODO: Do we somehow prefix the metadata here
```
bcdurak, line 507: TODO Reminder comment

src/zenml/models/v2/core/run_wait_condition.py
```diff
@@ -0,0 +1,362 @@
...
+    def run(self) -> "PipelineRunResponse":
+        """The `run` property.
+
+        Returns:
+            the value of the property.
+        """
+        return self.get_resources().run
+
+
+class RunWaitConditionFilter(ProjectScopedFilter):
```
bcdurak: Since the run wait conditions have `run_metadata`, you can also add the `RunMetadataFilterMixin` to the mix here.

docs/book/how-to/steps-pipelines/wait_resume.md
schustmi: We decided this would mirror step behaviour, and if the same wait condition is happening again, we will reuse the previous answer.

src/zenml/models/v2/core/pipeline_run.py
schustmi: I think this PR is big enough as is, adding unrelated step status updates to it makes no sense. But in general I agree.

src/zenml/cli/pipeline.py
schustmi: Correct. No :)

src/zenml/models/v2/core/run_wait_condition.py
schustmi: Nice catch, added it!

src/zenml/orchestrators/base_orchestrator.py
schustmi: Yeah no clue how to handle it, i'll leave it for future us unless you have a good idea :D

src/zenml/models/__init__.py
schustmi: Removed them here. In general, I think it would make much sense to have one place to import everything related to wait conditions, even the enums should be in the models files. But for now for consistency I think you're right!

src/zenml/cli/pipeline.py
schustmi: Implemented this

docs/book/how-to/steps-pipelines/wait_resume.md
bcdurak: Thanks, make sense.

src/zenml/models/v2/core/run_wait_condition.py
```diff
@@ -0,0 +1,367 @@
...
+    FILTER_EXCLUDE_FIELDS = [
+        *ProjectScopedFilter.FILTER_EXCLUDE_FIELDS,
+        *RunMetadataFilterMixin.FILTER_EXCLUDE_FIELDS,
+        "resolved_by",
+        "pipeline_run",
+    ]
+    CLI_EXCLUDE_FIELDS = [
+        *ProjectScopedFilter.CLI_EXCLUDE_FIELDS,
+        *RunMetadataFilterMixin.CLI_EXCLUDE_FIELDS,
+    ]
```
bcdurak, line 295: Since the `RunMetadataFilterMixin` has `API_MULTI_INPUT_PARAMS` I would also add for future safety:

```python
    API_MULTI_INPUT_PARAMS: ClassVar[List[str]] = [
        *ProjectScopedFilter.API_MULTI_INPUT_PARAMS,
        *RunMetadataFilterMixin.API_MULTI_INPUT_PARAMS,
    ]
```

schustmi: Added it, thanks!

PR #4593, opened 2026-03-10
Fix S3 double-close AssertionError on cleanup

### LINKED ISSUE #4586: S3ArtifactStore: double-close causes `AssertionError('Session was never entered')` on cleanup
## Description

After a successful pipeline run using an S3 artifact store, the following error appears during Python shutdown/garbage collection:

```
Task exception was never retrieved
future: <Task finished name='Task-76' coro=<ClientCreatorContext.__aexit__() done, defined at aiobotocore/session.py:35> exception=AssertionError('Session was never entered')>
  File "aiobotocore/session.py", line 36, in __aexit__
    await self._client.__aexit__(exc_type, exc_val, exc_tb)
  File "aiobotocore/client.py", line 644, in __aexit__
    await self._endpoint.http_session.__aexit__(exc_type, exc_val, exc_tb)
  File "aiobotocore/httpsession.py", line 111, in __aexit__
    assert self._sessions is not None, 'Session was never entered'
AssertionError: Session was never entered
```

The pipeline completes successfully — this is a noisy post-run error, not a functional failure.

## Root Cause

This is a **double-close bug** in `ZenMLS3Filesystem`. The aiobotocore session's `__aexit__` is called twice on the same client context:

### Lifecycle:

1. **Session creation** — `s3fs.S3FileSystem.set_session()` creates the client, calls `__aenter__()`, and registers a `weakref.finalize()` callback capturing the `_s3creator` object **by value**.

2. **First close (explicit)** — ZenML's `S3ArtifactStore.cleanup()` → `ZenMLS3Filesystem._close()`:
   ```python
   async def _close(self) -> None:
       if self._s3creator is not None:
           await self._s3creator.__aexit__(None, None, None)  # First exit — OK
           self._s3creator = None  # Clears the instance attribute...
           self._s3 = None         # ...but does NOT disarm the weakref finalizer
   ```

3. **Second close (finalizer)** — When the `ZenMLS3Filesystem` is garbage collected, the weakref finalizer fires with the **original** `_s3creator` object (captured by value at registration time). It calls `close_session(loop, original_s3creator)`, which does:
   ```python
   loop.create_task(s3.__aexit__(None, None, None))  # Second exit — CRASH
   ```
   At this point `http_session._sessions` is already `None` (set by the first exit), so aiobotocore's assertion fails.

### Why the error message is misleading

The assertion `self._sessions is not None, 'Session was never entered'` fires both when the session was never entered AND when it was already exited (since `__aexit__` sets `_sessions = None`). In this case, the session WAS entered and properly exited — it's just being exited a second time.

## Background

The `_close()` / `close()` / `close_session()` methods were introduced in commit `1de7adfb67` ("Fix S3 artifact store memory leak", June 2024). The original s3fs design has only ONE close path — the weakref finalizer. ZenML added a second explicit close path but didn't coordinate it with the existing finalizer, creating two competing cleanup owners for the same resource.

In newer versions of aiobotocore (3.x), the `__aexit__` assertion became strict enough to surface this as a visible error.

## Reproduction

Run any pipeline with an S3 artifact store. After successful completion, the error appears during garbage collection / process shutdown.

Environment:
- s3fs 2026.2.0
- aiobotocore 3.2.1
- Python 3.13

## Suggested Fix

**Two-pronged approach:**

### 1. Make `close_session()` idempotent (minimal fix)

Guard against already-closed session state before attempting `__aexit__`:

```python
@staticmethod
def close_session(loop: Any, s3: Any) -> None:
    # Skip if session was already closed by explicit cleanup
    try:
        http_session = s3._client._endpoint.http_session
        if http_session._sessions is None:
            return
    except AttributeError:
        pass
    # ... rest of existing close_session logic ...
```

### 2. Disarm the finalizer after explicit close (clean fix)

Store the `weakref.finalize()` handle and call `finalizer.detach()` in `_close()`. This requires overriding `set_session()` to capture the finalizer handle, or re-registering the finalizer with ZenML's own tracking.

Either fix alone resolves the issue; both together provide defense in depth.

strickvl:
## Summary
- Fixes the double-close bug where both `ZenMLS3Filesystem._close()` (explicit cleanup) and the weakref finalizer `close_session()` call `__aexit__` on the same aiobotocore session, causing `AssertionError("Session was never entered")` during garbage collection
- Adds a shared `_safe_aexit_s3_creator()` wrapper that catches the specific aiobotocore double-close assertion and treats it as a harmless no-op, while re-raising any other errors
- Extracts `_normalize_s3_path()` helper used by `listdir()`

## Details
After a successful pipeline run using an S3 artifact store, the following error would appear during Python shutdown:

```
AssertionError: Session was never entered
```

**Root cause:** ZenML's `S3ArtifactStore.cleanup()` calls `_close()` which does `await _s3creator.__aexit__(...)` and nulls the instance attribute. But the s3fs weakref finalizer still holds the *original* creator object (captured by value at registration time). When GC later fires `close_session()`, it calls `__aexit__` a second time on the already-closed session, and aiobotocore 3.x raises an assertion.

**Fix:** Both `_close()` and `close_session()` now route through `_safe_aexit_s3_creator()`, which wraps `__aexit__` in a try/except that suppresses only the known double-close assertion. This avoids deeper coupling to s3fs internals (like disarming the weakref finalizer).

Closes #4586

## Test plan
- [x] Regression test: explicit `_close()` then finalizer `close_session()` does not raise
- [x] Regression test: `close_session()` called twice does not raise
- [x] Regression test: unrelated `AssertionError` still propagates
- [x] All existing S3 artifact store tests pass
- [x] `mypy` passes on changed file
- [x] `bandit` passes (no medium+ issues)

### DIFF
--- a/src/zenml/integrations/s3/artifact_stores/s3_artifact_store.py
+++ b/src/zenml/integrations/s3/artifact_stores/s3_artifact_store.py
@@ -98,6 +98,9 @@ class ZenMLS3Filesystem(s3fs.S3FileSystem):  # type: ignore[misc]
 
         Args:
             s3_creator: The aiobotocore async context manager to exit.
+
+        Raises:
+            AssertionError: If the assertion is unrelated to double-close.
         """
         try:
             await s3_creator.__aexit__(None, None, None)


schustmi:
Seems like a lot of code to suppress one error :sweat_smile:
Also relies on some private s3fs methods as far as I can see, which might change in any release.

PR #4594, opened 2026-03-10
Bugfix/scheduling improvements

Json-Andriopoulos:
## Describe changes

- [x] Extend RBAC checks on snapshot attachment with ModelVersion.create
- [x] Introduce `ENV_ZENML_STORE_NO_RBAC` as a flag for store resolution (applied in combination with https://github.com/zenml-io/zenml-cloud-infra/pull/133)
- [x] Fix snapshot.latest_run to include triggered runs

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
+++ b/src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
@@ -21,7 +21,7 @@ from sqlalchemy import TEXT, CheckConstraint, Column, String, UniqueConstraint
 from sqlalchemy.dialects.mysql import MEDIUMTEXT
 from sqlalchemy.orm import joinedload, object_session, selectinload
 from sqlalchemy.sql.base import ExecutableOption
-from sqlmodel import Field, Relationship, asc, col, desc, select
+from sqlmodel import Field, Relationship, and_, asc, col, desc, or_, select
 
 from zenml.config.pipeline_configurations import PipelineConfiguration
 from zenml.config.pipeline_spec import PipelineSpec
@@ -268,15 +268,22 @@ class PipelineSnapshotSchema(BaseSchema, table=True):
                         == col(PipelineRunSchema.snapshot_id),
                     )
                     .where(
-                        # The snapshot for this run used this snapshot as a
-                        # source (e.g. run triggered from the server,
-                        # invocation of a deployment). We currently do not
-                        # include runs created directly from a snapshot (e.g.
-                        # run directly, scheduled runs), as these happen before
-                        # the user officially creates (= assigns a name to) the
-                        # snapshot.
-                        col(PipelineSnapshotSchema.source_snapshot_id)
-                        == self.id,
+                        # We currently do not include runs created directly from an un-named
+                        # snapshot (e.g. run directly, scheduled runs), as these happen before
+                        # the user officially creates (= assigns a name to) the snapshot.
+                        or_(
+                            # The snapshot for this run used this snapshot as a
+                            # source (e.g. run triggered from the server,
+                            # invocation of a deployment).
+                            col(PipelineSnapshotSchema.source_snapshot_id)
+                            == self.id,
+                            # Or the snapshot itself for runs created directly
+                            # from a named snapshot (e.g. triggers).
+                            and_(
+                                col(PipelineSnapshotSchema.name).is_not(None),
+                                col(PipelineSnapshotSchema.id) == self.id,
+                            ),
+                        )
                     )
                     .order_by(desc(PipelineRunSchema.created))
                     .limit(1)


src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -229,6 +229,13 @@ def attach_trigger_to_snapshot(
         project_id=snapshot.project_id,
     )
 
+    if snapshot.pipeline_configuration.model:
```
schustmi, line 232: There's a little nuance that isn't covered here: Users might also specify a model version that already exists, in which case we'd need to check READ permissions for that model version. But: what if someone actually creates that model version after attaching? Then our check here would be wrong.

Also, creating a model version *might* also implicitly create a model, which has its own RBAC permissions. All this logic is implemented in the RBACSqlZenStore, but I'm not sure how to best handle it in this endpoint tbh.

src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
```diff
@@ -268,15 +268,22 @@ def latest_run(self) -> Optional["PipelineRunSchema"]:
...
+                        or_(
+                            # The snapshot for this run used this snapshot as a
+                            # source (e.g. run triggered from the server,
+                            # invocation of a deployment).
+                            col(PipelineSnapshotSchema.source_snapshot_id)
+                            == self.id,
+                            # Or the snapshot itself for runs created directly
+                            # from a named snapshot (e.g. triggers).
+                            and_(
+                                col(PipelineSnapshotSchema.name).is_not(None),
```
schustmi, line 283: We might not even need this condition here I think. If a user explicitly fetches an unnamed snapshot (not even possible in the UI, only via an explicit arg in the CLI/SDK), we might as well give them the latest run?

src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
```diff
@@ -268,15 +268,22 @@ def latest_run(self) -> Optional["PipelineRunSchema"]:
...
-                        # The snapshot for this run used this snapshot as a
-                        # source (e.g. run triggered from the server,
-                        # invocation of a deployment). We currently do not
-                        # include runs created directly from a snapshot (e.g.
-                        # run directly, scheduled runs), as these happen before
-                        # the user officially creates (= assigns a name to) the
-                        # snapshot.
-                        col(PipelineSnapshotSchema.source_snapshot_id)
-                        == self.id,
+                        # We currently do not include runs created directly from an un-named
```
schustmi, line 271: This code solves the latest run of a snapshot, but there are many more inconsistencies in the UI:
- When displaying a list of runs, the UI includes the `source_snapshot` of a run in the `snapshot` column
- When on a snapshot detail page and viewing the runs of that snapshot, the frontend lists runs with `source_snapshot_id=...`

If we adjust the behaviour for `snapshot.latest_run`, I think we also need to address these.

For example:
- populate `run.source_snapshot` with the same value as `run.snapshot` in case the run was triggered, and the snapshot is named (Alternatively, we can handle this case in the frontend I guess)
- adjust the filter query if `source_snapshot_id` is set

Json-Andriopoulos: Here is an alternative approach, let me know what you think @schustmi :

1) Consider that any server-side pipeline run always gets generated from a generated snapshot of the source snapshot.
2) Instead of handling it for triggered runs everywhere, I refactor triggers to follow this pattern.
3) On attach snapshot, users provide a run configuration. We create on the spot a snapshot from original_snapshot + run_configuration and attach this instead.
4) GIven that runs come from snapshots with the proper source reference any source-related filtering and rendering would work.
5) Bonus: Users can configure run behaviour separately but they can do of-course once per attachment.
6) We need to change the way trigger.snapshots is resolved but should be fairly simpler.

schustmi: @Json-Andriopoulos So every attachment would create a snapshot, but not every run, correct?
The only weird thing IMO would be that they attach a snapshot, but a different one is actually attached, right?

So they'd attach a named snapshot, but we would actually create a new, unnamed snapshot here and attach that instead. Or would you also update the trigger responses and filters and so on to always include the `source_snapshot` of the attached snapshot?

Json-Andriopoulos: Yes, it would need a bit of weird handling on the trigger objects to fetch the source snapshot instead but other than that things won't need to change in most parts of the code.

src/zenml/zen_server/routers/trigger_endpoints.py
Json-Andriopoulos: Yeah, RBAC also does usage_reporting and entitlement checks that the SqlZenStore doesn't 🤔 . I don't think this will work, we will need the RBAC store available during the execution unfortunately.

src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
schustmi: Yeah true, maybe that's the best solution 👍

src/zenml/zen_stores/base_zen_store.py
```diff
@@ -155,7 +157,11 @@ def get_store_class(store_type: StoreType) -> Type["BaseZenStore"]:
             TypeError: If the store type is unsupported.
         """
         if store_type == StoreType.SQL:
-            if os.environ.get(ENV_ZENML_SERVER):
+            # Use RBAC on the server, avoid RBAC store for workers.
+
+            if handle_bool_env_var(
+                ENV_ZENML_SERVER
+            ) and not handle_bool_env_var(ENV_ZENML_STORE_NO_RBAC, False):
```
stefannica, line 164: I think we already covered this: you don't need to disable the RBAC zen store as long as you initialize the RBAC component.

stefannica (changes requested):

src/zenml/zen_stores/base_zen_store.py
schustmi: RBAC requires an auth context to work though, but might be solved by some hacky way of setting the auth context in the executor.

Json-Andriopoulos:
Closing this PR: RBAC failures are now handled properly and latest_run resolution will be part of a dedicated bigger PR.

PR #4598, opened 2026-03-11
Diverse changes to sdkdocs generation

AlexejPenner:
Clean up and harden SDK docs build                                                                                                      
                                                                                                                                          
Summary                                                                                                                                 
                  
 - Fix migration side-effect during docs builds: Guard run_migrations() in env.py with DISABLE_DATABASE_MIGRATION env var instead of deleting source files before building. Removes fragile rm -rf of migration files from generate-docs.sh.
 - Prevent backports from stealing the latest docs alias: publish_api_docs.yml now compares the version being built against the highest git tag and only applies the latest alias if it's truly the newest release.
 - Add version validation in CI: Rejects non-semver branch names with a clear error before publishing docs.

  Additional fixes

 - Fix Plausible analytics domain (apidocs.zenml.io → sdkdocs.zenml.io)
 - Disable broken "Edit this page" link
 - Update cli.md from deprecated mkdocstrings selection/rendering syntax to modern options: block
 - Fix broken --ignored_modules argparse flag in mkdocstrings_helper.py (type=List[str] → nargs="*")
 - Remove dead loop in mkdocstrings_helper.py
 - Add !^migrations$ filter to mkdocs.yml to hide migrations from rendered SDK docs
 - Add DISABLE_DATABASE_MIGRATION=1 to serve_api_docs.sh
 - Simplify docs/README.md local build instructions to just bash scripts/serve_api_docs.sh
 - Fix deprecated echo ::set-output → $GITHUB_OUTPUT in CI workflow


.github/workflows/publish_sdk_docs.yml
```diff
@@ -20,7 +20,26 @@ jobs:
         id: get_version
```
schustmi, line 20: Maybe also rename this file to sdk_docs?

PR #4604, opened 2026-03-13
Add pipeline run stop support for Vertex AI orchestrator

martinpanacek789:
## Summary

- Implements `_stop_run()` on `VertexOrchestrator`, enabling users to stop running Vertex AI pipeline runs via CLI (`zenml pipeline runs stop`) or the dashboard
- Handles both static pipelines (`PipelineJob.cancel()`) and dynamic pipelines (`CustomJob.cancel()`)
- Follows the same pattern as the existing Kubernetes orchestrator implementation
- No new dependencies, schema changes, or migrations needed — all infrastructure (CLI, API endpoint, status mapping) already exists

## Test plan

- [x] `mypy` passes on the changed file
- [x] `bash scripts/format.sh` produces no changes
- [x] Unit tests added and passing (graceful stop, forceful static, forceful dynamic, missing run ID)
- [ ] Test stopping a static Vertex AI pipeline run (PipelineJob) on live environment
- [ ] Test stopping a dynamic Vertex AI pipeline run (CustomJob) on live environment
- [ ] Verify dashboard stop button appears and works for Vertex AI runs

### DIFF
--- a/src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
+++ b/src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
@@ -842,6 +842,56 @@ class VertexOrchestrator(ContainerizedOrchestrator, GoogleCredentialsMixin):
         job_name = step_run.run_metadata[STEP_JOB_NAME_METADATA_KEY]
         client.cancel_custom_job(name=job_name)
 
+    def _stop_run(
+        self, run: "PipelineRunResponse", graceful: bool = False
+    ) -> None:
+        """Stops a pipeline run on Vertex AI.
+
+        Args:
+            run: The pipeline run to stop.
+            graceful: If True, does nothing (lets the pipeline finish
+                naturally). If False, cancels the Vertex AI job.
+
+        Raises:
+            ValueError: If the orchestrator run ID cannot be found.
+        """
+        if graceful:
+            logger.info(
+                "Graceful stop requested - letting the Vertex AI pipeline "
+                "finish naturally."
+            )
+            return
+
+        if METADATA_ORCHESTRATOR_RUN_ID in run.run_metadata:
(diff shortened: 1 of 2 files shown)


src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -842,6 +842,56 @@ def stop_isolated_step(self, step_run: "StepRunResponse") -> None:
...
+
+        Raises:
+            ValueError: If the orchestrator run ID cannot be found.
+        """
+        if graceful:
+            logger.info(
+                "Graceful stop requested - letting the Vertex AI pipeline "
+                "finish naturally."
+            )
+            return
```
schustmi, line 863: I think this should work, but only as long as step heartbeat is enabled. So we might need to check for that here, and fail if it's not as otherwise I don't think we can gracefully shut down on vertex. This is something that @bcdurak can confirm maybe.

src/zenml/integrations/gcp/orchestrators/vertex_orchestrator.py
```diff
@@ -842,6 +842,56 @@ def stop_isolated_step(self, step_run: "StepRunResponse") -> None:
...
+
+        if METADATA_ORCHESTRATOR_RUN_ID in run.run_metadata:
+            run_id = run.run_metadata[METADATA_ORCHESTRATOR_RUN_ID]
+        elif run.orchestrator_run_id is not None:
+            run_id = run.orchestrator_run_id
```
schustmi, line 868: This should usually be fine (as in, they should be the same value), but it runs the risk that some users (on purpose or accidentally) modify the orchestrator run ID in the metadata.

For that reason, I would suggest we swap these conditions around and use the `run.orchestrator_run_id` as the primary source of truth if it exists.

PR #4605, opened 2026-03-13
Fix `api_request_timeout` typing in K8S orchestrator

avishniakov:
## Describe changes
I fixed the typing issue of `api_request_timeout` in K8S orchestrator/step operator, so the timeout is really applied.
Weird enough, but still the fact - we observed on testing that `kwargs["_request_timeout"] = api_request_timeout` is not working at all, if the float is passed, but if we use `kwargs["_request_timeout"] = int(api_request_timeout)` all works as it should. Thus raising this PR to fix it.

## Pre-requisites

## Types of changes


avishniakov:
@schustmi , it is for you, right? 🙂

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -178,7 +178,7 @@ class KubernetesOrchestratorSettings(BaseSettings):
         default=30,
         description="When stopping a pipeline run, the amount of seconds to wait for a step pod to shutdown gracefully.",
     )
-    api_request_timeout: Optional[float] = Field(
+    api_request_timeout: Optional[int] = Field(
```
schustmi: This will be breaking for anyone already having a float stored in the DB for this attribute of their component. Any idea how we can make it backwards compatible?

avishniakov: Well, `kwargs["_request_timeout"] = int(api_request_timeout)` - it will be a one-line PR then 🙂 I just wanted to make it right, shall I go for dirty type casting instead?

avishniakov: Let me do it

schustmi:
I'd probably say the cleanest way would be to make it an attribute, add a validator that converts from float -> int in case the value is a float (and maybe logs a debug message for it)

avishniakov:
Hey @schustmi , thanks for reminding me! I did an attempt to do what you told me to 🙂 Hope I got you right, if not - LMK, and I will adjust.

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -277,6 +280,37 @@ def _convert_pod_failure_policy(cls, value: Any) -> Any:
...
+        Returns:
+            The validated value in integer format.
+
+        Raises:
+            TypeError: If the value is not an integer or float or None.
+        """
+        if not (value is None or isinstance(value, (int, float))):
+            raise TypeError(
+                f"Expected `api_request_timeout` to be an int or float, got {type(value)}."
+            )
```
schustmi: I think pydantic handles this already, no need to validate manually.

src/zenml/integrations/kubernetes/flavors/kubernetes_orchestrator_flavor.py
```diff
@@ -277,6 +280,37 @@ def _convert_pod_failure_policy(cls, value: Any) -> Any:
...
+        if value is not None and value <= 0:
+            logger.warning(
+                "Non-positive value for `api_request_timeout` ignored."
+            )
+            return None
```
schustmi: I'd say get rid of this, and instead use `from pydantic import PositiveInt` and use this in the field.

avishniakov: Updated, thanks!

PR #4607, opened 2026-03-14
Make CLI more agent-friendly (quick wins)

strickvl:
## Summary

This PR makes the ZenML CLI more reliable for automation, coding agents, and CI jobs while preserving the existing human-friendly table UX.

It improves machine-readable output, makes prompt-heavy flows safer in non-interactive environments, and adds dry-run previews for a focused set of mutating commands.

### Machine-friendly CLI behavior
- adds `ZENML_CLI_MACHINE_MODE`
- adds a central `is_machine_mode()` helper
- makes commands that support `--output` default to JSON in machine mode
- blocks confirmation prompts with a structured error in machine mode; destructive operations must be confirmed explicitly via the command's flag (usually `--yes`)
- fails fast on interactive prompt flows instead of hanging, unless the prompt has a default value (in which case the default is used)
- emits structured JSON errors for ZenML CLI-managed errors in machine mode

### Structured output for single-item commands
Migrates the remaining Rich-only `describe` commands so they can produce structured output:
- `zenml artifact version describe`
- `zenml authorized-device describe`
- `zenml service-account describe`
- `zenml service-account api-key describe`
- `zenml code-repository describe`

These commands still preserve the existing key/value table presentation for normal table output.

### Hardening prompt-heavy CLI flows
Wires machine-mode behavior through remaining prompt-heavy paths, including:
- service connector register/update flows
- stack import / cloud wizard flows
- stack component connector/resource selection
- login menu / API key entry
- optional analytics email prompt
- secret interactive flows
- user password prompts

### Dry-run previews for mutating commands
Adds `--dry-run` support for:
- `zenml stack register`
- `zenml stack update`
- `zenml stack delete`
- `zenml pipeline run`
- `zenml secret create`
- `zenml secret update`

Dry-run behavior validates inputs and returns a structured preview without performing the mutation.

A few intentional safety details:
- secret values are always redacted in dry-run output
- stack environment variable previews only show set/remove intent, not raw values
- `stack delete --dry-run` skips destructive confirmation prompts because no side effects occur
- `pipeline run --dry-run` validates import/config/preparation and intentionally skips stack activation, snapshot creation, and submission

## Tests and docs
- adds focused CLI regression coverage for:
  - JSON/YAML `--columns` filtering
  - machine-mode defaults and helper behavior
  - structured JSON error output
  - prompt blocking and auto-confirm behavior
  - dry-run previews for secret, stack, and pipeline commands
- documents `ZENML_CLI_MACHINE_MODE` in the environment variable docs
- updates the global settings docs to point users at machine mode for automation use cases

## Why

ZenML already had a strong centralized output pipeline, but there were still a few places where the CLI was human-first in ways that make automation unreliable:
- some describe commands only rendered Rich tables
- prompt-heavy flows could block indefinitely
- error output was human-readable only
- machine-friendly defaults were not available as a single opt-in mode
- mutating commands had no safe preview path for validation before execution

## Verification

Ran:
- `bash scripts/format.sh --no-yamlfix ...`
- `uv run mypy src/zenml/cli/utils.py src/zenml/cli/secret.py src/zenml/cli/pipeline.py src/zenml/cli/stack.py`
- `bash scripts/docstring.sh "src/zenml/cli/utils.py src/zenml/cli/secret.py src/zenml/cli/pipeline.py src/zenml/cli/stack.py"`
- `uv run pytest tests/integration/functional/cli/test_utils.py tests/integration/functional/cli/test_secret.py tests/integration/functional/cli/test_stack.py tests/integration/functional/cli/test_pipeline.py`

Result:
- `96 passed`

## Notes for reviewers

- The table UX for the migrated describe commands is intentionally preserved; structured output is used for non-table formats only.
- The dry-run implementation is deliberately scoped to the commands above rather than trying to introduce a broader preview API across the entire client/store stack in one go.

## Verification after merging develop

After resolving the develop merge on 2026-06-05, ran:
- `uv run python -c ...` parse check over the six resolved CLI test files: passed
- `uv run ruff check --select F,I ...` over the six resolved CLI test files: passed
- focused `uv run pytest ...` over the merged CLI machine-mode/dry-run/list tests: 14 passed

## CI follow-up

Fixed the Dependabot config failure by quoting `.github/dependabot.yml` `schedule.time` as `"07:00"`. Verified locally that YAML parses the value as a string.

## Changes after review feedback

Addressed all review threads (commit `2ee2b2d6dd`):

- **Machine mode no longer auto-confirms destructive operations.** `confirmation()` now raises a structured `MachineModeConfirmationError` in machine mode; agents must pass the command's explicit flag (usually `--yes`) to proceed. This mirrors how `gh`/`terraform` behave non-interactively. To keep every command usable, added the missing `--yes` flags to `annotator dataset delete`, `zenml downgrade`, `user update` (admin demotion), and `stack deploy`, and fixed `tag delete` so `--yes` is a true bypass even when the tag has attached resources.
- **`stack deploy` refuses machine mode entirely** (rather than proceeding via `--yes`), because it opens a browser and blocks until the cloud deployment completes — an agent would hang indefinitely.
- **`stack register --dry-run` now validates multi-instance components** (alerter, step operator, experiment tracker, sandbox). Previously the dry-run exit happened before those were resolved, so a dry run could report success for a stack that would fail to register.
- **`stack update --dry-run` validates every value of multi-value component flags**, not just the first. The preview now emits a list per component type; failures surface as structured errors instead of tracebacks.
- **Interactive wait-condition resolution is blocked in machine mode** with a pointer to the non-interactive form (`zenml pipeline runs wait-conditions resolve <ID> --resolution ... [--result ...]`). The inner prompts were intentionally left as raw `click.prompt` so machine mode can never silently auto-continue a wait condition via its `c` default.
- **`prompt()` returns a provided default in machine mode** instead of erroring — a default the user could accept by pressing Enter is equally safe for automation.
- **`secret.py` interactive flows route through `cli_utils.prompt`** for consistency/defense in depth (the flows were already unreachable in machine mode behind existing guards).
- Regression tests added/updated for all of the above; docs updated to describe the new confirmation behavior.

## Follow-up robustness fixes (self-review, commit `7d1628ad6d`)

- Empty list results now emit a parseable `{"items": [], "pagination": ...}` payload for JSON/YAML instead of nothing on stdout (previously `json.loads` on an empty result crashed scripts). This also fixed pagination metadata being silently dropped for empty pages (`Page.__len__` makes an empty page falsy).
- `pipeline run --dry-run --stack ...` now prepares the pipeline inside `temporary_active_stack`, matching the real execution path, so validation runs against the requested stack instead of the currently active one.
- Dry-run component previews serialize `flavor_name` (string) instead of `ComponentResponse.flavor` (a full `FlavorResponse` object).
- Fixed the remaining stale "auto-accepts confirmations" wording in `global-settings.md`.

## Composite-review follow-ups (commit `957b884204`)

- `secret list` with zero results now emits a parseable empty JSON payload in machine mode (it previously short-circuited with a plain-text warning).
- The "no items found" message is back for every output format on empty list results (it goes to stderr, so parseable stdout stays clean); JSON/YAML additionally emit the empty payload. This fixes CSV/TSV having gone completely silent on zero results.
- Fixed the `Page`-is-falsy bug in `handle_output`'s pagination footer (same class as the `prepare_output` fix).
- `pipeline run --dry-run` warning text is now accurate: preparation is only claimed when it actually ran (dynamic pipelines skip it), and the temporary stack activation is described correctly.
- Added machine-mode blocking regression tests for `stack delete` and admin demotion via `user update`.

## Intentional behavior changes worth noting (release notes)

- **`tag delete --yes` is now an unconditional bypass.** Previously it re-prompted when the tag still had attached resources; the attached-count is now part of the confirmation message instead. Required for machine mode and consistent with every other delete command.
- **`confirmation()` now forwards `default=` to the underlying prompt.** The old implementation dropped caller kwargs, so interactive call sites passing `default=True` (e.g. connector selection) now accept on Enter instead of re-prompting.
- **`user create --password ""` no longer prompts** — it creates an account with a `NULL` password (unusable for login until activated). Scripts that relied on `--password ""` triggering an interactive prompt should pass no flag instead.
- **`describe` table rendering changed slightly** via the `serialize_pydantic_model` refactor: nested non-response models (e.g. `Source` fields) now render as plain dicts instead of Pydantic reprs, and this applies to all `print_pydantic_model` callers, not only the migrated commands.


src/zenml/cli/utils.py
```diff
@@ -234,7 +292,37 @@ def exception(exception: Exception) -> NoReturn:
     else:
         error_message = str(exception)
 
-    error(error_message)
+    error(error_message, error_type=type(exception).__name__)
+
+
+def prompt(
```
bcdurak, line 312: Not sure if this is generally a good idea, but if there is a default value, you can perhaps opt to use the provided `default` value as the default. WDYT?

src/zenml/cli/utils.py
```diff
@@ -169,7 +172,56 @@ def confirmation(text: str, *args: Any, **kwargs: Any) -> bool:
     Returns:
         Boolean based on user response.
     """
-    return Confirm.ask(text, console=console)
+    if is_machine_mode():
```
bcdurak, line 181: I find this to be a bit dangerous tbh. This means any mutation or deletion operation will be confirmed right away, right? In machine mode, `zenml stack delete` or `zenml clean` would be just confirmed right away.

src/zenml/cli/utils.py
bcdurak: `rich.prompt.Confirm` instead of `click.confirm` here

src/zenml/cli/pipeline.py
```diff
@@ -32,6 +32,7 @@
     RunWaitConditionResolution,
     RunWaitConditionStatus,
 )
+from zenml.execution.pipeline.utils import prevent_pipeline_execution
 from zenml.logger import get_logger
 from zenml.models import (
```
bcdurak, line 37: Just leaving this note in a random line in this document. We have new interactive operations that need to be blocked gracefully in this file, like the resolution of the wait condition.

src/zenml/cli/secret.py
bcdurak: To keep it consistent with other methods, I would still change this to: `cli_utils.prompt("Please enter a secret key")`

bcdurak: There are two more in the update function.

src/zenml/cli/utils.py
```diff
@@ -3042,6 +3170,23 @@ def wrapper(function: F) -> F:
     return inner_decorator
 
 
+def dry_run_option(func: F) -> F:
```
bcdurak, line 3306: I see that this is only implemented/used for a handful of CLI commands. Can you explain why do we need the dry run option? What kind of use case does this help with?

src/zenml/cli/stack.py
```diff
@@ -545,11 +799,61 @@ def register_stack(
...
+                    "name": stack_name,
+                    "provider": provider,
+                    "connector": connector,
+                    "components": component_previews,
+                    "secrets": validated_secret_names,
+                    "set_active": set_stack,
+                },
+                details=details,
+            )
+            return
```
bcdurak, line 901: Due to the delayed review by me, something is being skipped here. Alerter, step operator and experiment tracker checks happen after the `return` here now. In case of a dry-run they are being completely ignored.

src/zenml/cli/stack.py
```diff
@@ -862,6 +1166,70 @@ def update_stack(
...
+                )
+            except KeyError as err:
+                cli_utils.exception(err)
+
+            _ensure_stack_can_be_modified(target_stack)
+
+            component_previews = {}
+            for component_type, component_ids in updates.items():
+                component_response = client.get_stack_component(
+                    component_type, component_ids[0]
```
bcdurak: Similar to the problem above, this can be problematic for alerters, step operators and experiment trackers as only the first one would be "validated" here.

bcdurak (commented):
Sorry about the very delayed review 🙏 I still have some stuff to review but I left some comments and questions already.

src/zenml/cli/utils.py
strickvl: Yeah that's sort of the point. How would you rather do it? Have an explicit `--force-dangerous-operation` flag or something?

src/zenml/cli/utils.py
strickvl: You want to see what it's going to do, but you don't want it to actually make the state changes

strickvl: Just like with terraform etc

AGENTS.md
```diff
@@ -241,8 +241,10 @@ def stop(h: Union[FileHandler, RemoteHandler]) -> None:
 
 ### Prerequisites
 - Set up a Python environment with ZenML dev dependencies
-- Install ZenML in development mode: `pip install -e ".[dev]"`
+- Install ZenML in development mode with the `local`, `server`, and `dev` extras — the `dev` extra alone is NOT enough to run tests or the local server:
+  `uv pip install --python .venv/bin/python -e ".[local,server,dev]"`
 - Most scripts require these dependencies to be available
+- If `.venv/bin/pytest` or `.venv/bin/mypy` are suddenly missing mid-session, the venv was likely rebuilt — rerun the install command above rather than installing tools one by one
```
bcdurak, line 247: Should we add the explanation about the machine mode here as well?

bcdurak: Also additionally, how can this happen?

src/zenml/cli/utils.py
bcdurak: I am not 100% sure tbh, just wanted to add a warning here. Adding another flag would not solve it anyhow, as the agent would just add the new flag.

src/zenml/cli/user_management.py
```diff
@@ -165,14 +165,21 @@ def create_user(
         is_admin: Whether the user should be an admin.
     """
     client = Client()
-    if not password:
+    if password is None:
```
bcdurak, line 168: There is a behavioural change here for `zenml user create baris --password ""`.

src/zenml/cli/utils.py
bcdurak: I see your point. In the terraform case, it can be helpful as the infrastructure might change, even in a destructive way. However, for ZenML, even with the set of dry-runnable commands that we have, we are not dealing with state changes, right? Additionally, if you look at the `stack.py`, we are adding a large bulk of code for these operations and I am not sure if the simulation of these operations is worth the cost of all the extra code.

bcdurak (commented):
I like the machine mode but I am still a bit unsure about the dry run. Would be happy to discuss it anytime 👍

src/zenml/cli/utils.py
strickvl: I mean in the end, you either want your agent to be able to act on your behalf or you don't. And you either set it up in such a way that it's safe or that you use a good model that doesn't do dumb stuff somehow?

PR #4610, opened 2026-03-17
Follow snapshot.source pattern for triggered runs

Json-Andriopoulos:
## Describe changes

This story introduces an improvement on the trigger/snapshot attachment logic. Users can now provide a configuration object that helps them customize the parameters for all the triggered runs of this attachment.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/trigger.py
+++ b/src/zenml/cli/trigger.py
@@ -230,10 +230,18 @@ def delete_schedule_trigger(schedule_id: UUID, soft: bool = True) -> None:
     required=False,
     help="Path to config file (PipelineRunConfiguration) for triggered runs.",
 )
+@click.option(
+    "--allow-replace",
+    type=click.BOOL,
+    required=False,
+    help="Allow replacement if attachment already exists.",
+    is_flag=True,
+)
 def attach_schedule_trigger(
     schedule_id: UUID,
     snapshot_id: UUID,
     config_path: str | None = None,
+    allow_replace: bool = False,
 ) -> None:
     """Attach a schedule to a snapshot.
 
@@ -241,6 +249,7 @@ def attach_schedule_trigger(
         schedule_id: The ID of the schedule.
         snapshot_id: The ID of the snapshot.
         config_path: The path to the config file to use.
+        allow_replace: Allow replacement if attachment already exists.
     """
     try:
         Client().attach_trigger_to_snapshot(
@@ -249,6 +258,7 @@ def attach_schedule_trigger(
             run_configuration=PipelineRunConfiguration.from_yaml(config_path)
             if config_path
             else None,
+            allow_replace=allow_replace,
         )
     except Exception as e:
         cli_utils.exception(e)
--- a/src/zenml/zen_server/routers/trigger_endpoints.py
+++ b/src/zenml/zen_server/routers/trigger_endpoints.py
@@ -210,6 +210,7 @@ def attach_trigger_to_snapshot(
     trigger_id: UUID,
     snapshot_id: UUID,
     run_configuration: PipelineRunConfiguration | None = None,
+    allow_replace: bool = False,
     _: AuthContext = Security(authorize),
 ) -> None:
     """Attaches a trigger to a snapshot.
@@ -218,6 +219,7 @@ def attach_trigger_to_snapshot(
         trigger_id: The ID of the trigger.
         snapshot_id: The ID of the snapshot.
         run_configuration: Configuration for the follow-up trigger runs.
+        allow_replace: Allow replacement if attachment already exists.
 
     Raises:
         IllegalOperationError: If the trigger is already attached to the snapshot.
@@ -226,13 +228,18 @@ def attach_trigger_to_snapshot(
 
     snapshot = zen_store().get_snapshot(snapshot_id=snapshot_id, hydrate=True)
 
+    snapshot_replaced_id = None
+
     for s in trigger.snapshots:
         if snapshot_id in {s.source_snapshot_id, s.id}:
-            raise IllegalOperationError(
-                f"Trigger {trigger_id} is already attached to snapshot {snapshot_id}. "
-                f"To attach {snapshot_id} with a different configuration, please detach "
-                f"the current attachment first."
-            )
+            if not allow_replace:
+                raise IllegalOperationError(
+                    f"Trigger {trigger_id} is already attached to snapshot {snapshot_id}. "
+                    f"To attach {snapshot_id} with a different configuration, please detach "
+                    f"the current attachment first or use the `allow_replace` flag."
+                )
+            else:
+                snapshot_replaced_id = s.id
 
     check_entitlement(feature=SCHEDULE_FEATURE)
 
@@ -268,6 +275,11 @@ def attach_trigger_to_snapshot(
         stack=stack,
         run_configuration=run_configuration,
     )
+
+    if snapshot_replaced_id is not None:
+        zen_store().detach_trigger_from_snapshot(
+            trigger_id=trigger_id, snapshot_id=snapshot_replaced_id
+        )
 
     zen_store().attach_trigger_to_snapshot(
         trigger_id=trigger_id,
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -7725,6 +7725,7 @@ class SqlZenStore(BaseZenStore):
         trigger_id: UUID,
         snapshot_id: UUID,
         run_configuration: PipelineRunConfiguration | None = None,
+        allow_replace: bool = False,
     ) -> None:
         """Attaches (links) a trigger to a snapshot.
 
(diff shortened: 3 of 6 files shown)


src/zenml/config/pipeline_run_configuration.py
```diff
@@ -32,14 +32,11 @@
 from zenml.utils.tag_utils import Tag
 
 
-class PipelineRunConfiguration(
+class BasePipelineRunConfiguration(
```
Json-Andriopoulos: @schustmi Would like your feedback here. What else do you believe could be part of this object? The point here is to avoid passing parameters that conflict with the repeated execution (such as `run_name`).

src/zenml/cli/trigger.py
```diff
@@ -221,17 +225,22 @@ def delete_schedule_trigger(schedule_id: UUID, soft: bool = True) -> None:
 @schedule.command("attach", help="Attach schedule to snapshot")
 @click.argument("schedule_id", type=UUID)
 @click.argument("snapshot_id", type=UUID)
-def attach_schedule_trigger(schedule_id: UUID, snapshot_id: UUID) -> None:
+@model_options(BasePipelineRunConfiguration)
```
schustmi: In other commands (e.g. `zenml pipeline run ...`) we usually allow users to pass the path to a config file instead. I like having them as CLI options as well, but for more complex configs I think it doesn't really work (e.g. how to pass settings, if anything a json string would work but we would need to manually parse that I think?)

src/zenml/config/pipeline_run_configuration.py
schustmi: In general, I guess all of them work. Even run name can be templated (e.g. put {date} or {time} in the run name which gets replaced at runtime.). So not sure we even need this extra class maybe?

There are some niche cases that are not allowed (e.g. DockerSettings, but those are just one key in the settings dict which in general is allowed). The server is already validating these though, so if you call the "create snapshot from snapshot" functionality on attach, it should already fail with invalid configs.

src/zenml/models/v2/core/triggers.py
```diff
@@ -570,14 +571,23 @@ def run_once_start_time(self) -> datetime | None:
...
         return self.get_resources().snapshots
 
+    @property
+    def executable_snapshots(self) -> list["PipelineSnapshotResponse"]:
+        """Implements the 'executable_snapshots' property.
+
+        Returns:
+            A list of snapshots the triggers is attached to.
+        """
+        return self.get_resources().executable_snapshots
```
schustmi, line 589: Do we really need both here in the response?

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -133,6 +133,7 @@ class TriggerSchema(NamedSchema, table=True):
 
     # ------------------ RELATIONSHIPS --------------------------------------------
 
+    # Reference to the original snapshot objects used by users to create attachments.
```
schustmi: It's not right, these are already the new snapshots that we created from the original ones?

schustmi: I guess for snapshots attached before this feature is released, this will differ and it will actually be the snapshot that the user attached.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -7216,6 +7223,11 @@ def attach_trigger_to_snapshot(
             if not trigger:
                 raise KeyError(f"Trigger {trigger_id} doesn't exist.")
 
+            for s in trigger.snapshots:
+                if snapshot_id in {s.source_snapshot_id, s.id}:
+                    # Skip attaching already attached snapshot.
```
schustmi: Minor, but the user could attach the same snapshot twice with different config right?

src/zenml/zen_stores/schemas/trigger_schemas.py
Json-Andriopoulos: Yeah comment is misleading. Correct for previous version, for this version it will be the "executable" snapshots.

src/zenml/models/v2/core/triggers.py
Json-Andriopoulos: Yes scheduler needs the executable snapshots but I also think it would be good to keep the reference for debugging.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Initially they couldn't because (trigger_id, snapshot_id) was the primary key. Now they can but the problem is if you misconfigure the attachment you don't have a way to "detach" just the one. Needs a bit of thinking, so for I am going to leave at max 1 configurations per snapshot attachment.

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -203,37 +208,58 @@ def delete_trigger(
...
     Args:
         trigger_id: The ID of the trigger.
         snapshot_id: The ID of the snapshot.
+        run_configuration: Configuration for the follow-up trigger runs.
     """
+    trigger = zen_store().get_trigger(trigger_id=trigger_id, hydrate=True)
+
+    snapshot = zen_store().get_snapshot(snapshot_id=snapshot_id, hydrate=True)
+
+    check_entitlement(feature=SCHEDULE_FEATURE)
```
stefannica, line 244: can you move this a bit higher, so you would not potentially waste any DB calls if the feature is disabled ?

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -203,37 +208,58 @@ def delete_trigger(
...
     verify_permission_for_model(
         model=snapshot,
         action=Action.READ,
     )
 
+    verify_permission(
+        resource_type=ResourceType.PIPELINE_SNAPSHOT,
+        action=Action.CREATE,
+        project_id=snapshot.project_id,
+    )
```
stefannica, line 260: I'm not 100% sure about this check. If you think about it, it's not the user who creates the snapshot, but zenml internally. This should just be aligned with all the other similar use-cases, for example running a snapshot - if you have the same check there, then it makes sense to keep this one too.

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -203,37 +208,58 @@ def delete_trigger(
...
+    # Validates and creates a runnable snapshot from the source snapshot + run configuration
+    build, stack, model_version = validate_snapshot_for_server_execution(
+        snapshot=snapshot,
+        run_configuration=run_configuration,
+    )
+    target_snapshot = create_snapshot_from_source(
+        snapshot=snapshot,
+        stack=stack,
+        run_configuration=run_configuration,
+    )
```
stefannica, line 277: if run_configuration is omitted, does it still make sense to create a new snapshot instead of using the existing one ? of course, this makes it easier to differentiate between executable and original snapshots down the line.

src/zenml/zen_stores/sql_zen_store.py
stefannica: return-ing here instead of throwing an error will leave the user wondering what happened.

on a related note: is there a way to update the run configuration of an existing snapshot attachment ? shouldn't there be one ?

src/zenml/zen_server/routers/trigger_endpoints.py
schustmi: Debatable I'd say, entitlement check is a call to cloudapi which internally calls stigg

src/zenml/zen_server/routers/trigger_endpoints.py
Json-Andriopoulos: Yes it still makes sense because we need to run the `ensure_async_orchestrator` operation.

Json-Andriopoulos: Also because we need a single way to handle things across the code base, reusing the same snapshot will cause multiple minor issues across front & backend where source_snapshot_id is expected.

src/zenml/zen_server/routers/trigger_endpoints.py
Json-Andriopoulos: This is consistent with the checks we do on similar endpoints.

src/zenml/zen_server/routers/trigger_endpoints.py
Json-Andriopoulos: Kept `entitlement_check` lower as it is a more expensive call (2 request hops).

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: @stefannica We can follow 2 paths. 1) The defensive way: we raise an error to let the user know that attachment did not succeed. 2) The friendly way: We re-attach and apply the new configuration under the hood. In the first approach the user would have to detach then attach with new configuration. I like this explicit approach more as it makes it more difficult to accidentally override/delete a configuration.

stefannica: Just for the record, we agreed on the following compromise: the user shall communicate via a flag (e.g. `allow_replace`) if it's alright to replace an existing snapshot-schedule relationship and the code shall either fail or pass depending on that flag value.

This allows the same endpoint/zen store API call to be also used to update snapshot-schedule relationships in addition to creating new ones.

PR #4618, opened 2026-03-18
Feature/kitaru analytics support

safoinme:
## Pre-requisites

## Types of changes


src/zenml/analytics/enums.py
```diff
@@ -97,3 +97,11 @@ class AnalyticsEvent(str, Enum):
...
     CREATED_TRIGGER = "Trigger created"
     DELETED_TRIGGER = "Trigger deleted"
+
+    # Kitaru
+    KITARU_CLI_INVOKED = "Kitaru CLI invoked"
+    KITARU_MCP_SERVER_STARTED = "Kitaru MCP server started"
+    KITARU_MCP_TOOL_CALLED = "Kitaru MCP tool called"
+    KITARU_FLOW_DEPLOYED = "Kitaru flow deployed"
+    KITARU_FLOW_REPLAYED = "Kitaru flow replayed"
+    KITARU_CONNECTED = "Kitaru server connected"
```
schustmi: Do we also need these here? This means for every new even we want in kitaru, we'd need to modify zenml.

Or can we somehow keep these in the kitaru repo and only add the new sources here?

src/zenml/zen_server/middleware.py
```diff
@@ -255,9 +257,13 @@ async def infer_source_context(request: Request, call_next: Any) -> Any:
         the response to the request.
     """
     try:
+        default_source = os.environ.get(
```
schustmi, line 260: So kitaru will now set this environment variable whenever it's imported, right?

schustmi: Also it would need to set the source context var for python/cli right?

bcdurak: Exactly. Kitaru needs to set this after zenml is imported too.

src/zenml/analytics/enums.py
bcdurak: I was thinking about the same but If the plan is to use the zenml analytics client (and its utils), we might have to add them here.

src/zenml/zen_server/middleware.py
schustmi: just make it Union[AnalyticsEvent, str]

safoinme: Yes in kitaru repo this is added

PR #4641, opened 2026-03-23
Bugfix/archived schedule issues

Json-Andriopoulos:
## Describe changes

This PR addresses the 2 following behavioral issues associated with schedule deletion:

- Deletion by ID prefix was not working for archived schedules.
- Reuse of an archived schedule name was blocked.

PR also includes documentation updates regarding the behavior of schedules post-deletion (archived schedule renaming explained, hard deletion of archived objects guideline, discovery of archived objects, etc.).

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/pipeline.py
+++ b/src/zenml/cli/pipeline.py
@@ -656,18 +656,29 @@ def schedule() -> None:
         "is_archived",
     ],
 )
+@click.option(
+    "--is-archived",
+    help="Hide archived, show archived or show both archived and non-archived.",
+    type=click.Choice(["true", "false", "both"], case_sensitive=False),
+    default="false",
+)
 def list_schedules(
-    columns: str, output_format: OutputFormat, **kwargs: Any
+    columns: str, output_format: OutputFormat, is_archived: str, **kwargs: Any
 ) -> None:
     """List all pipeline schedules.
 
     Args:
         columns: Columns to display in output.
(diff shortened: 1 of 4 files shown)


src/zenml/cli/pipeline.py
```diff
@@ -674,6 +666,9 @@ def list_schedules(
         output_format: Format for output (table/json/yaml/csv/tsv).
         **kwargs: Keyword arguments to filter schedules.
     """
+    if kwargs.get("is_archived") is None:
+        kwargs["is_archived"] = False
+
```
stefannica: This makes it impossible to list schedules regardless of archived status. To truly make this work, you need to take `is_archived` off the list of CLI filters and implement your own arguments to handle all three cases: true/false/both.

See `CLI_EXCLUDE_FIELDS`.

src/zenml/models/v2/core/schedule.py
```diff
@@ -361,8 +366,7 @@ class ScheduleFilter(ProjectScopedFilter):
         description="The time at which the schedule should run once",
         union_mode="left_to_right",
     )
-
-    is_archived: bool = Field(
-        default=False,
+    is_archived: bool | None = Field(
+        default=None,
```
stefannica, line 370: Don't you think it's a good idea to keep `False` as a default, given that you now handle this through CLI argument? Every piece of code that calls the zen store API directly to list schedules should continue to exclude archived filters unless explicitly asked not to.

PR #4648, opened 2026-03-26
Introduce new Databricks step-operator and fix `tags` and `custom-tags`

safoinme:
## Summary

- Introduces a **Databricks step operator** so you can run individual pipeline steps on Databricks (for example when the rest of the stack uses another orchestrator, or you want specific steps to execute in the Databricks runtime).
- Adds **optional tag settings** on the **Databricks orchestrator** so you can label jobs and cluster resources for cost, ownership, and governance in Databricks.

### Databricks step operator

**What it is.** A new **step operator** flavor so individual pipeline steps can run on Databricks while the rest of the pipeline stays on your current orchestrator. It fits teams that use Databricks for part of the workload (Spark, training, a specific runtime) but do not want the whole graph scheduled as one Databricks job.

**When to use it vs the orchestrator.** Use the **step operator** if only *some* steps should run on Databricks. Use the **orchestrator** if Databricks should drive the *entire* pipeline and scheduling story.

**How it behaves (at a glance).** ZenML packages your project, uploads it to the workspace, and submits a **one-off Databricks job run** for the chosen step. The step runs the same way in spirit as the Databricks orchestrator’s remote execution model, but scoped to that step’s stack and settings.

**What you need.** Install the Databricks integration, have a Databricks workspace and credentials that can submit jobs, and use a **remote** artifact store on the stack (same practical bar as other remote step operators in ZenML).

**Typical flow.** Register a `databricks` step operator on the stack, then mark steps with the operator name:

```python
from zenml import step

@step(step_operator="databricks_step_operator")
def my_step():
    ...
```

**Configuration.** Cluster and execution options (Spark version, workers or autoscaling, node types, Spark config/env, cluster policy, init scripts, image-related settings, and similar) are available via **`DatabricksStepOperatorSettings`** and can be attached per step through `settings={"step_operator": ...}` so only those steps pay the cost of that cluster shape.

**Limitations to be aware of.** This path targets **ad-hoc step runs**, not a long-lived Databricks job with full job-level scheduling and retry semantics, so job-oriented controls that apply to scheduled or persistent jobs may not apply here. Use the **Databricks-specific** settings for cluster sizing; generic ZenML `resource_settings` are not mapped to Databricks cluster shape.

*Full walkthrough, CLI `zenml step-operator register` examples, and warnings (e.g. wheel cleanup under the workspace path) are in the component guide for the Databricks step operator.*

### Orchestrator: `custom_tags` and `job_tags` (side change)

Two optional tag maps on the Databricks orchestrator settings help you align ZenML runs with how you track cost and ownership in Databricks:

| Setting | Applies to | Databricks limit | Use case |
|--------|------------|------------------|----------|
| `custom_tags` | Cluster resources (e.g. AWS instances, EBS volumes) | 45 tags | Cost allocation, governance |
| `job_tags` | Databricks job (and cluster tags as documented) | 25 tags | Project tracking, ownership |

### Usage (orchestrator tags)

```python
from zenml.integrations.databricks.flavors import DatabricksOrchestratorSettings

databricks_settings = DatabricksOrchestratorSettings(
    custom_tags={"cost_center": "ml-team", "environment": "production"},
    job_tags={"project": "recommendation-engine", "owner": "data-team"},
)
```

## Pre-requisites

## Types of changes


Json-Andriopoulos:
@claude Please review this

src/zenml/integrations/databricks/orchestrators/databricks_orchestrator.py
```diff
@@ -392,21 +401,47 @@ def _upload_and_run_pipeline(
         databricks_client = self._get_databricks_client()
         spark_conf = settings.spark_conf or {}
 
-        policy_id = settings.policy_id or None
-        for policy in databricks_client.cluster_policies.list():
-            if policy.name == "Job Compute":
-                policy_id = policy.policy_id
+        policy_id = settings.policy_id
```
Json-Andriopoulos: Maybe something like this would be better:

~~~
def get_policy_id(databricks_client, name: str) -> str | None:
    return next(
        (p.policy_id for p in databricks_client.cluster_policies.list() if p.name == name),
        None,
    )

policy_id = settings.policy_id or get_policy_id(databricks_client, "Job Compute")

if policy_id is None:
    raise ValueError(
        "Could not find the `Job Compute` policy in Databricks. "
        "Either create a `Job Compute` policy or specify a "
        "`policy_id` in the orchestrator settings."
    )
~~~

src/zenml/integrations/databricks/flavors/databricks_orchestrator_flavor.py
```diff
@@ -102,6 +137,65 @@ class DatabricksOrchestratorSettings(BaseSettings):
...
+            default=None,
+            description="Access control list for the Databricks job. Grants "
+            "permissions to users, groups, or service principals. By default, "
+            "only the job creator (service principal) can access the job. "
+            "Example: [{'group_name': 'users', 'permission_level': 'CAN_VIEW'}]",
+        )
+    )
+
+    # Job-level execution control
+    timeout_seconds: Optional[int] = Field(
```
Json-Andriopoulos: This comment applies to all numerical fields - we should add some validations as part of the field definition. For example here, I assume timeout_seconds it a positive integer so (ge=0).

src/zenml/integrations/databricks/flavors/databricks_shared_settings.py
```diff
@@ -0,0 +1,198 @@
...
+        default=None,
+        description="Databricks cluster policy ID for governance and cost control.",
+    )
+    autotermination_minutes: Optional[int] = Field(
+        default=None,
+        description="Minutes of inactivity before automatic cluster termination. "
+        "Helps control costs by shutting down idle clusters.",
+    )
+    autoscale: Tuple[int, int] = Field(
+        default=(0, 1),
```
Json-Andriopoulos: min_workers=0 seems like a dodgy default. In most cases (min=1, max=1) would be more appropriate no? Ignore the comment if this is the default settings value for databricks. If it so, maybe we could have a constant declaration (e.g. DEFAULT_DATABRICKS_AUTOSCALE) so that it acts a signal that we follow the library/framework defaults.

src/zenml/orchestrators/wheel_build_utils.py
```diff
@@ -0,0 +1,151 @@
...
+from zenml.logger import get_logger
+from zenml.utils.io_utils import copy_dir
+from zenml.utils.source_utils import get_source_root
+
+logger = get_logger(__name__)
+
+DEFAULT_PACKAGE_NAME = "zenmlproject"
+
+
+def sanitize_name(name: str) -> str:
```
Json-Andriopoulos, line 37: Regex is weird. Doing the same operations with native python is much more readable:

```
import string

allowed = set(string.ascii_lowercase + string.digits + "-")

def sanitize_name(name: str) -> str:
    name = "".join(c if c in allowed else "-" for c in name.lower())
    return name.strip("-")
```

PS: verify before copying this :)

src/zenml/orchestrators/wheel_build_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    """Sanitize a name for use in generated package identifiers.
+
+    Args:
+        name: Arbitrary input name.
+
+    Returns:
+        Sanitized name.
+    """
+    name = re.sub(r"[^a-z0-9-]", "-", name.lower())
```
Json-Andriopoulos: This works only with ascii characters, is that intended?

src/zenml/integrations/databricks/step_operators/databricks_step_operator.py
```diff
@@ -0,0 +1,381 @@
...
+            ignored_settings.append("max_concurrent_runs")
+        if settings.max_retries is not None:
+            ignored_settings.append("max_retries")
+        if settings.min_retry_interval_millis is not None:
+            ignored_settings.append("min_retry_interval_millis")
+        if settings.retry_on_timeout is not None:
+            ignored_settings.append("retry_on_timeout")
+
+        if ignored_settings:
+            logger.warning(
```
Json-Andriopoulos: We should add this information on the field description within settings, for each of the ignored fields

src/zenml/integrations/databricks/flavors/databricks_shared_settings.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+    spark_env_vars: Optional[Dict[str, str]] = Field(
+        default=None,
+        description="Environment variables for the Spark driver and executors. "
+        "Example: {'SPARK_WORKER_MEMORY': '4g', 'SPARK_DRIVER_MEMORY': '2g'}",
+    )
+    schedule_timezone: Optional[str] = Field(
+        default=None,
+        description="Timezone for scheduled pipeline execution. "
+        "Uses IANA timezone format (e.g., 'America/New_York').",
```
Json-Andriopoulos: We should validate this.

src/zenml/integrations/databricks/flavors/databricks_shared_settings.py
```diff
@@ -0,0 +1,198 @@
...
+        default=None,
+        description="Docker image URL for the cluster. Must be accessible "
+        "from the Databricks workspace. "
+        "Example: 'my-registry.com/my-image:latest'",
+    )
+    docker_image_username: Optional[str] = Field(
+        default=None,
+        description="Username for authenticating to the Docker registry",
+    )
+    docker_image_password: Optional[str] = Field(
```
Json-Andriopoulos: This should  be a Secret Field

stefannica (changes requested):
## Critical Issues (must fix before merge)

### C1. `databricks_orchestrator_flavor.py` will crash at import — incomplete refactor leaves dead code that references removed imports

`databricks_orchestrator_flavor.py` (lines 31-196) retains the old class bodies (`DatabricksAvailabilityType`, `DatabricksPermissionLevel`, `DatabricksAccessControlRequest`, and the first `DatabricksOrchestratorSettings(BaseSettings)` definition), but the imports they depend on (`get_logger`, `StrEnum`, `BaseModel`, `Field`, `BaseSettings`, `Dict`, `List`, `Tuple`) were removed from the import block during the refactor.

Line 31 will raise `NameError: name 'get_logger' is not defined` the moment any code imports this module, which kills the entire Databricks integration.

Additionally, `DatabricksOrchestratorSettings` is defined twice — the second definition (line 197, inheriting `DatabricksBaseSettings`) silently shadows the first. The first 160 lines of class definitions are dead code.

```31:196:src/zenml/integrations/databricks/flavors/databricks_orchestrator_flavor.py
logger = get_logger(__name__)


class DatabricksAvailabilityType(StrEnum):
    # ... 160+ lines of dead code referencing removed imports ...

class DatabricksOrchestratorSettings(BaseSettings):
    # ... this entire block is dead ...
    retry_on_timeout: Optional[bool] = Field(...)

class DatabricksOrchestratorSettings(DatabricksBaseSettings):
    """Databricks orchestrator settings."""
```

```suggestion
# Delete everything from line 31 through line 196 (inclusive).
# Keep only:
class DatabricksOrchestratorSettings(DatabricksBaseSettings):
    """Databricks orchestrator settings."""
```

### C2. `convert_step_to_task()` builds libraries locally then discards them — uses `_get_databricks_libraries()` for the `libraries=` kwarg

In `databricks_utils.py`, `convert_step_to_task()` builds a local `db_libraries` list (lines 235-246) and then never uses it. Instead, it passes `libraries=_get_databricks_libraries(...)` to `DatabricksTask`. The local list is dead code from the refactor.

This is not functionally broken because `_get_databricks_libraries` produces the correct result, but it's confusing and wastes cycles. Just remove the dead local build.

```235:253:src/zenml/integrations/databricks/utils/databricks_utils.py
    db_libraries = []               # <-- dead code
    if libraries:
        for library in libraries:
            if library.endswith(".whl"):
                db_libraries.append(Library(whl=library))
            else:
                db_libraries.append(Library(pypi=PythonPyPiLibrary(library)))
    if zenml_project_wheel:
        db_libraries.append(Library(whl=zenml_project_wheel))
    db_libraries.append(
        Library(pypi=PythonPyPiLibrary(f"zenml=={__version__}"))
    )
    return DatabricksTask(
        task_key=task_name,
        job_cluster_key=job_cluster_key,
        libraries=_get_databricks_libraries(  # <-- actual used value
            libraries=libraries,
            zenml_project_wheel=zenml_project_wheel,
        ),
```

```suggestion
    return DatabricksTask(
        task_key=task_name,
        job_cluster_key=job_cluster_key,
        libraries=_get_databricks_libraries(
            libraries=libraries,
            zenml_project_wheel=zenml_project_wheel,
        ),
        python_wheel_task=PythonWheelTask(
            ...
        ),
        ...
    )
```

---

## Warnings (should fix soon)

### W1. Documentation: `step_operator=True` is not valid ZenML API

The new step operator docs (`docs/book/component-guide/step-operators/databricks.md`, line 68) show:

```python
@step(step_operator=True)
def trainer(...) -> ...:
```

The ZenML `@step` decorator expects the step operator name as a string (e.g., `step_operator="databricks"`) — not `True`. Other step operator docs in the repo (Run:AI, AzureML) all use string names. This will mislead users.

```suggestion
@step(step_operator="databricks")
def trainer(...) -> ...:
    """Run this step on Databricks."""
    ...
```

(Same issue on line 101 of the same file.)

### W2. Documentation: `num_workers="3"` is a string but the field expects `int`

In `docs/book/component-guide/orchestrators/databricks.md`, the new tagging example has:

```python
num_workers="3",
```

But `DatabricksBaseSettings.num_workers` is typed `Optional[int]`. Pydantic will coerce it, but the docs should model the correct type for clarity.

```suggestion
num_workers=3,
```

### W3. `docker_image_password` in `DatabricksBaseSettings` should be a `SecretField`

The `docker_image_password` field (line 195 of `databricks_shared_settings.py`) is a plain `Optional[str]`. Since this is a credential, it should use `SecretField` to prevent it from being serialized in plaintext to the store, consistent with how `client_secret` is handled in the config classes.

```suggestion
from zenml.utils.secret_utils import SecretField

docker_image_password: Optional[str] = SecretField(
    default=None,
    description="Password for authenticating to the Docker registry",
)
```

### W4. `availability_type` setting is accepted but never wired into `ClusterSpec`

`DatabricksBaseSettings.availability_type` exists in the shared settings but `build_databricks_cluster_spec()` never passes it to `ClusterSpec`. Users can configure it and get no error, but it has no effect. I acknowledge this was pre-existing, but since this PR refactored the cluster spec builder it's a good opportunity to either wire it in or remove it.

### W5. `_get_databricks_libraries()` missing docstring detail

The helper `_get_databricks_libraries()` has a one-line docstring without `Args:` / `Returns:` documentation. While private, it is called from multiple functions. A minimal docstring would aid maintainability.

---

## Suggestions / Nits

### nit: `DatabricksOrchestratorFlavor.config_class` docstring says "KubeflowOrchestratorConfig"

Line 285 of `databricks_orchestrator_flavor.py`:
```
"""Returns `KubeflowOrchestratorConfig` config class.
```

This is a pre-existing copy-paste artifact. While you're touching this file, consider fixing it:

```suggestion
"""Returns DatabricksOrchestratorConfig config class.
```

### nit: `DatabricksAccessControlRequest` model validation

The docstring says "Exactly one of group_name, user_name, or service_principal_name must be specified." but there is no Pydantic `model_validator` enforcing this. Users can pass zero or multiple identifiers with no error. Consider adding a validator, or weakening the docstring to "At least one of..." to match reality.

### nit: `DatabricksStepOperatorConfig.client_id` type annotation

```python
client_id: str = SecretField(default=None)
client_secret: str = SecretField(default=None)
```

The type annotation says `str` but the default is `None`. This should be `Optional[str]` for correctness. Same pattern exists in the orchestrator config (pre-existing), but worth aligning when touching related code.

### nit: Wheel build utils — `os.chdir()` in `create_wheel()` is not thread-safe

`create_wheel()` calls `os.chdir(temp_dir)` and restores via `finally`. This mutates global process state and is not safe if multiple threads build wheels concurrently. The step operator calls this from the main orchestrator process, so it's probably fine for now, but consider running `pip wheel` with `cwd=temp_dir` via the `subprocess.run` `cwd` parameter instead:

```suggestion
result = subprocess.run(
    ["pip", "wheel", "."],
    check=True,
    capture_output=True,
    cwd=temp_dir,
)
```

This eliminates the `os.chdir` / `os.getcwd` dance entirely.

---

## Follow-up Questions for the Author

1. Was the intent to keep the old `DatabricksOrchestratorSettings(BaseSettings)` class body (lines 31-196) in `databricks_orchestrator_flavor.py`? It looks like a merge artifact from the incomplete migration to `DatabricksBaseSettings`. Can you confirm it should be removed?

2. The step operator uses `jobs.submit()` (one-time runs) while the orchestrator uses `jobs.create()` + `jobs.run_now()`. Have you tested that both ACL models (`iam.AccessControlRequest` vs `jobs.JobAccessControlRequest`) work correctly end-to-end with their respective Databricks API calls?

3. The `_resolve_policy_id` function falls back to scanning all policies for one named "Job Compute". Is this name guaranteed to exist in all Databricks workspaces, or is it Azure-specific? The error message is clearer now than before (suggesting the user specify a `policy_id`), which is good.

4. `DatabricksBaseSettings.autoscale` defaults to `(0, 1)` — is `min_workers=0` a valid value for Databricks autoscaling? Some Databricks configurations require `min_workers >= 1`.

---

## Checklist to Unblock

1. Fix C1: Remove dead code (lines 31-196) from `databricks_orchestrator_flavor.py`.
2. Fix C2: Remove dead local `db_libraries` build in `convert_step_to_task()`.
3. Fix W1: Change `step_operator=True` to `step_operator="databricks"` in docs.
4. Fix W2: Change `num_workers="3"` to `num_workers=3` in docs.
5. Consider W3 (SecretField for `docker_image_password`) — this is security-relevant.

src/zenml/integrations/databricks/flavors/databricks_shared_settings.py
safoinme: databricks can scale down to 0 actually since this is serverless job

src/zenml/orchestrators/wheel_build_utils.py
safoinme: yes

src/zenml/integrations/databricks/flavors/databricks_orchestrator_flavor.py
```diff
@@ -13,96 +13,101 @@
...
-
-
-class DatabricksOrchestratorSettings(BaseSettings):
-    """Databricks orchestrator base settings.
-
-    Configuration for Databricks cluster and Spark execution settings.
-    Field descriptions are defined inline using Field() descriptors.
+    These fields apply only to persistent Databricks Jobs created by the
+    orchestrator. They are not available on the Databricks step operator,
+    which submits one-time runs that have no Job resource to attach them to.
```
stefannica:
```suggestion
```

slop

src/zenml/integrations/databricks/orchestrators/databricks_orchestrator.py
```diff
@@ -119,11 +104,16 @@ def _get_databricks_client(
...
+        if (
+            self.config.client_id is not None
+            and self.config.client_secret is not None
+        ):
+            return DatabricksClient(
+                host=self.config.host,
+                client_id=self.config.client_id,
+                client_secret=self.config.client_secret,
+            )
+        return DatabricksClient(host=self.config.host)
```
stefannica, line 116: why not do the same thing you do for the step operator and store the client on the instance instead of recreating it ?

stefannica (changes requested):

stefannica (approved):
I've gone through a full AI assisted code review and couldn't find any issues that are must-fix. I will enclose the full report of the research and review here.

# Code review walkthrough session

## Metadata

| Field | Value |
|-------|--------|
| Repository | `/home/silent-assassin/aspyre/src/zenml` |
| Base ref | `develop` |
| Head branch | `fix/add-tags-to-databricks-orchestrator` |
| Date | 2026-04-29 |
| Participants | (to be filled) |

**Discovery (readonly):** `git diff develop...fix/add-tags-to-databricks-orchestrator --stat` — 26 files, ~3717 insertions, ~419 deletions.

---

## Stage 1 — Introduction

### Explainer summary

**Problem and intent:** The branch extends the Databricks integration so remote jobs are easier to govern, label, and operate: job tags and custom tags (and related Databricks job options) flow from stack settings into the Jobs API, and the integration gains a **Databricks step operator** alongside a refactored **shared settings** layer used by both the orchestrator and the new operator. Packaging for remote runs is clarified by **wheel build utilities** split out of the generic wheeled orchestrator, and **logging** behavior is adjusted in support of these paths. **Documentation** adds a step-operator guide and refreshes the orchestrator guide; **unit tests** cover new flavors, utils, step operator, orchestrator, logger, and wheel helpers.

**Themes (not exhaustive):**

1. **Shared Databricks settings** — `databricks_shared_settings.py` centralizes models (e.g. tags, custom tags, access control, cluster-related options) consumed by orchestrator and step operator flavors.
2. **Orchestrator** — `databricks_orchestrator.py` and `databricks_orchestrator_flavor.py` updated to use shared settings and to pass through job configuration (including tags) to Databricks job creation.
3. **New: Databricks step operator** — New stack component, entrypoint config, and utils wiring parallel to the orchestrator path.
4. **Utils and packaging** — `databricks_utils.py` grows substantially; `wheel_build_utils.py` is introduced and `wheeled_orchestrator.py` is slimmed to delegate to it.
5. **Logger** — `logger.py` changes with dedicated unit tests.
6. **Docs and registration** — `databricks.md` (orchestrator), new `step-operators/databricks.md`, TOC, integration `__init__` / flavors.

**Size and shape:** Large, multi-surface feature branch (orchestrator + new component + shared models + cross-cutting logger/wheel refactors), not a single-hunk “tag only” fix.

### Session notes / questions

- (User) Scope confirmation: any paths or commits to deprioritize (e.g. E2E template auto-update on branch)?
- (Open) PR/ticket link if available for metadata.

### Decisions

- Report path: `~/aspyre/docs/code-reviewer/review-session-fix-add-tags-to-databricks-orchestrator-20260429.md` (default pattern).

---

## Stage 2 — Public documentation pass (UX, product, and DevX)

### Explainer summary

Stage 2 is anchored in a **user-facing** description of what this branch adds and how it changes the **public** Databricks surface: a new **step operator**, a **unified settings model** shared with the **orchestrator**, **new optional knobs** (tags, ACLs, timeouts, Docker, init scripts, job concurrency and retries, and more), and **compatibility** notes (including import and validation). The subsections below are the canonical public reference for the session; **user journeys** follow and reuse the same vocabulary.

### Breaking, compatibility, and public feature reference

#### What ships in product terms

- A new **stack component:** **Databricks step operator** (flavor `databricks` on the `step_operator` component type) so **individual steps** can run on Databricks while the **active orchestrator** stays non-Databricks or different.
- A **shared Pydantic settings base** (`DatabricksBaseSettings` in `databricks_shared_settings.py`) that powers both `DatabricksOrchestratorSettings` and `DatabricksStepOperatorSettings`, with **one consistent field set** and field-level descriptions of who consumes each value (orchestrator-only vs both).
- **New optional** orchestrator- and step-level **configuration** on that base: resource **tagging** (`custom_tags`, `job_tags` where applicable), **job access control** entries, **timeouts**, **concurrency and retries** for jobs (orchestrator), **init scripts** (DBFS-only), **custom Docker** image and registry auth, **driver** node type, and related cluster and Spark options not present on `develop`’s old orchestrator settings class.
- **Documentation** in the component guide: `docs/book/component-guide/step-operators/databricks.md` (new) and `docs/book/component-guide/orchestrators/databricks.md` (updated, including a tagging section).
- **Wider** `zenml.integrations.databricks.flavors` exports (see [Package exports](#package-exports-zenmlintegrationsdatabricksflavors)).
- A **small** public-adjacent change: **console logging** may behave slightly differently in edge cases (see [Other compatibility notes](#other-compatibility-notes)).

#### New component: Databricks step operator (public surface)

| Item | User-facing detail |
|------|---------------------|
| Integration | `zenml integration install databricks` (installs `databricks-sdk` and integration deps). |
| Flavor | `databricks` (same string as the Databricks **orchestrator** flavor, but a different **stack component type**). |
| Stack config class | `DatabricksStepOperatorConfig` with `host`, and optional `client_id` and `client_secret` (service principal style auth; same story as the orchestrator in the guide). |
| Per-step / pipeline settings class | `DatabricksStepOperatorSettings` (inherits the shared `DatabricksBaseSettings` fields; attach via `settings={"step_operator": ...}` on `@step` or the pipeline as appropriate to your ZenML version patterns). |
| Register (CLI) | `zenml step-operator register <NAME> --flavor=databricks --host=... --client_id=... --client_secret=...` (see book for placeholders). |
| How steps bind | `@step(step_operator="<registered_name>")` |
| Stack requirement called out in docs | **Remote artifact store** so the orchestrator-side run and Databricks can both use artifacts. No container registry or image builder required for this flavor. |
| Execution model (user story) | ZenML builds a **wheel**, **uploads** to the workspace, submits a **one-time** Databricks run for the step, runs the ZenML entrypoint. Wheels live under `Workspace/Shared/.zenml`; successful runs may **delete** the step’s wheel directory; **failed** runs may **leave** it for debugging. |
| Docs | `docs/book/component-guide/step-operators/databricks.md` — includes warnings about which shared fields **do not** apply in one-time run mode (see [Step operator: fields not applied](#step-operator-fields-not-applied-in-one-time-run-mode)). |

#### Shared settings: `DatabricksBaseSettings` (complete field reference)

Both `DatabricksOrchestratorSettings` and `DatabricksStepOperatorSettings` subclass this base. Unless a row says otherwise, the field is a normal optional Pydantic field with defaults where noted. **Databricks platform limits** (e.g. tag caps) are enforced at Databricks; ZenML surfaces them in field descriptions and may reject at validation in some cases.

**Cluster, Spark, and instance options**

- `spark_version` (optional `str`) — Apache Spark / Databricks runtime for the cluster; if omitted, the workspace default applies.
- `num_workers` (optional `int`, ≥ 0) — fixed worker count; do not combine with autoscaling; see docs for fixed vs `autoscale`.
- `autoscale` (`(min, max)` workers, default `(0, 1)`) — autoscaling bounds. Default is intentionally driver-friendly (driver-only with optional one worker at upper bound, per model description). Validators require non-negative counts and `min ≤ max`.
- `node_type_id` (optional `str`) — worker node Databricks **instance** type.
- `driver_node_type_id` (optional `str`) — **driver** node type if different from workers.
- `policy_id` (optional `str`) — Databricks **cluster policy** for governance and guardrails.
- `autotermination_minutes` (optional `int`, ≥ 0) — idle shutdown for cost control.
- `single_user_name` (optional `str`) — **single-user** access mode for the cluster in Databricks terms.
- `spark_conf` (optional `str → str` map) — Spark configuration key-values.
- `spark_env_vars` (optional `str → str` map) — environment variables for Spark driver/executors.
- `init_scripts` (optional list of `str`) — **must** be `dbfs:/…` paths; other prefixes fail validation in ZenML.
- `availability_type` (optional enum) — `ON_DEMAND`, `SPOT`, or `SPOT_WITH_FALLBACK` for instance availability.

**Tagging and project metadata**

- `custom_tags` (optional `str → str` map) — up to **45** tags; applied to **underlying** cloud resources (e.g. instances, EBS) as described in the model and book.
- `job_tags` (optional `str → str` map) — up to **25** tags; described as applied to the **Databricks job** and **forwarded as cluster** tags. **Model text:** used **only** by the **orchestrator**; see step-operator [exceptions](#step-operator-fields-not-applied-in-one-time-run-mode).

**Scheduling, jobs, and governance (orchestrator and/or job lifecycle)**

- `schedule_timezone` (optional `str`) — **IANA** timezone (validated in ZenML). **Used** when the **orchestrator** runs **scheduled** pipelines (per model and book). **Not** applied in step one-time mode (see [below](#step-operator-fields-not-applied-in-one-time-run-mode)). Required for Databricks schedule UX per orchestrator book.
- `access_control_list` (optional list of `DatabricksAccessControlRequest`) — each entry grants a **group**, **user email**, or **service principal** (exactly one principal per entry) a `permission_level` (`CAN_VIEW`, `CAN_MANAGE_RUN`, `CAN_MANAGE`, `IS_OWNER`). If omitted, only the job creator has access, per model description. Usable for **orchestrated** jobs; the step operator also passes **ACLs** and **timeouts** for its submit path (code paths differ but the **settings** are shared).
- `timeout_seconds` (optional `int`, ≥ 0) — timeout for a **run**; `0` = no limit (per field text).
- `max_concurrent_runs` (optional `int` in `[1, 1000]`) — **orchestrator only** per model description. Not applied in step one-time mode [below](#step-operator-fields-not-applied-in-one-time-run-mode).
- `task_timeout_seconds` (optional `int`, ≥ 0) — per **task** (step) in the Databricks **job** model.
- `max_retries` (optional `int`, ≥ **-1**; `-1` = unlimited) — **orchestrator only** per model. Not in step one-time [below](#step-operator-fields-not-applied-in-one-time-run-mode).
- `min_retry_interval_millis` (optional `int`, ≥ 0) — **orchestrator only**; not in step one-time [below](#step-operator-fields-not-applied-in-one-time-run-mode).
- `retry_on_timeout` (optional `bool`) — **orchestrator only**; requires `max_retries` in model description. Not in step one-time [below](#step-operator-fields-not-applied-in-one-time-run-mode).

**Container image for the cluster**

- `docker_image_url` (optional `str`) — image the cluster can pull from a registry.
- `docker_image_username` / `docker_image_password` (optional; password is a secret) — must be **both** set or both omitted; ZenML validation enforces the pair.

#### Step operator: fields not applied in one-time run mode

The step operator book lists settings that are **not** applied when executing **one-off** Databricks **submit run** style execution (scheduling, long-lived **job** features, and certain job-level **tags/retries** that belong to a durable job definition in Databricks). **Not applied (per published doc string):** `schedule_timezone`, `job_tags`, `max_concurrent_runs`, `max_retries`, `min_retry_interval_millis`, `retry_on_timeout`.

**Additionally:** generic ZenML `resource_settings` are **not** mapped to Databricks cluster size; the Databricks-specific settings above are the levers (callout in the same doc).

#### Orchestrator: what is new or expanded vs `develop`

On **`develop`**, `DatabricksOrchestratorSettings` defined only: `spark_version`, `num_workers`, `node_type_id`, `policy_id`, `autotermination_minutes`, `autoscale`, `single_user_name`, `spark_conf`, `spark_env_vars`, `schedule_timezone`, and `availability_type` (and `DatabricksAvailabilityType` lived in the orchestrator flavor module).

**Preserved (same names, still on the orchestrator settings / config after refactor):** all of those **except** the enum type’s **module** location (see [Imports and enums](#imports-and-enums)) — the **field names and intent** for settings remain, with **stricter** validation in places (see [Validation](#validation-behavior-changes)). **Default** for `autoscale` remains the same numeric default `(0, 1)`.

**New (optional) on the orchestrator via the shared base (user-visible capabilities):** `custom_tags`, `job_tags`, `access_control_list`, `timeout_seconds`, `max_concurrent_runs`, `task_timeout_seconds`, `max_retries`, `min_retry_interval_millis`, `retry_on_timeout`, `driver_node_type_id`, `init_scripts`, `docker_image_url`, `docker_image_username`, `docker_image_password`, plus the expanded descriptions and in-model limits for `num_workers` / `autotermination_minutes` / etc.

**Orchestrator book updates** add a “Tagging Databricks resources” section and tie scheduling to IANA timezones. GPU examples remain aligned with the shared settings.

#### Imports and enums

**Breaking change (integrator-facing):** `DatabricksAvailabilityType` was **moved** from `databricks_orchestrator_flavor.py` to `databricks_shared_settings.py`. The type is **unchanged** at runtime (same name, same values); only the **Python module** that defines it changed. Any code that still does `from zenml.integrations.databricks.flavors.databricks_orchestrator_flavor import DatabricksAvailabilityType` will **fail** with an import error after upgrade. **Mitigation:** `from zenml.integrations.databricks.flavors import DatabricksAvailabilityType` or import from `databricks_shared_settings` (the `flavors` package re-exports the symbol; see the Package exports subsection in this document). There is nothing to “fix” in ZenML beyond the move itself; this is a **documented breaking change** for downstream importers. **Release notes:** call this out explicitly (see [DOC-RELEASE-1] in Backlog).

- New public types (additional): `DatabricksPermissionLevel`, `DatabricksAccessControlRequest` (see exports).

#### Validation behavior changes (no field removals)

- `schedule_timezone` must be a **valid IANA** ID where set; `develop` did not validate in ZenML before. Values that Databricks might have received anyway can now **fail in ZenML** at settings parse time.
- `autoscale` and `init_scripts` and **Docker** credentials are validated in the **shared** model; invalid combinations **fail early** in ZenML.
- `num_workers` and other numeric fields use `ge=0` (or stated bounds) on the **branch**; extreme invalid values that previously only failed downstream may now fail in ZenML.

#### Package exports (`zenml.integrations.databricks.flavors`)

The package `__init__` exports: `DatabricksOrchestrator{Flavor,Config,Settings}`, `DatabricksStepOperator{Flavor,Config,Settings}`, `DatabricksModelDeployer{Flavor,Config}`, and from shared: `DatabricksAccessControlRequest`, `DatabricksAvailabilityType`, `DatabricksPermissionLevel` (per current `__all__` in the branch).

#### Other compatibility notes

- **Console logging** ([`src/zenml/logger.py`]): step-name prefixing no longer **swallows** arbitrary exceptions; stdout/stderr wrap handles **circular import** to `LoggingContext` in a more explicit way. Affects only unusual extensions or tests that assumed silent failure.

#### Summary table (breaking vs non-breaking, user lens)

| Topic | Verdict for typical users |
|--------|---------------------------|
| Existing orchestrator field **names** on the prior settings set | **Present**; nothing removed. |
| New **optional** settings and step operator | **Additive**; opt-in. |
| `DatabricksAvailabilityType` **import path** (moved to `databricks_shared_settings`) | **Breaking** for anyone who imported it from `databricks_orchestrator_flavor`; use `from zenml.integrations.databricks.flavors import DatabricksAvailabilityType` (or `databricks_shared_settings`). **Document in release notes.** |
| Stricter IANA and structured validation | **Can break** atypical or previously invalid config strings. |
| Logger | **Edge** behavior only. |

These user journeys re-use the vocabulary and field names from the **Breaking, compatibility, and public feature reference** at the start of this stage; that section is the full catalog, journeys are narrative.

### User journey candidates

1. A **data engineer** keeps **Kubeflow (or any non-Databricks orchestrator)** for the graph but needs **one Spark-heavy or Databricks-only step**—they register the Databricks **step operator**, add it to the stack, and mark that step with `step_operator="…"`.
2. A **platform owner** standardizes **cost and ownership tags** on **full pipelines** run by the Databricks **orchestrator** using `custom_tags` / `job_tags` in pipeline or stack settings.
3. An **integrator** reads the **new flavor exports** (`DatabricksAccessControlRequest`, permission/availability enums) to type **access control** or **cluster policy** in code.
4. A **new adopter** follows `docs/book/component-guide/step-operators/databricks.md`: `zenml integration install databricks`, `zenml step-operator register`, then `@step(step_operator=…)`.

N1 — **Wrong workspace or leaked credentials:** stack config points `host` / OAuth **client_id** / **client_secret** at production; jobs and wheel uploads go to the wrong tenant or secrets sit in a shared stack (governance and blast radius).

N2 — **Access control mistakes:** `access_control_list` grants **CAN_MANAGE** / **IS_OWNER** to a **group** or **service principal** that should not own every pipeline job.

N3 — **Databricks platform limits / validation:** **tag count** (e.g. 25 / 45), invalid tag keys, or policy conflicts cause **job submit** to fail at the Databricks API boundary.

N4 — **Assumption mismatch on the step operator:** user sets **orchestrator-only** fields (e.g. `job_tags`, `schedule_timezone`, retry knobs) on **step** settings expecting them to apply; they are **ignored** for one-off step runs (docs warn; runtime may log).

### User journeys — expanded walkthroughs

#### Journey 1 — Offload one step to Databricks (primary)

| Aspect | Notes |
|--------|--------|
| Intent | Run a **single** pipeline step on Databricks while the **rest of the run** uses the **active orchestrator** (not Databricks). |
| How UX changed | **Before (develop):** No first-party ZenML Databricks **step operator**; teams either used the Databricks **orchestrator** for the whole graph or used other patterns. **After (branch):** Register a **step operator** with flavor `databricks`, add it to a stack with a **remote** artifact store, and attach `@step(step_operator="your_component_name")`. |
| Preconditions | Databricks **workspace**; credentials with permission to **submit runs**; `zenml integration install databricks`; **remote** artifact store on the stack. |
| Scene | A pipeline’s **feature** step should run where **Spark** and org data policies already live; **orchestrator** stays e.g. local or another cloud. |
| Act I (baseline) | Only **orchestrator-level** Databricks or workarounds; no **selective** step targeting. |
| Act II (new) | **Register** the component (see component guide for full CLI), **add to stack**, then in code: `from zenml import step` and `@step(step_operator="databricks_step_operator")` (name is whatever you registered). **Per-step** cluster shape via `DatabricksStepOperatorSettings` in `settings={"step_operator": …}`. |
| Config / contract (user-observable) | **Step operator config:** `host`, optional `client_id` / `client_secret` (same pattern as orchestrator). **Per-step settings (shared model):** cluster/Spark fields as in the reference (e.g. `spark_version`, `num_workers` or `autoscale`, `node_type_id`, `driver_node_type_id`, `policy_id`, `spark_conf`, `spark_env_vars`, `init_scripts` with `dbfs:/…`, `availability_type`, `custom_tags`, Docker image and registry auth, `access_control_list`, `timeout_seconds`, `task_timeout_seconds`). **Not applied** in one-time step mode (per book): `schedule_timezone`, `job_tags`, `max_concurrent_runs`, `max_retries`, `min_retry_interval_millis`, `retry_on_timeout`. **Not mapped:** generic ZenML `resource_settings` → Databricks cluster size. |
| Per-outcome | **Success:** Databricks **run** for that step; metadata such as run URL can surface on the step run. **Wheels** live under workspace path documented in the guide; success may **clean**; failure may **leave** artifacts for debug. **Ignored settings:** if orchestrator-only fields are set, behavior matches docs (no job-level tags on step mode). **Failure:** Databricks API errors (auth, policy, quota), or stack validation (e.g. local components with Databricks orchestrator—**step operator** has its own validator; check docs for *your* stack mix). |
| Caveats and N# | N1: **Confirm host and secrets** for the right workspace. N3: Databricks **limits** on tags and resources. N4: **Read** the warnings block in the step-operator doc so **job_tags** on a step is not confused with **orchestrator** job tags. |

**Copy-paste shape (illustrative names):**

```python
from zenml import step
from zenml.integrations.databricks.flavors import DatabricksStepOperatorSettings

databricks_settings = DatabricksStepOperatorSettings(
    spark_version="15.3.x-scala2.12",
    num_workers=3,
    node_type_id="Standard_D4s_v5",
    spark_conf={},
)

@step(
    step_operator="databricks_step_operator",
    settings={"step_operator": databricks_settings},
)
def databricks_step():
    ...
```

#### Journey 2 — Orchestrator tags (before / after, shorter)

| Aspect | Notes |
|--------|--------|
| Intent | **Tag** Databricks **jobs** and **cluster resources** for **finance / governance** when using the **Databricks orchestrator**. |
| Act I | **Before:** No `job_tags` / `custom_tags` on `DatabricksOrchestratorSettings` (in this form). **After:** Optional `Dict[str, str]` maps; **`job_tags`** for the **job** (orchestrated pipelines); **`custom_tags`** for **resource-level** tags (e.g. AWS), subject to Databricks **limits** in field descriptions. |
| Act II | `DatabricksOrchestratorSettings(custom_tags=…, job_tags=…)` in pipeline or stack **settings** for the orchestrator. |
| N# | N3: Enforce **tag** budgets and naming standards org-wide; **API** rejects over-limit or invalid keys. |

### Session notes / questions

- (Open) Exact **metadata keys** end users see on the step run in the **Dashboard** (run URL, job id) — confirm against released UI copy.
- (User) Any **naming** confusion between flavor **`databricks`** on **orchestrator** vs **step operator** in support channels?
- **UX / API design (2026-04-29):** `DatabricksStepOperatorSettings` inherits the full `DatabricksBaseSettings` surface, including fields the step operator **does not apply** (`schedule_timezone`, `job_tags`, `max_concurrent_runs`, `max_retries`, `min_retry_interval_millis`, `retry_on_timeout`). They are ignored at submit time with a log warning. Rationale in implementation is DRY / shared validation vs duplicating the cluster/Spark model. **Reviewer position:** that trades **user-facing clarity** (IDE autocomplete, mental model, “if it is on the class, it should work”) for **fewer duplicated models**; treating it as an acceptable long-term state is a **poor** default. **Preferred follow-up:** a slimmer public settings type for the step operator (e.g. compose a shared *cluster-only* mixin or nested model) so orchestrator-only job lifecycle fields are **not** on the step operator’s public class. Tracked in **Backlog** as a review flag / action item.
- **Breaking / release notes (2026-04-29):** `DatabricksAvailabilityType` **moved** to a new module (`databricks_shared_settings`); old import from `databricks_orchestrator_flavor` **breaks**. No product behavior change to the enum itself—**integrator-only** break. **Action:** list under **breaking changes** in release notes with the two supported import styles. Tracked as **[DOC-RELEASE-1]**.

### Decisions

- Stage 2 treats **logger** / **wheel** refactors as **non–user-journey** for the public pass; they remain in scope for Stages 3–4.
- Session captured **two** follow-ups: (1) step operator settings bloat **[REVIEW-UX-1]**, (2) **release-note** breaking entry for `DatabricksAvailabilityType` import move **[DOC-RELEASE-1]**.

---

## Stage 3 — Design / architecture

### 1. Summary (design intent)

**Problem at the architecture level:** Run ML code on Databricks with (a) **full-pipeline** jobs created via the Jobs API and (b) **single-step** execution via submit-run, while sharing **one** cluster/settings model and **one** utility layer so orchestrator and step operator do not fork Spark/cluster/Docker/tag logic.

**Major moving parts**

- `DatabricksBaseSettings` — single Pydantic model for cluster, Spark, tags, Docker, ACL, timeouts, and orchestrator-only job fields (`databricks_shared_settings.py`).
- `databricks_utils.py` — builds `ClusterSpec`, task vs submit-task payloads, ACL lists, wheel upload paths, run status mapping; both execution paths call into it.
- `DatabricksOrchestrator` — extends `WheeledOrchestrator`; creates a **Databricks job** (`jobs.create`), then `run_now`; tasks built with `convert_step_to_task`.
- `DatabricksStepOperator` — extends `BaseStepOperator`; calls `jobs.submit` with a single task from `convert_step_to_submit_task`.
- `wheel_build_utils.py` — wheel creation / package naming / repo copy; **extracted** from `WheeledOrchestrator` for reuse (orchestrator and step operator both use `create_wheel` / `prepare_repository_copy_for_wheel`).
- `logger.py` — safer console formatting and import-cycle handling (cross-cutting, not Databricks-specific domain).

**Out of scope for this branch (no changes in diff):** ZenML **server** REST API, **SQL** store, **migrations**, **dashboard** contracts. Stack components are still registered and stored as today; only **client-side** orchestration and integration code change.

---

### 2. Database level

**None.** No Alembic migrations, no new tables or columns, no altered store queries in the three-dot diff vs `develop`.

---

### 3. Domain models and REST API

**No new or changed ZenML server domain models or HTTP routes** in this diff. Public **Python** surface is the integration: `DatabricksOrchestrator{Config,Settings}`, `DatabricksStepOperator{Config,Settings}`, and shared types in `databricks_shared_settings.py` (see Stage 2). Serialization of stack components uses existing ZenML stack component mechanisms.

---

### 4. Shared settings and validation

`DatabricksBaseSettings` centralizes fields and validators (`autoscale`, `init_scripts` `dbfs:/`, `schedule_timezone` IANA, Docker username/password pairing):

```95:96:src/zenml/integrations/databricks/flavors/databricks_shared_settings.py
class DatabricksBaseSettings(BaseSettings):
    """Databricks execution settings shared by orchestrator and step operator."""
```

`DatabricksOrchestratorSettings` and `DatabricksStepOperatorSettings` are thin subclasses (same field set via inheritance — Stage 2 / [REVIEW-UX-1] documents the UX tradeoff).

---

### 5. Zen store / “store” in this design

**Not applicable** in the sense of a new `ZenStore` method. Persistence of runs/steps remains the existing ZenML client and server path. Databricks-specific **run metadata** (orchestrator URL, step run Databricks `run_id`, `job_id`, `run_page_url`, wheel directory) is written through existing **metadata** / step-run update paths (`publish_step_run_metadata`, `run_metadata` on the in-memory step run model in the step operator).

---

### 6. Databricks utils: cluster spec and ACL

`build_databricks_cluster_spec` maps `DatabricksBaseSettings` to a Databricks `ClusterSpec`, including `custom_tags`, Docker image, init scripts, autoscale vs `num_workers`, cloud availability attributes, policy resolution:

```537:609:src/zenml/integrations/databricks/utils/databricks_utils.py
def build_databricks_cluster_spec(
    databricks_client: DatabricksClient,
    settings: DatabricksBaseSettings,
    env_vars: Dict[str, str],
) -> ClusterSpec:
    ...
    return ClusterSpec(
        ...
        custom_tags=settings.custom_tags,
        ...
    )
```

`build_access_control_list` turns Pydantic `DatabricksAccessControlRequest` into SDK `iam.AccessControlRequest` objects for both `jobs.create` and `jobs.submit` callers.

`job_tags` are applied on the **orchestrator** path when creating the job (see below), not on the cluster spec in the snippet above (orchestrator uses `tags=settings.job_tags` on `jobs.create`).

---

### 7. Orchestrator flow (code anchors)

- **Schedule guardrails:** cron / `schedule_timezone` validated before building tasks (`databricks_orchestrator.py` — schedule and timezone checks in the `prepare` path; see lines ~200–218 in branch for Databricks schedule requirements).
- **Task graph:** `convert_step_to_task` with per-step retry fields from **orchestrator** settings (`max_retries`, `min_retry_interval_millis`, `retry_on_timeout`, `task_timeout_seconds`).
- **Job creation:** `jobs.create` with `tags=settings.job_tags`, `access_control_list`, `timeout_seconds`, `max_concurrent_runs`:

```386:395:src/zenml/integrations/databricks/orchestrators/databricks_orchestrator.py
        job = databricks_client.jobs.create(
            name=pipeline_name,
            tasks=tasks,
            job_clusters=[job_cluster],
            schedule=databricks_schedule,
            tags=settings.job_tags,
            access_control_list=build_access_control_list(settings),
            timeout_seconds=settings.timeout_seconds,
            max_concurrent_runs=settings.max_concurrent_runs,
        )
```

- **Run:** `jobs.run_now(job_id=...)` when `job_id` is present.
- **Wheels:** path under `DATABRICKS_WHEELS_DIRECTORY_PREFIX/.../orchestrator`; upload failure and failed submission delete the workspace directory (orchestrator).

---

### 8. Step operator flow (code anchors)

- **Validator:** remote **artifact store** only (`databricks_step_operator.py` `StackValidator` on `stack.artifact_store.config.is_local`).
- **Submit:** after wheel build/upload, `convert_step_to_submit_task` + `build_databricks_cluster_spec` for **new_cluster**; `jobs.submit` with ACL and run-level timeout:

```224:248:src/zenml/integrations/databricks/step_operators/databricks_step_operator.py
                task = convert_step_to_submit_task(
                    task_name=str(info.step_run_id),
                    command=entrypoint_command[0],
                    arguments=entrypoint_arguments,
                    libraries=collect_requirements(
                        docker_settings=info.config.docker_settings,
                        stack=Stack.from_model(stack_model),
                    ),
                    zenml_project_wheel=databricks_wheel_path,
                    new_cluster=build_databricks_cluster_spec(
                        databricks_client=self.client,
                        settings=settings,
                        env_vars=environment,
                    ),
                    timeout_seconds=settings.task_timeout_seconds,
                )
                submitted_run = self.client.jobs.submit(
                    access_control_list=build_access_control_list(settings),
                    run_name=f"{info.run_name}-{info.pipeline_step_name}",
                    tasks=[task],
                    timeout_seconds=settings.timeout_seconds,
                )
```

- **Metadata:** `run_id`, optional `job_id` and `run_page_url` from `get_run`, `wheel_directory` for post-success cleanup; `get_status` / `cancel` use `jobs.get_run` / `cancel_run` by stored `run_id`.

**Semantic difference from orchestrator:** `jobs.submit` is not a durable **job definition** with `job_tags` / schedule / `max_concurrent_runs` on the same object model as `jobs.create` — product docs (Stage 2) list which fields are **not** meaningful for ad-hoc submit; [REVIEW-UX-1] flags exposing orchestrator-only fields on the step settings class.

---

### 9. Wheel packaging refactor

`WheeledOrchestrator` delegates wheel helpers to `zenml/orchestrators/wheel_build_utils.py` (`create_wheel`, `prepare_repository_copy_for_wheel`, `sanitize_name`, etc.). The **step operator** imports the same helpers so it does not duplicate packaging logic. Single place to fix wheel generation behavior later.

---

### 10. Flow diagram (orchestrator vs step operator, condensed)

```mermaid
flowchart TB
  subgraph shared [Shared]
    BS[DatabricksBaseSettings]
    U[build_databricks_cluster_spec + ACL + task helpers]
  end
  subgraph orch [Orchestrator]
    WO[WheeledOrchestrator + wheel_build_utils]
    J[jobs.create + run_now]
  end
  subgraph step [Step operator]
    SO[DatabricksStepOperator]
    S[jobs.submit]
  end
  BS --> U
  U --> J
  U --> S
  WO --> J
  SO --> S
```

---

### 11. Boundaries, failure, and N# mapping (Stage 2)

- **N1 (wrong workspace / credentials):** `DatabricksClient` built from `host` and optional OAuth fields on **stack component** config; both paths call Databricks SDK on that client.
- **N2 (ACL):** `build_access_control_list` passes through to `jobs.create` and `jobs.submit` — same principal validation in Pydantic on `DatabricksAccessControlRequest`.
- **N3 (limits / policy):** cluster spec and job tags flow to Databricks API; failures surface as SDK/HTTP errors from the client.
- **Partial failure:** orchestrator deletes wheel directory on failed upload or failed job submission; step operator similar around submit; step operator can **clean** wheel dir after successful run via metadata-stored path.

---

### 12. Multi-tenancy, workspace paths, and Databricks resource identity (review discussion)

**Goal of this subsection:** record how **isolation** works (or does not) when many ZenML **servers**, **stack components**, and **pipeline runs** target the same Databricks **workspace** and **credentials**.

**Workspace (Databricks)**  
One `host` + one auth identity = one **Workspace** boundary in the Databricks product. The integration does not create “sub-workspaces.” All API calls for a given step operator config go through that single `WorkspaceClient`.

**Wheel upload path (step operator)**  
`databricks_step_operator.py` builds:

`{DATABRICKS_WHEELS_DIRECTORY_PREFIX}/{sanitize_name(pipeline.name)}/{sanitize_name(str(run_id))}/steps/{step_run_id}`

- **Prefix** is **global** in code today (`/Workspace/Shared/.zenml`) — see **[REVIEW-CONFIG-1]**.
- There is **no** ZenML **server id**, **stack component id**, or **workspace** name in the path; separation across independent ZenML deployments is **not** by a dedicated tenant folder, only by **UUIDs** (`run_id`, `step_run_id`) and the **sanitized pipeline name** segment.
- **Different step runs** (normal case): different `run_id` and/or `step_run_id` → **different** directories. **Same** step operator, many runs: **no** path collision in normal use.
- **`sanitize_name`:** non-alphanumeric characters collapse; **different** pipeline **names** can map to the **same** first path segment, but **different** runs still differ in the `run_id` segment.
- **Two** stack **components** with the same `host` and same SP: **shared** tree; paths do not encode **which** registered `step_operator` name was used. Isolation is still by run/step **UUIDs** in the path, not by component.

**Databricks objects per `jobs.submit` (step operator)**  
- **Run:** each `submit` returns a **new** Databricks `run_id` (new **Run** in the Jobs/Workflows UI). **Not** one long-lived run shared across steps.
- **Task:** `task_key` = `str(info.step_run_id)` (ZenML step run id) — unique per step run in ZenML.
- **Cluster:** each task is built with **`new_cluster=ClusterSpec`** — a **job-scoped** cluster for that run; typical **provision → execute → terminate** lifecycle. **Not** a single user-visible “ZenML shared cluster” reused for every step in the current design.
- **`run_name`:** `f"{info.run_name}-{info.pipeline_step_name}"` — **display** in Databricks; **not** a global unique key; multiple rows in the run list are still **distinct** runs by `run_id`.

**What is “shared” vs not**  
- **Shared** across all users of the same config: Databricks **identity**, **fixed** `.zenml` root (until configurable), and the **Workspace** as a whole.  
- **Not** shared between two **successful** distinct step runs: **run_id**, wheel **path** (per run/step), and **per-run** cluster lifecycle.

**Security / governance** hard boundaries (Databricks-side, not added by this PR): separate **workspaces**, **service principals**, **policies**, and **ACLs** for true multi-tenant separation at **cloud** level.

---

### 13. Cluster startup latency and possible future configuration

**Current behavior:** each step operator submit uses **`new_cluster`** in `convert_step_to_submit_task` (see §8), so Databricks **provisions** a new job cluster (per policy) for that run. That implies **startup latency** and **teardown** after the run; it favors **isolation** over **reusing** warm compute.

**Product follow-up (session, not in current PR):** it may be worth **extending** Databricks settings so users can opt into **faster** or **shared** compute, e.g.  
- **instance / cluster pool** id to shrink cold-start for **new** clusters, and/or  
- an **existing cluster id** (or Databricks’ “existing cluster” task shape) to **attach** runs to a long-lived or shared cluster instead of **`new_cluster` every time**,  
with clear tradeoffs (isolation, cost, noisy neighbor, who owns cluster lifetime). Tracked as **[REVIEW-PERF-1]** in Backlog.

### Session notes / questions (Stage 3)

- Confirm current branch still matches **line ranges** if large refactors land after this session (re-`git diff` / re-open files).
- Optional: add **sequence diagram** in Canvas for `submit` vs `create+run_now` if PR description needs a figure.
- **Workspace prefix:** wheel upload root is **not** configurable today (`DATABRICKS_WHEELS_DIRECTORY_PREFIX`). **Follow-up:** make it a **setting** — see **[REVIEW-CONFIG-1]** in Backlog.
- **Cluster reuse / startup (2026-04-29):** the design always uses **`new_cluster`** per step submit; there is **no** option for **instance pool** or **existing shared cluster** in the current `SubmitTask` wiring. A **worthwhile** follow-up is to add configuration for **pool id** and/or **existing cluster id** (and document tradeoffs) to **reduce** provisioning delay and support reuse — see **§13** and **[REVIEW-PERF-1]**.

### Decisions (Stage 3)

- Treat this PR as **integration-orchestration-only** for design review: **no** persistence-layer review checklist required beyond “unchanged.”
- Multi-tenancy and Databricks **resource identity** discussion is **canonically** captured in **§12–§13**; path-prefix and performance follow-ups in **Backlog** (**[REVIEW-CONFIG-1]**, **[REVIEW-PERF-1]**).

---

## Stage 4 — Implementation (code review)

### Explainer summary

This stage records a **thorough implementation review** of `develop...fix/add-tags-to-databricks-orchestrator` using the **stefan-reviewer** subagent (deep diff read, system-aware findings), then merged with this session’s prior flags (**[REVIEW-UX-1]**, **[DOC-RELEASE-1]**, **[REVIEW-CONFIG-1]**, **[REVIEW-PERF-1]**). It is **not** a line-by-line walkthrough of every hunk; it is **findings, gaps, and a thin file map** so implementers and reviewers can act on the list.

**Verdict (reviewer):** **COMMENT** — strong structure and test volume; address **major** items (policy resolution errors, default node type / cloud, hardcoded prefix, import migration) and several **minors** before or soon after merge per team bar.

---

### Implementation map (where to look)

| Area | Primary symbols / files |
|------|-------------------------|
| Shared settings + validation | `DatabricksBaseSettings`, `databricks_shared_settings.py`; flavors `__init__.py` exports |
| Cluster + task builders | `build_databricks_cluster_spec`, `convert_step_to_{task,submit_task}`, `_resolve_policy_id`, `_get_cloud_availability_attributes`, `databricks_utils.py` |
| Orchestrator | `DatabricksOrchestrator`, `_upload_and_run_pipeline`, `jobs.create` / `run_now`, `databricks_orchestrator.py` |
| Step operator | `DatabricksStepOperator.submit` / `get_status` / `cancel`, `databricks_step_operator.py` |
| Entrypoint (remote) | `DatabricksStepOperatorEntrypointConfiguration`, `databricks_step_operator_entrypoint_config.py` |
| Wheel packaging | `wheel_build_utils.py`, `WheeledOrchestrator` delegation |
| Cross-cutting | `logger.py` (circular import / `ConsoleFormatter`) |
| Tests | `tests/unit/integrations/databricks/`, `tests/unit/orchestrators/test_wheel_build_utils.py`, `tests/unit/test_logger.py` |

---

### Findings — major

| ID | Location (approx.) | Issue | Suggestion |
|----|--------------------|--------|------------|
| **M1** | `databricks_utils.py` `_resolve_policy_id` (~447–461) | When `policy_id` is set, `cluster_policies.get(policy_id=...)` is **uncaught**; invalid id or **permission** errors surface as raw SDK errors inside `build_databricks_cluster_spec`. | Try/except with a clear `ValueError` (or drop eager validation and let cluster creation fail with context). |
| **M2** | `databricks_utils.py` `DATABRICKS_WHEELS_DIRECTORY_PREFIX` (~73); orchestrator + step operator path builders | Hardcoded `/Workspace/Shared/.zenml` — no org/tenant override. | Same as **[REVIEW-CONFIG-1]**: configurable prefix with current default. |
| **M3** | `databricks_utils.py` `DATABRICKS_DEFAULT_NODE_TYPE_ID` (~75), used in `build_databricks_cluster_spec` | Default **`Standard_D4s_v5`** is **Azure**-skewed; AWS/GCP workspaces may error if `node_type_id` unset. | Require `node_type_id` with a clear validation message, or cloud-aware default from `host`, or **document** the Azure assumption. |
| **M4** | `DatabricksAvailabilityType` module move | Direct import from `databricks_orchestrator_flavor` **breaks**; `flavors` re-exports help but **deep** imports fail. | Same as **[DOC-RELEASE-1]**; optional **compat re-export** in `databricks_orchestrator_flavor.py` to avoid breaking integrators. |

---

### Findings — minor / warnings

| ID | Location (approx.) | Issue | Suggestion |
|----|--------------------|--------|------------|
| W1 | Step operator / utils typing | Lowercase `set[str]`, `tuple[...]` vs `typing` style elsewhere. | Consistency pass (non-blocking) if the repo style guide prefers one form. |
| W2 | `databricks_utils.py` `sanitize_labels` (~688–703) | Docstring still links **Kubernetes** label docs; body comments say Databricks. | Point to Databricks tag rules or a neutral description. |
| W3 | `databricks_orchestrator_flavor.py` config | `client_id` / `client_secret` are **Optional**; misconfiguration may only fail at **Databricks** auth time. | `model_validator`: both set or both unset (mirror Docker credential validation on shared settings). |
| W4 | `databricks_step_operator.py` `submit` (~260+) | `submitted_run.response` cast to `SubmitRunResponse`; depends on SDK return shape. | Defensive check or assert after `jobs.submit` if SDK types vary. |
| W5 | `databricks_orchestrator.py` wheel path | `sanitized` pipeline + run id path segments — **low** collision risk; worth awareness. | Document or test edge names. |
| W6 | `_get_cloud_availability_attributes` (~496–518) | **Heuristic** `host` substring matching for Azure/GCP vs AWS. | Document; optional explicit `cloud` setting follow-up. |
| W7 | `wheel_build_utils.py` | `python -m pip wheel` — **improvement** over fragile `pip` / cwd patterns. | None; positive note. |

---

### Findings — nits

| ID | Location (approx.) | Issue | Suggestion |
|----|--------------------|--------|------------|
| N1 | `databricks_shared_settings.py` | Line length / docstring polish. | Wrap to 80 cols where required by project. |
| N2 | `databricks_orchestrator_flavor.py` | Docstring wording for `config_class` return. | Tighten phrasing. |
| N3 | `databricks_step_operator.py` top-level `from zenml import __version__` | Eager import at module load. | Local import in `submit` if profiling matters (low). |
| N4 | `databricks_utils.py` `_resolve_policy_id` (~465) | **`assert policy.policy_id is not None`** in non-test code. | Explicit branch / `continue` or raise. |
| N5 | `databricks_orchestrator.py` vs step operator | Spark env merge: loop vs `dict.update`. | Align style with step operator. |
| N6 | `databricks_step_operator_flavor.py` | Logo URL under **orchestrator** asset path. | Optional dedicated asset (cosmetic). |

---

### Testing gaps (from review)

1. `_resolve_policy_id` when `cluster_policies.get` **raises** (M1).
2. `_get_cloud_availability_attributes` for **GCP** / **AWS** host patterns (not only Azure default-policy tests).
3. `map_databricks_run_to_execution_status` — e.g. `SUCCESS_WITH_FAILURES`, additional cancel / skip states if product requires distinction.
4. `get_databricks_wheel_source` with env set vs unset.
5. Databricks **client** construction when **only** `host` (no `client_id` / `client_secret`).
6. Orchestrator path with **schedule** (cron + `schedule_timezone`) end-to-end in a focused test.
7. Step operator `submit` **finally** `fileio.rmtree` on temp dir (if not already asserted).
8. Step operator `cancel` **NotFound** warning path.

---

### Security and tenancy (implementation lens)

- **Shared** `/Workspace/Shared/...` — readable/writable per workspace **ACLs**; document risk for multi-team workspaces (aligns with Stage 3 **§12** and **[REVIEW-CONFIG-1]**).
- **SecretField** for Docker password and Databricks secrets — good.
- **ACL** model: exactly one principal per `DatabricksAccessControlRequest` — good.
- `delete_workspace_directory` best-effort cleanup — appropriate for non-fatal teardown.

---

### Refactoring opportunities (non-blocking)

1. **Shared** `create_databricks_client(host, client_id, client_secret)` to dedupe orchestrator `._get_databricks_client` and step operator `.client` property.
2. **Context manager** for “upload then delete on failure” if cleanup triples further (optional).
3. `DATABRICKS_ZENML_DEFAULT_CUSTOM_REPOSITORY_PATH` single import — already centralized.

---

### Session notes / decisions (Stage 4)

- **Review method:** stefan-reviewer subagent on the same **three-dot** diff vs `develop`; line numbers are **approximate** to the branch at review time — revalidate after large rebases.
- **Overlap:** M2 = **[REVIEW-CONFIG-1]**; M4 = **[DOC-RELEASE-1]**; M3 and M1 are **new** tracked implementation risks in this table.
- **Next actions for author:** use the **Checklist** block in the subagent output (M1, M3, M4, W2, W3, tests) plus nits as time allows.

### Decisions (Stage 4)

- Stage 4 **complete** for this walkthrough: implementation findings, test gaps, and security notes are **documented above**; prior Backlog items remain until closed in product/PR work.

---

## Backlog

- Stage 2 open: Dashboard metadata labels for Databricks step runs; flavor name `databricks` on two component types (support FAQ).
- Stage 3: completed in this document (no DB / no REST; integration + two Databricks API paths).
- **Stage 4: completed** — stefan-reviewer implementation pass + session **Stage 4** section; open work is the **findings table** and **Backlog** items (**[REVIEW-UX-1]** through **[REVIEW-PERF-1]**, plus M1/M3 from Stage 4 table if not deduplicated to existing flags).

### Code review flags / action items

- **[REVIEW-UX-1] Databricks step operator settings surface (severity: design / follow-up, not merge-blocking unless team policy says otherwise).** The step operator’s public `DatabricksStepOperatorSettings` exposes fields that the step operator **does not** apply (shared inheritance from `DatabricksBaseSettings`). This optimizes for **code reuse** at the cost of **user-facing contract clarity** (false affordances, reliance on runtime warnings and docs). **Action:** either (a) refactor to a slimmer public model for the step operator that **omits** orchestrator-only fields while sharing implementation via a **cluster-only** sub-model or mixins, or (b) explicitly document and accept the pattern in ADR style—**not** let it pass unremarked. **Principle:** settings visible on a flavor’s settings class should **do** something for that flavor, or not appear on that class.

- **[DOC-RELEASE-1] `DatabricksAvailabilityType` import path (breaking for integrators).** The enum was **moved** from `databricks_orchestrator_flavor` to `databricks_shared_settings`. **No** behavioral change to the type; **any** external code that imported it from the old module file path will break on upgrade. **Required:** add a **Breaking changes** (or **Migration**) bullet in the **release notes** with: (1) what changed, (2) old import example that fails, (3) supported replacements: `from zenml.integrations.databricks.flavors import DatabricksAvailabilityType` and/or `from zenml.integrations.databricks.flavors.databricks_shared_settings import DatabricksAvailabilityType`. **No code change required** in ZenML beyond what the PR already does; this is **documentation and comms** only.

- **[REVIEW-CONFIG-1] Databricks workspace wheel directory prefix (multi-tenancy / ops).** The integration uses a **fixed** constant `DATABRICKS_WHEELS_DIRECTORY_PREFIX` (currently `/Workspace/Shared/.zenml`) in `databricks_utils.py` for both orchestrator and step operator uploads. All ZenML runs using the same Databricks workspace and identity therefore share a **common** root with no per–ZenML-server, per-tenant, or per–stack-component segment. **Request:** make this path **user-configurable** (e.g. a setting on `DatabricksOrchestratorConfig` / `DatabricksStepOperatorConfig` or on shared settings, with a safe default matching today’s value) so teams can isolate subtrees, align with Databricks folder policies, or run multiple independent ZenML deployments against one workspace with **separate** roots. Track as a product/config follow-up if not in scope for the current PR.

- **[REVIEW-PERF-1] Databricks instance pool and/or existing cluster (faster / reusable compute).** Today the step path (and orchestrator job clusters) use **`new_cluster=ClusterSpec`**; each run provisions **new** job compute, which implies **provisioning latency** and **no** reuse of a long-lived cluster. **Request:** optional settings to (a) attach a **cluster instance / driver pool** id to speed up **new** cluster creation where Databricks supports it, and/or (b) allow an **existing cluster id** (Jobs API `existing_cluster_id` style) for tasks that should **not** create a new cluster per run, with **documented** tradeoffs (isolation, cost, noisy neighbor, operational ownership of the shared cluster). Align orchestrator and step operator behavior if both gain the same knobs. See Stage 3 **§12–§13** for context.

PR #4649, opened 2026-03-27
Fix deployer dashboard HTML when Jinja2/Starlette template loading breaks

stefannica:
## Describe changes
The FastAPI deployment app’s dashboard catch-all route used Jinja2Templates.TemplateResponse(). On some environments (e.g. Kubernetes with newer Jinja2), that path can raise TypeError: unhashable type: 'dict' during template load/render, which prevents the server from starting when the dashboard is enabled.

What changed:
* Replaced `templates.TemplateResponse(...)` with explicit `templates.get_template("index.html")`, `template.render(request=..., service_info=...)`, and `HTMLResponse(content=html)`.
* Behavior for clients is unchanged: same template, same context (request, service_info), same HTML response.

Why this approach: Rendering via `Template.render` and returning `HTMLResponse` avoids Starlette’s `TemplateResponse` wrapper and keeps template loading independent of per-request context being threaded through APIs that interact badly with Jinja’s template cache across Jinja2/Starlette versions.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/deployers/server/fastapi/app.py
+++ b/src/zenml/deployers/server/fastapi/app.py
@@ -209,6 +209,13 @@ class FastAPIDeploymentAppRunner(BaseDeploymentAppRunner):
 
             # everything else is directed to the index.html file that hosts the
             # single-page application - this is to support client-side routing
+
+            # Using get_template + Template.render + HTMLResponse instead of
+            # Jinja2Templates.TemplateResponse. Starlette's _TemplateResponse
+            # path can interact with Jinja 3.1.x template caching around globals
+            # and raise TypeError (e.g. unhashable dict) in some environments.
+            # Explicit load and render preserves the same template context
+            # (request, service_info) that TemplateResponse would set.
             template = templates.get_template("index.html")
             html = template.render(
                 request=request,


stefannica:
Thanks for the detailed review.

> The switch from `templates.TemplateResponse()` to `templates.get_template().render()` in `catch_all_endpoint` bypasses Starlette's response wrapper, but there's no error handling added around either `get_template("index.html")` or `template.render()`. If the underlying issue causing template loading to break (as the PR title describes) still occurs, a `TemplateNotFound` or rendering exception will still propagate as an unhandled 500 — the fix appears to address a specific Starlette/FastAPI wiring issue without documenting or guarding against the root cause.

`TemplateResponse` ultimately called `template.render()` in Starlette’s `_TemplateResponse`, so `TemplateNotFound` / render failures would have surfaced the same way before; this change doesn’t worsen that. 

> It would be worth adding a comment explaining _why_ `TemplateResponse` was failing (e.g., a conflict with how Starlette resolves the template context in this particular deployment setup), since the behavioral difference isn't obvious from the diff alone.

Done

> Additionally, `jinja2.Template.render()` is synchronous and called inside an `async def` function — while this is likely fine for a small HTML file, it does block the event loop; `template.render_async()` with `await` would be the more correct approach in an async context.

I kept sync render() to match Starlette’s existing behavior for this response type. render_async would need an async-enabled Jinja environment and is out of scope for this bugfix.

> Finally, consider whether the `request` object being passed to `template.render()` behaves identically here as it did via `TemplateResponse`, particularly if the template uses any Starlette-specific URL helpers that rely on the request being set in a specific context.

We still pass `request` (and `service_info`) into `render()`, same as the context `TemplateResponse` assembled, so `url_for` and other `@pass_context` helpers that read `context["request"]` behave the same.

PR #4651, opened 2026-03-29
Introduce a container engine abstraction and add podman support

stefannica:
## Describe changes

The responsibilities related to container image management involving OCI-compatible clients such as Docker were spread across several components (the image builder, the container registry) and the old `zenml.utils.docker_utils` helpers. That layout made it hard to extend ZenML for engines that are not the Docker daemon + docker-py stack (for example **Podman**).

This change introduces a dedicated **`ContainerEngine`** abstraction in **`zenml.container_engines`**, with **`DockerContainerEngine`** and **`PodmanContainerEngine`** implementations. Local OCI operations—build, tag, push, “is this image local?”, and repository digest resolution live on that surface. Engines are **not** bound to a single registry in their constructors; registry credentials and validation flow through the stack’s **container registry** (and connector-driven CLI login where applicable) on each operation.

**`LocalImageBuilder`** stays the stack component users configure for **local** builds, but it resolves the active backend via **`get_container_engine()`**, so there is still a single **local** image builder flavor—no separate “local Podman” flavor.

Users choose Docker vs Podman through **global ZenML configuration**: the **`container_engines`** setting (typed as **`ContainerEngineType`**: `docker` or `podman`), also overridable via **`ZENML_CONTAINER_ENGINE`**. When unset, ZenML **auto-detects** (Docker first, then Podman). Default effective behavior remains Docker-first for existing environments.

### Implementation details

**`zenml.container_engines`** provides:

- **`ContainerEngine`** (ABC) — shared contract and helpers such as **`.parse_dockerignore()`** / **`.sanitize_tag()`** where they belong for all engines.
- **`DockerContainerEngine`** — Docker CLI and **docker-py** (build via API or optional `docker build` subprocess, push, inspect, registry digest lookup when meaningful).
- **`PodmanContainerEngine`** — **Podman CLI** only (build with tar-on-stdin like `docker build -`, `podman tag`, `podman image inspect` for the local-image heuristic, `podman login` + `podman push`).

**`get_container_engine()`** / **`check_container_engine()`** centralize selection, availability checks, and (for call sites that must pin the daemon API) **`get_container_engine(ContainerEngineType.DOCKER)`**.

**`LocalImageBuilder.build`** uses **`get_container_engine()`** for build and push; pipeline utilities (**`PipelineDockerImageBuilder`**, **`BuildConfiguration.compute_settings_checksum`**, etc.) call the same engine API for **tag**, **parent pull vs local** heuristics, and **`get_image_repo_digest`** where the checksum should include a parent digest for **local** builds.

**IMPORTANT**: Podman still cannot reliably report remote manifest digests (see [containers/podman#14779](https://github.com/containers/podman/issues/14779)), so the Podman path continues to use **uniquely generated tags** as a stable push identity—aligned with the approach already used elsewhere (e.g. AWS image builder).

The **container registry** stack component remains required for pushes in this flow and is still used to:

- **Validate** that the image name matches the configured registry URI.
- Run **`prepare_image_push`** and any flavor-specific pre-push logic.
- Supply **credentials** for private registries.

**`zenml.utils.docker_utils` is removed** from the library. Behavior that still belongs in product code now lives under **`zenml.container_engines`** (base + Docker implementation). Remaining ad-hoc helpers used only by the tests deployment harnesses (for example **`build_image`**-style utilities) have been replaced with the use of `DockerContainerEngine.build`.

Docker-specific integrations that must talk to the Docker daemon API (**Docker deployer**, **local Docker orchestrator**, **container service**, server Docker provider, CLI checks, **BentoML**, Docker service connector verification, etc.) are updated to use **`container_engines`**—typically **`get_container_engine(ContainerEngineType.DOCKER)`** where a Podman backend would not be valid—instead of importing **`docker_utils`**.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/service_connectors/docker_service_connector.py
+++ b/src/zenml/service_connectors/docker_service_connector.py
@@ -315,7 +315,7 @@ class DockerServiceConnector(ServiceConnector):
             is_available, error_message = engine.check_availability()
             if not is_available:
                 raise RuntimeError(
-                    f"The container engine is not available: {error_message}"
+                    f"No container engine available: {error_message}"
                 )
 
         return [resource_id] if resource_id else []


src/zenml/image_builders/base_image_builder.py
```diff
@@ -98,6 +99,102 @@ def build(
...
+            source: Existing local image reference.
+            target: Full image name including tag to apply.
+
+        Raises:
+            NotImplementedError: If the image builder does not support tagging
+                images.
+        """
+        raise NotImplementedError(
+            "Tagging images is not supported by this image builder. Use the "
+            "local image builder instead."
```
AlexejPenner: Base class shouldn't reference Local by name imo. This creates downward coupling and will become misleading if more subclasses are updated to use this. I suggest using `abstractmethod` or a generic message here and in the other methods

stefannica: done

src/zenml/utils/pipeline_docker_image_builder.py
```diff
@@ -251,16 +251,18 @@ def build_docker_image(
...
                 if container_registry:
-                    image_name_or_digest = container_registry.push_image(
-                        target_image_name
+                    image_name_or_digest = image_builder.push_image(
+                        target_image_name, container_registry
                     )
                 else:
                     image_name_or_digest = target_image_name
+            else:
+                image_name_or_digest = target_image_name
```
stefannica: Note for @schustmi : this means that unless you have a local image builder, the parent image will not be re-tagged and pushed to the container registry. I don't fully understand the implications of this. Should we do something else here, like warn the user ?

src/zenml/config/build_configuration.py
```diff
@@ -109,9 +109,15 @@ def compute_settings_checksum(
...
                 hash_.update(f.read())
 
-        if self.settings.parent_image and stack.container_registry:
-            digest = stack.container_registry.get_image_repo_digest(
-                self.settings.parent_image
+        if (
+            self.settings.parent_image
+            and stack.container_registry
+            and stack.image_builder
+            and stack.image_builder.is_building_locally
```
stefannica: Another note for @schustmi: this means that when a remote image builder is used, the parent image's digest will not be included in the hash. What is the impact of this ? Should I log a warning here or something ?

src/zenml/image_builders/local_image_builder.py
```diff
@@ -134,9 +274,138 @@ def build(
...
+        if self.config.container_runtime == LocalContainerRuntime.DOCKER:
+            return docker_utils.get_image_repo_digest(
+                image_name,
+                docker_client=self._get_docker_client(container_registry),
+            )
+
+        # Podman is currently unable to report the correct remote repo digest
+        # (see https://github.com/containers/podman/issues/14779). We return
+        # None in this case.
+        return None
```
schustmi: Raise a not implemented error, because it's not implemented?

src/zenml/image_builders/base_image_builder.py
```diff
@@ -98,6 +98,83 @@ def build(
             The Docker image repo digest or name.
         """
 
+    def tag_image(self, source: str, target: str) -> None:
```
schustmi: To be honest, all of the methods that you added here seem like no concern of the image builder abstraction to me.

One thing that makes this very clear is that none of the methods are implemented for any image builder other than the local one.

The repo digest for example is very clearly a concept of the container registry, which is also implementing the algorithm on how to assign that repo digest.

src/zenml/utils/pipeline_docker_image_builder.py
schustmi: I mean we had a solution that was working for all image builders before, maybe we just use that one? 😅

src/zenml/config/build_configuration.py
schustmi: See my other comments, this method makes simply no sense on the image builder and belongs on the container registry, where this PR for some reason removes it.

src/zenml/image_builders/local_image_builder.py
schustmi: But the real solution is simply the one we had before. Have it on the container registry, because that one can simply figure out how to get the repo digest, it doesn't matter what tool built it. Because it's the digest of the image in the repository stored in the `container registry`.

src/zenml/config/global_config.py
```diff
@@ -120,6 +122,7 @@ class GlobalConfiguration(BaseModel, metaclass=GlobalConfigMetaClass):
     user_email: Optional[str] = None
     user_email_opt_in: Optional[bool] = None
     analytics_opt_in: bool = True
+    container_engines: Optional[ContainerEngineType] = None
```
schustmi: Should this actually be a plural or only just a single `container_engine`?

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
```diff
@@ -84,6 +88,28 @@ def validator(self) -> Optional[StackValidator]:
             required_components={StackComponentType.IMAGE_BUILDER}
         )
 
+    @property
+    def docker_engine(self) -> DockerContainerEngine:
```
schustmi, line 92: In principal, we could also use podman to run the containers, right?

src/zenml/service_connectors/docker_service_connector.py
```diff
@@ -363,19 +303,19 @@ def _verify(
...
-        else:
-            assert resource_id is not None
-            self._authorize_client(docker_client, resource_id)
-            docker_client.close()
+        if ENV_ZENML_SERVER not in os.environ:
+            engine = get_container_engine()
+            is_available, error_message = engine.check_availability()
+            if not is_available:
+                raise RuntimeError(
+                    f"The container engine is not available: {error_message}"
```
schustmi:
```suggestion
                    f"No container engine available: {error_message}"
```

src/zenml/config/global_config.py
stefannica: This is in fact a bug. Should be `container_engine`. The env variable I documented doesn't even match this at all. Thanks for catching it !

src/zenml/orchestrators/local_docker/local_docker_orchestrator.py
stefannica: Correct! My not-so-secret future plan is to extend the container engine abstraction to support managing workloads too, then simply plug in the container engine into this orchestrator and the docker deployer and anywhere else this is needed.

schustmi: Nice!

schustmi (approved):

PR #4671, opened 2026-04-01
Allowing multiple components of the same type for a stack

### LINKED ISSUE #4629: Implement support for multiple stack components of the same type

_No response_

### Feature Description

Initial investigation can start with supporting step operators, experiment trackers and alerters. Eventually, the default should be support for multiple components (unless the type is something like an orchestrator).

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

bcdurak:
This PR allows people to register more than one component per type for specific types, namely Alerters, Step Operators, and Experiment Trackers.

###  Notable changes

- `StackCompositionSchema` was extended so a stack can store multiple components of the same type while still marking one default per type.
- A migration is added that adds this new column for all existing compositions.
- In the `StackResponse`, the first component within each list marks the default component.
- `Stack.components` is changed from “one component per type” to “list of components per type”. 
- `Stack.all_components` is implemented for callers that want flattened iteration.
- Backward-compatible singular accessors still work, e.g. `stack.alerter`, `stack.step_operator`, `stack.experiment_tracker` still return the default/first component.
- The CLI commands have been updates.
- Step decorator behaviour changed:
  - `@step(step_operator=True)` uses only the default step operator.
  - `@step(experiment_tracker=True)` activates all attached experiment trackers.
  - otherwise, you need to point to a component by name.
- Settings behaviour changed. You can adjust settings with:
  - `type`
  - `type.flavor`
  - `type:name` (this is what they resolve into if you use shortcuts now)
- Additional cleanup and bug fixes.

### TODO and Open Questions

- Adding settings for unactivated component probably should raise a warning
- Should we add a warning for components where two instances of the same flavor would be problematic? Like `mlflow`?


docs/book/component-guide/step-operators/README.md
```diff
@@ -40,6 +40,10 @@ zenml step-operator flavor list
 
 You don't need to directly interact with any ZenML step operator in your code. As long as the step operator that you want to use is part of your active [ZenML stack](https://docs.zenml.io/user-guides/production-guide/understand-stacks), you can simply specify it in the `@step` decorator of your step.
 
+{% hint style="info" %}
+Stacks can have multiple step operators attached. If you use `@step(step_operator=True)`, ZenML uses the default attached step operator, which is the first step operator of that type in the stack. If you want to use a non-default step operator, specify it explicitly by name in the `@step` decorator.
```
schustmi:
```suggestion
Stacks can have multiple step operators attached. If you use `@step(step_operator=True)`, ZenML uses the default attached step operator, which is the first step operator in the stack. If you want to use a non-default step operator, specify it explicitly by name in the `@step` decorator.
```

docs/book/how-to/steps-pipelines/configuration.md
```diff
@@ -143,8 +147,11 @@ Settings in ZenML are categorized into three main types:
   * `DeploymentSettings` for pipeline deployment configuration - can only be set at the pipeline level
 
 * **Stack-component-specific settings** for configuring behaviors of components in your stack:
-  * These use the pattern `<COMPONENT_CATEGORY>` or `<COMPONENT_CATEGORY>.<COMPONENT_FLAVOR>` as keys
-  * Examples include `experiment_tracker.mlflow` or just `step_operator`
+  * These use the pattern `<COMPONENT_CATEGORY>`, `<COMPONENT_CATEGORY>.<COMPONENT_FLAVOR>`, or `<COMPONENT_CATEGORY>:<COMPONENT_NAME>` as keys
+  * `<COMPONENT_CATEGORY>` targets the default attached component of that type
+  * `<COMPONENT_CATEGORY>.<COMPONENT_FLAVOR>` only works if exactly one attached component of that flavor exists in the stack
+  * `<COMPONENT_CATEGORY>:<COMPONENT_NAME>` targets a specific attached component by name
+  * Examples include `experiment_tracker`, `experiment_tracker.wandb`, or `step_operator:vertex`
```
schustmi, line 154: I think this should point out the `.` and `:` difference, very easy to miss

docs/book/how-to/steps-pipelines/configuration.md
```diff
@@ -133,6 +133,10 @@ def my_step():
 
 This approach allows you to use different components for different steps in your pipeline while also customizing their runtime behavior.
 
+{% hint style="info" %}
+If your stack contains multiple step operators, `@step(step_operator=True)` uses the default step operator, which is the first attached step operator of that type in the stack. If your stack contains multiple experiment trackers, `@step(experiment_tracker=True)` activates all attached experiment trackers for that step.
```
schustmi:
```suggestion
If your stack contains multiple step operators, `@step(step_operator=True)` uses the default step operator, which is the first attached step operator in the stack. If your stack contains multiple experiment trackers, `@step(experiment_tracker=True)` activates all attached experiment trackers for that step.
```

src/zenml/orchestrators/step_launcher.py
```diff
@@ -78,19 +82,19 @@ def _get_step_operator(
...
+        return step_operator
 
-    if step_operator_name and step_operator_name != step_operator.name:
+    step_operator = stack.step_operators.get(step_operator_name)
+    if step_operator is None:
         raise RuntimeError(
-            f"No step operator named '{step_operator_name}' in active "
-            f"stack '{stack.name}'."
+            f"No step operator named '{step_operator_name}' found in active "
+            f"stack '{stack.name}'. Available step operators: {stack.step_operators.keys()}."
```
schustmi: This will show as  `dict_keys([...])`, instead `list(stack.step_operators)` reads much nicer

src/zenml/stack/stack.py
```diff
@@ -279,131 +321,267 @@ def _raise_type_error(
...
+            ),
+            log_store=cast(
+                Optional["BaseLogStore"], log_store[0] if log_store else None
+            ),
         )
 
     @property
-    def components(self) -> Dict[StackComponentType, "StackComponent"]:
-        """All components of the stack.
+    def components(self) -> Dict[StackComponentType, List["StackComponent"]]:
```
schustmi: Breaking change, also some other changes in this class.

src/zenml/stack/stack.py
```diff
@@ -143,20 +152,34 @@ def __init__(
...
-        self._model_deployer = model_deployer
-        self._experiment_tracker = experiment_tracker
-        self._alerter = alerter
-        self._annotator = annotator
-        self._data_validator = data_validator
-        self._model_registry = model_registry
-        self._image_builder = image_builder
-        self._deployer = deployer
-        self._log_store = log_store
+        self._orchestrator = [orchestrator]
```
schustmi: Wouldn't it be easier to store them in a `_components` dictionary?

src/zenml/zen_stores/schemas/stack_schemas.py
```diff
@@ -284,8 +305,28 @@ def to_model(
...
+                StackComponentType, List["StackComponentSchema"]
+            ] = defaultdict(list)
+
+            sorted_compositions = sorted(
+                self.stack_compositions,
+                key=lambda composition: (
+                    composition.default_for_type is None,
+                ),
+            )
+            for composition in sorted_compositions:
```
schustmi, line 319: `get_query_options` needs to be adjusted for more efficient caching, especially now that we loop over compositions. For a stack with 10 components, this would now be 11 DB queries instead of 1.

src/zenml/zen_stores/schemas/stack_schemas.py
```diff
@@ -108,6 +126,16 @@ class StackSchema(NamedSchema, table=True):
     components: List["StackComponentSchema"] = Relationship(
```
schustmi, line 126: Do we still need this relationship?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -10239,19 +10240,21 @@ def create_stack(self, stack: StackRequest) -> StackResponse:
...
 
-                new_stack_schema = StackSchema.from_request(
-                    request=stack,
-                    components=defined_components,
-                )
+                for composition in self._create_stack_compositions(
+                    stack_id=new_stack_schema.id,
+                    component_ids_by_type=components_mapping,
+                ):
+                    session.add(composition)
```
schustmi: Seems like the method could add those themselves to make it a little simpler?

schustmi (changes requested):

docs/book/how-to/steps-pipelines/configuration.md
bcdurak: Explicitly explained it now.

src/zenml/orchestrators/step_launcher.py
bcdurak: Fixed this as well.

src/zenml/zen_stores/schemas/stack_schemas.py
bcdurak: We still use it in a few places 🤔

src/zenml/zen_stores/sql_zen_store.py
bcdurak: Refactored it as well.

src/zenml/zen_stores/schemas/stack_schemas.py
bcdurak: As discussed, I have added the component to the `get_query_options`. However, I am still a bit unsure about the `flavor_schema`. Each time we do `StackSchema.to_model()`, we will have the components but not their flavor schema which is used to form their bodies. That might end up making N queries as well.

src/zenml/stack/stack.py
bcdurak: Can you think of a way we can do this in a non-breaking way? If I leave the `components` as it is for instance (or make it a Union), I feel like things will become much more uglier.

bcdurak: I made it non-breaking now. Instead if someone uses the components property we throw a deprecation error.

src/zenml/stack/stack.py
bcdurak: I have refactored this. Feel free to take a look.

src/zenml/stack/stack.py
```diff
@@ -179,17 +205,31 @@ def from_model(cls, stack_model: "StackResponse") -> "Stack":
...
             }
+            stack_components: Dict[
+                StackComponentType, List["StackComponent"]
+            ] = defaultdict(list)
+
+            for component_model in hydrated_component_models:
+                hydrated_model = hydrated_component_models_by_id.get(
+                    component_model.id
+                )
+                if hydrated_model is None:
```
schustmi, line 226: I don't get this, seems lke this logically isn't possible?

schustmi: Actually the whole point of the `hydrated_component_models_by_id` I don't get. It does nothing :D

src/zenml/stack/stack.py
```diff
@@ -571,11 +983,15 @@ def log_store(self) -> "BaseLogStore":
         Returns:
             The log store of the stack.
         """
-        if not self._log_store:
+        if StackComponentType.LOG_STORE not in self._components:
             self.validate_log_store()
 
-        assert self._log_store is not None
-        return self._log_store
+        assert self._components.get(StackComponentType.LOG_STORE, None)
```
schustmi, line 989: Why is this necessary?

schustmi (changes requested):

src/zenml/stack/stack.py
bcdurak: You are right. I messed up the iteration. It should be fixed now.

src/zenml/stack/stack.py
bcdurak: It was just a control mechanism that we implemented back in the day. Each stack should have a log store after it has been validated.

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -548,8 +552,9 @@ def snapshot_request_from_source_snapshot(
         pipeline_update_exclude.add("parameters")
 
     if config.settings:
-        convert_component_shortcut_settings_keys(
-            settings=config.settings, stack=source_snapshot.stack
+        settings_utils.normalize_stack_component_setting_keys(
```
schustmi, line 613: Ah I just realized something: This is a breaking change. This will set the new settings keys in the snapshot, and that snapshot will be executed by a potentially old client which doesn't know how to interpret them. We can probably adjust this, so it does this depending on the zenml version that's stored in the snapshot?

schustmi (changes requested):

src/zenml/zen_server/pipeline_execution/utils.py
bcdurak: Good catch. I've implemented the required changes. Feel free to take another look.

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -90,6 +90,52 @@
 RUNNER_IMAGE_REPOSITORY = "zenml-runner"
 
 
+def use_legacy_stack_component_setting_keys_for_runner(
```
schustmi: `for_runner` doesn't really make sense here I think? Just `use_legacy_stack_component_setting_keys` or `should_use_legacy_stack_component_setting_keys` is clearer IMO

bcdurak: I have already changed it.

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -90,6 +90,52 @@
...
+    Args:
+        snapshot: The snapshot to check.
+
+    Returns:
+        Whether stack settings should use legacy keys for the snapshot runner.
+    """
+    runner_version: Optional[str] = None
+
+    if snapshot.build is not None:
+        runner_version = snapshot.build.zenml_version
```
schustmi: It's not about the runner version. It's about client version inside the docker images

bcdurak: Renamed it to `zenml_version`.

schustmi (approved):

PR #4676, opened 2026-04-02
feat: add structured logs to zenml-server and OTel instrumentation

### LINKED ISSUE #4675: Add structured log

_No response_

### Feature Description

Introducing structured logging in zenml-server using `structlog`

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

amitvikramraj:
## Changes I made

* Uses structlog logger instead of Python's stdlib logger. struclog logger object provides backwards compatibility with the stdlib logger.
* Server-side modules (`zen_server/`, `zen_stores/`) now emit structured events with keyword arguments instead of f-string messages.
* Request context (`request_id`, `method`, `path`, etc.) is automatically propagated via structlog contextvars — no manual interpolation needed.
* Added OpenTelemetry exporter for logs, traces and metrics.
* Keeps the old log coloring and rendering the console
* Also includes a few small improvements:
    * `X-Request-ID` in error responses to easily be able to filter logs
    * Added rate limit warning messages in logs

## Why

* Server logs were unstructured f-strings with manually threaded request IDs across 11+ files.
* This made them impossible to query in by tools like Loki/Garfana and painful to maintain.
* `structlog` provides parseable JSON output in the server and colored console output locally, with automatic context propagation.

## Before / After

**Before:**
```
[d5638eeb] API STATS - GET /api/v1/pipelines AUTHORIZING  [ active_requests: 1 memory_used_mb: 343.77 thread_count: 6 open_fds: 28 fd_limit: 65535 ]
[d5638eeb] SQL STATS - GET /api/v1/pipelines 'SqlZenStore.list_runs' STARTED  [ active_connections: 0 idle_connections: 3 overflow_connections: -17 ]
[d5638eeb] SQL STATS - GET /api/v1/pipelines 'SqlZenStore.list_runs' COMPLETED in 12.34ms  [ active_connections: 0 idle_connections: 3 overflow_connections: -17 ]
[d5638eeb] API STATS - 200 GET /api/v1/pipelines took 38.42ms  [ active_requests: 1 memory_used_mb: 343.77 thread_count: 6 ]
```

**After — console:**
```
2026-04-02T09:21:52Z [debug    ] request.received               [zenml.zen_server.middleware] client_ip=172.217.26.123 method=GET path=/api/v1/pipelines request_id=d5638eeb
2026-04-02T09:21:52Z [debug    ] sql.session.started            [zenml.zen_stores.sql_zen_store] active_connections=0 caller=SqlZenStore.list_runs idle_connections=3 request_id=d5638eeb
2026-04-02T09:21:52Z [debug    ] sql.session.completed          [zenml.zen_stores.sql_zen_store] caller=SqlZenStore.list_runs duration_ms=12.34 error=false request_id=d5638eeb
2026-04-02T09:21:52Z [debug    ] request.completed              [zenml.zen_server.middleware] duration_ms=38.42ms method=GET path=/api/v1/pipelines request_id=d5638eeb status_code=200
```

**After — JSON** (server default, queryable in Loki/Grafana):
```json
{"event":"request.received","level":"debug","logger":"zenml.zen_server.middleware","method":"GET","path":"/api/v1/pipelines","request_id":"d5638eeb","client_ip":"172.217.26.123","timestamp":"2026-04-02T09:21:52Z"}
{"event":"sql.session.started","level":"debug","logger":"zenml.zen_stores.sql_zen_store","caller":"SqlZenStore.list_runs","active_connections":0,"idle_connections":3,"request_id":"d5638eeb","timestamp":"2026-04-02T09:21:52Z"}
{"event":"sql.session.completed","level":"debug","logger":"zenml.zen_stores.sql_zen_store","caller":"SqlZenStore.list_runs","duration_ms":12.34,"error":false,"request_id":"d5638eeb","timestamp":"2026-04-02T09:21:52Z"}
{"event":"request.completed","level":"debug","logger":"zenml.zen_server.middleware","method":"GET","path":"/api/v1/pipelines","status_code":200,"duration_ms":"38.42ms","request_id":"d5638eeb","timestamp":"2026-04-02T09:21:52Z"}
```

## Steps to reproduce:

1. Run `docker compose up --build` — ZenML UI starts on port `3001`, Grafana UI runs on port `3002`
2. Open Grafana → Loki → Verify the logs

TODO: Later remove the docker-compose.


docker/zenml-server-dev.Dockerfile
```diff
@@ -88,9 +88,9 @@ COPY --chown=$USERNAME:$USER_GID src/zenml_cli/__init__.py ./src/zenml_cli/
 # NOTE: we uninstall zenml at the end because we install it separately in the
 # final stage
 RUN pip install --upgrade pip uv setuptools \
-    && uv pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex] "alembic==1.15.2" \
-    && uv pip uninstall zenml \
-    && uv pip freeze > requirements.txt
+  && uv pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex,otel] "alembic==1.15.2" \
```
Json-Andriopoulos: I am thinking `otel` might as well be part of the server dependencies.

src/zenml/zen_server/auth.py
```diff
@@ -87,7 +87,7 @@
     zen_store,
 )
 
-logger = get_logger(__name__)
+logger = structlog.get_logger(__name__)
```
Json-Andriopoulos: We shouldn't do that. We should configure the logging module at `root` level, then any logger we instantiate with the standard `logging.getLogger(__name__) ` instruction has the formatter changes in place.

Json-Andriopoulos: Something like that:

```
import logging
import structlog

shared = [
    structlog.stdlib.add_log_level,
    structlog.processors.TimeStamper(fmt="iso"),
]

formatter = structlog.stdlib.ProcessorFormatter(
    foreign_pre_chain=shared,
    processors=[
        structlog.stdlib.ProcessorFormatter.remove_processors_meta,
        structlog.processors.JSONRenderer(),
    ],
)

handler = logging.StreamHandler()
handler.setFormatter(formatter)

root = logging.getLogger()
root.handlers = [handler]
root.setLevel(logging.INFO)

logger = logging.getLogger(__name__)
logger.info("hello from stdlib logging")
```

That doesn't propagate changes regarding logging across our codebase but does the work we want (structured, JSON logs).

amitvikramraj: But if we call the std logger, then we would not be able to pass extra context.
```py
logger = logging.getLogger(__name__)
logger.info("hello from stdlib logging", ctx="something", another_ctx="something_else")
# ^^^this would not work
```

docker/zenml-server-dev.Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN set -ex && \
   apt-get update && \
   apt-get upgrade -y && \
   if [ "$INSTALL_DEBUG_TOOLS" = "true" ]; then \
-    apt-get install -y curl net-tools nmap inetutils-ping default-mysql-client mariadb-client git ; \
+  apt-get install -y curl net-tools nmap inetutils-ping default-mysql-client mariadb-client git ; \
```
stefannica: why were these changes necessary ? indentation was good here before

docker/zenml-server-dev.Dockerfile
```diff
@@ -88,9 +88,9 @@ COPY --chown=$USERNAME:$USER_GID src/zenml_cli/__init__.py ./src/zenml_cli/
 # NOTE: we uninstall zenml at the end because we install it separately in the
 # final stage
 RUN pip install --upgrade pip uv setuptools \
-    && uv pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex] "alembic==1.15.2" \
-    && uv pip uninstall zenml \
-    && uv pip freeze > requirements.txt
+  && uv pip install .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex] "alembic==1.15.2" \
```
stefannica: same here

docker/zenml-server-dev.Dockerfile
```diff
@@ -152,16 +152,18 @@ EXPOSE 8080
 
 ENTRYPOINT ["/entrypoint.sh"]
 CMD ["uvicorn", "zenml.zen_server.zen_server_api:app", "--log-level", "debug", "--no-server-header", "--proxy-headers", "--forwarded-allow-ips", "*", "--reload", "--port", "8080", "--host",  "0.0.0.0"]
+# CMD ["opentelemetry-instrument", "uvicorn", "zenml.zen_server.zen_server_api:app", "--log-level", "debug", "--no-server-header", "--proxy-headers", "--forwarded-allow-ips", "*", "--reload", "--port", "8080", "--host",  "0.0.0.0"]
```
stefannica: unused left-over comment

docker/zenml-server-dev.Dockerfile
```diff
@@ -152,16 +152,18 @@ EXPOSE 8080
...
 # and then remove the sources
 RUN pip install --no-deps --no-cache .[server,secrets-aws,secrets-gcp,secrets-azure,secrets-hashicorp,s3fs,gcsfs,adlfs,connectors-aws,connectors-gcp,connectors-azure,azureml,sagemaker,vertex] \
-    && rm -rf src README.md pyproject.toml
+  && rm -rf src README.md pyproject.toml
 
 EXPOSE 8080
 
 ENTRYPOINT ["/entrypoint.sh"]
 CMD ["uvicorn", "zenml.zen_server.zen_server_api:app", "--log-level", "debug", "--no-server-header", "--proxy-headers", "--forwarded-allow-ips", "*", "--port", "8080", "--host",  "0.0.0.0"]
+# CMD ["opentelemetry-instrument", "uvicorn", "zenml.zen_server.zen_server_api:app", "--log-level", "debug", "--no-server-header", "--proxy-headers", "--forwarded-allow-ips", "*", "--port", "8080", "--host",  "0.0.0.0"]
```
stefannica: unused left-over comment

src/zenml/zen_server/auth.py
```diff
@@ -56,7 +57,6 @@
     CredentialsNotValid,
     OAuthError,
 )
-from zenml.logger import get_logger
```
stefannica, line 59: Currently, all ZenML code uses the zenml.logger module and utilities. You can use that to your advantage: instead of replacing this with structlog in various modules, you should go to the source and update the zenml.logger functionality to use structlog instead when configured, which means that you automatically get structured logs everywhere instead of just a handful of modules.

Keep in mind that the same logger module is also used by client code, where you still want to keep unstructured logs as the default.

src/zenml/zen_server/otel.py
```diff
@@ -0,0 +1,194 @@
...
+        from opentelemetry.sdk.trace import TracerProvider
+        from opentelemetry.sdk.trace.export import BatchSpanProcessor
+    except ImportError:
+        logger.debug(
+            "OpenTelemetry SDK packages not installed — skipping "
+            "instrumentation.  Install the [otel] extra to enable."
+        )
+        return
+
+    service_name = os.environ.get("OTEL_SERVICE_NAME", "zenml-server")
```
stefannica: it's not a good idea to have all these environment variables hidden inside the code here. it's best to add them to the `ServerConfiguration` in `zenml.config.server.config` where they are centralized, automatically set from environment and discover-able  by users.

pyproject.toml
```diff
@@ -82,6 +83,12 @@ server = [
     "tldextract~=5.1.0",
     "itsdangerous~=2.2.0",
     "croniter>=6.0.0",
+    "opentelemetry-distro>=0.59b0",
```
stefannica, line 95: we already have OTEL dependencies in our client. Please check that they match version-wise.

stefannica (changes requested):
I really love where this PR is going. You have successfully replaced the entire custom API request tracing mess with centralized structured logging. Some things to consider:

* test that the request trace logs are still generated and have all the necessary info. The context vars you used need to travel from async-io coroutines down to worker threads and the connection you made through structlog might fail to cover everything.
* is there a way to hide structlogs from the actual code that emits logs ? I want it to be an implementation detail rather than something that the regular python modules need to be aware of and import directly. The key lies in reusing the existing centralized logging module zenml.logging.

docker/zenml-server-dev.Dockerfile
amitvikramraj: it's weird, i think it's my local formatter might have done. I'll undo these formatting changes

docker/zenml-server-dev.Dockerfile
amitvikramraj: I was trying to auto-intrument FastAPI using OTel but it didn't work because of the `--reload` flag. I'll remove it.

src/zenml/zen_server/auth.py
amitvikramraj: Thanks for this, it's a good idea. Earlier, I purposely didn't touch it for the same reason, since it is being used on the client side as well.

pyproject.toml
amitvikramraj: Could I get a link to where that dependency is? Is it the sdk dependency here at line 41 in toml file `"opentelemetry-sdk==1.38.0"` ?

src/zenml/zen_server/otel.py
amitvikramraj: addressed it!

src/zenml/zen_server/auth.py
amitvikramraj: addressed this

src/zenml/zen_server/auth.py
Json-Andriopoulos: You can't pass them this way sure, but you can use other structlog utility like bind_context_vars to do so. I imagine we wouldn't pass context vars this way either. A more sensible way to do it is:

```
with context_vars_set():  # -> Some context manager that sets context vars on the logger
    do_something()
    logger.info()
    try:
        do()
    except:
        logger.error()
    finally:
        logger.debug()
```

With this pattern all log calls pick up the context here, no reason to pass them explicitly.

Json-Andriopoulos: Here is how I would approach migration to structlog. This is up for discussion btw, don't want to enforce it as a pattern:

1) I would keep on using the same log statements, keep on using python's native logger.
2) I would use structlog to re-configure loggers (JSO formatter etc.)
3) I would introduce a few utilities (context manager, fastapi middleware, etc.) that help me introduce context vars on a few, controlled places that inject context for all subsequent log records (e.g trace-id on the beginning of any API call).

What that gives us: We get JSONified logs, we get context that is necessary for debugging (e.g. track all log statements for a particular request) and we do it with limited changes. I am a bit sceptic if I want to introduce changes across the codebase to comply with structlog's logger.

The difference boils down here:

```logger.info("hello from stdlib logging", ctx="something", another_ctx="something_else")```

I disagree that we should do it. We would need to change hundreds of log statements and the benefit is minimal. If you have context specific to one log statement then just add it in the message as you would today. If not, if the context is shared do it with context manager and all log statements within that block will have the context variables:

```
with bind_context(ctx_var="test"):
    logger.info("Hello")
    do()
    logger.warning("Oh oh")
```

amitvikramraj: To me, this feels quite a bit counterintuitive:

In the current structlog configuration:
* [1] We are still using the same logger statement. structlog exposes a wrapper class called [`BoundLogger`](https://www.structlog.org/en/stable/contextvars.html) that mirror the Python's stdlib logging API.

* The way I think about passing context information is that there are 2 types:
   * common context like request_id, trace_id, request method, request_path, etc. – we propagate them via middlewares instead of manually injecting them everywhere like we were doing earlier. For this we use structlog's `bind_contextvars()` method.
        ```py
        async def record_requests(request: Request, call_next: Any) -> Any:
            request_context = RequestContext(request=request)
            request_manager().current_request = request_context

            structlog.contextvars.clear_contextvars()
            structlog.contextvars.bind_contextvars(
                request_id=request_context.request_id,
                trace_id=request_context.trace_id,
                method=request.method,
                path=request.url.path,
                client_ip=(request.client.host if request.client else "unknown"),
            )
        ```


   * scoping context values to a specific operation. For this, structlog exposes a method `bound_contextvars()`, different from the `bind_contextvars()` which inject the context into all the logs:
        ```py
        with structlog.contextvars.bound_contextvars(project_id=str(project_id)):
            result = await service.do_work(project_id)
            logger.info("work.completed")

        # Inside the with block every log picks up project_id. Outside the block it's gone.
        ```
    * https://www.structlog.org/en/stable/contextvars.html

* Re: changing log statements
    * YES, I agree that we would have to change a couple of log statements, but TBH I feel this is a good thing because there was so much repetitive work happening in the way older logs were written. Using structlog logger makes it so much redundant and cleaner.

        ```py
        # Older log
        logger.debug(
            f"[{request_context.log_request_id}] ENDPOINT STATS - "
            f"{request_context.log_request} "
            f"async {func.__name__} STARTED "
            f"{get_system_metrics_log_str(request_context.request)}"
        )
        # Newer log – SO much simplerrr
        logger.debug(
            "endpoint.async.started",
            endpoint=func.__name__,
            **get_system_metrics(),
        )
        ```
* Re: passing unnecessary context to other logs.
    * The extra kwargs that we pass in a log statement using the structlog logger stay within that specific log event.
        ```py
        logger.info("hello from stuctlog logger", ctx="A", another_ctx="B")
        # → {"event": "something", "ctx": "A", "another_ctx": "B", request_id=..., method=..., ...}
        
        logger.info("another log statement")
        # → {"event": "next_event", request_id=..., method=...}   ← no ctx, no another_ctx
        ```

stefannica:
Closed in favor of: https://github.com/zenml-io/zenml/pull/4781

PR #4690, opened 2026-04-06
fix: replace bare `except:` with `except BaseException:` for explicit exception handling

harshadkhetpal:
## Summary

This PR replaces bare `except:` clauses with explicit `except BaseException:` in two files:

- `src/zenml/orchestrators/step_launcher.py` (line 536)
- `src/zenml/integrations/kaniko/image_builders/kaniko_image_builder.py` (line 294)

## Motivation

Bare `except:` clauses are discouraged by PEP 8 (E722) because they catch all exceptions including `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which can mask critical errors and make debugging difficult.

In both cases here, the pattern is `except: ... raise` (catch-all for cleanup, then re-raise), so `except BaseException:` preserves the exact same behavior while being explicit about intent.

## Changes

- **step_launcher.py**: `except:  # noqa: E722` → `except BaseException:` (also removes the now-unnecessary noqa comment)
- **kaniko_image_builder.py**: `except:` → `except BaseException:`

## Testing

These are safe, behavior-preserving changes — `except BaseException:` catches the same set of exceptions as bare `except:`. No functional change.

### DIFF
--- a/src/zenml/integrations/kaniko/image_builders/kaniko_image_builder.py
+++ b/src/zenml/integrations/kaniko/image_builders/kaniko_image_builder.py
@@ -291,7 +291,7 @@ class KanikoImageBuilder(BaseImageBuilder):
 
             try:
                 return_code = p.wait()
-            except:
+            except BaseException:
                 p.kill()
                 raise
 


schustmi:
> Bare except: clauses are discouraged by PEP 8 (E722) because they catch all exceptions including SystemExit, KeyboardInterrupt, and GeneratorExit, which can mask critical errors and make debugging difficult.

This is simply wrong. Your PR is not based on develop. The CLA is not signed.

PR #4704, opened 2026-04-08
Fix Skypilot Kubernetes orchestrator ignoring down and idle_minutes_to_autostop settings

XnetLoL:
## Summary

- The Kubernetes branch of `SkypilotBaseOrchestrator.submit_pipeline` hardcoded `down=False` and `idle_minutes_to_autostop=None`, silently ignoring whatever the user set in `SkypilotBaseOrchestratorSettings`.
- A follow-up guard (`launch_kwargs.pop("idle_minutes_to_autostop", None)`) reinforced the same silent ignore.
- The fix hoists both assignments above the `if isinstance(self.cloud, sky.clouds.Kubernetes)` branch and removes the pop, so Kubernetes behaves identically to other clouds.

**Why this is safe:**
SkyPilot on Kubernetes supports autodown (`down=True` with `idle_minutes_to_autostop`). The `AUTODOWN` feature is not in Kubernetes's `_CLOUD_UNSUPPORTED_FEATURES` list. Only `AUTOSTOP` (`down=False` with a non-None idle timeout) is unsupported — in that case SkyPilot will raise an error, which is the correct behaviour (fail loudly instead of silently doing nothing). The default settings (`down=True`, `idle_minutes_to_autostop=30`) work as expected.

**Breaking change:**
Previously, the docs listed `idle_minutes_to_autostop` as unsupported on Kubernetes and the orchestrator enforced this by ignoring user settings — clusters would stay up indefinitely after a run. With this fix, the default settings (`down=True`, `idle_minutes_to_autostop=30`) now take effect on Kubernetes too. Users who relied on clusters staying up must now explicitly set `down=False` and `idle_minutes_to_autostop=None` in their orchestrator settings to preserve the old behaviour.

## Pre-requisites

- [x] I have based my new branch on `develop` and the open PR is targeting `develop`. If your branch wasn't based on develop, read the [Contribution guide on rebasing branch to develop](https://github.com/zenml-io/zenml/blob/main/CONTRIBUTING.md#-pull-requests-rebase-your-branch-on-develop).

## Types of changes

## Test plan

- [ ] Existing SkyPilot integration tests pass in CI.
- [ ] Manually verified with a local k3s cluster + SkyPilot API server that the cluster is torn down after the idle timeout following a pipeline run.


safoinme (approved):
Thank you for this PR, looks good on my side

safoinme:
waiting for CI to pass and we merge @XnetLoL might be also worth updating branch

XnetLoL:
Done, rebased onto latest develop. @safoinme

bcdurak:
hey @XnetLoL, thank you for your contribution. The changes are merged and will be part of the next release.

PR #4716, opened 2026-04-09
Server-side replays

schustmi:
## Describe changes
This PR enables replaying pipeline runs from the server.

Key differences between replays and re-runs:
- Replays allow **skipping steps**. Skipping a step means it will not be re-executed, even if cache was disabled for the step or if some of the inputs changed in the meantime.
- Replays allow overriding **step input artifacts**. Compared to re-runs, replays not only allow overriding step parameters but even override inputs that came from an upstream step in the original run.

**Other required changes**:
- Update materializers for JSON serializable types to work from the server. This means they cannot use `fileio` anymore as that depends on the filesystem registry which is disabled on the server. Instead, they now use the artifact store directly.

## Pre-requisites

## Types of changes


stefannica (changes requested):
# Code review walkthrough session

## Metadata

| Field | Value |
|-------|--------|
| Repository | `/home/silent-assassin/aspyre/src/zenml` |
| Base ref | `origin/develop` (merge-base `1c6b4a0efcb8a288fb6e26116f37d0bea81a65d3`) |
| Head branch | `feature/server-side-replays` @ `8f3944b86ef77f7e4c8e84c46967e405d0cf73a1` |
| PR | [zenml-io/zenml#4716](https://github.com/zenml-io/zenml/pull/4716) — *Server-side replays* |
| Date | 2026-04-22 |
| Diff | 13 files, +403 / −28 (vs merge-base) |

---

## Stage 1 — Introduction

### Explainer summary

**What problem this solves**

The PR adds **server-side replay** of pipeline runs: the ZenML server can start
a new run from a **stored pipeline snapshot** with semantics that go beyond a
typical re-run. Replays can **skip steps** (without relying on the cache) and
**override step input artifacts**—including values that in the original run
came from upstream steps. A second commit makes **JSON-oriented
materialization** work in the server process by routing through the **artifact
store** instead of `fileio` / the filesystem registry (which is not available
on the server), and aligns JSON **external artifact** handling with that path.

**Size and shape**

- **Two commits:** *Server-side replays*; *Upload json artifacts to the
  artifact store*.
- **~403 lines added** across **13 files**; **no database migrations** in this
  diff.
- Two themes: (1) **public API + Zen store + HTTP + server execution** for
  replay; (2) **materializers and artifact utilities** for server-safe JSON
  I/O.

**Major paths (thematic, not line-by-line)**

| Area | Role |
|------|--------|
| `ReplayRunConfiguration` in `pipeline_run_configuration.py` | New config type extending `PipelineRunConfiguration` with `skip_successful_steps`, `steps_to_skip`, and `step_input_overrides`. |
| `ZenmlClient.replay_pipeline_run` in `client.py` | Resolves the run, optional YAML config, then `zen_store.replay_run`; optional `synchronous` wait. |
| `zen_store_interface.replay_run` + `rest_zen_store` + `sql_zen_store` | Store contract; REST client `POST` to run replay route; local SQL store implementation. |
| `constants.REPLAY` + `runs_endpoints.py` | Path segment `/replay` on runs; `POST` handler when workload manager is enabled, delegating to snapshot execution. |
| `zen_server/pipeline_execution/utils.py` | `run_snapshot` (and call chain) extended to pass **replay configuration** into the new run. |
| `orchestrators/step_runner.py` | Integrates replay behavior at step execution. |
| `materializers/*`, `artifacts/external_artifact.py`, `artifacts/utils.py` | Server-appropriate materialization; JSON artifact upload to the artifact store. |

**One-line outcome**

You can **replay** a prior run from the **server** with explicit **skips** and
**input overrides**, and the server can **materialize / persist** JSON-backed
values without a local client filesystem.

**Gates and preconditions (product-facing)**

- The source run must have a **pipeline snapshot**; the HTTP handler rejects
  replay if there is none.
- The `POST /runs/{run_id}/replay` route is registered only when
  `server_config().workload_manager_enabled` is true.
- **Authorization** checks include creating **pipeline runs** and **pipeline
  snapshots** for the project; an **entitlement** check is tied to the
  run-template / triggers feature (same as related server features in this
  area).
- `RestZenStore.replay_run` maps **HTTP 405** to a clear **RuntimeError** for
  servers that do not support replay; `ZenmlClient.replay_pipeline_run` is
  decorated to fail on a **local SQL Zen store** (server / REST path only).

### Session notes / questions

- *(Add during review.)*

### Decisions

- *(On finalize / stage handoff.)*

---

## Stage 2 — Public documentation pass (UX, product, and DevX)

**Pillar:** public surface only—HTTP, SDK, and Pydantic types users import. No
store internals.

### Explainer summary

This release adds a **server-side** way to start a new pipeline run from a
**stored snapshot** of an older run, with a dedicated configuration type
(`ReplayRunConfiguration`) for skip rules and per-step input overrides. The
**Python client** is the main ergonomic entry (`ZenmlClient.replay_pipeline_run`);
integrators can also call **`POST /api/v1/runs/{run_id}/replay`** with the same
body shape. The route and client path require a **capable server** (workload
manager on; snapshot present) and the right **permissions** and **entitlement**.
There is **no new CLI** in this diff. JSON-heavy execution on the server is
covered by materializer/artifact changes so behavior matches “runs on the
server,” not a local working tree.

### Breaking / compatibility

| # | Surface | What breaks or changes incompatibly | Affected who / what | Mitigation, migration, or version note |
|---|---------|----------------------------------------|----------------------|----------------------------------------|
| 1 | `BaseZenStore` / custom Zen store implementations | New **abstract** method `replay_run(...)` must be implemented. | Anyone maintaining a **custom** `ZenmlStore` subclass (rare) | Implement `replay_run` (e.g. delegate to REST) or stay on a version before the change. |
| 2 | Materializers / JSON on server | JSON built-in and base materializer paths **use the artifact store** and upload semantics where applicable instead of `fileio`-only behavior. | Code that **subclassed** materializers and **assumed** `fileio` in server context | Re-test; align with new loading/saving behavior. In practice server-side user code rarely hit the old path. |

**Default for app users:** new APIs and types are **additive**; existing runs and
client calls that never mentioned replay are unchanged. **No change** to
documented OpenAPI for older endpoints beyond the new route (when the server
feature is enabled).

*None identified* for typical SDK + server users beyond the rare custom-store /
advanced materializer cases above.

### User journey candidates

1. **Platform engineer** on a team using **ZenML Pro / server** **re-executes** a run that **failed halfway** with **`client.replay_pipeline_run`**, choosing either **default** behavior or options like **`skip_successful_steps`** (see **Journey 1**), optionally **`synchronous=True`**, **without** a laptop in the loop.

2. **Integrator** builds an internal tool that calls the **HTTP API** directly to **`POST /api/v1/runs/{run_id}/replay`** with a JSON body built from the same fields as `ReplayRunConfiguration` (e.g. `steps_to_skip`).

3. **Operator** points a **local client** at `zenml connect` to a **REST store**, tries replay, and learns whether the server **supports** replay (405 or clear error) vs is **fully** configured (snapshot, workload manager, entitlements).

4. **Data scientist** keeps using **local SQLite** Zen store for experiments: calling **`replay_pipeline_run`** **raises** `TypeError` (must be connected via the **API**), so they keep using **local** re-run or other flows.

N1 — A **privileged** caller replays a run and uses **`step_input_overrides`** to **replace upstream inputs** and **`steps_to_skip`** to **skip validation steps**, producing a “successful” run with **unreviewed** data in production metadata—**governance and trust** rest on **RBAC, entitlements, and who can create runs**.

N2 — Same **org** **upgrades the server** without the **workload manager** route: replay **disappears** (405 or “not supported”) and automation that assumed the new endpoint **fails** until the server is configured like **staging**.

### User journeys — expanded walkthroughs

#### Journey 1 — When you replay a run, what really happens? (end-user, SDK)

**What you are trying to do**

You already have a **run** in ZenML. Something went wrong, or you want a **new**
run that starts from the same **definition** of the pipeline, **executed on the
server**. You use **`client.replay_pipeline_run(...)`** and optionally pass
**options** to control which steps are **skipped** or which **inputs** you want
**different** the second time.

**What you need first (plain checklist)**

- The run you choose must have a **saved snapshot** of the pipeline, or replay
  is not available for that run.
- The client must be **connected to the ZenML server** (not a fully local-only
  store for this call).
- Your user must be allowed to **start runs** the same way you would for any
  other **server** pipeline, and the server product must have this feature
  **available** in your org.

**Smallest call (default replay, no custom options)**

```python
from zenml.config.pipeline_run_configuration import ReplayRunConfiguration

client = Client()
replayed = client.replay_pipeline_run("my-original-run-id-or-name")
```

If you do not pass a config, the client still builds a **default** (empty) replay
config behind the scenes. The important part for *behavior*: **this does
not** turn on *“auto-skip the steps that already went well last time”*.

**Idea 1: normal execution vs “skip successful steps”**

| Idea | In plain terms |
|------|----------------|
| **A — Default replay (no or empty “replay options”)** | The **new** run is planned like a **fresh** run: **every** step is scheduled to run **again**, unless your usual **caching** rules and a **cache match** make ZenML mark a step as **cached** (reusing a previous result **without re-running the code**). The **caching** rules you set on the run or the step (cache on vs off) apply here, like any other run. |
| **B — `skip_successful_steps=True` (you turn this on explicitly)** | Steps that **finished successfully in the *first* run** can be **left out** in the *new* run: ZenML can **reuse** the **outputs of that first run** for the new one for those steps. The **failed** step, or a step that never ran, is **not** “skipped that way” — you generally **re-run** what failed, which is the usual “fix the middle, continue” story. You also **override** a step’s **inputs** if you need new data; in that case ZenML will **not** use “skip for that step” the same way if it would have blocked the new data (the step is **re-run** so the overrides apply). |
| **C — `steps_to_skip=...` (names of steps to skip)** | You name **exact** steps in the new run that should **not execute**; use it when you want a **fully explicit** list, not only “whatever succeeded first time.” |
| **D — `step_input_overrides=...`** | You give a step **new inputs** for the new run, including when the input would otherwise come from an **upstream** step (see **Step input overrides** below for **UUIDs**, **values**, and **`ExternalArtifact`**). |

**Idea 2: “step cache” is not the same as “skip successful in replay”**

- **Cache** = “this step, with the **same** inputs/code fingerprint, was **run**
  before, so I can **reuse** that result.” If you turned **caching off** for a
  step, the **cache shortcut** does not apply, but the step is still a **normal
  runnable** step in the new run.
- **Replay: skip successful / by name** = a **separate** decision you can turn
  on, so a step does **not** re-run **because you asked replay to skip it
  (subject to the rules)**, and **reuses the first run’s** recorded
  inputs/outputs, **independent of** the cache.

**Example A — run failed in the *middle* (default replay, no `skip_…` options)**

*First run:*

| Order | Step | Result in the first run |
|-------|------|----------------------------|
| 1 | Load data | Succeeded |
| 2 | Validate | Succeeded |
| 3 | Train | **Failed** (this is the problem you are fixing) |
| 4 | Register model | **Never** ran in the first run |

*You call* `replay_pipeline_run("that-run")` *with no extra config.*

*Second run (intuitive, no implementation jargon):*

| Order | Step | What you will tend to see |
|-------|------|----------------------------|
| 1 | Load data | Treated as a **normal** step. It may **run** again, or, if your **caching** is on and ZenML **finds a match**, the step can show as **cached** and skip code. **Replay is not** auto-skipping it “because the first time succeeded” unless you set the **skip-successful** option. |
| 2 | Validate | Same as *Load data*. |
| 3 | Train | **Re-run** in the new run (or cached only if the **same** cache rules and match as usual). That is the step you are fixing. |
| 4 | Register model | **Runs** the first time in this *second* run (it was **not** a successful step in the first run, so the “that step already went well” style of skip would not **cover** it anyway). |

**Example B — you want the early steps to stay “done” without depending on
cache, and to focus on *Train***

Call with **`skip_successful_steps=True`** (names are whatever your pipeline
uses; this is the option’s meaning, not a copy-paste of step ids):

```python
replayed = client.replay_pipeline_run(
    "my-original-run",
    run_configuration=ReplayRunConfiguration(
        skip_successful_steps=True,
    ),
)
```

**Example C — only a named subset should be left out of the re-run**

```python
replayed = client.replay_pipeline_run(
    "my-original-run",
    run_configuration=ReplayRunConfiguration(
        steps_to_skip={"load_data", "validate"},
    ),
)
```

(Use the **real** invocation / step **names** from *your* pipeline project.)

**Step input overrides — what you are allowed to pass (SDK)**

You are **not** “rewiring” the pipeline on paper: the same step **names** and
**order** in the **snapshot** still hold. `step_input_overrides` means: “For
**this** step (by its **invocation** **name** in the pipeline) and **this** input
argument (the **name** in your `@step` function), use **this** data instead
of the value ZenML would normally resolve (including from an **upstream** step
in the first run).” If the step *would* have been **skipped** as
“already successful” but you give it **overrides** for that step, the step is
**re-run** so the new data is used (new inputs “win” over skip in that
situation).

**Shape in `ReplayRunConfiguration`:** a map **step invocation name** →
(**input** **name** as in your code → **value**). The outer keys and inner keys
must match your **actual** pipeline; the examples use placeholders like
`"train_model"` and `"dataset"`.

**What values are supported (families, not every edge case):**

1. **An artifact that already lives in the store (by id)** — **UUID** of an
   **artifact version** (a string in YAML/JSON, a `UUID` in Python). Use this
   when the file, table, or model is **already** stored. That includes
   reusing the **output** of a **step in the *previous* run***: you look up
   the **artifact version id** from the **old** run in the **UI** or with
   the **Client** and pass that id. There is no separate “type” for
   “came from another step”; it is the **same** as reusing any stored
   version.

2. **`ArtifactVersionResponse` from the Client** — the object returned e.g. by
   `Client().get_artifact_version(...)` when you already have the id, or from
   listing a step’s outputs. Same as (1) in the server; convenient in
   **Python** so you do not hand-carry raw ids in every place.

3. **A new in-memory value to upload** — For **`Client().replay_pipeline_run`**
   the config is sent to the server as **JSON** (see the table below). Use
   **JSON-friendly** values (**dict**, **list**, **str**, **int**,
   **bool**, **null**) or **UUIDs** of existing versions. **Rich** Python
   types (**pandas** `DataFrame`, **NumPy** arrays as objects, **Pydantic**
   models, arbitrary classes) usually **cannot** be sent that way; **upload**
   them **yourself** (e.g. `ExternalArtifact(...).upload_by_value()` or any
   flow that produces an **artifact version id**), then pass **only** the
   **UUIDs** in `step_input_overrides`. Other ZenML entry points (e.g. local
   `pipeline.replay()`) may **materialize** on the **client** before a
   **local** run—**not** the same as `replay_pipeline_run` over **REST**.
   Types that **cannot** be **materialized** for your stack still **fail** at
   run time.

**Examples (illustrative — use your own step and input names):**

```python
from uuid import UUID
from zenml.config.pipeline_run_configuration import ReplayRunConfiguration
from zenml.artifacts.external_artifact import ExternalArtifact
# from zenml.models import ArtifactVersionResponse  # if you use response objects

client = Client()
```

*Reuse an already stored version by id (including one produced in the
original run):*

```python
ReplayRunConfiguration(
    step_input_overrides={
        "train_model": {
            "dataset": UUID("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"),
        }
    }
)
```

*Point to a version you fetched in code (same effect as a bare id):*

```python
av = client.get_artifact_version("<artifact-version-uuid>")

ReplayRunConfiguration(
    step_input_overrides={
        "train_model": {"dataset": av},
    }
)
```

*Inject new config or small data (uploaded as a new artifact in the
store for this run):*

```python
ReplayRunConfiguration(
    step_input_overrides={
        "train_model": {
            "hyperparams": {"epochs": 50, "lr": 0.01},
        }
    }
)
```

*Example with NumPy / `ExternalArtifact` (see table: **REST** **replay**
prefers **JSON-safe** values or **UUIDs**; a raw `ndarray` may not **serialize**
to the replay **JSON** **body**—use a **list**, **pre-upload** + **UUID**, or
another entry point):*

```python
import numpy as np

ReplayRunConfiguration(
    step_input_overrides={
        "preprocess": {
            "x": [1, 2, 3],  # JSON-friendly; or pass a UUID of a pre-uploaded artifact
        }
    }
)

# Pre-upload then pass id, or use ExternalArtifact in a client-only flow:
# artifact_id = ExternalArtifact(value=np.array([1,2,3])).upload_by_value()
# then step_input_overrides={"preprocess": {"x": artifact_id}}
```

*Putting it on `replay_pipeline_run`:*

```python
from zenml.config.pipeline_run_configuration import ReplayRunConfiguration

client = Client()
client.replay_pipeline_run(
    "my-original-run",
    run_configuration=ReplayRunConfiguration(
        step_input_overrides={...}  # one of the shapes above
    ),
)
```

**Materialization and the REST replay request (`replay_pipeline_run`)**

`Client().replay_pipeline_run` calls `zen_store.replay_run`, which **POST**s
`ReplayRunConfiguration` using a **JSON**-safe dump of the model
(`model_dump(mode="json")`). **It does not** run a **client-side** “upload
**non-JSONable** values to the artifact **store** first, then replace them
with ids” step. For **DataFrames**, **ndarrays** (as objects), **custom**
classes, **etc.**, **you** are expected to **create** **artifact** **versions**
first (e.g. `ExternalArtifact(value=...).upload_by_value()` on the **client**,
or reuse an **existing** version id) and put **only** the **UUIDs** in
`step_input_overrides`. **That** is different from `Pipeline.replay()` in
**local** code, which can **upload** on the **client** as part of its own
**replay** path—**not** the same as `replay_pipeline_run` over the **API**.

The table below still applies: **JSON-like** **inline** values are **only**
**materialized** on the **server** **after** a successful **POST**; **UUIDs**
are not re-uploaded in that request.

| What you pass in `step_input_overrides` | What goes over the network | Where new artifacts are created for that override |
|----------------------------------------|----------------------------|--------------------------------------------------|
| **Only artifact version UUIDs** (and no other “new” values) | The **JSON** body lists those **ids**; nothing to **upload** in the replay handler for those fields. | **Nowhere** in this call—the versions **already** exist in the **artifact** **store**. |
| **JSON-like values** (e.g. **dict**, **list**, **str**, **int**, **bool**, **null**) | The **full** values appear in the **JSON** **body** of `POST .../runs/{id}/replay`. | **On the server** after the request: values are **materialized** and **written** to the **artifact** **store** for the **snapshot’s** **stack** (same idea as other server snapshot runs that need the store). |
| **Non-JSON-friendly objects** (e.g. **pandas** `DataFrame`, **NumPy** `ndarray` as an object, **Pydantic** model instances, custom classes) | They **do not** travel as live objects in the **REST** replay request; the **client** usually **cannot** build a valid **JSON** **body** for them, or the request is **rejected** before the server runs. | **Not** on the **server** in that request. You **materialize** **client-side** (or in another process), obtain **artifact version** **UUIDs**, and put **only** those **UUIDs** in `step_input_overrides` for `replay_pipeline_run`. |

**In one sentence:** for **server** **replay** via the **SDK**, **JSON-safe**
**inline** values and **raw** **UUIDs** are what the wire supports as-is;
**everything else** should become **UUIDs** **before** the call, or you use a
**different** ZenML path that is **not** the single **REST** **replay** round
trip.

A **raw** `POST` to the same route (no Python SDK) follows the same rule: the
**body** is **JSON**, so the middle row matches **JSON** **values** and the
**first** and **third** rows match **strings**-as-**UUID** and
**“pass ids only for rich data”** **respectively**.

**Config fields, one line each**

| Field | In end-user terms |
|-------|-------------------|
| `skip_successful_steps` | “**Skip** the work for steps that **succeeded in the *first* run** and **reuse** their first-run results in the *new* run, where the rules allow.” **Does *not* apply to** a step that **failed** or was **not run** in the first run, in the sense we care about. |
| `steps_to_skip` | “**Never re-run** these **named** steps, period (by name in your project).” |
| `step_input_overrides` | “For a given **step** and **input** name, use **this** data in the new run” (by **UUID**, **Client** artifact object, or **value** to **upload**—see the **Step input overrides** subsection above). **Overrides** and **“skip that step”** interact: new inputs mean the step **executes** so the new data is used. |
| Other run settings (name, **cache** for the run, per-step **cache**, …) | Same as always; they work **on top of** the replay list above. **Cache** decides whether ZenML can **reuse** a past result for a step when caching is **on** and a **match** exists; **skip successful** is a **separate** “**reuse the first run’s** step result for the new run” path you **opt in** to. |

**If something is wrong, what it feels like (no class names required)**

| Situation | What you can tell the user to expect |
|----------|--------------------------------------|
| **Success** | A **new** run in the list with a **new** id, same as any other new run. |
| **“No snapshot for this run”** | The old run is not a valid source for this feature; you need a run that **captured a snapshot** when it was created, or a different way to re-run. |
| **“Not when connected to the local store only”** (message about **server** / **API**) | Replay this way needs the **server**-connected client; for **local** experiments, use the **other** flows you use today. |
| **“Replay not supported”** or **old server** | The **environment** (version, admin settings) is not the same as one where you already saw it work. Align with your **staging** or **doc**-listed requirements, or **re-run** the pipeline the older way. |
| **Not allowed to start a run** | The same **access** and **governance** you already use for other **sensitive** actions on the server. |

**N1 and N2 (in one breath)**

- **N1** — Skips and **input changes** are **strong**; **governance** is the same
  as for *who may start a run* and *who may use* server features—not an extra
  in-product “are you sure?” for each flag.
- **N2** — Pipelines in **automation** should not assume the replay **endpoint**
  exists in every **environment** until the **ops** person confirms the **same
  server** setup you tested.

You can still add a `config_path=...` YAML, or a dict, instead of
`ReplayRunConfiguration(...)`; **do not** set both an in-file and an in-code
**full** config in one call, or the client will **error** with a clear
**“pick one”** message.

**HTTP (for other tools):** `POST /api/v1/runs/{run_id}/replay` with an
optional **JSON** body: same field names and shapes as the Python replay
config, or an empty body to mean **defaults** as in **Example A** above. Normal
**auth** for your `v1` run APIs.

### Session notes / questions

- **Q:** For an input that normally comes from a **previous** step, does the
  **step input override** or the **upstream** artifact win? **A:** The **override**
  wins for that **input** **name** (it replaces the resolved artifact). A
  **wrong** name does not change the real slot.

- **Q:** With **`skip_successful_steps=True`**, if a step **succeeded** in the
  first run but this replay has **`step_input_overrides`** for that step, is it
  **skipped** or **re-run**? **A:** **Re-run** with the new inputs; **overrides**
  take precedence over **skip** for that step (log may note the step was not
  skipped because overrides are present).

- **Q (HTTP / REST, not SDK):** What has to work **server-side** when
  `step_input_overrides` contains **new** values (not only artifact **UUIDs**)?
  **A:** Those values are **materialized** and **written** in the **server**
  process using the **artifact** **store** on the **snapshot’s** **build**
  **stack** (the same stack path used for other server snapshot execution, not
  a separate “original run only” store). Materialization uses the **normal**
  **default** **materializers** for the **types** the API can express (for a
  **JSON** body, that is mostly **JSON-like** values; **UUID-only** overrides
  skip the upload path). The **stack** must already be **runnable** from the
  server (existing validation); if the server **cannot** **read/write** the
  artifact **store** as configured, **upload** **fails** and replay does not
  start—this is **broader** than replay (any server run that needs the store)
  but **inline** **values** in overrides add a **write** during the **replay**
  **request**.

- **Improvement proposition (not implemented today):** **`replay_pipeline_run`**
  could **materialize** any **JSON-serializable** (or otherwise client-handled)
  value in `step_input_overrides` on the **client** (upload to the **artifact**
  **store** using the **client’s** stack / credentials) and send **only**
  **artifact** **version** **UUIDs** in the **REST** **body**, instead of
  embedding **raw** **dicts** / **scalars** for the **server** to **upload**
  in `_maybe_upload_step_input_overrides`. **Rationale:** when **only** the
  **client** may **reach** the **artifact** **store** (or **fully**
  **authenticate** to it) and the **server** is **not** **trusted** or **able**
  to **write** there, **server-side** **inline** **uploads** are a **hard**
  **requirement** that some **deployments** **cannot** meet. **Wire** **format**
  would stay **ids-only** for overrides, shrinking **server** **privilege** to
  “accept **UUIDs**” for those fields. **Trade-offs:** **client** must have
  **write** access; **duplication** of **materialization** **logic** next to
  `pipeline.replay()`’s `_prepare_step_input_overrides` **pattern**; need a
  **clear** **contract** for **HTTP** **callers** who have **no** **Python**
  (they still **pre-upload** and pass **ids**).

- **Action items:** **Docs (current):** document that **inline** **JSON** in
  `step_input_overrides` implies **server**-side **artifact** **store** **write**
  on **replay** (see **Journey 1** table). **Product / SDK (proposed):** evaluate
  **client-side** **pre-upload** for **JSON-like** **override** **values** in
  `replay_pipeline_run` so the **server** **only** receives **UUIDs**—per
  **improvement** **proposition** above. **No** code change in **this** review
  **session**.

### Decisions

- *(On finalize.)*

---

## Stage 3 — Design / architecture

### Explainer summary

This stage maps Stage 2 user journeys to the code path: **REST** → **Zen
store** → **`run_snapshot`** and snapshot materialization → orchestration
( **`skip_successful_steps`**, **`step_input_overrides`**, cache). This PR
does not add Alembic migrations; **`original_run_id`** is an existing field
used for replay lineage when `run_snapshot` receives `original_run`.

### 1. Design intent and scope

- **In scope:** Server-side replay from a hydrated run snapshot;
  `ReplayRunConfiguration` merged into the new snapshot’s
  `PipelineConfiguration`; `RestZenStore.replay_run` (HTTP POST);
  `SqlZenStore.replay_run` unimplemented (local store only);
  `StepRunRequestFactory.should_skip_step` and `resolve_step_inputs` for
  skip/override semantics; materializer and artifact helper changes for
  server I/O.
- **Out of scope in this diff:** Dashboard / Cloud UI “Re-run” flows that
  start from template or snapshot forms (separate entry points; see Stage 2
  session notes).

### 2. Database

- **This PR:** No new tables or columns in the diff.
- **Lineage:** `PipelineRun.original_run_id` (FK) is set when
  `run_snapshot` passes `original_run` into `create_placeholder_run` via
  `original_run_id=original_run.id`. Schema/ORM for that field predates this
  feature branch.

### 3. Domain model and REST

- `ReplayRunConfiguration` (e.g. `pipeline_run_configuration.py`) extends
  `PipelineRunConfiguration` with `skip_successful_steps`, `steps_to_skip`,
  and `step_input_overrides` (values may be inline material values or UUIDs
  of existing artifact versions; non-UUID values are uploaded server-side—see
  §5).
- **Route:** `POST .../runs/{run_id}/replay` in `runs_endpoints.py`, only
  registered when `server_config().workload_manager_enabled`.
- **Handler** loads the run (hydrated), requires `run.snapshot`, checks
  permissions for creating snapshot and run, `check_entitlement` for run
  template / triggers, then calls `run_snapshot` with
  `replay_configuration` and `original_run`.

```626:631:src/zenml/zen_server/routers/runs_endpoints.py
        return run_snapshot(
            snapshot=run.snapshot,
            auth_context=auth_context,
            replay_configuration=run_configuration,
            original_run=run,
        )
```

### 4. Zen store contract and implementations

- **Interface** (`zen_store_interface.py`): abstract
  `replay_run(run_id, run_configuration: ReplayRunConfiguration) -> PipelineRunResponse`.
- **RestZenStore** posts to the replay URL with the configuration body; 405
  is surfaced as a `RuntimeError` if the server build does not expose replay.
- **SqlZenStore** always raises: replay is a server workload path, not
  available through the local SQL store (`Client.replay_pipeline_run` is
  blocked for SQL store users).

```5134:5150:src/zenml/zen_stores/sql_zen_store.py
    def replay_run(
        self,
        run_id: UUID,
        run_configuration: ReplayRunConfiguration,
    ) -> NoReturn:
        """Replay a pipeline run.

        Args:
            run_id: The ID of the pipeline run to replay.
            run_configuration: Replay configuration.

        Raises:
            NotImplementedError: Always.
        """
        raise NotImplementedError(
            "Replaying a pipeline run is not possible with a local store."
        )
```

### 5. Server pipeline execution: snapshot creation and replay

- **`run_snapshot`** merges `replay_configuration` into
  `run_configuration` (or takes configuration from a snapshot run request),
  validates via `validate_snapshot_for_server_execution`, and by default
  calls `create_snapshot_from_source` to fork a new snapshot. It then
  `create_placeholder_run` with `original_run_id` when `original_run` is set,
  and builds the runner workload. **`is_replay`** and **`WorkloadType`** use
  `replay_configuration is not None`: an HTTP call with an **omitted** body
  gives `None`, so the run is not treated as a replay for workload
  keying; the SDK path typically sends a `ReplayRunConfiguration` instance
  (possibly empty with defaults) so `is_replay` is true.

```276:313:src/zenml/zen_server/pipeline_execution/utils.py
    if replay_configuration and not original_run:
        raise ValueError("Original run is required to replay a pipeline run.")

    run_configuration: Optional[PipelineRunConfiguration] = (
        replay_configuration
        or (request.run_configuration if request else None)
    )
    ...
    placeholder_run = create_placeholder_run(
        snapshot=target_snapshot,
        trigger_info=trigger_info,
        logs=LogsRequest(source=LOGS_RUNNER_SOURCE),
        original_run_id=original_run.id if original_run else None,
    )
```

```342:347:src/zenml/zen_server/pipeline_execution/utils.py
    is_replay = replay_configuration is not None
    workload_id = (
        placeholder_run.id if trigger_id or is_replay else target_snapshot.id
    )
    workload_type = (
        WorkloadType.RUN if trigger_id or is_replay else WorkloadType.SNAPSHOT
    )
```

- **`create_snapshot_from_source`** for `ReplayRunConfiguration` first runs
  **`_maybe_upload_step_input_overrides`**: if any override is not a UUID, it
  loads the stack’s artifact store and uses `ExternalArtifact(value=...).upload_by_value` to persist values and replace them with version IDs. Then
  **`snapshot_request_from_source_snapshot`** `model_dump`s
  `PipelineConfiguration` fields (including overrides as UUIDs) and merges
  them onto the source snapshot’s `pipeline_configuration`.

```169:180:src/zenml/zen_server/pipeline_execution/utils.py
    if isinstance(run_configuration, ReplayRunConfiguration):
        run_configuration = _maybe_upload_step_input_overrides(
            run_configuration=run_configuration,
            stack=stack,
        )

    snapshot_request = snapshot_request_from_source_snapshot(
        source_snapshot=snapshot,
        config=run_configuration or PipelineRunConfiguration(),
        template_id=template_id,
    )
```

```200:235:src/zenml/zen_server/pipeline_execution/utils.py
    if not run_configuration.step_input_overrides:
        return run_configuration

    if all(
        isinstance(value, UUID)
        for overrides in run_configuration.step_input_overrides.values()
        for value in overrides.values()
    ):
        return run_configuration

    artifact_store = load_artifact_store(
        stack.components[StackComponentType.ARTIFACT_STORE][0].id,
        zen_store=zen_store(),
    )
    ...
            else:
                artifact_version_id = ExternalArtifact(
                    value=value
                ).upload_by_value(artifact_store=artifact_store)

            override_ids[invocation_id][input_name] = artifact_version_id

    return run_configuration.model_copy(
        update={"step_input_overrides": override_ids}
    )
```

```633:646:src/zenml/zen_server/pipeline_execution/utils.py
    pipeline_update = config.model_dump(
        include=set(PipelineConfiguration.model_fields),
        exclude=pipeline_update_exclude,
        exclude_unset=True,
        exclude_none=True,
    )
    if pipeline_secrets := pipeline_update.get("secrets", []):
        pipeline_update["secrets"] = [
            zen_store().get_secret_by_name_or_id(secret).id
            for secret in pipeline_secrets
        ]
    pipeline_configuration = pydantic_utils.update_model(
        source_snapshot.pipeline_configuration, pipeline_update
    )
```

### 6. Orchestration: skip vs cache vs override

- **`StepRunRequestFactory.should_skip_step`** (`step_run_utils.py`): apply
  explicit `steps_to_skip` first; if `skip_successful_steps` is false, do
  not auto-skip; if there are `step_input_overrides` for the invocation, log
  a warning and **run** the step; otherwise skip when the original run’s
  step succeeded.

```103:139:src/zenml/orchestrators/step_run_utils.py
    def should_skip_step(self, invocation_id: str) -> bool:
        """Check whether a step should be skipped.

        Args:
            invocation_id: The invocation ID to check.

        Returns:
            Whether the step should be skipped.
        """
        if invocation_id in self.snapshot.pipeline_configuration.steps_to_skip:
            return True

        if not self.snapshot.pipeline_configuration.skip_successful_steps:
            return False

        if self._get_input_overrides(invocation_id):
            logger.warning(
                "Step `%s` should be skipped, but there are input overrides "
                "configured. The step will be executed.",
                invocation_id,
            )
            return False

        try:
            original_step_run = self._get_original_step_run(invocation_id)
        except Exception as e:
            logger.warning(
                "Failed to fetch original step run `%s` while trying to skip "
                "successful steps: %s. The step will be executed.",
                invocation_id,
                str(e),
            )
            return False

        return bool(
            original_step_run and original_step_run.status.is_successful
        )
```

- **`resolve_step_inputs`** applies `input_overrides` after other input
  resolution, loading each ID as a full `StepRunInputResponse` with
  `input_type=OVERRIDE` (so overrides **win** over upstream wiring for those
  names).

```196:203:src/zenml/orchestrators/input_utils.py
    if input_overrides:
        for name, id_ in input_overrides.items():
            input_artifacts[name] = [
                StepRunInputResponse(
                    input_type=StepRunInputArtifactType.OVERRIDE,
                    **Client().get_artifact_version(id_).model_dump(),
                )
            ]
```

### 7. Materializers and artifacts (this PR)

- `base_materializer`, `built_in_materializer`, `external_artifact`,
  `artifacts/utils`, and `step_runner` changes align **saving/loading
  through the artifact store** with server execution; details follow the
  file-level diff (same theme as override upload: server must read/write
  bytes consistently with the client path).

### 8. End-to-end flow (condensed)

```mermaid
sequenceDiagram
  participant C as Client / RestZenStore
  participant R as runs_endpoints.replay_run
  participant S as run_snapshot
  participant U as _maybe_upload
  participant Z as create_snapshot (zen_store)
  participant W as _build_and_run
  C->>R: POST /api/v1/runs/{id}/replay
  R->>S: run_snapshot(replay_configuration, original_run)
  S->>U: if ReplayRunConfiguration
  S->>Z: create_snapshot_from_source
  S->>S: create_placeholder_run(original_run_id)
  S->>W: runner workload
```

### Stage 3 — Session notes / follow-ups

- **Empty body vs SDK default:** `is_replay = replay_configuration is not None`
  treats a missing JSON body as “not a replay” for `workload_id` and
  `WorkloadType`, which can diverge from the user’s mental model. Worth
  aligning with OpenAPI and all clients if replay is a strict product
  invariant.
- **Override upload location:** Stage 2 improvement—client pre-uploads all
  override material so the POST carries only UUIDs, avoiding
  `upload_by_value` on the server in locked-down stores.

### Decisions

- *(On finalize.)*

---

## Stage 4 — Implementation

### Explainer summary

This stage is a **code review** pass over the **actual** diff (13 files,
+403/−28 vs merge-base `1c6b4a0e…`), using Stages 2–3 and session notes as the
**expected behavior** bar. **Finding:** orchestration files such as
`step_run_utils.py` / `input_utils.py` are **not** in this PR—**skip/override**
semantics were **already** on `develop`; this branch wires **API + store +
`run_snapshot` + upload path + JSON materialization**.

### 1. What the diff does (mechanical)

| Area | Change |
|------|--------|
| `ReplayRunConfiguration` | New Pydantic model with `skip_successful_steps`, `steps_to_skip`, `step_input_overrides`. |
| `Client.replay_pipeline_run` | Resolve run, build config, `zen_store.replay_run`, optional wait. |
| `ZenStoreInterface` + `RestZenStore` + `SqlZenStore` | Abstract `replay_run`; REST `POST …/runs/{id}/replay` with `model_dump(mode="json")`; local `NotImplementedError`. |
| `runs_endpoints` | Conditional route + permissions + `check_entitlement` + `run_snapshot(…, replay_configuration, original_run)`. |
| `pipeline_execution/utils` | `run_snapshot` optional `request`, replay branch, `original_run_id`; `_maybe_upload_step_input_overrides` in `create_snapshot_from_source`. |
| Artifacts / materializers | `upload_by_value` / `save_artifact` accept explicit `artifact_store`; JSON via `artifact_store.open`; `BaseMaterializer` size via store; `step_runner` passes stack’s store. |

### 2. Strengths

- **Clear layering:** HTTP → `run_snapshot` → `create_snapshot_from_source` →
  optional upload → same snapshot-creation path as other server runs.
- **REST body serialization** uses `body.model_dump(mode="json")` in
  `RestZenStore.post`, which matches the report’s **JSON-only** wire story and
  forces UUIDs to strings for transport.
- **Explicit artifact store** for server upload path avoids “active stack on
  server” ambiguity for `_maybe_upload_step_input_overrides` and aligns
  `ExternalArtifact.upload_by_value` with the **stack** loaded for the
  snapshot.
- **Guard:** `replay_configuration` without `original_run` raises
  `ValueError` early.
- **Smallest diff** on `step_runner`: one argument threads the stack’s store
  into `_store_artifact_data_and_prepare_request`, consistent with the
  artifact utils change.

### 3. Issues and risks (reviewer view)

**3.1 — `is_replay` vs empty HTTP body (Stage 3 follow-up, **medium**)**  
`is_replay = replay_configuration is not None` is **false** when a client sends
`POST …/replay` with **no** body, even though `original_run` is set and the
**user intent** is still “replay.” Effect: `workload_id` / `WorkloadType` use
the **non-replay** branch (snapshot id, `WorkloadType.SNAPSHOT`), while
`original_run_id` is still set on the placeholder run. If downstream code or
**observability** assumes `WorkloadType.RUN` for *all* replays, **HTTP and SDK**
behave **differently**. **Suggestion:** treat **“replay endpoint was used”**
as replay for workload purposes (e.g. `is_replay = original_run is not None` or
a dedicated flag from the router), or **document** that an empty body is not
equivalent to the SDK and may change workload metadata.

**3.2 — `RuntimeError` for missing snapshot (`runs_endpoints`, **low–medium**)**  
`if not run.snapshot: raise RuntimeError(...)` may surface as a **5xx** unless
the FastAPI layer maps it. Prefer **`HTTPException(400 or 422, …)`** with a
stable error code for API clients (Stage 2 “what it feels like” table).

**3.3 — Entitlement name vs feature (`check_entitlement`, **low / product**)**  
Replay reuses `RUN_TEMPLATE_TRIGGERS_FEATURE_NAME` like template runs. **OK** for
consistency with `pipeline_snapshot` run path, but the **name** may confuse
admins (“triggers” vs “replay”). Worth a **product note** in ops docs so
on-call knows which switch gates replay.

**3.4 — Server-side override upload (Stage 2 improvement proposition, **context-dependent**)**  
`_maybe_upload_step_input_overrides` **writes** to the stack’s artifact store
during **request handling**. That matches the report’s **governance** story
(server must be able to write), but **fails** deployments where only clients
may write (reported improvement: **client pre-upload**, **UUID-only** body).
Not a bug in the PR’s stated model; **flag** for **enterprise** hardening.

**3.5 — `BaseMaterializer._get_storage_size` (**low**)**  
Recursive directory walk via `listdir` + `os.path.join` replaces `fileio.size`
for metadata. **Risk:** very large trees or **unusual** store semantics
(symlinks, slow `listdir`) could make metadata extraction costly or
non-terminating in pathological cases. Unlikely for typical step outputs;
consider a **depth cap** or **fallback** if this shows up in profiling.

**3.6 — `built_in_materializer` JSON helpers (**low**)**  
`_read_json` / `_write_json` use `artifact_store.open` without the previous
`yaml_utils` code path. **Check:** `ensure_ascii` and encoding behavior match
**legacy** artifacts; add a **round-trip** or **regression** test for old runs
if not already present.

**3.7 — Abstract `replay_run` (**by design, Stage 2**)**  
Custom `ZenStore` implementations must implement `replay_run` or break. The PR
is consistent with the table in Stage 2; no code issue—call out in **release
notes**.

### 4. Suggestions (concrete, optional)

1. **Unify workload replay semantics** for `POST` with/without body (see 3.1)
   or document OpenAPI as **“empty body = snapshot-style workload”**.
2. **Map validation failures** to 4xx with structured errors in `replay_run`.
3. **Tests:** unit or integration for **(a)** replay route with/without
   `ReplayRunConfiguration` body, **(b)** `_maybe_upload` with UUID-only vs
   inline values, **(c)** `RestZenStore.replay_run` 405 → `RuntimeError` message.
4. **Docs:** one paragraph that **inline** `step_input_overrides` values imply
   **server** artifact store **writes** (Stage 2 action item).
5. **Future:** optional **client-only UUID** path for overrides (Stage 2
   proposition) to shrink server trust boundary.

### 5. Session notes (Stage 4)

- **Report alignment:** Stages 2–3 described orchestrator **behavior**; this
  PR does **not** change `should_skip_step` / `resolve_step_inputs` (verify on
  `develop` if diffs are compared in review tools).
- **Decisions for authors:** (1) product stance on `is_replay` + empty body;
   (2) HTTP status for “no snapshot”; (3) test matrix for new paths.

### Decisions

- *(On finalize.)*

stefannica (changes requested):
# Summary


### Code review — follow-ups (recommended)

1. **`is_replay` vs empty HTTP body** — Today,
   `is_replay = (replay_configuration is not None)`. A **POST with no body**
   can still be a user-intent replay but gets **workload** metadata
   (`workload_id` / `WorkloadType`) **differing** from the **SDK** path, while
   `original_run_id` is still set. **Decide:** treat `original_run is not None`
   (or a router flag) as replay for workload purposes, **or** document in
   **OpenAPI** that empty body is not equivalent to the SDK.
2. **Error mapping** — `RuntimeError` for “no snapshot” may present as **5xx**;
   prefer **4xx** with a **stable** error code for API clients.
3. **Tests** — Add coverage for: replay **with/without** JSON body;
   `_maybe_upload` (UUID-only vs inline values); **405** → user-facing error
   on `RestZenStore.replay_run`.
4. **Docs (short)** — One explicit paragraph: inline override **values** ⇒
   **server-side artifact store writes** during replay; UUID-only overrides
   avoid that write path.
5. **Materializers** — Low risk: confirm **legacy** JSON artifacts still
   round-trip after switching to `artifact_store.open`; add regression tests if
   missing. **Optional:** guard **huge** directories in `_get_storage_size` if
   profiling shows pain.

### Optional product / technical debt (not blocking merge by default)

- **Client-only override upload** (Stage 2): materialize JSON-like overrides on
  the **client** and send **only artifact version UUIDs** so the server need not
  write to the store on replay (enterprise / locked-down store scenarios).
- **Release notes:** call out **custom Zen store** `replay_run` requirement and
  **materializer** behavior for server context.

stefannica (approved):

PR #4729, opened 2026-04-15
Remove the unique constraint from the `LogsSchema`

bcdurak:
## Describe changes
The logs schema had a unique log key. But in cases like restarts, the same pipeline run gets used which already has a logs entry with the same log key (for instance "orchestrator" in this case). We want to enable the creation of log entries for repeated executions, thus we are removing the unique constraint. 

## Pre-requisites

## Types of changes


schustmi (approved):

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -458,76 +461,109 @@ def run_logs(
...
+        elif logs_id:
+            logs = search_logs_by_id(run.log_collection, logs_id)
+            if logs is None:
+                raise KeyError(
+                    f"No logs found for ID '{logs_id}' in run {run_id}"
+                )
+        else:
+            raise ValueError("Either source or logs_id must be provided.")
+    else:
+        raise KeyError(f"No logs found for run {run_id}.")
```
schustmi: If a run has no log collection, there might still be runner logs. this error would prevent that I think

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -458,76 +461,109 @@ def run_logs(
...
-        )
-        is_legacy_run_with_runner_logs = (
-            snapshot.template_id or snapshot.source_snapshot_id or run.trigger
-        )
+    logs: Optional["LogsResponse"] = None
+    if run.log_collection:
+        if source:
+            logs = search_logs_by_source(run.log_collection, source)
+            if logs is None:
+                raise KeyError(
```
schustmi, line 495: If the source is `runner`, this will fail even though it's handled later.

I'd suggest we keep the whole logic that existed before maybe, and just handle the ID case in front?

Seems very easy:
```
if logs_id:
  logs = search_logs_by_source(run.log_collection or [], source)
else:
  # previous logic
```

schustmi (changes requested):

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: Putting the ID in the front was my first instinct as well. The only reason why I have jumped through so many hoops for the case that people would provide an ID that points to one of the new logs responses with source == runner. This should be good now.

src/zenml/zen_server/routers/runs_endpoints.py
bcdurak: You are right about the error. I have removed it now.

schustmi (approved):

schustmi (approved):

PR #4736, opened 2026-04-17
Fix typo in step creation error message

shaun0927:
## Describe changes

Corrects two small issues in the step-creation error path in `SqlZenStore._update_step_run_status_upgrading_if_necessary` (`src/zenml/zen_stores/sql_zen_store.py:10915`):

1. `Cannot creat step` → `Cannot create step` (typo).
2. A missing space before the execution-mode interpolation caused the rendered message to read `...execution mode is{mode}.` instead of `...execution mode is {mode}.`.

The neighbouring message at line 10904 already uses the correct spelling (`Cannot create step '{...}' for pipeline in ...`), so the existing working path stays aligned with this one.

## Pre-requisites

- [ ] Documentation is not affected (message-level change only).

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -10912,9 +10912,9 @@ class SqlZenStore(BaseZenStore):
 
                 if execution_mode != ExecutionMode.CONTINUE_ON_FAILURE:
                     raise IllegalOperationError(
-                        f"Cannot creat step '{step_run.name}' for the run '{run.name}' "
-                        "because the run is in a FAILED state and the execution mode is"
-                        f"{execution_mode}."
+                        f"Cannot create step '{step_run.name}' for the run '{run.name}' "
+                        "because the run is in a FAILED state and the execution "
+                        f"mode is {execution_mode}."
                     )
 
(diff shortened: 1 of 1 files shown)


schustmi (approved):

JiwaniZakir (commented):
The typo fix in `sql_zen_store.py` is correct — `creat` → `create` in the `create_run_step` method. The message reformatting is also an improvement, though worth noting the original had a subtle bug beyond just the typo: the third line was a plain string literal (not an f-string) immediately followed by an f-string on the next line, meaning there was no space between `"is"` and `{execution_mode}` in the concatenated result. The new version correctly separates `"execution "` and `f"mode is {execution_mode}."` across the two strings, which reads more naturally.

One remaining nit: the error message says the run `is in a FAILED state` but doesn't mention the *current* execution mode value in a way that helps users act on it — consider whether surfacing `execution_mode` by name is sufficient or whether a hint like `"use ExecutionMode.CONTINUE_ON_FAILURE to allow steps to proceed after failure"` would be more actionable. That said, this is out of scope for a typo fix and can be addressed separately.

bcdurak:
Merged the PR now. @shaun0927 thank you for your contribution.

PR #4737, opened 2026-04-17
Avoid KeyError in template config schema when step has parameter_spec without config parameters

shaun0927:
## Describe changes

`generate_config_schema` in `src/zenml/zen_stores/template_utils.py` crashes with `KeyError('{step_name}_parameters')` when a step carries a `parameter_spec` but has no matching entries in `config.parameters`.

### Where it breaks

The substitution loop unconditionally pops the step's parameter schema `$def`:

```python
# src/zenml/zen_stores/template_utils.py:408-409
# Remove the original parameter schema #def that we just replaced
root_definitions.pop(f"{step_name}_parameters")
```

That entry is only added a few hundred lines above, inside a guarded block that only runs when `step.config.parameters` is truthy:

```python
# src/zenml/zen_stores/template_utils.py:242-260
if step.config.parameters:
    ...
    parameters_class = create_model(
        f"{step_name}_parameters", **parameter_fields
    )
```

The outer loop, however, only checks `step.spec.parameter_spec`:

```python
# src/zenml/zen_stores/template_utils.py:368-370
for step_name, step in step_configurations.items():
    if not step.spec.parameter_spec:
        continue
```

So any step that reaches the substitution block with an empty `config.parameters` (for example when regenerating a config schema from a previously stored spec, where `parameter_spec` is persisted but the resolved `config.parameters` dict may not be) hits the `pop` on a key that was never inserted, aborting the whole schema generation.

### Fix

Use `pop(..., None)` so the substitution path is safe regardless of whether the step's `config.parameters` produced the intermediate definition. A regression test in `tests/unit/zen_stores/test_template_utils.py::test_generate_config_schema_with_parameter_spec_and_no_config_parameters` exercises the previously failing scenario and fails on `develop` without this fix.

## Pre-requisites

- [x] I have added a regression test that reproduces the crash on `develop`.

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/template_utils.py
+++ b/src/zenml/zen_stores/template_utils.py
@@ -405,7 +405,11 @@ def _replace_step_parameter_definitions(
         step_schema.setdefault("properties", {})["parameters"] = (
             parameter_schema
         )
-        # Remove the original parameter schema #def that we just replaced
-        root_definitions.pop(f"{step_name}_parameters")
+        # The `{step_name}_parameters` entry is only added to `$defs` when
+        # the step has configured parameters (see the `create_model` call
+        # above). Steps that carry a `parameter_spec` without matching
+        # `config.parameters` - for example templates regenerated from a
+        # stored spec - never produced that entry, so pop defensively.
+        root_definitions.pop(f"{step_name}_parameters", None)
 
     return schema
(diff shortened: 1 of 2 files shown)


schustmi:
@shaun0927 I'm closing this, I don't think the case that this PR describes and tries to fix. `parameters` and `parameter_spec` depend on each other, and will only be set at the same time. If you find an actual case where this is causing a bug, please feel free to reopen this.

PR #4741, opened 2026-04-17
Clarify that DynamicPipelineRunner.wait() timeout is a soft timeout

shaun0927:
## Describe changes

Per @schustmi's guidance on #4738, this PR updates the `DynamicPipelineRunner.wait()` docstring (`src/zenml/execution/pipeline/dynamic/runner.py:1027`) to match the actual (intentional) polling behaviour.

### Before

```
timeout: Maximum time in seconds to poll before pausing.
```

That reads as a hard ceiling. The polling loop, however, is:

```python
# src/zenml/execution/pipeline/dynamic/runner.py:1121-1123
# We keep polling until the deadline is reached and all ongoing
# steps have finished.
while time.time() < deadline or self.has_in_progress_steps():
    ...
```

So a `wait(timeout=30)` call in a dynamic pipeline will keep polling for as long as any sibling step (e.g. a concurrent training future) is still in progress - which can be minutes or hours. The behaviour is intentional: the timeout is meant to bound *this* wait condition's own budget, not to hard-cancel unrelated concurrent work that just happens to run on the same pipeline.

### After

The docstring now spells out that `timeout` is a soft timeout, describes the interaction with in-progress sibling steps, and points callers who need a strict upper bound at the workaround (drain sibling futures first). No behaviour change.

### Scope

- Docstring-only; no code paths, public APIs, or tests changed.
- Matches maintainer direction on #4738: https://github.com/zenml-io/zenml/issues/4738#issuecomment-4266519563

## Pre-requisites

- [x] Docs-only change (no user-facing behaviour change, no test needed).

## Types of changes

- [x] Documentation Update

Closes #4738 (docs angle).

### DIFF
--- a/src/zenml/execution/pipeline/dynamic/runner.py
+++ b/src/zenml/execution/pipeline/dynamic/runner.py
@@ -1026,10 +1026,24 @@ class DynamicPipelineRunner:
     ) -> Any:
         """Create and poll a run wait condition.
 
+        The ``timeout`` argument is a soft timeout: once the deadline
+        passes, the runner stops scheduling further *wait* polling cycles,
+        but it keeps polling as long as sibling steps in the same dynamic
+        pipeline are still in progress. This avoids hard-cancelling
+        concurrent work (e.g. a long-running training step running next to
+        a ``wait(timeout=30)`` for external input) just because the wait
+        condition's own budget ran out. Callers that want a hard upper
+        bound on ``wait`` should drain any concurrent step futures before
+        calling ``wait``.
+
         Args:
             schema: Optional expected output type for the resolved result.
             type: Wait condition type.
-            timeout: Maximum time in seconds to poll before pausing.
+            timeout: Soft timeout in seconds. After this many seconds the
+                runner stops accepting new poll cycles for this wait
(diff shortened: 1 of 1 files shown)


JiwaniZakir (commented):
The updated `timeout` arg description in the `Args` block contains a self-referential note — "See the method-level docstring above for the full semantics" — which is slightly awkward since a reader already has to be in that same docstring to reach the `Args` section. It would be cleaner to either inline the key caveat or drop the forward reference entirely.

More importantly, the new method-level docstring doesn't address what happens if sibling steps never reach a terminal state (e.g., they hang or enter an error loop that keeps them "in-progress" in). Under the described semantics, a `wait(timeout=30)` could block forever in that scenario, which is a meaningful behavioral edge case that callers need to know about. Consider explicitly stating whether there is any secondary hard ceiling, or whether this is a known limitation.

The guidance to "drain any concurrent step futures before calling `wait`" is actionable advice, but it doesn't point to the specific API or pattern for doing so. A brief `See also` reference to the relevant method (or example) would make this self-contained for anyone who encounters the timeout semantics for the first time.

schustmi (approved):

PR #4743, opened 2026-04-20
Add acknowledgeable trigger dispatch status and improve CLI commands

stefannica:
## Summary

- Triggers now persist a **per–attached-snapshot dispatch state** so you can see whether the last run was **succeeded** (`SUCCESS`), **skipped** because of concurrency policy (`SKIPPED_CONCURRENCY`), or **failed** (`ERROR`).
- Failed dispatches store **richer error context** (message, type, optional severity and stack trace, timestamps, and consecutive count). You can **acknowledge** and clear that stored error context with the new flow—without re-attaching triggers or changing configuration.
- **Trigger CLI** commands are easier to use: you can pass **trigger and snapshot** as **name or full UUID** for attach, detach, update, delete, and `clear-errors` (exact match for names; the Python Client still supports **ID prefix** and optional name-prefix behavior where documented on `get_*_trigger`).

### Status values

- `SUCCESS` — last dispatch for that trigger–snapshot link completed successfully.
- `SKIPPED_CONCURRENCY` — dispatch was skipped due to the trigger’s concurrency policy.
- `ERROR` — dispatch failed; see stored error details (below).

### Error context on the link

- `last_error_message` — user-facing summary of what went wrong.
- `last_error_type` — classifier to group similar failures.
- `last_error_severity` — optional severity.
- `last_error_at` / `first_error_at` — window for the current error streak.
- `last_error_count` — how many times the **same error type** occurred in a row (useful for one-off blips vs repeated incidents).
- `last_error_stack_trace` — optional, truncated for storage.

### Repeated failures

- If the same `last_error_type` recurs, `last_error_count` increments and streak timestamps are updated.
- If a **different** error type appears, the streak resets (count back to 1, new `first_error_at` / `last_error_at`).
- A later **`SUCCESS`** updates `last_status` to `SUCCESS` but **does not** automatically strip the last error payload; use **clear** when you are done with incident handling (or rely on the acknowledge flow in UI/CLI).

### Acknowledge / clear errors

- **`Client.clear_trigger_dispatch_error`** and **`zenml trigger … clear-errors`** clear stored **error details** for one attachment or for **all** snapshots on a trigger, without deleting the trigger or detaching snapshots—suited to “alarm cleared” workflows.

### CLI usability

- **Update** / **delete** / **attach** / **detach** / **platform-event update** take a trigger (and snapshot where applicable) as **string name or ID**.
- **Delete** default is **soft** (archive). Use **`--hard`** for permanent removal; use **`--archived`** when the target is already archived (e.g. hard-delete an archived trigger).

### CLI examples

```bash
# List triggers
zenml trigger schedule list --active=true
zenml trigger platform-event list --active=true

# Update by name (bool options use = for Click)
zenml trigger schedule update nightly-training --active=false
zenml trigger platform-event update on-pipeline-complete --active=true

# Attach / detach: <trigger> first, then <snapshot>
zenml trigger schedule attach nightly-training fraud-pipeline-snapshot
zenml trigger schedule detach nightly-training fraud-pipeline-snapshot

# Config only on attach
zenml trigger schedule attach nightly-training fraud-pipeline-snapshot \
  --config=run-config.yml

# Delete: default = soft (archive); --hard = permanent; --archived = target archived rows
zenml trigger schedule delete nightly-training
zenml trigger schedule delete old-schedule --archived --hard
zenml trigger platform-event delete on-pipeline-complete --hard

# Clear acknowledged dispatch error details (command name: clear-errors)
# All snapshots for this trigger:
zenml trigger schedule clear-errors nightly-training
# One trigger–snapshot pair:
zenml trigger schedule clear-errors nightly-training fraud-pipeline-snapshot

# Same for platform event triggers
zenml trigger platform-event clear-errors on-pipeline-complete
zenml trigger platform-event clear-errors on-pipeline-complete my-snapshot
```

## Pre-requisites

## Types of changes


src/zenml/cli/trigger.py
```diff
@@ -261,23 +265,31 @@ def make_delete_command() -> click.Command:
...
+    def delete_trigger(trigger_name_or_id: str, soft: bool = True) -> None:
         """Delete a trigger trigger.
 
         Args:
-            trigger_id: The ID of the trigger.
+            trigger_name_or_id: The name or ID of the trigger.
             soft: Deletion mode.
         """
         try:
+            trigger_id = (
```
Json-Andriopoulos, line 279: This will not work. On archival the `name` of the trigger is changed to `<original_name>_<hash>`. The reason behind it is you may want to reuse the original name after archival.So `get_schedule_trigger` by name will not work. We have 2 possible paths:

1) Keep this delete by ID
2) Try to search by name as a prefix and delete if len(results) == 1.

If you try the 2nd method you might need to hack around the TriggerFilter `is_archived` option. This is a non-nullable boolean set by default to False (however you need to search by name on both archived and not archived sets).

Json-Andriopoulos (changes requested):

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -324,19 +328,46 @@ def to_model(
...
+                )
+                snapshots.append(display_snapshot)
+                display_snapshot_id_by_executable_id[snapshot.id] = (
+                    display_snapshot.id
+                )
+
+            snapshot_dispatch_states: dict[
+                UUID, TriggerSnapshotDispatchState
+            ] = {}
+            for snapshot_link in self.snapshot_links:
```
Json-Andriopoulos, line 350: Great catch attaching the state to the displayed snapshot!

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -324,19 +328,46 @@ def to_model(
...
-                    else s.to_model()
-                    for s in self.snapshots
-                ],
-                executable_snapshots=[s.to_model() for s in self.snapshots],
+                snapshots=snapshots,
+                executable_snapshots=executable_snapshots,
                 latest_run=latest_run.to_model()
                 if latest_run is not None
                 else None,
+                snapshot_dispatch_states=snapshot_dispatch_states,
```
Json-Andriopoulos, line 370: Just a heads up here. If you `delete` the display, original snapshot the current behavior is bugged. We display the attached snapshot as the original one so behavior when testing this feature might also be weird. Currently working on fixing this.

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -367,3 +368,42 @@ def detach_trigger_from_snapshot(
...
+) -> None:
+    """Clears recorded dispatch error details for a trigger-snapshot link.
+
+    Args:
+        trigger_id: The ID of the trigger.
+        snapshot_id: The ID of the snapshot.
+    """
+    trigger = zen_store().get_trigger(trigger_id=trigger_id, hydrate=True)
+
+    check_entitlement(feature=SCHEDULE_FEATURE)
```
Json-Andriopoulos: Not sure if we need this here. For economy of outgoing calls we only kept it as a check for endpoints that lead to extra executions/runs (create, update, attach). Endpoints that clean-up the state (delete, detach) we do not guard - I think it is ok for a user to dismiss/delete information even if they are un-subscribed to a feature.

src/zenml/zen_stores/schemas/trigger_schemas.py
```diff
@@ -265,14 +270,13 @@ def update(self, trigger_update: TriggerUpdate) -> "TriggerSchema":
...
-        if trigger_update.name is not None:
-            self.name = trigger_update.name
-
-        if trigger_update.concurrency is not None:
-            self.concurrency = trigger_update.concurrency
+        for field, value in trigger_update.model_dump(
+            exclude_unset=True,
+            include=set(TriggerBase.model_fields.keys()),
+        ).items():
+            if field in ["type"]:
```
Json-Andriopoulos, line 277: There is also `flavor` which we should probably avoid setting.

Json-Andriopoulos: Also what happens if one future field here needs additional handling (e.g. a JSON field would need json.dumps(field)). I prefer explicit statements to be honest though this is not a hard restriction.

src/zenml/models/v2/core/triggers.py
```diff
@@ -50,6 +51,115 @@
...
+        description=(
+            "Friendly user-facing message describing the latest error."
+        ),
+    )
+    last_error_type: str | None = Field(
+        default=None,
+        max_length=ERROR_TYPE_MAX_LENGTH,
+        description="Implementation-level error classifier.",
+    )
+    last_error_at: datetime | None = Field(
```
Json-Andriopoulos, line 112: Maybe the notion of severity would be useful here as well.

src/zenml/models/v2/core/triggers.py
```diff
@@ -50,6 +51,115 @@
...
+        default=None,
+        description="Timestamp of the latest recorded error.",
+    )
+    last_error_count: int = Field(
```
Json-Andriopoulos, line 116: I know I am the one who suggested this but maybe displaying counts will be wildly confusing for the users of a SaaS platform. Kind of thinking that since we display the last appearance the users already get a notion if the issue still persists and we do not need to expose more information.

src/zenml/zen_server/routers/trigger_endpoints.py
```diff
@@ -367,3 +368,42 @@ def detach_trigger_from_snapshot(
         trigger_id=trigger_id,
         snapshot_id=snapshot_id,
     )
+
+
+@router.delete(
+    "/{trigger_id}"
```
Json-Andriopoulos: TRIGGERS + "/{trigger_id}" (This was just changed, please rebase from develop to pick up the change).

src/zenml/zen_stores/schemas/trigger_schemas.py
stefannica: @Json-Andriopoulos we shouldn't ever display the attached snapshot instead of the original one. If the original snapshot is deleted, then this should be interpreted as breaking the link between trigger and snapshot and therefore there should be no relationship left between the executable snapshot and the trigger anymore.

src/zenml/zen_server/routers/trigger_endpoints.py
stefannica: good catch, will remove

src/zenml/zen_stores/schemas/trigger_schemas.py
stefannica: When we get to this, we already filter by the `**TriggerBase**` field (see the `include=` option above), so `flavor` is not included to begin with.

src/zenml/models/v2/core/triggers.py
stefannica: Maybe, but it's not required and I'd rather keep this as simple as possible for the first version, then iterate on it as we get more user feedback.

src/zenml/zen_stores/schemas/trigger_schemas.py
stefannica: If you look at other schemas, you'll see this exact pattern: a `for field, value in trigger_update.model_dump(` loop with exceptional attributes being treated inline.

src/zenml/models/v2/core/triggers.py
stefannica: Ok for taking it out, not married to it.

src/zenml/zen_server/routers/trigger_endpoints.py
stefannica: fixed

src/zenml/zen_stores/schemas/trigger_schemas.py
stefannica: :pray:

src/zenml/models/v2/core/triggers.py
stefannica: Done

src/zenml/cli/trigger.py
stefannica: I think I nailed it, but it took a bit of doing:

* I made `is_archived` a global filter setting, so it can work together with other filters when OR is used, such as when prefix-matching happens
* had to fix an inheritance bug
* made some changes to the delete trigger CLI command to make it explicit when an archived trigger is deleted

Json-Andriopoulos (approved):

src/zenml/models/v2/core/triggers.py
stefannica: Ended up adding it for UI purposes.

PR #4745, opened 2026-04-20
Helper endpoint: Trigger SourceType supported events

Json-Andriopoulos:
## Describe changes
Helper endpoint for client usage. Provides meaningful breakdown of supported target events by source type. 

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/utils/enum_utils.py
+++ b/src/zenml/utils/enum_utils.py
@@ -13,7 +13,8 @@
 #  permissions and limitations under the License.
 """Util functions for enums."""
 
-from enum import Enum
+from abc import ABCMeta, abstractmethod
+from enum import Enum, EnumMeta
 from typing import List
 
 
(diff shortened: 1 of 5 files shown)


Cahllagerfeld (commented):
One comment from my side, we're not having this `helpers` router yet. I wonder if we should nest this under the `triggers` router, because also `/helpers/supported-events` is not having any reference which events its referring to 🤔 

Ill leave this up to you ultimately, just one thought

src/zenml/zen_server/routers/helper_endpoints.py
```diff
@@ -0,0 +1,32 @@
...
+from zenml.enums import SourceType
+from zenml.zen_server.exceptions import error_response
+
+router = APIRouter(
+    prefix=API + VERSION_1 + HELPERS,
+    tags=["helpers"],
+    responses={401: error_response, 403: error_response},
+)
+
+@router.get(
```
schustmi: Wouldn't this make more sense as part of the triggers router somehow? Or is this already in preparation for these events being used in multiple places?

src/zenml/utils/enum_utils.py
```diff
@@ -45,3 +46,35 @@ def values(cls) -> List[str]:
...
+class ABCEnumMeta(ABCMeta, EnumType):
+    """Metaclass combining ABCMeta and EnumType."""
+
+
+class DescribedValuesEnum(StrEnum, metaclass=ABCEnumMeta):
+    """Enum abstraction displaying enum descriptions."""
+
+    @classmethod
+    @abstractmethod
+    def value_index(cls) -> dict[str, str]:
```
schustmi: Maybe `value_descriptions`?

Also, do we really need both methods as a dict and list of dicts?
Is the list return better than just giving e.g. the frontend the dict directly?

src/zenml/zen_server/routers/helper_endpoints.py
Json-Andriopoulos: I guess it makes more sense yeah.

src/zenml/utils/enum_utils.py
Json-Andriopoulos: 2 things: 

1) I guess it may not make a difference here but I always prefer to send a list of objects instead of a mapping. Reason is, if we need to make structural changes, any change in the mapping is breaking while objects can be freely extended. I like treating this as an enforced pattern and not judge per case.
2) If you forget to add/update the description for a new enum value this still returns the value (with a null description which is fine). If I make it a mapping, and someone forgets to update it, it becomes a blocking bug.

schustmi (approved):

PR #4748, opened 2026-04-20
k8s deployer: select Service on deployment-id, not overridable app label

### LINKED ISSUE #4740: Kubernetes deployer Service selector mismatch (ZenML 0.94.2)
We found a bug in the Kubernetes deployer where the Service created for a deployment has empty endpoints, making the deployment unreachable despite the pod being healthy.

Root cause: 
The Service template (service.yaml.j2) uses app: {{ name }} as its selector, where name is the deployment-specific resource name (e.g. zenml-<uuid>). However, in the Deployment template (deployment.yaml.j2), the pod labels set app: {{ name }} first, then two subsequent Jinja2 loops can overwrite it:

The labels dict loop — if settings.labels contains an app key
The pod_settings.labels loop — if the deployer stack component has app in pod_settings.labels

Since YAML duplicate keys resolve to the last occurrence, the pod ends up with a different app value than what the Service selector expects. The Service matches nothing, endpoints are empty, and any client gets ConnectError.

This is easy to hit in practice: it's common to set app: <service-name> in pod_settings.labels on the deployer component for observability/labeling purposes, which silently breaks the Service routing.

Affected code:
(built-in templates in zenml/integrations/kubernetes/templates/kubernetes/):

service.yaml.j2 — selector relies on app: {{ name }}
deployment.yaml.j2 — labels and pod_settings.labels loops can overwrite app

Suggested fix:
Either filter app out of both loops in deployment.yaml.j2
{% for key, value in labels.items() if key != 'app' %}
{% for key, value in pod_settings.labels.items() if key != 'app' %}Or change the Service selector to use zenml-deployment-idinstead of app, which is unique per deployment and not subject to override.

We worked around it with a custom deployment template using the filter approach.

joaquinhuigomez:
The generated Service selected on `app: {{ name }}`, but the pod template sets the same `app` label first and then lets user-supplied `labels` and `pod_settings.labels` overwrite it. If either dict contains an `app` key, the pod ends up with a different `app` label than the Service selector expects, leaving the Service with zero endpoints even though the pod is healthy.

Select on the two labels the Deployment's `matchLabels` already uses — `managed-by` and `zenml-deployment-id` — since those are deployment-unique and not part of the user-controlled labels dict.

Fixes #4740


joaquinhuigomez:
Still relevant — the selector mismatch reproduces on develop. Happy to rebase if reviewer bandwidth opens up.

safoinme (approved):
Look good, Thank you for fixing this

bcdurak:
hey @joaquinhuigomez, thank you for your contribution. The changes are merged and will be part of the next release.

PR #4754, opened 2026-04-21
Bugfix/curate deleted trigger assocs

Json-Andriopoulos:
## Describe changes

Bugfix and improvements branch:

1) On snapshot deletion propagate and drop trigger_snapshot associations of the snapshot object.
2) Make pipeline_run.pipeline_id as body field as a speed up/performance improvement.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -5057,7 +5057,6 @@ class SqlZenStore(BaseZenStore):
             snapshot_id: The ID of the snapshot (to be deleted or un-named).
             session: SQLAlchemy session
         """
-
         linked_snapshot_ids = session.exec(
             select(PipelineSnapshotSchema.id).where(
                 PipelineSnapshotSchema.source_snapshot_id == snapshot_id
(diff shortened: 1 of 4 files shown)


Json-Andriopoulos:
Making this draft and pointing it to `feature/max-runs-schedule-stop-criteria` as it should be merged right after it.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5368,6 +5413,16 @@ def delete_snapshot(self, snapshot_id: UUID) -> None:
                 snapshot.source_snapshot_id = None
                 session.add(snapshot)
 
+            # Remove the attached trigger snapshots
+
+            session.execute(
```
schustmi, line 5418: Right now, one can attach an unnamed snapshot to a trigger. is that something that we should stop as well somehow, or just remove them when "deleting"?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5046,6 +5046,36 @@ def _snapshot_exists(
         with Session(self.engine) as session:
             return session.exec(query).first() is not None
 
+    def _drop_snapshot_trigger_assoc(
```
Json-Andriopoulos, line 5049: @schustmi Does any other place come to mind where we would need to call this?

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Good point, I will add a guard to make sure the snapshot is named 👍

PR #4769, opened 2026-04-24
Increase the max size of zenml secrets stored in the SQL secrets store

stefannica:
## Describe changes
Increases the max allowed size of the zenml secrets to 64kb when the zenml server's database is used as a secrets storage backend. The limit is applied to the summed sizes of all keys and values associated with the same secret object.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/f2c8e4a1903b_increase_secret_values_mediumtext.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+
+
+def downgrade() -> None:
+    """Narrow `secret.values` back to unbounded TEXT-style storage."""
+    with op.batch_alter_table("secret", schema=None) as batch_op:
+        batch_op.alter_column(
+            "values",
+            existing_type=_MEDIUM_STRING,
+            type_=sa.TEXT(),
```
schustmi, line 41: what happens here if values exceed the limit? Should we just prevent the downgrade to avoid this case at all?

stefannica: If the values exceed the limit, the downgrade will exit with an error, which is exactly what we want.

schustmi (approved):

PR #4771, opened 2026-04-27
Check status changed before dispatching

Json-Andriopoulos:
## Describe changes

Checks status change before dispatching. Resolves duplicate submissions.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -12220,10 +12220,13 @@ class SqlZenStore(BaseZenStore):
         session.commit()
         if not did_update:
             return
-        EventDispatcher().handle_run_status_update(
-            run=pipeline_run.to_model(include_metadata=True)
-        )
+
         new_status = ExecutionStatus(pipeline_run.status)
+
+        if new_status != previous_status:
+            EventDispatcher().handle_run_status_update(
+                run=pipeline_run.to_model(include_metadata=True)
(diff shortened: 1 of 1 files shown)


src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -12220,9 +12220,13 @@ def _update_pipeline_run_status(
         session.commit()
         if not did_update:
             return
-        EventDispatcher().handle_run_status_update(
-            run=pipeline_run.to_model(include_metadata=True)
-        )
+
+        if requested_status is not None and previous_status != ExecutionStatus(
```
schustmi: I think it'll be safer to use `new_status` (=`pipeline_run.status`) instead of requested status.

In static pipelines, the `requested_status` is often `None` (e.g. when a step gets updated), and we internally compute the new status value depending on all the step statuses of the run.

schustmi (changes requested):

src/zenml/zen_stores/sql_zen_store.py
schustmi: For example: The last step of a run finishes. It will call this method with `requested_status=None`, but `pipeline_run.status` will still change to `completed`.

Json-Andriopoulos: Good point!

schustmi (approved):

PR #4778, opened 2026-04-28
Add isnull/isnotnull filter operators

### LINKED ISSUE #4761: Allow filtering by null values

demian-overflow:
## Summary

Resolves #4761.

The base filter framework treats `None` as "no filter" (skipped at `BaseFilter._generate_filter_list`), so there is currently no way to express "this column is NULL" through the public API. This PR adds two unary operators — `isnull:` and `isnotnull:` — and a `NullFilter` that emits `column.is_(None)` / `column.is_not(None)` regardless of the column's datatype.

End-to-end, this means:

```python
PipelineRunFilter(orchestrator_run_id="isnull:")
# -> WHERE pipeline_run.orchestrator_run_id IS NULL

PipelineRunFilter(orchestrator_run_id="isnotnull:")
# -> WHERE pipeline_run.orchestrator_run_id IS NOT NULL
```

## Why a separate Filter class

Each datatype-specific factory in `FilterGenerator.define_filter` (datetime, UUID, int, bool, str) casts the user-supplied value (`int(value)`, `datetime.strptime(value)`, …). Those casts would crash on the empty value that follows `isnull:`. Routing null operators to a dedicated `NullFilter` short-circuits before any casting and keeps the rest of the type-specific factories untouched.

## Scope notes

- Works on every filter field that is *not* in `FILTER_EXCLUDE_FIELDS` and accepts the standard `op:value` string protocol (i.e., the field type is `Optional[str]` or `Optional[Union[X, str]]`).
- Fields handled by `get_custom_filters` (e.g., `source_snapshot_id`, `templatable`, the various JOIN-based filters in `PipelineRunFilter`) keep their bespoke logic and would need per-field follow-ups to honor `isnull:`. The `source_snapshot_id` example from the issue belongs to that bucket — it can be migrated separately, now that the framework primitive exists.
- Pure `Optional[bool]` / `Optional[UUID]` fields without a `str` fallback in their `Union` won't accept the operator string at the pydantic layer; a future PR can widen those types if there's demand.

## Backwards compatibility

Pure addition: two new enum members and one new Filter subclass. No existing filter inputs change behavior.

## Test plan
- [x] `pytest tests/unit/models/test_filter_models.py` (41 tests, all green)
- [x] `pytest tests/unit/utils/` (162 tests pass)
- [x] Manual SQL render check via `BaseFilter.generate_filters` confirms `IS [NOT] NULL` for `orchestrator_run_id` on `PipelineRunSchema`
- [x] `ruff check`, `ruff format`, `mypy`, `pydoclint` clean on changed files

### DIFF
--- a/src/zenml/enums.py
+++ b/src/zenml/enums.py
@@ -366,6 +366,8 @@ class GenericFilterOps(StrEnum):
     LTE = "lte"
     LT = "lt"
     IN = "in"
+    ISNULL = "isnull"
+    ISNOTNULL = "isnotnull"
 
 
 class SorterOps(StrEnum):
--- a/src/zenml/models/v2/base/filter.py
+++ b/src/zenml/models/v2/base/filter.py
@@ -144,6 +144,35 @@ class Filter(BaseModel, ABC):
         """
 
 
+class NullFilter(Filter):
+    """Filter that emits `IS NULL` / `IS NOT NULL` for any nullable column.
+
+    Routed by `FilterGenerator.define_filter` whenever the user supplies the
+    `isnull:` or `isnotnull:` operator, regardless of the column's datatype.
+    These operators are unary, so the filter carries no value.
+    """
+
+    ALLOWED_OPS: ClassVar[List[str]] = [
+        GenericFilterOps.ISNULL,
+        GenericFilterOps.ISNOTNULL,
+    ]
+
+    value: None = None
+
+    def generate_query_conditions_from_column(self, column: Any) -> Any:
+        """Generate query conditions for a null check on any column.
+
+        Args:
+            column: The column of an SQLModel table on which to filter.
+
+        Returns:
+            The query condition.
+        """
+        if self.operation == GenericFilterOps.ISNOTNULL:
+            return column.is_not(None)
+        return column.is_(None)
+
+
 class BoolFilter(Filter):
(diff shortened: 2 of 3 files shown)


schustmi (approved):

schustmi:
@demian-overflow Thanks for the PR! Code looks good to me, I'll do some tests later this week and will merge once I'm done with it.

bcdurak:
hey @demian-overflow, thank you for your contribution. As @schustmi mentioned, the changes look good, however, we have already started working on this issue in a larger scale. You can see the changes here [in this PR](https://github.com/zenml-io/zenml/pull/4566). I will take a look to see if I can bring the changes together.

bcdurak:
Hey @demian-overflow, as I have mentioned, since we are dealing with this issue in a different PR, I will close this one. In any case, thanks a lot for your contribution ❤️

PR #4790, opened 2026-04-30
Updating the `opentelemetry-sdk` dependency

### LINKED ISSUE #4753: Bump the `opentelemetry-sdk` version
We need to unpin `opentelemetry-sdk` as a zenml dependency, or at least allow `opentelemetry-sdk==1.39.0` to be used (currently pinned to `opentelemetry-sdk==1.38.0`).

## Why `opentelemetry-sdk==1.38.0`?

There is **no comment in `pyproject.toml`** explaining the pin. The only `git log` hit for that dependency is the original **log store abstraction** work (`#4111`). So the likely reasons are the usual ones: **reproducible installs**, **avoiding surprise upgrades** while the Logs SDK was still changing, and **keeping `uv.lock` stable**. It is not documenting a known incompatibility with 1.39+.

## How ZenML uses the SDK (relevant to an upgrade)

| Area | What you use |
|------|----------------|
| **`OtelLogStore`** | `LoggerProvider`, `LoggingHandler`, `Resource`, `BatchLogRecordProcessor` (subclassed), `Logger.emit`, and **`LoggingHandler._translate`** (private API). |
| **`OtelBatchLogRecordProcessor`** | Overrides `flush`; uses **`_batch_processor._worker_awaken`** (private API) for non-blocking flush. |
| **`OTLPLogExporter` / `ArtifactLogExporter` / `DatadogLogExporter`** | Implement `LogExporter` with **`export(batch: Sequence[LogData])`**, and read **`log_data.log_record`**, **`instrumentation_scope`**, resource, attributes, etc. |

The integration is **centered on custom `LogExporter` implementations and the batch log pipeline**, not traces/metrics.

## Can we loosen to `>=1.39.0`?

**Not without code changes.** The **1.39.0** line on the releases page includes an explicit **breaking change**:

- **`LogData` is removed**; exporters should take **`Sequence[ReadableLogRecord]`** instead, and the migration guide describes the new shapes (`ReadableLogRecord` / `ReadWriteLogRecord`) and that **`instrumentation_scope` now lives on that readable type** rather than on the old `LogData` wrapper.

Our code still imports and types against `LogData` in:

- `artifact_log_exporter.py`
- `otel_log_exporter.py`
- `datadog/datadog_log_exporter.py` (via `_encode_log` overrides)

Until those are migrated per the upstream migration notes, **`opentelemetry-sdk>=1.39.0` will not be a safe “loosen only” change**.

## Other release notes worth attention (after 1.38)

1. **`BatchLogRecordProcessor` default `schedule_delay_millis` (spec alignment)**  
   A later release notes that the **SDK default** was changed **5000 ms → 1000 ms** so behavior matches the spec unless configured otherwise.  
   **ZenML’s `OtelLogStoreConfig` sets `schedule_delay_millis` explicitly** (default from `LOGS_OTEL_SCHEDULE_DELAY_MILLIS`, currently **5000** in `constants.py`), so **your default export cadence should stay as today** unless someone bypasses that config path. Users who only relied on the **raw OTel env default** without ZenML’s field could see a behavior shift—that is mostly outside your wrapper.

2. **`LoggingHandler` deprecation**  
   The SDK is steering people toward **`opentelemetry-instrumentation-logging`**. `OtelLogStore` still uses `LoggingHandler`; this is **deprecation / future migration**, not an immediate break.

3. **Python `CRITICAL` → OTel `FATAL` severity text**  
   Mapping was aligned with the spec. Your **`ArtifactLogExporter._map_severity_to_level`** already treats **`FATAL`** like critical, so this is **unlikely to be a problem**.

4. **Private SDK internals**  
   Use of **`_translate`** on `LoggingHandler` and **`_worker_awaken`** on the batch processor is **brittle** on any upgrade: not called out in every release note, but worth treating as ongoing maintenance risk.

## Practical takeaway

- **Reason for 1.38.0 pin:** best explained as **stability at import time** when the feature landed, not a documented permanent ceiling.
- **Loosening to `>=1.39.0`:** do it **together with** migrating custom exporters off **`LogData`** to **`ReadableLogRecord`** (and any renamed result types the SDK expects), plus a quick pass on **`LoggingHandler` / private attributes** after running tests.

bcdurak:
## Describe changes

We need to unpin opentelemetry-sdk as a zenml dependency, or at least allow opentelemetry-sdk==1.39.0 to be used (currently pinned to opentelemetry-sdk==1.38.0).

There is a detailed explanation about it in the linked issue as well.

BREAKING CHANGE: This PR changes the pin of the dependency `opentelemetry-sdk` from `1.38.0` to `1.40.0`.

## Pre-requisites

## Types of changes


bcdurak:
@stefannica, fyi, since this is changing the implementation of the `OtelLogStore`, I am categorizing this as a breaking change.

stefannica (approved):

PR #4800, opened 2026-05-06
Extra info on kubernetes-killed jobs

Json-Andriopoulos:
## Describe changes

Provides richer information on system failures  of dynamic pipeline jobs (e.g. Kubernetes kills pod due to OOM)

Example of an isolated step failing due to OOM:

Before:

Screenshot 2026-05-11 at 3 53 25 PM

After:

Screenshot 2026-05-11 at 3 53 46 PM

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/kubernetes/kube_utils.py
+++ b/src/zenml/integrations/kubernetes/kube_utils.py
@@ -890,6 +890,59 @@ def get_container_termination_reason(
     )
 
 
+def get_pod_failure_details(
+    pod: k8s_client.V1Pod, container_name: str
+) -> Optional[str]:
+    """Get best-effort pod failure details.
+
+    Args:
+        pod: The pod to get failure details for.
+        container_name: The container name.
+
+    Returns:
+        Failure details for the pod, if available.
+    """
+    try:
+        details: List[str] = []
+        container_state = get_container_status(pod, container_name)
+        if container_state and container_state.terminated:
+            terminated = container_state.terminated
+            container_details = [
+                terminated.reason or "Unknown",
+                f"exit_code={terminated.exit_code}",
+            ]
+            if terminated.message:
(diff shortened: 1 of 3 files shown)


src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1140,6 +1145,100 @@ def check_job_status(
     return JobStatus.RUNNING, None
 
 
+def build_step_run_failure_exception_info(
```
schustmi: This does not seem like the right place. This file I think should be limited to kubernetes utilities, while these two functions here are purely ZenML step status related.

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1140,6 +1145,100 @@ def check_job_status(
...
+    except Exception as e:
+        logger.warning(
+            "Failed to fetch step run `%s` before publishing Kubernetes "
+            "failure details: %s",
+            step_run.id,
+            e,
+        )
+        return
+
+    if step_run.exception_info:
```
schustmi: The whole point of `exception_info` was to reconstruct the original exception that occurred in the step, so it can be re-raised in different environments.

Even with the check here, this is super prone to race conditions and overwriting potential real exception info.

Were there any concrete error cases that this PR should surface to the user? Maybe there are other solutions.

schustmi (changes requested):

src/zenml/integrations/kubernetes/kube_utils.py
Json-Andriopoulos: Yeah, this is probably over-engineered. I think just adding better logging on the key orchestrator entrypoints to show more details on interrupted jobs (e.g. OOM) should be good enough.

schustmi (approved):

PR #4804, opened 2026-05-07
Allow streaming events while running a pipeline

schustmi:
## Describe changes

Live event streaming on pipeline runs over SSE. Can be enabled by setting `stream_broker_implementation_source`.

  ## Producer

Call  `zenml.streaming.publish({...}, kind=..., correlation_id=..., index=...)` from inside any step or dynamic pipeline.

  ## Server

  Add `StreamBroker` interface and `RedisStreamsBroker` implementation. `StreamBroadcaster` runs one broker reader per stream, fans out to N SSE consumers, handles catch-up, gap signalling, idle cleanup, per-stream consumer cap.

  ## Endpoint

  GET /api/v1/runs/{id}/events/stream

  Multi-value filters: `?kinds=`, `?step_names=`, `?correlation_ids=`. Resume via `Last-Event-ID` header or `?since=`.

## Pre-requisites

## Types of changes


src/zenml/zen_server/streaming/brokers/redis_streams.py
```diff
@@ -0,0 +1,286 @@
...
+            response = await client.xrevrange(stream_key, count=1)
+        except Exception as exc:
+            raise BrokerConnectionError(str(exc)) from exc
+
+        if not response:
+            return None
+
+        return _decode(response[0][0])
+
+    async def delete_stream(self, stream_key: str) -> None:
```
Json-Andriopoulos, line 229: We should make sure this is called on run completion, run deletion.

schustmi: Wouldn't the 1h timeout just eventually clean the stream? I actually reverted any changes that called this, so once a run completes we keep the stream around a little for potential consumers.

schustmi: It certainly comes with many complications:
- Immediate deletion would mean any currently subscribed consumers might not receive the latest events before the stream gets deleted. Also publishing would be more complicated, as we'd need to guarantee all events are flushed before sending a run status update.
- Runs get deleted also when some foreign key relations (snapshot, pipeline) get deleted. Catching all those cases will be tricky and is not reasonable IMO.
- If we delay stream deletion by some grace period, what happens if a new consumer subscribes to the stream (on a different server replica). For them, the stream would just shut down in the middle eventually, right?

Json-Andriopoulos: Ah, we use expire to do that lazily, sorry I missed that. I think that covers my concern.

src/zenml/constants.py
```diff
@@ -662,3 +665,9 @@ def handle_float_env_var(var: str, default: float = 0.0) -> float:
 RESOURCE_POOL_FEATURE = "resource_pool"
 
 LOGS_RUNNER_SOURCE = "runner"
+
+STREAM_EVENT_MAX_BATCH_SIZE = 1000
+STREAM_EVENT_PAYLOAD_BYTES_MAX = 64 * 1024
+STREAM_PUBLISHER_BATCH_SIZE: int = handle_int_env_var(
```
Json-Andriopoulos: We should validate this is > 0 .

src/zenml/zen_server/routers/runs_endpoints.py
```diff
@@ -666,3 +708,244 @@ def replay_run(
...
+
+    try:
+        ids = await stream_broker().publish(
+            stream_key_for_run(pipeline_run_id), payloads
+        )
+    except Exception as exc:
+        logger.exception("Broker publish failed for run %s", pipeline_run_id)
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail=f"Broker publish failed: {exc}",
```
Json-Andriopoulos: I am not sure what information the exception message might leak here. Maybe a more controlled message would be better?

src/zenml/zen_server/streaming/sse.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    try:
+        while True:
+            done, _ = await asyncio.wait(
+                {next_task},
+                timeout=heartbeat_seconds,
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            if not done:
+                # TODO: consider checking the run status here and synthesizing
```
Json-Andriopoulos, line 178: This is interesting I like it. We can reach such conditions, I think it would be good to have it implemented as part of this PR.

src/zenml/streaming/publishing.py
```diff
@@ -0,0 +1,368 @@
...
+    """
+    if kind in _RESERVED_KINDS:
+        raise ValueError(
+            f"`{kind}` is reserved for server-emitted control frames "
+            f"({sorted(_RESERVED_KINDS)})"
+        )
+    ctx = _resolve_publish_context()
+    if ctx is None:
+        logger.debug(
+            "`publish(...)` called outside a pipeline run, ignoring event: %s",
```
Json-Andriopoulos, line 357: This should be a warning I think

src/zenml/zen_server/streaming/sse.py
schustmi: The only reason I didn't implement this was that I'm concerned about overloading our server. 
- there might be multiple consumers of the same stream on different replicas
- for each consumer, this would send a request for the run status every 30 seconds or whatever the configured heartbeat interval
- the current implementation scales pretty well with multiple consumers, as each replica only subscribes to each stream only once. but this would now add DB load relative to the consumer count

From my understanding (but I might be wrong), idling consumers waiting on an empty stream shouldn't consume any noticeable resources.

On the other hand, it usually won't really matter for the consumer if the connection isn't closed as soon as the run ends. The user will usually be on a website that displays the stream, and once they change the URL the browser will close the connection.

So essentially, we'd add these potentially many run status checks for little benefit, and the happy case should already publish the end frame just fine, so all of this would be just to cover some unlikely edge cases.

What do you think?

src/zenml/streaming/publishing.py
```diff
@@ -0,0 +1,368 @@
...
+
+_PAYLOAD_ADAPTER: TypeAdapter[Dict[str, Any]] = TypeAdapter(Dict[str, Any])
+
+# Mirrors `SSEEventName` values in zen_server/streaming/types.py.
+_RESERVED_KINDS: FrozenSet[str] = frozenset(
+    {"end", "gap", "error", "cursor", "system"}
+)
+
+
+def _check_payload_size(payload: Dict[str, Any]) -> None:
```
Json-Andriopoulos: Minor: I am thinking this function should either be a validator of the StreamEvent itself or it should be a method of the StreamPublisher.

schustmi: The reason this isn't a validator on the stream event class is because otherwise we would run the json-encoding twice on the server side. I'll move it to the publisher class

src/zenml/zen_server/streaming/sse.py
Json-Andriopoulos: Hmm I was thinking about a more infrequent HB (5, 10 minutes would make sense) to avoid the mid/long-term effect of dangling sessions but even that is probably a stretch. An alternative here is: We don't care as long as we can inspect it. For instance, there is a server performance issue how can I rule out SSE, queues etc?

Json-Andriopoulos (approved):

src/zenml/zen_server/routers/runs_endpoints.py
schustmi: Done

src/zenml/zen_server/streaming/sse.py
schustmi: @Json-Andriopoulos What do you mean by not caring here? Not caring about idle connections? Or not caring about DB calls?

Json-Andriopoulos: Idle connections :)

PR #4813, opened 2026-05-08
Upgrade orjson for dependency audit

strickvl:
## Description

This is the second remediation PR stacked on top of the dependency audit workflow branch:

- Base: `fix/pytest-audit-compatibility`
- Previous PR: #4812

It resolves the `orjson` audit finding only. PyJWT and Starlette findings are intentionally left for follow-up stacked PRs.

## Changes

- Replace `orjson~=3.10.0` with `orjson>=3.11.6,<3.12.0`.

This allows resolution of the audit-fixed version for `CVE-2025-67221` while keeping the dependency within the 3.11 line.

## Validation

Targeted local validation:

- `uv run --extra dev --extra server --extra local pytest tests/unit/zen_server/test_exceptions.py -q`
  - Result: `36 passed`

Dependency-audit smoke for isolated `.[server,dev,local]`:

- `orjson` resolved to `3.11.8`
- `uv pip check` passed
- `pip-audit` no longer reports the `orjson` finding
- Remaining expected findings:
  - `pyjwt 2.7.0` / `CVE-2026-32597`
  - `starlette 0.45.3` / `CVE-2025-54121`
  - `starlette 0.45.3` / `CVE-2025-62727`

Integration tests were not run locally; CI should validate broader behavior.

## Release notes

No release notes needed; this is dependency security maintenance.

PR #4814, opened 2026-05-08
Upgrade PyJWT for dependency audit

### LINKED ISSUE #4782: Relax `pyjwt==2.7.*` pin in `[server]` extra (blocks security upgrade for `mcp`)
## Summary

The `[server]` extra in `zenml` 0.94.3 pins `pyjwt[crypto]==2.7.*` exactly. This pin currently blocks two security-relevant upgrades for downstream consumers:

1. **`pyjwt` itself has a known vulnerability**: [CVE-2026-32597](https://nvd.nist.gov/vuln/detail/CVE-2026-32597), fixed in `pyjwt 2.12.0`. The current `==2.7.*` pin makes it impossible for downstream projects to pick up the patched version.
2. **It transitively blocks `mcp` upgrades**: `mcp >= 1.20.0` requires `pyjwt[crypto]>=2.10.1`. As a result, projects using both `zenml[server]` and `mcp` cannot upgrade beyond `mcp 1.19.0`, which has [CVE-2025-66416](https://github.com/modelcontextprotocol/python-sdk/security/advisories) (fixed in `mcp 1.23.0`).

## Request

Could the pin be widened to something like `pyjwt[crypto]>=2.7,<3` (or whichever upper bound is safe based on ZenML's actual API usage of pyjwt)? If `2.7.*` is intentional because of a known incompatibility with newer pyjwt versions, it would help to document that in the dep declaration or pyproject comments.

## Context

Discovered while running `pip-audit` on [Kitaru](https://github.com/zenml-io/kitaru), which depends on `zenml[local]>=0.94.3` and the `mcp` extra. We can patch most CVEs via lockfile refresh, but these two are upstream-blocked.

Happy to open a PR if helpful — just need to know what the upper bound should be.

strickvl:
## Description

This is the third remediation PR stacked on top of the dependency audit workflow branch:

- Base: `fix/orjson-audit-remediation`
- Previous PRs: #4812, #4813

It resolves the PyJWT audit finding only. Starlette findings are intentionally left for the next framework-focused PR.

## Changes

- Replace `pyjwt[crypto]==2.7.*` with `pyjwt[crypto]>=2.13.0,<2.14.0`.
  - 2.13.0 patches several advisories that affect `<= 2.12.1`, so an earlier
    cap of `<2.13.0` would have left the dependency audit red on PyJWT:
    - `GHSA-jq35-7prp-9v3f` — algorithm allow-list bypass via `PyJWK`/`PyJWKClient`
    - `GHSA-993g-76c3-p5m4` — `PyJWKClient` SSRF + token forgery
    - `GHSA-fhv5-28vv-h8m8` — `PyJWKClient` JWKS DoS
    - `GHSA-w7vc-732c-9m39` — DoS via Base64URL decode in detached JWS
  - None of these are reachable in ZenML's runtime (HS256 with a static
    `jwt_secret_key`, no `PyJWKClient`), but `pip-audit` matches on installed
    version regardless, so the floor is set to `>=2.13.0` to force the patched
    line.
- Modernize `JWTToken.decode_token()` to use PyJWT's supported `options={...}`
  mechanism instead of the deprecated/ineffective `verify=...` argument.
- Make `verify=False` **all-or-nothing**: it now skips the signature *and*
  every other claim check (exp, aud, iss). Validating claims such as audience
  on a token whose signature is unverified gives no real guarantee, since an
  unsigned token can be forged with any claims. PyJWT disables the remaining
  checks automatically once `verify_signature` is False, so only that one flag
  is set. This also matches the ZenML Pro server, which reuses this
  implementation and expects no verification when `verify=False`.

## Regression coverage

- wrong-signature tokens are still rejected by default
- `verify=False` skips signature verification (returns the decoded claims)
- `verify=False` now skips *all* claim checks (wrong-audience token decodes)
- download-token generation/verification
- download-token extra-claim mismatch rejection

## Validation

- `ruff format` + `ruff check --select I` on `jwt.py` and `test_jwt.py`: clean
- `pytest tests/unit/zen_server/test_jwt.py -q`: `11 passed`
- Verified the `verify_signature: False` cascade (all-or-nothing) behaves
  identically on PyJWT 2.7.0, 2.12.1, and 2.13.0.
- CI should validate the broader `.[server,dev,local]` resolution against
  PyJWT 2.13.x and the rest of the dependency tree.

## Release notes

No release notes needed; this is dependency security maintenance.


stefannica (approved):
I reviewed this against our **current** `develop` tree to see where we issue and verify JWTs, and to judge **backward compatibility for long-lived bearer tokens** after moving to PyJWT 2.12.x and replacing `verify=` with `options=`.

### Summary

The dependency bump and the `options={...}` migration are directionally correct. After tracing **all** PyJWT issuance in the server, **access tokens that this server has ever issued through `JWTToken` should remain decodable** under PyJWT 2.10+ rules (notably `sub` must be a string when present). **Download JWTs** are a separate, narrower payload and do not go through `JWTToken.decode_token`; they should also remain valid unless PyJWT changes default checks for claims they do not carry.

I do not see a structural reason this would invalidate existing CLI or device bearer tokens **as long as** the same `jwt_secret_key`, algorithm, and issuer/audience semantics remain on the server.

---

### Dependency change

| | Version |
|--|--|
| **Before** | `pyjwt[crypto]==2.7.*` |
| **After (per PR)** | `pyjwt[crypto]>=2.12.0,<2.13.0` (e.g. 2.12.1) |

That crosses **2.10.0**, where PyJWT began validating `sub` / `jti` when those claims appear ([changelog](https://pyjwt.readthedocs.io/en/latest/changelog.html#v2-10-0)), which caused real ecosystem breakage for non-string `sub` ([discussion](https://github.com/jpadilla/pyjwt/issues/1017)). That is the main **token-level** risk when upgrading; our issuance path is relevant below.

---

### Where we issue JWTs (source of truth for compatibility)

There are only **two** production patterns that call `jwt.encode` in the ZenML server:

| Token kind | Issuer | `sub` and other registered claims |
|------------|--------|-----------------------------------|
| **OAuth / API / device / cookie access token** | `generate_access_token` in `auth.py` builds `JWTToken(...).encode(...)` | `JWTToken.encode` sets `sub` to `str(self.user_id)`, plus string `iss` / `aud` from server config, optional `exp`, and optional UUID claims as **strings** (`device_id`, `api_key_id`, etc.). See ```198:226:src/zenml/zen_server/jwt.py``` and ```1080:1089:src/zenml/zen_server/auth.py```. |
| **Short-lived download token** | `generate_download_token` in `auth.py` | Payload is `exp`, `download_type`, `resource_id`, plus optional `extra_claims`. There is **no** `sub` (or `iss` / `aud`) in the default payload. See ```1134:1147:src/zenml/zen_server/auth.py```. |

No other `jwt.encode` call sites showed up under `src/zenml/` besides tests.

**Important detail for `sub`:** `encode` starts from `self.claims.copy()` but then **always** assigns `claims["sub"] = str(self.user_id)` (```198:202:src/zenml/zen_server/jwt.py```), so callers cannot accidentally leave a non-string `sub` in the serialized token even if `claims` were populated oddly.

**Decoding paths:**

- Bearer access tokens: `JWTToken.decode_token` (your PR updates this), used from `auth.py` when the `Authorization` bearer is not an API key prefix (see ```271:273:src/zenml/zen_server/auth.py```).
- Download tokens: `verify_download_token` uses `jwt.decode` with **only** `algorithms=...` (```1171:1177:src/zenml/zen_server/auth.py```), not `JWTToken.decode_token`, so the `verify` → `options` change in `JWTToken` does not affect download verification directly.

---

### PyJWT changes vs this PR

1. **`verify=` on `jwt.decode`**  
   Deprecated for a long time in favor of `options`. Aligning `JWTToken.decode_token` with `options={...}` matches current PyJWT and matches what we already do in tests when decoding with verification disabled (see `tests/unit/zen_server/test_jwt.py` helper decodes using `options={...}`). Good change.

2. **`sub` validation from PyJWT 2.10+**  
   Because every access token from `JWTToken.encode` uses **`sub` as a string UUID**, we align with PyJWT’s stricter checks. That directly mitigates the main real-world breakage people saw when moving to 2.10.x.

3. **Download tokens without `sub`**  
   PyJWT’s added `sub` validation applies when the claim is **present**; tokens without `sub` are not forced to add one. Download JWTs therefore should not suddenly fail solely because of `sub` typing.

---

### Will existing long-lived JWTs (CLI, device, etc.) still work after deploy?

**Expected answer: yes**, for tokens issued by this codebase’s `JWTToken` path, **provided** deployment-time invariants are unchanged:

- Same `jwt_secret_key` and algorithm (default HS256 from `constants.py`).
- Same effective **`iss`** and **`aud`** as when the token was minted (defaults tie to deployment id in `ServerConfiguration.get_jwt_token_*`).
- Token not expired beyond configured leeway, same as today.

Reasoning is not hand-waving: **every** access token goes through `JWTToken.encode`, which **forces** string `sub` and string `iss` / `aud`. Older PyJWT already enforced signature and those claims when decoding with `audience=` and `issuer=`; the upgrade does not redefine the JWS format or the signing key. Stricter PyJWT behavior mainly affects **invalid** tokens (wrong `sub` type, etc.), which our issuer path does not produce.

---

### Conclusion

From repository inspection, **we only mint access JWTs through `JWTToken.encode`, which always emits a string `sub` and standard string `iss` / `aud`**. That lines up with PyJWT 2.10+ expectations, so **existing valid bearer tokens should continue to validate after this upgrade**, assuming server secrets and issuer/audience configuration are unchanged. The `verify` → `options` update is appropriate and reduces reliance on deprecated API.

Looks good to me from a backward-compatibility perspective.

src/zenml/zen_server/jwt.py
```diff
@@ -82,6 +82,14 @@ def decode_token(
...
         """
         config = server_config()
+        verification_options = {
+            "verify_signature": verify,
+            "verify_exp": True,
+            "verify_nbf": True,
+            "verify_iat": True,
+            "verify_aud": True,
+            "verify_iss": True,
+        }
```
stefannica:
```suggestion
           verify: Whether to verify the token.

       Returns:
           The decoded JWT access token.

       Raises:
           CredentialsNotValid: If the token is invalid.
       """
config = server_config()
        verification_options = {
            "verify_signature": verify,
        }
```

On second thought, I'm just a tiny bit concerned here that the `verify` flag coming from this function arguments only controls signature verification.  Even when I set `verify` to false, there is still some verification taking place for the other fields like the audience and expiration. I understand this isn't really used anywhere currently in the ZenML code. Verify is always set to true. And it was ignored anyway by the PyJWT library. I would still recommend that we make verify an all or nothing verification flag even though the docstring says it's only verifying the signature. I'm using this exact same implementation in the ZenML Pro server and there I expect no verification to happen when `verify` is set to False.

stefannica (changes requested):
I had a change of mind...

PR #4815, opened 2026-05-08
Upgrade FastAPI and Starlette for dependency audit

### LINKED ISSUE #4792: Pinned starlette version has a "High" vulnerability

_No response_

### System Information

_No response_

### Issue Description

Zenml pins a version of FastAPI which in turn pins starlette to 0.45.3, which is affected by https://github.com/advisories/GHSA-7f5h-v6xp-fcq8

Dependencies should be relaxed to allow starlette>=0.49.1 to fix the issue

### Code Snippet [Optional]

```markdown
import zenml

# Your code here
```

### Code of Conduct

strickvl:
## Description

This is the final remediation PR stacked on top of the dependency audit workflow branch:

- Base: `fix/pyjwt-audit-remediation`
- Previous PRs: #4812, #4813, #4814

It resolves the remaining Starlette audit findings by moving FastAPI to a narrow range that resolves to an audit-safe Starlette version through FastAPI's own dependency metadata.

## Changes

- Change FastAPI bound:
  - from `fastapi>=0.100,<=0.115.8`
  - to `fastapi>=0.120.1,<0.121.0`

FastAPI `0.120.x` declares `starlette>=0.40.0,<0.50.0`. With the current resolver, `fastapi>=0.120.1,<0.121.0` installs Starlette `0.49.3`, which is above the vulnerable versions reported by `pip-audit`.

This keeps the framework move constrained while avoiding a direct ZenML dependency on Starlette, which is a nested dependency managed by FastAPI.

## Validation

Implementation-agent validation in isolated temp venv with `.[server,dev,local]` before the follow-up review change:

| Check | Result |
|---|---|
| Dependency resolution/install | Passed |
| Installed FastAPI | `0.120.4` |
| Installed Starlette | `0.49.3` |
| Installed PyJWT | `2.12.1` |
| Installed orjson | `3.11.8` |
| Installed pytest | `9.0.3` |
| `uv pip check` | Passed |
| `pytest tests/unit/zen_server/test_exceptions.py tests/unit/zen_server/test_jwt.py -q` | `47 passed` |
| Zen server app/middleware import smoke | Passed |
| `pip-audit --path ... --strict` | `No known vulnerabilities found` |
| `git diff --check` | Passed |

The smoke check covered imports/setup for FastAPI, `StaticFiles`, `FileResponse`, default `ORJSONResponse`, and the expected middleware stack.

Follow-up validation after removing the direct Starlette dependency:

```text
fastapi==0.120.4
starlette==0.49.3
```

Integration tests were not run locally; CI should validate broader REST/server behavior, especially real request/download flows around `FileResponse`, static assets, middleware, and background cleanup.

## Release notes

No release notes needed; this is dependency security maintenance.


stefannica (approved):

stefannica:
Did you test this @strickvl

pyproject.toml
```diff
@@ -79,7 +79,8 @@ local = [
 ]
 server = [
     "zenml[local,native-schedules]",  # Includes all the DB dependencies
-    "fastapi>=0.100,<=0.115.8",
+    "fastapi>=0.120.1,<0.121.0",
+    "starlette>=0.49.1,<0.50.0",
```
stefannica:
```suggestion
```

I don't think it's a good idea to manage nested dependencies. That is a rabbit hole I'd rather not go down into. We should stick to first-level dependencies.

I already checked that `fastapi>=0.120.1,<0.121.0` gets us starlette in the range `>=0.40.0,<0.51.0`, which includes the minimal version that we want: `0.49.1`.

stefannica (changes requested):
On second thought, I'd rather not have to manage nested dependencies.

bcdurak:
@stefannica @strickvl The next release will have a breaking title and this is a good chance to clean up and merge this PR (and all the other dependency audit related PRs). Could you guys take a look when you have a chance?

pyproject.toml
strickvl: Good point. I pushed this just now.

stefannica (approved):

PR #4833, opened 2026-05-12
Only show OSS dashboard warning for non-default projects

### LINKED ISSUE #4713: UX zenml project set default shows Pro warning before success message -- confusing message order

_No response_

### System Information

OS: Ubuntu 22.04
Python: 3.11+
ZenML: latest (open source)
Note: Issue is in CLI output ordering

### Issue Description

Running `zenml project set default` on open source ZenML 
shows this output:

  The ZenML projects feature is available only on ZenML Pro.
  Please visit https://zenml.io/pro to learn more.
  ✔ The active project has been set to default

Problems with this output:
→ Warning appears BEFORE the success message
→ New user reads warning first and thinks command failed
→ Then sees success message and is confused
→ "Did it work? Do I need Pro? What just happened?"
→ Creates a negative first impression unnecessarily

The command actually succeeded but the output makes 
it look like it failed. This is especially confusing 
for new users going through the quickstart guide.

Expected behavior — one of these options:
1. Show success message FIRST, then the Pro note
2. Make the Pro note clearly informational not a warning
3. Add clarification: "Note: advanced project features 
   require Pro, but basic usage works on open source"

Suggested better output:
  ✔ The active project has been set to default
  ℹ️  Note: advanced project features are available 
     on ZenML Pro. Visit https://zenml.io/pro

### Code Snippet [Optional]

```markdown
# Command run:
zenml init && zenml project set default

# Current output (confusing order):
The ZenML projects feature is available only on ZenML Pro.
Please visit https://zenml.io/pro to learn more.
✔ The active project has been set to default

# Expected output (clear order):
✔ The active project has been set to default
ℹ️  Note: advanced project features require ZenML Pro.
```

### Code of Conduct

qubeena07:
Fixes #4713

This addresses the second round of reviewer feedback on the OSS dashboard warning shown during project CLI operations.

Changes in this update:

Renamed the warning function to warn_if_project_not_visible_on_oss since the old name was misleading. The function now takes project_name as a required argument instead of optional, because passing nothing was pointless and the old early return on None was a smell.

Removed the warning from list projects and delete project commands. Listing does not change context so there is nothing to warn about, and deletion does not need a visibility notice.

Fixed a double warning in the describe command. When no project name is given, the code accesses client.active_project which already fires the warning internally via client.py. Calling the function again right after was printing the message twice.

Updated the warning text in client.py to be accurate. The old message implied the project feature requires Pro entirely, which is wrong. Pipelines run fine in OSS with any project. The limitation is only that the OSS dashboard will not show runs from non-default projects. The new message says exactly that.


bcdurak:
Hey @qubeena07, thank you for your contribution 🙂 As stated in the original issue, the warning message is also a bit misleading and needs to be adjusted. To handle it a bit better, here are a few notes:

- In the OSS version of ZenML, you can create/manage/delete projects through the CLI.
- While a new project is set, you can run your workflow without any problems.
- However, the ZenML OSS dashboard only works with the default project.

So, here is what I would recommend:

- Convert the order of the success and the warning message, which you have already done. 
- Additionally, only show the warning message, if someone tries to create/set/describe a project that is not the `default` project.

Hope that helps. Happy coding.

bcdurak (changes requested):
I have left some notes above.

qubeena07:
Hi @bcdurak, thanks for the feedback!

I've made both changes you suggested:

1. The success message already appeared before the warning (done in the previous commit), so that ordering is in place.
2. The warning now only shows when the project being operated on is not the default project. If someone runs `zenml project set default` or registers a project named `default`, no warning appears. The warning only fires for non-default projects, since that is where the OSS dashboard limitation actually applies.

Also updated the warning text to be more informative, explaining that pipelines will still run but won't be visible in the OSS dashboard, rather than just pointing to Pro.

Could you take a look when you get a chance? Happy to adjust anything.

src/zenml/cli/utils.py
```diff
@@ -2737,13 +2737,32 @@ def format_value(value: Any, indent_level: int = 0) -> str:
     console.print("  [dim]Deprovision and delete this deployment[/dim]")
 
 
-def check_zenml_pro_project_availability() -> None:
-    """Check if the ZenML Pro project feature is available."""
+def check_zenml_pro_project_availability(
```
bcdurak: I think the name here is a bit misleading. I would suggest renaming it to something like `warn_if_project_not_visible_on_oss`.

bcdurak: I would also change to input `project_name` to be a required value, as without a project name, it will just end up returning.

src/zenml/cli/project.py
```diff
@@ -81,6 +80,7 @@ def list_projects(
         row_generator=cli_utils.generate_project_row,
         active_id=active_project_id,
     )
+    check_zenml_pro_project_availability()
```
bcdurak: So, there is an important thing to consider here. If you check the client implementation, we have [a similar check and warning](https://github.com/zenml-io/zenml/blob/62d7ca8639ce56bf9b00a4c3c7891b8dd6b0e148/src/zenml/client.py#L1190) there. That means, every time you access this property and the active project is not the default one, it will show the warning. This is why you don't need this here.

bcdurak: Also, I would rephrase the warning in the client based on our discussion.

src/zenml/cli/project.py
```diff
@@ -228,3 +228,4 @@ def delete_project(project_name_or_id: str) -> None:
             )
         except Exception as e:
             cli_utils.exception(e)
+    check_zenml_pro_project_availability(project_name=project_name_or_id)
```
bcdurak: Since this is a deletion command, I would not show a warning here.

src/zenml/cli/project.py
```diff
@@ -192,13 +192,13 @@ def describe_project(project_name_or_id: Optional[str] = None) -> None:
...
         project_name_or_id: The name or ID of the project to set as active.
     """
-    check_zenml_pro_project_availability()
     client = Client()
     if not project_name_or_id:
         active_project = client.active_project
         cli_utils.print_pydantic_models(
             [active_project], exclude_columns=["created", "updated"]
         )
+        check_zenml_pro_project_availability(project_name=active_project.name)
```
bcdurak: The comment that I made about the `client.active_project` applies here as well. In this `if` clause, this would throw the warning once and then the client would throw a similar warning again.

bcdurak (changes requested):
Hey @qubeena07, thank you for the updates. I left some comments again.

bcdurak:
@qubeena07 I forgot to mention it previously. Can you also sign the CLA?

qubeena07:
@bcdurak  Yes, I signed the CLA. Thank you.

src/zenml/cli/utils.py
```diff
@@ -2737,13 +2737,30 @@ def format_value(value: Any, indent_level: int = 0) -> str:
...
     if not client.zen_store.get_store_info().is_pro_server():
         warning(
-            "The ZenML projects feature is available only on ZenML Pro. "
-            "Please visit https://zenml.io/pro to learn more."
+            f"The ZenML OSS dashboard only works with the default project "
+            f"('{DEFAULT_PROJECT_NAME}'). You are currently working with "
+            f"project '{project_name}'. Your pipelines will run, but you "
+            f"won't be able to view them in the OSS dashboard. "
+            f"To use multiple projects with full dashboard support, "
+            f"please visit https://zenml.io/pro to learn more."
```
bcdurak, line 2783:
```suggestion
            f"The ZenML OSS dashboard only works with the default project "
            f"('{DEFAULT_PROJECT_NAME}'). If you use any other projects "
            f"such as ('{project_name}'), your pipelines will run, but you "
            f"won't be able to view them in the OSS dashboard. "
            f"To use multiple projects with full dashboard support, "
            f"please visit https://zenml.io/pro to learn more."
```

This warning does not only show when they are actively using a project but also when they register it or when they describe (without activation). This is why I would rephrase this slightly.

bcdurak (commented):
@qubeena07 sorry to bother you with this again, but according to our CLA-assistant, it seems like you haven't signed it yet. Could you perhaps double check? Additionally, I left a small comment, aside from that, everything looks good.

qubeena07:
I have read the CLA Document and I hereby sign the CLA. 
@bcdurak my CLA assistant shows  "You have agreed to the CLA for zenml-io/zenml" but the bot still says CLA not signed.

qubeena07:
Hi! Thanks for the feedback. I have addressed the comment by updating the warning text to better reflect that the message appears at registration time and not just during active use. I also removed the redundant warning calls from the set and describe commands since client.active_project already handles that. The CLA email issue should be resolved now too. Please let me know if anything else needs to be changed!

bcdurak (changes requested):
@qubeena07 Thanks for the updates! Most of the feedback looks addressed now.

One small thing: I think there was a misunderstanding around `set` and `describe`. I didn’t mean to remove the warning from those commands entirely. The warning should still be shown when someone runs `zenml project set <non-default-project>` or `zenml project describe <non-default-project>`.

The duplicate-warning concern only applied to `zenml project describe` without an argument, because that path accesses `client.active_project`, which already emits the warning internally.

So I think the intended behavior should be:

- [X] `register <non-default-project>`: show the warning after the success message 
- [ ] `set <non-default-project>`: show the warning after the success message 
- [ ] `describe <non-default-project>`: show the warning 
- [ ] `describe` without an argument: don’t add an extra CLI warning, since `client.active_project` already handles it
- [X] `delete`: no warning needed

Could you add the warning back for the explicit `set` and `describe <project>` paths?

qubeena07:
Hi @bcdurak, thanks for clarifying! I have added the warning back to both `set` and `describe <project>`. The no-arg describe path is left without a manual call since `client.active_project` already handles it internally.

bcdurak (approved):
@qubeena07, looks good to me. Will merge this now. Thank you once again for your contribution 🚀

PR #4834, opened 2026-05-13
Fix CLI page info formatting

strickvl:
## Summary
- Replace literal Markdown-style backticks in the CLI pagination summary with Rich cyan markup.
- Preserve visual emphasis for the page/count values while ensuring no backticks are rendered in terminal output.
- Add a unit regression test for `print_page_info()` that captures rendered Rich console output.

## Root cause
`print_page_info()` used Markdown-style backticks around the page and item count, but `declare()` prints plain strings through Rich `console.print()`. Rich does not treat Markdown backticks as inline-code markup in this path, so the backticks were shown literally in commands like `zenml stack list`.

## Testing
- `uv run --extra local --with pytest --with pytest-mock pytest tests/unit/cli/test_utils.py -q`
- `uv run --with ruff bash scripts/format.sh --no-yamlfix src/zenml/cli/utils.py tests/unit/cli/test_utils.py`

### DIFF
--- a/src/zenml/cli/utils.py
+++ b/src/zenml/cli/utils.py
@@ -2754,8 +2754,8 @@ def print_page_info(page: "Page[Any]") -> None:
         page: The page object containing pagination information.
     """
     declare(
-        f"Page `({page.index}/{page.total_pages})`, "
-        f"`{page.total}` items found for the applied filters."
+        f"Page [cyan]({page.index}/{page.total_pages})[/cyan], "
+        f"[cyan]{page.total}[/cyan] items found for the applied filters."
     )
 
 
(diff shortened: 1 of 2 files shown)


src/zenml/cli/utils.py
```diff
@@ -2754,8 +2754,8 @@ def print_page_info(page: "Page[Any]") -> None:
         page: The page object containing pagination information.
     """
     declare(
-        f"Page `({page.index}/{page.total_pages})`, "
-        f"`{page.total}` items found for the applied filters."
+        f"Page ({page.index}/{page.total_pages}), "
+        f"{page.total} items found for the applied filters."
```
amitvikramraj:
```py
    declare(
        f"Page [cyan]({page.index}/{page.total_pages})[/cyan], "
        f"[cyan]{page.total}[/cyan] items found for the applied filters."
    )
```

amitvikramraj (approved):

PR #4838, opened 2026-05-15
Richer WanDB integration

### LINKED ISSUE #4825: Improve W&B run identity, metadata, and deep-linking

_No response_

### Feature Description

Make W&B run initialization richer without changing the integration shape: deterministic W&B run IDs from ZenML step-run IDs, resume="allow" for retries, better default group, job_type, config, ZenML model/pipeline/step metadata, and bidirectional links between ZenML and W&B.

Users get cleaner W&B projects, fewer duplicate runs on retry, better filtering/comparison, and easier navigation without writing extra W&B boilerplate.

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

Json-Andriopoulos:
## Describe changes

Improved the W&B experiment tracker with richer ZenML context and run control.

ZenML now adds pipeline and step metadata to W&B runs by default, groups runs by pipeline run, records richer W&B identifiers back on ZenML step metadata, and supports advanced settings such as custom groups, job types, run config, explicit or deterministic run IDs, resume behavior, and safe passthrough wandb.init kwargs. This makes W&B runs easier to filter, trace back to ZenML, and manage across retries.

## Pre-requisites

## Types of changes

### DIFF
--- a/docs/book/component-guide/experiment-trackers/wandb.md
+++ b/docs/book/component-guide/experiment-trackers/wandb.md
@@ -220,6 +220,8 @@ wandb_settings = WandbExperimentTrackerSettings(
 )
 ```
 
+ZenML truncates W&B tags to 64 characters and removes duplicates after truncation, so long tags with the same first 64 characters collapse into one W&B tag.
+
 When ZenML metadata is enabled, W&B config receives flat keys such as `zenml_pipeline_name`, `zenml_pipeline_run_name`, `zenml_pipeline_run_id`, `zenml_step_name`, `zenml_latest_step_run_id`, and `zenml_latest_step_run_version`. ZenML dashboard links are added when they can be derived from the active server. These values are written to W&B config, not W&B summary, so metric summaries stay focused on experiment results.
 
 To use W&B in a minimal mode without ZenML-added grouping, tags, or config, disable metadata:
--- a/docs/book/component-guide/experiment-trackers/wandb.md
+++ b/docs/book/component-guide/experiment-trackers/wandb.md
@@ -197,6 +197,17 @@ def my_step(
     ...
 ```
 
+ZenML disables W&B SDK console output by default with settings equivalent to `wandb.Settings(console="off", silent=True)`. This keeps W&B status and progress messages out of ZenML step logs. To show W&B SDK console output again, override these settings:
+
+```python
+wandb_settings = WandbExperimentTrackerSettings(
+    settings=wandb.Settings(
+        console="auto",
+        silent=False,
+    ),
+)
+```
+
 ZenML manages common `wandb.init(...)` fields for you. In addition to the W&B entity and project configured on the stack component, `WandbExperimentTrackerSettings` supports:
 
 ```python
--- a/docs/book/component-guide/experiment-trackers/wandb.md
+++ b/docs/book/component-guide/experiment-trackers/wandb.md
@@ -164,7 +164,7 @@ Or on the ZenML dashboard as metadata of a step that uses the tracker:
 Alternatively, you can see an overview of all experiment runs at https://wandb.ai/{ENTITY\_NAME}/{PROJECT\_NAME}/runs/.
 
 {% hint style="info" %}
-The naming convention of each Weights & Biases experiment run is `{pipeline_run_name}_{step_name}` (e.g. `wandb_example_pipeline-25_Apr_22-20_06_33_535737_tf_evaluator`) and each experiment run will be tagged with both `pipeline_name` and `pipeline_run_name`, which you can use to group and filter experiment runs.
+The naming convention of each Weights & Biases experiment run is `{pipeline_run_name}_{step_name}` (e.g. `wandb_example_pipeline-25_Apr_22-20_06_33_535737_tf_evaluator`). By default, ZenML groups W&B runs by the ZenML pipeline run name, adds human-readable tags (`zenml`, the pipeline name, and the pipeline run name), and stores durable ZenML identifiers in the W&B config.
 {% endhint %}
 
 #### Additional configuration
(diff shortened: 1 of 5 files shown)


docs/book/component-guide/experiment-trackers/wandb.md
```diff
@@ -197,6 +197,64 @@ def my_step(
...
+To use W&B in a minimal mode without ZenML-added grouping, tags, or config, disable metadata:
+
+```python
+wandb_settings = WandbExperimentTrackerSettings(
+    enable_zenml_metadata=False,
+    tags=["plain-wandb"],
+)
+```
+
+By default, ZenML leaves W&B run IDs unset (`run_id_strategy="wandb_generated"`), preserving W&B-generated IDs. If you want deterministic IDs, choose one of the opt-in strategies. For example, this configuration keeps retries for the same ZenML step invocation attached to the same W&B run:
```
bcdurak: Before digging into the code, I have tested almost everything that is listed here in the docs and everything seems to work seamlessly. The only thing which confused me a bit was the `run_id_strategy`.

First of all, it would be nice to mention that `wandb_generated`, `pipeline_step` and `step_run` **were the three options** and nothing more. The first thing I did after seeing this was to check the code if there was any more. Additionally, the naming sounds a bit off to me. "Pipeline step" and "step run" are not very clear imo, but I do not have a better option atm.

Also, when we initially implemented the WandB experiment tracker, we didn't have the retry logic and we use a combination of the pipeline run name and the step name as the run name in WandB. With the retry logic, we end up with run names that are identical if you set the run id strategy to `step_run`:

Image

We can probably fix this by using the `step_run_id` however this might mess up the `pipeline_step` strategy. 🤔

docs/book/component-guide/experiment-trackers/wandb.md
```diff
@@ -197,6 +197,64 @@ def my_step(
...
+Advanced users can attach logging to an existing W&B run, or pass unmanaged W&B initialization fields through `init_kwargs`:
+
+```python
+wandb_settings = WandbExperimentTrackerSettings(
+    run_id="existing-wandb-run-id",
+    init_kwargs={"notes": "Backfill evaluation metrics."},
+)
+```
+
+ZenML validates conflicts early:
```
bcdurak, line 276: Additionally, I experienced some weird behaviour with the logs. I am running a pipeline like this:

```python
import wandb

@step(experiment_tracker="wandb")
def train() -> int:
    wandb.log({"train_metric": 1})
    return 1


@step(experiment_tracker="wandb")
def evaluate(x: int) -> None:
    wandb.log({"eval_metric": x + 1})


@pipeline(enable_cache=False)
def grouped_wandb_pipeline():
    evaluate(train())


if __name__ == "__main__":
    grouped_wandb_pipeline()
```

The console output looks alright but the logs on the ZenML side are presented as errors and have weird artifacts like `mmmm`: 😄 

Image

src/zenml/integrations/wandb/flavors/wandb_experiment_tracker_flavor.py
```diff
@@ -42,17 +44,70 @@
...
+    run_id_strategy: Literal[
+        "wandb_generated", "step_run", "pipeline_step"
+    ] = Field(
+        "wandb_generated",
+        description=(
+            "Strategy used to derive Wandb run IDs. The default leaves IDs "
+            "to Wandb."
+        ),
+    )
+    resume: Optional[Literal["allow", "must", "never", "auto"]] = Field(
```
bcdurak, line 78: I have been testing different options here. If you set `resume` to `must`, it requires an ID of an existing run as well. Would be nice to add it as an early validation (at least the fact that people set the `run_id` along with `resume=must`).

src/zenml/integrations/wandb/experiment_trackers/wandb_experiment_tracker.py
```diff
@@ -103,14 +91,26 @@ def get_step_run_metadata(
...
+        run_project: Optional[str] = None
+        run_entity: Optional[str] = None
+        run_job_type: Optional[str] = None
 
         # Try to get the run name and URL from WandB directly
         current_wandb_run = wandb.run
         if current_wandb_run:
+            run_id = getattr(current_wandb_run, "id", None)
+            run_path = getattr(current_wandb_run, "path", None)
             run_url = current_wandb_run.get_url()
```
bcdurak: I am getting some warnings that `get_url` is deprecated. While we are at it, we can also replace that with `current_wandb_run.url` I think.

src/zenml/integrations/wandb/experiment_trackers/run_initialization.py
```diff
@@ -0,0 +1,206 @@
...
+from zenml.utils import dashboard_utils
+
+if TYPE_CHECKING:
+    from zenml.config.step_run_info import StepRunInfo
+
+
+logger = get_logger(__name__)
+
+
+def sanitize_tag(tag: str) -> str:
```
bcdurak, line 38: Not sure if this is an issue we can fix, but this can end up wielding two different tags into the same tag. Just wanted to mention it here.

src/zenml/integrations/wandb/experiment_trackers/run_initialization.py
```diff
@@ -0,0 +1,206 @@
...
+        tag: The tag to sanitize.
+
+    Returns:
+        The sanitized tag.
+    """
+    return tag[:64]
+
+
+@dataclass(frozen=True)
+class WandbInitialization:
```
bcdurak: Why do we need this class explicitly? The usage seems a bit redundant.

src/zenml/integrations/wandb/experiment_trackers/run_initialization.py
bcdurak: Oh, I see that we are deduplicating them already. Perhaps we can mention this behaviour in the docs?

bcdurak (changes requested):
Looks good so far. Leaving a few comments:

docs/book/component-guide/experiment-trackers/wandb.md
Json-Andriopoulos: Yeah, this is WanDB caused, when logging it appends logs to stderr even though these are info/debug logs. Also uses carriage returns/progress formatting hence the weird logs. I think this happened prior to the changes as well. Will make wandb console logs configurable (switched off by default) and document accordingly.

bcdurak: Can we also mention this in the docs? Some of our users set up alerting mechanisms with their log stores in case of error messages. Would be nice to warn them ahead of time.

bcdurak: Oh, I commented here before checking the changes as this seemed unresolved. I see that you already mention this in the docs, it would still be nice to tell them what they would see if they enable it again.

docs/book/component-guide/experiment-trackers/wandb.md
bcdurak: How about `reuse_on_retry` and `new_on_retry` as alternative names for the strategies?

bcdurak (commented):
Left a few more comments on the unresolved threads above, so I am sending this as a ping. Feel free to re-request the review.

bcdurak (approved):

PR #4839, opened 2026-05-16
fix(users): correct copy-paste bug in active status update

andesyteoss:
## Describe changes

Fix a copy-paste bug in `update_user()` where the "active status change" block incorrectly writes `is_admin` instead of `active` to `safe_user_update`, preventing admins from activating or deactivating user accounts via the API.

## Bug details

In `src/zenml/zen_server/routers/users_endpoints.py`, the `update_user` function constructs a `safe_user_update` object that only carries fields the caller is authorized to change. There are two consecutive blocks — one for `is_admin` changes (line ~543) and one for `active` changes (line ~561). Both blocks correctly validate permissions, but the `active` block contains a copy-paste error from the `is_admin` block:

```python
# Validate an active status change
if user_update.active is not None and user.active != user_update.active:
    # ... permission checks (these are correct) ...

    # Accept the admin status update          <-- wrong comment
    safe_user_update.is_admin = user_update.is_admin  # <-- BUG: should be .active
```

After all validation, `safe_user_update` is what gets persisted. Because `safe_user_update.active` is never set, the user's active status is never actually updated — the request silently succeeds but changes `is_admin` a second time (to the same value it was already set to in the block above, so no visible side effect there).

## Security impact

This bug means **admin users cannot deactivate compromised or offboarded accounts** through the API. If an account needs to be disabled (e.g., a team member leaves, credentials are compromised), the `PUT /api/v1/users/{user_id}` endpoint with `{"active": false}` will appear to succeed but the account remains active and can continue to authenticate.

While this is not an authorization bypass in the traditional sense (no unauthorized access is granted), the inability to deactivate accounts is a meaningful security gap — it removes a key administrative control for incident response.

## Fix

The fix is a two-line change:
- `safe_user_update.is_admin = user_update.is_admin` → `safe_user_update.active = user_update.active`
- Updated the comment from "admin status" to "active status"

## Testing

- Verified the fix applies cleanly on `develop`.
- Manually traced the data flow: `user_update.active` is now correctly propagated to `safe_user_update.active`, which is the object passed to the store's `update_user()` method.
- Confirmed that the `is_admin` block (lines 543–558) remains unchanged and correct.
- Confirmed no other references to `safe_user_update.active` exist outside this block, so this is the only place the field needs to be set.

## Proof of Concept

To reproduce the bug on an unpatched instance:

```python
# As an admin user, attempt to deactivate another user
import requests

ZENML_SERVER = "http://localhost:8080"
ADMIN_TOKEN = "<admin-api-token>"
TARGET_USER_ID = "<user-id-to-deactivate>"

# Send deactivation request
resp = requests.put(
    f"{ZENML_SERVER}/api/v1/users/{TARGET_USER_ID}",
    headers={"Authorization": f"Bearer {ADMIN_TOKEN}"},
    json={"active": False},
)
print(resp.status_code)  # 200 — appears to succeed

# Verify the user is still active
resp = requests.get(
    f"{ZENML_SERVER}/api/v1/users/{TARGET_USER_ID}",
    headers={"Authorization": f"Bearer {ADMIN_TOKEN}"},
)
print(resp.json()["active"])  # True — deactivation silently failed
```

The `update_user` function is defined at line ~440 of `users_endpoints.py` and handles `PUT /api/v1/users/{user_name_or_id}`.

## Adversarial review

Before submitting, we considered whether this bug might be unreachable or mitigated elsewhere. The `active` field validation block is only entered when `user_update.active is not None and user.active != user_update.active` — so the buggy code does execute when an admin tries to change a user's active status. There is no other code path in the endpoint or in the SQL store layer that separately handles active status updates. The bug is straightforward and reproducible.

## Pre-requisites

  - [x] N/A — bug fix only, no API or feature changes.

## Types of changes


andesyteoss:
I've signed the CLA — could you please [recheck](https://cla-assistant.io/check/zenml-io/zenml?pullRequest=4839) the status? Thanks!

schustmi (approved):

stefannica (approved):

andesyteoss:
Appreciate the review.

PR #4844, opened 2026-05-18
Support python 3.14

schustmi:
## Describe changes
Adds support for python 3.14.

Breaking changes:
- SQLModel bumped from version `0.18.0` -> `0.38.0`. This only affects environments in which the `local` or `server` extras are installed.

## Pre-requisites

## Types of changes


pyproject.toml
```diff
@@ -87,7 +88,7 @@ server = [
     "python-multipart~=0.0.9",
     "pyjwt[crypto]==2.7.*",
     "fastapi-utils",
-    "orjson~=3.10.0",
+    "orjson>=3.10.0,<4.0.0",
```
bcdurak: Only a small nitpick here. I am not a big fan of relaxed dependencies like these. As far as I can see the latest version that they have is `3.11.9`. If they release `3.12.0` with any problematic breaking changes, we can not make sure if everything will work. I would suggest doing something like `>=3.10.0,<3.12.0`, but I will leave it up to you.

bcdurak (approved):
It makes me a bit uneasy to modify existing migration scripts but I couldn't find a scenario which would fail. Considering all the other changes, the sooner we merge this the better. I left a small comment but other than that things look good.

pyproject.toml
schustmi: If they follow semantic versioning correctly, this should be a safe pin because anything lower than 4.0 is a non-breaking change. I'll still lower the range to what you suggested just to be safe.

PR #4845, opened 2026-05-19
Better string parsing during interactive wait condition resolving

schustmi:
## Describe changes
When resolving a wait condition with a string schema using the interactive terminal input, users previously had to input a quoted string (e.g. `"my answer"`, because a raw string would fail the JSON parsing. This PR detects that and uses the raw string as answer in case the schema allows it.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/pipeline.py
+++ b/src/zenml/cli/pipeline.py
@@ -41,11 +41,13 @@ from zenml.models import (
     PipelineRunUpdate,
     PipelineSnapshotFilter,
     RunWaitConditionFilter,
+    RunWaitConditionResponse,
     ScheduleFilter,
 )
 from zenml.pipelines.pipeline_definition import Pipeline
 from zenml.utils import run_utils, source_utils, uuid_utils
 from zenml.utils.dashboard_utils import get_deployment_url
+from zenml.utils.json_utils import parse_value_for_schema
 from zenml.utils.yaml_utils import write_yaml
 
 logger = get_logger(__name__)
@@ -1019,7 +1021,7 @@ def _interactive_resolve_wait_conditions(
     skipped_condition_ids: set[UUID] = set()
 
     def _resolve_single_condition_interactively(
-        condition: Any,
+        condition: "RunWaitConditionResponse",
         idx: int,
         total: int,
     ) -> str:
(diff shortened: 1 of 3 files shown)


bcdurak (changes requested):
An issue that I found with this is, when you are waiting for a string input and you provide a value like `123` or `true`, this results into:


```
Expected JSON schema:
{
  "type": "string"
}
Press Enter on an empty line to submit null.
Press Ctrl+C to abort.
> 123
Failed to resolve: Value `123` does not match expected schema.
> true
Failed to resolve: Value `True` does not match expected schema.
> null
Failed to resolve: Value `None` does not match expected schema.
```

Also an additional question, I think this only affects the interactive terminal input. Should we also handle it the same way when people try to resolve through the CLI?

schustmi:
@bcdurak Yeah not sure about it, do you think if the schema is string I should just always treat the input as a string? All of those were also broken before, but I guess I can fix it also as part of this.

I'll see if those also apply to the CLI.

schustmi:
@bcdurak Actually for all your examples, I can't really solve them easily without doing actual schema validation. E.g. if you provide an int, how should I treat it? If the schema is `str`, I should instead convert to a string. If the schema is `Union[int, str]` I should keep it. No way to do that unless I actually have a way of doing schema validation, which I don't.

schustmi:
The only thing I could check potentially: If the schema is purely a `str` (no optional, no union), then i can always use the raw value. WDYT?

bcdurak:
@schustmi Yeah, it is a bit of a tricky one. The thing is, previously there was also only one way to define a string. If you would do `"test"` it would work but `test` wouldn't. Now that it does, it makes it a bit confusing how to treat `123` if the schema is a string.

As you mentioned, without a schema validation, it is definitely not easy. But I like your last suggestion. I think it would cover most of the use-cases.

schustmi:
@bcdurak Implemented something reasonable without actually doing a full validation I think

bcdurak:
@schustmi I will take a look as soon as possible. In the meanwhile, there are some linting issues just so you know.

bcdurak (approved):

PR #4852, opened 2026-05-21
Fix deployer run context lifecycle

### LINKED ISSUE #4846: K8s deployer: `RunContext` singleton torn down per-request by stack orchestrator, causing `RuntimeError: Run context not initialized` under concurrency
## Summary

When a pipeline is served by the Kubernetes deployer and the **active stack** uses a non-local orchestrator (`kubernetes` in our case), the dict returned by the pipeline's `on_init` hook is **not** the one read by the step at request time. Under any meaningful concurrency, `get_step_context().pipeline_state` raises `RuntimeError: Run context not initialized` for the majority of requests.

The dict is loaded correctly into `RunContext` at pod boot, but it is destroyed and re-created for every `/invoke` call by the step runner using the **stack's** orchestrator (not the deployer's `SharedLocalOrchestrator`). Because `RunContext` is a process-wide singleton, concurrent requests race on it: one request's `RunContext._clear()` runs while another request is mid-step, and that other request crashes.

In a 100-call load test at 16 concurrent in-flight requests, only ~5% of calls succeed; the rest fail. A single-call smoke test always succeeds (it wins the race trivially), which is why this is easy to miss in initial validation.

## Versions

- ZenML server: 0.94.3
- ZenML client / deployer pod: 0.94.2
- Stack: `kubernetes` deployer + `kubernetes` orchestrator (RKE2)
- Python: 3.12
- Confirmed unchanged on `develop` at the relevant code paths (see "Confirmed on develop" below).

## Reproduction

Minimal pipeline:

```python
from zenml import pipeline, step, get_step_context
from zenml.integrations.kubernetes.flavors.kubernetes_deployer_flavor import (
    KubernetesDeployerSettings,
)


def init_state() -> dict:
    return {"value": 42}


def cleanup_state() -> None:
    pass


@step
def read_state() -> int:
    return get_step_context().pipeline_state["value"]


@pipeline(
    name="repro-shared-state",
    on_init=init_state,
    on_cleanup=cleanup_state,
    settings={"deployer.kubernetes": KubernetesDeployerSettings(service_port=8000)},
)
def repro_pipeline() -> int:
    return read_state()


repro_pipeline.deploy(deployment_name="repro-shared-state")
```

Driver, called from inside the cluster:

```python
import concurrent.futures, requests
URL = "http://zenml-<deployment-id>.zenml.svc.cluster.local:8000/invoke"

def call(_):
    body = requests.post(URL, json={"parameters": {}}, timeout=60).json()
    return body.get("success", False), body.get("error")

with concurrent.futures.ThreadPoolExecutor(16) as pool:
    results = list(pool.map(call, range(100)))

errors = [e for ok, e in results if not ok]
print(f"errors: {len(errors)}/100")
print(errors[0] if errors else "")
```

Expected: `errors: 0/100`.
Observed: `errors: ~95/100`, each carrying:

```
RuntimeError: Run context not initialized. The run state is only available in the context of a running pipeline.
  File ".../zenml/steps/step_context.py", line 232, in pipeline_state
    return get_or_create_run_context().state
  File ".../zenml/steps/step_context.py", line 97, in state
    raise RuntimeError(...)
```

Pod log analysis (single replica, no restarts, 10-minute uptime, 100 requests):

| Marker | Count |
|---|---|
| Pod-boot init from `PipelineDeploymentService.initialize` | 1 |
| `Executing the pipeline's init hook...` total | 97 |
| User-side init log markers (model load, DB connect, etc.) | 97 |
| `POST /invoke` requests | 100 |
| `Run context not initialized` exceptions logged | 114 |
| Step entrypoint reached (first user log line) | 62 |

So init is running per request, **in addition to** the boot-time init. Some requests fail before the step entrypoint is even reached (we observed downstream resource contention during the per-request init itself, e.g. ArangoDB rejecting bursts of new connections), so the failure surface is broader than just \"step reads empty state\".

## Root cause

`RunContext` (`src/zenml/steps/step_context.py:78`) uses `SingletonMetaClass`, which holds a single instance per process via a class attribute (`src/zenml/utils/singleton.py:49-66`). It is shared across all FastAPI worker threads. (`ThreadLocalSingleton` exists in the same module but is not used here.)

The deployer's intended design (`src/zenml/deployers/server/service.py`):

1. At boot, `PipelineDeploymentService.initialize()` calls `BaseOrchestrator.run_init_hook(self.snapshot)` to populate the singleton once (`service.py:328`).
2. Per request, `_execute_with_orchestrator` instantiates a `SharedLocalOrchestrator` whose `run_init_hook` / `run_cleanup_hook` are no-ops (`service.py:76-104`), so the boot-time singleton survives across calls.

The break is in `src/zenml/orchestrators/step_runner.py:223-230` and `352-358`:

```python
if (
    pipeline_run.snapshot
    and self._stack.orchestrator.run_init_cleanup_at_step_level
):
    self._stack.orchestrator.run_init_hook(snapshot=pipeline_run.snapshot)
# ... step body ...
# finally:
if (
    pipeline_run.snapshot
    and self._stack.orchestrator.run_init_cleanup_at_step_level
):
    self._stack.orchestrator.run_cleanup_hook(snapshot=pipeline_run.snapshot)
```

`self._stack` is set in `StepLauncher.__init__` as `Stack.from_model(snapshot.stack)` (`step_launcher.py:145`) — i.e. the **snapshot's** stack. The `SharedLocalOrchestrator` the deployer instantiates inline (`service.py:565-574`) is never part of any stack and is never seen by `step_runner`. So `self._stack.orchestrator` resolves to the user-configured K8s orchestrator, which:

- inherits `run_init_cleanup_at_step_level = True` from `BaseOrchestrator` (`base_orchestrator.py:708-724` — only `LocalOrchestrator` overrides this to `False`),
- inherits the base implementations of `run_init_hook` / `run_cleanup_hook` (`base_orchestrator.py:727-795`).

Per request:

1. step_runner calls the stack orchestrator's `run_init_hook` → base impl runs, repopulates the singleton.
2. step runs, reads `pipeline_state` → works **only if no other concurrent request has triggered teardown yet**.
3. step_runner's `finally` calls `run_cleanup_hook` → `RunContext._clear()` sets `__singleton_instance = None`.

Under concurrency, requests overlap on the shared singleton: request A's `_clear()` runs while request B's step is mid-flight, and request B then hits `get_or_create_run_context()` which constructs a fresh, uninitialized `RunContext` and `.state` raises.

The boot-time init dict is effectively never read by step code in this configuration — it's replaced by per-request init dicts that are then cleared.

## Expected behavior

When a pipeline is served by a deployer:

- The dict returned by `on_init` should be readable from `get_step_context().pipeline_state` in every step invocation, on every concurrent request, for the lifetime of the pod.
- Per-request init/cleanup must not be triggered by the stack orchestrator's `run_init_cleanup_at_step_level=True` while the deployer is the active execution path — the deployer's `SharedLocalOrchestrator` should own that lifecycle.

## Possible fixes (non-prescriptive)

1. **Have the step runner consult the executing orchestrator**, not `self._stack.orchestrator`. Under the deployer's runtime, that would be `SharedLocalOrchestrator`, whose overrides correctly no-op step-level init/cleanup.
2. **Make `RunContext` reentrant** — e.g. ref-counted `initialize` / `_clear` so concurrent step executions on the same process don't race on teardown.
3. **Have `BasePipelineDeploymentService` flip the stack orchestrator's `run_init_cleanup_at_step_level` to `False`** while it serves (or wrap it so the step runner sees False).
4. **Move per-request state to a per-request scope** (thread-local or `contextvars`) and expose long-lived deployer state via a dedicated accessor like `get_deployment_state()`.

## Confirmed on `develop`

Fetched `src/zenml/steps/step_context.py`, `src/zenml/orchestrators/base_orchestrator.py`, `src/zenml/orchestrators/step_runner.py`, and `src/zenml/deployers/server/service.py` from `develop` via the GitHub contents API and diffed against 0.94.2 — byte-identical at the relevant code paths. The 0.94.3 and 0.94.4 release notes do not mention any change to `pipeline_state` / `RunContext` / init-hook plumbing. A search across issues and merged PRs for \"Run context not initialized\", \"pipeline_state\", \"RunContext\", and \"singleton run context\" returns no results — this appears to be unreported.

## Workaround

Storing the model and database connection as module-level globals in the file containing the init hook, and reading them directly from the step instead of via `pipeline_state`, bypasses `RunContext` and is robust against the singleton race. Not pretty, but works without ZenML changes.

strickvl:
## What this does

Fixes #4846 by making run-context cleanup conditional on whether the matching init call actually initialized `RunContext`.

The concrete change is:

- `BaseOrchestrator.run_init_hook(...)` now returns:
  - `True` when it initializes `RunContext`
  - `False` when `RunContext` was already initialized
- Step/run cleanup now skips only on exact `False`.
  - `True` still runs cleanup.
  - Legacy custom overrides returning `None` still run cleanup, preserving existing behavior.
- The earlier `lifecycle_orchestrator` threading approach was removed from `BaseOrchestrator.run_step(...)`, `launch_step(...)`, `StepLauncher`, `StepRunner`, and dynamic runner launch calls.
- The same init-result cleanup rule is applied to nearby init/cleanup pairs in `StepRunner`, `LocalOrchestrator`, `DynamicPipelineRunner`, and `PipelineDeploymentService`.
- Init-failure cleanup now lives inside `BaseOrchestrator.run_init_hook(...)`: if init creates a partial `RunContext` and then raises, the paired cleanup hook path runs and the partial context is cleared before the original init failure is re-raised.
- Deployment service cleanup skips if service startup never successfully reached the init ownership point, and cleanup is idempotent per service instance.
- Dynamic child placeholder runs now preserve deployment trigger metadata, so deployment-triggered child runs keep using the deployment-owned `RunContext` lifecycle.

This prevents Kubernetes deployer `/invoke` requests from clearing service-level `RunContext` state that was initialized when the deployment pod started.

## Why

A deployed pipeline request enters through `SharedLocalOrchestrator`, but the deployment snapshot still preserves the original stack. If that stack contains the Kubernetes orchestrator, inline request execution can reach step-level cleanup behavior that calls `RunContext._clear()`.

Under concurrent `/invoke` calls, one request could clear the process-wide `RunContext` while another request was still reading `get_step_context().pipeline_state`, causing:

```text
RuntimeError: Run context not initialized. The run state is only available in the context of a running pipeline.
```

With this change, request-level init sees that the deployment service already initialized `RunContext`, returns `False`, and the matching request-level cleanup is skipped.

## Notes

- This intentionally preserves compatibility for custom orchestrators whose `run_init_hook(...)` overrides still return `None`.
- This addresses review feedback by keeping cleanup ownership tied to the `run_init_hook(...)` return value instead of adding extra caller-side lifecycle bookkeeping.
- Dynamic deployment runs and their child runs preserve deployment-trigger metadata so they do not accidentally take over `RunContext` lifecycle ownership from the deployment service.

## Verification

```bash
uv run ruff format --check src/zenml/deployers/server/service.py src/zenml/execution/pipeline/dynamic/runner.py src/zenml/orchestrators/base_orchestrator.py src/zenml/orchestrators/local/local_orchestrator.py src/zenml/orchestrators/step_runner.py tests/unit/deployers/server/test_service.py tests/unit/execution/pipeline/dynamic/test_runner.py tests/unit/orchestrators/local/test_local_orchestrator.py tests/unit/orchestrators/test_base_orchestrator.py tests/unit/orchestrators/test_step_runner.py
uv run ruff check src/zenml/deployers/server/service.py src/zenml/execution/pipeline/dynamic/runner.py src/zenml/orchestrators/base_orchestrator.py src/zenml/orchestrators/local/local_orchestrator.py src/zenml/orchestrators/step_runner.py
uv run pytest tests/unit/orchestrators/test_base_orchestrator.py tests/unit/deployers/server/test_service.py tests/unit/execution/pipeline/dynamic/test_runner.py tests/unit/orchestrators/local/test_local_orchestrator.py tests/unit/orchestrators/test_step_runner.py
```

Latest local focused test run: `50 passed`.


safoinme (approved):

schustmi (commented):
@strickvl Can't we do this a little simpler? As in: The deployment should create the run state when it gets started. Each pipeline run will then see that the run context is already initialized, and will therefore ignore not run the hook again?

That would avoid passing around a completely unrelated object through multiple classes.

More specifically for the case described here:
- `orchestrator.run_init_hook()` returns a bool whether it actually initialized something.
- we only run the teardown if it returned True

src/zenml/deployers/server/service.py
```diff
@@ -299,6 +303,22 @@ class PipelineInvokeResponse(BaseDeploymentInvocationResponse):
...
+        Args:
+            app_runner: The deployment application runner used with this service.
+            **kwargs: Additional keyword arguments for the deployment service.
+        """
+        super().__init__(app_runner=app_runner, **kwargs)
+        self._run_context_initialized_by_service: Optional[bool] = None
+        self._init_hook_started_by_service = False
+        self._init_hook_completed_by_service = False
+        self._run_context_was_initialized_before_init_hook = False
+        self._cleanup_hook_ran_by_service = False
```
schustmi: Nope. For sure we're not modifying adding 5 different attributes for this, in multiple places. 100% not.

schustmi (changes requested):

bcdurak:
Hey @strickvl, as we are aiming for a release this week, I was wondering if we can find some time to get this in as well?

PR #4856, opened 2026-05-22
Fix GKE private cluster access via DNS control plane endpoint in GCP service connector

stefannica:
## Summary

Fixes GCP service connector failures for **private GKE clusters** that use Google’s **DNS-based control plane endpoint**. ZenML now connects the same way as `gcloud container clusters get-credentials --dns-endpoint`, instead of always using the cluster’s private IP.

## Problem

When you use a GCP service connector to get Kubernetes credentials for a GKE cluster, ZenML must reach the cluster API to verify access and to hand runners a usable `server` URL.

For many **private** clusters (common in newer GCP setups), `cluster.endpoint` is a **private IP** only reachable from inside the VPC. The ZenML server (for example ZenML Cloud) is usually **outside** that network, so credential requests **time out** even when:

- GCP credentials are valid
- The cluster is healthy
- `kubectl` works locally with `--dns-endpoint`

Those local flows use the cluster’s **DNS hostname** (`*.gke.goog`), not the private IP. The connector did not.

## What changes for users

When issuing Kubernetes client credentials for GKE, the connector now:

1. **Uses the DNS control plane hostname** when GKE exposes `control_plane_endpoints_config.dns_endpoint_config.endpoint` (same FQDN as in the GCP API and `gcloud --dns-endpoint`).
2. **Falls back** to the previous behavior (`cluster.endpoint` + cluster CA) when no DNS endpoint is configured (public clusters and older setups).

You should be able to verify connectors, fetch connector clients, and run stack components against private GKE clusters that rely on the DNS endpoint—without putting the ZenML server on the cluster’s private network.

## Technical notes

- DNS path: `https://<fqdn>`, no cluster CA (system/Google PKI, aligned with DNS endpoint access).
- IP path: unchanged (`https://<cluster.endpoint>` + `master_auth.cluster_ca_certificate`).
- Unit tests added for DNS preference, IP fallback, and missing endpoint.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
+++ b/src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
@@ -2237,7 +2237,7 @@ class GCPServiceConnector(ServiceConnector):
             )
             if dns_endpoint_config:
                 dns_host = dns_endpoint_config.endpoint.strip()
-                if dns_host and dns_endpoint_config.allow_external_traffic:
+                if dns_host:
                     logger.debug(
                         "Using GKE DNS control plane endpoint for Kubernetes API "
                         "access: %s",
--- a/src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
+++ b/src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
@@ -46,6 +46,7 @@ from google.auth._default import (
 )
 from google.auth.transport.requests import Request
 from google.cloud import artifactregistry_v1, container_v1, storage
+from google.cloud.container_v1.types import Cluster as GKECluster
 from google.cloud.location import locations_pb2
 from google.oauth2 import credentials as gcp_oauth2_credentials
 from google.oauth2 import service_account as gcp_service_account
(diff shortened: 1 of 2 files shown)


src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
```diff
@@ -2201,6 +2202,56 @@ def _verify(
...
+            A tuple of ``(server_url, certificate_authority)``. The
+            certificate authority is the base64-encoded cluster CA for
+            IP-based endpoints, or ``None`` for DNS-based endpoints (clients
+            should use the system CA bundle).
+
+        Raises:
+            AuthorizationException: If no Kubernetes API endpoint is available.
+        """
+        dns_host = (
+            cluster.control_plane_endpoints_config.dns_endpoint_config.endpoint.strip()
```
bcdurak: With the GCP integration, we allow: `google-cloud-container>=2.21.0`. However, `cluster.control_plane_endpoints_config` is only accessible starting from `2.52.0`. With earlier versions, this would eventually lead to an error. I think we need to handle this gracefully.

src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
```diff
@@ -2201,6 +2202,56 @@ def _verify(
...
+        )
+        if dns_host:
```
bcdurak: Here, as far as I can see, we can extend the check a little bit. When there is a DNS hostname, we opt to use it. However, if the `cluster.control_plane_endpoints_config.dns_endpoint_config.allow_external_traffic` is set to False, this option will not work. Instead, in cases such as these, we can still fallback to the old method.

bcdurak (changes requested):

bcdurak:
Also, just as a heads-up, the linting is failing due to some formatting issues.

src/zenml/integrations/gcp/service_connectors/gcp_service_connector.py
bcdurak: As the fix is applied, resolving this.

bcdurak (approved):

PR #4858, opened 2026-05-26
fix: scoped sqlalchemy otel instrumentation to server sql engine

### LINKED ISSUE #4863: Fix sqlalchemy otel instrumentation in zenml-server

amitvikramraj:
## Describe changes

* This updates the OSS server OpenTelemetry SQLAlchemy setup to instrument the concrete ZenML SQL store engine instead of relying on global SQLAlchemy instrumentation. Global instrumentation was only producing `connect` spans in local workspace traces. Targeting the initialized `SqlZenStore.engine` adds query execution spans such as `SELECT`, `INSERT`, and `UPDATE` with `db.statement`, making database activity visible in traces.

* Tracking the library instrumentors during otel configure and cleaning them on shutdown

* Added stable OpenTelemetry resource attributes (`service.version`, `deployment.environment.name`, and `zenml.deployment.type`) so traces, metrics, and logs can be filtered/grouped by ZenML version and deployment context without adding high-cardinality per-span attributes.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/otel.py
+++ b/src/zenml/zen_server/otel.py
@@ -54,6 +54,7 @@ _otel_providers: list[Any] = []
 # list of callbacks to uninstrument the libraries when we shutdown OTel.
 _otel_uninstrument_callbacks: list[Callable[[], None]] = []
 
+
 def configure_otel(app: "FastAPI") -> None:
     """Set up OpenTelemetry tracing, metrics, and log export.
 
@@ -146,7 +147,6 @@ def shutdown_otel() -> None:
         and not _otel_uninstrument_callbacks
     ):
         return
-
 
     # Uninstrument the libraries.
     while _otel_uninstrument_callbacks:
--- a/src/zenml/zen_server/otel.py
+++ b/src/zenml/zen_server/otel.py
@@ -43,6 +43,8 @@ if TYPE_CHECKING:
     from fastapi import FastAPI
     from opentelemetry.sdk.resources import Resource
 
+    from zenml.config.server_config import ServerConfiguration
+
 logger = get_logger(__name__)
 
 _otel_configured = False
@@ -95,7 +97,7 @@ def configure_otel(app: "FastAPI") -> None:
     try:
         from opentelemetry.sdk.resources import Resource
 
-        resource = Resource.create({"service.name": config.otel_service_name})
+        resource = Resource.create(_get_resource_attributes(config))
     except ImportError:
         logger.debug(
             "OpenTelemetry SDK packages not installed — skipping "
@@ -382,3 +384,14 @@ def _instrument_libraries(app: "FastAPI") -> None:
         pass
     except Exception:
         logger.exception("Failed to instrument SQLAlchemy with OpenTelemetry.")
+
+
+def _get_resource_attributes(config: "ServerConfiguration") -> dict[str, str]:
+    """Get stable resource attributes shared by all OTel signals."""
+    import zenml
+
+    return {
+        "service.name": config.otel_service_name,
+        "service.version": zenml.__version__,
+        "deployment.environment.name": str(config.deployment_type),
+    }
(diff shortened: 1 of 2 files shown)


src/zenml/zen_server/otel.py
```diff
@@ -331,10 +331,23 @@ def _instrument_libraries(app: "FastAPI") -> None:
             SQLAlchemyInstrumentor,
         )
 
-        SQLAlchemyInstrumentor().instrument()
+        from zenml.zen_server.utils import initialize_zen_store, zen_store
+
+        # if the zen store is not initialized, initialize it. Else, use the existing store.
+        # The zen-store is initialized during the server startup, but otel is configured
```
Json-Andriopoulos: Not sure I like the out-of-order initialization of the store. Why not configure otel on the server startup to respect the initialization order?

amitvikramraj: I was thinking about the order as well. So,

* Currently, OTel is intentionally configured to run before the app startup, so that it can capture startup logs and other telemetry(if present).
* And then I thought, ideally, the SQL Zen store initialisation should be idempotent, so probably it wouldn't matter much if we initialise it before the server starts.
* Although if we are okay to lose tracking some of the initial server logs, then we can move the `configure_otel` function inside the app startup phase.

amitvikramraj:
image
image

Sample trace map showing not only the HTTP request but also which SQL query ran and how long it took.

cc: @stefannica

src/zenml/zen_server/otel.py
stefannica: I agree with @Json-Andriopoulos, this should feel wrong to you. You're implementing some low-level concerns such as logging, but now you also took over initializing the most important design component of all as part of this module. This doesn't belong here.

The initialization of both OTEL and Zen store is orchestrated in the FastAPI startup hook, correct? Perhaps that's where you should link them together: first initialize OTEL, then initialize the Zen store and plug the resulting store instance into the SQLAlchemy instrumentor after the fact. Is that possible ?

amitvikramraj: This did feel wrong, which is why I simply opened the PR without prior discussion to get your guys' comments. Instrumenting SQL store separately is the better way.

amitvikramraj: @Json-Andriopoulos @stefannica I addressed the changes.

src/zenml/zen_server/otel.py
```diff
@@ -135,15 +142,33 @@ def shutdown_otel() -> None:
...
+    if (
+        not _otel_configured
+        and not _otel_log_handler
+        and not _otel_providers
+        and not _otel_uninstrument_callbacks
+    ):
         return
 
+    # Uninstrument the libraries.
+    while _otel_uninstrument_callbacks:
```
Json-Andriopoulos: minor: while `any condition` really freaks me out, no matter how well written the while loop is or how tested.  When possible, always opt out for `for` loops - much safer. For instance:

```
for callback in _otel_uninstrument_callbacks:
    try:
        callback()
    except Exception as exc:
        ....

_otel_uninstrument_callbacks = []   # set empty after you are done
```

src/zenml/zen_server/otel.py
```diff
@@ -318,23 +375,23 @@ def _instrument_libraries(app: "FastAPI") -> None:
...
-        )
 
-        SQLAlchemyInstrumentor().instrument()
-    except ImportError:
-        logger.debug(
-            "OpenTelemetry SQLAlchemy instrumentation package not installed. "
-            "Install `opentelemetry-instrumentation-sqlalchemy`."
-        )
-        pass
+def _get_resource_attributes(config: "ServerConfiguration") -> dict[str, str]:
```
Json-Andriopoulos: This method feels a bit unnecessary, it is just a dict builder with one usage.

Json-Andriopoulos (approved):
LGTM. A few minor comments, please address them before merging @amitvikramraj

src/zenml/zen_server/otel.py
amitvikramraj: I'll make the changes but May I ask the reasoning behind this, to understand a bit more?

src/zenml/zen_server/otel.py
amitvikramraj: Added this to make it look a bit cleaner. I'll put it inplace.

PR #4860, opened 2026-05-26
Run statistics endpoint

schustmi:
Add a `POST /api/v1/runs/statistics` endpoint that returns grouped aggregations over pipeline runs.

- **Group by**: status, pipeline, snapshot, source snapshot, stack, trigger, user, time bucket (hour/day/week/month), metadata value, tag. Multiple groupings allowed.
- **Metrics**: `avg`/`sum`/`min`/`max` over `duration`, `step_count` (retries collapsed), `cached_step_count`, `output_artifact_count` or a numeric metadata value. Per-group `run_count` is always returned.
- **Limitation**: metadata metrics only aggregate entries stored as `int` or `float`. Other types are skipped. Tag grouping is multi-bucket: a run with N tags contributes to N buckets, so sum of `run_count` across tag buckets can exceed the number of filtered runs.

### DIFF
--- a/src/zenml/models/__init__.py
+++ b/src/zenml/models/__init__.py
@@ -17,61 +17,60 @@
 
 # V2 Base
 from zenml.models.v2.base.base import (
-    BaseZenModel,
+    BaseDatedResponseBody,
+    BaseIdentifiedResponse,
     BaseRequest,
     BaseResponse,
-    BaseUpdate,
-    BaseIdentifiedResponse,
     BaseResponseBody,
     BaseResponseMetadata,
     BaseResponseResources,
-    BaseDatedResponseBody,
+    BaseUpdate,
+    BaseZenModel,
 )
+from zenml.models.v2.base.filter import (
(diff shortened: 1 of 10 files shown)


bcdurak (approved):
I haven't had the time to fully test this over the workspace, but the code looks good and I will approve this. 

However, I still have one nitpicky thing. If we are dealing with a time grouping, I think we give the results of the earliest entries. I would propose doing the opposite. But, I'll leave that up to you.

PS: Thanks for ordering the `models/__init__.py` 😄

schustmi:
> I haven't had the time to fully test this over the workspace, but the code looks good and I will approve this.
> 
> However, I still have one nitpicky thing. If we are dealing with a time grouping, I think we give the results of the earliest entries. I would propose doing the opposite. But, I'll leave that up to you.
> 
> PS: Thanks for ordering the `models/__init__.py` 😄

Made the change with the ordering :)
I think ruff or something in our CI script did it actually :)

PR #4866, opened 2026-05-27
Add Sandbox stack component (core abstraction)

htahir1:
## Summary

Adds a `Sandbox` stack component for running untrusted (typically LLM-generated) code in an isolated environment, plus the built-in `local` flavor that runs commands as subprocesses. Cloud-backed flavors (Modal, k8s_agent_sandbox) follow in stacked PRs #4867 and #4870.

Three abstractions:
- `BaseSandbox`: the stack component. Call `Client().active_stack.sandbox.create_session()` from inside a step.
- `SandboxSession`: a live, bounded interaction with `with` semantics and many `exec` calls.
- `SandboxProcess`: a handle to a running command exposing line-delimited stdout/stderr, `wait()`, `kill()`, `exit_code`, and a `collect()` drain-everything convenience.

`SandboxSnapshot` carries snapshot/restore state for flavors that support it. Sandbox stdout/stderr surfaces on the step as a dedicated `sandbox:<id>` log source through a new `LoggingContext.begin()/end()/emit()` side-channel.

Optional methods (`attach`, `restore`, `create_snapshot`, `aexec`, `upload_file`, `download_file`, `destroy`) raise `NotImplementedError`. Flavors opt in.

Stacked merge order: #4867 then #4870.

## Merge order (stack)

```
develop
   ├─ 2️⃣ #4867 Modal flavor
   │  └─ 3️⃣ #4869 Harbor POC
```

**This PR merges first.** Everything below it in the stack is blocked on it.


src/zenml/utils/logging_utils.py
```diff
@@ -278,14 +332,10 @@ def __exit__(
                 metadata={"zenml.event.type": "exception"},
             )
 
+        self.end()
```
schustmi: Why is self.end() not called in the `try..finally..` block below where the original call happened? Right now, if end fails the remaining cleanup logic would not run.

src/zenml/sandboxes/snapshot.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Round-trips via ``session.snapshot() → BaseSandboxSnapshot`` and
+    ``sandbox.restore(snapshot) → SandboxSession``. Each flavor subclasses
+    this model and ships a dedicated materializer for it.
+
+    The ``provider`` field must match the flavor name; ``BaseSandbox.restore``
+    rejects cross-flavor snapshots.
+    """
+
+    provider: str = Field(
```
schustmi: Maybe `sandbox_flavor` or something similar? Or even the component UUID? What if I have two different components of the same flavor, with different service connectors. Component A might not be able to load a snapshot of component B?

src/zenml/sandboxes/process.py
```diff
@@ -0,0 +1,167 @@
...
+        Tuple of ``(joined_string, truncated_flag)``.
+    """
+    parts: List[str] = []
+    total = 0
+    truncated = False
+    for line in stream:
+        total += len(line)
+        if total > max_chars:
+            truncated = True
+            continue
```
schustmi, line 193: This logic makes no sense at all. Why would we skip some intermediate lines that exceed the total limit, but then continue on with subsequent lines. The user will read the output but won't know theres a gap in there.

src/zenml/sandboxes/process.py
```diff
@@ -0,0 +1,167 @@
...
+            stderr_truncated=stderr_truncated,
+        )
+
+
+def _drain_capped(stream: Iterator[str], max_chars: int) -> Tuple[str, bool]:
+    """Drain a line iterator fully, capping the returned string at ``max_chars``.
+
+    Always consumes to completion (StopIteration) so the underlying
+    provider's wait() is not blocked. Lines that would push the total
+    over ``max_chars`` are dropped, but iteration continues.
```
schustmi: This I don't get. How does not consuming to completion block the `wait(..)`?

src/zenml/sandboxes/process.py
```diff
@@ -0,0 +1,167 @@
...
+            and per-stream truncation flags.
+        """
+        stdout, stdout_truncated = _drain_capped(self.stdout(), max_chars)
+        stderr, stderr_truncated = _drain_capped(self.stderr(), max_chars)
+        exit_code = self.wait()
+        if self._session is not None:
+            self._session._log_exec_result(
+                exit_code=exit_code,
+                started_at=self._started_at,
+            )
```
schustmi: If I call collect multiple times, we would log this multiple times even though it's the same result, correct?

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        return not self.is_remote
+
+
+class BaseSandboxSettings(BaseSettings):
+    """Per-step / per-pipeline overrides for a Sandbox component."""
+
+    base_image: Optional[str] = Field(
+        default=None,
+        description="Image for the Session. None → flavor default; the "
```
schustmi: What does `None->flavor default` mean? Each flavor just hard-codes an image somewhere?

`<step>`: I hate this. If anything it should be a different attribute, IMO it shouldn't even exist. That image is certainly not the one wanted in the sandbox, and only some lazy workaround. Just using a plain ubuntu as default would be better.

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+    )
+    environment: Dict[str, str] = Field(
+        default_factory=dict,
+        description="Per-step env vars merged into the Session on top of "
+        "the component's `StackComponent.environment` and resolved "
+        "secrets (settings override on key collision). For secret-backed "
+        "values, attach a secret to the component via "
+        "`zenml sandbox register --secret=...`.",
+    )
+    copy_local_env: bool = Field(
```
schustmi: Again, I don't think this should exist.

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+        default=None,
+        description="Session-level timeout passed through to the provider's "
+        "TTL knob. None lets the provider apply its own default.",
+    )
+
+
+class BaseSandbox(StackComponent, ABC):
+    """Base class for all ZenML sandbox components.
+
+    See ADR 0001 (component-vs-launcher) and ADR 0002 (attach vs restore) in
```
schustmi: In general, the docstrings in this PR are a complete mess. Referencing some plan that no one cares. Literally 99% useless, you can let it rewrite everyhing to have precise, useful information.

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+
+        Returns:
+            The effective settings for the next Session.
+        """
+        from zenml.steps.step_context import StepContext
+
+        settings_cls = self.settings_class or BaseSandboxSettings
+
+        if override is not None:
+            settings_dict = self.config.model_dump(exclude_unset=True)
```
schustmi: This doesn't use our partial dict overrides, behaves different than all other places where we overwrite configs.

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+            try:
+                env.update(Client().get_secret(secret_id).secret_values)
+            except Exception as e:
+                logger.warning(
+                    "Could not resolve sandbox component secret %s: %s. "
+                    "Skipping.",
+                    secret_id,
+                    e,
+                )
+        if settings is not None and settings.environment:
```
schustmi: self.environment, settings.environment, but secrets only on self. A complete mess. Either we (ab)use the env + secrets defined on the component, which are meant to be set in the environment where the component is being used. Or we don't, and duplicate them on the config/settings. But for sure we don't do the halfway mess implemented here.

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+    )
+    copy_local_env: bool = Field(
+        default=False,
+        description="If True, propagates the step process's full local env "
+        "into the Session. Layered FIRST in the env merge — explicit "
+        "component env, component secrets, and per-step "
+        "settings.environment all override values copied from the local "
+        "env. Convenient for prototyping; off by default for security.",
+    )
+    timeout_seconds: Optional[int] = Field(
```
schustmi: Already ignored by local sandbox in this PR. This should not be in the global sandbox config/settings in that case.

src/zenml/sandboxes/base.py
```diff
@@ -0,0 +1,370 @@
...
+logger = get_logger(__name__)
+
+
+# Sentinel: when used as ``base_image``, the Session uses the image the
+# current ZenML step is running in. Falls back to flavor default with a
+# warning if the step is not containerized. See docs/book/component-guide/sandboxes/dev/plan.md "Session Environment".
+STEP_IMAGE = "<step>"
+
+
+class BaseSandboxConfig(StackComponentConfig):
```
schustmi: Config should always inherit from settings in case the settings actually define some values. Otherwise, some internal logic breaks down.

Probably this happens in the flavors and therefore might not surface, still wrong.

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+    warning is emitted at session creation if a non-None / non-STEP_IMAGE
+    value is set.
+
+    Overrides ``copy_local_env`` default to ``True``: LocalSandbox runs
+    in the same shell environment as the step (no isolation), so
+    propagating PATH and friends matches user expectations. Explicit
+    settings still override.
+    """
+
+    copy_local_env: bool = True
```
schustmi: Why again?

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+)
+
+
+class LocalSandboxConfig(BaseSandboxConfig):
+    """Configuration for the local subprocess sandbox.
+
+    No fields. Env vars and secrets flow through the standard
+    ``StackComponent.environment`` / ``.secrets`` mechanism inherited
+    from ``BaseSandboxConfig``.
+    """
```
schustmi: This not inheriting from the settings can already causes major issues at runtime.

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+        """Initializes the process wrapper.
+
+        Args:
+            process: A live ``subprocess.Popen`` opened with text-mode
+                stdout/stderr pipes.
+            session: Owning session. When provided, stdout/stderr lines
+                are forwarded through ``session._wrap_stream`` so they
+                land in the per-session sandbox log source.
+        """
+        self._process = process
```
schustmi, line 86: why does this not set `started_at`?

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+        Args:
+            settings: Optional per-step ``BaseSandboxSettings`` /
+                ``LocalSandboxSettings``.
+
+        Returns:
+            A ``LocalSandboxSession`` ready for ``exec()``.
+        """
+        logger.warning(_NO_ISOLATION_WARNING)
+
+        eff = cast(LocalSandboxSettings, self.resolve_settings(settings))
```
schustmi: `eff`: No.

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+        return self.generate_default_sdk_docs_url()
+
+    @property
+    def logo_url(self) -> str:
+        """Dashboard logo URL.
+
+        Returns:
+            The flavor logo URL.
+        """
+        return "https://public-flavor-logos.s3.eu-central-1.amazonaws.com/orchestrator/local.png"
```
schustmi: Needs different image prob.

src/zenml/sandboxes/session.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        raise NotImplementedError(
+            f"{type(self).__name__} does not support file downloads."
+        )
+
+    @abstractmethod
+    def close(self) -> None:
+        """Releases this local handle. Cheap; safe to call multiple times.
+
+        Subclasses must call ``self._close_log_ctx()`` from their
```
schustmi: Risky. Better have subclasses override another method, and call the log context close and the subclass implementation in a method in the super class.

src/zenml/sandboxes/session.py
```diff
@@ -0,0 +1,425 @@
...
+        """Async sibling of ``exec``. Optional; default raises NotImplementedError."""
+        raise NotImplementedError(
+            f"{type(self).__name__} does not support async exec. "
+            "Use exec() instead."
+        )
+
+    def snapshot(self) -> BaseSandboxSnapshot:
+        """Captures Session state. Returns a serializable handle for restore.
+
+        Optional; default raises NotImplementedError.
```
schustmi: Useless information; Remove this. Everywhere.

src/zenml/sandboxes/session.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        from zenml.metadata.metadata_types import MetadataType, Uri
+        from zenml.utils.metadata_utils import log_metadata
+
+        prefix = f"sandbox.{self.id}"
+        metadata: Dict[str, MetadataType] = {
+            f"{prefix}.flavor": self._parent.flavor,
+        }
+        url = self._dashboard_url()
+        if url:
```
schustmi, line 308: if url := self._dashboard_url():

src/zenml/sandboxes/session.py
```diff
@@ -0,0 +1,425 @@
...
+                self.id,
+                e,
+            )
+            self._log_forwarding_disabled = True
+            return None
+
+        self._log_ctx = ctx
+        return ctx
+
+    def _emit_sandbox(
```
schustmi: emit_sandbox?

src/zenml/sandboxes/session.py
```diff
@@ -0,0 +1,425 @@
...
+            lines: Source line iterator from the flavor's process.
+            stream: ``"stdout"`` → INFO, ``"stderr"`` → WARNING.
+
+        Yields:
+            Each line from the source iterator, unchanged.
+        """
+        if self._log_forwarding_disabled:
+            yield from lines
+            return
+        level = logging.INFO if stream == "stdout" else logging.WARNING
```
schustmi: why is stdERR a warning?

src/zenml/sandboxes/session.py
```diff
@@ -0,0 +1,425 @@
...
+does NOT join the thread's active-context stack. Incidental
+``logger.info(...)`` calls in the step process (ZenML internals,
+OpenAI HTTP retries, etc.) stay in the regular step log; only the
+explicit ``emit_to(record)`` calls in this module land in the
+sandbox source.
+
+Step metadata (sandbox session id, flavor, optional dashboard URL) is
+published once on ``__enter__`` via ``_on_enter()`` and keyed by the
+session id so multiple sessions in one step don't overwrite each
+other.
```
schustmi: Remove all of it.

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+                stderr=subprocess.PIPE,
+                text=True,
+                bufsize=1,  # line-buffered
+            )
+        except (FileNotFoundError, OSError) as e:
+            raise SandboxExecError(
+                f"LocalSandbox exec failed to launch ({type(e).__name__}): {e}"
+            ) from e
+        process = LocalSandboxProcess(popen, session=self)
+        process._started_at = started_at
```
schustmi: pass in init

src/zenml/sandboxes/local_sandbox.py
```diff
@@ -0,0 +1,430 @@
...
+            if isinstance(command, list)
+            else shlex.split(command)
+        )
+        # Marker into the sandbox log so the source reads like a shell
+        # session: `$ <command>` line followed by its stdout/stderr.
+        self._log_command(argv)
+
+        effective_cwd = cwd or self._workdir
+        if not os.path.isabs(effective_cwd):
+            effective_cwd = os.path.join(self._workdir, effective_cwd)
```
schustmi: what is this doing?

if i don't pass a cwd, and self._workdir is not absolute, it just concatenates it twice.

also, the sandbox doesn't persist the CWD, correct?

schustmi (changes requested):

schustmi (approved):

htahir1:
@schustmi — relaying three core-scope findings from @strickvl's reviews of the downstream flavor PRs (full reviews: https://gist.github.com/strickvl/710208dc93772743788696f1377e6ad5). They all land in base files this PR owns, so wanted your call before anything moves:

1. **`_resolve_session_environment()` ignores component `--env`/`--secret`** (`base.py`): sessions only receive `settings.sandbox_environment`. Component-level `--env`/`--secret` do reach the *step process* (via `env_utils.get_runtime_environment`/`get_runtime_secret_environment`), but never the sandbox container itself. Should sessions additionally inherit component env + resolved secrets (per-session settings winning)? The modal docs were rewritten to describe the current behavior accurately in the meantime.

2. **Direct `session.close()` leaks the logging context** (`session.py`): `__exit__` calls `close()` then `_close_logging_context()`, but users calling `close()` directly leave the `sandbox:<session_id>` log source open. Either base `close()` could own final cleanup, or the contract could require concrete `close()` implementations to call `_close_logging_context()` — the Cloudflare flavor does the latter for now.

3. **LocalSandbox doesn't track spawned processes** (`local_sandbox.py`): `close()` removes the workdir and marks the session closed, but a `sleep 3600` launched via `exec()` keeps running after the context manager exits. Suggested: track `Popen` handles on the session and terminate still-running ones in `close()`/`destroy()`, with a test that closes a session while a child is alive.

None of these block the flavor PRs — happy to implement whichever you want, here or as a follow-up.

htahir1:
Resolution after discussing with @schustmi:

1. **Component `--env`/`--secret` not flowing into sessions** — intentional, no change. The modal docs already describe the actual behavior (step-process env via the standard component mechanism, `sandbox_environment` for in-sandbox vars).
2. **`close()` logging-context leak** — fixed in `c255b877` on this branch: `SandboxSession.close()` is now concrete and owns the logging-context cleanup in a `finally` around a new abstract `_close()` flavor hook; `__exit__` reduces to `close()`. Both the context-manager and direct-`close()` styles stay supported (the Harbor bridge holds a session across async `start()`/`stop()`, so a with-only API would break legitimate split-lifecycle users), and flavors can no longer forget the cleanup — a flavor still overriding `close()` instead of `_close()` can't even instantiate. All four flavors adopted the hook on their branches, and their `destroy()` implementations now route through `close()` so direct destroys flush the log source too.
3. **LocalSandbox process tracking** — fixed in the same commit: sessions track spawned `Popen` handles and terminate any still running on `close()`/`destroy()` (terminate, 5s grace, kill), with a test that closes a session while a `sleep 3600` child is alive.

htahir1:
Follow-up to the close-ownership change (`3fff1025`): `close()` is now fully terminal, not just for logging. The previous state was inconsistent — a "closed" session still accepted `exec`/`upload`/`download`/`create_snapshot` and ran them silently unlogged (the logging context is created lazily on first emit, so it would even resurrect). Now the base class owns a closed flag: `close()` is idempotent, and every session-using operation funnels through a base-level guard that raises `SandboxSessionClosedError` on a closed handle. Flavors implement `_exec`/`_create_snapshot`/`_upload_file`/`_download_file` hooks and structurally cannot bypass the check. `destroy()` stays callable on a closed handle (terminating the provider sandbox after releasing the local handle is legitimate). LocalSandbox dropped its private duplicate of the same machinery. All four flavors adopted the hooks on their branches.

PR #4867, opened 2026-05-27
Add Modal Sandbox flavor

htahir1:
## Summary

Adds the Modal flavor of the `Sandbox` stack component. Wraps Modal's [`Sandbox`](https://modal.com/docs/reference/modal.Sandbox) primitive in `BaseSandbox` / `SandboxSession` / `SandboxProcess` — the same interface as the LocalSandbox in the base PR, so example code is portable across flavors.

## What's in the box

- `ModalSandbox` — looks up the Modal `App` per call (honoring `modal_environment`), claims a `Sandbox` per session, multiplexes `exec()` calls through Modal's `ContainerProcess` API
- `ModalSandboxSession` — exposes Modal's `snapshot_filesystem()` as a `ModalSandboxSnapshot` artifact, plus `restore()` to boot a new session from that snapshot with the session environment re-applied (the basis for agent fan-out in #4868)
- `ModalSandboxProcess` — line-buffers Modal's chunk-based stdout/stderr into the `SandboxProcess` line-iterator contract; `kill()` sends EOF via `StreamWriter.write_eof()`/`drain()` (Modal has no per-command kill — `destroy()` terminates the whole sandbox)
- `src/zenml/integrations/modal/sandbox_utils.py` is **vendored byte-identical** from the Modal orchestrator branch (#4915) so the two PRs merge in either order with zero conflicts; this file should not be reviewed or edited here

## Settings

`ModalSandboxSettings` (per-step overridable): `image` (default `python:3.11-slim`), `cpu`, `memory`, `timeout` (sandbox TTL, default 1h — closing a session does NOT terminate the sandbox; only `destroy()` or the TTL does), plus `gpu` / `region` / `cloud` / `modal_environment` inherited from the step operator settings. `ModalSandboxConfig` inherits the settings class, so all of these can also be set at registration time (`zenml sandbox register --image=...`), plus `app_name`.

## Breaking changes
- Bumps the minimum modal version requirement to `>-=1.4`


src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,703 @@
...
+            # Round up: Modal rejects fractional MB; truncating could
+            # under-provision a tight memory request.
+            kwargs["memory"] = math.ceil(memory_mb)
+        if eff.region is not None:
+            kwargs["region"] = eff.region
+        if eff.cloud is not None:
+            kwargs["cloud"] = eff.cloud
+        return kwargs
+
+    @staticmethod
```
schustmi: The step is already running and consuming the resources, and now the sandbox will consume them again. That makes no sense.

src/zenml/integrations/modal/flavors/modal_sandbox_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+
+class ModalSandboxSettings(BaseSandboxSettings, ModalStepOperatorSettings):
+    """Per-step settings for the Modal sandbox."""
+
+    # gpu / region / cloud / modal_environment are inherited from
+    # ModalStepOperatorSettings so the two Modal stack components share a
+    # single source of truth. cpu_count / memory / gpu_count come from the
+    # active step's ResourceSettings.
+
+    base_image: Optional[str] = Field(
```
schustmi: Step sentinel should be removed.

Also, why `base_image` and `default_image`? Just call it image, the settings overwrite any config value anyway.

src/zenml/integrations/modal/flavors/modal_sandbox_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+    from zenml.integrations.modal.sandboxes import ModalSandbox
+
+
+class ModalSandboxSettings(BaseSandboxSettings, ModalStepOperatorSettings):
+    """Per-step settings for the Modal sandbox."""
+
+    # gpu / region / cloud / modal_environment are inherited from
+    # ModalStepOperatorSettings so the two Modal stack components share a
+    # single source of truth. cpu_count / memory / gpu_count come from the
+    # active step's ResourceSettings.
```
schustmi: Useless comment.

src/zenml/integrations/modal/flavors/modal_sandbox_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+        "'my-registry/my-image:tag'). Pass the module-local sentinel "
+        "'<step>' to reuse the active step's containerized-orchestrator "
+        "image. When None, the component's 'default_image' is used.",
+    )
+
+
+class ModalSandboxConfig(BaseSandboxConfig, ModalStepOperatorConfig):
+    """Configuration for the Modal sandbox component."""
+
+    # Inherits from ModalStepOperatorConfig so token_id / token_secret /
```
schustmi: I can see that it inherits from that class, by looking exactly two lines above this. The place where i'm supposed to look what a class inherits from, in the class definition.

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,703 @@
...
+        )
+
+    def upload_file(self, local_path: str, remote_path: str) -> None:
+        """Upload a local file into the Sandbox.
+
+        Args:
+            local_path: Path to the file on the caller's machine.
+            remote_path: Destination path inside the Sandbox.
+        """
+        # Uses Modal's Sandbox.filesystem.copy_from_local() API; the
```
schustmi: Why would anyone care about old code that got removed and isn't used anymore.

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,703 @@
...
+            local_path: Destination path on the caller's machine.
+        """
+        # Uses Modal's Sandbox.filesystem.copy_to_local() API; the older
+        # Sandbox.open() + FileIO.read() path is deprecated by Modal as
+        # of 2026-03-09.
+        self._sandbox.filesystem.copy_to_local(remote_path, local_path)
+
+    def close(self) -> None:
+        """No-op: Modal ``Sandbox`` has no client-side resources to drop."""
+        # The Sandbox keeps running on Modal until its TTL expires; use
```
schustmi: Sure it has, `self._sandbox`

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,703 @@
...
+            container_registry = self.stack.container_registry
+        except Exception:
+            container_registry = None
+        if container_registry is None:
+            return None
+        creds = container_registry.credentials
+        if not creds:
+            return None
+
+        import modal
```
schustmi: Import at top, like every other stack component.

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,703 @@
...
+                ``env=`` on ``Sandbox.create``. Skipped on the restore
+                path where injecting env into a frozen filesystem image
+                is rarely what the caller wants.
+
+        Returns:
+            Kwargs for ``modal.Sandbox.create``.
+        """
+        from zenml.config.resource_settings import ByteUnit
+        from zenml.integrations.modal.step_operators.modal_step_operator import (
+            get_gpu_values,
```
schustmi: -> utils file

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,703 @@
...
+        )
+
+        resource_settings = self._active_resource_settings()
+        memory_mb = resource_settings.get_memory(ByteUnit.MB)
+        # ModalSandboxSettings inherits from ModalStepOperatorSettings, so
```
schustmi: Well if the linter accepts it, why would we need a useless three line comment for it then?

schustmi (changes requested):
3 min review, the whole thing needs to be cleaned up

htahir1:
Thanks @strickvl — addressed in `1455fc04a2`.

**Fixed:**
- **restore() auth**: `modal.Image.from_id` now receives `client=modal_client` (verified the kwarg on the installed SDK); added a test asserting the configured client is used.
- **kill() contract**: `kill()` now terminates the owning Modal sandbox via `Sandbox.terminate()` — Modal has no per-command kill, and the EOF-only behavior could make base `collect()`'s kill-then-wait cleanup block until the TTL. Logged loudly, documented in a new Caveats bullet in the docs.
- **docs `--env`/`--secret` claim**: you were right that the paragraph oversold — component `--env`/`--secret` land in the **step process** environment (via `env_utils.get_runtime_environment`/`get_runtime_secret_environment`), not inside the Session. Rewrote the paragraph to say exactly that and to point at `sandbox_environment` for in-sandbox vars. Note the underlying mechanism does work — it's the step process, where agent frameworks run, that gets them.
- **docs example + smoke script leaks**: the docs example now destroys the session in a `finally` (with a note that `close()` keeps the sandbox alive for `attach()`); the smoke script destroys sessions in all checks except the attach one, which destroys at the end.

**Not taking:**
- **SDK contract/signature test**: CI runs mypy against the installed `modal` package (which ships types), so signature drift on `Sandbox.create` / `exec` / `Image.from_id` already fails the lint job — a dedicated signature test would duplicate that.

The "component env/secrets should merge into sandbox sessions" design question is core (`src/zenml/sandboxes/base.py`, #4866) — flagged there rather than diverging the flavor.

htahir1:
Thanks @strickvl — both round-2 findings were real; fixed in `cee2a1e8ae`.

- **kill() left the session handle open on a dead sandbox**: `kill()` now closes the session after terminating, so the next `session.exec()` raises `SandboxSessionClosedError` instead of passing the guard and failing later with a confusing Modal error. Test added (`test_kill_closes_the_session`).
- **destroy() hid termination failures**: it no longer logs-and-pretends. A failed `terminate()` raises `RuntimeError` (the sandbox keeps billing until TTL — that's worth surfacing) and the handle stays **open** so `destroy()` can be retried. Test added covering raise → handle open → retry succeeds.

htahir1:
Thanks @strickvl — all taken, fixed in `e6c1046840`.

- **`with` + `finally: destroy()`**: you're right that `__exit__` would close the handle right after a failed `destroy()` raised, defeating the keep-open-for-retry behavior we just added. The docs example, the PydanticAI example steps, and all four force-cleanup smoke checks now use plain `session = create_session(); try: ...; finally: session.destroy()`. The docs also spell out when the context manager *is* the right tool (close-and-attach-later flows) and why not to mix it with `destroy()`.
- **cpu/memory docs**: added both to the settings table and corrected the resources paragraph — they forward to `Sandbox.create()` and only fall back to Modal's defaults when unset.

htahir1:
Thanks @strickvl — the contract mismatch was real; fixed in `dc123b226c`.

- **`wait(timeout=)` portability**: implemented via `poll()` against a monotonic deadline (Modal's own `wait()` takes no timeout), returning the exit code or raising `TimeoutError`. The refuse-with-`NotImplementedError` behavior is gone; tests updated for both the completes-in-time and deadline-exceeded paths.
- **Smoke script `dst_path` UnboundLocalError**: fixed — initialized to `None`, cleanup skips unassigned paths.

**Not taking** (third time, but your own note agrees it's not a blocker): the `inspect.signature()` contract test — CI runs mypy against the installed typed `modal` package, which fails the lint job on exactly the drift such a test would catch.

(Your local `pymysql` issue: the unit suites run standalone via `PYTHONPATH=src pytest tests/unit/integrations/modal/` — the conftest deps are only needed for the harness-based integration tests.)

src/zenml/integrations/modal/__init__.py
```diff
@@ -24,6 +24,10 @@
 from zenml.stack import Flavor
 
 MODAL_STEP_OPERATOR_FLAVOR = "modal"
+# Same string as the step-operator flavor; intentional. The framework
+# disambiguates by StackComponentType.SANDBOX vs STEP_OPERATOR, so the
+# two flavors can coexist without renaming either.
```
schustmi:
```suggestion
```

src/zenml/integrations/modal/flavors/modal_sandbox_flavor.py
```diff
@@ -0,0 +1,138 @@
...
+    BaseSandboxConfig,
+    BaseSandboxFlavor,
+    BaseSandboxSettings,
+)
+
+if TYPE_CHECKING:
+    from zenml.integrations.modal.sandboxes import ModalSandbox
+
+
+class ModalSandboxSettings(BaseSandboxSettings, ModalStepOperatorSettings):
```
schustmi, line 36: IMO, instead of inheriting from step operator settings/config, the cleaner solution would be to have some base classes/mixins that we use both for the step operator and sandbox (and even orchestrator maybe)

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+
+        Args:
+            *args: Forwarded to ``StackComponent``.
+            **kwargs: Forwarded to ``StackComponent``.
+        """
+        super().__init__(*args, **kwargs)
+        # Explicit Modal client built lazily from the component's
+        # credentials and reused across sessions; rebuilt if it closes.
+        # The lock guards the build against concurrent create_session calls
+        # (e.g. fanned-out mapped steps sharing this component instance).
```
schustmi, line 307:
```suggestion
```

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+    ) -> SandboxSession:
+        """Boot a fresh Modal Sandbox.
+
+        Args:
+            settings: Optional per-step overrides for image / env.
+
+        Returns:
+            A ``ModalSandboxSession`` wrapping the live Modal Sandbox.
+        """
+        eff = cast(ModalSandboxSettings, self.resolve_settings(settings))
```
schustmi: `eff` is a useless name. same below

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+
+        Returns:
+            A new ``ModalSandboxSession`` initialized from the Image. No
+            in-memory state from the original Session is preserved.
+
+        Raises:
+            RuntimeError: If Modal can't load the Image (e.g. id GC'd).
+        """
+        self._validate_snapshot(snapshot)
+        eff = cast(ModalSandboxSettings, self.resolve_settings(None))
```
schustmi: just call it `settings`.

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+            container_registry
+            and container_registry.is_valid_image_name_for_registry(image)
+            and container_registry.credentials
+        ):
+            return container_registry.credentials
+        return None
+
+    def _build_create_kwargs(
+        self,
+        eff: ModalSandboxSettings,
```
schustmi: `settings`. insane name pick

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+                injected at create time).
+
+        Returns:
+            A ``ModalSandboxProcess`` wrapping the live command handle.
+
+        Raises:
+            SandboxExecError: If Modal rejects the launch.
+        """
+        argv: List[str] = (
+            list(command)
```
schustmi:
```suggestion
            command
```

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+
+        kwargs: Dict[str, Any] = {}
+        if cwd is not None:
+            kwargs["workdir"] = cwd
+        if env:
+            # Unlike Sandbox.create, Sandbox.exec takes no env=; per-exec
+            # env is injected via secrets=.
+            kwargs["secrets"] = sandbox_utils.create_runtime_secrets(
+                cast(Dict[str, Optional[str]], env)
+            )
```
schustmi: Have you verified this actually works?

Also, these secrets should probably be cleaned up? Otherwise, each sandbox exec leaves users with more modal secrets

schustmi: Also, not sure if possible, but can we name those secrets? If so, would be useful I think to indicate that these are not actually "secret"

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+        documented retryable ``RuntimeError`` (with the billing warning)
+        and leaves the handle open for retry.
+        """
+        session = cast("ModalSandboxSession", self._session)
+        logger.warning(
+            "Modal has no per-command kill; terminating the whole Modal "
+            "sandbox '%s' to stop the command.",
+            session.id,
+        )
+        session.destroy()
```
schustmi: Mhm this seems like a non-ideal solution to kill the whole sandbox. Let's imagine someone runs multiple execs in parallel, killing one of them will also stop all the others.

For sure not intended by the abstraction, I just don't know whether this is worse or just not implementing it at all and not supporting `kill()`.

Have you considered alternatives (not sure if these are possible) like:
- When we call exec, we potentially have the PID. We could launch another exec which kills the original process?

schustmi (changes requested):

scripts/test_modal_sandbox.py
```diff
@@ -0,0 +1,191 @@
+#  Copyright (c) ZenML GmbH 2026. All Rights Reserved.
```
schustmi: This file should be somewhere else, maybe in integration tests?

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,529 @@
...
+            if isinstance(command, list)
+            else shlex.split(command)
+        )
+        self._log_command(argv)
+
+        kwargs: Dict[str, Any] = {}
+        if cwd is not None:
+            kwargs["workdir"] = cwd
+        if env:
+            # Unlike Sandbox.create, Sandbox.exec takes no env=; per-exec
```
schustmi: Sandbox.exec takes `env=...` for me? is this something that modal introduced in a later release maybe?

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
schustmi: It seems like we don't really get the OS PID from modal. Something we could do though potentially:
- Inject an env var in each exec call `ZENML_PID=<UUID>`
- When `kill()` is called, we exec into the sandbox. Find a process for which the specific UUID is set in the env var.
- Run `pkill` to kill that process. potentially even `pkill -9` to force it.

schustmi: Again, not sure this complexity is what we actually want, or whether your approach or just not supporting `kill()` is better.

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
```diff
@@ -0,0 +1,522 @@
...
+        argv: List[str] = (
+            command if isinstance(command, list) else shlex.split(command)
+        )
+        self._log_command(argv)
+
+        kwargs: Dict[str, Any] = {}
+        if cwd is not None:
+            kwargs["workdir"] = cwd
+        if env:
+            kwargs["env"] = env
```
schustmi, line 224: See my earlier comment. I said `on my version, env exists`. Now i installed `1.0.0`, which is part of the range of our integration. it doesn't exist there. which means this now breaks for anyone having that version installed.

schustmi (changes requested):

src/zenml/integrations/modal/sandboxes/modal_sandbox.py
schustmi: I'm not sure if secrets is a good way to do it. Or if we want `SOME_ENV=1 grep ...`.

schustmi (approved):

PR #4868, opened 2026-05-27
Add PydanticAI + Sandbox example

htahir1:
## Summary

Worked example showing how to drive a PydanticAI agent through a ZenML `Sandbox`-backed pipeline. The agent has four tools (`run_python`, `run_shell`, `list_files`, `read_file`), each backed by `session.exec(...)` on the active stack's Sandbox. The agent tools talk only to the flavor-agnostic `BaseSandbox` interface. Snapshots are an optional capability (Modal implements them today): where available, the dependency install runs once and every subagent restores it for free; on flavors without snapshots the pipeline degrades gracefully — `prep_step` probes cheaply and each subagent installs its own deps in a fresh session (verified live against the Kubernetes sandbox flavor: probe → fallback → 3 concurrent subagents → reducer, 43.8s).

**Stacked on #4867** (Modal flavor). Transitively depends on **#4866** (core).

## Pipeline shape

```
prep_step ──► snapshot ─┐
                        ├─► subagent_step.map() ──► reducer_step
planner_step ──► subtasks ┘
```

1. `prep_step` boots a sandbox, `pip install`s the scientific stack once, snapshots the filesystem into a `ModalSandboxSnapshot` artifact
2. `planner_step` asks a small LLM to decompose the user goal into N independent subtasks
3. `subagent_step.map(snapshot=unmapped(snap), subtask=subtasks)` fans out — each subagent restores the shared snapshot and runs the agent loop on one subtask
4. `reducer_step` synthesizes the per-subagent answers into a final response

Real run against a Modal account: prep ~13s, planner ~5s, 3 subagents in parallel ~12–28s each, reducer ~9s — about **1m11s** end-to-end.

## What's interesting

- **Snapshot fan-out**: heavy setup runs once; each fanned-out subagent boots from the snapshot for free. The agent's system prompt reflects this — the scientific stack is preinstalled, `pip install` is only for additional packages.
- **Honest caching**: only the three LLM-dependent steps disable caching (`@step(enable_cache=False)`); `prep_step` stays cacheable, so re-runs reuse the snapshot instead of reinstalling.
- **Secrets stay out of images**: the OpenAI key reaches the sandbox via the component-level secret reference, never via `DockerSettings.environment` (which would bake it into a pushed image).
- **Per-session log forwarding**: each subagent's sandbox stdout/stderr lands in its own `sandbox:<session_id>` step log source; metadata records `sandbox.<session_id>.dashboard_url` as a clickable Modal link.
- **`step.map` + `unmapped`** broadcasts the snapshot to all N subagents; subtasks zip 1:1.

## Test plan

- [x] Smoke import (`python -c "import run"`) — settings construction validates
- [x] End-to-end run against Modal — completes in ~1m11s
- [x] ruff clean on `examples/sandbox_pydantic_ai/`

## Merge order (stack)

```
develop
   ├─ 2️⃣ #4867 Modal flavor
   │  └─ 3️⃣ #4869 Harbor POC
```

**This PR merges after #4866 → #4867.**


htahir1:
Thanks @strickvl — addressed in `3c522495` (now `4131129310` after the modal cascade).

**Fixed:**
- **Sandbox leak until TTL**: real money bug, good catch. `prep_step` now destroys its session right after the snapshot is taken, and `subagent_step` destroys in a `finally`. A default run no longer leaves 4 Modal sandboxes billing for the TTL.

**Not taking:**
- **OpenAI key instructions**: I believe this one is inverted. Component `--secret` values *are* resolved into the step process environment at runtime — `env_utils.get_runtime_secret_environment` collects secrets from every stack component and `step_runner.py` applies them via `temporary_environment(...)` around the user code, on any orchestrator. That's exactly where the PydanticAI calls run, and the README's "resolved into the step environment at runtime" sentence describes this mechanism. Happy to be corrected if you've seen it not resolve in practice.
- **`Field(min_length=3, max_length=3)` on subtasks**: the pipeline maps over however many subtasks the planner returns — 2 or 5 fan out fine; "exactly 3" is just prompt guidance for a predictable demo.
- **Pinning numpy/scipy/pydantic-ai**: repo examples deliberately float dependencies.

schustmi (approved):

PR #4869, opened 2026-05-28
Harbor evals on ZenML Sandbox

htahir1:
## Summary

A POC showing how to run [Harbor](https://harborframework.org) evaluation trials inside a ZenML pipeline, with the trial's container backed by whatever Sandbox flavor is on the active stack. This is the worked example that proves the Sandbox component composes with an external framework (alongside #4868, which drives a PydanticAI agent directly).

**Stacked on #4867** (Modal flavor). Transitively depends on **#4866** (core).

## Architecture

```
ZenML pipeline (single step, the user's entry point)
  -> Harbor (programmatic: tasks, agent, verifier, ATIF, reward)
    -> ZenMLSandboxEnvironment (implements harbor.BaseEnvironment)
      -> Client().active_stack.sandbox.create_session()
        -> Modal sandbox
```

The bridge file `examples/sandbox_harbor_poc/zenml_sandbox_env.py` implements the full `harbor.BaseEnvironment` contract — `start`/`stop`/`exec`/`upload_file`/`download_file`/`upload_dir`/`download_dir` — on top of `BaseSandbox`/`SandboxSession`, talking only to the public surface. Directory scaffolding delegates to Harbor's own base-class helper rather than reimplementing it. The modal-specific settings override is imported lazily, so the bridge itself has no hard dependency on the modal integration and works with any flavor.

## What this gives you over plain `harbor run`

- **Run lineage** — every trial is a ZenML pipeline run with dashboard, artifact store, replay, caching. The trial's full Harbor jobs directory (logs, trajectory, verifier output) is tarred and stored as a `harbor_trial_artifacts` output artifact before the local tempdir is cleaned up, so nothing about the trial is lost.
- **No per-trial rebuild** — Task `[environment].docker_image` flows through to the flavor's `image` setting; the Sandbox component owns the image.
- **Portable substrate** — swap the Sandbox flavor (Modal → Cloudflare → GKE Agent Sandbox) and Harbor sees nothing change.

Bundled task: `tasks/hello/` — hermetic, writes `42` to `/app/answer.txt`, verifier scores 1.0 on match. Runs under the `oracle` agent so no LLM keys are needed for smoke testing.

## Registry benchmarks (Terminal Bench & friends)

The task argument also accepts a `dataset:NAME@VERSION` spec naming a [Harbor registry](https://www.harborframework.com/) dataset:

```bash
python run.py dataset:terminal-bench-sample@2.0 oracle 1   # 10-task sample
python run.py dataset:terminal-bench@2.0 oracle 1          # full 89-task Terminal Bench 2.0
```

`build_matrix` expands the spec through **Harbor's own registry resolver**, so the task set and commit pins match `harbor run` exactly, then fans out one mapped trial step per benchmark task. Tasks travel as plain-string `git+URL@COMMIT:SUBPATH` refs — each step fetches only its own pinned task at runtime, nothing heavy ships with the example — and tasks that pin a prebuilt `docker_image` boot it directly via the bridge's existing settings override.

## Test plan

- [x] Smoke run: `python run.py` produces a green ZenML pipeline + a `JobResult` artifact with reward=1.0 + the trial-artifacts tarball
- [x] Registry task live on Modal: terminal-bench-2.0-sample `chess-best-move` (git-pinned task, prebuilt `ghcr.io/...` image) — reward 1.0
- [x] `dataset:terminal-bench@2.0` resolves all 89 tasks; `dataset:terminal-bench-sample@2.0` all 10
- [x] ruff clean on `examples/sandbox_harbor_poc/`
- [x] Modal unit suite (carried through the stacked branch) passes — 45/45

## Merge order (stack)

```
develop
   ├─ 2️⃣ #4867 Modal flavor
   │  └─ 3️⃣ #4869 Harbor POC
```

**This PR merges after #4866 → #4867.**


htahir1:
Thanks @strickvl — addressed in `de4616ce` (now `f41da5cff4` after the modal cascade).

**Fixed:**
- **`docker_image` on non-Modal flavors**: `_settings_override()` now checks the active sandbox component's flavor and raises a clear `NotImplementedError` ("docker_image is currently only supported with the Modal sandbox flavor") instead of handing Modal settings to whatever is on the stack.
- **Silently ignored task contract fields**: `start()` now fails fast when `allow_internet = false` is set (users must not believe network isolation is enforced when it isn't), and emits one warning listing any of `cpus`/`memory_mb`/`gpus` that are set. The per-exec `user` ignore was upgraded from debug to warning (it's not a field on Harbor's `EnvironmentConfig`, so it can't be checked at start time).
- **README**: prerequisites now spell out the active-stack-with-Modal-sandbox requirement (with register commands), the portability claim is narrowed to "smoke-tested with the Modal flavor" with a warning that LocalSandbox executes on the host, "full `BaseEnvironment` contract" is reworded to "the `BaseEnvironment` methods" with the known gaps listed, and the hard image requirements (`bash`, `timeout(1)`, `tar`) are documented.

**Not taking:**
- **asyncio.run() thread-helper**: the comment in `run.py` already acknowledges the nested-loop limitation; for a POC I'd rather keep the call obvious than add loop-detection machinery.
- **/tmp tarball cleanup on failed extraction**: the sandbox is destroyed after the trial either way — the leak dies with it.

htahir1:
Thanks @strickvl — this one's a reviewer-environment artifact, not a bug in the stack:

`src/zenml/integrations/modal/sandbox_utils.py` **is present at the exact head you reviewed**:

```
$ git ls-tree 88e0b5fae5 --name-only src/zenml/integrations/modal/ | grep sandbox_utils
src/zenml/integrations/modal/sandbox_utils.py
```

It entered this branch via the base-branch merge (it's owned by #4867 / vendored from the modal-orchestrator branch), so it doesn't appear in a `base...HEAD` diff or in this PR's "files changed" — which is presumably what your context builder fed the review. The import works at head; `from zenml.integrations.modal.sandboxes import ModalSandbox` is exercised by the 46-test modal suite that runs on this branch's CI.

To your question: yes, the stack is self-consistent — `sandbox_utils.py` is #4867's file (kept byte-identical with the modal-orchestrator branch so the two merge in either order), and this branch carries #4867 merged in.

schustmi (approved):

PR #4870, opened 2026-05-28
Add k8s_agent_sandbox integration: Sandbox flavor backed by Google's Agent Sandbox / Substrate

htahir1:
## Summary

Adds the `k8s_agent_sandbox` flavor — wraps Google's [Agent Sandbox on GKE](https://cloud.google.com/blog/products/containers-kubernetes/bringing-you-agent-sandbox-on-gke-and-agent-substrate) (the [`kubernetes-sigs/agent-sandbox`](https://github.com/kubernetes-sigs/agent-sandbox) operator and its [`k8s-agent-sandbox`](https://pypi.org/project/k8s-agent-sandbox/) Python SDK) in `BaseSandbox` / `SandboxSession` / `SandboxProcess`. Works against any Kubernetes cluster running the operator; on GKE the operator additionally supports Pod Snapshots (planned follow-up).

**Stacked on #4866.** Merge that first. Independent of the Modal flavor (#4867) — the two integrations import nothing from each other.

## Scope

V1 covers `create_session` / `exec` / `close` / `destroy`. The SDK's `commands.run` is one blocking HTTP POST, so the exec contract is honest about that — a single-shot `ExecutionResult` exposed via one-element `stdout()`/`stderr()` iterators (split into lines so the `SandboxProcess` line-iterator contract holds). Per-exec env vars are injected as an `export KEY=value; ` prefix, so they apply across the whole command chain (`&&`, `|`, `;`). Snapshot/restore, file I/O, and real streaming are out of scope; tracked as follow-ups.

## Two template modes

- **Template mode** — `template_name` references a pre-created `SandboxTemplate` custom resource
- **Inline mode** — synthesizes a one-off `SandboxTemplate` per session from `image` + the active step's `ResourceSettings` + optional `KubernetesPodSettings` + `sandbox_environment`, and cleans it up at session close. Cleanup re-enters the service-connector scope (`_kube_default_config`) so it never runs against the ambient kubeconfig.

Connection modes (1:1 with the SDK): `gateway` / `local_tunnel` / `direct` / `in_cluster`. Service-connector wiring is declared via `ServiceConnectorRequirements(resource_type="kubernetes-cluster")`, so both the `gcp` (GKE) and `kubernetes` (any cluster) connectors satisfy it. The `kubernetes` pin is inherited from `KubernetesIntegration.REQUIREMENTS` to prevent drift.

## DRY across the kubernetes integration

Reuses `kube_utils.convert_resource_settings_to_k8s_format` (cpu/memory/gpu mapping), `manifest_utils.add_pod_settings` (affinity/tolerations/node_selectors/volumes/env/security_context), and `kubernetes.client.V1PodSpec` + `ApiClient.sanitize_for_serialization` for the CR body. The untyped `k8s_agent_sandbox` SDK is handled via the repo-standard `ignore_missing_imports` mypy override in `pyproject.toml`.

## Test plan

- [x] 42 unit tests covering wiring, settings resolution, connection-config routing, inline template body construction (container shape, ports, probes, ResourceSettings merge, GPU routing, KubernetesPodSettings pass-through, env exports), connector-scoped cleanup ordering, orphan cleanup on `create_sandbox` failure, error UX for 404/403/409, session close/destroy split, exec argv handling
- [x] ruff, mypy, pydoclint clean
- [ ] Live-cluster e2e (kind + operator) — deferred, pre-GA gate
- [ ] Snapshot/restore via GKE Pod Snapshots (`PodSnapshotSandboxClient`) — follow-up PR

## Merge order (stack)

```
develop
   ├─ 2️⃣ #4867 Modal flavor
   │  └─ 3️⃣ #4869 Harbor POC
```

**This PR merges after #4866.** It does not depend on #4867 — but the head branch carries #4867's commits, so the diff currently shows modal files. Review only `src/zenml/integrations/k8s_agent_sandbox/**` (+ its tests/docs); the modal files belong to #4867 and disappear from this diff once it merges.


htahir1:
Thanks @strickvl — addressed in `a7ac839f` (now `f64285b8db` after merging the modal branch in, which also refreshes the stale modal code you saw in this diff — the app cache you flagged was already removed there, and the restore-client + kill fixes landed on #4867).

**Fixed:**
- **Env-key shell injection**: keys are validated against `^[A-Za-z_][A-Za-z0-9_]*$` before being interpolated into `export` statements; a key like `FOO; rm -rf /` now raises before anything reaches the sandbox. Test added.
- **`cwd` + `env` failure semantics**: composition is now `cd <cwd> && { <exports><cmd>; }`, so nothing runs in the wrong directory when `cd` fails — including commands with internal `;`/`&&` chains. Test asserts the exact composed string.
- **Process-global kube config race**: `_kube_default_config()` now serializes on a module-level lock. Checked all call sites for re-entrancy first — none nest (destroy exits its scope before `_delete_inline_template` opens a new one), so a plain `Lock` suffices.
- **Inline template cleanup ordering**: the tracker is now cleared only after a successful delete (or a 404 = already gone); transient failures keep it set so a later `destroy()` retries instead of orphaning the `SandboxTemplate`. Added a transient-failure-then-retry test.

**Not taking here:**
- **Component `--env`/`--secret` merge into `_resolve_session_environment`**: that's `src/zenml/sandboxes/base.py` — core abstraction owned by #4866, flagged there. Worth noting the values aren't dropped today: they're injected into the step process env by `env_utils.get_runtime_environment`/`get_runtime_secret_environment`; the open design question is whether sessions should *additionally* inherit them.

htahir1:
## Known issues to resolve before merge (found via live testing of the sibling kubernetes flavor #4926)

Two real concurrency/correctness issues in the connector path that the unit suite can't see. Filing here so they're tracked; **we'll fix both before this PR merges.**

### 1. The service connector is silently defeated by the SDK

`_kube_default_config()` sets the connector's config as `Configuration._default` for the scope, but inside that scope `_build_client()` → `SandboxClient()` → `K8sHelper.__init__` (k8s-agent-sandbox SDK) unconditionally calls:

```python
try:
    config.load_incluster_config()
except config.ConfigException:
    config.load_kube_config()
```

`load_kube_config()` **overwrites `Configuration._default`** with the ambient kubeconfig before the SDK constructs its `CustomObjectsApi`/`CoreV1Api`. Net effect: the SDK's own calls use ambient credentials, not the connector's — so a stack whose connector points at a *different* cluster than the local kubeconfig talks to the wrong cluster (or fails). The connector is this flavor's headline feature (`ServiceConnectorRequirements`), so this needs fixing. The SDK has no `ApiClient` injection seam, so the fix likely means constructing the SDK's client objects ourselves with the connector config, or upstreaming an injection point.

### 2. The global-config lock serializes all fan-out

`_kube_default_config_lock` is held across the **entire** `create_session()` (`k8s_agent_sandbox.py:657`), including `create_sandbox`'s claim creation and the `wait_for_sandbox_ready` watch (default `sandbox_ready_timeout=180s`). N concurrently-mapped trial/subagent steps therefore serialize to N × (schedule + ready-wait), and `destroy()`/`_delete_inline_template` contend on the same lock — one slow pod boot blocks every other session's teardown process-wide. Not corruption, but it nullifies parallel eval fan-out (the exact Harbor/JetBrains use case). Fix shape: narrow the locked region to just the config-capture + client-construction window (the `ApiClient` snapshots its config at construction via `Configuration.get_default_copy()`), so the ready-wait runs unlocked.

### Note (already fixed in this PR)
The `k8s-agent-sandbox` version pin was `>=0.1.0,<0.2`, which installs a flat package layout missing every module the flavor imports — repinned to `>=0.3,<0.5` in this branch.

PR #4875, opened 2026-05-29
Improved step and pipeline hooks

### LINKED ISSUE #4468: Display failure/success hooks and their output

_No response_

### Feature Description

Display failure/success hooks and their output.

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

schustmi:
Add a hook system that runs user code at run and step lifecycle points and records each firing as a queryable `HookInvocation`. Four hooks at each scope: `on_start`, `on_success`, `on_failure`, `on_end`, plus `on_pause` and `on_resume` at run scope for dynamic pipelines. Step hooks fire per execution attempt (`on_start`/`on_end`) or once at the terminal outcome (`on_success`/`on_failure`). Hook failures are swallowed and captured on the record with `status=FAILED`. Logs produced inside a hook are captured separately and stored on the invocation. Call `run_hook(func, ...)` from inside a step or dynamic pipeline to record an arbitrary `CUSTOM` invocation, with `store_return=True` to materialize the return as output artifacts. Query via `Client().list_hook_invocations(...)` or `zenml hook-invocation list`, filterable by run, step, hook type, and name.

- **Static vs dynamic `@pipeline(on_*=...)`:** on a static pipeline the hook is a per-step default that each step inherits unless it sets its own, and is not invoked at run-level. On a dynamic pipeline it fires once at the run level instead. `on_pause`/`on_resume` are ignored on static pipelines.
- **`on_init`/`on_cleanup` stay separate:** they are deployment setup and teardown hooks that run once per deployment service lifecycle, not per run, so they fall outside the per-run/per-step invocation model and are not recorded as `HookInvocation`.

**Breaking changes:**
- `on_success`/`on_failure` hooks for dynamic pipelines now do not propagate to steps anymore and instead get invoked at run level.


strickvl:
We'll probably want something about hooks in our https://github.com/zenml-io/skills. No big update, but just pointer to them + updates where it feels appropriate.

schustmi:
> We'll probably want something about hooks in our https://github.com/zenml-io/skills. No big update, but just pointer to them + updates where it feels appropriate.

Will adjust once the PR is final and we're clear about how the public API will look like (still one open question and pending reviews)

docs/book/how-to/steps-pipelines/hooks.md
```diff
@@ -0,0 +1,218 @@
...
+| `on_start` | Propagates to each step's `on_start` default | Fires once at the pipeline level |
+| `on_success` | Propagates to each step's `on_success` default | Fires once at the pipeline level |
+| `on_failure` | Propagates to each step's `on_failure` default | Fires once at the pipeline level |
+| `on_end` | Propagates to each step's `on_end` default | Fires once at the pipeline level |
+
+For a static pipeline, a pipeline-level hook is a default that every step inherits
+where it has not set its own. No pipeline-level rows are produced. For a dynamic
+pipeline, the hook fires once at the run level and produces `RUN_*` rows. It fires
+on every run, including each invocation of a deployed pipeline. Dynamic pipeline
+users who want per-step defaults wire each `@step` directly.
```
bcdurak, line 47: Within the scope of this feature, I would say this is quite a change. There are two things that I don't really like about this if I understand this correctly:

- Static pipelines do not have any run-level hooks (while dynamic pipelines do)
- It should be difficult to set a default hook for dynamic pipelines with many steps.

Not sure what the alternative would be but I wanted to get your opinion on this. 

PS: Also I would avoid using the word `rows` here.

docs/book/how-to/steps-pipelines/hooks.md
```diff
@@ -0,0 +1,218 @@
...
+from zenml.enums import ExecutionStatus
+
+def on_end(): ...
+def on_end(status: ExecutionStatus): ...
+def on_end(status: ExecutionStatus, exception: Optional[BaseException] = None): ...
+```
+
+`status` is the terminal status of the step attempt or run. `exception` is set
+only when the status is a failure. Write the two-argument form with a default on
+`exception` so the same function works on both the success and failure paths.
```
bcdurak: I like this already. What do you think implementing some easy way to have access to output artifacts here? People can potentially load the results to write the alerts, or potentially tag them, attach metadata to them, etc?

Also, what happens if I implement the one-argument form but the step ends with a failure?

docs/book/how-to/steps-pipelines/hooks.md
```diff
@@ -0,0 +1,218 @@
...
+  `status=FAILED` and execution proceeds.
+
+## Init and cleanup hooks
+
+`on_init` and `on_cleanup` are pipeline setup and teardown hooks. They initialize
+and tear down shared run state rather than reacting to a single run or step
+outcome, so they follow different rules from the lifecycle hooks above. They are
+**not** recorded as `HookInvocation` records. When `on_init` fails, the run still
+records `RUN_START`, `RUN_END`, and `RUN_FAILURE`, but never a row for `on_init`
+itself. Find the root cause on `pipeline_run.exception_info`.
```
bcdurak, line 155: Is there a reason why we are not recording them as `HookInvocations`?

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
```diff
@@ -0,0 +1,292 @@
+#  Copyright (c) ZenML GmbH 2025. All Rights Reserved.
```
bcdurak:
```suggestion
#  Copyright (c) ZenML GmbH 2026. All Rights Reserved.
```
😅

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
```diff
@@ -0,0 +1,292 @@
...
+        target_column="id",
+        ondelete="CASCADE",
+        nullable=False,
+    )
+    step_run_id: Optional[UUID] = build_foreign_key_field(
+        source=__tablename__,
+        target="step_run",
+        source_column="step_run_id",
+        target_column="id",
+        ondelete="SET NULL",
```
bcdurak: Should this cascade as well?

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
bcdurak: This is the same for other files as well. I will just leave this comment here.

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class HookInvocationSchema(BaseSchema, table=True):
+    """SQL Model for hook invocations of pipeline runs."""
+
+    __tablename__ = "hook_invocation"
+
+    # Fields
+    hook_type: str = Field(nullable=False)
+    name: Optional[str] = Field(nullable=True)
```
bcdurak, line 54: As far as I can see people can set the name for their hook invocation. Do you think there should be a unique constraint on the much like the step runs and pipeline runs?

src/zenml/zen_server/routers/hook_invocations_endpoints.py
```diff
@@ -0,0 +1,201 @@
...
+        hook_invocation: The hook invocation to create.
+
+    Returns:
+        The created hook invocation.
+    """
+    pipeline_run = zen_store().get_run(
+        hook_invocation.pipeline_run_id, hydrate=False
+    )
+
+    # The caller can only link output artifact versions they are allowed to
```
bcdurak, line 125: I don't mind the check but it made me wornder whether this is even possible? For this to fail, the hook should store the artifacts but not itself. If this fails, we would also have untethered artifacts, right?

src/zenml/models/v2/core/hook_invocation.py
```diff
@@ -0,0 +1,292 @@
...
+        Returns:
+            the value of the property.
+        """
+        return self.get_resources().log_collection
+
+
+# ------------------ Filter Model ------------------
+
+
+class HookInvocationFilter(ProjectScopedFilter):
```
bcdurak, line 272: I think `start_time`, `end_time` and `status` would be great additional parameters here as well.

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -824,11 +829,28 @@ def run_pipeline(self) -> None:
...
-                    self._orchestrator.run_init_hook(snapshot=self._snapshot)
+                    try:
+                        self._orchestrator.run_init_hook(
+                            snapshot=self._snapshot
+                        )
+                    except BaseException as e:
+                        # The init hook itself is not tracked, but its failure
+                        # ends the run, so the run end and failure hooks fire
+                        # before the failure propagates.
+                        self._fire_terminal_run_hooks(
```
bcdurak, line 860: Just a small warning here, this might fire before the run is actually put on a `FAILED` state.

src/zenml/hooks/execution.py
```diff
@@ -0,0 +1,289 @@
...
+            # Output parsing runs after the hook succeeded, so a parsing error
+            # does not mark the invocation as FAILED or discard the hook's
+            # status.
+            if store_return:
+                outputs = _parse_hook_outputs(loaded_func, result)
+            return result
+        finally:
+            if track:
+                try:
+                    record_hook_invocation(
```
bcdurak, line 258: By the time we reach this point, the hook executes and we create the artifacts, however if this process fails (can fail due to a bad name that doesn't work well with the `HOOK_NAME_PATTERN `), we do not have anything to look back to and the resulting artifacts are orphaned.

src/zenml/orchestrators/step_runner.py
```diff
@@ -303,18 +326,17 @@ def run(
...
-                                    success_hook_source,
-                                    step_exception=None,
-                                )
+                        self._fire_step_hook(
+                            self.configuration.end_hook_source,
+                            HookType.STEP_END,
+                            step_environment,
+                            optional_args=(ExecutionStatus.COMPLETED, None),
+                        )
+                        self._fire_step_hook(
```
bcdurak, line 332: This should run a bit later, right? At least after the output validation.

```python
from typing import Tuple

def success_hook() -> None:
    print("SUCCESS_HOOK_RAN")


@step(on_success=success_hook)
def bad_output_step() -> Tuple[int, int]:
    return 1


@pipeline(enable_cache=False)
def success_before_outputs_pipeline() -> None:
    bad_output_step()


if __name__ == "__main__":
    success_before_outputs_pipeline()
```

If I do this, the success hook run but the step fails due to mismatched outputs.

bcdurak (commented):
I am sharing my initial findings. I will play with this even more and test things out. Outside of the comments, the changes look really good. I am excited to merge this in.

docs/book/how-to/steps-pipelines/hooks.md
schustmi: Yes. Because hook invocations are linked to pipeline runs, and those hooks are specifically implemented for deployments to warm-start pipelines runs by e.g. loading a machine learning model in memory once and not every run.

src/zenml/zen_server/routers/hook_invocations_endpoints.py
schustmi: No, someone could simply send a request and include any ID of artifacts they don't have access to. This is the case that this code is covering.

docs/book/how-to/steps-pipelines/hooks.md
schustmi: > - Static pipelines do not have any run-level hooks (while dynamic pipelines do)
> - 
Correct, because it is literally impossible for most of our orchestrators.

> It should be difficult to set a default hook for dynamic pipelines with many steps.

Agreed, it is difficult. But once you have pipeline-level hooks, can you think of any reasonable use case where you would need one specific hook for every step?

`rows` -> Agreed

docs/book/how-to/steps-pipelines/hooks.md
schustmi: You mean access to the output artifacts of the step? If so, for a failed step they wouldn't exist 😉 

I guess we could give them an (optional) step run in successful hooks?

src/zenml/hooks/execution.py
schustmi: Same with steps I think. We can validate the name early, but server failures we can't really handle I think.

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
schustmi: I thought about it, but didn't really see a reason why we would need to limit it. WDYT?

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
schustmi: I think steps are only cascade deleted because a run is deleted, so the behaviour won't change. But I'll adjust it, logically it makes sense I think

bcdurak:
I forgot to write this down, but it might be nice to have some CLI functionality around the hook invocations as well.

src/zenml/hooks/execution.py
bcdurak: Agreed about the server failures but the name is something that we can validate quite early. I would suggest doing that.

src/zenml/orchestrators/step_runner.py
schustmi: I don't know, lots of our logic there is quite messy tbh. If I move it after the output validation and the output validation fails, no hook will fire at all.

Currently, I just put them where they've always been. The alternative would be to re-write out entire error handling here (which I think is a good idea, don't get me wrong) but I think that's out of scope of this PR.

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: If this fails, actually the whole run stays in running state forever I think. See my comment on the step runner, I think similar here

docs/book/how-to/steps-pipelines/hooks.md
bcdurak: As discussed, resolving this now.

src/zenml/execution/pipeline/dynamic/runner.py
bcdurak: Back when I was reviewing it, I have tried to test this locally and I think the run was in a `FAILED` state, at least in the dynamic case.

src/zenml/orchestrators/step_runner.py
bcdurak: I see. We can leave it as it is right now, but this would be a nice improvement IMO.

src/zenml/zen_stores/schemas/hook_invocation_schemas.py
bcdurak: I was just thinking about the visual presentations. But I guess you are right. We can leave it as is.

bcdurak (approved):

PR #4882, opened 2026-06-01
fix(orchestrators): guard against IndexError for bare list/tuple step annotations

### LINKED ISSUE #4881: BUG: IndexError in StepRunner._parse_inputs when step parameter uses bare list annotation
When a step function parameter is annotated with the bare `list` type (no type argument, e.g. `def my_step(x: list) -> None`) and the step receives multiple input artifacts, `StepRunner._parse_inputs` raises `IndexError: tuple index out of range` at `step_runner.py` line 506.

The branch `elif issubclass(arg_type, (list, tuple)):` correctly triggers for bare `list`, but then immediately does `get_args(annotation)[0]` without checking whether `get_args` returned anything. `typing.get_args(list)` returns an empty tuple `()`, so indexing it raises `IndexError`.

Traceback:
```
File "step_runner.py", line 506, in _parse_inputs
    item_arg_type = get_args(annotation)[0]
IndexError: tuple index out of range
```

Repro: annotate a step parameter with `list` (not `List[int]` or `list[int]`) and connect multiple upstream artifacts to it. The fix is a one-line guard: only index `get_args` when the result is non-empty, falling back to the raw type otherwise.

devteamaegis:
## What's broken

When a step parameter is annotated with bare `list` or `tuple` (no type argument, e.g. `def my_step(x: list) -> None`) and the step receives multiple input artifacts, `StepRunner._parse_inputs` raises `IndexError: tuple index out of range` at line 506. The `issubclass(arg_type, (list, tuple))` branch fires correctly for bare `list`, but then unconditionally does `get_args(annotation)[0]`. Since `typing.get_args(list)` returns an empty tuple `()`, indexing it crashes.

## Why it happens

`get_args` returns an empty tuple for unparameterized generic types; there was no guard before indexing position 0.

## Fix

In `step_runner.py`, the index now only happens when `get_args(annotation)` is non-empty. If it is empty (bare `list` or `tuple`), `item_arg_type` stays as the raw `arg_type` set at the top of the branch, so each artifact is loaded with its own stored type — matching the behaviour of the `Any` branch above.

## Test

Added `test_bare_list_annotation_does_not_raise_index_error` to `tests/unit/orchestrators/test_step_runner.py`. It reproduces the exact expression that previously crashed and asserts that the guard produces a valid `item_arg_type` without raising.

Fixes #4881


src/zenml/orchestrators/step_runner.py
```diff
@@ -503,8 +503,14 @@ def _parse_inputs(
...
-                        item_arg_type = get_args(annotation)[0]
-                        item_arg_type = resolve_type_annotation(item_arg_type)
+                        _type_args = get_args(annotation)
+                        if _type_args:
+                            item_arg_type = resolve_type_annotation(
+                                _type_args[0]
+                            )
+                        # else: bare `list`/`tuple` with no type argument —
+                        # item_arg_type stays as arg_type (set above), which
+                        # causes each artifact to be loaded with its own type.
```
schustmi, line 513: In the else case, `item_arg_type` is `list`, no? This would cause downstream materializers to try loading as list and also failing. Instead, we should probably set `item_arg_type` to `Any` or `None` in the else case, so that we actually use the data type stored with the artifact.

tests/unit/orchestrators/test_step_runner.py
```diff
@@ -239,3 +239,24 @@ def test_loading_input_artifact_without_specified_data_type(
     )
     assert isinstance(data, int)
     assert data == 42
+
+
+def test_bare_list_annotation_does_not_raise_index_error():
```
schustmi, line 244: This test just mirrors your code and does not actually test any of the changes you implemented as part of the step runner. I'd suggest you mirror the tests in `test_base_step.py` and run a full pipeline with your suggested step annotation and make sure that works end-to-end.

schustmi (changes requested):

schustmi:
Thanks for the PR @devteamaegis. Left a few comments on how to improve it, but it for sure is a bug in the current code.

Can you please also base your PR on the `develop` branch (and use that as a base for the PR) and sign the CLA. Thanks!

schustmi:
Closed in #4929

PR #4888, opened 2026-06-02
Fix/mysql workspace db backpressure

safoinme:
## Describe changes

This PR adds OSS-side backpressure and DB-pressure mitigation for MySQL-backed ZenML workspaces.

- Adds `pool_timeout` support to SQL Zen Store configuration and Helm values so database connection checkout waits are bounded.
- Limits request timeout/backpressure semantics to deduplicated, idempotency-key-backed requests, avoiding hidden duplicate mutations for non-deduplicated requests.
- Makes expired API transaction cleanup bounded and MySQL-friendly with batch limits and `SKIP LOCKED`.
- Adds schema indexes and an Alembic migration for hot `artifact_version` and `step_run` lookup paths.

## Validation

- `pytest tests/unit/zen_stores/test_sql_store_utils.py tests/unit/zen_server/test_request_management.py`: 7 passed.
- `ruff check` on changed source/test/migration files: passed.
- `ruff format --check` on changed source/test/migration files: passed.

## Rollout Notes

The migration creates indexes on existing workspace databases. For large MySQL workspaces, validate migration duration and query plans in staging/canary before broad rollout.

## Pre-requisites

## Types of changes


src/zenml/zen_stores/migrations/versions/9210b7665982_add_pagination_indexes_for_pipeline_.py
```diff
@@ -0,0 +1,152 @@
...
+    """Check whether an index exists on the current database connection."""
+    connection = op.get_bind()
+    inspector = inspect(connection)
+    return any(
+        index["name"] == index_name
+        for index in inspector.get_indexes(table_name)
+    )
+
+
+def _create_index_if_missing(
```
Json-Andriopoulos: I like this a lot, replayable migrations are much easier to maintain. I would move those functions to a common migration utility and instruct agents to use sucu functions to create replayable migrations.

src/zenml/zen_stores/migrations/versions/c465032f0ec0_add_pagination_indexes_for_pipeline_.py
```diff
@@ -0,0 +1,152 @@
...
+    _create_index_if_missing(
+        "step_configuration",
+        "ix_step_configuration_step_run_id",
+        ["step_run_id"],
+    )
+
+    _create_index_if_missing(
+        "step_run",
+        "ix_step_run_project_id_pipeline_run_id_name_created_id",
+        ["project_id", "pipeline_run_id", "name", "created", "id"],
```
Json-Andriopoulos, line 82: Multi-field indices are used in a leftmost contiguous order. What type of query are you trying to optimize with this index?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2381,17 +2387,80 @@ def _get_api_transaction(
...
+    ) -> bool:
+        """Check whether a completed API transaction has expired."""
+        return (
+            api_transaction_schema.completed
+            and api_transaction_schema.expired is not None
+            and api_transaction_schema.expired < utc_now()
+        )
+
+    @staticmethod
+    def _claim_expired_api_transaction(
```
Json-Andriopoulos, line 2402: This is the epitome of a function that should have detailed documentation: Either add a detailed docstring or add an inline comment like:

```
# Claim the transaction with a conditional UPDATE instead of first loading it
# and then mutating it. This makes the claim atomic across multiple workers:
# only one worker can transition a still-completed, still-expired transaction
# back to pending state. If another worker already claimed it, the predicates
# no longer match and rowcount will be 0.
```

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2381,17 +2387,80 @@ def _get_api_transaction(
...
                     col(ApiTransactionSchema.completed),
                     col(ApiTransactionSchema.expired) < utc_now(),
                 )
+                .order_by(
+                    col(ApiTransactionSchema.expired),
+                    col(ApiTransactionSchema.id),
+                )
+                .limit(batch_size)
+            )
+            if self.config.driver == SQLDatabaseDriver.MYSQL:
```
Json-Andriopoulos, line 2475: This should also have a comment to explain why we do this 2-step deletion instead of a simple delete.where() query.

src/zenml/zen_stores/migrations/versions/c465032f0ec0_add_pagination_indexes_for_pipeline_.py
safoinme: This targets the slow-log `step_run` listing shape:

WHERE step_run.project_id = ?
  AND step_run.pipeline_run_id = ?
  AND step_run.name = ? / IN (...)
ORDER BY step_run.created, step_run.id
LIMIT ...

It was the top offender in the RDS slow logs, with thousands of executions and ~174k-380k rows examined for usually 0-2 rows returned.

The order is intentional: `project_id`, `pipeline_run_id`, and `name` are equality predicates and form the leftmost prefix; `created, id` then match the pagination order. It is not intended as a general-purpose `step_run` index, only for this hydrated step-run listing query shape.

src/zenml/zen_server/request_management.py
```diff
@@ -279,10 +280,16 @@ def _cleanup_expired_transactions(self) -> None:
...
 
         start_time = time.perf_counter()
-        zen_store().cleanup_expired_api_transactions()
+        deleted_count = zen_store().cleanup_expired_api_transactions(
+            batch_size=self._cleanup_batch_size
+        )
         duration = time.perf_counter() - start_time
 
-        logger.debug(f"Cleaning up expired transactions took {duration:.2f}s")
+        logger.debug(
```
Json-Andriopoulos, line 288: Since this runs periodically the log here can get spammy. I'd add a condition on the deleted_count or duration (log only when it takes too long or when it deletes more than N records?)

src/zenml/zen_stores/migrations/versions/c465032f0ec0_add_pagination_indexes_for_pipeline_.py
Json-Andriopoulos: One thing we need to take into account is that indices are not free. The more indices we add the more space we use and the slower updates become. I am wondering if in this case a more reusable index like (project_id, pipeline_run_id,created,id) would serve more queries (e.g. get run failing steps, get run steps, etc.) and still make the problematic query fast.

PR #4892, opened 2026-06-03
Validate artifact content hash when loading pickled artifacts

qxyuan853:
The cloudpickle fallback materializer records the SHA-256 of the stored file (compute_content_hash) and, when a content hash was recorded for the artifact version, checks the stored bytes against it before reading them back. Artifacts with no recorded hash load unchanged.

The recorded hash reaches the materializer through a new BaseMaterializer.expected_content_hash field, set in _load_artifact_from_uri, the model-metadata loader, and step-input loading.

Adds unit tests for matching-hash loads, content-hash mismatches, and the no-hash path.

## Describe changes

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/materializers/cloudpickle_materializer.py
+++ b/src/zenml/materializers/cloudpickle_materializer.py
@@ -13,8 +13,10 @@
 #  permissions and limitations under the License.
 """Implementation of ZenML's cloudpickle materializer."""
 
+import hashlib
+import hmac
 import os
-from typing import Any, ClassVar, Tuple, Type
+from typing import Any, ClassVar, Optional, Tuple, Type
 
 import cloudpickle
 
@@ -31,6 +33,9 @@ logger = get_logger(__name__)
 
 DEFAULT_FILENAME = "artifact.pkl"
 DEFAULT_PYTHON_VERSION_FILENAME = "python_version.txt"
+
+# Chunk size used when streaming the stored artifact to compute its hash.
+_HASH_CHUNK_SIZE = 65536
 
 
 class CloudpickleMaterializer(BaseMaterializer):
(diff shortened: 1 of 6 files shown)


schustmi:
@qxyuan853 Can you please rebase on the develop branch and use that as the base of your PR.

qxyuan853:
> @qxyuan853 Can you please rebase on the develop branch and use that as the base of your PR.

Hi, I have rebased the PR on the develop branch.

src/zenml/materializers/cloudpickle_materializer.py
```diff
@@ -69,11 +81,22 @@ def load(self, data_type: Type[Any]) -> Any:
                 "versions. Attempting to load anyway..."
             )
 
-        # load data
         filepath = os.path.join(self.uri, DEFAULT_FILENAME)
         with self.artifact_store.open(filepath, "rb") as fid:
```
schustmi, line 85: Instead of loading the file again, can we potentially compute this already during `save()` and store it on the materializer instance?

qxyuan853: Sounds good. What do you think of this approach (https://github.com/zenml-io/zenml/pull/4946)?

I pushed a new commit (https://github.com/zenml-io/zenml/pull/4946/changes/58b9109d2597cee6e1624a1fe012d9665a029566) that computes the hash while save() writes the bytes, using a lightweight _HashingWriter wrapper around the store handle. The hash is cached on the materializer, so compute_content_hash() can return it directly without reading the file again.

It only falls back to re-reading when hashing an instance that was not written by the current materializer. The hash is still computed over the actual stored bytes, so load() can verify them before unpickling.

schustmi:
Closed in favor of #4946

PR #4900, opened 2026-06-05
Fix breaking CI checks

### LINKED ISSUE #4886: Clean up the failing tests on the CI

_No response_

### System Information

_No response_

### What happened?

.

### Reproduction steps

1.
2.
3.
...

### Relevant log output

```shell

```

### Code of Conduct

Json-Andriopoulos:
## Describe changes

- [x] Fixes local MLFLow tracking to use the recommended SQLite DB backend (Breaking since mlflow==3.13.0)
- [x] Fixes failing pyarrow mypy checks
- [x] Fixes MLFlow/MLServer version incompatibility (MLServer deprecated for MLFlow >= 3.13.0
- [x] Fixes mypy/croniter false positive failure

## Breaking Changes:

The PR introduces breaking changes for locally running MLFLow experiment trackers. As per the updated docs:

> If you use the default local MLflow configuration without setting a `tracking_uri`, ZenML stores new MLflow tracking metadata in a SQLite database at `<LOCAL_ARTIFACT_STORE>/mlflow.db` and stores MLflow artifacts in `<LOCAL_ARTIFACT_STORE>/mlflow_artifacts`.
Earlier ZenML versions used MLflow's file store at `<LOCAL_ARTIFACT_STORE>/mlruns` for this default local setup. Existing local runs in that directory are not migrated automatically and won't appear in the new default SQLite-backed MLflow UI. The files remain on disk and can still be inspected separately with MLflow's file-store backend.


src/zenml/integrations/mlflow/services/mlflow_deployment.py
```diff
@@ -115,24 +116,35 @@ class MLFlowDeploymentConfig(LocalDaemonServiceConfig):
...
 
         Raises:
-            ValueError: if mlserver packages are not installed
+            ValueError: if mlserver packages are not installed or supported.
         """
         if mlserver is True:
-            # For the mlserver deployment, the mlserver and
-            # mlserver-mlflow packages need to be installed separately
-            # because they rely on an older version of Pydantic.
+            if version.parse(MLFLOW_VERSION) >= version.parse("3.13.0"):
```
schustmi, line 134: Do you think that there could be a case where this validation runs in a different environment than the actual deployment, meaning there could be different MLFlow versions installed?

src/zenml/integrations/mlflow/experiment_trackers/mlflow_experiment_tracker.py
```diff
@@ -53,6 +53,8 @@
 MLFLOW_TRACKING_TOKEN = "MLFLOW_TRACKING_TOKEN"
 MLFLOW_TRACKING_INSECURE_TLS = "MLFLOW_TRACKING_INSECURE_TLS"
 MLFLOW_BACKEND_STORE_URI = "_MLFLOW_SERVER_FILE_STORE"
+LOCAL_MLFLOW_BACKEND_FILENAME = "mlflow.db"
```
schustmi, line 56: How do the changes in this file work with the existing data stored in the file-based storage that we used on previous versions?

I'm assuming it will just ignore them and write to the DB from the moment users upgrade, correct? If yes, that's somewhat "breaking"/should be mentioned in the docs/release notes I guess?

Json-Andriopoulos: Good catch, yes I will add a note

src/zenml/integrations/mlflow/services/mlflow_deployment.py
Json-Andriopoulos: Hmm seems more like a corner case but good catch. We should guard against that scenario as well.

schustmi (approved):

PR #4903, opened 2026-06-08
feat: updated logging instumentation to use newer otel version 1.40.0

### LINKED ISSUE #4753: Bump the `opentelemetry-sdk` version

amitvikramraj:


pyproject.toml
```diff
@@ -100,10 +100,10 @@ server = [
 ]
 otel = [
     # Install otel dependencies to enable OpenTelemetry instrumentation in the server
-    "opentelemetry-exporter-otlp-proto-http>=1.38.0",
-    "opentelemetry-instrumentation-fastapi>=0.59b0",
-    "opentelemetry-instrumentation-requests>=0.59b0",
-    "opentelemetry-instrumentation-sqlalchemy>=0.59b0",
+    "opentelemetry-exporter-otlp-proto-http>=1.40.0",
```
bcdurak: What do you think about pinning these requirements? In the main package, these are already pinned and `1.41.0` already has some changes that would break it for us.

bcdurak: Perhaps we can use ranges such as `1.40.0,<1.41.0`?

src/zenml/zen_server/otel.py
```diff
@@ -287,24 +280,38 @@ def _configure_logs(
...
-        # init_logging() attaches ZenML's console and log-storage handlers to
-        # the root logger, and server startup configures uvicorn to propagate
-        # there too. Attach OTel to the same logger so one logging pipeline
-        # exports ZenML, uvicorn, and third-party records with the same filters.
+
+        # Edit the OTel Log Handler level and add our context filters to it.
         root_logger = logging.getLogger()
-        root_logger.addHandler(otel_handler)
+        for handler in root_logger.handlers:
+            if isinstance(handler, LoggingHandler):
```
bcdurak: A small nitpick here: We mutate not just the handler that is added by the `instrument(..)` logic, but all the handlers that are of type `LoggingHandler`. Since we use this logic during server spin up, this should be ok for now but we can maybe try to find a cleaner way here.

bcdurak: Perhaps we can use the old logic by using handlers again through `from opentelemetry.instrumentation.logging.handler import LoggingHandler
`. WDYT?

amitvikramraj: No, no, the type `LoggingHandler` is specific to OpenTelemetry, which is a subclass of stdlib `logging.Handler`, and other logging handlers would be of type `logging.Handler` not `LoggingHandler`. So this should not cause any problem. Basically, the newer logic is the same as the older one. It just looks different because of the breaking change in the OTel version

pyproject.toml
amitvikramraj: Yes, we could. I just thought that since these are optional dependencies, it wouldn't make much sense to pin them. But if you give the go-ahead, I can pin them!

bcdurak: What do you mean by optional dependencies? Don't we have to install them to be able to use otel on the server side?

src/zenml/zen_server/otel.py
bcdurak: I do understand the difference but it is possible other dependencies might install their own `LoggingHandler` from otel, right? In that case, we will be mutating their handlers as well.

pyproject.toml
amitvikramraj: I meant optional, as in from the 'zenml' package perspective

src/zenml/zen_server/otel.py
amitvikramraj: TBH, I do not know. Unless a package heavily customises OTel's LoggingHandler, this is highly unlikely.

bcdurak:
@amitvikramraj I have updated my branch, so I have updated yours as well to solve the linting issues.

src/zenml/zen_server/otel.py
```diff
@@ -290,10 +293,11 @@ def _configure_logs(
         otel_handler = LoggingHandler(
             level=get_logging_level().value,
             logger_provider=logger_provider,
+            log_code_attributes=True,
         )
         # Attach the ZenML filters that add structlog contextvars
         # and step name to the log record to the OTel handler.
-        otel_handler = add_zenml_filters(otel_handler)  # type: ignore[assignment]
+        add_zenml_filters(otel_handler)  # type: ignore[assignment]
```
bcdurak:
```suggestion
        add_zenml_filters(otel_handler)
```

bcdurak (changes requested):

bcdurak (approved):

PR #4904, opened 2026-06-08
Run arbitrary (non-)python commands as steps

schustmi:
This PR adds code to run arbitrary commands as steps in a pipeline using the `CommandStep(...)` class.
Command steps do **not** require ZenML in the execution environment.

**Example:**
```python
from zenml import CommandStep, pipeline

def some_func():
  print("hello from python function")

bash_command_step = CommandStep(command=["bash", "-c", "echo 'hello from bash"])
function_command_step = CommandStep(some_func)

@pipeline(dynamic=True, enable_cache=False)
def command_step_pipeline() -> None:
    bash_command_step()
    function_command_step.submit()

if __name__ == "__main__":
    command_step_pipeline()
```

**Limitations**:
- Command steps do not support inputs and outputs
- Your code is not downloaded into the execution environment
- Logs of command steps are not tracked
- Steps in static pipelines without step operator are not supported

### DIFF
--- a/docs/book/how-to/steps-pipelines/command_steps.md
+++ b/docs/book/how-to/steps-pipelines/command_steps.md
@@ -23,6 +23,29 @@ if __name__ == "__main__":
 
 The command is whatever you would type in a shell, split into a list. The image just needs to contain the binary you invoke. Shell features like pipes and `&&` need an explicit shell, for example `["bash", "-c", "a | b"]`.
 
+## Running a Python function
+
+Instead of a command, you can pass a Python function. ZenML extracts the source code of the function and runs it with `python -c`:
+
+```python
+from zenml import CommandStep
+
+def train() -> None:
+    import json
+
+    print(json.dumps({"status": "training"}))
+
+train_step = CommandStep(command=train)
+```
+
+The extracted source code is the entire program, nothing around the function travels with it. This means the function must be self-contained:
+
+- All imports must happen inside the function body. Module-level imports are not available.
+- The function must not reference module-level variables, constants, or other functions.
+- The function must not take any parameters, be decorated, or be defined inside another function.
+
+ZenML rejects parameters, decorators, and references to enclosing functions when the step is created. The execution environment only needs a `python` binary on the `PATH`, ZenML does not have to be installed.
+
 ## Running without ZenML in the image
 
 A regular step runs your Python function inside the container, so the image has to contain `zenml`, your step code, and all of its dependencies. A command step does not. ZenML treats the command as a black box and never imports anything inside the container, so you can run an image that does not have `zenml` installed.
@@ -74,3 +97,4 @@ Environment variables and secrets are passed to the command as environment varia
 - Logs of command steps are not tracked by ZenML. They stay in the backend's native logging (for example CloudWatch or pod logs).
 - Step hooks are not allowed.
 - Steps in static pipelines without a step operator are not supported.
+- Functions passed as commands must be self-contained (see [Running a Python function](#running-a-python-function)).
--- a/docs/book/how-to/steps-pipelines/command_steps.md
+++ b/docs/book/how-to/steps-pipelines/command_steps.md
@@ -46,7 +46,7 @@ The extracted source code is the entire program, nothing around the function tra
 
 ZenML rejects parameters, decorators, and references to enclosing functions when the step is created. The execution environment only needs a `python` binary on the `PATH`, ZenML does not have to be installed.
 
-## Running without ZenML in the image
+## Running without ZenML in the image (dynamic pipelines only)
 
 A regular step runs your Python function inside the container, so the image has to contain `zenml`, your step code, and all of its dependencies. A command step does not. ZenML treats the command as a black box and never imports anything inside the container, so you can run an image that does not have `zenml` installed.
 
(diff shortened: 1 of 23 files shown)


docs/book/how-to/steps-pipelines/command_steps.md
```diff
@@ -0,0 +1,101 @@
...
+train = CommandStep(
+    command=["python", "train.py"],
+    step_operator="sagemaker",
+    settings={
+        "docker": DockerSettings(skip_build=True, parent_image="my-registry/train:latest")
+    },
+)
+```
+
+The image must be pullable by the backend and already carry your code and libraries. The container owns its own credentials for anything it accesses.
```
bcdurak: Like we discussed, I am leaving a reminder here to write a warning.

src/zenml/steps/base_step.py
```diff
@@ -311,6 +311,15 @@ def load_from_source(cls, source: Union[Source, str]) -> "BaseStep":
                 f"got `{type(obj).__name__}`."
             )
 
+    @classmethod
+    def _load_from_source(cls) -> "BaseStep":
```
bcdurak, line 315: The name and the docstring of this method is a bit confusing IMO. We don't really deal with a source class here right?

src/zenml/steps/command_step.py
```diff
@@ -0,0 +1,190 @@
...
+            runtime_parameters: Runtime parameters passed to the step.
+
+        Raises:
+            StepInterfaceError: If success or failure hooks are configured.
+        """
+        super()._validate_configuration(
+            config=config, runtime_parameters=runtime_parameters
+        )
+
+        if config.failure_hook_source or config.success_hook_source:
```
bcdurak: If step-level hooks are not supported, should we also reject `start_hook_source` and `end_hook_source` here?

src/zenml/config/compiler.py
```diff
@@ -574,6 +579,20 @@ def _compile_step_invocation(
...
                 skip_input_validation=skip_input_validation,
             )
+
+        if step_configuration_overrides.command:
+            # Do not allow explicit hooks on a command step. In case they get
+            # applied from the pipeline level, we do not fail but just ignore
+            # them. Otherwise, the hook defaults from the pipeline would be
+            # unusable if the pipeline contains a command step.
+            if (
+                step_configuration_overrides.failure_hook_source
```
bcdurak, line 589: Same as before with the end and start hooks.

src/zenml/integrations/azure/step_operators/azureml_step_operator.py
```diff
@@ -226,9 +227,14 @@ def submit(
             azureml_client, settings, default_compute_name=f"zenml_{self.id}"
         )
 
+        job_name_prefix = str(info.step_run_id)[:8]
+        job_name = (
```
bcdurak, line 231: Not sure if this is going to be very readable. Maybe we can put the step_name before the run name here?

src/zenml/steps/command_step.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+    if func.__name__ == "<lambda>":
+        raise StepInterfaceError(
+            "Lambdas cannot be used as command step functions."
+        )
+
+    if hasattr(func, "__wrapped__"):
+        raise StepInterfaceError(
+            f"The command step function `{func.__name__}` is decorated."
```
bcdurak:
```suggestion
            f"The command step function `{func.__name__}` cannot be decorated."
```

src/zenml/steps/command_step.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+    if func.__closure__ is not None:
+        raise StepInterfaceError(
+            f"The command step function `{func.__name__}` references "
```
bcdurak:
```suggestion
            f"The command step function `{func.__name__}` cannot reference "
```

src/zenml/steps/command_step.py
```diff
@@ -0,0 +1,190 @@
...
+            "variables from an enclosing function."
+        )
+
+    if inspect.signature(func).parameters:
+        raise StepInterfaceError(
+            f"The command step function `{func.__name__}` has parameters."
```
bcdurak:
```suggestion
            f"The command step function `{func.__name__}` cannot have parameters."
```

src/zenml/steps/command_step.py
```diff
@@ -0,0 +1,190 @@
...
+        source = textwrap.dedent(inspect.getsource(func))
+    except (OSError, TypeError):
+        raise StepInterfaceError(
+            "Failed to get the source code of the command step function "
+            f"`{func.__name__}`."
+        )
+
+    if source.lstrip().startswith("@"):
+        raise StepInterfaceError(
+            f"The command step function `{func.__name__}` is decorated."
```
bcdurak:
```suggestion
            f"The command step function `{func.__name__}` cannot be decorated."
```

bcdurak (approved):
Leaving a bunch of nitpicky comments 😄 Feel free to ignore them as you see fit.

src/zenml/steps/base_step.py
schustmi: True, but the important part here is that it came from a source, which means the command step needs to set some default values. The source already got resolved to the class this is called on, but i think it's still useful to know where it's coming from/why we need to set the defaults.

docs/book/how-to/steps-pipelines/command_steps.md
schustmi: Done, added it in the headline actually.

PR #4907, opened 2026-06-08
Add Cloudflare integration: R2 artifact store, container registry, sandbox, and orchestrator

htahir1:
## Summary

The complete `cloudflare` integration in one PR — four components that make Cloudflare a first-class ZenML provider:

- **R2 artifact store** (`artifact_stores/r2_artifact_store.py`) — S3-compatible artifact store backed by Cloudflare R2, with the R2 endpoint handling (`https://<account_id>.r2.cloudflarestorage.com`) behind familiar s3-style configuration.
- **Container registry** (`flavors/cloudflare_container_registry_flavor.py`).
- **Sandbox flavor** (`sandboxes/cloudflare_sandbox.py`) — runs isolated/LLM-generated code in Cloudflare Sandboxes via a bridge Worker over plain HTTP+SSE; no TypeScript on our side, no Python dependency beyond `httpx`.
- **Orchestrator flavor** (`orchestrators/cloudflare_orchestrator.py`) — runs each pipeline step inside a Cloudflare Sandbox through the same bridge. Folded in from the former `feature/sandbox-cloudflare-orchestrator` branch so the whole integration lands in one PR. See [Orchestrator](#orchestrator-early-stage) for its maturity caveats.

## Sandbox: how it works

ZenML talks to the Cloudflare-provided **bridge Worker** (deployed once via `npx wrangler deploy` from `cloudflare/sandbox-sdk`). Steps do `with Client().active_stack.sandbox.create_session() as s: s.exec(...)`.

**Auth crosses three boundaries:**
1. **ZenML → bridge** (this PR): static bearer token (`SANDBOX_API_KEY`) stored as a ZenML `SecretField` on the component; https enforced.
2. **Bridge → Cloudflare account**: the Worker's own deploy-time binding — ZenML never sees account-level keys.
3. **Sandbox → external world**: Cloudflare's outbound proxy, documented seam, not v1.

**No service connector in this PR (deliberate):** an earlier draft included one, but it only passed static credentials through (no short-lived brokering) and its top-level boto3 import meant connector-type registration silently failed on servers built without the AWS extras. Components take credentials directly; a proper connector using Cloudflare's `temp-access-credentials` API (scoped, short-lived R2 keys) is a follow-up.

**Design decisions worth reviewer attention:**
- Only idempotent methods retry (GET/PUT/DELETE on 429/502/503/504); POSTs (create/exec) never retry — a failed POST may already have executed, so retrying risks double execution. DELETE treats 404/410 as success so destroy retries can't dead-loop.
- `destroy()` raises on failure and keeps the handle open for retry; `close()` is terminal for the local handle but leaves the remote sandbox running until `destroy()` or the bridge TTL; local SSE streams die with the handle.
- Upload size is checked against the bridge's 32 MiB body limit before reading the file; remote paths are sanitized against `..` traversal.

## Orchestrator (early-stage)

The orchestrator flavor runs each step inside a Cloudflare Sandbox via the same bridge Worker. It is **explicitly a proof of concept** and labelled as such in the code:

- **Synchronous, client-driven.** It drives the sandboxes from the client and blocks until each step finishes (`is_synchronous` returns `True`); there's no async submission or server-side status polling yet.
- **Owns its own bridge connection** rather than sharing the sandbox component's.
- **No dedicated tests yet** and **not re-validated against a live bridge since consolidation** — it's here so the full integration is reviewable in one place, not because it's production-ready. Hardening (async submission, status/cancel, resource translation) is follow-up before it graduates from POC.

## Verified end-to-end

Deployed a real bridge Worker and ran a 3-step ZenML pipeline spinning up 2 sandboxes (~20s): env injection round-trips, exit codes propagate, sandboxes destroyed cleanly. The example pipeline uses the R2 artifact store from this same PR. _(The store/registry/sandbox path is what was live-verified; the orchestrator flavor still needs a live re-validation pass — see above.)_

## Test plan

- 64 sandbox unit tests (`httpx.MockTransport`, no live Cloudflare in CI) — SSE parsing, retry policy per method, path sanitization, session lifecycle/leak cleanup, file caps, destroy/close semantics
- Integration tests for the store and flavors under `tests/integration/integrations/cloudflare/`
- Docs pages for the store, registry, and sandbox
- Orchestrator flavor: POC, no dedicated tests yet (see above)

## Merge order (stack)

```
develop  (✅ #4866 Sandbox core, #4868 PydanticAI example already merged)
├─ #4867 Modal flavor
│  └─ #4869 Harbor example
├─ #4870 k8s Agent Sandbox flavor
└─ this PR — Cloudflare integration (R2 store / registry / sandbox / orchestrator)
```

Base is now `develop` (auto-retargeted after #4866 merged), so the diff shows only the Cloudflare work.


htahir1:
Thanks @strickvl — addressed in `328b55089b`.

**Fixed:**
- **Logging context on direct `close()`**: `CloudflareSandboxSession.close()` now closes the ZenML logging context in a `finally` (the second call from `__exit__` is a no-op). Two lifecycle tests added — direct `close()` ends the log source, and `close()` + context-manager exit doesn't raise. Suite is at 59 tests.

**Relayed to core (#4866) rather than fixed here** — both of your must-fixes are in `src/zenml/sandboxes/`, which this PR deliberately doesn't touch:
- base `close()` vs `__exit__` logging-context ownership (`session.py`)
- LocalSandbox not tracking/terminating spawned `Popen` processes on `close()`/`destroy()` (`local_sandbox.py`), plus the suggested kill-while-running test
- the `Stack.from_components` single-vs-list sandbox question (deprecated path in core `stack.py`)

**On the product question**: yes, "close() keeps the provider sandbox alive, TTL is the backstop" is the intentional v1 contract — `close()` exists for detach-then-`attach()` flows, `destroy()` is the terminator. The examples and docs were updated (on the modal/example branches) to `destroy()` in one-shot flows so users don't pay for the distinction by accident.

htahir1:
Thanks @strickvl — both findings were real; fixed in `a7591423fd`.

- **close() leaked live SSE streams**: the session now tracks its processes and `kill()`s unfinished ones in `_close()`, so the local streaming response + pump thread die with the handle instead of lingering until the bridge timeout. Regression test added (`test_close_kills_unfinished_exec_streams`).
- **destroy() swallowed delete failures**: it now raises `RuntimeError` on a failed `delete_sandbox()` and keeps the handle (including bridge-session state) **open** so destroy can be retried — no more fake success while the sandbox keeps running. Test added covering raise → handle open → retry succeeds.

On the question: destroy is no longer best-effort — it either terminates or raises, matching the base docs' "terminates the sandbox" promise. (The test-harness `pymysql` import issue you hit is a known local-env quirk; the unit suite runs with `PYTHONPATH=src pytest tests/unit/integrations/cloudflare/` — 61 passing on this head.)

htahir1:
Thanks @strickvl — both taken, fixed in `32d71e7629` (cloudflare) and on the core branch (local sandbox).

- **Upload memory spike**: `_upload_file()` now checks `os.path.getsize()` against the 32 MiB bridge limit before reading anything, so an oversized file raises immediately instead of being loaded into memory just to be rejected. Regression test added using a sparse file (`truncate`, no real I/O).
- **Unreaped SIGKILLed children**: `LocalSandboxSession._close()` now `wait()`s after the `kill()` escalation so the child doesn't linger as a zombie until GC. (Skipped the SIGTERM-ignoring-child test — a `trap '' TERM` child plus two timeout windows makes for a slow/flaky unit test; the existing close-terminates-child test covers the path.)

htahir1:
Thanks @strickvl — both must-fixes taken, fixed in `5073f4c72d`.

- **Secrets on the remote command line**: you're right that the `env KEY=VALUE` argv inlining was the wrong trade-off. The bridge has no per-exec env API, so `env=` on `session.exec(...)` now raises `NotImplementedError` with a pointer at the session-scoped `sandbox_environment` setting (which travels via the bridge session API, never on a command line). Docs updated; test asserts no request is even issued.
- **Unbounded pump buffers**: capped at 100k lines per stream — past that, oldest undrained lines drop with a one-time warning, so a `wait()`-only caller of a noisy command can't OOM the step process. Regression test added.

**Not taking**: streaming downloads with byte counting — the bridge enforces its 32 MiB body cap server-side, so the in-memory read is bounded by construction; client-side double-enforcement would be plumbing without a failure mode it prevents.

**On the question**: the bridge is Cloudflare's reference Worker (`cloudflare/sandbox-sdk`), so the API isn't ours to extend — a first-class per-exec env field would be an upstream request. If/when it lands, re-enabling `env=` is a small, contained change.

PR #4919, opened 2026-06-10
Fix the rate-limiting

stefannica:
## Summary

Fixes ZenML server login rate limiting behind reverse proxies by removing direct trust in raw `X-Forwarded-For` headers and making Uvicorn proxy-header handling explicit in the Helm chart.

The limiter keys failed login attempts by client IP. In Kubernetes/Ingress deployments, ZenML usually receives traffic from an ingress controller, gateway, or other proxy rather than directly from the client. Proxies may add an `X-Forwarded-For` header to preserve client IP information:

```text
X-Forwarded-For: <client-ip>, <proxy-hop-1>, <proxy-hop-2>
```

However, clients can also send this header themselves. If an ingress/gateway appends to an existing value instead of sanitizing or replacing it, part of the chain can be attacker-controlled:

```text
X-Forwarded-For: <spoofed-ip>, <real-client-ip>
```

Parsing this header in ZenML application code is unsafe. The correct source of truth is `request.client.host`, after Uvicorn has applied its configured proxy-header trust policy. If proxy headers are disabled, Uvicorn ignores `X-Forwarded-*`; if enabled, Uvicorn resolves the client host according to `--forwarded-allow-ips`.

This PR updates rate limiting and device authorization to use the ASGI/Uvicorn-resolved client host instead of reading raw forwarded headers directly. It also makes proxy-header behavior configurable in Helm and disables proxy-header trust by default for safer Kubernetes deployments.

## Changes

- Updated the login rate limiter to use `request.client.host` as resolved by Uvicorn/FastAPI.
- Removed direct `X-Forwarded-For` parsing from the rate limiter.
- Removed direct `X-Forwarded-For` parsing from device authorization IP/geolocation handling.
- Added regression coverage that verifies raw `X-Forwarded-For` does not override the ASGI client host.
- Made Uvicorn proxy-header behavior configurable in the Helm chart:
  - `server.proxyHeaders.enabled`
  - `server.proxyHeaders.forwardedAllowIPs`
- Changed the Helm default to disable proxy-header trust:
  - `server.proxyHeaders.enabled: false`
- Render `--no-proxy-headers` explicitly when proxy headers are disabled.
- Render `--proxy-headers --forwarded-allow-ips <value>` when proxy headers are enabled.
- Added a Helm warning explaining that proxy headers should only be enabled when:
  - `forwardedAllowIPs` is set to a strict trusted proxy IP/CIDR list, or
  - the ingress/gateway sanitizes or overwrites incoming `X-Forwarded-*` headers before forwarding to ZenML.

Side changes:
- Updated Helm notes to report Gateway API URLs correctly.
- Fixed a potential `500 Internal Server Error` in the token endpoint.

## Breaking Changes

This changes the default behavior for Helm deployments. The chart now renders `--no-proxy-headers` by default, so Uvicorn will ignore `X-Forwarded-*` headers unless the operator explicitly enables them.

Previously, the server image started Uvicorn with `--proxy-headers --forwarded-allow-ips "*"`, which trusted forwarded headers from any peer. That made deployments more likely to see the original client IP automatically, but it was unsafe unless the ingress/gateway sanitized client-supplied `X-Forwarded-For` headers first.

With the new default:

```yaml
server:
  proxyHeaders:
    enabled: false
```

ZenML will see the direct peer address, usually the ingress/gateway IP in Kubernetes. This is safer against forged `X-Forwarded-For` values, but it may group many users into the same IP-based rate-limit bucket.

Deployments that need ZenML to see real client IPs must opt in explicitly:

```yaml
server:
  proxyHeaders:
    enabled: true
    forwardedAllowIPs: "<trusted proxy IPs or CIDRs>"
```

Only use `forwardedAllowIPs: "*"` if the ingress/gateway reliably strips, overwrites, or sanitizes incoming `X-Forwarded-*` headers before forwarding to ZenML.

For ingress-nginx, align this with its forwarded-header settings. If nginx is the public edge, avoid trusting arbitrary upstream forwarded headers; if it is behind another trusted load balancer, configure `use-forwarded-headers`, `enable-real-ip`, and `proxy-real-ip-cidr` so only trusted upstream ranges can supply client IP headers. ingress-nginx documents these settings in its ConfigMap options. Source: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/

For Envoy / Envoy Gateway, configure trusted client IP detection at the gateway/proxy layer, for example via trusted hops or trusted CIDRs, so Envoy determines the client address from a trusted chain rather than accepting arbitrary client-provided XFF. Envoy’s XFF handling is based on trusted proxy configuration; the app should consume the resolved downstream/client address, not parse raw XFF itself. Source: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers.html#x-forwarded-for

## Pre-requisites

## Types of changes

### DIFF
--- a/helm/templates/server-deployment.yaml
+++ b/helm/templates/server-deployment.yaml
@@ -129,6 +129,8 @@ spec:
             - --proxy-headers
             - --forwarded-allow-ips
             - {{ $server.proxyHeaders.forwardedAllowIPs | quote }}
+            {{ else }}
+            - --no-proxy-headers
             {{ end }}
             - --port
             - "8080"
--- a/helm/templates/server-deployment.yaml
+++ b/helm/templates/server-deployment.yaml
@@ -121,6 +121,19 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ $server.image.repository }}:{{ $server.image.tag | default .Chart.Version }}"
           imagePullPolicy: {{ $server.image.pullPolicy }}
+          args:
+            - uvicorn
+            - zenml.zen_server.zen_server_api:app
+            - --no-server-header
+            {{ if $server.proxyHeaders.enabled }}
+            - --proxy-headers
+            - --forwarded-allow-ips
+            - {{ $server.proxyHeaders.forwardedAllowIPs | quote }}
+            {{ end }}
+            - --port
+            - "8080"
+            - --host
+            - "0.0.0.0"
           volumeMounts:
             - name: zenml-config
               mountPath: /zenml/.zenconfig
--- a/helm/templates/NOTES.txt
+++ b/helm/templates/NOTES.txt
@@ -20,7 +20,20 @@ An enrollment key has been auto-generated for your server. Please use the follow
 {{- end }}
 
 {{- else }}
-{{- if $server.ingress.enabled }}
+{{- if $server.gateway.enabled }}
+{{- if $server.gateway.host }}
+{{- $gatewayScheme := "http" }}
+{{- if eq $server.gateway.sectionName "https" }}
+{{- $gatewayScheme = "https" }}
+{{- end }}
+
+Please access the ZenML server at the following URL {{- if .Release.IsInstall }} to activate the server and create an initial admin user account {{- end }}:
+
+  {{ $gatewayScheme }}://{{ $server.gateway.host }}{{ $server.gateway.path }}
+
+{{- else }}
+{{- end }}
+{{- else if $server.ingress.enabled }}
 {{- if $server.ingress.host }}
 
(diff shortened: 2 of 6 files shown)


schustmi (approved):

helm/templates/server-deployment.yaml
amitvikramraj, line 1: The Helm chart changes look good. Although now the app startup CMD, `uvicorn ....`, is tied to the Helm chart. And if someone, for whatever reason, changes the command `CMD ["uvicorn", ...]` in zenml Docker image, I guess they will have to update it here as well. Although it looks unlikely.

amitvikramraj (approved):

PR #4926, opened 2026-06-11
Kubernetes sandbox

schustmi:
This PR adds a `kubernetes` sandbox flavor that backs each sandbox session with a dedicated pod on a Kubernetes cluster. Sessions support streamed command execution and re-attaching to a running session via `attach(session_id)`.

**Example:**
```bash
zenml integration install kubernetes
zenml sandbox register k8s-sb --flavor=kubernetes --connector=<CONNECTOR_NAME>
zenml stack update --sandbox k8s-sb
```

**Limitations:**
- `snapshot()` / `restore()` and file upload/download are not supported.
- `kill()` only stops streaming output. The command keeps running until the pod is deleted.
- Session pod images must provide `/bin/sh`.

### DIFF
--- a/docs/book/component-guide/sandboxes/README.md
+++ b/docs/book/component-guide/sandboxes/README.md
@@ -108,6 +108,7 @@ Every value you put in a Session's environment (`sandbox_environment`) is **read
 ## Available flavors
 
 - **[Local](local.md)** — subprocess-based; **no isolation**; built-in. Intended for examples, unit tests, and development against the abstraction.
+- **[Kubernetes](kubernetes.md)** — pod-backed sessions executed through Kubernetes exec.
 
 ## Develop a custom Sandbox
 
--- a/docs/book/component-guide/sandboxes/kubernetes.md
+++ b/docs/book/component-guide/sandboxes/kubernetes.md
@@ -33,7 +33,7 @@ Some configuration options for the Kubernetes sandbox can only be set through th
 The following configuration options can be set either through the sandbox config or overridden using `KubernetesSandboxSettings`:
 
 - **`sandbox_environment`**: Environment variables to set in session pods and command executions.
-- **`image`** (default: "python:3.11-slim"): The container image used for session pods. The image must provide `/bin/sh`, so distroless images are not supported.
+- **`image`** (default: "python:3.11-slim"): The container image used for session pods. The image must provide `/bin/sh` and `base64`, so distroless images are not supported.
 - **`pod_settings`**: Node selectors, labels, affinity, tolerations, secrets, environment variables, image pull secrets, the scheduler name and additional arguments to apply to the session pods. These can be either specified using the Kubernetes model objects or as dictionaries.
 - **`service_account_name`**: The name of a Kubernetes service account to use for session pods. If not configured, pods use the namespace default service account.
 - **`automount_service_account_token`** (default: False): If `True`, a Kubernetes API token is mounted into session pods, allowing code running in the sandbox to call the Kubernetes API with the service account's RBAC permissions.
@@ -76,7 +76,7 @@ Without these controls, sandbox pods may still reach internal services or extern
 | `destroy()` | ✅ | Deletes the backing pod. |
 | `attach(session_id)` | ✅ | Re-attaches to a running session pod. |
 | `snapshot()` / `restore()` | ❌ | Not implemented. |
-| `upload_file` / `download_file` | ❌ | Not implemented. |
+| `upload_file` / `download_file` | ✅ | Transfers file content through Kubernetes exec. |
 
 ## Attaching to a running session
 
--- a/docs/book/component-guide/sandboxes/kubernetes.md
+++ b/docs/book/component-guide/sandboxes/kubernetes.md
@@ -76,7 +76,7 @@ Without these controls, sandbox pods may still reach internal services or extern
 | `destroy()` | ✅ | Deletes the backing pod. |
 | `attach(session_id)` | ✅ | Re-attaches to a running session pod. |
 | `snapshot()` / `restore()` | ❌ | Not implemented. |
-| `upload_file` / `download_file` | ✅ | Transfers file content through Kubernetes exec. |
+| `upload_file` / `download_file` | ✅ | Transfers file content through Kubernetes exec. Uploads are limited to 256 MiB per file. |
 
 ## Attaching to a running session
 
--- a/docs/book/component-guide/sandboxes/kubernetes.md
+++ b/docs/book/component-guide/sandboxes/kubernetes.md
@@ -1,0 +1,94 @@
+---
+description: Pod-backed sandbox sessions for isolated command execution on Kubernetes.
+---
+
+# Kubernetes Sandbox
+
+The Kubernetes sandbox flavor creates one Kubernetes pod per sandbox session and executes each `session.exec(...)` command inside that pod using Kubernetes `exec`.
+
+## How to register
+
+```bash
+zenml integration install kubernetes
+zenml sandbox register k8s-sb --flavor=kubernetes --connector=<CONNECTOR_NAME>
+zenml stack update --sandbox k8s-sb
+```
+
+The [Service Connector](https://docs.zenml.io/how-to/infrastructure-deployment/auth-management/service-connectors-guide) must provide access to a Kubernetes cluster. Alternatively, you can register the sandbox without a connector:
+
+```bash
+zenml sandbox register k8s-sb --flavor=kubernetes --kubernetes_context=<CONTEXT_NAME>
+```
+
+In this case, the sandbox uses your local kubeconfig credentials. `kubernetes_context` selects the kubeconfig context to use and defaults to the active one.
+
+## Additional configuration
+
+Some configuration options for the Kubernetes sandbox can only be set through the sandbox config when you register it (and cannot be changed per-session through the settings):
+
+- **`incluster`** (default: False): If `True`, the sandbox will load the in-cluster Kubernetes configuration and create session pods in the same cluster it is running in, ignoring the `kubernetes_context`.
+- **`kubernetes_context`**: The name of the Kubernetes context to use for creating session pods (ignored if using a service connector or `incluster`).
+- **`kubernetes_namespace`** (default: "zenml"): The Kubernetes namespace in which session pods are created. The namespace must already exist in the Kubernetes cluster.
+
+The following configuration options can be set either through the sandbox config or overridden using `KubernetesSandboxSettings`:
+
+- **`sandbox_environment`**: Environment variables to set in session pods and command executions.
+- **`image`** (default: "python:3.11-slim"): The container image used for session pods. The image must provide `/bin/sh`, so distroless images are not supported.
+- **`pod_settings`**: Node selectors, labels, affinity, tolerations, secrets, environment variables, image pull secrets, the scheduler name and additional arguments to apply to the session pods. These can be either specified using the Kubernetes model objects or as dictionaries.
+- **`service_account_name`**: The name of a Kubernetes service account to use for session pods. If not configured, pods use the namespace default service account.
+- **`automount_service_account_token`** (default: False): If `True`, a Kubernetes API token is mounted into session pods, allowing code running in the sandbox to call the Kubernetes API with the service account's RBAC permissions.
+- **`privileged`** (default: False): If the container should be run in privileged mode.
+- **`pod_startup_timeout`** (default: 120): The maximum time (in seconds) to wait for a session pod to become running.
+- **`api_request_timeout`**: Timeout (in seconds) for Kubernetes API requests. If not set, client defaults are used.
+
+## Security model
+
+This flavor is designed to make safer defaults possible, but cluster policy remains the primary security boundary.
+
+### ZenML-side safeguards (default behavior)
+
+- Service account token is not mounted in sandbox pods (`automount_service_account_token=False`).
+- `sandbox_environment` values are visible to executed code. Do not inject secrets you do not want sandbox code to read.
+
+For untrusted code, set `service_account_name` to a dedicated least-privilege service account instead of relying on the namespace default.
+
+### Required cluster-admin controls
+
+For untrusted LLM code, configure at least:
+
+- **Least-privilege RBAC** for sandbox service accounts.
+- **Pod Security Admission** (or equivalent policy engine) to forbid privileged containers and host namespace escapes.
+- **Runtime hardening policy** (e.g. Kyverno/Gatekeeper) to enforce non-root, read-only root filesystem, seccomp, and no privilege escalation where required.
+- **Network policies** with explicit egress/ingress rules (default-deny plus allow-list).
+- **ResourceQuota / LimitRange** to constrain runaway compute and memory usage.
+- **Image policy controls** (trusted registry, signature/provenance checks if available).
+
+Without these controls, sandbox pods may still reach internal services or external networks based on cluster defaults.
+
+## Feature set
+
+| Feature | Kubernetes | Notes |
+|---|---|---|
+| `create_session()` | ✅ | Creates a pod and waits until it is running. |
+| `exec()` | ✅ | Uses Kubernetes exec websocket streaming. |
+| Streaming output | ✅ | Stdout/stderr stream line-by-line through `SandboxProcess`. |
+| Sandbox log forwarding | ✅ | Output is forwarded into `sandbox:<session_id>` step logs. |
+| `destroy()` | ✅ | Deletes the backing pod. |
+| `attach(session_id)` | ✅ | Re-attaches to a running session pod. |
+| `snapshot()` / `restore()` | ❌ | Not implemented. |
+| `upload_file` / `download_file` | ❌ | Not implemented. |
+
+## Attaching to a running session
+
+`attach(session_id)` re-creates a session handle for an existing session pod. Attaching fails if the pod does not exist, is not running, or was created by a different sandbox component.
+
+{% hint style="warning" %}
+An attached session shares the pod with the original session handle. Calling `destroy()` on either handle deletes the pod, after which `exec()` calls on the other handle fail.
+{% endhint %}
+
+## Lifecycle behavior
+
+- `close()` closes only the local session handle. It does not delete the pod.
+- `destroy()` deletes the backing pod and then closes the handle.
+- `kill()` only stops streaming output. Kubernetes exec does not propagate signals on disconnect, so the command keeps running inside the pod until the pod is deleted.
+
--- a/docs/book/component-guide/toc.md
+++ b/docs/book/component-guide/toc.md
@@ -50,6 +50,7 @@
   * [Develop a Custom Log Store](log-stores/custom.md)
 * [Sandboxes](sandboxes/README.md)
   * [Local](sandboxes/local.md)
+  * [Kubernetes](sandboxes/kubernetes.md)
 * [Step Operators](step-operators/README.md)
   * [Amazon SageMaker](step-operators/sagemaker.md)
   * [AzureML](step-operators/azureml.md)
--- a/src/zenml/integrations/kubernetes/__init__.py
+++ b/src/zenml/integrations/kubernetes/__init__.py
@@ -26,6 +26,7 @@ from zenml.stack import Flavor
 KUBERNETES_ORCHESTRATOR_FLAVOR = "kubernetes"
 KUBERNETES_STEP_OPERATOR_FLAVOR = "kubernetes"
 KUBERNETES_DEPLOYER_FLAVOR = "kubernetes"
+KUBERNETES_SANDBOX_FLAVOR = "kubernetes"
 
 
 class KubernetesIntegration(Integration):
@@ -46,12 +47,14 @@ class KubernetesIntegration(Integration):
         from zenml.integrations.kubernetes.flavors import (
             KubernetesDeployerFlavor,
             KubernetesOrchestratorFlavor,
+            KubernetesSandboxFlavor,
             KubernetesStepOperatorFlavor,
         )
 
         return [
             KubernetesDeployerFlavor,
             KubernetesOrchestratorFlavor,
+            KubernetesSandboxFlavor,
             KubernetesStepOperatorFlavor,
         ]
 
--- a/src/zenml/integrations/kubernetes/flavors/__init__.py
+++ b/src/zenml/integrations/kubernetes/flavors/__init__.py
@@ -23,6 +23,11 @@ from zenml.integrations.kubernetes.flavors.kubernetes_orchestrator_flavor import
     KubernetesOrchestratorFlavor,
     KubernetesOrchestratorSettings,
 )
+from zenml.integrations.kubernetes.flavors.kubernetes_sandbox_flavor import (
+    KubernetesSandboxConfig,
+    KubernetesSandboxFlavor,
+    KubernetesSandboxSettings,
+)
 from zenml.integrations.kubernetes.flavors.kubernetes_step_operator_flavor import (
     KubernetesStepOperatorConfig,
     KubernetesStepOperatorFlavor,
@@ -36,6 +41,9 @@ __all__ = [
     "KubernetesOrchestratorFlavor",
     "KubernetesOrchestratorConfig",
     "KubernetesOrchestratorSettings",
+    "KubernetesSandboxConfig",
+    "KubernetesSandboxFlavor",
+    "KubernetesSandboxSettings",
     "KubernetesStepOperatorConfig",
     "KubernetesStepOperatorFlavor",
     "KubernetesStepOperatorSettings",
--- a/src/zenml/integrations/kubernetes/flavors/kubernetes_sandbox_flavor.py
+++ b/src/zenml/integrations/kubernetes/flavors/kubernetes_sandbox_flavor.py
@@ -1,0 +1,174 @@
+#  Copyright (c) ZenML GmbH 2026. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at:
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""Kubernetes sandbox flavor."""
+
+from typing import TYPE_CHECKING, Optional, Type
+
+from pydantic import Field, PositiveInt
+
+from zenml.constants import KUBERNETES_CLUSTER_RESOURCE_TYPE
+from zenml.integrations.kubernetes import KUBERNETES_SANDBOX_FLAVOR
+from zenml.integrations.kubernetes.pod_settings import KubernetesPodSettings
+from zenml.models import ServiceConnectorRequirements
+from zenml.sandboxes.base import (
+    BaseSandboxConfig,
+    BaseSandboxFlavor,
+    BaseSandboxSettings,
+)
+
+if TYPE_CHECKING:
+    from zenml.integrations.kubernetes.sandboxes import KubernetesSandbox
+
+
+class KubernetesSandboxSettings(BaseSandboxSettings):
+    """Settings for the Kubernetes sandbox."""
+
+    image: str = Field(
+        default="python:3.11-slim",
+        description="Container image used for sandbox session pods.",
+    )
+    pod_settings: Optional[KubernetesPodSettings] = Field(
+        default=None,
+        description="Pod configuration overrides for sandbox session pods.",
+    )
+    service_account_name: Optional[str] = Field(
+        default=None,
+        description="Service account assigned to sandbox session pods.",
+    )
+    automount_service_account_token: bool = Field(
+        default=False,
+        description="Whether to mount a Kubernetes API token into sandbox "
+        "pods, which grants in-pod cluster API access.",
+    )
+    privileged: bool = Field(
+        default=False,
+        description="Whether sandbox session containers run in privileged mode.",
+    )
+    pod_startup_timeout: PositiveInt = Field(
+        default=120,
+        description="Maximum number of seconds to wait for a sandbox pod to "
+        "reach the running phase.",
+    )
+    api_request_timeout: Optional[PositiveInt] = Field(
+        default=None,
+        description="Timeout for Kubernetes API requests in seconds.",
+    )
+
+
+class KubernetesSandboxConfig(BaseSandboxConfig, KubernetesSandboxSettings):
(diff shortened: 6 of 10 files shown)


htahir1:
Reviewed against develop at `2d88f95f18`. Nice and lean overall — correct use of the `_exec`/`_close`/`_destroy` lifecycle hooks, secure pod defaults (no token automount, `require_service_account`), startup-failure pod cleanup, and 404-tolerant destroy. The two must-fixes below are bugs we already hit and fixed on the sibling sandbox PRs, so there's precedent to crib from.

## Must-fix

1. **Shell injection via env *keys*** (`kubernetes_sandbox.py:258`, `_build_shell_command`): values are `shlex.quote`d but keys are interpolated raw into `export`. Repro of the exact composition:
   ```
   env={"A; whoami #": "v"}  →  export A; whoami #='v' && x
   ```
   — the key escapes the export and comments out the command. Same bug class as the one fixed in the k8s_agent_sandbox flavor (#4870): validate keys against `^[A-Za-z_][A-Za-z0-9_]*$` and raise on violation.

2. **`cwd` guard escapes for raw string commands** (`kubernetes_sandbox.py:266`): the fragments join with `&&`, but a *string* command is passed through unmodified, so internal `;` breaks the guard:
   ```
   command="a; b", cwd="/work"  →  cd /work && a; b
   ```
   `b` runs in the wrong directory when `cd` fails. #4870 fixed this with a newline-terminated brace group (`cd <cwd> && { <cmd>\n}` — the newline also survives trailing comments/`;`/`&`), see `k8s_agent_sandbox.py:_exec`.

## Suggestions

3. **Unbounded stream queues**: `_drain_streams` feeds `queue.Queue()` with no cap, so a `wait()`-only caller of a noisy command grows memory without bound. The Cloudflare flavor (#4907) hit exactly this in review and now caps at 100k lines/stream, dropping oldest with a one-time warning — same approach would fit here.

4. **`kill()` semantics doc**: closing the websocket detaches the stream but the command keeps running inside the pod (k8s exec has no signal propagation on disconnect). The base contract reads as "kill terminates"; worth an explicit limitation bullet in `kubernetes.md` mirroring the Cloudflare one ("kill() stops streaming; the command runs on until pod deletion"). The `collect()` drain-failure path is safe either way since `wait()` unblocks on ws close.

5. **No unit tests**: the sibling flavors each ship a mocked suite (session lifecycle, exec composition, startup-failure cleanup, destroy 404 tolerance — ~50 tests each as a template). At minimum the two must-fixes above deserve regression tests.

## Questions

- `require_service_account=True` + `service_account_name=None` defaults mean a fresh `zenml sandbox register --flavor=kubernetes` fails on first `create_session()` until users set a service account or flip the flag. Deliberate secure-by-default, or worth defaulting `require_service_account=False` for the out-of-box path?
- The session pod runs `/bin/sh -c "while true; do sleep 30; done"` — worth a one-line image requirement note (`/bin/sh` must exist; distroless images won't work) in the docs?

Heads-up on merge order: this PR and #4870 both add entries to `sandboxes/README.md` and `toc.md`, so whichever lands second gets a trivial list conflict.

htahir1:
Re-reviewed at `b5e230db34`. Everything from round 1 is addressed — env-key validation, the newline-terminated brace group (each with regression tests, including the trailing-comment/`;`/`&` cases), the stream-buffer cap with overflow warning, the `kill()` limitation and `/bin/sh`+`base64` image requirements documented, and the `require_service_account` UX question resolved by dropping the field. The new `attach()` (with component-ownership + running-phase checks) and the `head -c N` trick for stdin EOF are both nice.

One real finding in the new file-transfer code, suggestion-level:

**Whole-file memory residency in `_upload_file`/`_download_file`** (`kubernetes_sandbox.py:458/488`): upload does `f.read()` + full base64 (~2.3× file size in RAM as strings); download accumulates the encoded stream via `stdout += chunk` in `_collect_exec_output` (which is also O(n²) string concat across many chunks) and then decodes in one shot. Unlike the Cloudflare bridge there's no protocol size cap here, so a multi-GB transfer OOMs the step process. If large transfers are out of scope for v1, a size guard + docs line would do (that's what #4907 went with); otherwise upload streams naturally by reading 3·N-byte chunks → encode → `write_stdin`, and `_collect_exec_output` could at least build a list and `"".join` it.

Two nits, take or leave:
- A websocket drop mid-upload leaves the remote `head` blocked on stdin until pod deletion — same family as the documented `kill()` limitation, probably worth half a sentence in the docs file-transfer section.
- `_upload_file` raises `SandboxExecError` via `_collect_exec_output` but the docstring has no `Raises:` section (pydoclint will flag DOC501 if it runs on this file).

LGTM otherwise — happy to approve once the memory point has an answer (guard, stream, or "out of scope, documented").

htahir1:
Live-verified against this branch on the `kubernetes-sandbox` staging workspace (`test` stack: local orchestrator + this sandbox + S3, all via the aws connector): the Harbor POC example from #4869 ran unmodified — session `k8s-8f43d3eb1e82` pod created, 7 execs through the websocket, verifier reward **1.000**, pod destroyed on trial stop, 14s end-to-end including pod startup. Same example, zero changes, previously ran on Modal — the cross-flavor portability story holds. Run: https://staging.cloud.zenml.io/workspaces/kubernetes-sandbox/projects/f481ca0f-20b5-4d37-9dbe-0f46b8a21b4c/runs/880de620-2de9-43c1-9d00-6b7d4e8d015b

htahir1:
Live-testing follow-up — found a real **concurrency must-fix** that mocked tests can't see:

**Concurrent sessions on one component break the shared `ApiClient`.** Repro on the staging stack: a dynamic pipeline fanning out 3 mapped steps, each opening its own sandbox session from the same `KubernetesSandbox` component, in the same process (ZenML runs mapped steps as concurrent inline threads). One trial succeeds (reward 1.0); the other two fail with kubernetes-client `(0)`-status errors on `connect_get_namespaced_pod_exec`, `read_namespaced_pod`, *and* `delete_namespaced_pod`, all shaped like:

```
(0)
Reason: Handshake status 200 OK -+-+- {'audit-id': ..., 'content-type': 'application/json', ...}
```

(a websocket upgrade that came back `200` instead of `101` — the classic symptom of interleaved requests on a shared connection). A single-session run of the identical pipeline is green, so it's purely the concurrency.

Root cause is almost certainly `KubernetesSandbox._k8s_client` being cached once on the component and shared across threads — the kubernetes python `ApiClient` is not thread-safe. The k8s-agent-sandbox flavor (#4870) hit the adjacent version of this and serializes its global-config mutation behind a lock; here the cleaner fix is probably a per-session (or thread-local) `ApiClient` from the connector instead of the cached singleton, since connectors hand out fresh clients cheaply.

Suggested regression test: two threads, each `create_session()` + `exec()` against a mocked stream factory asserting no shared-client crosstalk — plus this is exactly the parallel-eval use case (the Harbor/JetBrains fan-out), so it's worth fixing before merge.

For balance: everything else held up beautifully under live fire — 15/15 lifecycle checks (exec/env/cwd-guard/injection-rejection/binary file round-trip via the base64 path/timeout/kill/attach/destroy/terminal-close) and the single-trial Harbor run, all on a real cluster.

htahir1:
Addendum to the concurrency finding: a second 3-way-concurrent run (PydanticAI agent fan-out — same component, 3 simultaneous sessions, but exec-heavy rather than file-transfer-heavy) **passed** end-to-end in 36s. So the shared-client failure is a race, not deterministic — the Harbor workload that reproduces it does concurrent `upload_file`/`download_file` (stdin websockets) alongside exec streams, which seems to be what wins the corruption window. Repro recipe if useful: 3 mapped steps each doing create_session → upload → exec → download → destroy. Run: https://staging.cloud.zenml.io/workspaces/kubernetes-sandbox/projects/f481ca0f-20b5-4d37-9dbe-0f46b8a21b4c/runs/edb47c63-e4d0-44ed-b91d-c567d002889a

htahir1:
Root-caused and fixed the concurrency race — validated live. @schustmi here's everything you need:

## Root cause (exact)

It's not general `ApiClient` thread-unsafety — it's `kubernetes.stream`. From `kubernetes/stream/stream.py`:

```python
prev_request = api_client.request
try:
    api_client.request = functools.partial(websocket_request, configuration, binary=binary)
    out = api_method(*args, **kwargs)
finally:
    api_client.request = prev_request
```

Every exec call **swaps `request` on the client object in place** for its duration. With the component-cached client shared across threads: a plain REST call landing inside another thread's swap window goes through `websocket_call` (the `(0)` errors on `read/delete_namespaced_pod`), and an exec whose swap was clobbered goes out as plain HTTP — the server answers `200` instead of `101`, hence `Handshake status 200 OK`. Both observed symptoms, one mechanism. No amount of locking around *your* calls fixes this cleanly; the client must not be shared.

## The fix

Per-session clients, no component-level cache:
- `get_kube_client()` builds a **fresh** `ApiClient` every call (connector / kubeconfig / incluster, logic otherwise unchanged).
- `create_session()` builds one client for the session and uses it for pod create, readiness wait, startup-failure cleanup, and hands a private `CoreV1Api` to the session.
- `KubernetesSandboxSession.__init__` takes `core_api` and owns it (exec streams, file transfer, `_destroy`); the parent-reaching `_core_api` property is gone.
- `attach()` likewise builds the session a private client.

Sessions are single-threaded by construction (one step thread each), so a session-private client is sufficient — and it costs one `connector.connect()` per session instead of per call.

Full patch (applies clean on `b5e230db34`, includes the test fixture adaptation — three attach tests stubbed the old shared property):

<details><summary>4926-race-fix.patch</summary>

```diff
diff --git a/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py b/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py
index f839eb7ebd..abe2a04823 100644
--- a/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py
+++ b/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py
@@ -293,6 +293,7 @@ class KubernetesSandboxSession(SandboxSession):
         pod_name: str,
         namespace: str,
         parent: "BaseSandbox",
+        core_api: k8s_client.CoreV1Api,
     ) -> None:
         """Initialize a Kubernetes sandbox session.
 
@@ -301,21 +302,16 @@ class KubernetesSandboxSession(SandboxSession):
             pod_name: Name of the backing Kubernetes pod.
             namespace: Kubernetes namespace of the pod.
             parent: The sandbox component that created this session.
+            core_api: Session-private API client. ``kubernetes.stream``
+                temporarily monkeypatches ``ApiClient.request`` for the
+                duration of every exec call, so an ``ApiClient`` must
+                never be shared across concurrent sessions/threads.
         """
         self._pod_name = pod_name
         self._namespace = namespace
+        self._core_api = core_api
         super().__init__(id=id, parent=parent)
 
-    @property
-    def _core_api(self) -> k8s_client.CoreV1Api:
-        """Return Kubernetes Core API client.
-
-        Returns:
-            The CoreV1Api client.
-        """
-        sandbox = cast(KubernetesSandbox, self._parent)
-        return sandbox.core_api
-
     @staticmethod
     def _build_shell_command(
         command: Union[str, List[str]],
@@ -521,8 +517,6 @@ class KubernetesSandboxSession(SandboxSession):
 class KubernetesSandbox(BaseSandbox):
     """Kubernetes pod-backed sandbox."""
 
-    _k8s_client: Optional[k8s_client.ApiClient] = None
-
     @property
     def config(self) -> KubernetesSandboxConfig:
         """Kubernetes sandbox configuration.
@@ -542,7 +536,13 @@ class KubernetesSandbox(BaseSandbox):
         return KubernetesSandboxSettings
 
     def get_kube_client(self) -> k8s_client.ApiClient:
-        """Get the Kubernetes API client.
+        """Build a fresh Kubernetes API client.
+
+        A new client is built per call instead of cached on the
+        component: ``kubernetes.stream`` swaps ``ApiClient.request``
+        for the duration of every exec websocket call, so a client
+        shared across concurrent sessions corrupts both plain REST
+        calls and websocket handshakes.
 
         Returns:
             The Kubernetes API client.
@@ -552,11 +552,7 @@ class KubernetesSandbox(BaseSandbox):
         """
         if self.config.incluster:
             kube_utils.load_kube_config(incluster=True)
-            self._k8s_client = k8s_client.ApiClient()
-            return self._k8s_client
-
-        if self._k8s_client and not self.connector_has_expired():
-            return self._k8s_client
+            return k8s_client.ApiClient()
 
         connector = self.get_connector()
         if connector:
@@ -566,14 +562,12 @@ class KubernetesSandbox(BaseSandbox):
                     f"Expected a k8s_client.ApiClient while trying to use "
                     f"the linked connector, but got {type(client)}."
                 )
-            self._k8s_client = client
-        else:
-            kube_utils.load_kube_config(
-                context=self.config.kubernetes_context,
-            )
-            self._k8s_client = k8s_client.ApiClient()
+            return client
 
-        return self._k8s_client
+        kube_utils.load_kube_config(
+            context=self.config.kubernetes_context,
+        )
+        return k8s_client.ApiClient()
 
     @property
     def core_api(self) -> k8s_client.CoreV1Api:
@@ -626,8 +620,10 @@ class KubernetesSandbox(BaseSandbox):
                 resolved_settings.automount_service_account_token
             )
 
+        session_api_client = self.get_kube_client()
+        session_core_api = k8s_client.CoreV1Api(session_api_client)
         kube_utils.create_pod(
-            core_api=self.core_api,
+            core_api=session_core_api,
             namespace=self.config.kubernetes_namespace,
             pod_manifest=pod_manifest,
             api_request_timeout=resolved_settings.api_request_timeout,
@@ -635,7 +631,7 @@ class KubernetesSandbox(BaseSandbox):
 
         try:
             kube_utils.wait_pod(
-                kube_client_fn=self.get_kube_client,
+                kube_client_fn=lambda: session_api_client,
                 pod_name=pod_name,
                 namespace=self.config.kubernetes_namespace,
                 exit_condition_lambda=lambda pod: (
@@ -648,7 +644,7 @@ class KubernetesSandbox(BaseSandbox):
         except Exception:
             try:
                 kube_utils.delete_pod(
-                    core_api=self.core_api,
+                    core_api=session_core_api,
                     pod_name=pod_name,
                     namespace=self.config.kubernetes_namespace,
                     api_request_timeout=resolved_settings.api_request_timeout,
@@ -667,6 +663,7 @@ class KubernetesSandbox(BaseSandbox):
             pod_name=pod_name,
             namespace=self.config.kubernetes_namespace,
             parent=self,
+            core_api=session_core_api,
         )
 
     def attach(self, session_id: str) -> SandboxSession:
@@ -682,8 +679,9 @@ class KubernetesSandbox(BaseSandbox):
         Returns:
             A Kubernetes sandbox session.
         """
+        session_core_api = k8s_client.CoreV1Api(self.get_kube_client())
         pods = kube_utils.retry_on_api_exception(
-            self.core_api.list_namespaced_pod,
+            session_core_api.list_namespaced_pod,
             api_request_timeout=self.config.api_request_timeout,
         )(
             namespace=self.config.kubernetes_namespace,
@@ -720,4 +718,5 @@ class KubernetesSandbox(BaseSandbox):
             pod_name=pod.metadata.name,
             namespace=self.config.kubernetes_namespace,
             parent=self,
+            core_api=session_core_api,
         )
diff --git a/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py b/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py
index 3e30b9652b..bd2be2c087 100644
--- a/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py
+++ b/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py
@@ -15,7 +15,7 @@
 
 import queue
 from typing import List, Optional
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 from uuid import uuid4
 
 import pytest
@@ -170,6 +170,18 @@ def _make_sandbox(pods: List[k8s_client.V1Pod]) -> MagicMock:
     return sandbox
 
 
+def _attach(sandbox: MagicMock, session_id: str):
+    """Call attach with the session-private CoreV1Api stubbed."""
+    from zenml.integrations.kubernetes.sandboxes import kubernetes_sandbox
+
+    with patch.object(
+        kubernetes_sandbox.k8s_client,
+        "CoreV1Api",
+        return_value=sandbox.core_api,
+    ):
+        return KubernetesSandbox.attach(sandbox, session_id)
+
+
 def test_attach_returns_session_for_running_pod() -> None:
     """Test that attach returns a session handle for a running pod."""
     sandbox = _make_sandbox([])
@@ -177,7 +189,7 @@ def test_attach_returns_session_for_running_pod() -> None:
         items=[_session_pod("k8s-abc", str(sandbox.id))]
     )
 
-    session = KubernetesSandbox.attach(sandbox, "k8s-abc")
+    session = _attach(sandbox, "k8s-abc")
 
     assert isinstance(session, KubernetesSandboxSession)
     assert session.id == "k8s-abc"
@@ -190,7 +202,7 @@ def test_attach_fails_without_pod() -> None:
     sandbox = _make_sandbox([])
 
     with pytest.raises(RuntimeError, match="No sandbox pod found"):
-        KubernetesSandbox.attach(sandbox, "k8s-abc")
+        _attach(sandbox, "k8s-abc")
 
 
 def test_attach_fails_for_other_component() -> None:
@@ -198,7 +210,7 @@ def test_attach_fails_for_other_component() -> None:
     sandbox = _make_sandbox([_session_pod("k8s-abc", str(uuid4()))])
 
     with pytest.raises(RuntimeError, match="different sandbox component"):
-        KubernetesSandbox.attach(sandbox, "k8s-abc")
+        _attach(sandbox, "k8s-abc")
 
 
 def test_attach_fails_for_non_running_pod() -> None:
@@ -209,4 +221,4 @@ def test_attach_fails_for_non_running_pod() -> None:
     )
 
     with pytest.raises(RuntimeError, match="is not running"):
-        KubernetesSandbox.attach(sandbox, "k8s-abc")
+        _attach(sandbox, "k8s-abc")

```
</details>

## Validation

- Repro workload (3 concurrent mapped steps, each create_session → upload → exec → download → destroy, same component): **pre-fix ~1-in-3 trials failed per run; post-fix 9/9 trials reward 1.000 across 3 consecutive runs, zero handshake errors** ([run1](https://staging.cloud.zenml.io/workspaces/kubernetes-sandbox/projects/f481ca0f-20b5-4d37-9dbe-0f46b8a21b4c/runs/), runs from ~19:47–19:50 UTC in the workspace).
- Your 14 unit tests pass against the patch.

Feel free to apply it as-is or reshape — the one design point worth preserving is *no shared `ApiClient` across anything that can exec concurrently*.

htahir1:
@schustmi read whoe comment history for problems and fixes i found

schustmi:
> Root-caused and fixed the concurrency race — validated live. @schustmi here's everything you need:
> 
> ## Root cause (exact)
> It's not general `ApiClient` thread-unsafety — it's `kubernetes.stream`. From `kubernetes/stream/stream.py`:
> 
> ```python
> prev_request = api_client.request
> try:
>     api_client.request = functools.partial(websocket_request, configuration, binary=binary)
>     out = api_method(*args, **kwargs)
> finally:
>     api_client.request = prev_request
> ```
> 
> Every exec call **swaps `request` on the client object in place** for its duration. With the component-cached client shared across threads: a plain REST call landing inside another thread's swap window goes through `websocket_call` (the `(0)` errors on `read/delete_namespaced_pod`), and an exec whose swap was clobbered goes out as plain HTTP — the server answers `200` instead of `101`, hence `Handshake status 200 OK`. Both observed symptoms, one mechanism. No amount of locking around _your_ calls fixes this cleanly; the client must not be shared.
> 
> ## The fix
> Per-session clients, no component-level cache:
> 
> * `get_kube_client()` builds a **fresh** `ApiClient` every call (connector / kubeconfig / incluster, logic otherwise unchanged).
> * `create_session()` builds one client for the session and uses it for pod create, readiness wait, startup-failure cleanup, and hands a private `CoreV1Api` to the session.
> * `KubernetesSandboxSession.__init__` takes `core_api` and owns it (exec streams, file transfer, `_destroy`); the parent-reaching `_core_api` property is gone.
> * `attach()` likewise builds the session a private client.
> 
> Sessions are single-threaded by construction (one step thread each), so a session-private client is sufficient — and it costs one `connector.connect()` per session instead of per call.
> 
> Full patch (applies clean on `b5e230db34`, includes the test fixture adaptation — three attach tests stubbed the old shared property):
> 
> 4926-race-fix.patch
> ```diff
> diff --git a/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py b/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py
> index f839eb7ebd..abe2a04823 100644
> --- a/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py
> +++ b/src/zenml/integrations/kubernetes/sandboxes/kubernetes_sandbox.py
> @@ -293,6 +293,7 @@ class KubernetesSandboxSession(SandboxSession):
>          pod_name: str,
>          namespace: str,
>          parent: "BaseSandbox",
> +        core_api: k8s_client.CoreV1Api,
>      ) -> None:
>          """Initialize a Kubernetes sandbox session.
>  
> @@ -301,21 +302,16 @@ class KubernetesSandboxSession(SandboxSession):
>              pod_name: Name of the backing Kubernetes pod.
>              namespace: Kubernetes namespace of the pod.
>              parent: The sandbox component that created this session.
> +            core_api: Session-private API client. ``kubernetes.stream``
> +                temporarily monkeypatches ``ApiClient.request`` for the
> +                duration of every exec call, so an ``ApiClient`` must
> +                never be shared across concurrent sessions/threads.
>          """
>          self._pod_name = pod_name
>          self._namespace = namespace
> +        self._core_api = core_api
>          super().__init__(id=id, parent=parent)
>  
> -    @property
> -    def _core_api(self) -> k8s_client.CoreV1Api:
> -        """Return Kubernetes Core API client.
> -
> -        Returns:
> -            The CoreV1Api client.
> -        """
> -        sandbox = cast(KubernetesSandbox, self._parent)
> -        return sandbox.core_api
> -
>      @staticmethod
>      def _build_shell_command(
>          command: Union[str, List[str]],
> @@ -521,8 +517,6 @@ class KubernetesSandboxSession(SandboxSession):
>  class KubernetesSandbox(BaseSandbox):
>      """Kubernetes pod-backed sandbox."""
>  
> -    _k8s_client: Optional[k8s_client.ApiClient] = None
> -
>      @property
>      def config(self) -> KubernetesSandboxConfig:
>          """Kubernetes sandbox configuration.
> @@ -542,7 +536,13 @@ class KubernetesSandbox(BaseSandbox):
>          return KubernetesSandboxSettings
>  
>      def get_kube_client(self) -> k8s_client.ApiClient:
> -        """Get the Kubernetes API client.
> +        """Build a fresh Kubernetes API client.
> +
> +        A new client is built per call instead of cached on the
> +        component: ``kubernetes.stream`` swaps ``ApiClient.request``
> +        for the duration of every exec websocket call, so a client
> +        shared across concurrent sessions corrupts both plain REST
> +        calls and websocket handshakes.
>  
>          Returns:
>              The Kubernetes API client.
> @@ -552,11 +552,7 @@ class KubernetesSandbox(BaseSandbox):
>          """
>          if self.config.incluster:
>              kube_utils.load_kube_config(incluster=True)
> -            self._k8s_client = k8s_client.ApiClient()
> -            return self._k8s_client
> -
> -        if self._k8s_client and not self.connector_has_expired():
> -            return self._k8s_client
> +            return k8s_client.ApiClient()
>  
>          connector = self.get_connector()
>          if connector:
> @@ -566,14 +562,12 @@ class KubernetesSandbox(BaseSandbox):
>                      f"Expected a k8s_client.ApiClient while trying to use "
>                      f"the linked connector, but got {type(client)}."
>                  )
> -            self._k8s_client = client
> -        else:
> -            kube_utils.load_kube_config(
> -                context=self.config.kubernetes_context,
> -            )
> -            self._k8s_client = k8s_client.ApiClient()
> +            return client
>  
> -        return self._k8s_client
> +        kube_utils.load_kube_config(
> +            context=self.config.kubernetes_context,
> +        )
> +        return k8s_client.ApiClient()
>  
>      @property
>      def core_api(self) -> k8s_client.CoreV1Api:
> @@ -626,8 +620,10 @@ class KubernetesSandbox(BaseSandbox):
>                  resolved_settings.automount_service_account_token
>              )
>  
> +        session_api_client = self.get_kube_client()
> +        session_core_api = k8s_client.CoreV1Api(session_api_client)
>          kube_utils.create_pod(
> -            core_api=self.core_api,
> +            core_api=session_core_api,
>              namespace=self.config.kubernetes_namespace,
>              pod_manifest=pod_manifest,
>              api_request_timeout=resolved_settings.api_request_timeout,
> @@ -635,7 +631,7 @@ class KubernetesSandbox(BaseSandbox):
>  
>          try:
>              kube_utils.wait_pod(
> -                kube_client_fn=self.get_kube_client,
> +                kube_client_fn=lambda: session_api_client,
>                  pod_name=pod_name,
>                  namespace=self.config.kubernetes_namespace,
>                  exit_condition_lambda=lambda pod: (
> @@ -648,7 +644,7 @@ class KubernetesSandbox(BaseSandbox):
>          except Exception:
>              try:
>                  kube_utils.delete_pod(
> -                    core_api=self.core_api,
> +                    core_api=session_core_api,
>                      pod_name=pod_name,
>                      namespace=self.config.kubernetes_namespace,
>                      api_request_timeout=resolved_settings.api_request_timeout,
> @@ -667,6 +663,7 @@ class KubernetesSandbox(BaseSandbox):
>              pod_name=pod_name,
>              namespace=self.config.kubernetes_namespace,
>              parent=self,
> +            core_api=session_core_api,
>          )
>  
>      def attach(self, session_id: str) -> SandboxSession:
> @@ -682,8 +679,9 @@ class KubernetesSandbox(BaseSandbox):
>          Returns:
>              A Kubernetes sandbox session.
>          """
> +        session_core_api = k8s_client.CoreV1Api(self.get_kube_client())
>          pods = kube_utils.retry_on_api_exception(
> -            self.core_api.list_namespaced_pod,
> +            session_core_api.list_namespaced_pod,
>              api_request_timeout=self.config.api_request_timeout,
>          )(
>              namespace=self.config.kubernetes_namespace,
> @@ -720,4 +718,5 @@ class KubernetesSandbox(BaseSandbox):
>              pod_name=pod.metadata.name,
>              namespace=self.config.kubernetes_namespace,
>              parent=self,
> +            core_api=session_core_api,
>          )
> diff --git a/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py b/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py
> index 3e30b9652b..bd2be2c087 100644
> --- a/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py
> +++ b/tests/unit/integrations/kubernetes/test_kubernetes_sandbox.py
> @@ -15,7 +15,7 @@
>  
>  import queue
>  from typing import List, Optional
> -from unittest.mock import MagicMock
> +from unittest.mock import MagicMock, patch
>  from uuid import uuid4
>  
>  import pytest
> @@ -170,6 +170,18 @@ def _make_sandbox(pods: List[k8s_client.V1Pod]) -> MagicMock:
>      return sandbox
>  
>  
> +def _attach(sandbox: MagicMock, session_id: str):
> +    """Call attach with the session-private CoreV1Api stubbed."""
> +    from zenml.integrations.kubernetes.sandboxes import kubernetes_sandbox
> +
> +    with patch.object(
> +        kubernetes_sandbox.k8s_client,
> +        "CoreV1Api",
> +        return_value=sandbox.core_api,
> +    ):
> +        return KubernetesSandbox.attach(sandbox, session_id)
> +
> +
>  def test_attach_returns_session_for_running_pod() -> None:
>      """Test that attach returns a session handle for a running pod."""
>      sandbox = _make_sandbox([])
> @@ -177,7 +189,7 @@ def test_attach_returns_session_for_running_pod() -> None:
>          items=[_session_pod("k8s-abc", str(sandbox.id))]
>      )
>  
> -    session = KubernetesSandbox.attach(sandbox, "k8s-abc")
> +    session = _attach(sandbox, "k8s-abc")
>  
>      assert isinstance(session, KubernetesSandboxSession)
>      assert session.id == "k8s-abc"
> @@ -190,7 +202,7 @@ def test_attach_fails_without_pod() -> None:
>      sandbox = _make_sandbox([])
>  
>      with pytest.raises(RuntimeError, match="No sandbox pod found"):
> -        KubernetesSandbox.attach(sandbox, "k8s-abc")
> +        _attach(sandbox, "k8s-abc")
>  
>  
>  def test_attach_fails_for_other_component() -> None:
> @@ -198,7 +210,7 @@ def test_attach_fails_for_other_component() -> None:
>      sandbox = _make_sandbox([_session_pod("k8s-abc", str(uuid4()))])
>  
>      with pytest.raises(RuntimeError, match="different sandbox component"):
> -        KubernetesSandbox.attach(sandbox, "k8s-abc")
> +        _attach(sandbox, "k8s-abc")
>  
>  
>  def test_attach_fails_for_non_running_pod() -> None:
> @@ -209,4 +221,4 @@ def test_attach_fails_for_non_running_pod() -> None:
>      )
>  
>      with pytest.raises(RuntimeError, match="is not running"):
> -        KubernetesSandbox.attach(sandbox, "k8s-abc")
> +        _attach(sandbox, "k8s-abc")
> ```
> 
> ## Validation
> * Repro workload (3 concurrent mapped steps, each create_session → upload → exec → download → destroy, same component): **pre-fix ~1-in-3 trials failed per run; post-fix 9/9 trials reward 1.000 across 3 consecutive runs, zero handshake errors** ([run1](https://staging.cloud.zenml.io/workspaces/kubernetes-sandbox/projects/f481ca0f-20b5-4d37-9dbe-0f46b8a21b4c/runs/), runs from ~19:47–19:50 UTC in the workspace).
> * Your 14 unit tests pass against the patch.
> 
> Feel free to apply it as-is or reshape — the one design point worth preserving is _no shared `ApiClient` across anything that can exec concurrently_.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Stop tagging me on slop that doesn't make sense, even after 1 second of looking at it.

htahir1:
@schustmi its not slop it literally failed all my harbor example. I'll hand it over to you - run the example end to end please, and see if it works

htahir1 (approved):
Approving — the race fix holds under sustained concurrent fan-out.

Validated live on the staging `kubernetes-sandbox` workspace against `45de1eebfe` (your session-private `CoreV1Api` + per-session `_client_lock`):

- **10 consecutive 3-trial Harbor campaigns = 30/30 trials reward 1.000, zero handshake errors, zero non-zero exits.** Each campaign is the exact workload that reproduced the race pre-fix (3 concurrently-mapped trial steps, each create_session → upload → exec → download → destroy on the same component). Pre-fix this lost ~1 trial in 3; 30/30 clean is strong signal the corruption window is closed.
- Earlier this round: 15/15 lifecycle smoke checks (exec/env/cwd-guard/env-key-rejection/binary file round-trip/timeout/kill/attach/destroy/terminal-close) and the PydanticAI fan-out example, all green on this flavor.

The lifecycle hooks, secure pod defaults, startup-failure cleanup, 404-tolerant destroy, and the newline-terminated brace group / env-key validation from the earlier rounds all look right. Nice work.

PR #4927, opened 2026-06-11
Enforce update permissions for tag-resource endpoints

mosskappa:
## Summary
- require UPDATE permission on the referenced resource before creating or deleting tag-resource relationships
- apply the check to single and batch tag-resource endpoints
- add a regression test covering denied and allowed RBAC paths

## Tests
- python -m pytest tests/unit/zen_server/test_tag_resource_rbac.py -q
- python -m pytest tests/integration/functional/zen_stores/test_zen_store.py::TestTagResource -q
- python -m ruff check src/zenml/zen_server/routers/tag_resource_endpoints.py tests/unit/zen_server/test_tag_resource_rbac.py
- python -m ruff format --check src/zenml/zen_server/routers/tag_resource_endpoints.py tests/unit/zen_server/test_tag_resource_rbac.py
- git diff --check

### DIFF
--- a/src/zenml/zen_server/routers/tag_resource_endpoints.py
+++ b/src/zenml/zen_server/routers/tag_resource_endpoints.py
@@ -26,6 +26,8 @@ from zenml.constants import (
 from zenml.models import TagResourceRequest, TagResourceResponse
 from zenml.zen_server.auth import AuthContext, authorize
 from zenml.zen_server.exceptions import error_response
+from zenml.zen_server.rbac.models import Action
+from zenml.zen_server.rbac.utils import batch_verify_permissions_for_models
 from zenml.zen_server.utils import (
     async_fastapi_endpoint_wrapper,
     zen_store,
@@ -36,6 +38,16 @@ router = APIRouter(
     tags=["tag_resources"],
     responses={401: error_response, 403: error_response},
 )
+
+
+def _verify_tag_resources_update_permission(
+    tag_resources: List[TagResourceRequest],
+) -> None:
+    """Verify update permission on all resources being tagged."""
+    resources = zen_store().get_resources_from_tag_resources(
+        tag_resources=tag_resources
+    )
+    batch_verify_permissions_for_models(models=resources, action=Action.UPDATE)
 
 
 @router.post(
@@ -55,6 +67,7 @@ def create_tag_resource(
     Returns:
         A tag resource response.
     """
+    _verify_tag_resources_update_permission(tag_resources=[tag_resource])
     return zen_store().create_tag_resource(tag_resource=tag_resource)
 
 
@@ -75,10 +88,8 @@ def batch_create_tag_resource(
     Returns:
         A list of tag resource responses.
     """
-    return [
-        zen_store().create_tag_resource(tag_resource=tag_resource)
-        for tag_resource in tag_resources
-    ]
+    _verify_tag_resources_update_permission(tag_resources=tag_resources)
+    return zen_store().batch_create_tag_resource(tag_resources=tag_resources)
 
 
 @router.delete(
@@ -95,6 +106,7 @@ def delete_tag_resource(
     Args:
         tag_resource: The tag resource relationship to delete.
     """
+    _verify_tag_resources_update_permission(tag_resources=[tag_resource])
     zen_store().delete_tag_resource(tag_resource=tag_resource)
 
 
@@ -112,4 +124,5 @@ def batch_delete_tag_resource(
     Args:
         tag_resources: A list of tag resource requests.
     """
+    _verify_tag_resources_update_permission(tag_resources=tag_resources)
     zen_store().batch_delete_tag_resource(tag_resources=tag_resources)
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -15205,6 +15205,74 @@ class SqlZenStore(BaseZenStore):
 
         return resource_type_to_schema_mapping[resource_type]
 
+    def _get_resources_from_tag_resources(
+        self,
+        tag_resources: List[TagResourceRequest],
+        session: Session,
+    ) -> List[BaseSchema]:
+        """Get resources referenced by tag resource requests in bulk.
+
+        Args:
+            tag_resources: The tag resource requests.
+            session: The database session to use.
+
+        Returns:
+            The resources referenced by the tag resource requests.
+
+        Raises:
+            KeyError: If a referenced resource does not exist.
+        """
+        resource_ids_by_type: Dict[TaggableResourceTypes, Set[UUID]] = (
+            defaultdict(set)
+        )
+        for tag_resource in tag_resources:
+            resource_ids_by_type[tag_resource.resource_type].add(
+                tag_resource.resource_id
+            )
+
+        resources_by_key: Dict[
+            Tuple[TaggableResourceTypes, UUID], BaseSchema
+        ] = {}
+        for resource_type, resource_ids in resource_ids_by_type.items():
+            schema_class = self._get_schema_from_resource_type(resource_type)
+            query = select(schema_class).where(
+                col(schema_class.id).in_(resource_ids)
+            )
+            resources = session.exec(query).all()
+            for resource in resources:
+                resources_by_key[(resource_type, resource.id)] = resource
+
+        resolved_resources: List[BaseSchema] = []
+        for tag_resource in tag_resources:
+            key = (tag_resource.resource_type, tag_resource.resource_id)
+            try:
+                resolved_resources.append(resources_by_key[key])
+            except KeyError:
+                raise KeyError(
+                    f"{tag_resource.resource_type.value} with ID "
+                    f"`{tag_resource.resource_id}` not found."
+                )
+
+        return resolved_resources
+
+    def get_resources_from_tag_resources(
+        self,
+        tag_resources: List[TagResourceRequest],
+    ) -> List[Any]:
+        """Get resource models referenced by tag resource requests in bulk.
+
+        Args:
+            tag_resources: The tag resource requests.
+
+        Returns:
+            The resource models referenced by the tag resource requests.
+        """
+        with Session(self.engine) as session:
+            resources = self._get_resources_from_tag_resources(
+                tag_resources=tag_resources, session=session
+            )
+            return [resource.to_model() for resource in resources]
+
     def _get_tag_schema(
         self,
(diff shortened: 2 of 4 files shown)


src/zenml/zen_server/routers/tag_resource_endpoints.py
```diff
@@ -38,6 +41,36 @@
...
+        tag_resource.resource_id, hydrate=False
+    )
+    verify_permission_for_model(model=resource, action=Action.UPDATE)
+
+
+def _verify_tag_resources_update_permission(
+    tag_resources: List[TagResourceRequest],
+) -> None:
+    """Verify update permission on all resources being tagged."""
+    for tag_resource in tag_resources:
```
schustmi: This should do a batch request to avoid unnecessary requests to the RBAC service -> `batch_verify_permissions_for_models`

schustmi (changes requested):
@mosskappa Can you please also rebase on develop and use that as the base of your PR. Thanks!

mosskappa:
Rebased this PR on `develop` and updated the PR base branch accordingly.

I reran the focused checks after the rebase:

- `python -m pytest tests/unit/zen_server/test_tag_resource_rbac.py -q`
- `python -m pytest tests/integration/functional/zen_stores/test_zen_store.py::TestTagResource -q`
- `python -m ruff check src/zenml/zen_server/routers/tag_resource_endpoints.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `python -m ruff format --check src/zenml/zen_server/routers/tag_resource_endpoints.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `git diff --check`

mosskappa:
Updated the batch create/delete permission path to resolve all referenced resources first and call `batch_verify_permissions_for_models(..., action=Action.UPDATE)` instead of verifying each model individually.

Reran:

- `python -m pytest tests/unit/zen_server/test_tag_resource_rbac.py -q`
- `python -m pytest tests/integration/functional/zen_stores/test_zen_store.py::TestTagResource -q`
- `python -m ruff check src/zenml/zen_server/routers/tag_resource_endpoints.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `python -m ruff format --check src/zenml/zen_server/routers/tag_resource_endpoints.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `git diff --check`

stefannica:
Hello @mosskappa,

Thanks for adding the RBAC checks here. The overall approach makes sense: when attaching or detaching a tag from a resource, we should verify that the user has `UPDATE` permissions on the affected resource.

One thing you should adjust is how this works for the batch endpoints. In `batch_create_tag_resource` / `batch_delete_tag_resource`, we should avoid resolving every referenced resource through individual SQL Zen Store `get` calls. For large batches, that turns into one DB lookup per tag-resource entry, which is unnecessarily expensive.

This is a bit more involved than a regular endpoint-level PR. At first glance, this sounds like a good fit for list calls: group the incoming `TagResourceRequest`s by `resource_type`, collect the IDs for each type, and call the corresponding `list_*` method with an `id=oneof:[...]` filter. However, most of these resources are project-scoped, and the normal list APIs require a single project context. The tag-resource request only contains `resource_type`, `resource_id`, and `tag_id`; it does not contain the project of the target resource. That means regular project-scoped list calls are not a reliable way to resolve an arbitrary batch of resource IDs, especially if the batch contains resources from multiple projects.

Instead, this likely needs to be handled inside the SQL Zen Store with some database-level batching. Could you add SQL-store-level functionality that resolves a batch of tag-resource requests in bulk?

Concretely, something along these lines:

1. Add a helper/method in `SQLZenStore` that accepts a list of `TagResourceRequest`s.
2. Group the requests by `resource_type`.
3. Use the existing resource-type-to-schema mapping functionality, e.g. `_get_schema_from_resource_type(...)`, to determine the SQL schema class for each group.
4. For each resource type, issue one DB query using `schema.id.in_(...)` for the collected resource IDs.
5. Return the resolved resource schemas/models mapped back to the original tag-resource requests.
6. The endpoint can then convert those resolved resources into RBAC `Resource`s and run the existing bulk `check_permissions(..., action=Action.UPDATE)` logic.

This keeps the RBAC behavior the same, but avoids N individual resource lookups and also avoids trying to force this through project-scoped list APIs when we do not know the target projects ahead of time.

There is already tag-resource handling in the SQL store around resource-type/schema resolution, so hopefully this can reuse existing pieces like `_get_schema_from_resource_type(...)` rather than introducing a completely separate mapping.

That said, this is more advanced than the original endpoint change because it touches the SQL store and requires careful DB-level handling. If you do not feel comfortable taking on that part, no worries at all. Just let us know, and we can take over this follow-up on the maintainer side.

mosskappa:
Thanks for the detailed direction. I updated the batch path to avoid per-entry resource lookups:

- added SQL-store-level bulk resolution for `TagResourceRequest` lists, grouped by `resource_type`
- each resource type now uses a single `id.in_(...)` query for the collected resource IDs
- the endpoint uses the bulk-resolved resource models for `batch_verify_permissions_for_models(..., action=Action.UPDATE)`
- `batch_create_tag_resource` now reuses the same bulk resolver before creating the tag-resource schemas
- added a store integration test that patches `_get_schema_by_id` to fail if batch tag-resource creation falls back to individual resource get calls
- expanded the endpoint RBAC regression to cover batch create/delete as well

Reran:

- `python -m pytest tests/unit/zen_server/test_tag_resource_rbac.py -q`
- `python -m pytest tests/integration/functional/zen_stores/test_zen_store.py::TestTagResource -q`
- original local bypass reproducer against the patched branch (expected fixed-control result: attach now returns 403)
- `python -m ruff check src/zenml/zen_server/routers/tag_resource_endpoints.py src/zenml/zen_stores/sql_zen_store.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `python -m ruff check tests/integration/functional/zen_stores/test_zen_store.py --ignore D100,D101,D102,D103,D202,D205` (to avoid unrelated pre-existing docstring findings in this file)
- `python -m ruff format --check src/zenml/zen_server/routers/tag_resource_endpoints.py src/zenml/zen_stores/sql_zen_store.py tests/unit/zen_server/test_tag_resource_rbac.py tests/integration/functional/zen_stores/test_zen_store.py`
- `python -m py_compile src/zenml/zen_server/routers/tag_resource_endpoints.py src/zenml/zen_stores/sql_zen_store.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `git diff --check`

src/zenml/zen_server/routers/tag_resource_endpoints.py
```diff
@@ -75,6 +128,7 @@ def batch_create_tag_resource(
     Returns:
         A list of tag resource responses.
     """
+    _verify_tag_resources_update_permission(tag_resources=tag_resources)
     return [
         zen_store().create_tag_resource(tag_resource=tag_resource)
         for tag_resource in tag_resources
```
stefannica: Would you be so kind to fix this so that it calls this instead ?

```
return zen_store().batch_create_tag_resource(tag_resources=tag_resources)
```

src/zenml/zen_server/routers/tag_resource_endpoints.py
```diff
@@ -38,6 +44,52 @@
...
+
+def _verify_tag_resources_update_permission(
+    tag_resources: List[TagResourceRequest],
+) -> None:
+    """Verify update permission on all resources being tagged."""
+    store = zen_store()
+    bulk_resource_getter = getattr(
+        store, "get_resources_from_tag_resources", None
+    )
+    if bulk_resource_getter:
```
stefannica: This is a very awkward way of calling the new method you added to SQL Zen store. Why not call it directly ? There is no reason for this type of getter-based backwards compatibility logic here.

src/zenml/zen_server/routers/tag_resource_endpoints.py
```diff
@@ -38,6 +44,52 @@
...
+    )
+
+
+def _verify_tag_resource_update_permission(
+    tag_resource: TagResourceRequest,
+) -> None:
+    """Verify update permission on the resource being tagged."""
+    resource = _get_tagged_resource(tag_resource=tag_resource)
+    verify_permission_for_model(model=resource, action=Action.UPDATE)
+
```
stefannica: Now that you added code inside the SQL zen store to deal with resolution of tagged resources, this code here is unnecessary and duplicated. You can simply use `_verify_tag_resources_update_permission` instead and just pass a one-item list.

stefannica (changes requested):
Thank you for the update. A few remaining improvement suggestions, but otherwise the approach looks functionally correct.

mosskappa:
Thanks, I addressed the remaining cleanup suggestions:

- `batch_create_tag_resource` now calls `zen_store().batch_create_tag_resource(tag_resources=tag_resources)` directly.
- Removed the getter-based fallback and now call `zen_store().get_resources_from_tag_resources(...)` directly.
- Removed the duplicated single-resource resolver in the router. Single create/delete now pass a one-item list into `_verify_tag_resources_update_permission(...)`, so all endpoint paths use the same SQL-store-backed resource resolution.

Reran:

- `python -m pytest tests/unit/zen_server/test_tag_resource_rbac.py -q`
- `python -m pytest tests/integration/functional/zen_stores/test_zen_store.py::TestTagResource -q`
- original local bypass reproducer against the patched branch (expected fixed-control result: attach returns 403)
- `python -m ruff check src/zenml/zen_server/routers/tag_resource_endpoints.py src/zenml/zen_stores/sql_zen_store.py tests/unit/zen_server/test_tag_resource_rbac.py`
- `python -m ruff format --check src/zenml/zen_server/routers/tag_resource_endpoints.py src/zenml/zen_stores/sql_zen_store.py tests/unit/zen_server/test_tag_resource_rbac.py tests/integration/functional/zen_stores/test_zen_store.py`
- `git diff --check`

stefannica (approved):
This looks great now, thank you for your contribution and prompt follow-ups!

mosskappa:
Thanks again. I noticed the latest CI run exposed a test isolation issue in the new regression test: it was setting ZenML test paths at import time, which could leak into unrelated unit tests in the full suite. I updated the test to use the existing clean client fixture and restore the temporary server/RBAC state after the request checks.

Local verification after the update:

- tag-resource RBAC regression test
- the previously affected unit tests for artifact store registration, stack component secret reference handling, and run configuration loading
- ruff check for the touched server/store/regression-test files
- ruff format --check
- git diff --check

schustmi (approved):

PR #4938, opened 2026-06-12
Improved tag resource RBAC checks and consistent RBAC access to tags

stefannica:
**Summary**

This PR moves tag-resource RBAC enforcement out of the tag-resource endpoints and into `RBACSqlZenStore`, keeping RBAC-specific behavior isolated in the RBAC store layer. It also opens up `get_tag` to match the existing strategy that tags are server-wide resources everyone can create and view without RBAC checks.

**What Changed**

- Removed endpoint-level permission checks from `tag_resource_endpoints.py`.
- Added RBAC checks to `RBACSqlZenStore.batch_create_tag_resource(...)` and `RBACSqlZenStore.batch_delete_tag_resource(...)`.
- Reused the existing SQL store `_get_resources_from_tag_resources(...)` bulk resolver instead of hydrating response models.
- Added a private `SqlZenStore._batch_create_tag_resource(...)` helper that accepts already-resolved resources.
- Opened up `get_tag` so tag reads are consistent with tag creation/listing as globally visible server-wide resources.
- Updated the tag-resource RBAC regression test to exercise `RBACSqlZenStore` directly.

**Why This Is Better**

Previously, the endpoint fetched full resource models just to check `UPDATE` permissions before tagging or untagging. That mixed RBAC concerns into endpoint code and caused unnecessary schema-to-model hydration.

The new flow keeps endpoints clean and pushes authorization to the RBAC store, where it belongs. It resolves tag resources once, builds lightweight RBAC `Resource` objects from the resolved schemas, and reuses those same resolved resources for the actual create/delete operation.

Opening `get_tag` keeps tag behavior internally consistent: tags are shared server-wide metadata objects that users can create and view, while RBAC is enforced when attaching or detaching a tag from a protected resource.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/rbac/rbac_sql_zen_store.py
+++ b/src/zenml/zen_server/rbac/rbac_sql_zen_store.py
@@ -14,16 +14,12 @@
 """RBAC SQL Zen Store implementation."""
 
 from typing import (
-    TYPE_CHECKING,
     List,
     Optional,
-    Sequence,
-    Set,
     Tuple,
 )
 from uuid import UUID
 
-from zenml.enums import TaggableResourceTypes
 from zenml.logger import get_logger
 from zenml.models import (
     ModelRequest,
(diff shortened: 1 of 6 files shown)


src/zenml/zen_server/rbac/rbac_sql_zen_store.py
```diff
@@ -14,35 +14,144 @@
...
+        """
+        with Session(self.engine) as session:
+            resolved_resources = self._get_resources_from_tag_resources(
+                tag_resources=tag_resources, session=session
+            )
+            resources = self._get_rbac_resources_from_tag_resources(
+                tag_resources=tag_resources,
+                resolved_resources=resolved_resources,
+            )
+            batch_verify_permissions(resources=resources, action=Action.UPDATE)
```
schustmi: This is not correct behaviour, as it ignores the rule that ownership of an entity automatically gives permissions to read/update/delete that entity. That's why `batch_verify_permissions_for_model(...)` should be used instead here.

schustmi (changes requested):

src/zenml/zen_server/rbac/rbac_sql_zen_store.py
stefannica: done. I'm looking at owners first and only calling RBAC for entities that have a different owner.

schustmi: I see that you're looking at user_id on the schema instead of using `batch_verify_permissions_for_model` for the ownership checks. This saves us a few milliseconds of converting to models, but the downside is that this is now a special case handling here that we need to catch if we make any changes to the RBAC logic. This is not an operation that is called from the hot path of running a regular pipeline so IMO the minor time savings we get are not that crucial, but maybe I'm wrong. I'll leave it up to you

schustmi (approved):

src/zenml/zen_server/rbac/rbac_sql_zen_store.py
stefannica: I moved as much RBAC specific code as possible from the RBAC zen store and into the RBAC "core" that is used by all other endpoints.

schustmi (approved):

bcdurak (approved):

PR #4942, opened 2026-06-15
Add DB indexes and request timeout fix

safoinme:
## Summary

- Adds an Alembic migration for targeted database indexes that support common pagination and hydration query shapes.
- Adds matching SQLModel index metadata for pipeline runs, pipeline snapshots, step configurations, step runs, and artifact versions.
- Updates request management so only deduplicated/cacheable requests can return a timeout/backpressure response while the underlying operation continues; non-deduplicated requests keep the HTTP connection and operation lifetime aligned.
- Adds focused unit coverage for the new index metadata and request timeout behavior.

## Why

This is split out from the broader MySQL workspace DB backpressure work so reviewers can inspect the pagination/index and request-management changes independently from the expired API transaction cleanup work in #4943.

The pagination indexes intentionally keep `id` where ZenML sorts by `created, id`; `id` is the stable tie-breaker used by the generic filter sorting path.

## Scope note

This PR does **not** add API transaction cleanup behavior or API transaction cleanup indexes. The existing `ix_api_transaction_completed_expired` index is already present on `develop` from the original API transaction table migration.

## Reviewer focus

- `src/zenml/zen_stores/migrations/versions/c465032f0ec0_add_pagination_indexes_for_pipeline_.py`
  - Confirm index names/columns match the schema declarations.
  - Confirm the migration only includes pagination/hydration-related indexes.
  - Confirm downgrade drops only the indexes added by this migration.
- `src/zenml/zen_server/request_management.py`
  - Confirm timeout behavior remains limited to deduplicated/cacheable requests.
- `tests/unit/zen_stores/test_sql_store_utils.py`
  - Confirm index expectations match intended query shapes.

## Validation

Local checks run:

- `git diff --check`: passed
- In-memory Python compile check for touched Python files: passed
- Diff-only search for `pool_timeout`, `poolTimeout`, `max_overflow`, `maxOverflow`, and `ZENML_STORE_MAX_OVERFLOW`: no branch-added matches

Not run:

- Focused pytest command could not run because the existing worktree `.venv` does not have `pytest` installed:
  - `.venv/bin/python -m pytest tests/unit/zen_server/test_request_management.py tests/unit/zen_stores/test_sql_store_utils.py -q --no-provision`
  - failed with `No module named pytest`

## Follow-up

The expired API transaction cleanup branch is stacked in #4943.


src/zenml/zen_stores/migrations/versions/c465032f0ec0_add_pagination_indexes_for_pipeline_.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def upgrade() -> None:
+    """Upgrade database schema and/or data, creating a new revision."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    create_index_if_missing(
+        "api_transaction",
+        "ix_api_transaction_completed_expired",
+        ["completed", "expired"],
+    )
```
stefannica: This doesn't belong here, right ?

stefannica (approved):

Json-Andriopoulos (approved):

PR #4943, opened 2026-06-15
Add catch-up cleanup for expired API transactions

safoinme:
## Summary

- Reworks expired API transaction cleanup from one fixed delete per interval into bounded catch-up cleanup.
- Adds configurable cleanup controls for per-SQL-batch size and per-pass time budget.
- Keeps each SQL delete bounded while allowing a cleanup pass to drain multiple full batches when there is backlog.
- Adds expired transaction reclaim logic so completed expired idempotency transactions can be safely re-executed and stale cached results are removed.

## Dependency

This PR is stacked on #4942 / `fix/mysql-db-indexes-request-management`.

The API transaction cleanup index itself is **not added here** because `ix_api_transaction_completed_expired` already exists on `develop` from the original API transaction table migration. This PR uses that existing index for the bounded cleanup scan.

## Behavior

Before this change, cleanup capacity was fixed at one batch per interval. With the default interval, backlog could grow indefinitely if expired transactions accumulated faster than that single batch capacity.

Now each cleanup pass:

1. Deletes at most `api_transaction_cleanup_batch_size` rows per SQL batch.
2. Repeats while batches are full, which indicates backlog remains.
3. Stops once a partial batch is deleted or `api_transaction_cleanup_time_budget` is reached.

This avoids unbounded single deletes while allowing catch-up throughput when the database handles cleanup quickly.

## Reviewer focus

- `src/zenml/zen_server/request_management.py`
  - Confirm the cleanup loop does not schedule overlapping cleanup tasks in one process.
  - Confirm the time-budget loop cannot run indefinitely.
- `src/zenml/zen_stores/sql_zen_store.py`
  - Confirm the two-step select/delete keeps each SQL delete bounded.
  - Confirm MySQL `SKIP LOCKED` remains limited to the select query.
  - Confirm expired transaction reclaim is atomic and deletes stale cached results.
- `src/zenml/config/server_config.py`
  - Confirm the new config defaults and caps are safe.

## Validation

Local checks run:

- `git diff --check`: passed
- In-memory Python compile check for touched Python files: passed
- Diff-only search for `pool_timeout`, `poolTimeout`, `max_overflow`, `maxOverflow`, and `ZENML_STORE_MAX_OVERFLOW`: no branch-added matches

Not run:

- Focused pytest command could not run because the existing worktree `.venv` does not have `pytest` installed:
  - `.venv/bin/python -m pytest tests/unit/zen_server/test_request_management.py tests/unit/zen_stores/test_sql_store_utils.py -q --no-provision`
  - failed with `No module named pytest`


Json-Andriopoulos (approved):

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2421,17 +2423,94 @@ def _get_api_transaction(
...
+    ) -> int:
+        """Delete a bounded batch of completed API transactions that expired."""
+        if (
+            batch_size < 1
+            or batch_size > MAX_ZENML_SERVER_API_TXN_CLEANUP_BATCH_SIZE
+        ):
+            raise ValueError(
+                "Expired API transaction cleanup batch size must be between "
+                f"1 and {MAX_ZENML_SERVER_API_TXN_CLEANUP_BATCH_SIZE}."
+            )
```
stefannica: sounds like something to add to the server settings validator instead of here where it can fail silently.

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2421,17 +2423,94 @@ def _get_api_transaction(
...
+        if result_schema is not None:
+            session.delete(result_schema)
+
+        session.commit()
+        session.refresh(api_transaction_schema)
+        return api_transaction_schema
+
+    def cleanup_expired_api_transactions(
+        self,
+        batch_size: int = DEFAULT_ZENML_SERVER_API_TXN_CLEANUP_BATCH_SIZE,
```
stefannica: I suggest simplifying this: make this mandatory and get rid of the default value and validation. you're calling this from only one place anyway.

src/zenml/config/server_config.py
```diff
@@ -393,9 +402,17 @@ class ServerConfiguration(BaseModel):
...
         DEFAULT_ZENML_SERVER_API_TXN_CLEANUP_INTERVAL
     )
+    api_transaction_cleanup_batch_size: PositiveInt = Field(
+        default=DEFAULT_ZENML_SERVER_API_TXN_CLEANUP_BATCH_SIZE,
+        le=MAX_ZENML_SERVER_API_TXN_CLEANUP_BATCH_SIZE,
+    )
+    api_transaction_cleanup_time_budget: PositiveFloat = Field(
+        default=DEFAULT_ZENML_SERVER_API_TXN_CLEANUP_TIME_BUDGET,
+        le=MAX_ZENML_SERVER_API_TXN_CLEANUP_TIME_BUDGET,
+    )
```
stefannica, line 418: one other validation to add: `api_transaction_cleanup_time_budget` must not be higher than `api_transaction_cleanup_interval`

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -2450,24 +2529,58 @@ def get_or_create_api_transaction(
...
-                # continue using it
-                session.rollback()
-                api_transaction_schema = self._get_api_transaction(
-                    api_transaction_id=api_transaction_schema.id,
-                    method=api_transaction.method,
-                    url=api_transaction.url,
-                    session=session,
+            for retry in range(2):
+                api_transaction_schema = ApiTransactionSchema.from_request(
+                    api_transaction
```
stefannica, line 2483: could you please avoid the "reclaim" complications and keep this simple ?

to avoid race conditions here, you can simply do this 2-times loop until either create or get succeeds. and it doesn't matter if you get an expired transaction either. the expiration aspect is only relevant for cleanup and should play no role in whether a cached request is usable or not.

stefannica (changes requested):

src/zenml/zen_server/request_management.py
```diff
@@ -120,10 +123,13 @@ def is_cacheable(self) -> bool:
...
-        # Only cache requests that are authenticated and are part of a
-        # transaction.
+        # Read-only requests are already idempotent. Routing them through the
+        # SQL-backed transaction cache turns high-frequency control-plane
+        # polling into writes and cleanup work without adding mutation safety.
         return (
-            self.auth_context is not None and self.transaction_id is not None
+            self.auth_context is not None
+            and self.transaction_id is not None
+            and self.request.method.upper() not in _SAFE_REQUEST_METHODS
```
stefannica: there is no reason to not cache requests for GET calls. In fact, these are the very reason why this API transaction feature was created: to cache big step run and snapshot responses.

stefannica (changes requested):

stefannica (approved):

PR #4946, opened 2026-06-16
Validate artifact content hash when loading pickled artifacts

qxyuan853:
## What

The cloudpickle fallback materializer now records the SHA-256 of the stored
artifact file and verifies it before the file is read back on load. When a
content hash is recorded for the artifact version, the stored bytes are checked
against it; artifacts with no recorded hash load unchanged.

The hash is computed in a single streaming pass while the artifact is written in
`save()` (a small `_HashingWriter` wraps the store write handle), and cached on
the materializer, so loading needs no extra read of the (possibly remote)
artifact. The recorded hash reaches the materializer through a
`BaseMaterializer.expected_content_hash` field, set in `_load_artifact_from_uri`,
the model-metadata loader, and step-input loading.

## Why

Gives the pickle-based fallback an integrity check that ties the stored file to
the content hash recorded for its artifact version, so a stored file that no
longer matches what was recorded is not read back. Computing the hash during
`save()` avoids re-reading the artifact on load.

## Tests

Unit tests for matching-hash loads, content-hash mismatches, the no-hash path,
and save/re-read hash equivalence.

### DIFF
--- a/src/zenml/materializers/cloudpickle_materializer.py
+++ b/src/zenml/materializers/cloudpickle_materializer.py
@@ -13,8 +13,10 @@
 #  permissions and limitations under the License.
 """Implementation of ZenML's cloudpickle materializer."""
 
+import hashlib
+import hmac
 import os
-from typing import Any, ClassVar, Tuple, Type
+from typing import IO, Any, ClassVar, Optional, Tuple, Type
 
 import cloudpickle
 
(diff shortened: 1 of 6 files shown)


src/zenml/materializers/cloudpickle_materializer.py
```diff
@@ -108,13 +183,53 @@ def save(self, data: Any) -> None:
...
+                    digest.update(chunk)
+            return digest.hexdigest()
+        except Exception as e:
+            logger.warning(
+                "Could not compute the content hash for the artifact at "
+                "'%s'. (%s)",
+                self.uri,
+                e,
+            )
+            return None
```
schustmi, line 235: IMO, this fallback should be removed. We can just return `self._content_hash` in this function. Any call to `save` will always compute it, so no need for this duplicate code here.

src/zenml/materializers/cloudpickle_materializer.py
```diff
@@ -32,6 +34,56 @@
 DEFAULT_FILENAME = "artifact.pkl"
 DEFAULT_PYTHON_VERSION_FILENAME = "python_version.txt"
 
+# Chunk size used when streaming the stored artifact to compute its hash.
+_HASH_CHUNK_SIZE = 65536
+
+
+class _HashingWriter:
```
schustmi, line 41: Great idea, I think we can make this more generic and reusable:
- Instead of tying this to hashing, the class could receive a callback that it invokes every time `write(..)` is called. That way, it doesn't need to know about hashes etc.
- This should be a generic class so that it supports correct type checking for both `IO[str]` and `IO[bytes]`.
- Lastly, I would suggest removing all the docstrings that tie this to artifact stores etc, and move the class to our `io_utils` file so others can reuse it.

schustmi (changes requested):
Sorry for the late review @qxyuan853, only got around to it now. Left some comments, but I much prefer this approach :)

src/zenml/materializers/cloudpickle_materializer.py
qxyuan853: Thanks! I agree. Those changes make the class much more generic and reusable.

Also, I wanted to follow up on the security issue. Would it be possible to assign a CVE ID for it? I previously submitted a CVE request following the security policy via email (yyjLF@proton.me). I’d really appreciate it if you could help with the assignment if appropriate.

schustmi: ZenML is meant to be used in a private shared environment, meaning that deploying and using ZenML comes with the obvious responsibilities of ensuring proper security is put in place to protect its adjacent shared data sources. ZenML is not meant to replace these responsibilities.

More importantly, this vulnerability can only be exploited if the attacker can access or inject malicious files into a shared environment.

I still think this PR is valuable and it provides protection against accidental modifications of your artifacts, but if someone has access to your artifact store and ZenML server, they could even work around this. For this reason, we consider this as a functionality improvement and not actually fixing a vulnerability.

schustmi: @qxyuan853 Are you still willing to work on this PR or should we close it?

qxyuan853: Hi @schustmi, thanks for checking in — yes, I'm still willing to work on this PR. Just to clarify one thing on my side: do you need me to submit a new PR, or should I keep updating this one? I had previously assumed you'd make the adjustments directly, so I wanted to confirm before proceeding.

schustmi: Continuing on this PR would be fine. I'll leave the changes up to you if you want to do them, if not I'd do them myself on a separate PR :)

qxyuan853: Thank you! I think it'd be easier if you just handled it directly. Appreciate your help on this :)

PR #4950, opened 2026-06-16
Improve internal future awaiting error handling

schustmi:
Previously, we always re-raised the exception that happened during step execution when it caused a pipeline to fail. This PR adjusts this in the following way:
- When a step is explicitly awaited (run synchronously or `future.wait()`/`future.result()`), we re-raise the step exception
- When a step is implicitly awaited (future passed to downstream step, or the future being awaited after the pipeline function returns), a `StepExecutionException` with a generic message is raised instead.

**Breaking changes:**
- If users had a `try...except` block around their pipeline execution, their code could break because different exceptions get raised.

### DIFF
--- a/src/zenml/exceptions.py
+++ b/src/zenml/exceptions.py
@@ -130,6 +130,24 @@ class RunInterruptedException(ZenMLBaseException):
     """Raised when a ZenML step gets interrupted for an unknown reason."""
 
 
+class StepExecutionException(ZenMLBaseException):
+    """Raised when a step fails and its result was not explicitly awaited."""
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        original_exception: Optional[BaseException] = None,
+    ):
+        """Initializes the exception.
+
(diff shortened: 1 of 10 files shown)


src/zenml/exceptions.py
```diff
@@ -130,6 +130,27 @@ class RunInterruptedException(ZenMLBaseException):
...
+
+        Args:
+            message: Message with details of the exception.
+            original_exception: The original step failure that this wraps.
+        """
+        super().__init__(message)
+        self.original_exception = original_exception
+
+
+StepExecutionException.__setstate__
```
bcdurak: Just out of curiosity, what is this for?

bcdurak (approved):
Aside from a small nitpicky question, looks good.

src/zenml/exceptions.py
schustmi: Oops, I was testing something else and must have left this in there. Removed it now

PR #4955, opened 2026-06-17
Use uv as a default when fetching installed python packages

schustmi:
Use `uv` as a default when fetching installed python packages


bcdurak (approved):
Nice, finally 🏎️

PR #4964, opened 2026-06-18
Fix logging shutdown deadlock

schustmi:
#4866 swapped the order in `LoggingContext.__exit__` to fix the following bug: Before that PR, logs that happen during the flush happening for a logging context ended up in the logs of the parent context.

This order change however introduced a deadlock in combination with `fsspec` based artifact log stores and debug logs enabled. This PR now fixes that by disabling the log store during shutdown.

### DIFF
--- a/src/zenml/utils/logging_utils.py
+++ b/src/zenml/utils/logging_utils.py
@@ -16,10 +16,21 @@
 import logging
 import os
 import threading
+from contextlib import contextmanager
 from contextvars import ContextVar
 from datetime import datetime
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Type,
+    Union,
+    cast,
+)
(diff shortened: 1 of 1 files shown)


src/zenml/utils/logging_utils.py
```diff
@@ -299,31 +316,40 @@ def __exit__(
...
-                        step_names_in_console.reset(self._step_names_token)
-                        self._step_names_token = None
+        # Disable the context before shutting down. This is needed to avoid
+        # deadlocks that would happen if the shutdown process emits logs to
+        # the active logging context. This is for example the case for asy
+        # fsspec based artifact stores, which copy the context vars of the
+        # thread they're called from and then emit debug logs.
+        with self.disabled(), self._lock:
+            try:
+                self.end()
```
bcdurak, line 344: If `self.end()` raises during deregister, end-of-stream emit, flush, exporter shutdown, or artifact-store finalize, `super().__exit__()` is skipped. That means the `ContextVar` remains set to this exiting context. After the `disabled()` manager unwinds, `_disabled` is restored, so later logs can route into a stale/half-deregistered context.

bcdurak (changes requested):

bcdurak (approved):

PR #4969, opened 2026-06-18
Improve listing pipelines sorted by latest run

schustmi:
Restructure the latest-run subquery and add a supporting index to improve the query plan when sorting pipelines by their latest run.

**Explanation:**
The latest-run subquery joined pipeline_run into the pipeline scan and grouped by pipeline.id, which made MySQL scan every run of every pipeline to compute `MAX(created)`. The new query aggregates over pipeline runs first (which uses the new `pipeline_id, created` index), and then joins the result with the outer query.

**Improvements:**
- locally: 20k pipelines, ~5x improvement on the raw DB query
- remote: 5k pipelines, 1.2mil runs, ~20x improvement on the client method (median 7s -> 330ms)

### DIFF
--- a/src/zenml/models/v2/core/pipeline.py
+++ b/src/zenml/models/v2/core/pipeline.py
@@ -397,7 +397,7 @@ class PipelineFilter(ProjectScopedFilter, TaggableFilter):
         Returns:
             The query with sorting applied.
         """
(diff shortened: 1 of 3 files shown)


bcdurak:
Can't accept this without knowing what comes after `remote:` 😅

bcdurak (approved):

PR #4977, opened 2026-06-22
Support for resource pools v2

stefannica:
## Overview

This PR updates resource pool support to use the new ZenML Pro Resource Manager as the source of truth for scheduling, allocation, lease management, and eviction decisions. The ZenML workspace now keeps only the runtime-facing resource request surface needed by pipeline execution, while the old resource pool control-plane entities are removed from the OSS API, client, CLI, RBAC, DB schemas, and local store implementation.

It also connects allocated resources all the way into dynamic step execution. Resource settings can now describe richer resource demands, ZenML creates Resource Manager-backed requests for eligible dynamic steps, waits for allocation before submitting remote work, and passes the resulting allocation into orchestrators or step operators that explicitly support it. Kubernetes is the first implementation to opt in: it applies Resource Manager-selected component settings and translates allocated CPU, memory, and GPU capacity into Kubernetes pod requests and limits.

```mermaid
flowchart TD
    A[Dynamic step run is created] --> B[ZenML builds Resource Manager demands<br/>from ResourceSettings]
    B --> C[Workspace creates resource request<br/>through Resource Manager]
    C --> D{Allocated?}

    D -- queued/pending --> E[StepLauncher polls and renews<br/>pending request lease]
    E --> D

    D -- allocated --> F[StepLauncher receives allocated<br/>ResourceRequestResponse]
    F --> G{Execution target}

    G -- step operator --> H[submit_with_allocation]
    G -- dynamic orchestrator --> I[submit_isolated_step_with_allocation]

    H --> J[Kubernetes applies target settings<br/>and pod resource overrides]
    I --> J
    J --> K[Remote step starts]
    K --> L[Heartbeat renews lease while running]
    L --> M[Resources released on completion<br/>or cancellation]
```

Finally, resource request leases are integrated with the existing step heartbeat mechanism. Heartbeats now keep allocated requests alive, report runtime state back to Resource Manager, and turn reclaim or eviction states into graceful step cancellation.

```mermaid
sequenceDiagram
    participant Step as Running step
    participant HB as Heartbeat worker
    participant Store as ZenML SQL store
    participant RM as Resource Manager

    Step->>HB: start heartbeat worker
    loop every heartbeat interval
        HB->>Store: update_step_heartbeat(timeout)
        Store->>RM: renew resource request lease
        RM-->>Store: current request status
        alt request still valid
            Store-->>HB: step remains running
        else request preempting / cancelled / expired / released
            Store-->>HB: return cancelling status
            HB->>Step: interrupt main thread
        end
    end
    Step->>Store: step shutdown / final status
    Store->>RM: release resource request
```

## Breaking Changes

This change erases all data from the previous resource pools v1 implementation (pools, policies, resource requests. There is no migration path provided.

## Implementation Details

- Extends `ResourceSettings` with a richer resource demand model:
  - adds `PoolResourceDemand` and flexible `resources` input support for resource names, name-to-quantity maps, single demand objects, and lists of demands
  - supports exact resource names, resource kinds, units, capacity classes, resource selectors, and class selectors
  - adds `gpu_class`, `allocation_wait_timeout_seconds`, `initialization_lease_seconds`, and `reclaim_tolerance`
  - keeps backwards-compatible input handling for legacy `pool_resources` and `preemptible`
  - converts configured resources into canonical `ResourceRequestDemand` objects for Resource Manager requests

- Updates dynamic runtime selection:
  - dynamic steps that request basic infrastructure resources (`cpu`, `memory`, or `gpu`) now resolve to `StepRuntime.ISOLATED` when the orchestrator supports isolated steps
  - opaque/custom pool resources do not force isolation by themselves

- Adds allocation-aware launch hooks:
  - introduces `supports_resource_pool_allocation` on base orchestrators and step operators, defaulting to `False`
  - adds `submit_isolated_step_with_allocation(...)` on `BaseOrchestrator`
  - adds `submit_with_allocation(...)` and `launch_with_allocation(...)` on `BaseStepOperator`
  - preserves backwards compatibility by delegating to the existing methods for components that do not opt in

- Wires allocations through the dynamic step launcher:
  - waits for a dynamic step’s resource request to become allocated before launching remote execution
  - renews the request lease while waiting for allocation
  - honors allocation wait and initialization lease timeouts from `ResourceSettings`
  - passes the allocated `ResourceRequestResponse` into the selected step operator or dynamic orchestrator submission path

- Adds Kubernetes allocation support:
  - marks the Kubernetes orchestrator and Kubernetes step operator as supporting resource pool allocations
  - applies Resource Manager target component settings on top of static stack/component settings
  - merges Resource Manager allocation-derived pod resources on top of existing pod requests and limits
  - converts allocated CPU, memory, and GPU quantities into Kubernetes-compatible resource keys and units

- Minimizes the workspace resource pool API:
  - removes legacy resource pool, policy, grant, queue, and demand control-plane models and endpoints from the workspace
  - removes the resource pool CLI group and old client/store methods
  - keeps resource request listing and retrieval for runtime visibility
  - adds `renew` and `release` resource request operations

- Reuses heartbeats for lease renewal and eviction:
  - extends heartbeat requests with `heartbeat_liveness_timeout_seconds`
  - renews resource request leases from SQL store heartbeat updates
  - reports runtime state transitions such as pending, submitted, and running
  - keeps the heartbeat worker alive until step shutdown completes, so leases continue to renew during graceful termination
  - maps Resource Manager eviction/reclaim statuses to step cancellation
  - validates that reclaim-tolerant isolated steps have heartbeat enabled

- Reflects ZenML Pro account privileges:
  - maps the external Pro user `is_superuser` property to ZenML’s internal `is_admin` user property during auth/user synchronization

- Defines a resource pool store interface that delegates runtime request storage to Resource Manager:
  - adds a Resource Manager-backed resource pool store facade (implemented separately in https://github.com/zenml-io/zenml-cloud-plugins/pull/55)
  - stores only the external `resource_request_id` on step runs instead of local resource request rows

## Pre-requisites

## Types of changes


src/zenml/orchestrators/base_orchestrator.py
```diff
@@ -557,10 +558,25 @@ def can_run_isolated_steps(self) -> bool:
...
             getattr(self.submit_isolated_step, "__func__", None)
             is not BaseOrchestrator.submit_isolated_step
+            or getattr(
+                self.submit_isolated_step_with_allocation, "__func__", None
+            )
+            is not BaseOrchestrator.submit_isolated_step_with_allocation
         )
 
+    @property
+    def supports_resource_pool_allocation(self) -> bool:
```
Json-Andriopoulos, line 568: Minor: To reduce the code a little bit, why not follow this pattern:

```
class Base:
    SUPPORTS_FEATURE_X: bool = False

    @property
    def feature_x_is_supported(self) -> bool:
        return self.SUPPORTS_FEATURE_X
```

Then every class that needs to set it to true just sets the class variable:

```
class ChildClass(Base):
    SUPPORTS_FEATURE_X: bool = True
```

This should help get rid of all those properties and compress the code a bit. Also keeps the implementation in one place, so if we want to generalise it to support a richer feature support mapping per orchestrator (like multiple feature support options) we can also do it in one place.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -480,12 +487,15 @@ def _run_step_with_step_operator(
         self,
         step_operator_name: Optional[str],
         step_run_info: StepRunInfo,
+        allocated_resource_request: Optional[ResourceRequestResponse],
```
Json-Andriopoulos, line 484: Very minor but you can probably get an agent to do it: We should move towards a more python3.10+ compatible typing approach `ResourceRequestResponse | None`.

src/zenml/config/resource_settings.py
```diff
@@ -65,6 +73,59 @@ def byte_value(self) -> int:
 
 
 MEMORY_REGEX = r"^[0-9]+(" + "|".join(unit.value for unit in ByteUnit) + r")$"
+BASIC_RESOURCE_KINDS = frozenset({"cpu", "gpu", "memory"})
```
Json-Andriopoulos: I am wondering if those should be under constants.py?

docs/book/getting-started/zenml-pro/resource-pools-user-guide.md
```diff
@@ -0,0 +1,197 @@
...
+## Custom resources
+
+Use custom resources when your admin created descriptors outside the stock
+CPU, memory, and GPU language. For example, a training license:
+
+```python
+from zenml.config import ResourceSettings
+from zenml.models import ResourceRequestDemand
+
+license_settings = ResourceSettings(
```
Json-Andriopoulos, line 84: This example does not work. Running it as is gets:

```
ValidationError: 4 validation errors for ResourceSettings
resources.str
  Input should be a valid string [type=string_type, input_value=[ResourceRequestDemand(re...e, class_selector=None)], input_type=list]
    For further information visit https://errors.pydantic.dev/2.12/v/string_type
resources.dict[str,constrained-int]
  Input should be a valid dictionary [type=dict_type, input_value=[ResourceRequestDemand(re...e, class_selector=None)], input_type=list]
    For further information visit https://errors.pydantic.dev/2.12/v/dict_type
resources.PoolResourceDemand
  Input should be a valid dictionary or instance of PoolResourceDemand [type=model_type, input_value=[ResourceRequestDemand(re...e, class_selector=None)],
input_type=list]
    For further information visit https://errors.pydantic.dev/2.12/v/model_type
resources.list[PoolResourceDemand].0
  Input should be a valid dictionary or instance of PoolResourceDemand [type=model_type, input_value=ResourceRequestDemand(res...ne, class_selector=None),
input_type=ResourceRequestDemand]
    For further information visit https://errors.pydantic.dev/2.12/v/model_type
```

src/zenml/integrations/kubernetes/kube_utils.py
```diff
@@ -1643,6 +1649,215 @@ def apply_default_resource_requests(
...
+    """
+    if not allocated_resource_request:
+        return settings
+
+    component_settings = (
+        allocated_resource_request.get_resources().component_settings
+    )
+
+    settings_dict = settings.model_dump(exclude_unset=True)
+    settings_dict.update(component_settings)
```
Json-Andriopoulos, line 1678: This shallow merge may potentially be unsafe: E.g. 

```
settings_dict = {
    "pod_settings": {
        "labels": {"team": "ml"},
        "tolerations": [...],
    }
}

component_settings = {
    "pod_settings": {
        "node_selectors": {"instance-type": "gpu-pool"}
    }
}
```

Would produce:

```
{
    "pod_settings": {
        "node_selectors": {"instance-type": "gpu-pool"}
    }
}
```

I think we need a utility that recursively merges the dictionariesl.

docs/book/getting-started/zenml-pro/resource-pools-user-guide.md
stefannica: done

src/zenml/orchestrators/step_launcher.py
```diff
@@ -786,76 +808,136 @@ def _get_initial_step_run_status(self) -> ExecutionStatus:
...
+                        runtime_state=ResourceRequestRuntimeState.SUBMITTED,
+                    ),
+                )
                 logger.info(
                     "Resource request `%s` for step `%s` was approved.",
                     resource_request.id,
-                    step_run_info.pipeline_step_name,
+                    step_name,
                 )
                 publish_utils.publish_step_run_status_update(
```
Json-Andriopoulos, line 899: This is more related to resource pools v1 I think but there is a logical bug/side-effect here. Consider the following scenario:

1) Resource Manager & Actual infra capacity diverge. What seems to be available in the resource pool is not actually available in the Kubernetes cluster.
2) We update the status here to `running` but the pod can not be actually allocated and remains pending. Instead of provisioning, the status seems running and the step appears to be stack mid-execution.

Json-Andriopoulos (changes requested):

src/zenml/orchestrators/step_launcher.py
stefannica: I'm aware of this behavior, and I even had some changes that were meant to fix it, but they were too complex and too disruptive. I took them out of this PR to reduce to cognitive burden required to test it/ review it. I have a follow-up PR that will correct this after the MVP version is released.

src/zenml/config/resource_settings.py
```diff
@@ -234,38 +370,133 @@ def get_memory(
...
+                    class_name=self.gpu_class,
+                )
+            )
 
         if self.cpu_count is not None:
-            merged["mcpu"] = math.ceil(self.cpu_count * 1000)
+            demands.append(
+                ResourceRequestDemand(
+                    kind="cpu",
+                    quantity=math.ceil(self.cpu_count),
```
Json-Andriopoulos: This may be confusing for users. If I request 0.5 CPU I wouldn't expect to occupy 1 CPU. I think we should keep cpu_count as integer and maybe add an mcpu field or convert fractions to mCPU under the hood.

src/zenml/integrations/kubernetes/kube_utils.py
stefannica: This one is tricky: it's not so much a bug as it is a feature. I don't want to have to deep merge user settings and resource pool settings, because that would potentially allow users to circumvent the settings that the resource pool allocation logic "forces" the orchestrator to use and gain access to any resources they want. Not to mention it may lead to all sorts of chimera monsters that result from incorrectly merging two deeply nested settings.

To avoid all that, I chose to use a shallow merge. Not unsafe. Intentional.

src/zenml/config/resource_settings.py
stefannica: `cpu_count` was already a float, I didn't introduce this type as a change in this PR. But you're right here in saying that we should use the right mcpu unit if it's not an integer value. I'll update the code.

src/zenml/orchestrators/base_orchestrator.py
stefannica: Three reasons why I would like to keep using properties instead of the pattern you suggest:

1. we use properties instead of class variables in a few other places in ZenML (flavor, stack component)
2. deciding whether a feature is supported or not might be a dynamic decision, not always a static one (the class var approach doesn't support that)
3. as a personal preference, I hate class variables and I try to avoid them as much as possible. They can easily get out of hand and make the code an undocumented mess. The property is something that is clear and has clear docstrings documenting what it does.

src/zenml/execution/pipeline/dynamic/compilation.py
```diff
@@ -320,12 +320,12 @@ def get_step_runtime(
     if not orchestrator.can_run_isolated_steps:
         return StepRuntime.INLINE
 
-    runtime = step_config.runtime
+    if step_config.resource_settings.has_basic_resource_demands:
+        return StepRuntime.ISOLATED
```
Json-Andriopoulos: Minor, let me know what you think. Let's say as a user I do:

```
@step(runtime=StepRuntime.INLINE)
def inline_step():
    pass

@step(runtime=StepRuntime.ISOLATED)
def iso_step_1():
    pass

@step(runtime=StepRuntime.ISOLATED)
def some_other_iso_step():
    pass

@pipeline(settings={"resources": "..."})
def pipeline():
    pass
```

What happens here is that the inline step even though defined as such is executed in isolation as it inherits the pipeline's ResourceSettings. What the user may expect is something different: isolated steps to use the ResourceSettings and the inline step to run as inline and ignore the ResourceSettings of the pipeline.

I think the best course is either a) Warn about running is isolation b) Consider as INLINE and maybe apply empty ResourceSettings.

stefannica: Yes, correct. I'd pick a) and warn people.

src/zenml/orchestrators/step_launcher.py
```diff
@@ -786,76 +808,136 @@ def _get_initial_step_run_status(self) -> ExecutionStatus:
...
-            elif resource_request.status == ResourceRequestStatus.CANCELLED:
+            if resource_request.status == ResourceRequestStatus.CANCELLED:
+                reason = resource_request.status_reason or "Unknown reason"
                 raise RuntimeError(
                     f"Resource request `{resource_request.id}` for step "
-                    f"`{step_run_info.pipeline_step_name}` was cancelled."
+                    f"`{step_name}` was cancelled: {reason}"
                 )
 
             logger.info(
```
Json-Andriopoulos: A low priority workload may wait for several hours before being prioritized which will generate hundreds of logs like this:

```
[INFO] [2026-08-26T07:43:59.420217Z] Step `low_priority_workload` has started.
[INFO] [2026-08-26T07:43:59.420345Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:43:59.420387Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:44:04.514520Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:44:04.514670Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:44:09.645226Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:44:24.792120Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:44:44.958874Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
[INFO] [2026-08-26T07:45:05.117035Z] Waiting for resource request `af4fee29-cb0e-4b9f-a48e-6e1388b199df` of step `low_priority_workload` to be approved...
```

I think we should cache it and display once or more infrequently

src/zenml/client.py
```diff
@@ -2276,235 +2272,68 @@ def delete_stack_component(
...
         """
-        return self.zen_store.get_resource_pool_subject_policy(
-            policy_id=policy_id, hydrate=hydrate
+        return self.zen_store.get_resource_request(
+            resource_request_id=resource_request_id,
+            hydrate=hydrate,
         )
 
-    def list_resource_pool_subject_policies(
+    def list_resource_requests(
```
Json-Andriopoulos, line 2292: project would also be useful here

src/zenml/integrations/kubernetes/kube_utils.py
Json-Andriopoulos: While I get the reasoning, as is it remains unacceptable imo.  This is breaking existing functionality - we should at least warn users about this behavior on both the docs and the runtime.

src/zenml/orchestrators/step_launcher.py
Json-Andriopoulos: Cool, resolving this. Please remember to add a backlog item :)

src/zenml/config/resource_settings.py
Json-Andriopoulos: Maybe we can do it as a follow-up - it doesn't need to block the PR.

src/zenml/client.py
stefannica: This was also a bit incompletely implemented. The CLI commands went straight to the zen store instead of calling Client API methods, so I added those too.

stefannica: done

src/zenml/orchestrators/step_launcher.py
stefannica: done

src/zenml/execution/pipeline/dynamic/compilation.py
stefannica: done

src/zenml/config/resource_settings.py
stefannica: done

Json-Andriopoulos (approved):

PR #4978, opened 2026-06-22
Allow overwriting step inputs for all invocations when replaying

schustmi:
When replaying a run, you can now override a step input for **every invocation of a step** using `step_default_input_overrides`, keyed by the step's name. The existing per-invocation `step_input_overrides` still take precedence per input key.

**Limitations:**
- Step names are not enforced to be unique. If two different steps share a name, a `step_default_input_overrides` entry applies to every step with that name.

### DIFF
--- a/docs/book/user-guide/tutorial/replaying-runs-steps.md
+++ b/docs/book/user-guide/tutorial/replaying-runs-steps.md
@@ -173,6 +173,52 @@ Client().replay_pipeline_run(
 {% endtab %}
 {% endtabs %}
 
+#### Overriding an input for every invocation of a step
+
+If you want to override inputs for **every invocation** of a step, you can
+use `step_default_input_overrides`. It is keyed by the step's name instead of
+the invocation ID. Per-invocation overrides in `step_input_overrides` take
+precedence per input key.
+
+{% hint style="warning" %}
+The override applies to every invocation that shares the given name, and ZenML
+does not enforce that a name maps to a single step function. If two different
+steps share a name (for example two functions named `train` in different
+modules both default to the name `train`), the override is applied to both.
+Give such steps distinct names if you want to target only one of them.
+{% endhint %}
+
+{% tabs %}
+{% tab title="From local code" %}
+```python
+training_pipeline.replay(
+    pipeline_run="run_name_or_id",
+    step_default_input_overrides={
+        "llm_call": {
+            "model": "claude-fable-5"
+        }
+    },
+)
+```
+{% endtab %}
+{% tab title="From the server" %}
+```python
+from zenml import Client
+
+Client().replay_pipeline_run(
+    name_id_or_prefix="run_name_or_id",
+    run_configuration={
+        "step_default_input_overrides": {
+            "llm_call": {
+                "model": "claude-fable-5"
+            }
+        }
+    },
+)
+```
+{% endtab %}
+{% endtabs %}
+
 ---
 
 ## Replaying a single step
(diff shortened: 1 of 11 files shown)


docs/book/user-guide/tutorial/replaying-runs-steps.md
```diff
@@ -173,6 +173,52 @@ Client().replay_pipeline_run(
 {% endtab %}
 {% endtabs %}
 
+#### Overriding an input for every invocation of a step
```
bcdurak, line 176: I mentioned this yesterday but will write it here as well for visibility. This is a really neat feature that can serve a lot of different people. With the way the docs are structured right now, I think this needs to be in the main "ZenML" tab.

docs/book/user-guide/tutorial/replaying-runs-steps.md
```diff
@@ -173,6 +173,52 @@ Client().replay_pipeline_run(
...
+Client().replay_pipeline_run(
+    name_id_or_prefix="run_name_or_id",
+    run_configuration={
+        "step_default_input_overrides": {
+            "train": {
+                "training_data": my_dataframe,
+            }
+        }
+    },
+)
```
bcdurak, line 217: I have had a hunch about this one and I've tested it with a local server. I think it should have said something like "replaying isn't allowed on this server", but it failed before even reaching that point. 

The reason is how we handle the upload. When you use `training_pipeline.replay(...)` like in the example above, we call `upload_input_artifact_overrides(...)` before serializing the request that is meant to be sent to the server. If you do this over the client directly, it fails with: `PydanticSerializationError: Unable to serialize unknown type: <class 'pandas.DataFrame'>`.

I think the same comment applies to the existing overrides as well.

src/zenml/orchestrators/step_run_utils.py
```diff
@@ -115,7 +116,12 @@ def should_skip_step(self, invocation_id: str) -> bool:
         if not self.snapshot.pipeline_configuration.skip_successful_steps:
             return False
 
-        if self._get_input_overrides(invocation_id):
+        if self._get_input_overrides(
+            invocation_id,
+            step=step,
+            warn_on_invalid_keys=False,
+            include_step_defaults=False,
```
bcdurak, line 123: I was looking into when this value could be `False` and I have discovered the following behaviour:

- I run `my_pipeline` with `my_step` successfully.
- If I call the replay with `step_default_input_overrides`, `my_step` does not run again: 
  ```
  my_pipeline.replay(
    pipeline_run=run.id,
    skip_successful_steps=True,
    step_default_input_overrides={
        "my_step": {"value": "1"},
    },
  )
  ```
- However if I call it with `step_input_overrides`, it does run again:
  ```
  my_pipeline.replay(
    pipeline_run=run.id,
    skip_successful_steps=True,
    step_input_overrides={
        "my_step": {"value": "2"},
    },
  )
  ```

I think this is a bit inconsistent.

bcdurak (changes requested):
I have tested this in different settings, it mostly looks good. But I think we need to address some issues.

docs/book/user-guide/tutorial/replaying-runs-steps.md
schustmi: Updated the docs

src/zenml/orchestrators/step_run_utils.py
schustmi: As discussed, inconsistent but intended :)

bcdurak: I admit defeat 😃

bcdurak (approved):

PR #4982, opened 2026-06-23
Misc DB improvements

schustmi:
- Use query options when listing artifact versions
- Only use `DISTINCT` in query if the query joins any table. In all other cases, the distinct is unnecessary and just slows down the query


bcdurak (approved):

bcdurak (approved):
Good changes!

PR #4990, opened 2026-06-24
Add isolated step cleanup hooks

schustmi:
This PR provides some hooks that can be used in step operator or orchestrator implementations to run some cleanup after an isolated step finishes. For example, clean up secrets created for container registry authentication.

### DIFF
--- a/src/zenml/execution/pipeline/dynamic/runner.py
+++ b/src/zenml/execution/pipeline/dynamic/runner.py
@@ -599,6 +599,10 @@ class DynamicPipelineRunner:
                             wrap_step_failure(exc, invocation_id=invocation_id)
                         )
 
+                # Either infra or DB status indicates that the step has
+                # finished -> we run the step cleanup
+                self._cleanup_isolated_step(step_run)
+
                 # Get the updated status that we might have just published
                 db_status = step_run.status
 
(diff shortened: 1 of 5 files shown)


src/zenml/orchestrators/step_launcher.py
```diff
@@ -535,11 +535,16 @@ def _run_step_with_step_operator(
...
-                    environment=environment,
-                )
+                try:
+                    step_operator.launch(
+                        info=step_run_info,
+                        entrypoint_command=entrypoint_command,
+                        environment=environment,
+                    )
+                finally:
+                    step_operator.cleanup_step_submission(
```
bcdurak: I think this is covered in all the other cases but since `step_operator.cleanup_step_submission` is called directly here, a failure in the cleanup logic can mask any success/failure in the actual `launch` method, right?

src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -552,6 +552,10 @@ def _monitoring_loop(self) -> None:
 
                 self._steps_to_monitor.pop(invocation_id, None)
 
+                # Either infra or DB status indicates that the step has
+                # finished -> we run the step cleanup
+                self._cleanup_isolated_step(step_run)
```
bcdurak: Right below this line, the status updates happen. For instance, if the `db_status` is `STOPPING`, we will publish a stopped step run. However, before doing that we pass a `step_run` to the clean up logic that might still have a `RUNNING` status. Do you think we can do the clean up after the status update?

bcdurak (changes requested):

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: Done

src/zenml/orchestrators/step_launcher.py
schustmi: Done

bcdurak (approved):

PR #4991, opened 2026-06-24
Improved snapshot fetching

schustmi:
Results on DB with ~1M snapshots:
- ~10% quicker for hydrated list calls (page size 20)
- ~15% quicker for unhydrated list calls (page size 20)
- ~20% quicker for hydrated list calls (page size 200)
- ~30% quicker for unhydrated list calls (page size 200)

Unhydrated improvements are for relatively small values in `pipeline_configuration` and `client_environment` columns, so might be even more for most users.

### DIFF
--- a/src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
+++ b/src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
@@ -19,7 +19,7 @@ from uuid import UUID
 
 from sqlalchemy import TEXT, CheckConstraint, Column, String, UniqueConstraint
 from sqlalchemy.dialects.mysql import MEDIUMTEXT
-from sqlalchemy.orm import joinedload, object_session, selectinload
+from sqlalchemy.orm import defer, object_session, selectinload
 from sqlalchemy.sql.base import ExecutableOption
 from sqlmodel import Field, Relationship, asc, col, desc, select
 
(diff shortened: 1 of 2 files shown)


src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
```diff
@@ -372,21 +372,29 @@ def get_query_options(
...
+                    ),
+                    defer(jl_arg(PipelineSnapshotSchema.client_environment)),
                 ]
             )
 
         if include_resources:
             options.extend(
                 [
-                    joinedload(jl_arg(PipelineSnapshotSchema.user)),
+                    selectinload(jl_arg(PipelineSnapshotSchema.stack)),
```
bcdurak, line 390: Seeing this reminded me of the changes for the artifact versions yesterday, should we add the tags and deployment here as well?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -5334,6 +5334,7 @@ def list_snapshots(
                 table=PipelineSnapshotSchema,
                 filter_model=snapshot_filter_model,
                 hydrate=hydrate,
+                apply_query_options_from_schema=True,
```
bcdurak, line 5337: Very very minor nitpick: We do the trick here for the list calls which depends on the `include_metadata` parameter. Theoretically speaking, can we also use a similar trick in the `get_snapshot` call like: 

```python
snapshot = self._get_schema_by_id(
    resource_id=snapshot_id,
    schema_class=PipelineSnapshotSchema,
    session=session,
    query_options=PipelineSnapshotSchema.get_query_options(
        include_metadata=hydrate,
        include_resources=True,
    ),
)
```

bcdurak (commented):
Leaving some comments regarding some things I am curious about. Other than that, the changes look alright.

src/zenml/zen_stores/sql_zen_store.py
schustmi: No, in the `get_snapshot` case it actually makes the timing worse.

src/zenml/zen_stores/schemas/pipeline_snapshot_schemas.py
schustmi: done

bcdurak (approved):

PR #4992, opened 2026-06-24
OAuth2 service connector

schustmi:
This PR adds a generic OAuth2 service connector, which can be configured either with a static token, client ID and secret or client ID and refresh token.

Limitations:
- Rotating refresh tokens are not supported yet, as there is no public API for updating the secrets that back credentials of a service connector. Can be extended in the future.

### DIFF
--- a/src/zenml/service_connectors/oauth2_service_connector.py
+++ b/src/zenml/service_connectors/oauth2_service_connector.py
@@ -1,0 +1,385 @@
+#  Copyright (c) ZenML GmbH 2026. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at:
+#
+#       https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing
+#  permissions and limitations under the License.
+"""OAuth2 Service Connector."""
+
+from datetime import datetime, timedelta
+from typing import Any, Dict, List, Optional, Tuple
+
+import requests
+from pydantic import Field
+
+from zenml.exceptions import AuthorizationException
+from zenml.logger import get_logger
+from zenml.models import (
+    AuthenticationMethodModel,
+    ResourceTypeModel,
+    ServiceConnectorTypeModel,
+)
+from zenml.service_connectors.service_connector import (
+    AuthenticationConfig,
+    ServiceConnector,
+)
+from zenml.utils.enum_utils import StrEnum
+from zenml.utils.secret_utils import PlainSerializedSecretStr
+from zenml.utils.time_utils import utc_now
+
+logger = get_logger(__name__)
+
+OAUTH2_CONNECTOR_TYPE = "oauth2"
+OAUTH2_API_RESOURCE_TYPE = "oauth2-api"
+OAUTH2_SESSION_EXPIRATION_BUFFER = 15  # 15 minutes
+
+
+class OAuth2AuthenticationMethods(StrEnum):
+    """OAuth2 authentication methods."""
+
+    CLIENT_CREDENTIALS = "client-credentials"
+    REFRESH_TOKEN = "refresh-token"
+    TOKEN = "token"
+
+
+class OAuth2BaseConfig(AuthenticationConfig):
+    """OAuth2 base configuration."""
+
+    api_url: str = Field(title="API base URL")
+
+
+class OAuth2TokenConfig(OAuth2BaseConfig):
+    """OAuth2 token configuration."""
+
+    access_token: PlainSerializedSecretStr = Field(title="OAuth2 access token")
+
+
+class OAuth2GrantConfig(OAuth2BaseConfig):
+    """OAuth2 base configuration for grants that mint access tokens."""
+
+    token_url: str = Field(title="OAuth2 token endpoint URL")
+    client_id: str = Field(title="OAuth2 client ID")
+    scope: Optional[str] = Field(
+        default=None,
+        title="Space-separated OAuth2 scopes",
+    )
+
+
+class OAuth2ClientCredentialsConfig(OAuth2GrantConfig):
+    """OAuth2 client credentials configuration."""
+
+    client_secret: PlainSerializedSecretStr = Field(
+        title="OAuth2 client secret",
+    )
+    audience: Optional[str] = Field(default=None, title="OAuth2 audience")
+
+
+class OAuth2RefreshTokenConfig(OAuth2GrantConfig):
+    """OAuth2 refresh token configuration."""
+
+    refresh_token: PlainSerializedSecretStr = Field(
+        title="OAuth2 refresh token",
+    )
+    client_secret: Optional[PlainSerializedSecretStr] = Field(
+        default=None,
+        title="OAuth2 client secret",
+    )
+
+
+OAUTH2_SERVICE_CONNECTOR_TYPE_SPEC = ServiceConnectorTypeModel(
+    name="OAuth2 Service Connector",
+    connector_type=OAUTH2_CONNECTOR_TYPE,
+    description="""
+The ZenML OAuth2 Service Connector authenticates to an HTTP API using OAuth2.
+
+The client credentials and refresh token methods exchange long-lived
+credentials for a temporary access token at a token endpoint. Connector
+consumers only ever receive the temporary token, never the underlying
+credentials. A pre-obtained access token can also be supplied directly.
+
+Connecting returns an authenticated requests Session with the bearer token set
+on the Authorization header.
+""",
+    emoji=":closed_lock_with_key:",
+    auth_methods=[
+        AuthenticationMethodModel(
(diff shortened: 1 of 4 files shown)


src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+        expires_at: Optional[datetime] = None
+        expires_in = payload.get("expires_in")
+        if expires_in is not None:
+            expires_at = utc_now(tz_aware=True) + timedelta(
+                seconds=int(expires_in)
+            )
+
+        return token, expires_at
+
+    def _get_token(self) -> Tuple[str, Optional[datetime]]:
```
bcdurak, line 343: There is an interesting interaction that I found with how we use `_get_token`. Right now, `get_connector_client()` calls `_verify()` and `_verify()` calls `_get_token()`, which fetches a token and puts in the cache. Afterwards, `_get_connector_client()` calls `_get_token()` again. If the cached token has less than 15 minutes of lifetime, the second `_get_token()` call refuses to reuse the token that was just fetched during verification and fetches another one. Not sure if this is a big issue, it could mean duplicate calls for 5-to-15 minute access tokens, so I just wanted to put a comment here.

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+        Args:
+            kwargs: Additional implementation specific keyword arguments.
+
+        Returns:
+            An authenticated requests Session.
+        """
+        token, _ = self._get_token()
+
+        session = requests.Session()
+        session.headers["Authorization"] = f"Bearer {token}"
```
bcdurak, line 414: Very small nitpick. If we are doing this, It might be nice to verify that the `token_type` of the response is always `Bearer` when we parse it.

bcdurak (approved):
Putting two small comments here, feel free to ignore. Otherwise, it looks good.

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+client credentials grant.
+""",
+            config_class=OAuth2ClientCredentialsConfig,
+        ),
+        AuthenticationMethodModel(
+            name="OAuth2 Refresh Token",
+            auth_method=OAuth2AuthenticationMethods.REFRESH_TOKEN,
+            description="""
+Exchange a refresh token for a temporary access token using the OAuth2 refresh
+token grant. Rotating refresh tokens are not supported.
```
stefannica, line 179:
```suggestion
token grant. Rotating refresh tokens is not supported.
```

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+        Returns:
+            The access token and its expiration timestamp, if known.
+        """
+        if isinstance(self.config, OAuth2TokenConfig):
+            # Use the user-configured expires_at timestamp on the connector
+            # configuration for static tokens.
+            return self.config.access_token.get_secret_value(), self.expires_at
+
+        key = self.auth_method
+        if key in self._token_cache:
```
stefannica, line 355: Just to clarify, this cache is not reused across different clients. What happens inside the server is every time a client makes a call to "get connector client" or to "verify" or to anything that has to do with this service connector, a new service connector instance is created for the lifetime of that request alone. The service connector instance won't be shared across different requests from different clients or even from the same client for that matter. So the only reason why this cache is useful is to not make multiple requests to the target OAuth2 API as part of the processing of a single client API call like "get connector client".

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+            raise AuthorizationException(
+                f"The OAuth2 token endpoint at {cfg.token_url} returned an "
+                f"invalid access token."
+            )
+
+        expires_at: Optional[datetime] = None
+        expires_in = payload.get("expires_in")
+        if expires_in is not None:
+            expires_at = utc_now(tz_aware=True) + timedelta(
+                seconds=int(expires_in)
```
stefannica: If `expires_in` is not a valid integer, this will still surface as an authorization error, but I think it's better if you catch this explicitly.

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+            response = requests.post(
+                cfg.token_url,
+                data=data,
+                auth=auth,
+                timeout=30,
+            )
+            response.raise_for_status()
+            payload = response.json()
+            token = payload["access_token"]
+        except (requests.RequestException, KeyError, ValueError) as e:
```
stefannica: non-dict JSON payloads can raise TypeError too

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+                data["client_id"] = cfg.client_id
+
+        if cfg.scope:
+            data["scope"] = cfg.scope
+
+        try:
+            response = requests.post(
+                cfg.token_url,
+                data=data,
+                auth=auth,
```
stefannica, line 291: This uses the so-called basic authentication, essentially passing the authorization credentials using a basic header:

```
POST /oauth/token HTTP/1.1
Host: auth.example.com
Content-Type: application/x-www-form-urlencoded
Authorization: Basic base64(urlencode(client_id):urlencode(client_secret))

grant_type=client_credentials&scope=read+write
```

However, there is a second way to authenticate to OAuth2 servers that includes the client ID and client secret inside the URL call:

```
POST /oauth/token HTTP/1.1
Host: auth.example.com
Content-Type: application/x-www-form-urlencoded

grant_type=client_credentials&
client_id=my-client&
client_secret=my-secret&
scope=read+write
```


I think we should support both and have this configured, have this modeled as part of the service connector configuration. Something like a `client_auth_method` maybe ?

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+        cfg = self.config
+        assert isinstance(cfg, OAuth2GrantConfig)
+
+        data: Dict[str, str] = {}
+        auth: Optional[Tuple[str, str]] = None
+
+        if isinstance(cfg, OAuth2ClientCredentialsConfig):
+            data["grant_type"] = "client_credentials"
+            if cfg.audience:
+                data["audience"] = cfg.audience
```
stefannica, line 260: A simple improvement that would make this compatible with more services out there: add generic token endpoint parameters.

Keep the current first-class fields like scope and audience, but add something like:

```
token_params: Dict[str, str] = Field(default_factory=dict)
```

This covers common provider-specific fields:
* resource for some Microsoft/Azure AD v1 style flows
* audience for Auth0-style APIs, if you don’t want it hardcoded as a top-level field
* vendor-specific flags
* custom API/resource identifiers

I’d validate that these params cannot override reserved fields unless intentionally allowed:

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+            else:
+                # Public clients without a secret send the client ID in the
+                # request body.
+                data["client_id"] = cfg.client_id
+
+        if cfg.scope:
+            data["scope"] = cfg.scope
+
+        try:
+            response = requests.post(
```
stefannica, line 288: Another improvement idea: add optional token request headers.

Some providers or gateways require extra headers.

```
token_headers: Dict[str, str] = Field(default_factory=dict)
```

I would not let users override `Authorization` casually because it's used for the basic authentication case.

src/zenml/service_connectors/oauth2_service_connector.py
```diff
@@ -0,0 +1,385 @@
...
+                cfg.token_url,
+                data=data,
+                auth=auth,
+                timeout=30,
+            )
+            response.raise_for_status()
+            payload = response.json()
+            token = payload["access_token"]
```
stefannica: You could also validate unexpected `token_type`, especially if it is present and not `Bearer`

stefannica (changes requested):
I totally love this. I left some improvement suggestions that will open this up for use with even more providers and services out there.

src/zenml/service_connectors/oauth2_service_connector.py
schustmi: Yes, I thought it might be worth just having it anyway as it doesn't really cost us anything. I think the other service connectors also keep a cache on the instance.

src/zenml/service_connectors/oauth2_service_connector.py
schustmi: done

src/zenml/service_connectors/oauth2_service_connector.py
schustmi: done

src/zenml/service_connectors/oauth2_service_connector.py
schustmi: This applies to any token expiration that is less than the cache duration, and also applies when not having a cache at all. So I don't think there's really anything we can do?

src/zenml/service_connectors/oauth2_service_connector.py
schustmi: done

stefannica (approved):

PR #4993, opened 2026-06-24
Add opt-in Modal app cleanup after sandbox completion

raza-khan0108:
## Describe changes

After a step sandbox finishes, the Modal app entry remains in the Modal
dashboard as `deployed, Tasks: 0`. This is purely visual clutter — idle
apps incur no compute charges — but accumulates quickly for users running
many ZenML pipelines or steps on Modal (each step run creates one app
named `zenml-<step-run-id>-<step-name>`).

This PR adds a new `stop_app_after_run: bool = False` setting to
`ModalSettingsMixin` (shared by both the step operator and orchestrator
flavors). When enabled, ZenML overrides `cleanup_step_submission()` — the
hook introduced in #4990 — to call `modal app stop --yes <app-name>` via
subprocess after the sandbox reaches a terminal state. The app entry moves
from **deployed** → **stopped** in the Modal dashboard. Stopped apps are
still visible in the "recently stopped" section with their full log history
intact, so debugging is unaffected.

### Why subprocess instead of the Python SDK?

`modal.App` has no public `.stop()` method. The Modal CLI is the only
supported programmatic path for stopping a deployed app. We call it via
`subprocess.run()` and inject `MODAL_TOKEN_ID` / `MODAL_TOKEN_SECRET` from
ZenML's configured credentials so the call works regardless of ambient auth
state.

### Changes

- **`modal_base_flavor.py`**: add `stop_app_after_run: bool = False` to
  `ModalSettingsMixin`
- **`sandbox_utils.py`**: add `stop_modal_app()` helper that runs the Modal
  CLI and forwards token credentials
- **`modal_step_operator.py`**: override `cleanup_step_submission()` (from
  #4990) — early-returns when the setting is `False`, otherwise calls
  `stop_modal_app()` with warning-only error handling so cleanup can never
  mask the step result

### Usage

```python
@step(
    step_operator="my-modal-operator",
    settings={
        "step_operator": ModalStepOperatorSettings(stop_app_after_run=True)
    },
)
def my_step() -> None: ...

### DIFF
--- a/src/zenml/integrations/modal/flavors/modal_base_flavor.py
+++ b/src/zenml/integrations/modal/flavors/modal_base_flavor.py
@@ -68,6 +68,15 @@ class ModalSettingsMixin(BaseSettings):
         f"Examples: 3600 (1 hour), 7200 (2 hours), {DEFAULT_TIMEOUT_SECONDS} (24 hours maximum). "
         "Execution will be terminated if it exceeds this timeout",
     )
+    stop_app_after_run: bool = Field(
+        False,
+        description="Stop the Modal app after the sandbox reaches a terminal "
+        "state. When True, ZenML calls `modal app stop` on the app created "
+        "for this run, removing it from the Modal dashboard's deployed-apps "
+        "list. Stopped apps are still visible in the 'recently stopped' "
+        "section and their logs remain accessible. Defaults to False to "
+        "preserve the previous behavior.",
+    )
 
 
 class ModalCredentialsMixin(BaseSettings):
--- a/src/zenml/integrations/modal/flavors/modal_base_flavor.py
+++ b/src/zenml/integrations/modal/flavors/modal_base_flavor.py
@@ -68,15 +68,6 @@ class ModalSettingsMixin(BaseSettings):
         f"Examples: 3600 (1 hour), 7200 (2 hours), {DEFAULT_TIMEOUT_SECONDS} (24 hours maximum). "
         "Execution will be terminated if it exceeds this timeout",
     )
-    stop_app_after_run: bool = Field(
-        False,
-        description="Stop the Modal app after the sandbox reaches a terminal "
-        "state. When True, ZenML calls `modal app stop` on the app created "
-        "for this run, removing it from the Modal dashboard's deployed-apps "
-        "list. Stopped apps are still visible in the 'recently stopped' "
-        "section and their logs remain accessible. Defaults to False to "
-        "preserve the previous behavior.",
-    )
 
 
 class ModalCredentialsMixin(BaseSettings):
--- a/src/zenml/integrations/modal/sandbox_utils.py
+++ b/src/zenml/integrations/modal/sandbox_utils.py
@@ -14,6 +14,7 @@
 """Shared Modal Sandbox helpers."""
 
 import math
+import os
 import time
 from typing import Any, Dict, List, Optional, Protocol, Tuple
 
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by schustmi** [comment]
@raza-khan0108 A few things please:
- sign the CLA
- rebase on develop and use it as a base for your branch
- We'll merge #4990 zero, which introduces a method specifically for such cleanup, which means we don't have to abuse the `get_status` method, which is both ugly as well as unreliable. Once this is merged, I'd ask you to update your PR to use the cleanup method to delete the app.

**Comment by schustmi** [comment]
#4990 is merged now, so as soon as you rebase on develop you can use it.

**Comment by raza-khan0108** [comment]
Hi @schustmi — addressed all the feedback:

- ✅ Signed the CLA
- ✅ Rebased onto `develop` and changed the base branch
- ✅ Replaced the `get_status()` hook with `cleanup_step_submission()` from #4990
- ✅ Moved `stop_app_after_run` into `ModalSettingsMixin` in `modal_base_flavor.py` so it's shared by both the step operator and orchestrator

The PR description has been updated to reflect the new implementation. Let me know if anything else needs changing!

File: src/zenml/integrations/modal/sandbox_utils.py [hunk 1]
```diff
@@ -398,3 +399,56 @@ def terminate_sandbox(
...
+
+    env = os.environ.copy()
+    token_pair = resolve_modal_token_pair(
+        token_id=token_id, token_secret=token_secret
+    )
+    if token_pair:
+        env[MODAL_TOKEN_ID_ENV_KEY] = token_pair[0]
+        env[MODAL_TOKEN_SECRET_ENV_KEY] = token_pair[1]
+
+    result = subprocess.run(cmd, capture_output=True, text=True, env=env)
```
schustmi: Is it not possible to do this via the modal python sdk?

raza-khan0108: I looked into using the Modal Python SDK first, but there doesn’t appear to be a public API to stop an already deployed app programmatically (e.g., no `App.stop()` equivalent). The supported path today is `modal app stop`, so this PR uses the CLI via `subprocess.run()`.

I also made sure the call is non-blocking for pipeline outcome: cleanup is best-effort (warning-only on failure), and credentials are passed explicitly via `MODAL_TOKEN_ID` / `MODAL_TOKEN_SECRET` from the configured ZenML integration so it doesn’t depend on ambient auth state.

If Modal adds a public SDK method for this, I’m happy to switch this implementation to that.

src/zenml/integrations/modal/sandbox_utils.py
```diff
@@ -398,3 +399,56 @@ def terminate_sandbox(
...
+            ``None``, the ambient CLI credential is used.
+        token_secret: Modal token secret to forward as
+            ``MODAL_TOKEN_SECRET``. When ``None``, the ambient CLI
+            credential is used.
+
+    Raises:
+        RuntimeError: If the ``modal app stop`` command exits with a
+            non-zero status code.
+    """
+    import subprocess
```
schustmi: Import at the top

src/zenml/integrations/modal/sandbox_utils.py
```diff
@@ -398,3 +399,56 @@ def terminate_sandbox(
...
+    """Stop a deployed Modal app by name using the Modal CLI.
+
+    Moves the app from 'deployed' to 'stopped' in the Modal dashboard,
+    eliminating the visual clutter of idle ``Tasks: 0`` app entries.
+    Stopped apps remain visible in the 'recently stopped' section with
+    their logs intact, so debugging is not affected.
+
+    This should only be called after the associated Modal sandbox has
+    reached a terminal state, because stopping the app does not wait for
+    running sandboxes to finish.
```
schustmi:
```suggestion
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -267,3 +267,45 @@ def cancel(self, step_run: "StepRunResponse") -> None:
...
+
+    def cleanup_step_submission(self, step_run: "StepRunResponse") -> None:
+        """Stop the Modal app after a sandbox run completes.
+
+        When ``stop_app_after_run`` is enabled in the step settings, this
+        calls ``modal app stop`` on the per-step app, moving its entry from
+        'deployed' to 'stopped' in the Modal dashboard. Stopped apps remain
+        visible in the 'recently stopped' section with their full log history
+        intact, so debugging is unaffected. Failures are logged as warnings
+        and never re-raised so that cleanup cannot mask the step result.
```
schustmi:
```suggestion
```

src/zenml/integrations/modal/step_operators/modal_step_operator.py
```diff
@@ -267,3 +267,45 @@ def cancel(self, step_run: "StepRunResponse") -> None:
...
+
+        Args:
+            step_run: The finished step run.
+        """
+        settings = cast(ModalStepOperatorSettings, self.get_settings(step_run))
+        if not settings.stop_app_after_run:
+            return
+
+        app_name = f"zenml-{step_run.id}-{step_run.name}"[:64]
```
schustmi: This should be extracted in some helper function if it's used in multiple places.

src/zenml/integrations/modal/flavors/modal_base_flavor.py
```diff
@@ -68,6 +68,15 @@ class ModalSettingsMixin(BaseSettings):
         f"Examples: 3600 (1 hour), 7200 (2 hours), {DEFAULT_TIMEOUT_SECONDS} (24 hours maximum). "
         "Execution will be terminated if it exceeds this timeout",
     )
+    stop_app_after_run: bool = Field(
```
schustmi: This is added on the base settings class, but only respected when using the step operator. I think we should move it to the step operator settings, or implement it for all flavors inheriting from this (sandbox, orchestrator, step operator).

Also, I think `stop_app` or something similar makes more sense, as for example the step operator stops it after a step finishes, not a run.

src/zenml/integrations/modal/flavors/modal_base_flavor.py
```diff
@@ -68,6 +68,15 @@ class ModalSettingsMixin(BaseSettings):
...
+        False,
+        description="Stop the Modal app after the sandbox reaches a terminal "
+        "state. When True, ZenML calls `modal app stop` on the app created "
+        "for this run, removing it from the Modal dashboard's deployed-apps "
+        "list. Stopped apps are still visible in the 'recently stopped' "
+        "section and their logs remain accessible. Defaults to False to "
+        "preserve the previous behavior.",
```
schustmi:
```suggestion
        description="Stop the Modal app after the sandbox reaches a terminal "
        "state.",
```

schustmi (changes requested):

raza-khan0108:
I've updated the PR to address all your comments:

1. **Setting Moved & Renamed**: Moved `stop_app_after_run` from `ModalSettingsMixin` to `ModalStepOperatorSettings` and renamed it to `stop_app` with the suggested concise description.
2. **Helper Function for App Names**: Extracted `sandbox_utils.get_modal_app_name(...)` to deduplicate app name formatting across `submit()` and `cleanup_step_submission()`.
3. **Cleaned up Imports & Docstrings**: Moved `subprocess` import to the top of `sandbox_utils.py` and trimmed docstrings per suggestions.
4. **Environment Metadata Handling**: `cleanup_step_submission()` now checks step run metadata for `STEP_MODAL_ENVIRONMENT_METADATA_KEY` first to ensure cleanup targets the exact Modal environment used at submission time.
5. **Robust CLI Execution**: Added a 30s timeout to `stop_modal_app()` and added fallback handling for older Modal CLI versions (< 1.4.2) that don't support `--yes`.
6. **Docs & Tests**: Updated `docs/book/component-guide/step-operators/modal.md` to document `stop_app` and added unit tests covering `get_modal_app_name` and `cleanup_step_submission()`.

PR #4994, opened 2026-06-25
More query improvements

schustmi:
Follow up to #4982, this PR adds more query option and index improvements.

- Only load required columns during DAG query
- `selectinload` run metadata for all entities
- add missing `selectinload`s for artifact version and model version
- use `joinedload` instead of `selectinload` for `get_step_run` and `get_pipeline_run` for certain 1:1 relationships
- fix: release connection during alembic migration


bcdurak (approved):
Good changes, really looking forward to releasing all the improvements.

PR #4995, opened 2026-06-26
Dynamic pipeline start ordering

schustmi:
This PR adds an option to specify `start_after=...` when calling steps in dynamic pipelines. This allows specifying which steps should start before/after other steps when running them concurrently.

Breaking change:
- The `start_after` keyword is now a reserved keyword for steps. Any pipelines containing steps with a keyword argument with this name will fail from this point on.

### DIFF
--- a/docs/book/how-to/steps-pipelines/dynamic_pipelines.md
+++ b/docs/book/how-to/steps-pipelines/dynamic_pipelines.md
@@ -321,6 +321,28 @@ The `StepFuture` object provides several methods:
 When using `step.submit()`, steps with `runtime="isolated"` will execute in separate containers/processes, while steps with `runtime="inline"` will execute in separate threads within the orchestration environment.
 {% endhint %}
 
+### Ordering submitted steps
+
+A submitted step starts as soon as its inputs are available. To impose an order between steps that have no data dependency, use `after` or `start_after`:
+
+- `after=` waits for the upstream step to **finish** before starting.
+- `start_after=` waits for the upstream step to **start** before starting.
+
+`start_after` is useful when an upstream step is long-running and you want a dependent to run alongside it rather than after it. A common case is a step that brings up a service and a second step that uses it:
+
+```python
+@pipeline(dynamic=True)
+def serve_and_query():
+    server = serve_model.submit()  # long-running
+    # Starts once the server step is running, not when it finishes.
+    query = run_inference.submit(start_after=server)
+    query.wait()
+```
+
+Using `after=server` here would deadlock, since the dependent would wait for the long-running server step to finish. Both parameters accept a single future or a list, and you can combine them: `run_inference.submit(after=preprocess, start_after=server)`. `start_after` is available on `step.submit(...)`, `step.map(...)`, `step.product(...)`, and on a direct synchronous `step(...)` call, where the entrypoint blocks until the upstream has started. The upstream can be another step or a submitted child pipeline, for example `run_inference(start_after=serve_pipeline.submit())`.
+
+`start_after` orders execution, it does not probe readiness. The dependent starts once the upstream step has launched (for isolated steps, once it is submitted to the infrastructure), which does not guarantee that whatever the upstream sets up is ready to serve. Add your own connection retries if the dependent needs to reach a service the upstream starts. A failed upstream counts as started, so a `start_after` dependent is released rather than blocked when the upstream fails. Circular `start_after` dependencies are not detected and will stall the involved steps.
+
 ### Child pipelines inside dynamic pipelines
 
 Dynamic pipelines can call other dynamic pipelines from their `@pipeline`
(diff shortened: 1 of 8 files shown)


docs/book/how-to/steps-pipelines/dynamic_pipelines.md
```diff
@@ -321,6 +321,28 @@ The `StepFuture` object provides several methods:
 When using `step.submit()`, steps with `runtime="isolated"` will execute in separate containers/processes, while steps with `runtime="inline"` will execute in separate threads within the orchestration environment.
 {% endhint %}
 
+### Ordering submitted steps
+
+A submitted step starts as soon as its inputs are available. To impose an order between steps that have no data dependency, use `after` or `start_after`:
+
+- `after=` waits for the upstream step to **finish** before starting.
+- `start_after=` waits for the upstream step to **start** before starting.
```
bcdurak, line 329: I like the feature however, intuitively `after` and `start_after` sound like they would do the exact same thing, but I can not think of a better alternative. Perhaps `after_start`?

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
```diff
@@ -321,6 +321,28 @@ The `StepFuture` object provides several methods:
...
+```python
+@pipeline(dynamic=True)
+def serve_and_query():
+    server = serve_model.submit()  # long-running
+    # Starts once the server step is running, not when it finishes.
+    query = run_inference.submit(start_after=server)
+    query.wait()
+```
+
+Using `after=server` here would deadlock, since the dependent would wait for the long-running server step to finish. Both parameters accept a single future or a list, and you can combine them: `run_inference.submit(after=preprocess, start_after=server)`. `start_after` is available on `step.submit(...)`, `step.map(...)`, `step.product(...)`, and on a direct synchronous `step(...)` call, where the entrypoint blocks until the upstream has started. The upstream can be another step or a submitted child pipeline, for example `run_inference(start_after=serve_pipeline.submit())`.
```
bcdurak, line 342: Should we categorize this as a breaking change?

Within any pipeline that had a step with a parameter called `start_after` will probably fail now. For instance, this works with the latest release:

```python

@step
def print_message(start_after: str) -> None:
    print(start_after)

@pipeline
def my_pipeline() -> None:
    print_message(start_after="hello")

if __name__ == "__main__":
    my_pipeline()
```

However, it fails with a `RuntimeError: start_after is only supported in dynamic pipelines.` now.

bcdurak: A similar issue affects the dynamic pipelines as well.

bcdurak (commented):
Hey @schustmi, just gonna leave a few comments here as I found a potentially breaking change. I would just like to briefly discuss how we should label or handle it. Otherwise, I have tested the changes and they look good.

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
schustmi: I thought about it, but seems so incredibly rare that I though I'd ignore it. If you want, i can mark it as a breaking change though.

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
schustmi: `after_start` seems more confusing to me tbh :D

`step(start_after=another_step())` I thought this made sense to me tbh

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
bcdurak: I think that would be the fair solution here. Potentially, we can still let the static pipelines function as they are functioning right now but I don't like that too much. IMO, it is completely fine if we leave things as is and just mark it breaking.

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
schustmi: Sure it's still a little unclear what `after` is doing, but `start_after` actually seems pretty clear?

bcdurak: Yeah, it just sounds like they would do the same thing 😄 As I mentioned, I don't have a better alternative. At some point, `on_start` came to mind but we reach over to the hook territory there 😄

bcdurak: We can leave it as is. I can't think of a better alternative if you can't as well.

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
schustmi: Marked as breaking

bcdurak (approved):

PR #5003, opened 2026-07-01
Async snapshot execution

Json-Andriopoulos:
## Describe changes

Enables support for async snapshot execution on the ZenML server. When the feature is enabled, the ZenML server queues the snapshot for execution (async execution) instead of synchronously handling it server side.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/config/server_config.py
+++ b/src/zenml/config/server_config.py
@@ -334,11 +334,7 @@ class ServerConfiguration(BaseModel):
     workload_manager_implementation_source: Optional[str] = None
     resource_pool_implementation_source: Optional[str] = None
     stream_broker_implementation_source: Optional[str] = None
-    # TODO: Configure this through workspace provisioning before merging. This
-    # temporary default intentionally requires the Pro plugin during development.
-    snapshot_run_dispatcher_implementation_source: Optional[str] = (
-        "zenml_cloud_plugins.snapshot_run_dispatcher.RedisSnapshotRunDispatcher"
-    )
+    snapshot_run_dispatcher_implementation_source: Optional[str] = None
     streaming_heartbeat_seconds: float = Field(default=30.0, gt=0.0)
     streaming_max_subscribers_per_stream: int = Field(default=100, gt=0)
     streaming_broadcaster_idle_grace_seconds: float = Field(
--- a/src/zenml/config/server_config.py
+++ b/src/zenml/config/server_config.py
@@ -166,6 +166,8 @@ class ServerConfiguration(BaseModel):
             server.
         workload_manager_implementation_source: Source pointing to a class
             implementing the workload management interface.
+        snapshot_run_dispatcher_implementation_source: Source pointing to a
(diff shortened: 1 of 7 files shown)


src/zenml/config/server_config.py
```diff
@@ -332,6 +334,11 @@ class ServerConfiguration(BaseModel):
     workload_manager_implementation_source: Optional[str] = None
     resource_pool_implementation_source: Optional[str] = None
     stream_broker_implementation_source: Optional[str] = None
+    # TODO: Configure this through workspace provisioning before merging. This
```
schustmi: I think we should set this to the local one maybe as a default for now?

schustmi: Or even leave it empty and just use local if nothign is set for it?

src/zenml/zen_server/pipeline_execution/snapshot_run_dispatcher.py
```diff
@@ -0,0 +1,80 @@
...
+
+from zenml.exceptions import MaxConcurrentTasksError
+
+
+class SnapshotRunExecutionRequest(BaseModel):
+    """Durable identifiers required to execute a prepared snapshot run."""
+
+    run_id: UUID
+    snapshot_id: UUID
+    source_snapshot_id: UUID
```
schustmi: Unless I'm missing something, this seems to be only used in one very defensive error check, which IMO could be removed entirely. This is a request to execute a snapshot, how the snapshot was built (e.g. from which source) shouldn't even matter anymore at this point IMO.

src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -309,146 +332,374 @@ def run_snapshot(
...
+        response_run = prepared.run
+    else:
+        response_run = zen_store().update_run(
+            run_id=prepared.run.id,
+            run_update=PipelineRunUpdate(
+                status_reason=SNAPSHOT_RUN_QUEUED_STATUS_REASON
+            ),
+        )
+        try:
+            snapshot_run_dispatcher().submit(prepared.request)
```
schustmi: When using the local dispatcher, we throw away and refetch all the models (run, snapshot, ...) here.

IMO, we should either:
- pass them along, and not refetch. maybe even as part of the request object as optional attributes, which we drop when putting it on redis?
- make the sync case above also do the same. this would reduce the code complexity a lot IMO, as `prepare_snapshot_run` would just return the request, and we wouldn't need three methods to execute snapshot runs (_execute_prepared_snapshot_run, _execute_snapshot_run, execute_snapshot_run)

schustmi (changes requested):

src/zenml/config/server_config.py
Json-Andriopoulos: Yeap, that was a temporary change for it to be deployable without control plane changes. The final behaviour will be: either load from source or default to the local in-memory version.

schustmi (approved):

PR #5005, opened 2026-07-01
Fix configuration of docker credentials for building/pushing container images

stefannica:
## Describe changes

Fixes https://github.com/zenml-io/zenml-jetbrains/issues/50

This PR addresses a couple of things related to how zenml configures credentials for Docker as part of the process of building and pushing container images:

1. configuring credentials for the docker python client was done incorrectly. The URL extracted from the container registry used in combination with username/password credentials was never actually used by the python client because of a registry URI mismatch. For example:

* the registry URI expected by the docker python client (hostname only, similar to the one used by the local docker CLI/SDK): ` europe-west3-docker.pkg.dev`
* the actual URI configured for the python client via `self.client.login` in the `DockerContainerEngine` class contained the entire repository path: `europe-west3-docker.pkg.dev/zenml-core/zenml-f0a1916fb0ad/zenml:super_simple_pipeline-orchestrator`

This caused the docker python client to completely ignore the configured credentials and rely on those lifted from the local docker SDK configuration. In the past, this lead to all sorts of problems involving expired, missing or incorrect credentials, which is why we decided a long time ago to always configure the local docker CLI/SDK credentials at the same time as the python client (see point 2).

2. the local docker CLI/SDK credentials are always configured in addition to those for the python client. This can overwrite the credentials that the user stored there previously. It is also completely unnecessary unless the "use_subprocess" option is used in the local image builder.

The solutions to these issues are:

1. using the proper hostname when configuring credentials for the docker python client.
2. stop configuring credentials for the local docker CLI/SDK unless absolutely necessary (i.e. only when the "use_subprocess" option is set).

## Side changes

* fix the zenml development install script to include the `docstring_parser_fork` fix even if no integrations are installed
* remove unused `npm install html-minifier` from github actions to address zizmor vulnerability

## Pre-requisites

## Types of changes

### DIFF
--- a/.github/workflows/publish_api_docs.yml
+++ b/.github/workflows/publish_api_docs.yml
@@ -21,11 +21,6 @@ jobs:
         run: |
           BRANCH="${GITHUB_REF_NAME}"
           echo ::set-output name=VERSION::${BRANCH#release/}
-      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6.4.0
-        with:
-          node-version: '14'
-      - run: npm install
-      - run: npm install html-minifier -g
       - name: Set up Python 3.11
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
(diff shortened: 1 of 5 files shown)


strickvl (commented):
The deletion for zizmor makes sense to me. I think that html-minifier stuff was left over. It doesn't appear to be called in the `scripts/generate-docs.sh` either. I didn't look at any of the rest of the code, though.

schustmi (approved):

PR #5006, opened 2026-07-02
Changed default otel-service-name as workspace name for zenml-pro workspaces

amitvikramraj:


src/zenml/constants.py
```diff
@@ -24,6 +24,7 @@
 T = TypeVar("T")
 
 DEFAULT_ZENML_SERVER_OTEL_SERVICE_NAME = "zenml-server"
+"""Default OTEL_SERVICE_NAME for non-Pro ZenML Workspace."""
```
bcdurak:
```suggestion
# Default OTEL_SERVICE_NAME for non-Pro ZenML Workspace.
DEFAULT_ZENML_SERVER_OTEL_SERVICE_NAME = "zenml-server"
```

src/zenml/config/server_config.py
```diff
@@ -784,10 +785,19 @@ def get_server_config(cls) -> "ServerConfiguration":
                     workspace_id=str(server_pro_config.workspace_id),
                 )
             )
+            if server_pro_config.organization_name:
+                server_config.metadata.update(
```
Json-Andriopoulos: I don't get this pattern. It creates a dict, then updates the existing dict with one key-value. I think it unnecessarily obfuscates the code. Why not do this instead?

```
server_config.metadata["workspace_name"] = "..."
```

amitvikramraj: I have no idea. I was following the existing pattern.😄 do you want me to update this?

src/zenml/constants.py
amitvikramraj: this addes a docsting to the constant name. You can hover over it in your IDE and see the constant definition which can be useful if someone is importing this value. But yeah it's fine, adding it as a comment is also fine.

schustmi: I think neither are necessary, the name is more than self-explanatory :D

amitvikramraj: I added this because the default has now two behaviour – for PRO workspaces the default name is the workspace name that we get from config. For non-pro workspace, the default is basically this constant value

src/zenml/config/server_config.py
amitvikramraj: okay I addressed this

src/zenml/constants.py
schustmi: I'd consider it differently: This is the default otel service name. ZenML Pro servers overwrite it. Users might overwrite it. It's still the default, if no one else overrides it.

Json-Andriopoulos (approved):

PR #5011, opened 2026-07-02
Add dependency audit release waivers

strickvl:
## What this changes

Adds explicit, time-limited advisory waivers to the dependency audit workflow so release PRs can proceed when a blocking vulnerability is acknowledged but cannot safely be fixed during release prep.

Concretely:
- adds `.github/dependency-audit-waivers.json` with temporary Starlette advisory waivers expiring on `2026-07-31`
- teaches `scripts/dependency_audit_severity_gate.py` to classify matching findings as `waived_blocking` instead of `blocking`
- applies waivers only in the `release-pr` context
- sends a Discord notification only when a release PR actually uses waived findings
- reruns the workflow when the `release` label is added or removed, so waiver eligibility is recalculated

## Why

The current audit failure is caused by Starlette advisories whose fixes require a FastAPI range bump. That bump is risky enough for ZenML server users that it should happen in a dedicated, tested dependency PR rather than during release prep.

This keeps normal PRs strict while giving release PRs an explicit, expiring exception with visibility.

## Verification

- `uv run --extra dev --extra local --extra server pytest tests/unit/scripts/test_dependency_audit_severity_gate.py --no-provision -q`
- `uv run --extra dev --extra local --extra server bash scripts/format.sh`
- Smoke checked against downloaded release audit report from `/Users/strickvl/Desktop/dependency-audit-reports/pip-audit.json`:
  - normal context: `blocking=2`, `waived_blocking=0`
  - `release-pr` context: `blocking=0`, `waived_blocking=2`

### DIFF
--- a/.github/dependency-audit-waivers.json
+++ b/.github/dependency-audit-waivers.json
@@ -1,0 +1,18 @@
+{
+  "waivers": [
+    {
+      "package": "starlette",
+      "advisories": [
+        "CVE-2026-48818",
+        "GHSA-wqp7-x3pw-xc5r",
+        "PYSEC-2026-249",
+        "CVE-2026-54283",
+        "GHSA-82w8-qh3p-5jfq"
+      ],
+      "allowed_contexts": ["release-pr"],
+      "expires_on": "2026-07-31",
+      "owner": "server-team",
+      "reason": "Starlette fixes require a FastAPI range bump. FastAPI upgrades affect ZenML server compatibility and need a dedicated test plan outside release prep."
+    }
+  ]
+}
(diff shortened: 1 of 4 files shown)


schustmi (approved):
Approving, but if this is just for the release process, I think I can also just ignore it. At least it surfaces those issues during the release process, might be a good reminder for me to make sure we fix it in the next release even?

strickvl:
> Approving, but if this is just for the release process, I think I can also just ignore it. At least it surfaces those issues during the release process, might be a good reminder for me to make sure we fix it in the next release even?

Yeah I'm happy to close this if it means we are more likely to fix these :)

schustmi:
I don't want to make any promises, but I think it would at least marginally increase the probability of us fixing these :)

strickvl:
Closing this for the benefit of all humanity

PR #5023, opened 2026-07-02
Fix the in-memory docker credentials properly

stefannica:
## Describe changes
The docker python library is hard-coded to give precedence to credentials extracted from the locally configured credentials store, when one is found, even when the DockerClient is explicitly given new credentials for an existing registry via `client.login()` calls.

The order of priority that the docker python library uses to load credentials is the following:

* lowest prio: the "auths" dictionary that stores credentials supplied at runtime with `client.login()` (i.e. what zenml uses)
* next: the credentials store configured by the user
* highest prio: the credential "helpers" configured by the user

To properly fix this issue without dealing too much with the docker library internals, this PR simulates an in-memory "credentials helper" that merely serves the `client.login()` credentials again, only with the highest priority. This is done in three steps:

1. defining a custom credentials helper implementations that serves credentials from this "auths" dictionary that stores  `client.login()` credentials:

```
class _AuthsCredentialStore:

    def __init__(self, auths: Dict[str, Dict[str, Any]]) -> None:
        self._auths = auths

    def get(self, server: str) -> Dict[str, str]:
        return self._auths.get(server) // more or less

    def list(self) -> Dict[str, str]:
        return self._auths // more or less
```

2. configure the store as a helper when the docker client is initialized:

```
    @property
    def client(self) -> DockerClient:
            self._client = DockerClient.from_env()
            self._client.api._auth_configs_stores[_ZENML_CREDENTIAL_STORE] = (
                _AuthsCredentialStore(auth_config.auths)
            ) // more or less
```

3. every time `client.login()` is called for a server, we tie it to the credentials helper we just created:

```
client.login(
                username=username,
                password=password,
                registry=docker_registry,
                reauth=True,
            )
client.api._auth_configs.cred_helpers[docker_registry] = _ZENML_CREDENTIAL_STORE
```

## Pre-requisites

## Types of changes


src/zenml/container_engines/docker_engine.py
```diff
@@ -122,6 +183,20 @@ def client(self) -> DockerClient:
                 "Could not create a Docker client from the environment. Is your "
                 "Docker daemon running?"
             ) from e
+
+        # Install a fake credential store backed by docker-py's in-memory
```
schustmi:
```suggestion
        # Install a fake credential helper backed by docker-py's in-memory
```

schustmi (approved):

stefannica:
@schustmi I added some guardrails to catch errors that may occur if the credentials helper "hack" I implemented fails due to changes in the docker python library internals. It will fall back to using the CLI to log in, so everything should keep running just as before. Please re-review.

schustmi (approved):

PR #5025, opened 2026-07-03
Expire tokens after password changes and API key rotations

stefannica:
## Describe changes
This fixes stale session reuse after user passwords are changed or API keys are rotated.

The change adds token revocation state without tracking every issued token individually:

- User accounts now store `password_changed_at`.
- JWTs now include an `iat` claim with subsecond precision.
- User tokens issued before `password_changed_at` are rejected.
- API keys now track a `key_generation`.
- JWTs derived from API keys include a compact `key_gen` claim.
- API-key tokens are accepted only for the current key generation, or the immediately previous generation while the configured retain period is still active.
- Legacy tokens without `iat` / `key_gen` remain valid.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_stores/schemas/user_schemas.py
+++ b/src/zenml/zen_stores/schemas/user_schemas.py
@@ -231,8 +231,7 @@ class UserSchema(NamedSchema, table=True):
 
             if field == "password":
                 setattr(self, field, user_update.create_hashed_password())
-                if value is not None:
-                    self.password_changed_at = utc_now()
+                self.password_changed_at = utc_now()
             elif field == "activation_token":
                 setattr(
                     self, field, user_update.create_hashed_activation_token()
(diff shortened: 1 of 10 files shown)


src/zenml/zen_stores/schemas/user_schemas.py
```diff
@@ -226,6 +231,8 @@ def update_user(self, user_update: UserUpdate) -> "UserSchema":
 
             if field == "password":
                 setattr(self, field, user_update.create_hashed_password())
+                if value is not None:
+                    self.password_changed_at = utc_now()
```
schustmi: If `value is None`, what would happen here? We would set `self.password` to `None`, right? I didn't check what this does internally, but seems like it still changed the password and should invalidate old tokens?

schustmi (changes requested):

src/zenml/zen_stores/schemas/user_schemas.py
stefannica: You're right, the condition is not necessary here.

stefannica: Done

schustmi (approved):

PR #5027, opened 2026-07-03
Remove placeholder data on queue full errors

### LINKED ISSUE #5026: Missing local snapshot dispatcher run cleanup

_No response_

### System Information

_No response_

### What happened?

On a workspace without the local snapshot dispatcher, when I try to trigger more snapshot than the dispatcher concurrency limit (2 by default) using the `Client().trigger_pipeline(...)`, the following happens:
- The first two get accepted and succeed
- The third call of the endpoint creates a run, and then runs into the concurrency limit. It changes the run status to `Failed` with a status reason of `Snapshot execution queue is full.`
- The endpoint returns a 409 response
- The Client retries the request after some time
- The same Client call now creates multiple runs in a failed state

### Reproduction steps

- ZenML Pro workspace
- 1 replica only
- Trigger more than two snapshots at the same time

### Relevant log output

```shell

```

### Code of Conduct

Json-Andriopoulos:
## Describe changes

Deletes placeholder on queue full error. This handles the scenario of generating multiple retried failed runs when the queue is full.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/zen_server/pipeline_execution/utils.py
+++ b/src/zenml/zen_server/pipeline_execution/utils.py
@@ -352,13 +352,18 @@ def run_snapshot(
         try:
             snapshot_run_dispatcher().submit(execution_request)
         except SnapshotRunQueueFullError:
-            zen_store().update_run(
-                run_id=execution_request.run_id,
-                run_update=PipelineRunUpdate(
-                    status=ExecutionStatus.FAILED,
-                    status_reason="Snapshot execution queue is full.",
-                ),
-            )
+            try:
+                zen_store().delete_run(run_id=execution_request.run_id)
+                if create_new_snapshot:
+                    zen_store().delete_snapshot(
+                        snapshot_id=execution_request.snapshot_id
+                    )
+            except Exception:
+                logger.exception(
+                    "Failed to clean up snapshot run %s after queue "
+                    "rejection.",
+                    execution_request.run_id,
+                )
             raise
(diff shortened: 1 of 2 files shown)


src/zenml/zen_server/pipeline_execution/utils.py
```diff
@@ -352,13 +352,18 @@ def run_snapshot(
...
             snapshot_run_dispatcher().submit(execution_request)
         except SnapshotRunQueueFullError:
-            zen_store().update_run(
-                run_id=execution_request.run_id,
-                run_update=PipelineRunUpdate(
-                    status=ExecutionStatus.FAILED,
-                    status_reason="Snapshot execution queue is full.",
-                ),
-            )
+            try:
```
schustmi, line 355: do you think it makes sense to even fail early using some helper method on the dispatcher that checks whether it can dispatch, to avoid creating the snapshot/run? Or not worth the code complexity for the little benefit?

Json-Andriopoulos: Hmm I'd say let's get the fix out and we can improve it with an early check, yes. Checking queue availability should be significantly lighter/faster that creating the run in both scenarios though I don't expect queue full errors to happen that often for it to have a server impact.

schustmi: Makes sense

schustmi (approved):

PR #5034, opened 2026-07-06
Support deleting artifact data from server API

Json-Andriopoulos:
## Describe changes

Supports server-side artifact version data deletion. Prior to this feature users were able to delete artifact store data from the Client. Deletion operation is dual: 1) Metadata is deleted from the server 2) Data is deleted client-side using the stack's artifact store.

Supporting data deletion server-side unblocks the following:

1) Deleting artifact version metadata & data from the UI
2) Deleting artifact version metadata & data from a thin client without the stack's dependencies.

## Pre-requisites

## Types of changes

### DIFF
--- a/src/zenml/cli/artifact.py
+++ b/src/zenml/cli/artifact.py
@@ -252,6 +252,62 @@ def update_artifact_version(
         cli_utils.declare(f"Artifact version '{artifact_version.id}' updated.")
 
 
+@version.command("delete", help="Delete an artifact version by ID.")
+@click.argument("artifact_version_id")
+@click.option(
+    "--delete-data",
+    is_flag=True,
+    help="Also delete the artifact data from the artifact store.",
+)
+@click.option(
+    "--server-side",
+    is_flag=True,
+    help="Delete artifact data from the server instead of the local stack.",
+)
+@click.option(
+    "--yes",
+    "-y",
+    is_flag=True,
+    help="Don't ask for confirmation.",
+)
+def delete_artifact_version(
+    artifact_version_id: str,
+    delete_data: bool = False,
+    server_side: bool = False,
+    yes: bool = False,
+) -> None:
+    """Delete an artifact version by ID.
+
+    Args:
+        artifact_version_id: The ID of the artifact version to delete.
+        delete_data: If set, also delete the artifact data from the artifact
+            store.
+        server_side: If set, delete artifact data from the server instead of
+            the local stack.
+        yes: If set, don't ask for confirmation.
+    """
+    if not yes:
+        confirmation = cli_utils.confirmation(
+            f"Are you sure you want to delete artifact version "
+            f"'{artifact_version_id}'?"
+        )
+        if not confirmation:
+            cli_utils.declare("Artifact version deletion canceled.")
+            return
+
+    try:
+        Client().delete_artifact_version(
+            name_id_or_prefix=artifact_version_id,
+            delete_metadata=True,
+            delete_from_artifact_store=delete_data,
+            server_side=server_side,
+        )
+    except Exception as e:
+        cli_utils.exception(e)
+    else:
+        cli_utils.declare(f"Artifact version '{artifact_version_id}' deleted.")
+
+
 @artifact.command(
     "prune",
     help=(
--- a/src/zenml/cli/artifact.py
+++ b/src/zenml/cli/artifact.py
@@ -289,11 +289,17 @@ def update_artifact_version(
     is_flag=True,
     help="Ignore errors and continue with the next artifact version.",
 )
+@click.option(
+    "--server-side",
+    is_flag=True,
+    help="Delete artifact data from the server instead of the local stack.",
+)
 def prune_artifacts(
     only_artifact: bool = False,
     only_metadata: bool = False,
     yes: bool = False,
     ignore_errors: bool = False,
+    server_side: bool = False,
 ) -> None:
     """Delete all unused artifacts and artifact versions.
 
@@ -309,6 +315,8 @@ def prune_artifacts(
         yes: If set, don't ask for confirmation.
         ignore_errors: If set, ignore errors and continue with the next
             artifact version.
+        server_side: If set, delete artifact data from the server instead of
+            the local stack.
     """
     client = Client()
     unused_artifact_versions = depaginate(
@@ -334,6 +342,7 @@ def prune_artifacts(
                 name_id_or_prefix=unused_artifact_version.id,
                 delete_metadata=not only_artifact,
                 delete_from_artifact_store=not only_metadata,
+                server_side=server_side,
             )
             unused_artifact = unused_artifact_version.artifact
             if not unused_artifact.versions and not only_artifact:
--- a/src/zenml/cli/artifact.py
+++ b/src/zenml/cli/artifact.py
@@ -262,7 +262,7 @@ def update_artifact_version(
 @click.option(
     "--server-side",
     is_flag=True,
-    help="Delete artifact data from the server instead of the local stack.",
+    help="Flag to decide whether the deletion should happen client-side or server-side.",
 )
 @click.option(
     "--yes",
@@ -282,8 +282,7 @@ def delete_artifact_version(
         artifact_version_id: The ID of the artifact version to delete.
         delete_data: If set, also delete the artifact data from the artifact
             store.
-        server_side: If set, delete artifact data from the server instead of
-            the local stack.
+        server_side: If set, artifact deletion will happen on the server side.
         yes: If set, don't ask for confirmation.
     """
     if not yes:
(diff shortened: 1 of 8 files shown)


src/zenml/zen_server/routers/artifact_version_endpoints.py
```diff
@@ -233,18 +235,44 @@ def update_artifact_version(
...
-        id=artifact_version_id,
-        get_method=zen_store().get_artifact_version,
-        delete_method=zen_store().delete_artifact_version,
+    artifact_version = zen_store().get_artifact_version(
+        artifact_version_id, hydrate=True
     )
+    verify_permission_for_model(artifact_version, action=Action.DELETE)
+
+    if delete_metadata:
+        zen_store().delete_artifact_version(artifact_version.id)
```
bcdurak: I didn't check the other parts of the PR just yet, but wanted to give an early feedback here. I think if someone wants to delete an artifact version from the artifact store they should need access to the artifact store (and its connector if it has one).

bcdurak (changes requested):

src/zenml/cli/artifact.py
```diff
@@ -252,6 +252,62 @@ def update_artifact_version(
...
+@click.argument("artifact_version_id")
+@click.option(
+    "--delete-data",
+    is_flag=True,
+    help="Also delete the artifact data from the artifact store.",
+)
+@click.option(
+    "--server-side",
+    is_flag=True,
+    help="Delete artifact data from the server instead of the local stack.",
```
bcdurak: The help message and the docstring below sound a bit misleading IMO.

bcdurak: Same in the prune method as well.

src/zenml/zen_server/routers/artifact_version_endpoints.py
```diff
@@ -233,18 +235,58 @@ def update_artifact_version(
...
+            artifact_version.artifact_store_id, hydrate=True
+        )
+        verify_permission_for_model(
+            artifact_store_model,
+            action=Action.READ,
+        )
+        if artifact_store_model.connector:
+            verify_permission_for_model(
+                artifact_store_model.connector,
+                action=Action.READ,
```
bcdurak, line 289: I am not fully sure about this but shouldn't we check for the `Action.CLIENT` here?

src/zenml/zen_server/routers/artifact_version_endpoints.py
```diff
@@ -233,18 +235,58 @@ def update_artifact_version(
...
+            )
+
+    if delete_metadata:
+        zen_store().delete_artifact_version(artifact_version.id)
+        delete_model_resource(artifact_version)
+
+    if delete_from_artifact_store and artifact_version.artifact_store_id:
```
bcdurak: Like in the client, I think this step needs to happen first. We should keep the artifact in the DB if the deletion fails.

src/zenml/zen_stores/zen_store_interface.py
```diff
@@ -670,11 +667,19 @@ def update_artifact_version(
         """
 
     @abstractmethod
-    def delete_artifact_version(self, artifact_version_id: UUID) -> None:
+    def delete_artifact_version(
+        self,
+        artifact_version_id: UUID,
+        delete_metadata: bool = True,
+        delete_from_artifact_store: bool = False,
```
bcdurak: The deletion from the artifact store happens either in the client or in the endpoint. The zen store interface should not be affected by these parameters IMO.

bcdurak: `delete_from_artifact_store` only gets rejected from the SQLZenStore anyhow.

src/zenml/cli/artifact.py
```diff
@@ -252,6 +252,62 @@ def update_artifact_version(
...
 
+@version.command("delete", help="Delete an artifact version by ID.")
+@click.argument("artifact_version_id")
+@click.option(
+    "--delete-data",
+    is_flag=True,
+    help="Also delete the artifact data from the artifact store.",
+)
+@click.option(
+    "--server-side",
```
bcdurak, line 263: If people set `--server-side` without `--delete-data`, it will silently be ignored.

src/zenml/zen_server/routers/artifact_version_endpoints.py
bcdurak: Maybe even both?

bcdurak (changes requested):
Happy to discuss the comments. Hope they are helpful.

src/zenml/cli/artifact.py
Json-Andriopoulos: What would the alternative be? Metadata deletion is always server side. I don't want to raise warning or even worse a validation for something that is just a redundant flag 🤔

src/zenml/zen_server/routers/artifact_version_endpoints.py
Json-Andriopoulos: I don't agree or we need to change our existing patterns. Any stack component operation I checked that is server-side basically relies (directly or implicitly) on Action.READ.

src/zenml/cli/artifact.py
Json-Andriopoulos: Why is it misleading? I am not sure what a good alternative would be. Maybe I can rephrase it as "When set, executes data deletion server-side instead of relying on local stack"? I am open to suggestions

src/zenml/zen_server/routers/artifact_version_endpoints.py
bcdurak: I am not 100% sure myself, but it might be a nice indication to change our existing patterns as well. Otherwise, why do we even have a `Action.CLIENT` check?

src/zenml/cli/artifact.py
bcdurak: When you mention a `local` stack, it makes me think of the default one, like a local artifact store and orchestrator. That aside, in both cases, you are deleting the data from the same stack anyhow. It's just about where the operation takes place. I would rephrase to something like: `Flag to decide whether the deletion should happen client-side or server-side`.

src/zenml/cli/artifact.py
bcdurak: I was thinking of adding a warning but I would leave it up to you.

src/zenml/zen_server/routers/artifact_version_endpoints.py
```diff
@@ -233,18 +235,63 @@ def update_artifact_version(
...
+                artifact_store_model.connector,
+                action=Action.READ,
+            )
+            verify_permission_for_model(
+                artifact_store_model.connector,
+                action=Action.CLIENT,
+            )
+
+
+    if delete_from_artifact_store and artifact_version.artifact_store_id:
```
bcdurak: Identical `if` from the block above.

src/zenml/client.py
```diff
@@ -6013,46 +6014,78 @@ def delete_artifact_version(
...
-            raise ValueError(
-                "The metadata of artifact versions that are used in runs "
-                "cannot be deleted. Please delete all runs that use this "
-                "artifact first."
+        if delete_metadata:
+            unused_versions = self.list_artifact_versions(
+                id=artifact_version.id, only_unused=True, size=1
+            )
+            if not unused_versions.items:
+                raise ValueError(
```
bcdurak, line 6064: We do this check-and-fail on the client side but not on the server side I think. Should we add there as well? WDYT?

bcdurak (commented):
A few final comments.

src/zenml/zen_server/routers/artifact_version_endpoints.py
```diff
@@ -233,19 +235,85 @@ def update_artifact_version(
...
+                artifact_store_model.connector,
+                action=Action.CLIENT,
+            )
+
+    # Deletion operations.
+
+    if delete_from_artifact_store:
+        try:
+            artifact_store = load_artifact_store(
+                artifact_store_id=artifact_version.artifact_store_id,
```
bcdurak, line 298: Linting issue here as `artifact_store_id` could be None.

bcdurak (changes requested):
Quick heads-up, there are some formatting and linting issues.

src/zenml/zen_server/routers/artifact_version_endpoints.py
```diff
@@ -233,19 +235,85 @@ def update_artifact_version(
...
+                action=Action.READ,
+            )
+            verify_permission_for_model(
+                artifact_store_model.connector,
+                action=Action.CLIENT,
+            )
+
+    # Deletion operations.
+
+    if delete_from_artifact_store:
```
bcdurak: The same duplicate `if` check is happening here as well.

bcdurak (approved):

PR #5036, opened 2026-07-06
Reduce unnecessary server requests

schustmi:
- Cache project when fetching via `ProjectScopedResponse`
- Cache RestZenStore store info unless explicitly requesting refresh
- Store stack response on stack object

Local orchestrator, `N` steps, no step output artifacts or tags/metadata, bare setup.
- Before `26 + 14N` requests
- After `18 + 6N` requests


bcdurak (approved):
Great changes all around.

PR #5038, opened 2026-07-07
In-process execution caching

schustmi:
This PR caches the pipeline run and completed step run responses in-process for the duration of a run, so pipelines that execute steps in one process (local orchestrator, inline steps in dynamic pipelines) make fewer server requests.

Results: See CI comment below


src/zenml/orchestrators/step_run_utils.py
```diff
@@ -213,7 +222,7 @@ def populate_request(
         input_artifacts = input_utils.resolve_step_inputs(
             step=step,
             pipeline_run=self.pipeline_run,
-            step_runs=step_runs,
+            step_runs=self._step_runs,
```
bcdurak, line 225: As far as I can see this is now using the step run within the execution context. The function `resolve_step_inputs` mention that `step_runs` will be updated in-place. However, if the cache is empty as we call this function, the changes will not be stored because the function starts with `step_runs = step_runs or {}`. I would rather do a `None` check there and initialize the context properly.

src/zenml/utils/dict_utils.py
```diff
@@ -15,10 +15,51 @@
...
+        Args:
+            *args: Positional arguments for the update.
+            **kwargs: Keyword arguments for the update.
+        """
+        super().update(*args, **kwargs)
+        self._evict()
+
+    def _evict(self) -> None:
+        """Evict the oldest entries beyond the maximum size."""
+        while len(self) > self._max_size:
```
bcdurak, line 66: Any negative `_max_size` would fail here. It would be nice to have an early validation for this.

bcdurak (commented):
I couldn't find any concrete cases why we would need to do this but still wanted to ask if we should put any sort of `lock`s around the context as it is a shared dict. Can you think of anything?

src/zenml/utils/dict_utils.py
schustmi: done

src/zenml/orchestrators/step_run_utils.py
schustmi: done

bcdurak (approved):
Good to go 👍

PR #5039, opened 2026-07-07
Support webhook integrations

### LINKED ISSUE #4796: Webhook triggers

_No response_

### Feature Description

Extend the trigger functionality with an additional Trigger type, Webhooks.

### Problem or Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Code of Conduct

Json-Andriopoulos:
## Describe changes

Introduces a new project-level component called webhook-integrations. Users can use the feature to create an integration point within ZenMl for events on extternal platforms (like Github).

The PR focuses on getting CRUD functionality in place (CLI, SDK, Rest, SQL). Integration with trigger code is out of scope and will be a follow-up branch.

## Pre-requisites

## Types of changes


htahir1:
@Json-Andriopoulos nice work on this - one request while the adapter design is still soft.

We have concrete customer demand (can't name names here) for exactly this flow with Firecrawl's monitoring webhooks: a monitor detects a page change, posts a `monitor.page` event, and that should trigger a pipeline snapshot. I just built an example around it on `example/firecrawl-web-monitor` (#5065) - right now it works around the missing trigger integration by pulling checks via the Firecrawl API, but webhook triggers are the architecture we actually want to sell there.

Two things in the current `BaseWebhookAdapter` would need overrides for Firecrawl, and they're probably representative of a lot of non-GitHub providers:

1. **Auth**: Firecrawl doesn't sign payloads with HMAC - it sends whatever custom headers you configure on the monitor (typically a shared bearer token). So `authenticate()` assuming an HMAC signature header shouldn't be baked into the base contract, just into the GitHub adapter.
2. **Event type location**: `parse()` requires the event type in a request header (`self.event_header`), but Firecrawl puts it in the JSON body (`"type": "monitor.page"`). An adapter should be able to derive `event_type` from the parsed payload.

Both are cheap to generalize now and annoying to retrofit once adapters exist downstream. Happy to contribute a `FirecrawlWebhookAdapter` as the first non-GitHub adapter once the trigger follow-up branch is ready - the example gives us a live end-to-end test bed for it.

src/zenml/zen_server/rbac/utils.py
```diff
@@ -522,6 +522,8 @@ def get_resource_type_for_model(
         TagRequest,
         TagResponse,
         TriggerRequest,
+        WebhookIntegrationRequest,
```
schustmi, line 525: There is also `_get_resource_type_schema_mapping` in this file which needs updating I think?

src/zenml/client.py
```diff
@@ -6832,6 +6840,194 @@ def delete_code_repository(
         )
         self.zen_store.delete_code_repository(code_repository_id=repo.id)
 
+    # ------------------------- Webhook integrations -------------------------
+
+    def create_webhook_integration(
```
schustmi: Should all of these only be enabled when connected to a rest store? We use a `_fail_for_sql_zen_store` decorator in some other client methods that don't make sense when not using a server.

src/zenml/client.py
```diff
@@ -6832,6 +6840,194 @@ def delete_code_repository(
...
+            name_id_or_prefix: The integration name, ID, or ID prefix.
+            name: The new integration name.
+            active: The new active state.
+            secret: A new direct signing secret or ZenML secret reference.
+            project: The project name or ID.
+
+        Returns:
+            The updated webhook integration.
+        """
+        integration = self.get_webhook_integration(
```
schustmi: Could maybe short-circuit here if it's already an ID, to avoid one request?

src/zenml/cli/webhook_integration.py
```diff
@@ -0,0 +1,207 @@
+# Copyright (c) ZenML GmbH 2026. All Rights Reserved.
```
schustmi: Missing license (also in some other files)

src/zenml/cli/webhook_integration.py
```diff
@@ -0,0 +1,207 @@
...
+from zenml.cli.utils import OutputFormat, list_options
+from zenml.client import Client
+from zenml.console import console
+from zenml.enums import CliCategories, WebhookType
+from zenml.exceptions import IllegalOperationError
+from zenml.models import WebhookIntegrationFilter, WebhookIntegrationResponse
+
+
+@cli.group(
+    "webhook-integration", cls=TagGroup, tag=CliCategories.MANAGEMENT_TOOLS
```
schustmi: maybe `webhook` to make it nicer to use?

schustmi: I think this maybe applies in most places, is there any reason not to use `webhook` directly but `webhook integration`?

src/zenml/cli/webhook_integration.py
```diff
@@ -0,0 +1,207 @@
...
+    """Add the project name to a webhook integration list row.
+
+    Args:
+        integration: The webhook integration to format.
+        _output_format: The requested output format.
+
+    Returns:
+        Additional row values for the webhook integration.
+    """
+    return {"project": integration.project.name}
```
schustmi: Any reason this is necessary? Isn't it the same for every row?

src/zenml/models/v2/core/webhook_integration.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+class WebhookIntegrationCreateResponse(BaseZenModel):
+    """Creation result with a generated secret when applicable."""
+
+    integration: WebhookIntegrationResponse
+    secret: PlainSerializedSecretStr | None = None
+
+
+class WebhookIntegrationSecretRequest(BaseZenModel):
```
schustmi: Maybe `WebhookIntegrationRotateSecretRequest`?

src/zenml/webhooks/adapters.py
```diff
@@ -0,0 +1,270 @@
...
+from abc import ABC, abstractmethod
+from collections.abc import Mapping
+from typing import Any
+
+from pydantic import BaseModel
+
+from zenml.enums import WebhookType
+
+
+class WebhookAuthenticationError(ValueError):
```
schustmi: Haven't gotten to the endpoints yet, but our endpoint utilities convert any `ValueError` or subclasses to a `400` response status. Probably we want a 301 in this case though?

src/zenml/zen_stores/schemas/webhook_integration_schemas.py
```diff
@@ -0,0 +1,181 @@
...
+    user: UserSchema | None = Relationship()
+    secret_id: UUID = build_foreign_key_field(
+        source=__tablename__,
+        target=SecretSchema.__tablename__,
+        source_column="secret_id",
+        target_column="id",
+        ondelete="CASCADE",
+        nullable=False,
+    )
+    webhook_type: str = Field(index=True)
```
schustmi: We don't usually use `index=True` on fields, as the index name will then depend on the DB flavor and future migrations that want to edit/delete it will be more painful to write. Instead, declaring the index in `__table_args__` with an explicit name is preferred.

src/zenml/zen_stores/schemas/webhook_integration_schemas.py
```diff
@@ -0,0 +1,181 @@
...
+        ondelete="SET NULL",
+        nullable=True,
+    )
+    user: UserSchema | None = Relationship()
+    secret_id: UUID = build_foreign_key_field(
+        source=__tablename__,
+        target=SecretSchema.__tablename__,
+        source_column="secret_id",
+        target_column="id",
+        ondelete="CASCADE",
```
schustmi: Do we really absolutely need to cascade if the secret gets deleted. That could happen by accident right, which would lose every trace of the webhook being used also in e.g. trigger executions?

Wouldn't setting it to inactive and failing when trying to active again be enough?

src/zenml/zen_stores/schemas/webhook_integration_schemas.py
```diff
@@ -0,0 +1,181 @@
...
+from zenml.zen_stores.schemas.user_schemas import UserSchema
+
+
+class WebhookIntegrationSchema(NamedSchema, table=True):
+    """SQL schema for a project-scoped webhook integration."""
+
+    __tablename__ = "webhook_integration"
+    __table_args__ = (
+        UniqueConstraint(
+            "name",
```
schustmi: If we flip around the constraint to `project_id, name` the project_id foreign key can reuse the index created by it. Or did you choose this order on purpose?

src/zenml/zen_server/routers/webhook_integration_endpoints.py
```diff
@@ -0,0 +1,332 @@
...
+
+    Args:
+        integration: The webhook integration creation request.
+
+    Returns:
+        The created integration and any generated signing secret.
+    """
+    verify_permission_for_model(model=integration, action=Action.CREATE)
+    _verify_webhook_secret_reference_access(integration.secret)
+    return zen_store().create_webhook_integration(integration)
```
schustmi: I'd call the `dehydrate_..` function here, just in case we decide to add some references to other models in the resources in the future.

src/zenml/webhooks/adapters.py
schustmi: This gets manually converted from what I can see, but maybe still better to inherit from one of our auth exceptions?

src/zenml/zen_server/routers/webhook_integration_endpoints.py
```diff
@@ -0,0 +1,332 @@
...
+    Args:
+        webhook_type: The provider type from the public endpoint path.
+        integration_id: The webhook integration ID.
+        request: The raw HTTP request.
+
+    Returns:
+        An empty accepted response.
+    """
+    body = await request.body()
+    return await run_in_threadpool(
```
schustmi, line 240: Any reason this endpoint is async and then runs this in a thread?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8111,6 +8130,371 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+        Returns:
+            The webhook integration.
+        """
+        with Session(self.engine) as session:
+            schema = self._get_schema_by_id(
+                resource_id=integration_id,
+                schema_class=WebhookIntegrationSchema,
+                session=session,
+            )
+            return schema.to_model(include_metadata=hydrate)
```
schustmi: `include_resource=True` missing

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8111,6 +8130,371 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+                )
+                self._update_secret_values(
+                    secret_id=schema.secret_id,
+                    values=values,
+                    overwrite=True,
+                )
+            schema.update(update)
+            session.add(schema)
+            session.commit()
+            return schema.to_model(include_metadata=True)
```
schustmi: include_resources=True

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8111,6 +8130,371 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+        Args:
+            integration_id: The webhook integration ID.
+        """
+        with Session(self.engine) as session:
+            schema = self._get_schema_by_id(
+                resource_id=integration_id,
+                schema_class=WebhookIntegrationSchema,
+                session=session,
+            )
+            self._delete_secret_schema(
```
schustmi: Here we just rely on the cascade behaviour, right? Wouldn't it be safer to also delete the webhook integration first, in case someone changes that in the future?

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8111,6 +8130,371 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+                    secret=update.secret.get_secret_value(), session=session
+                )
+                self._update_secret_values(
+                    secret_id=schema.secret_id,
+                    values=values,
+                    overwrite=True,
+                )
+            schema.update(update)
+            session.add(schema)
+            session.commit()
```
schustmi, line 8287: If the commit fails, shouldn't we delete the potentially newly created secret, similar to the create method?

src/zenml/zen_stores/schemas/webhook_integration_schemas.py
schustmi: I guess this is fine because it's an internal secret, still not sure it's absolutely necessary

src/zenml/zen_stores/sql_zen_store.py
```diff
@@ -8111,6 +8130,371 @@ def create_run_metadata(self, run_metadata: RunMetadataRequest) -> None:
...
+            )
+            values = self._get_secret_values(schema.secret_id)
+
+        if _WEBHOOK_SECRET_VALUE_KEY in values:
+            return values[_WEBHOOK_SECRET_VALUE_KEY]
+
+        try:
+            reference_id = UUID(values[_WEBHOOK_SECRET_REFERENCE_ID_KEY])
+            reference_key = values[_WEBHOOK_SECRET_REFERENCE_VALUE_KEY]
+            referenced_values = self._get_secret_values(reference_id)
```
schustmi: Not sure if this is going to hurt with a potentially large amount of events coming in (imagine a GH webhook for lots of events in a huge organization). Every authentication check would now require one extra secret call that could be avoided by not allowing referencing existing secrets (or allow referencing it on creation, but don't point to it but instead just duplicate the value).

Would also reduce some of the complexity in the rotate handling. WDYT?

schustmi (changes requested):

src/zenml/cli/webhook_integration.py
Json-Andriopoulos: Hmm I opted for `webhook-integration` to avoid mixups with `trigger webhook` but thinking about it now I dont have a strong opinion. Will apply

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Hmm, duplicating the value won't work that easily. We would need to keep it updated when it is rotated. I like referencing the secret as UX but you are right unless a reliable cacheing layer is available we should avoid it for now.

src/zenml/client.py
Json-Andriopoulos: Good point, yeah it doesn't make sense unless you have an accessible server.

src/zenml/zen_stores/sql_zen_store.py
schustmi: What I meant was essentially: We just copy it from their original secret, but from that point on it's completely disconnected. As in, rotating the original secret wouldn't work, they would need to rotate it via the webhook rotate request. But I guess that's also confusing.

In the current implementation though, the user wouldn't even know what secret they've connected to the webhook, which is also not ideal IMO.

src/zenml/zen_stores/schemas/webhook_integration_schemas.py
Json-Andriopoulos: Given that it is internal, I am thinking setting to restrict and blocking deletion when a webhook is using it is the best way forward. Wdyt?

schustmi: Yeah makes sense I think

src/zenml/webhooks/adapters.py
Json-Andriopoulos: Hmm, that was separated on purpose as authentication here is not against the ZenML server but a separate intake endpoint. Don't have a strong opinion on it though, will apply the change.

schustmi: No strong opinion on my end either, feel free to also leave as is.

src/zenml/zen_server/routers/webhook_integration_endpoints.py
Json-Andriopoulos: Yeap, we need request.body() which is async. 

Alternatively we could do: 
```
async def read_raw_body(request: Request) -> bytes:
    return await request.body()


@app.post("/webhook")
def webhook(
    body: Annotated[bytes, Depends(read_raw_body)],
):
    # This code is synchronous.
    print(body)

    return {
        "length": len(body),
        "text": body.decode("utf-8"),
    }
``` 
but it doesn't serve much of a purpose :/ . I am open to suggestions.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: Hmm, valid but making this operation atomic is more complex that is sounds. Current code separate secret & metadata updates which is the simplest solution imo.

src/zenml/zen_stores/sql_zen_store.py
Json-Andriopoulos: So, current code optimizes a bit the Db and Store I/O by reducing operations to a minimum:

1) We drop support for more expensive secret reference
2) We make webhook integration resolution at the intake endpoint super light
3) We make stats update 1-pass (converted nested stats to flat table).

PR #5051, opened 2026-07-09
Improve DAG endpoint performance

schustmi:
For large DAGs, python overhead of parsing step configurations and creating ORM objects dominated the runtime of the endpoint. This PR adds two improvements:
- Only parse/validate necessary fields of the step configuration.
- Don't create ORM objects for step artifacts and instead work on the raw column values.

Benchmark on a 2893 node / 5141 edge DAG:

| | Total | DB | Python | DB Statements |
|---|---|---|---|---|
| Before | 1039 ms | 178 ms | 861 ms | 18 |
| After | 505 ms | 130 ms | 375 ms | 14 |


bcdurak (approved):
Good stuff.

PR #5052, opened 2026-07-09
Dynamic pipeline improvements

schustmi:
- Add support for `CONTINUE_ON_FAILURE` execution mode. In dynamic pipelines, this works as follows: If a sync step fails, it still causes the pipeline function to fail and therefore no new steps being launched. If an async step fails, it does not cause any failure and the pipeline function as well as all ongoing/queued steps continue executing.
- Model implicit step dependencies: Any step that now has an implicit upstream dependency to the last sync step that finished executing. This means nodes in the DAG are now more well connected and represent better the runtime behaviour of the actual pipeline.

### DIFF
--- a/src/zenml/execution/pipeline/dynamic/runner.py
+++ b/src/zenml/execution/pipeline/dynamic/runner.py
@@ -317,24 +317,16 @@ class DynamicPipelineRunner:
         if not snapshot.stack:
             raise RuntimeError("Missing stack for snapshot.")
 
-        if (
-            snapshot.pipeline_configuration.execution_mode
-            == ExecutionMode.CONTINUE_ON_FAILURE
-        ):
-            logger.warning(
-                "The `%s` execution mode is not supported for "
-                "dynamic pipelines right now. "
-                "The `%s` execution mode will be used instead.",
-                snapshot.pipeline_configuration.execution_mode,
-                ExecutionMode.STOP_ON_FAILURE,
-            )
-
         self._parent_runner = parent_runner
         self._snapshot = snapshot
         self._pipeline: Optional["DynamicPipeline"] = None
         self._fail_fast = (
             snapshot.pipeline_configuration.execution_mode
             == ExecutionMode.FAIL_FAST
+        )
+        self._continue_on_failure = (
+            snapshot.pipeline_configuration.execution_mode
+            == ExecutionMode.CONTINUE_ON_FAILURE
         )
 
         worker_count = handle_int_env_var(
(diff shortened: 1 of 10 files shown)


src/zenml/execution/pipeline/dynamic/runner.py
```diff
@@ -2276,11 +2402,11 @@ def _on_step_finished(self, step_run: "StepRunResponse") -> None:
             or step_run.status == ExecutionStatus.STOPPED
         ):
             exception = self._get_step_exception(step_run=step_run)
-            self.mark_node_failed(node_id=step_run.name)
-            self.record_failure(
+            self._fail_concurrent_node(
```
bcdurak, line 2422: Not sure how likely this is, but if the status of the step is `STOPPED` here, the step will be treated similar to the async failed steps and the run will be marked as `COMPLETED`, right? Is that the expected behaviour here?

bcdurak (changes requested):
The changes look mostly good. As I was testing it, I wanted to test the changes on the DAG connections  and I may have found something. Take a look at this pipeline:

```python
from typing import List

from zenml.enums import ExecutionMode


@step
def step_01_gate() -> int:
    return 1


@step
def step_02_values() -> List[int]:
    return [1, 2, 3]


@step
def step_03_fail_on_two(value: int) -> int:
    if value == 2:
        raise RuntimeError("boom")
    return value


@step
def step_04_after_map() -> None:
    print("should this run?")


@pipeline(
    dynamic=True,
    enable_cache=False,
    execution_mode=ExecutionMode.CONTINUE_ON_FAILURE,
)
def map_failure_pipeline() -> None:
    step_01_gate()
    values = step_02_values()
    mapped = step_03_fail_on_two.map(value=values)
    step_04_after_map.submit(after=mapped)

if __name__ == "__main__":
    map_failure_pipeline()
```

When I run this, the execution gets stuck after the execution of `step_03`. Besides this, there's also a question that I left below.

src/zenml/execution/pipeline/dynamic/runner.py
schustmi: That is correct for this execution mode I think, but the whole stopping implementation at the moment is anyway very vague for dynamic pipelines. I'm not sure what the actual expected behaviour is. I assume a stopped step should always force the end of the run, even when using `CONTINUE_ON_FAILURE`?

What is probably happening today:
- Pipeline run gets set to `STOPPING`
- Isolated steps might receive it using the heartbeat, which would trigger the code you commented on. The step is in `STOPPED` state.
- If no new steps get created, the run finishes as `COMPLETED`
- If new sync steps get created, the server rejects them (no new steps for stopping runs) which probably causes the run to become `STOPPED` (our code publishes `FAILED` state, server updates to `STOPPED`)
- New async steps, no idea. Probably the run would be `COMPLETED`.

schustmi: If we think `STOPPING` > `CONTINUE_ON_FAILURE`, I think we might be able to just add a `force: bool` flag that we set in the stopping case, which always causes it to fail the pipeline run.

schustmi:
@bcdurak Your example did catch a race condition, should be fixed now!

bcdurak (changes requested):
Found another hanging case:

```python
from typing import List
from zenml.enums import ExecutionMode

@step
def values() -> List[int]:
    return [1, 2, 3]

@step
def fail_on_two(value: int) -> int:
    if value == 2:
        raise RuntimeError("boom")
    return value

@step
def consume_all(values: List[int]) -> None:
    print(f"consume_all got: {values}")

@pipeline(
    dynamic=True,
    enable_cache=False,
    execution_mode=ExecutionMode.CONTINUE_ON_FAILURE,
)
def sync_after_failed_map_result() -> None:
    mapped = fail_on_two.map(value=values())
    consume_all(mapped)

if __name__ == "__main__":
    sync_after_failed_map_result()
```

bcdurak (approved):
Can not find anything else 😄

PR #5053, opened 2026-07-09
Fix dynamic pipeline step config race condition

schustmi:
This PR fixes a race condition where a step run existed in the DB, but was missing a configuration (which was committed later in the same method). This caused issues in the DAG endpoint, which failed with a 500 error in this case.

### DIFF
--- a/src/zenml/zen_stores/sql_zen_store.py
+++ b/src/zenml/zen_stores/sql_zen_store.py
@@ -11611,6 +11611,27 @@ class SqlZenStore(BaseZenStore):
             )
 
             session.add(step_schema)
+
+            if step_run.dynamic_config:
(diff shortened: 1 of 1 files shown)


src/zenml/zen_stores/sql_zen_store.py
bcdurak, line 11639: I am not sure about this but if the process might fail with this moved logic, the error message might be a misleading here.

bcdurak (approved):

src/zenml/zen_stores/sql_zen_store.py
schustmi: I think the new logic should never fail, the only realistic possibility would be due to some integrity error, but that one wouldn't happen because the step run id is unique and we're creating the step in the same transaction.

PR #5062, opened 2026-07-10
fix: decouple the zenml logging level setting from ZENML_DEBUG env var

amitvikramraj:
The `ZENML_DEBUG` environment variable has been historically tied to two unrelated concerns:

1. it decides which telemetry target to use  (setting `ZENML_DEBUG` when telemetry is enabled will send telemetry events to the ZenML development target instead of the regular production target). 
2. it changes the logging verbosity to DEBUG when set (unless already set via its usual `ZENML_LOGGING_VERBOSITY` env variable).

This has been a constant cause of confusion and conflict, making it difficult to debug issues related to logging verbosity and telemetry and carrying a high risk of sending telemetry events to the wrong target when changes are made with the sole intention to affect the logging verbosity.

This PR makes a stand and finally detaches `ZENML_DEBUG` from the logging verbosity concern, leaving `ZENML_LOGGING_VERBOSITY` as the one and only way used to control logging verbosity.

IMPORTANT: `ZENML_DEBUG` is still kept in all the original places where it was used, even when telemetry is explicitly turned off through `ZENML_ANALYTICS_OPT_IN=false`, because of these two important aspects:

* client-side telemetry is decided by the server's telemetry setting in a client-server setup, even when the client telemetry is explicitly turned off with `ZENML_ANALYTICS_OPT_IN=false`
* the UI uses the server's telemetry setting and the `ZENML_DEBUG` environment variable to decide where to send the UI telemetry events.

### DIFF
--- a/.github/workflows/base-package-functionality.yml
+++ b/.github/workflows/base-package-functionality.yml
@@ -39,7 +39,7 @@ jobs:
     name: test-base-package-functionality
     runs-on: ${{ inputs.os }}
     env:
-      ZENML_DEBUG: 1
+      ZENML_LOGGING_VERBOSITY: debug
       ZENML_ANALYTICS_OPT_IN: false
       PYTHONIOENCODING: utf-8
       UV_HTTP_TIMEOUT: 600
--- a/.github/workflows/base-package-functionality.yml
+++ b/.github/workflows/base-package-functionality.yml
@@ -34,13 +34,14 @@ on:
         description: Git branch or ref
         type: string
         required: true
+env:
+  ZENML_LOGGING_VERBOSITY: debug
+  ZENML_ANALYTICS_OPT_IN: false
 jobs:
   test-base-package-functionality:
     name: test-base-package-functionality
     runs-on: ${{ inputs.os }}
     env:
-      ZENML_LOGGING_VERBOSITY: debug
-      ZENML_ANALYTICS_OPT_IN: false
       PYTHONIOENCODING: utf-8
       UV_HTTP_TIMEOUT: 600
       OBJC_DISABLE_INITIALIZE_FORK_SAFETY: 'YES'
--- a/.github/workflows/base-package-functionality.yml
+++ b/.github/workflows/base-package-functionality.yml
@@ -35,6 +35,7 @@ on:
         type: string
         required: true
 env:
+  ZENML_DEBUG: 1
   ZENML_LOGGING_VERBOSITY: debug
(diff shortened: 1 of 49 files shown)


.github/workflows/base-package-functionality.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     name: test-base-package-functionality
     runs-on: ${{ inputs.os }}
     env:
-      ZENML_DEBUG: 1
+      ZENML_LOGGING_VERBOSITY: debug
       ZENML_ANALYTICS_OPT_IN: false
```
stefannica:
```suggestion

env:
  ZENML_LOGGING_VERBOSITY: debug
  ZENML_ANALYTICS_OPT_IN: false

jobs:
  test-base-package-functionality:
    name: test-base-package-functionality
    runs-on: ${{ inputs.os }}
    env:
```

stefannica (approved):

tests/unit/test_general.py
```diff
@@ -17,16 +17,19 @@
...
 from zenml.materializers.materializer_registry import materializer_registry
 from zenml.metadata.metadata_types import MetadataTypeTuple
 
 
 def test_debug_mode_enabled_for_tests():
-    """Checks that the ZENML_DEBUG is set in the tests."""
-    assert os.environ[ENV_ZENML_DEBUG] == "true"
+    """Checks that the ZENML_DEBUG is set in the tests or ZENML_ANALYTICS_OPT_IN is set to false."""
+    assert (
+        os.environ[ENV_ZENML_DEBUG] == "true"
```
bcdurak: It's gonna fail.

bcdurak (changes requested):

helm/templates/_environment.tpl
```diff
@@ -643,6 +643,8 @@ SSL_CERT_FILE: "/updated-certs/ca-certificates.crt"
 {{- end }}
 {{- if $server.debug }}
 ZENML_LOGGING_VERBOSITY: "DEBUG"
```
bcdurak, line 646: Shouldn't this change the `ZENML_DEBUG`?

stefannica: We deprecated it in favor of `logging.verbosity`.

tests/unit/test_general.py
stefannica: Reverted

bcdurak (approved):

PR #5064, opened 2026-07-10
Add Slurm integration (step operator + orchestrator)

### LINKED ISSUE #5063: Add a Slurm integration (step operator over SSH/local sbatch)
## Feature Description

Add a `slurm` integration so ZenML steps can run on Slurm clusters — the scheduler of most HPC and on-prem GPU environments:

- **`SlurmStepOperator` (v0)** — run individual (typically GPU/training) steps as Slurm jobs: render an `sbatch` script from step resources and settings, submit, poll status, stream the job output, cancel on abort.
- **A small `SlurmClient` abstraction** with two transports from day one:
  - `ssh` — submit to a remote login/controller node (reusing the SSH connection layer from the existing `ssh` integration: host/key config, secret-backed keys, paramiko client).
  - `local` — plain subprocess for when the orchestrator itself runs on the login node (the pattern HPC teams use today with cron/Snakemake/dask-jobqueue).
- **`SlurmOrchestrator` (v1, follow-up)** — map the pipeline DAG onto `sbatch --dependency=afterok:<id>` chains so Slurm itself orchestrates the run.

## Problem or Use Case

Slurm is the de facto scheduler for scientific computing and for a large share of enterprise/on-prem GPU clusters (post-training and fine-tuning workloads included). Teams in these environments currently cannot target their cluster from a ZenML pipeline at all — there is no Slurm flavor of any stack component.

Comparable tools validate both the demand and the approach: Metaflow ships `metaflow-slurm` (an out-of-tree extension that SSHes to the cluster and submits shell scripts via `sbatch`), and Prefect has an open docs issue asking for even a tutorial, with the community pattern being a worker on the login node + dask-jobqueue. A first-class, in-tree stack component with ZenML's secret-backed connection config would be ahead of both.

### Execution model (the key design decision)

Slurm compute nodes generally cannot run Docker (no root), so unlike every existing step operator, the Slurm operator is **container-free by design** in v0:

- The user provides an `env_setup_command` (e.g. `source /shared/venvs/zenml/bin/activate`) pointing at a Python environment on the cluster's shared filesystem with `zenml` installed.
- The generated sbatch script activates that environment and runs the ZenML step entrypoint; pipeline code reaches the node via ZenML's code-upload/download path (no image build).
- Container support for HPC (Apptainer/Singularity, Pyxis/enroot) is a v1 setting, not a v0 blocker.

### Settings

`partition`, `time_limit`, `account`, `qos`, `nodes`/`ntasks`, `extra_sbatch_directives` (escape hatch), `env_setup_command`, `workdir` (shared filesystem path), plus `ResourceSettings` mapping (`cpu_count` → `--cpus-per-task`, `memory` → `--mem`, `gpu_count` → `--gres=gpu:N`).

### Status and logs

- Prefer structured CLI output where available (`sbatch --parsable`, `squeue --json` / `sacct --json` on Slurm ≥ 21.08), with plain-format fallbacks for older clusters.
- Do not assume `sacct` (accounting is often not configured): poll `squeue`, and detect terminal state via an exit-code sentinel file written by the job script.
- Stream the job's output file back during polling for live logs.

## Alternatives Considered

- **PySlurm (C bindings)** — must run inside the cluster (munge key, matching `slurm.conf`, network membership) and is version-locked to the Slurm major release. Wrong fit for a client that submits from outside.
- **submitit** — assumes local `sbatch` and pickles closures to a shared FS; its core value (function pickling) is redundant next to ZenML's entrypoint mechanism.
- **slurmrestd (REST API)** — the principled remote API, but opt-in, admin-configured, version-churning per release, and usually unreachable from outside the cluster network (so SSH-tunneled anyway). Slots in later as a third `SlurmClient` transport if adoption earns it.
- **SSH + CLI** is what every comparable production tool converges on (Metaflow's extension, Nextflow, Snakemake profiles, Parsl's SSH provider).

## Additional Context

Reference points in the codebase:

- SSH connection layer to reuse: `src/zenml/integrations/ssh/` (`ssh_client.py`, `flavors/base.py` connection mixin) — note its execution model (SSH + `docker run`) does not transfer; only the transport does.
- Step operator lifecycle to implement: `src/zenml/step_operators/base_step_operator.py`.
- Recent new-integration shape to mirror: `src/zenml/integrations/digitalocean/` (#5047/#5054).

Known sharp edges to validate:

- Code + dependency delivery without containers (shared-FS environment is a documented v0 requirement).
- `sacct`/accounting availability differs per cluster; the sentinel-file fallback is required.
- Long queue times vs. polling timeouts (a queued-but-healthy job must not be treated as failed).
- Managed/zero-config Slurm offerings (e.g. cloud Slurm clusters) restrict images and expose only SSH — the integration must not assume admin access to the cluster.

## Priority

Medium — opens the HPC/on-prem GPU cluster segment that currently has no ZenML story.

htahir1:
Closes #5063

## What this does

Adds a **Slurm integration** so ZenML pipelines can run on an HPC cluster managed by the Slurm scheduler. It ships two stack components:

- **Step operator** (`slurm`) — runs a single step as a Slurm batch job (delegating the rest of the pipeline to another orchestrator).
- **Orchestrator** (`slurm`) — submits a whole pipeline as a graph of Slurm jobs wired together with `sbatch --dependency=afterok` chains.

Both run the step's **standard ZenML Docker image** on the compute node via a rootless HPC container runtime (Apptainer/Singularity by default, or NVIDIA Pyxis/enroot, or the Docker daemon where available), so code delivery uses ZenML's normal image build — nothing is shipped by the components. Submission goes over **SSH** to a login/controller node (reusing the `ssh` integration's `SSHClient` + connection mixin) or via **local `sbatch`** when the client already runs on the cluster.

## Design decisions

- **Docker image, not a code tarball.** An earlier approach shipped code as an archive on `PYTHONPATH`; that fought ZenML's entrypoint (which `chdir`s into the image's `/app`) and reimplemented delivery the framework already does. Switched to running the step image with an HPC container runtime.
- **No service connector.** Follows the `ssh` integration's `SecretField`/connection-mixin pattern rather than a bespoke connector.
- **Detached orchestrator, modeled on the current `ssh` orchestrator** (not the deprecated HyperAI). Submits the whole DAG upfront in topological order (via `topsorted_layers`); each step self-reports status, and the run id is the **placeholder run id** (unique per run), not the snapshot id.
- **Shared machinery factored out**: `flavors/base.py` (config/settings mixins), `slurm_job.py` (container-command + sbatch-script builders, `stage_and_submit`, stack validator), and `build_slurm_client` (single transport factory) are shared by both components.

## Reliability & security hardening

Following an external review, the failure/status and credential paths were hardened:

- **`fetch_status`** reconciles a detached run from its Slurm jobs (`squeue --jobs` with a `--states=all` lookup for jobs that already left the active queue), catching failures that occur before the ZenML entrypoint starts (bad image, runtime error) which a step can never self-report. The missing-sentinel window is treated as still-running, not terminal.
- **Exact job-id** status/cancellation; the Slurm job id is persisted as step-run metadata.
- **Private registry auth** per runtime (Apptainer/Singularity `*_DOCKER_USERNAME/PASSWORD`, enroot `.credentials`, Docker `config.json`) via owner-only files — no passwords on any command line.
- **Credential lifecycle**: owner-only staging (`mkdir` under `umask 077`, `0600` env file), an EXIT trap that scrubs credentials + records the exit code, staging rollback on submission failure, and an `afterany` cleanup/reaper job (depending on **all** step jobs) that scrubs credential files even for jobs cancelled before their trap runs.
- **Runtime safety**: env-var names validated and values shell-escaped (Pyxis `export`s quoted values instead of `source`ing a Docker-format file); Apptainer/Singularity run with `--no-eval`; CPU requests rounded up and memory emitted in MB so fractional/sub-GB resource settings produce valid `#SBATCH` directives.

## Testing

- **43 unit tests** cover config/validator, the client (submit/state/cancel by id), per-runtime container commands, registry auth (no password on the command line), env escaping, resource conversion, DAG dependency wiring, the reaper, and `fetch_status`. The Slurm CLI and SSH are never invoked — everything runs against an in-memory fake command runner.
- `ruff`, `ruff format`, and `mypy` are clean on the package.
- ⚠️ **Not yet validated on a real cluster.** Integrations that talk to external services are typically verified locally/in CI against live infra; that step is pending a deployed Slurm cluster. Please run full CI.

## Notes for reviewers

- Closest reference points are the `ssh` orchestrator/step operator (SSH + Docker, detached submit) and `ContainerizedOrchestrator`/`kubernetes` for the image/DAG plumbing.
- The run-metadata round-trip for the job-id map and the exact behavior on a real scheduler are the two things that can only be confirmed live.


src/zenml/integrations/slurm/orchestrators/slurm_orchestrator.py
```diff
@@ -0,0 +1,1378 @@
...
+        submitted: Dict[str, str] = {}
+        sensitive_paths: List[str] = []
+        cleanup_job_id: Optional[str] = None
+        cleanup_settings = cast(
+            SlurmOrchestratorSettings, self.get_settings(snapshot)
+        )
+        container_registry = stack.container_registry
+        assert container_registry is not None
+        try:
+            for step_name in self._sorted_step_names(steps):
```
schustmi: steps are already sorted

src/zenml/integrations/slurm/orchestrators/slurm_orchestrator.py
```diff
@@ -0,0 +1,1378 @@
...
+                if name in steps[other].spec.upstream_steps
+            ]
+
+        layers = topsorted_layers(
+            nodes=list(steps),
+            get_node_id_fn=lambda name: name,
+            get_parent_nodes=parents,
+            get_child_nodes=children,
+        )
+        return [name for layer in layers for name in layer]
```
schustmi:
```suggestion
```

src/zenml/integrations/slurm/orchestrators/slurm_orchestrator.py
```diff
@@ -0,0 +1,1378 @@
...
+
+        run_id = str(step_run_info.run_id)
+        run_dir = self._isolated_run_dir(
+            run_id=run_id, step_run_id=str(step_run_info.step_run_id)
+        )
+        client = build_slurm_client(self.config)
+        job_id: Optional[str] = None
+        sensitive_paths: List[str] = []
+        try:
+            job_id, sensitive_paths = self._submit_container_job(
```
schustmi: Submitting a slurm job from an active slurm job seems like a horrible idea which will lead to starvation.

The new job will be queued, and the active job will be waiting and blocking resources while all other jobs in the queue are executing.

schustmi (changes requested):

src/zenml/integrations/slurm/orchestrators/slurm_orchestrator.py
```diff
@@ -0,0 +1,1378 @@
...
+            )
+        stack = Stack.from_model(step_run_info.snapshot.stack)
+        container_registry = stack.container_registry
+        assert container_registry is not None
+
+        run_id = str(step_run_info.run_id)
+        run_dir = self._isolated_run_dir(
+            run_id=run_id, step_run_id=str(step_run_info.step_run_id)
+        )
+        client = build_slurm_client(self.config)
```
schustmi: Shouldn't this always use a local runner, because by definition we're not running in the slurm cluster already?

schustmi (changes requested):

Issue #4658, opened 2026-03-31
Audit GitHub Actions workflow permissions and persist-credentials usage


strickvl:
## Context

Our zizmor security linter config (`.github/zizmor.yml`) currently has two checks disabled with TODOs:

1. **`excessive-permissions`** (disabled) — Many workflows use default permissions rather than explicit minimal permissions. GitHub's default token has broad read/write access, which violates the principle of least privilege.

2. **`artipacked`** (disabled) — Most `actions/checkout` steps use the default `persist-credentials: true`, which leaves the `GITHUB_TOKEN` available to all subsequent steps. Workflows that don't need to push commits should set `persist-credentials: false`.

## Proposed work

Audit all 44 workflow files in `.github/workflows/` and:

### Permissions
- Add explicit `permissions:` blocks to every workflow
- Use the minimum required permissions for each workflow's actual needs
- Reference: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

### persist-credentials
- Add `persist-credentials: false` to every `actions/checkout` step that does NOT need to push commits
- Document which workflows legitimately need `persist-credentials: true` and why

### After completion
- Re-enable `excessive-permissions` and `artipacked` checks in `.github/zizmor.yml`
- Add any necessary exceptions with documented reasons

## Workflows that likely need push access
- `release.yml` (tagging, publishing)
- `update-templates-to-examples.yml` (pushing template sync commits)
- `snack-it.yml` (creating issues)
- `publish_*.yml` workflows

## Reference
- zizmor docs: https://woodruffw.github.io/zizmor/
- Current config: `.github/zizmor.yml` (lines 21-33)

qubeena07:
Starting on this, opened PR #5075 as a first batch covering the linting, unit test, and doc link checking workflows. Plan is to keep going in a few more small batches rather than one giant PR, since that matches what the issue already suggests. Will follow up with the CI entry point workflows and then the release and publish ones next.

PR #5075, opened 2026-07-13
Add explicit permissions to linting, unit test, and doc link workflows

qubeena07:
## Summary

First batch of the workflow permissions audit tracked in #4658.

These five workflows relied on the default GITHUB_TOKEN permissions, which are broader than they need:

- linting.yml
- unit test.yml
- spellcheck.yml
- check links.yml
- check markdown links.yml

Each now declares an empty workflow level permissions block and grants only what each job actually uses, following the same pattern already used in dependency audit.yml and zizmor.yml. Most jobs only need contents read for checkout. The jobs that post a PR comment (post summary in check links.yml and check links in check markdown links.yml) get pull requests write instead. Checkout steps that do not push commits now also set persist credentials to false.

More workflows remain to audit, this covers the CI linting and doc link checking group as a first incremental step per the issue's own guidance to address this incrementally across multiple PRs.

## Test plan

- [x] Ran zizmor locally against the changed files with the repo config, no findings
- [x] Ran yamlfix on the changed files, no changes needed
- [x] Reviewed each job's steps to confirm no push or write operation depends on the removed default permissions


qubeena07:
Hi @AlexejPenner @bcdurak is there any updates on this PR? Since github flagged it for stale PR.

bcdurak:
hey @qubeena07, sorry about the late response. We have [a quite significant change](https://github.com/zenml-io/zenml/pull/5075) lined-up for the CI. I would suggest doing this once the big refactor branch makes it in as it might result into quite a lot of conflicts of open points.

PR #5079, opened 2026-07-14
Allow passing raw values as step parameters in dynamic pipelines

schustmi:
Add env var to specify a threshold that decides whether a JSON-serializable raw value passed to a step in a dynamic pipeline gets treated as a parameter or artifact. Defaults to 0, which means all values get passed as artifacts to preserve backwards compatibilty.

There exist explicit versions to enforce param/artifact already:
```python
@pipeline(dynamic=True):
def p():
  my_step = my_step.with_options(parameters={"always_a_param": 1})
  my_step(always_an_artifact=ExternalArtifact(1))
```


docs/book/how-to/steps-pipelines/dynamic_pipelines.md
```diff
@@ -93,6 +93,25 @@ One caveat: when you use `child_pipeline.embed(...)`, the child pipeline's own
...
 parent run's configuration controls the steps that execute inline.
 
+### Step inputs as parameters
+
+Any value you pass to a step that is not the output of another step is uploaded
+to the artifact store as an external artifact, even a small `int`. Each upload
+costs a write to the artifact store and a request to the server, which adds up
+when a pipeline calls steps in a loop.
+
+Set the `ZENML_PARAMETER_SIZE_THRESHOLD` environment variable to pass
```
bcdurak, line 103: A very small nitpick here. As far as I can see, the resolution of this variable happens within the dynamic pipeline entrypoint, so the env var needs to be set in this environment. Setting it locally and then running the pipeline wouldn't have any effect. Might be nice to mention it here.

bcdurak (approved):

docs/book/how-to/steps-pipelines/dynamic_pipelines.md
schustmi: done

Issue #2262, opened 2024-01-10
Resolve Vertex Scheduler Character Limit Issue for Long Pipeline Names


strickvl:
## Open Source Contributors Welcomed!
Please comment below if you would like to work on this issue!

### Contact Details [Optional]

support@zenml.io

## What happened?
A user encountered an issue where the Vertex scheduler in ZenML does not accept pipeline names longer than 64 characters. This limitation requires users to shorten their pipeline names to successfully schedule their pipelines on Vertex AI, which can be restrictive and inconvenient.

## Task Description
Investigate and address the issue with the Vertex scheduler's character limit for pipeline names in ZenML. The goal is to either increase the character limit or implement a mechanism that automatically handles longer names without user intervention.

## Expected Outcome
Users should be able to schedule pipelines on Vertex AI without the constraint of a 64-character limit on pipeline names.
ZenML should provide a seamless experience, either by accommodating longer names or by intelligently managing them within the existing Vertex scheduler limitations.
This fix will enhance user convenience and flexibility in naming pipelines.

## Steps to Implement
- Reproduce the issue to understand the exact limitations and behavior of the Vertex scheduler with respect to pipeline names.
- Explore solutions such as increasing the character limit or automatically abbreviating/transforming long pipeline names while preserving their uniqueness and recognizability.
- Implement the chosen solution in a way that is transparent and intuitive for the user.
- Test the implementation with various pipeline name lengths to ensure compatibility and functionality.
- Update documentation to inform users about the handling of pipeline names in Vertex scheduler and any best practices or limitations.

## Additional Context
This issue is crucial for users who require descriptive and potentially lengthy names for their pipelines, especially in complex or large-scale MLOps environments.

## Code of Conduct
- [ ] I agree to follow this project's Code of Conduct

strickvl:
@christianversloot if I'm not completely mistaken, we had a conversation about this in Slack? Unfortunately the slack history hasn't been retained.

christianversloot:
Correct, but that was for SageMaker orchestrator. There, I believe we limit name length to circumvent this error. It could perhaps be inspiration for a fix here :)

strickvl:
@christianversloot ah yes! https://github.com/zenml-io/zenml/pull/1505 was the PR.

ashutosh887:
@strickvl @christianversloot 
I would like to work on this... Please assign!

Merthoshan:
has the bug been solved? if not i would like to solve it

[...]

PR #5080, opened 2026-07-15
WIP: Fix Vertex AI scheduler 64-character limit for long pipeline names (#2262)

hrushi2602-cyber:
## Description

**Relates to Issue:** #2262

### The Problem
When submitting a pipeline with a name longer than 64 characters to the Vertex AI orchestrator, the API rejects the payload due to strict `display_name` character limits. Simply slicing the string at 64 characters risks name collisions if multiple pipelines share a long prefix.

### The Solution
Introduced a local helper function, `sanitize_vertex_job_name`, inside `vertex_orchestrator.py` that intercepts both `orchestrator_run_name` and `pipeline_name`. 
* If the name is 64 characters or fewer, it passes through unchanged.
* If it exceeds 64 characters, it generates an MD5 hash of the full name, extracts an 8-character short hash, and appends it to the truncated string. 
* This ensures the final `display_name` is exactly 64 characters and guaranteed unique to avoid overwriting jobs.

### Current Status (WIP)
I am opening this as a Draft PR to show intent and verify this hashing approach aligns with the maintainers' expectations before wiring up the full test coverage. 

- [x] Implement hash-truncation logic
- [x] Apply to Vertex AI payload variables
- [ ] Add unit tests (Pending feedback on the approach)

Looking forward to any feedback!


schustmi (changes requested):
Hey @hrushi2602-cyber, thanks for the PR! Can you please rebase your changes on the `develop` branch and also `develop` as a base branch for your PR.

Regarding the implementation:
We already have something very similar to this implemented in the `get_orchestrator_run_name(...)` function (see `src/zenml/orchestrators/utils.py`). In my opinion, it would be nicest if we do the following:
- Add a new function in the `string_utils` module. This should be generic, receives as input some original string, a optional maximum length, a suffix length and maybe also a separator string (e.g. "-" or "_") . It should then generate a random suffix, truncate the original string so a suffix of that length fits and then merge them.
- We call this function both from `get_orchestrator_run_name(...)` as well as the one you introduced (`sanitize_vertex_job_name(...)`)

hrushi2602-cyber:
Thanks for the feedback! I've updated the PR based on your suggestions. Here is a summary of the changes:

Rebased and Targeted develop: I have successfully changed the PR base branch to develop and rebased my local commits on top of the latest changes from the upstream develop branch.

Centralized String Utility: Added a new generic function append_random_suffix in src/zenml/utils/string_utils.py. It handles suffix generation, strict length limit enforcement, and safe string merging. I also included a charset argument to ensure backward compatibility with existing components that might expect hexadecimal strings.

Refactored get_orchestrator_run_name: Updated src/zenml/orchestrators/utils.py to offload the string truncation and merging logic to the new utility. The original logic that dynamically shrinks the suffix (while enforcing the minimum of 8 characters) and uses the hexadecimal charset was fully preserved so this won't break downstream orchestrators.

Refactored sanitize_vertex_job_name: This function now acts as a pre-filter to enforce Vertex-specific rules (lowercase, alphanumeric, starting with a letter, replacing underscores with hyphens), and then defers to append_random_suffix to safely enforce the 64-character limit.

src/zenml/orchestrators/utils.py
```diff
@@ -64,24 +65,23 @@ def get_orchestrator_run_name(
...
-    return f"{pipeline_name}{suffix}"
-
+        # The -1 accounts for the "_" separator added by the utility function
+        suffix_length = min(32, max(8, max_length - len(pipeline_name) - 1))
+
+    return append_random_suffix(
+        original_string=pipeline_name,
+        suffix_length=suffix_length,
+        max_length=max_length,
+        separator="_",
```
schustmi: I think we should also use `separator=""` here for backwards compatibility. All orchestrators use this function, and they might not allow underscores.

src/zenml/utils/string_utils.py
```diff
@@ -27,6 +27,32 @@
...
+    original_string: str,
+    suffix_length: int,
+    max_length: Optional[int] = None,
+    separator: str = "-",
+    charset: str = string.ascii_lowercase + string.digits
+) -> str:
+    """
+    Appends a random suffix to a string, truncating the original if necessary.
+    """
+    # Use the provided charset instead of hardcoding alphanumeric
```
schustmi:
```suggestion
```

src/zenml/utils/string_utils.py
```diff
@@ -27,6 +27,32 @@
...
 
 
+def append_random_suffix(
+    original_string: str,
+    suffix_length: int,
+    max_length: Optional[int] = None,
+    separator: str = "-",
+    charset: str = string.ascii_lowercase + string.digits
+) -> str:
+    """
```
schustmi, line 48: Please adjust these to follow the google docstring format.

src/zenml/orchestrators/utils.py
```diff
@@ -64,24 +65,23 @@ def get_orchestrator_run_name(
...
+        charset="0123456789abcdef" # <--- Ensures 100% backward compatibility
```
schustmi:
```suggestion
        charset="0123456789abcdef"
```

schustmi (changes requested):
Few minor things, other than that looks good to me

hrushi2602-cyber:
All requested styling changes have been made! Here is what was updated in the latest commits:

Google Docstrings: Updated the append_random_suffix utility to strictly follow the Google docstring format.

Linter Fixes: Swapped single quotes to double quotes, removed the duplicate re import, and properly sorted the import block in vertex_orchestrator.py.

I ran the ruff checks locally on the files changed by me and everything is passing clean with Ruff now. Let me know if you need anything else.

hrushi2602-cyber:
@schustmi Just a gentle reminder! I just applied your suggested changes just check once and let me know... Just commenting in case it got burried under due to the weekend!!

hrushi2602-cyber:
@schustmi still waiting here....can I get a feedback please?

hrushi2602-cyber:
Would be great if someone suggests changes or helps merge, very politely, there's weeks of work done by me here...I really don't want it to be wasted...

PR #5082, opened 2026-07-15
Fix SSH orchestrator container name collisions on pipeline re-runs

kounelisagis:
The SSH orchestrator names each step's Compose service and container after the snapshot id (`<snapshot-id>-<step>`). But a single snapshot gets run by many pipeline runs — re-running from the dashboard, run templates, concurrent triggers — so the second launch tries to create containers that already exist from the first, and `docker compose up` fails on the host with `container name already in use`. In practice you can only run a given snapshot once before having to SSH in and clean up the leftover containers.

This keys the service names, container names and `depends_on` edges on the run id instead, so every run gets unique names. That's already what the per-run remote directory and the dynamic path do. The snapshot id still drives image selection and entrypoint arguments, which really are snapshot-scoped.

While in there: the dynamic path was writing its Docker `--env-file` by hand, so a newline (or NUL) in a value would silently corrupt the file — and that's the file carrying the store token. It now goes through the same `serialize_env_for_docker_env_file` helper the step operator already uses, which rejects those values with a clear error instead.

Added regression tests for both cases: two runs of the same snapshot no longer share container names, and a corrupting env value now raises rather than writing a broken file.


src/zenml/integrations/ssh/orchestrators/ssh_orchestrator.py
```diff
@@ -331,7 +333,11 @@ def _step_service(
 
         service: Dict[str, Any] = {
             "image": self.get_image(snapshot=snapshot, step_name=step_name),
-            "container_name": f"{snapshot.id}-{step_name}",
+            # Container names must be unique per run, not per snapshot: the same
+            # snapshot can be launched multiple times (re-runs, run templates,
+            # concurrent triggers). Keying on snapshot.id would collide on the
+            # remote host with "container name already in use".
+            "container_name": f"{run_id}-{step_name}",
```
schustmi: Maybe a helper `_get_container_name(snapshot_name, step_name)` so that we can change it in one place going forward?

src/zenml/integrations/ssh/orchestrators/ssh_orchestrator.py
```diff
@@ -331,7 +333,11 @@ def _step_service(
 
         service: Dict[str, Any] = {
             "image": self.get_image(snapshot=snapshot, step_name=step_name),
-            "container_name": f"{snapshot.id}-{step_name}",
+            # Container names must be unique per run, not per snapshot: the same
+            # snapshot can be launched multiple times (re-runs, run templates,
+            # concurrent triggers). Keying on snapshot.id would collide on the
+            # remote host with "container name already in use".
```
schustmi:
```suggestion
```

Useful to know in the PR, but doesn't really matter in code going forward IMO

schustmi (changes requested):
Thanks for the PR @kounelisagis! Left a few small comments

schustmi (approved):

PR #5086, opened 2026-07-16
Add project metadata

strickvl:
## Summary

- add arbitrary `project_metadata` to Project create, update, and hydrated response models
- persist metadata as portable JSON text with a rolling-deployment-safe migration
- preserve metadata when omitted and replace it, including clearing with `{}`, when explicitly supplied

## Testing

- focused model and schema unit tests
- direct SQL and local-server REST integration tests
- migration upgrade, existing-row backfill, downgrade, and old-worker insert checks


src/zenml/zen_stores/schemas/project_schemas.py
```diff
@@ -162,7 +173,11 @@ def update(self, project_update: ProjectUpdate) -> "ProjectSchema":
         for field, value in project_update.model_dump(
             exclude_unset=True
         ).items():
-            setattr(self, field, value)
+            if field == "project_metadata":
+                if value is not None:
+                    self.project_metadata = json.dumps(value)
```
schustmi: pydantic_encoder to support uuids etc?

src/zenml/zen_stores/schemas/project_schemas.py
```diff
@@ -62,6 +64,14 @@ class ProjectSchema(NamedSchema, table=True):
 
     display_name: str
     description: str
+    project_metadata: str = Field(
+        default="{}",
+        sa_column=Column(
+            Text().with_variant(MEDIUMTEXT, "mysql"),
+            nullable=False,
+            server_default=text("('{}')"),
```
schustmi: don't like this, just make it nullable IMO and if we want the response non-optional handle it in `to_model()`

src/zenml/zen_stores/schemas/project_schemas.py
```diff
@@ -62,6 +64,14 @@ class ProjectSchema(NamedSchema, table=True):
 
     display_name: str
     description: str
+    project_metadata: str = Field(
+        default="{}",
+        sa_column=Column(
+            Text().with_variant(MEDIUMTEXT, "mysql"),
```
schustmi: I think we usually use `String(length=MEDIUMTEXT_MAX_LENGTH).with_variant(
                MEDIUMTEXT, "mysql"
            ),`

in other tables, not sure how they differ but prob best to stay consistent

schustmi (changes requested):

schustmi (approved):

PR #5087, opened 2026-07-16
Escape $ in generated Docker Compose files

kounelisagis:
Docker Compose interpolates `$VAR` and `${VAR}` across the whole compose file before starting containers. The SSH and HyperAI orchestrators write already resolved step environments (secrets included) and bind mount paths straight into the file, so any value containing `$` was corrupted: a `pa$$word` secret reached the container as `pa$word`, host environment values leaked in, and values shaped like `${X:?}` aborted `docker compose up` and the whole run.

Both orchestrators now serialize through a shared `dump_compose_yaml` helper that escapes every `$` as `$$`, so values round trip verbatim. The escape covers the whole document because mount paths are affected too, not only the environment block.

Tests: helper round trip, plus an SSH regression asserting values containing `$` survive.

### DIFF
--- a/src/zenml/orchestrators/utils.py
+++ b/src/zenml/orchestrators/utils.py
@@ -19,6 +19,8 @@ import shlex
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, cast
 from uuid import UUID
 
+import yaml
+
 from zenml.client import Client
 from zenml.config.global_config import (
     GlobalConfiguration,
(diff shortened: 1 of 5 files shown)


tests/unit/integrations/ssh/orchestrators/test_ssh_orchestrator.py
```diff
@@ -223,6 +223,51 @@ def test_container_names_unique_across_runs_of_same_snapshot(self) -> None:
...
+        (secrets included), a raw ``$`` would be silently mangled or, for
+        forms like ``${X:?}``, abort ``docker compose up``. The uploaded file
+        must escape ``$`` so every value round-trips verbatim.
+        """
+        orch = _make_orchestrator()
+        snap = _fake_snapshot({"only": _fake_step()})
+        run = MagicMock()
+        run.id = uuid4()
+        hostile = {
+            "DB_PASSWORD": "pa$$word",
```
Json-Andriopoulos: This is what is causing the git guardian alarm I think. It is an obvious false positive but let's add more neutral values to get rid of it :)

src/zenml/orchestrators/utils.py
```diff
@@ -284,6 +286,31 @@ def shell_join(command: List[str]) -> str:
     return shlex.join(command)
 
 
+def dump_compose_yaml(compose: Dict[str, Any]) -> str:
```
Json-Andriopoulos: Let's apply the python3.10+ typing syntax here: `compose: dict[str, Any]`

src/zenml/orchestrators/utils.py
```diff
@@ -284,6 +286,31 @@ def shell_join(command: List[str]) -> str:
...
+    serializer emits.
+
+    Args:
+        compose: The Docker Compose definition.
+
+    Returns:
+        YAML text ready to write to a ``docker-compose.yml``.
+    """
+    yaml_text: str = yaml.dump(
+        compose, default_flow_style=False, sort_keys=False
```
Json-Andriopoulos: I think `default_flow_style=False` is redundant here.

tests/unit/orchestrators/test_utils.py
```diff
@@ -16,16 +16,63 @@
...
+    return yaml_text.replace("$$", "$")
+
+
+def test_dump_compose_yaml_escapes_dollar_signs():
+    """Every ``$`` is escaped so Compose does not interpolate our values."""
+    compose = {
+        "services": {
+            "step": {
+                "environment": {
+                    "DB_PASSWORD": "pa$$word",
```
Json-Andriopoulos: Same here, let's add more neutral values.

Json-Andriopoulos (commented):
Thanks for this PR @kounelisagis . I added some minor comments, feel free to address them at your own pace and ping me to approve & merge.

Json-Andriopoulos:
@kounelisagis One last thing: To fix the failing lint checks you can do the following: `./scripts/format.sh` and run `./scripts/lint.sh` to verify that they are resolved. Thanks

Json-Andriopoulos (approved):

PR #5085, opened 2026-07-16
Fix clean_requirements dropping > and != version specifiers

chuenchen309:
## Describe changes

`clean_requirements` de-duplicates a requirements list by extracting each package name and keying a dict on it. The name is extracted by splitting on each operator in turn, but the chain covers `>=` without covering a bare `>`, and covers `==` which does not match `!=`. Requirements using those two operators keep the operator and version in their key, so they never collide with the bare package name and both entries survive:

```python
clean_requirements(["package1", "package1>1.5.0"])
# actual:   ["package1", "package1>1.5.0"]
# expected: ["package1>1.5.0"]

clean_requirements(["package1", "package1!=1.5.0"])
# actual:   ["package1", "package1!=1.5.0"]
# expected: ["package1!=1.5.0"]
```

`<=` is unaffected because splitting on `<` already covers it — `>` is the only comparison operator that lacks that coincidental coverage. Both `>` and `!=` are legal PEP 440 specifiers.

The fix splits on `>` instead of `>=` (the former subsumes the latter) and adds a split on `!=`. The second condition (`any(op in req for op in ["=", ">", ...])`, which decides that a constrained requirement beats a bare one) already handles both operators and is unchanged.

**Reachability**: `databricks_utils.py:161` calls `clean_requirements(sorted(set(requirements)))` on a user's parsed requirements, so a `requirements.txt` containing `foo>1.0` alongside a bare `foo` currently ships both lines to Databricks.

**Tests**: the existing `test_clean_requirements` parametrize list covers `==`, `>=`, `<`, `~=`, `^=` and extras, but has no case for `>` or `!=` — exactly the gap. I added cases for both, plus a `<=` case pinning the behavior that already works. The three new operator cases fail on `develop` and pass with this change; the full file is 16/16 green.

## Pre-requisites
Please ensure you have done the following:
- [x] I have read the **CONTRIBUTING.md** document.
- [x] I have added tests to cover my changes.
- [x] I have based my new branch on `develop` and the open PR is targeting `develop`.
- [ ] **IMPORTANT**: I made sure that my changes are reflected properly in the following resources:
  - [ ] [ZenML Docs](https://docs.zenml.io) — internal helper, not documented; no doc change needed
  - [ ] Dashboard: not affected
  - [ ] Templates: not affected
  - [ ] [Projects](https://github.com/zenml-io/zenml-projects): not affected

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
[...]

PR #5089, opened 2026-07-16
Count an installed prerelease as installed

chuenchen309:
## What

`requirement_installed` reports an installed prerelease as not installed:

```python
# pkg 1.0.0rc1 is installed
>>> requirement_installed("pkg")
False        # expected True
```

`SpecifierSet.contains` excludes prereleases unless asked otherwise. That's the sensible default when *resolving* a requirement, but this function is asking a different question — "is this already installed?" — and the answer shouldn't depend on whether the installed version happens to be an rc.

The empty-specifier case is the sharp edge: a requirement with **no version constraint at all** still hits the default, so a bare `"pkg"` doesn't match an installed `1.0.0rc1`.

```
requirement 'pkg'         vs installed 1.0.0rc1 -> False   (prereleases=True: True)
requirement 'pkg'         vs installed 1.0.0    -> True
requirement 'pkg>=1.0'    vs installed 1.5.0rc1 -> False   (prereleases=True: True)
requirement 'pkg>=1.0'    vs installed 1.5.0    -> True
requirement 'pkg==1.0.0rc1' vs installed 1.0.0rc1 -> True   # explicit prerelease pin already works
```

## Why it matters

`Integration.check_installation` (`integration.py:69`) walks `get_dependencies()` recursively, so this isn't limited to what the user asked for — **one prerelease anywhere in an integration's transitive dependencies marks the whole integration uninstalled**, with only a `logger.debug` to explain it.

Also reachable from `cli/utils.py:1055`/`:1097` (the uv path), `:1127` (pip) and `:3269` (notebook), each of which reads as "is this tool available?".

## Fix

```python
return requirement.specifier.contains(dist.version, prereleases=True)
```

Explicit prerelease pins (`pkg==1.0.0rc1`) already worked and are unaffected — `packaging` enables prereleases automatically when the specifier itself names one. This only changes the cases that were answering the wrong question.

## Testing

Added `test_requirement_installed_accepts_prereleases`, parametrized over a bare requirement with an rc, a bare requirement with a beta, and a constrained requirement with an rc. `requirement_installed` had no coverage in `tests/unit/utils/test_package_utils.py` at all.

Verified all three fail first and pass after. `tests/unit/utils/` — 200 passed (excluding `test_visualization_utils.py`, which needs IPython that I don't have locally). `ruff check`/`format --check` and `mypy src/zenml/utils/package_utils.py` clean — the one `D100` in the test file is already on `develop`, which I checked against a clean tree rather than assuming.

## Release notes label

Can't add labels as an external contributor — reads like `no-release-notes` to me, but happy to defer.

---

This change was AI-assisted (Claude). I reproduced the behaviour myself against a real install, traced the `check_installation` path by reading it rather than taking it on faith, confirmed no open PR touches this function (#5085 touches the same file but not `requirement_installed`), and ran the tests and lint baselines above myself.

### DIFF
--- a/src/zenml/utils/package_utils.py
+++ b/src/zenml/utils/package_utils.py
@@ -78,7 +78,7 @@ def requirement_installed(requirement: Union[str, Requirement]) -> bool:
     except PackageNotFoundError:
         return False
 
-    return requirement.specifier.contains(dist.version)
+    return requirement.specifier.contains(dist.version, prereleases=True)
 
 
 def get_dependencies(
--- a/tests/unit/utils/test_package_utils.py
+++ b/tests/unit/utils/test_package_utils.py
@@ -17,6 +17,7 @@ from zenml import __version__ as current_zenml_version
 from zenml.utils.package_utils import (
     clean_requirements,
     get_package_information,
+    requirement_installed,
 )
 
 
@@ -82,3 +83,23 @@ def test_clean_requirements_mixed_types():
 def test_get_package_information_works():
     """Test that the package information is returned."""
     assert get_package_information()["zenml"] == current_zenml_version
+
+
+@pytest.mark.parametrize(
+    "requirement, installed_version",
+    [
+        ("package", "1.0.0rc1"),
+        ("package", "2.0.0b2"),
+        ("package>=1.0", "1.5.0rc1"),
+    ],
+)
+def test_requirement_installed_accepts_prereleases(
+    requirement: str, installed_version: str, mocker
+) -> None:
+    """Check that an installed prerelease counts as installed."""
+    mocker.patch(
+        "zenml.utils.package_utils.distribution",
+        return_value=mocker.Mock(version=installed_version),
+    )
+
+    assert requirement_installed(requirement) is True


schustmi (approved):

chuenchen309:
Two red checks here that I can't clear from a fork, in case they're noise to you or worth a look:

- **`run-slow-ci-label-is-set`** — "Please add the 'run-slow-ci' label to this PR before merging." External contributors can't set labels (`AddLabelsToLabelable` → `does not have the correct permissions`), so this needs a maintainer either way. Same for the required `release-notes`/`no-release-notes` label I mentioned in the description.
- **`pr-labeler`** — `HttpError: Resource not accessible by integration`. `pr_labeler.yml` runs on `pull_request: [opened]` and writes labels with `secrets.GITHUB_TOKEN`, which is read-only for a PR from a fork, so it can't succeed on any external PR. Not specific to this one — flagging it since a permanently-red check on every outside contribution seemed worth knowing about rather than assuming you already do.

Neither relates to the change itself; the unit tests are green.

chuenchen309:
Correction to the above — there's a third red check I missed; it went red the same minute I posted, so "two red checks" was wrong.

- **`Profile ZenML CLI Performance`** — fails at the `Checkout current branch` step with `error: pathspec 'fix/requirement-installed-prereleases' did not match any file(s) known to git`. `performance-profiling.yml:75` does `git checkout ${GITHUB_HEAD_REF}` against the upstream checkout, and a fork's branch doesn't exist there — so, like `pr-labeler`, it can't pass on any external PR. It never reached the profiling steps, so there's no performance signal in it either way.

Still nothing touching the change itself. Happy to leave all three alone — just didn't want to leave a miscount standing.

